PR #903, opened 2018-06-05
[RW-782] Gate homepage -> workspaces on data access, if enforceRegistered

calbach:
Demo (manually hacked with `hasDataAccess()` always returning `false`):
https://calbach-dot-all-of-us-workbench-test.appspot.com/

`enforceRegistered` is not `true` anywhere currently, though I would like to move to requiring it in staging/stable per RW-712.

These pages are changing, so the UI details I think are not too critical. In general, I do think we need a more directed initial profile setup flow than what we currently have, so ultimately this data access check might manifest differently. The main thing is just getting the plumbing and logic setup here. 

Also, update the button styling so the disabled homepage buttons don't look clickable.

### DIFF
--- a/ui/src/app/views/homepage/component.ts
+++ b/ui/src/app/views/homepage/component.ts
@@ -1,11 +1,13 @@
 import {Component, OnDestroy, OnInit, ViewChild} from '@angular/core';
 import {ActivatedRoute, Router} from '@angular/router';
 import {ProfileStorageService} from 'app/services/profile-storage.service';
+import {ServerConfigService} from 'app/services/server-config.service';
 import {BugReportComponent} from 'app/views/bug-report/component';
 
 import {
   BillingProjectStatus,
   BlockscoreIdVerificationStatus,
+  DataAccessLevel,
   Profile,
   ProfileService
 } from 'generated';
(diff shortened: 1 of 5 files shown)


ui/src/app/views/homepage/component.ts
```diff
@@ -66,9 +75,7 @@ export class HomepageComponent implements OnInit, OnDestroy {
           this.profileStorageService.reload();
         }, 10000);
       }
-      if (profile != null) {
-        this.FCAccountInitialized = true;
-      }
+      this.FCAccountInitialized = true;
```
blrubenstein: Can we just remove this variable entirely and rely only on hasCDRAccess? Basically, as it stands right now, when will this ever be false?

ui/src/app/views/homepage/component.ts
```diff
@@ -127,13 +134,27 @@ export class HomepageComponent implements OnInit, OnDestroy {
...
+
+  listWorkspaces(): void {
+   this.router.navigate(['workspaces']);
+  }
+
+  hasCdrAccess(): boolean {
+    if (!this.enforceRegistered) {
+      return true;
+    }
+    if (!this.profile) {
```
blrubenstein: I think we need to move this above `!this.enforceRegistered` or currently it will succeed if you don't have a profile yet, if enforce registered is disabled.

calbach: I was treating "hasCdrAccess" as orthogonal to the account initialization state so this was returning the right thing - but changed per above comment to remove FCAccountInitialized.

For reference, here are the effective predicates for the two buttons on screen:
- List workspaces: `FCAccountInitialized && (!enforceRegistered || dataAccess >= registered)`
- Add workspace: `BillingProjectReady && (!enforceRegistered || dataAccess >= registered)`

ui/src/app/views/homepage/component.ts
calbach: It's going to be false when the profile hasn't loaded yet. The home page does render before that. We still need error messaging to look at this - I switched the template to use `hasCdrAccess` as the source of truth for the list workspace page, but had to include the profile for the error messaging cases.

blrubenstein: Got it! I didn't notice that this was within that subscribe. LGTM.

ui/src/app/views/homepage/component.ts
blrubenstein: Looks great! Yeah, that looks good on the predicates.

calbach:
Note: there is an existing issue on the homepage which is that I can access the "workspaces" page via the sidebar. I'm going to leave that alone for now pending updated mocks, but we may need to address this later.

PR #904, opened 2018-06-06
[RW-705][risk=low] Adding Mandrill API

UrsaStutsman:
Obtained API key. All should be working now... But not sure about domain. See comments in DirectoryServiceImpl for details.


api/config/config_local.json
```diff
@@ -29,5 +29,8 @@
     "adminIdVerification": "stable-environment-manual-id-verification-requests@fake-research-aou.org",
     "supportGroup": "all-of-us-workbench-eng@googlegroups.com",
     "verifiedSendingAddress": "all-of-us-workbench-eng@googlegroups.com"
+  },
+  "mandrill": {
+    "apiKey": "test-api-key-should-go-here"
```
rushtong: The key should not be here. Look at `CloudStorageServiceImpl` for how we're grabbing the current mailchimp key. We should do the same thing for the mandrill key.

UrsaStutsman: kk

api/src/main/resources/mandrill_api.yaml
```diff
@@ -0,0 +1,110 @@
+#A subset of the Mandrill Api
```
calbach: Assuming you copied this file from somewhere, could you please put a link here to that location?

api/src/main/java/org/pmiops/workbench/mandrill/MandrillConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ScopedProxyMode;
+import org.springframework.web.context.annotation.RequestScope;
+
+@org.springframework.context.annotation.Configuration
+public class MandrillConfig{
+
+  private static final String MANDRILL_API_CLIENT = "mandrillApiClient";
+
+  @Bean(name=MANDRILL_API_CLIENT)
```
calbach: Does it actually fail without this qualifier? My guess is that the places where we've done this are all cargo-culted from the FirecloudConfig, where we actually need two different bindings for the same class (each with different credentials).

api/src/main/java/org/pmiops/workbench/mandrill/MandrillServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+    keyAndMessage.setKey(apiKey);
+    keyAndMessage.setMessage(email);
+    try {
+      MandrillMessageStatuses msgStatuses = mandrillApi.send(keyAndMessage);
+      log.log(Level.INFO, "Message Status: " + msgStatuses.toString());
+      return msgStatuses;
+    } catch (ApiException e){
+      log.log(Level.WARNING, "Sending email via Mandrill Failed.");
+      return null;
```
calbach: Surely this method should throw an exception if it fails to send the email

api/src/main/java/org/pmiops/workbench/mandrill/MandrillService.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Encapsulate mandrill API interaction details and provide a simple/mockable interface
+ * for internal use.
+ */
+public interface MandrillService {
+
+  MandrillMessageStatuses sendEmail(final MandrillMessage m);
+
+}
```
rushtong: What I would prefer is a more generic interface that hides the implementation details ... and then has a more specific implementation. For example, this is really just a `MailService` class. I would suggest that this interface have a `sendPasswordEmail()` method with minimal arguments, and then the impl class can do the message construction in a mandrill-specific way. This way we have a cleaner separation of concerns. The `DirectoryServiceImpl` class above should not have to know about Mandrill details to do it's primary work. As a caller of the `MailService`, the directory service should just ask it to do what it needs to do, however it wants to do it. The way this is written now (which is also what it was when sending a plain `Message` object), different callers need to have specialized knowledge that they really don't *need* to have and is better off encapsulated in the impl class.

api/src/main/java/org/pmiops/workbench/mandrill/MandrillServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Autowired
+  public MandrillServiceImpl(Provider<WorkbenchConfig> configProvider,
+                             Provider<MandrillApi> mandrillApiProvider) {
+    this.configProvider = configProvider;
+    this.mandrillApiProvider = mandrillApiProvider;
+  }
+
+  @Override
+  public MandrillMessageStatuses sendEmail(MandrillMessage email) {
```
rushtong: I'd like to see this return something more generic than `MandrillMessageStatuses`. It should never return null, and throw an exception if it fails. I don't have a good suggestion for what to return at the moment (looking at this not on a work computer right now). I'll look more into this tomorrow, but I suspect that there are different potential conditions that can be gleaned from the response that might be useful to transform into `ApiException`, or something similar.

api/src/main/java/org/pmiops/workbench/mandrill/MandrillServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+      return null;
+    }
+  }
+
+  private String readMandrillApiKey() {
+    String bucketName = configProvider.get().googleCloudStorageService.credentialsBucketName;
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+    String content = new String(storage.get(bucketName, "mandrill-keys.json").getContent());
+    JSONObject mandrillKeys = new JSONObject(content);
+    return mandrillKeys.getString("api-key");
```
rushtong: The above lines should all be in a method defined in `CloudStorageService` and not here. You should be calling that service to get the key value.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -157,21 +160,23 @@ private String randomString(){
...
-      Message msg = new MimeMessage(session);
-      msg.setFrom(new InternetAddress(workbenchConfig.admin.verifiedSendingAddress));
-      msg.addRecipient(Message.RecipientType.TO, new InternetAddress(
-        contactEmail, user.getName().getFullName()));
-      msg.setSubject("Your new All of Us Account");
-      msg.setText(messageBody);
-      mailServiceProvider.get().send(msg);
-    } catch (MessagingException | UnsupportedEncodingException e) {
-      throw new EmailException("Error sending password email", e);
+      mandrillService.sendEmail(message);
```
rushtong: The bulk of this method should be abstracted to the mail service implementation class. I also think it would be best to abstract the from email address to the configuration instead of hard-coding it here. That way they can be changed at any time just by updating the config in the credentials GCS bucket.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -11,7 +12,9 @@
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.exceptions.EmailException;
 import org.pmiops.workbench.exceptions.ExceptionUtils;
-import org.pmiops.workbench.mail.MailService;
+import org.pmiops.workbench.mandrill.model.MandrillMessage;
+import org.pmiops.workbench.mandrill.MandrillService;
+import org.pmiops.workbench.mandrill.model.RecipientAddress;
```
rushtong: There are a number of unused imports in this class.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -54,19 +57,19 @@
...
   private final Provider<WorkbenchConfig> configProvider;
-  private final Provider<MailService> mailServiceProvider;
   private final HttpTransport httpTransport;
   private final GoogleRetryHandler retryHandler;
+  private final MandrillService mandrillService;
 
   @Autowired
   public DirectoryServiceImpl(Provider<GoogleCredential> googleCredentialProvider,
-      Provider<WorkbenchConfig> configProvider, Provider<MailService> mailServiceProvider,
+      Provider<WorkbenchConfig> configProvider, MandrillService mandrillService,
```
rushtong: I prefer the `Provider` approach to service injection, but will defer to others on this point.

api/src/main/java/org/pmiops/workbench/mandrill/MandrillConfig.java
```diff
@@ -0,0 +1,28 @@
+package org.pmiops.workbench.mandrill;
+
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.mandrill.api.MandrillApi;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ScopedProxyMode;
+import org.springframework.web.context.annotation.RequestScope;
+
+@org.springframework.context.annotation.Configuration
```
rushtong: Nit: add an import and just use `@Configuration`

api/src/main/java/org/pmiops/workbench/mandrill/MandrillConfig.java
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  @Bean
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
+  public MandrillApi mandrillApi(@Qualifier(MANDRILL_API_CLIENT) ApiClient apiClient) {
+    MandrillApi api = new MandrillApi();
+    api.setApiClient(apiClient);
+    return api;
+  }
+}
```
rushtong: nit: line ending

api/src/main/java/org/pmiops/workbench/mandrill/MandrillService.java
```diff
@@ -0,0 +1,14 @@
...
+import org.pmiops.workbench.mandrill.model.MandrillMessage;
+import org.pmiops.workbench.mandrill.model.MandrillMessageStatuses;
+
+/**
+ * Encapsulate mandrill API interaction details and provide a simple/mockable interface
+ * for internal use.
+ */
+public interface MandrillService {
+
+  MandrillMessageStatuses sendEmail(final MandrillMessage m);
```
rushtong: This can be a void method. I don't think we need to return anything.

api/src/main/java/org/pmiops/workbench/mandrill/MandrillServiceImpl.java
rushtong: Looking at this with fresh eyes, I think this can be a void method as mentioned above.

api/src/main/java/org/pmiops/workbench/mandrill/MandrillServiceImpl.java
rushtong: Additionally, let's look at abstracting this a little further. If we expand the concept of `mandrill-keys.json` to be larger, say `mandrill-info.json` (or something better-named) and add kv-pairs for the api key, from address, and perhaps something else that might be needed in the future. In other systems I've worked on, we also needed automatic CC and BCC addresses added that were environment-specific.

api/src/main/resources/mandrill_api.yaml
```diff
@@ -0,0 +1,110 @@
...
+#the part we need
+
+swagger: '2.0'
+info:
+  title: Mandrill
+  version: "0.1"
```
rushtong: nit: Use single quotes throughout this file for consistency.

api/src/main/resources/mandrill_api.yaml
```diff
@@ -0,0 +1,110 @@
...
+paths:
+  /messages/send.json:
+    post:
+      tags:
+        - mandrill
+      operationId: send
+      summary: Send an email via mandrill
+      responses:
+        200:
+          description: "Email sent successfully"
```
rushtong: nit: Omit the use of quotes when not necessary.

api/src/main/resources/mandrill_api.yaml
```diff
@@ -0,0 +1,110 @@
...
+      from_email:
+        type: string
+        description: the from address
+      to:
+        type: array
+        items:
+          "$ref": "#/definitions/RecipientAddress"
+
+  RecipientAddress:
+    description: ''
```
rushtong: Can we get away without this if it's empty (and for others below)? There is an online swagger validator that might help us clean this up (https://swagger.io/tools/swagger-editor/)

api/src/main/java/org/pmiops/workbench/mandrill/MandrillService.java
```diff
@@ -0,0 +1,14 @@
+package org.pmiops.workbench.mandrill;
+
+import org.pmiops.workbench.mandrill.model.MandrillMessage;
+import org.pmiops.workbench.mandrill.model.MandrillMessageStatuses;
+
+/**
+ * Encapsulate mandrill API interaction details and provide a simple/mockable interface
+ * for internal use.
+ */
+public interface MandrillService {
```
rushtong: I don't think we actually need a new service for this that's mandrill-specific. All of this code could be added to the current `MailService` and `MailServiceImpl` classes.

api/src/main/resources/mandrill_api.yaml
UrsaStutsman: Actually didn't

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
UrsaStutsman: So should I store the email in the GCS bucket? Or in our configs?
As mentioned in the comments I actually just did this as a place holder, and I wasn't sure if we'd want different emails from different environments (helpful to devs?)
Also - I actually thought of this whole method as a bit of a placeholder to be better implemented when we actually plug in the REAL intro email (note this is still just the quick dirty email I just swapped out the google mail service for mandrill....

rushtong: We 100% need a `sendPasswordEmail` method. It should look a lot like what it does here, but instead be in the mail sending class instead of this class.

rushtong: I'm suggesting that we store the value of `message.setFromEmail("donotreply@allofus.org")` in the configuration instead of hard coding it here. That way, it is no longer a placeholder and is a real thing.

api/src/main/java/org/pmiops/workbench/google/CloudStorageService.java
```diff
@@ -11,6 +11,7 @@
 
   public String readInvitationKey();
   public String readBlockscoreApiKey();
+  public String readMandrillApiKey();
```
rushtong: Not a nit with this code since it is following convention, but with our interfaces in general. Most methods on interfaces (abstract, default, and static) are public by default so they don't need the public modifier.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -32,6 +34,16 @@ public String readBlockscoreApiKey() {
     return readToString(getCredentialsBucketName(), "blockscore-api-key.txt").trim();
   }
 
+  public String readMandrillApiKey() {
+    JSONObject mandrillKeys = new JSONObject(readToString(getCredentialsBucketName(), "mandrill-keys.json").trim());
+    return mandrillKeys.getString("api-key");
+  }
+
+  public String readMandrillFromEmail() {
```
rushtong: I've been thinking more about this and I'm less convinced that this is the right approach even though it allows us to change the address on the fly. I am now leaning toward moving this specific configuration to each of the config json files in `api/config`. It's the right thing conceptually, so let's try that.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -54,19 +48,19 @@
 
   private final Provider<GoogleCredential> googleCredentialProvider;
   private final Provider<WorkbenchConfig> configProvider;
-  private final Provider<MailService> mailServiceProvider;
   private final HttpTransport httpTransport;
   private final GoogleRetryHandler retryHandler;
+  private final Provider<MailService> mailServiceProvider;
```
rushtong: This can be reverted along with the deletion of line 57 to keep the # of changes smaller.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -54,19 +48,19 @@
...
   @Autowired
   public DirectoryServiceImpl(Provider<GoogleCredential> googleCredentialProvider,
       Provider<WorkbenchConfig> configProvider, Provider<MailService> mailServiceProvider,
       HttpTransport httpTransport, GoogleRetryHandler retryHandler) {
     this.googleCredentialProvider = googleCredentialProvider;
     this.configProvider = configProvider;
-    this.mailServiceProvider = mailServiceProvider;
     this.httpTransport = httpTransport;
     this.retryHandler = retryHandler;
+    this.mailServiceProvider = mailServiceProvider;
```
rushtong: Ditto - revert this move to keep the changes down.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -32,6 +34,16 @@ public String readBlockscoreApiKey() {
     return readToString(getCredentialsBucketName(), "blockscore-api-key.txt").trim();
   }
 
+  public String readMandrillApiKey() {
+    JSONObject mandrillKeys = new JSONObject(readToString(getCredentialsBucketName(), "mandrill-keys.json").trim());
```
rushtong: One thing I'd like to see us do is also try to refactor when possible. For example, look at all of these `readToString` calls and note that they all use `.trim()` at the end. Instead of following the same duplicative pattern, we should try to clean these up. In this case, if we updated that method to be
```
  String readToString(String bucketName, String objectPath) {
    Storage storage = StorageOptions.getDefaultInstance().getService();
    return new String(storage.get(bucketName, objectPath).getContent()).trim();
  }
```
we could remove all the extra `trim()`s in each of the calls to `readToString`. It's a small change, but I'd like us to get into the habit of incremental and continual improvement.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -3,31 +3,25 @@
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.http.HttpTransport;
+import com.google.api.gax.rpc.ApiException;
```
rushtong: Unused.

api/src/main/java/org/pmiops/workbench/mail/MailService.java
```diff
@@ -3,8 +3,13 @@
...
 
+import com.google.api.services.admin.directory.model.User;
+import org.pmiops.workbench.mandrill.model.MandrillMessage;
+import org.pmiops.workbench.mandrill.model.MandrillMessageStatuses;
+
 public interface MailService {
 
     void send(Message msg) throws MessagingException;
 
+    MandrillMessageStatuses sendEmail(String contactEmail, String password, User user) throws MessagingException;
```
rushtong: I think this is too-generic a method name. Under the hood, it sends out what will be the welcome to AoU email. There are a couple ways of doing this. One way (and what we do more often in this codebase) is to make the method name very reflective of what it does, i.e. `sendWelcomeEmail`. 

Another approach is to bundle up the arguments into a DTO object that has more implementation-specific information. Something like
```
    public enum Type {WELCOME}
    class MessageType {
        String contactEmail;
        String password; 
        User user;
        Type type;
    }
```
And then the implementation of `sendEmail` can look at the type and do something more type specific. As more email types are defined, we can implement them one at a time.

rushtong: I don't understand why we're returning `MandrillMessageStatuses` since it isn't used anywhere. If we were using it for some reason, I could see returning it. As the PR stands now, it's just cruft.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,5 +1,17 @@
...
+import org.pmiops.workbench.mandrill.ApiException;
+import org.pmiops.workbench.mandrill.model.MandrillApiKeyAndMessage;
+import org.pmiops.workbench.mandrill.model.MandrillMessage;
+import org.pmiops.workbench.mandrill.model.MandrillMessageStatuses;
+import org.pmiops.workbench.mandrill.model.RecipientAddress;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import javax.inject.Provider;
+import java.util.Arrays;
 import org.springframework.stereotype.Service;
```
rushtong: This line is duplicated on line 12

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +21,46 @@
...
+        this.mandrillApiProvider = mandrillApiProvider;
+        this.cloudStorageServiceProvider = cloudStorageServiceProvider;
+    }
+
     public void send(Message msg) throws MessagingException {
         Transport.send(msg);
     }
 
+    public MandrillMessageStatuses sendEmail(String contactEmail, String password, User user) throws MessagingException {
+        MandrillApi mandrillApi = mandrillApiProvider.get();
```
rushtong: Nit: `mandrillApi` is only used once, it can be in-lined on line 46

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +21,46 @@
...
+        String apiKey = cloudStorageServiceProvider.get().readMandrillApiKey();
+        MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+        keyAndMessage.setKey(apiKey);
+        MandrillMessage msg = buildMandrillMessage(contactEmail, password, user);
+        keyAndMessage.setMessage(msg);
+        try {
+            MandrillMessageStatuses msgStatuses = mandrillApi.send(keyAndMessage);
+            return msgStatuses;
```
rushtong: Nit: this can just be `return mandrillApi.send(keyAndMessage);` if we end up keeping the return (see my comment about that elsewhere).

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -126,10 +120,11 @@ public User createUser(String givenName, String familyName, String username, Str
...
       .setChangePasswordAtNextLogin(true);
     retryHandler.run((context) -> getGoogleDirectoryService().users().insert(user).execute());
-    //This is a temporary solution until Mandrill is implemented.
-    //This way if one merge happens and the other is delayed for any reason,
-    //we won't break in a major way.
-    sendPasswordEmail(contactEmail, password, user);
+    try {
+      mailServiceProvider.get().sendEmail(contactEmail, password, user);
+    } catch (MessagingException e) {
+      return user;
```
rushtong: Why are we swallowing the exception here?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +21,46 @@
...
+        } catch (ApiException e){
+            throw new MessagingException("Sending email failed.");
+        }
+    }
+
+    private MandrillMessage buildMandrillMessage(String contactEmail, String password, User user) {
```
rushtong: Similar to the comment about the generic-ness of the interface's method name, I think this, too, is too generic. It should be very clear that this method is building a welcome to AoU email message.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +20,47 @@
...
+        } catch (ApiException e){
+            throw new MessagingException("Sending email failed.");
+        }
+    }
+
+    private MandrillMessage buildWelcomeMessage(String contactEmail, String password, User user) {
+        MandrillMessage msg = new MandrillMessage();
+        RecipientAddress toAddress = new RecipientAddress();
+        toAddress.setEmail(contactEmail);
+        msg.setTo(Arrays.asList(toAddress));
```
rushtong: `Arrays.asList` -> `Collections.singletonList`

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +20,47 @@
...
     }
 
+    public void sendWelcomeEmail(String contactEmail, String password, User user) throws MessagingException {
+        String apiKey = cloudStorageServiceProvider.get().readMandrillApiKey();
+        MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+        keyAndMessage.setKey(apiKey);
+        MandrillMessage msg = buildWelcomeMessage(contactEmail, password, user);
+        keyAndMessage.setMessage(msg);
+        try {
+            mandrillApiProvider.get().send(keyAndMessage);
```
rushtong: I do have a question here. Is it possible for `MandrillMessageStatuses` to indicate some kind of error with a successful call to the `send()` method? If so, we should inspect the results and throw an error if there is one.

rushtong: Clarification. There are actually 2 error cases here. The API can throw an exception, which is handled in this try block. 

The second error case is more subtle. If the response from `mandrillApiProvider.get().send()` indicates that the message was not actually sent, then we need to log that and throw an exception as well. We don't need to ever return the response from `mandrillApiProvider.get().send()`, but we do need to inspect it and deal with it if there was a sending problem.

rushtong:
Unfortunately, we have zero tests for the mail service features. I think the complexity around sending mandrill emails requires that we mock out a few different scenarios - Success, an API exception, and the various conditions of `MandrillMessageStatus` such as status and reject reason.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -126,10 +123,12 @@ public User createUser(String givenName, String familyName, String username, Str
...
       .setChangePasswordAtNextLogin(true);
     retryHandler.run((context) -> getGoogleDirectoryService().users().insert(user).execute());
-    //This is a temporary solution until Mandrill is implemented.
-    //This way if one merge happens and the other is delayed for any reason,
-    //we won't break in a major way.
-    sendPasswordEmail(contactEmail, password, user);
+    try {
+      mailServiceProvider.get().sendWelcomeEmail(contactEmail, password, user);
+    } catch (MessagingException e) {
+      log.log(Level.SEVERE, "Welcome Email not sent: " + e.toString());
```
rushtong: I'm wondering if we need to log a bit more info here. In this case, we've swallowed the password so it will never be seen again. We also don't know who the user is, unless it's part of the message exception. In the case where we've lost the PW, I think that's a manageable situation because an admin can reset that in the gsuite interface. We still need to know the user so we can then follow up on that from a bug reporting standpoint.

Let's add some more info about the user that tried to register here. At a minimum, we need the contact email and the gsuite username somewhere in the message. When the exception eventually makes its way into a jira ticket, then we should have everything we need to follow up.

api/src/test/java/org/pmiops/workbench/mail/MailServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+  private static final String FAMILY_NAME = "Bobberson";
+  private static final String CONTACT_EMAIL = "bob@example.com";
+  private static final String PASSWORD = "secretpassword";
+  private static final String PRIMARY_EMAIL = "bob@researchallofus.org";
+  private static final String API_KEY = "this-is-an-api-key";
+
+
+  @Mock
+  private CloudStorageServiceImpl cloudStorageService;
+
```
rushtong: Minor nit: The convention in our tests is to have no line breaks between the mocks.

api/src/test/java/org/pmiops/workbench/mail/MailServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+    when(mandrillApiProvider.get()).thenReturn(mandrillApi);
+    when(cloudStorageServiceProvider.get()).thenReturn(cloudStorageService);
+    when(workbenchConfigProvider.get()).thenReturn(workbenchConfig);
+    service = new MailServiceImpl(Providers.of(mandrillApi),
+      Providers.of(cloudStorageService), Providers.of(workbenchConfig));
+    msgStatuses = new MandrillMessageStatuses();
+    msgStatuses.add(msgStatus);
+  }
+
+  @NotNull
```
rushtong: Not necessary

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -126,10 +123,13 @@ public User createUser(String givenName, String familyName, String username, Str
...
-    //This is a temporary solution until Mandrill is implemented.
-    //This way if one merge happens and the other is delayed for any reason,
-    //we won't break in a major way.
-    sendPasswordEmail(contactEmail, password, user);
+    try {
+      mailServiceProvider.get().sendWelcomeEmail(contactEmail, password, user);
+    } catch (MessagingException e) {
+      log.log(Level.SEVERE, String.format(
+        "Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, username, e.toString()));
+      throw new WorkbenchException(e);
```
danqrodney: So let's say Mandrill is down temporarily, and this request fails. The user is now in a bit of a pickle; they can't log in with their new account without having received an email with their credentials.

One way of mitigating this is doing retries here or inside sendWelcomeEmail, with some backoff. Another option (more work, probably not worth it now?) is to have the email be sent in a background task asynchronously that can be retried, so that longer outages of Mandrill can still be recovered from.

Anyway, we may want to do something here.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +22,54 @@
...
+        }
+    }
+
+    private MandrillMessage buildWelcomeMessage(String contactEmail, String password, User user) {
+        MandrillMessage msg = new MandrillMessage();
+        RecipientAddress toAddress = new RecipientAddress();
+        toAddress.setEmail(contactEmail);
+        msg.setTo(Collections.singletonList(toAddress));
+        String msgBody = "Your new account is: " + user.getPrimaryEmail() +
+          "\nThe password for your new account is: " + password;
```
danqrodney: I think ideally the message body would contain a link to go log into workbench?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +22,54 @@
...
     }
 
+    public void sendWelcomeEmail(String contactEmail, String password, User user) throws MessagingException {
+        String apiKey = cloudStorageServiceProvider.get().readMandrillApiKey();
+        MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+        keyAndMessage.setKey(apiKey);
+        MandrillMessage msg = buildWelcomeMessage(contactEmail, password, user);
+        keyAndMessage.setMessage(msg);
+        try {
+            MandrillMessageStatuses msgStatuses = mandrillApiProvider.get().send(keyAndMessage);
```
danqrodney: I assume this works locally?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +22,54 @@
...
+            for (MandrillMessageStatus msgStatus : msgStatuses) {
+                if (msgStatus.getRejectReason() != null) {
+                  throw new MessagingException(msgStatus.getRejectReason());
```
danqrodney: What are the possible reject reasons?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +22,54 @@
...
+                }
+            }
+        } catch (ApiException e) {
+            throw new MessagingException("Sending email failed.");
+        } catch (MessagingException e) {
+            throw new MessagingException("Sending email failed with message: " + e.getMessage());
```
danqrodney: Does the message contain the reject reason here?

Why are we catching and rethrowing in this case?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -9,8 +22,54 @@
...
+        keyAndMessage.setMessage(msg);
+        try {
+            MandrillMessageStatuses msgStatuses = mandrillApiProvider.get().send(keyAndMessage);
+            for (MandrillMessageStatus msgStatus : msgStatuses) {
+                if (msgStatus.getRejectReason() != null) {
+                  throw new MessagingException(msgStatus.getRejectReason());
+                }
+            }
+        } catch (ApiException e) {
+            throw new MessagingException("Sending email failed.");
```
danqrodney: Should we pass e as a second parameter here? If not, seems like we should log e here as the cause.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
UrsaStutsman: "hard-bounce", "soft-bounce", "spam", "unsub", "custom", "invalid-sender", "invalid", "test-mode-limit", "unsigned", or "rule". But given these are strings set by Mandrill, figured it was best not to "hard code" these in. But in the JSON response from them there's a field "reject_reason" that if it was rejected, the reason is here. Otherwise it's null.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
UrsaStutsman: This will be taken care of by Bru's PR that has the to spec welcome email.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
UrsaStutsman: Yup.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
UrsaStutsman: Going to add in some retries. Also going to move logging into MailService. Still up for conversation on how to handle longer outages...

api/config/config_stable.json
```diff
@@ -30,5 +30,8 @@
     "adminIdVerification": "stable-environment-manual-id-verification-requests@fake-research-aou.org",
     "supportGroup": "all-of-us-workbench-eng@googlegroups.com",
     "verifiedSendingAddress": "all-of-us-workbench-eng@googlegroups.com"
+  },
+  "mandrill": {
+    "fromEmail": "donotreply@fake-research-aou.org"
```
blrubenstein: I think we have been using `stable.fake-research-aou.org` for stable (and similarly for staging).

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
blrubenstein: Do we want to start using that here? Or as a followup? (Probably makes sense to do a followup)

UrsaStutsman: Think it will be easier merge wise to merge this first and then plug in your email. (Plus actually have to do the other hooking up covered by 707.)

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -126,10 +120,11 @@ public User createUser(String givenName, String familyName, String username, Str
       .setEmails(new UserEmail().setType("custom").setAddress(contactEmail).setCustomType("contact"))
       .setChangePasswordAtNextLogin(true);
     retryHandler.run((context) -> getGoogleDirectoryService().users().insert(user).execute());
-    //This is a temporary solution until Mandrill is implemented.
-    //This way if one merge happens and the other is delayed for any reason,
-    //we won't break in a major way.
-    sendPasswordEmail(contactEmail, password, user);
+    try {
+      mailServiceProvider.get().sendWelcomeEmail(contactEmail, password, user);
```
danqrodney: We may want to at some point add a field on the user object indicating the time at which this email was successfully sent.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+    public void sendWelcomeEmail(String contactEmail, String password, User user) throws MessagingException {
+        String apiKey = cloudStorageServiceProvider.get().readMandrillApiKey();
+        int retries = workbenchConfigProvider.get().mandrill.sendRetries;
+        MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+        keyAndMessage.setKey(apiKey);
+        MandrillMessage msg = buildWelcomeMessage(contactEmail, password, user);
+        keyAndMessage.setMessage(msg);
+        retry: do {
+            retries--;
+            log.log(Level.INFO, "retries: " + retries);
```
danqrodney: You can probably remove this?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+                    log.log(Level.WARNING, String.format(
+                      "ApiException: Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                    if (retries == 0) {
+                        log.log(Level.SEVERE, String.format(
+                          "ApiException: On Last Attempt! Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                        throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
+                    }
+                    break;
+
+                case REJECTED:
```
danqrodney: So "REJECTED" here means that a retry is bound to fail?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+                      "Messaging Exception: Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                    throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
+
+                case SUCCESSFUL:
+                    log.log(Level.INFO, String.format(
+                      "Welcome Email to '%s' for user '%s' was sent.", contactEmail, user.getName()));
+                    break retry;
+
+                default:
+                  log.log(Level.SEVERE, String.format(
```
danqrodney: If we don't expect this to ever happen, probably best to just throw a ServerErrorException() here.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+            log.log(Level.INFO, "trying send");
+            MandrillMessageStatuses msgStatuses = mandrillApiProvider.get().send(keyAndMessage);
+            for (MandrillMessageStatus msgStatus : msgStatuses) {
+                if (msgStatus.getRejectReason() != null) {
+                    return new ImmutablePair<String, String>("REJECTED", msgStatus.getRejectReason());
+                }
+            }
+        } catch (Exception e) {
+            log.log(Level.INFO, "in catch");
+            return new ImmutablePair<>("API_ERROR", e.toString());
```
danqrodney: Do we need to distinguish between different kinds of errors / exceptions here?

api/src/main/resources/mandrill_api.yaml
danqrodney: OK. A link to their API docs would be useful in any case.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+
+                case REJECTED:
+                    log.log(Level.SEVERE, String.format(
+                      "Messaging Exception: Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                    throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
+
+                case SUCCESSFUL:
+                    log.log(Level.INFO, String.format(
+                      "Welcome Email to '%s' for user '%s' was sent.", contactEmail, user.getName()));
+                    break retry;
```
danqrodney: You could probably just return here?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+            ImmutablePair attempt = trySend(keyAndMessage);
+            Status status = Status.valueOf(attempt.getLeft().toString());
+            switch (status) {
+                case API_ERROR:
+                    log.log(Level.WARNING, String.format(
+                      "ApiException: Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                    if (retries == 0) {
+                        log.log(Level.SEVERE, String.format(
+                          "ApiException: On Last Attempt! Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+                        throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
```
danqrodney: Why do we throw an exception in this case and the REJECTED case but not in the default case?

FWIW: I don't think we should return a 500 when we get back REJECTED here... as the input is valid, the user just gave an email address that we couldn't send stuff to. Seems better to just have the create user response indicate if there was a problem sending the email but return a 200?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: Agreed, or at least make it debug.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: If I read this right, in the API error case, we don't actually know what went wrong here. This is an error from the api client talking to the mandrill server for some reason. In this case, I do think we have to throw an exception. How that is returned to the user is abstracted to, in this case, the `DirectoryServiceImpl` class.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: Will defer to @UrsaStutsman on this, but I think it is true that rejected will always fail. This means that mandrill didn't like what we asked it to send, or didn't like our api key, or our license is not valid ... etc.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: Agreed.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: Agreed, but we do still need to log this so we have some way of following up via an error reporting process on this failed attempt.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: I would use the real `Status` here instead of doing string manipulation in other places.

api/src/main/resources/mandrill_api.yaml
rushtong: This is their current doc for this API call: https://mandrillapp.com/api/docs/messages.JSON.html

api/src/test/java/org/pmiops/workbench/mail/MailServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+    MandrillMessageStatuses msgStatuses = new MandrillMessageStatuses();
+    msgStatuses.add(msgStatus);
+    when(mandrillApi.send(any())).thenReturn(msgStatuses);
+    when(cloudStorageService.readMandrillApiKey()).thenReturn(API_KEY);
+
+    service = new MailServiceImpl(Providers.of(mandrillApi), Providers.of(cloudStorageService),
+      Providers.of(createWorkbenchConfig()));
+  }
+
+  private User createUser() {
```
rushtong: Nit: relocate this method above or below the other private method in this class.

rushtong:
General note: it looks like your IDE is set to 4 space indents. This project uses 2 space indents.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,120 @@
...
+        return msg;
+    }
+
+    private ImmutablePair<String, String> trySend(MandrillApiKeyAndMessage keyAndMessage) {
+        try {
+            log.log(Level.INFO, "trying send");
+            MandrillMessageStatuses msgStatuses = mandrillApiProvider.get().send(keyAndMessage);
+            for (MandrillMessageStatus msgStatus : msgStatuses) {
+                if (msgStatus.getRejectReason() != null) {
+                    return new ImmutablePair<String, String>("REJECTED", msgStatus.getRejectReason());
```
rushtong: `ImmutablePair<String, String>` -> `ImmutablePair<>`

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
UrsaStutsman: Yes - "REJECTED" means we communicated with Mandrill, but Mandrill rejected it. This can be for any number of reasons, this is why I'm logging the reject_reason from mandrill. Common reasons, email gets a "hard_bounce", ie does not exist, something is wrong with our api key, etc...

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,118 @@
...
+            "Messaging Exception: Welcome Email to '%s' for user '%s' not sent: %s", contactEmail, user.getName(), attempt.getRight().toString()));
+          throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
+
+        case SUCCESSFUL:
+          log.log(Level.INFO, String.format(
+            "Welcome Email to '%s' for user '%s' was sent.", contactEmail, user.getName()));
+          return;
+
+        default:
+          log.log(Level.SEVERE, String.format(
```
rushtong: I could have sworn I added a comment on this, but it seems to be lost. I think this needs to check the retries count and only log and fail if the count is at zero.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,118 @@
...
+      "\nThe password for your new account is: " + password;
+    msg.setHtml(msgBody);
+    msg.setSubject("Your new All of Us Account");
+    msg.setFromEmail(workbenchConfigProvider.get().mandrill.fromEmail);
+    return msg;
+  }
+
+  private ImmutablePair<Status, String> trySend(MandrillApiKeyAndMessage keyAndMessage) {
+    try {
+      log.log(Level.INFO, "trying send");
```
rushtong: Looks like debugging code

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,118 @@
...
+      MandrillMessageStatuses msgStatuses = mandrillApiProvider.get().send(keyAndMessage);
+      for (MandrillMessageStatus msgStatus : msgStatuses) {
+        if (msgStatus.getRejectReason() != null) {
+          return new ImmutablePair<>(Status.REJECTED, msgStatus.getRejectReason());
+        }
+      }
+    } catch (Exception e) {
+      log.log(Level.INFO, "in catch");
```
rushtong: Looks like debugging code

api/src/main/java/org/pmiops/workbench/mail/MailService.java
```diff
@@ -3,8 +3,11 @@
 import javax.mail.Message;
 import javax.mail.MessagingException;
 
+import com.google.api.services.admin.directory.model.User;
+
 public interface MailService {
 
     void send(Message msg) throws MessagingException;
 
+    void sendWelcomeEmail(String contactEmail, String password, User user) throws MessagingException;
```
rushtong: I think this file is using 4 space indents. Pretty sure that was my fault.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,118 @@
...
+
+        default:
+          log.log(Level.SEVERE, String.format(
+            "Welcome Email to '%s' for user '%s' was not sent. Default case.", contactEmail, user.getName()));
+          throw new MessagingException("Sending email failed: " + attempt.getRight().toString());
+      }
+    } while (retries > 0);
+  }
+
+  private MandrillMessage buildWelcomeMessage(String contactEmail, String password, User user) {
```
rushtong: This is a new comment, and I should have seen this earlier, sorry. I think we should check the validity of the email address here and throw an invalid argument exception if it is not valid.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -1,16 +1,118 @@
...
+    int retries = workbenchConfigProvider.get().mandrill.sendRetries;
+    MandrillApiKeyAndMessage keyAndMessage = new MandrillApiKeyAndMessage();
+    keyAndMessage.setKey(apiKey);
+    MandrillMessage msg = buildWelcomeMessage(contactEmail, password, user);
+    keyAndMessage.setMessage(msg);
+    do {
+      retries--;
+      ImmutablePair attempt = trySend(keyAndMessage);
+      Status status = Status.valueOf(attempt.getLeft().toString());
+//            Status status = Status.valueOf(attempt.getLeft().toString());
```
rushtong: Remove

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
rushtong: And, if we do add email validation, we should add a test method that exercises this condition as well.

PR #905, opened 2018-06-06
[RW-817][risk=no] PR Template

rushtong:
## Addresses 
https://precisionmedicineinitiative.atlassian.net/browse/RW-817

## Reviewers
* Is this correct?
* Is this too long (I have heard that it could be)
* Is it useful?

My goals here are to try and put some gentle guardrails around the process that might help reduce errors.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,9 +6,9 @@ Have you read README.md lately? If not, do that first.
 ---
 
 - [ ] Review code comments; remove done TODOs, create stories for remaining TODOs
+- [ ] Optionally link to the Jira Ticket for the convenience of the reviewer
 - [ ] Run all tests, both **unit** and **integration** tests
 - [ ] Preface the PR title with the Jira Ticket Number and security risk, i.e., "[RW-123][risk=no] My New Feature"
-- [ ] Optionally link to the Jira Ticket for the convenience of the reviewer
   - Risk factors: 
     - **no**: None 
     - **low**: TODO: Need a description of all risk factors
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-Teplace this this template with your PR description.
+Replace this this template with your PR description.
 Please remember to keep in mind the security levels outlined in [CONTRIBUTING.md](CONTRIBUTING.md) and to 
 include a risk tag of the form `[risk=no|low|moderate|severe]` in the PR title
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,26 +1,12 @@
 
 <your comments for this PR go here>
 
-Have you read README.md lately? If not, do that first.
+Have you read [CONTRIBUTING.md](CONTRIBUTING.md) lately? If not, do that first.
 
 ---
 
-- [ ] Review code comments; remove done TODOs, create stories for remaining TODOs
-- [ ] Optionally link to the Jira Ticket for the convenience of the reviewer
 - [ ] Run all tests, both **unit** and **integration** tests
-- [ ] Preface the PR title with the Jira Ticket Number and security risk, i.e., "[RW-123][risk=no] My New Feature"
-  - Risk factors: 
-    - **no**: None 
-    - **low**: TODO: Need a description of all risk factors
-    - **moderate**: TODO: Need a description of all risk factors
-    - **severe**: TODO: Need a description of all risk factors
-- [ ] Add description or comments on the PR explaining the hows/whys (if not obvious)
 - [ ] Request at least one reviewer, preferably two.
-- [ ] Initial review by LR and others.
-- [ ] Comment / review / update cycle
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+
+- [ ] Review code comments; remove done TODOs, create stories for remaining TODOs
+- [ ] Optionally link to the Jira Ticket for the convenience of the reviewer
+- [ ] Run all tests, both **unit** and **integration** tests
+- [ ] Preface the PR title with the Jira Ticket Number and security risk, i.e., "[RW-123][risk=no] My New Feature"
+  - Risk factors: 
+    - **no**: None 
+    - **low**: TODO: Need a description of all risk factors
+    - **moderate**: TODO: Need a description of all risk factors
+    - **severe**: TODO: Need a description of all risk factors
```
rushtong: I'm looking for suggestions on where to find this information.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
...
+<your comments for this PR go here>
+
+Have you read README.md lately? If not, do that first.
+
+---
+
+- [ ] Review code comments; remove done TODOs, create stories for remaining TODOs
+- [ ] Optionally link to the Jira Ticket for the convenience of the reviewer
+- [ ] Run all tests, both **unit** and **integration** tests
+- [ ] Preface the PR title with the Jira Ticket Number and security risk, i.e., "[RW-123][risk=no] My New Feature"
```
blrubenstein: My suggestion was going to be: (to even further minimize the no risk case.)
[]
[risk = low]
[risk = moderate]
[risk = severe]

calbach:
Thanks @rushtong  - my preference is that process documentation like this go in `CONTRIBUTING.md`. I've been hit with this PR template before on some of the Broad repo's, and in my opinion this is too much content to digest, let alone check off per PR. This is especially daunting for someone who is unfamiliar with the codebase/process and just wants to fix a typo in a README, for example.

Most of the bullets here are recommendations and general process things. These are very important to document for first-time contributors, but I think are overkill to check off for every PR. My preference would be that from this list, the only thing to keep would be the note about including the Jira ticket / "risk" prefix, since that's looking for a specific PR title syntax (and a link out to CONTRIBUTING.md). Everything else I think we should enforce with automation and the review process.

blrubenstein:
I agree with @calbach. I actually looked at this and thought it was a very long list of things, having not read through it yet. (And, having not actually worked in the FC code base, I also find that PR checklist highly intimidating.)

rushtong:
Thanks All - great feedback. Will migrate this to a `CONTRIBUTING.md` approach.

rushtong:
@blrubenstein @calbach - Please take another look, thank you!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+Please remember to keep in mind the security levels outlined in [CONTRIBUTING.md](CONTRIBUTING.md)
+
+* **no**: None 
+* **low**: Low chance of potential impact to, or exposure of patient data
+* **moderate**: Moderate chance of potential impact to, or exposure of patient data
+* **severe**: Severe chance of potential impact to, or exposure of patient data
+
+And to:
+
+* Get thumbs from all reviewers
```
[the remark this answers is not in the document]
rushtong: True, I think I might be very wrong here. I think our current policy might be a single reviewer. That feels a little light to me though, I would prefer a minimum of 2, but will defer to the group.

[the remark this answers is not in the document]
calbach: I'm definitely more accustomed to a requirement of 1 reviewer - though I expect it to be common-place to have multiple reviewers where it makes sense, e.g. a PR that touches areas with two different subject matter experts. My position is require 1 approval, and leave the rest to judgement.

I think if we're seeing siloing of reviewers, or not-very-thorough reviews, we need to address that as a team culture issue. There's certainly a spectrum of change significance - on the low end 2 reviewers is overkill and will slow us down (e.g. https://github.com/all-of-us/workbench/pull/944). Where in other cases 2 or more may be appropriate (e.g. this PR).

On a related note: I think marking only a single reviewer is generally ideal for code review. Assuming clear communication of when the author has finished addressing comments (a problem in GitHub IMO - another discussion), there is a always a clear owner for the next step in the code review process. With multiple reviewers, there's a bystander effect, or alternatively duplicated review work in the event of parallel reviews. The next best thing is marking multiple reviewers but assigning them disjoint review tasks e.g. "adding Dan for API changes, Bru for UI changes".

Exception: for trivial changes (1-liners), it's common and reasonable to simply shotgun over many reviewers to maximize visibility.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+Teplace this this template with your PR description.
```
blrubenstein: Nit: Typo?

rushtong: :-) Why yes, yes it is.

calbach:
Thanks for putting this together. LGTM

PR #906, opened 2018-06-06
[RW-777] [risk=no] Fix nav bar routing

blrubenstein:


blrubenstein:
Yeah, we just changed the routing to include an s, and missed some of the spots where it is used...

PR #908, opened 2018-06-07
[RW-712][risk=no] Enforce registered in staging

calbach:
This will have implications for the release QA process. This will also force us to start thinking about some new UX issues in the first-sign-in flow. Timing is not urgent if we think this will cause significant short-term pain before some of the account creation flow changes are in.

### DIFF
--- a/api/config/config_staging.json
+++ b/api/config/config_staging.json
@@ -3,7 +3,7 @@
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-staging-",
     "debugEndpoints": false,
-    "enforceRegistered": false,
+    "enforceRegistered": true,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-rw-staging-scripts/setup_notebook_cluster.sh",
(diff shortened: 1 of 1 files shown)


blrubenstein:
I'm actually most worried about shared accounts with this. Because I believe ID verification is a gating factor on registered access, and it would be nice to have a shared account that has the ID verification review authority. 

Other than that, I think contact email verification is the other thing that needs to go in, because we don't currently have a way to do that with our mailchimp deprecation. We should probably remove that need until we have a process in place for that.

calbach:
AFAIK I don't have any access to any "shared accounts". I've already put instructions in the playbook for doing manual ID verification, though it will be a bit of a pain initially.

Won't mailchimp keep working until we switch to the new thing? That said, I'm fine to wait on that.

calbach:
Ah, I just realized we have the temporary password via email thing setup now, and I think @rushtong mentioned work on the email verification status - so I think I'll wait on that before continuing with this PR.

calbach:
Alright, lets try this out. Since I'm scheduled to do the next release, I will be the one to suffer the initial QA impact.

PR #909, opened 2018-06-07
[RW-809] [Risk=low] Blrubenstein/retry errored clusters

blrubenstein:


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -110,6 +114,20 @@ public void setUserProvider(Provider<User> userProvider) {
       fcCluster = this.notebooksService.createCluster(
           project, NotebooksService.DEFAULT_CLUSTER_NAME, userProvider.get().getEmail());
     }
+    if (fcCluster.getStatus() != null && fcCluster.getStatus().equals(org.pmiops.workbench.notebooks.model.ClusterStatus.ERROR)) {
```
calbach: Can avoid the null check by reordering this: `ClusterStatus.ERROR.equals(fcCluster.getStatus());`, here and below

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -3,8 +3,7 @@
 import static com.google.common.truth.Truth.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
calbach: Per slack discussion, please avoid `*` imports. (we should turn on a linter rule for this)

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -254,4 +268,21 @@ public void testLocalize_noNotebooks() throws Exception {
...
+  }
+
+  @Test
+  public void testClusterRetriesThreeTimes() throws Exception {
+    when(notebookService.getCluster(WORKSPACE_NS, "all-of-us")).thenThrow(new NotFoundException());
+    when(notebookService.createCluster(eq(WORKSPACE_NS), eq("all-of-us"), any()))
+        .thenReturn(testFcClusterErrored);
+
+    Cluster defaultCluster = clusterController.listClusters().getBody().getDefaultCluster();
+    defaultCluster = clusterController.listClusters().getBody().getDefaultCluster();
```
calbach: for loop?

api/db/changelog/db.changelog-39-add-cluster-retry-number.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="brubenst" id="changelog-39-add-cluster-retry-number">
+    <addColumn tableName="user">
+      <column name="cluster_retry_number" type="INTEGER" />
```
calbach: naming nit: cluster_create_retries? The "_number" suffix doesn't feel quite right to me, possibly the redundancy with the column type, or slight ambiguity since "number" doesn't necessarily == count.

Including "create" would also be nice in case we need to retry anything else cluster related in the future.

blrubenstein:
@calbach should be good for another round of reviews!

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -110,6 +114,20 @@ public void setUserProvider(Provider<User> userProvider) {
       fcCluster = this.notebooksService.createCluster(
           project, NotebooksService.DEFAULT_CLUSTER_NAME, userProvider.get().getEmail());
     }
+    if (org.pmiops.workbench.notebooks.model.ClusterStatus.ERROR.equals(fcCluster.getStatus())) {
+      User user = this.userProvider.get();
```
calbach: Actually - should update this logic to break if we exceed 2 retries. Currently this would just keep incrementing cluster retries indefinitely on every poll.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -110,6 +114,20 @@ public void setUserProvider(Provider<User> userProvider) {
...
+      user.setClusterCreateRetries(user.getClusterCreateRetries() + 1);
+      this.userDao.save(user);
+      if (user.getClusterCreateRetries() <= 2) {
+        this.notebooksService.deleteCluster(project, NotebooksService.DEFAULT_CLUSTER_NAME);
```
calbach: I think it would make sense to log a warning when we do this delete along with the retry count. If you're feeling ambitious, you could also log `cluster.getErrors()`, which might give some insight into what happened.

PR #912, opened 2018-06-08
[RW-805][risk=no] Mail Service Refactor

rushtong:
## Addresses
Partly addresses https://precisionmedicineinitiative.atlassian.net/browse/RW-805
There is more work to do, so this PR does not complete 805. 

## Changes
* Send all mail via a mail service class instead of directly.
* Update integration and unit tests


blrubenstein:
Nit: Open bracket in PR title.

rushtong:
Thanks, fixed PR title.

api/src/test/java/org/pmiops/workbench/api/BugReportControllerTest.java
```diff
@@ -87,13 +87,18 @@ User user() {
...
+  private List<Message> sentMessages = new ArrayList<>();
 
   @Before
-  public void setUp() {
-    gaeHelper.setUp();
+  public void setUp() throws MessagingException {
+    sentMessages.clear();
+    Mockito.doAnswer(
+      invocation -> {
+        sentMessages.add(invocation.getArgumentAt(0, Message.class));
```
calbach: Note: I don't feel very strongly about this specific code, since the details here are changing to use Jira anyways.

In general, I'd prefer fakes to mocks. Usually it exercises more real codepaths and it avoids complex in-test setup to simulate behavior, like this argument capture. That said, the gae test helper is a lot to digest, and not super similar to our other test patterns, so I'm also OK with this mocking.

api/src/test/java/org/pmiops/workbench/api/BugReportControllerTest.java
```diff
@@ -87,13 +87,18 @@ User user() {
...
+        return null;
+      }
+    ).when(mailService).send(Mockito.any());
```
calbach: `any()` is statically imported.

api/src/test/java/org/pmiops/workbench/api/BugReportControllerTest.java
rushtong: Thanks for the explanation - it's appreciated. I primarily wanted to preserve the original intention of ensuring that the attachment tests, but didn't really consider the value of fakes vs mocks.

calbach: opt: since with this change we would no longer depend on the GAE local testing stuff, you could consider dropping these three gradle deps that I had added: https://github.com/all-of-us/workbench/blob/master/api/build.gradle#L235

PR #914, opened 2018-06-08
[RW-805][risk=no] Set user email verification in the right place

rushtong:
## Addresses
Partly addresses https://precisionmedicineinitiative.atlassian.net/browse/RW-805
There is more work to do, so this PR does not complete 805.
Relates to https://github.com/all-of-us/workbench/pull/912

## Changes
* Set the user's email verification status in the right place. We no longer need to use mailchimp to know if the user's email is verified now.
* Remove outdated test

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -33,7 +33,6 @@ import org.pmiops.workbench.exceptions.ForbiddenException;
 import org.pmiops.workbench.exceptions.GatewayTimeoutException;
 import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.exceptions.WorkbenchException;
-import org.pmiops.workbench.firecloud.ApiException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.firecloud.model.BillingProjectMembership.CreationStatusEnum;
 import org.pmiops.workbench.google.CloudStorageService;
@@ -236,6 +235,8 @@ public class ProfileController implements ProfileApiDelegate {
       }
 
       user.setFirstSignInTime(new Timestamp(clock.instant().toEpochMilli()));
+      // If the user is logged in, then we know that they have followed the account creation instructions sent to
+      // their initial contact email address.
       user.setEmailVerificationStatus(EmailVerificationStatus.SUBSCRIBED);
       try {
         return userDao.save(user);
@@ -314,7 +315,7 @@ public class ProfileController implements ProfileApiDelegate {
   }
 
   private ResponseEntity<Profile> getProfileResponse(User user) {
-    return ResponseEntity.ok(profileService.getProfile(user, /* checkEmailVerification */ true));
+    return ResponseEntity.ok(profileService.getProfile(user));
   }
 
   @Override
@@ -519,7 +520,7 @@ public class ProfileController implements ProfileApiDelegate {
     IdVerificationListResponse response = new IdVerificationListResponse();
     List<Profile> responseList = new ArrayList<Profile>();
     for (User user : userService.getNonVerifiedUsers()) {
-      responseList.add(profileService.getProfile(user, /* checkEmailVerification */ false));
+      responseList.add(profileService.getProfile(user));
     }
     response.setProfileList(responseList);
     return ResponseEntity.ok(response);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -236,6 +236,7 @@ public class ProfileController implements ProfileApiDelegate {
       }
 
       user.setFirstSignInTime(new Timestamp(clock.instant().toEpochMilli()));
+      user.setEmailVerificationStatus(EmailVerificationStatus.SUBSCRIBED);
       try {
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -98,27 +95,6 @@ public Profile getProfile(User user, boolean checkEmailVerification) {
     profile.setInstitutionalAffiliations(user.getInstitutionalAffiliations()
         .stream().map(TO_CLIENT_INSTITUTIONAL_AFFILIATION)
         .collect(Collectors.toList()));
-    EmailVerificationStatus userEmailVerificationStatus = user.getEmailVerificationStatus();
-    // if verification is pending or unverified, need to query MailChimp and update DB accordingly
-    // Note: This is pretty slow, we shouldn't be calling it on every Profile lookup. Leave it
-    // for now as we're in the process of switching off Mailchimp (related: RW-705).
-    if (checkEmailVerification &&
```
calbach: This parameter `checkEmailVerification` is no longer used, please remove.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -236,6 +236,7 @@ private User initializeUserIfNeeded() {
       }
 
       user.setFirstSignInTime(new Timestamp(clock.instant().toEpochMilli()));
+      user.setEmailVerificationStatus(EmailVerificationStatus.SUBSCRIBED);
```
calbach: nit: Perhaps a comment here to explain why signing implies email verification.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -98,27 +95,6 @@ public Profile getProfile(User user, boolean checkEmailVerification) {
...
-        !userEmailVerificationStatus.equals(EmailVerificationStatus.SUBSCRIBED)) {
-      if (userEmailVerificationStatus.equals(EmailVerificationStatus.UNVERIFIED) && user.getContactEmail() != null) {
-        try {
-          mailChimpService.addUserContactEmail(user.getContactEmail());
```
calbach: `mailChimpService` I believe is now unused by this service

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -467,7 +468,6 @@ private void verifyInvitationKey(String invitationKey){
     if (updatedProfile.getContactEmail() != null) {
       if (!updatedProfile.getContactEmail().equals(user.getContactEmail())) {
         mailChimpService.addUserContactEmail(updatedProfile.getContactEmail());
```
calbach: Is this still needed? Are we still putting researchers on the mailchimp list?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -467,7 +468,6 @@ private void verifyInvitationKey(String invitationKey){
...
-        user.setEmailVerificationStatus(EmailVerificationStatus.PENDING);
```
calbach: Shouldn't changing your email "unverify" your email? Possibly just a TODO for now, but I think it still needs to be handled somehow.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
rushtong: I have a large-ish PR for mailchimp removal coming and was planning on doing that there. Can do it in this one though.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: No, and I plan on removing it in a separate PR.

calbach: OK - sounds good

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: I believe this requires more refinement. We have a story that addresses changing contact email and I propose we define requirements for that story.

calbach: Sounds good.

rushtong: For reference, the contact email change story is https://precisionmedicineinitiative.atlassian.net/browse/RW-719

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
rushtong: Done.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
rushtong: Done

calbach:
FYI, the codacy rules are two that I've proposed we disable. I can do that right now if you agree:

https://docs.google.com/document/d/1o1nN4Oni9v6wgoYeFhG2yj47pqaD7XkVrRaTHqsdHhg/edit?hl=en#

rushtong:
I do agree with those rule changes, thank you. I was purposefully ignoring the failing check, but had meant to comment in the PR about it.

PR #915, opened 2018-06-08
[RW-745][risk=no] First pass at Codacy Java style fixes

calbach:
More style change proposals found here: https://docs.google.com/document/d/1o1nN4Oni9v6wgoYeFhG2yj47pqaD7XkVrRaTHqsdHhg/edit


calbach:
Fair point. To clarify, fully qualifying both is not incompatible with this codacy style pattern - the alternative fix would be to remove the import (importing AND fully qualifying should be an error).

In general in the codebase I've seen the convention of "import the API model, fully qualify the DB model" in controllers. This is my preference mainly to avoid verbosity but also once you've got the convention, it's easy enough to scan the short one vs the long one. Fully qualifying both is explicit, but I'm not sure about crystal clear - as I find it very hard to scan for what's usually a small delta of "db." in the package name (at least, that's the case for my IDE). Of course this approach is also more verbose as well.

PR #919, opened 2018-06-08
[RW-745][risk=no] Misc Codacy cleanups (Docker, CSS)

calbach:


api/src/dev/server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine
+FROM alpine:3.7
```
blrubenstein: I especially appreciate pinning the version here! Thank you CH!

rushtong: +1, not doing this has bitten me before.

PR #922, opened 2018-06-11
[RW-800][risk=no] added new criteria tree for PM.

freemabd1:


freemabd1:
This table data wasn't generated. I created it manually... but this should be included in the Jon S. scripts to generate the criteria table/counts

PR #923, opened 2018-06-11
[DO NOT MERGE] [RW-707][risk=low] updating account success page

UrsaStutsman:


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -445,6 +446,39 @@ private void verifyInvitationKey(String invitationKey){
...
+    }
+  }
+
+  private void updateUser(UpdateContactEmailRequest updateRequest) {
+    com.google.api.services.admin.directory.model.User googleUser =
+      directoryService.getUser(updateRequest.getUsername());
+    googleUser.setPrimaryEmail(updateRequest.getContactEmail());
+
+    User user = userDao.findUserByEmail(updateRequest.getUsername());
+    user.setContactEmail(updateRequest.getContactEmail());
```
rushtong: Do we want to verify the email address before saving it? In theory, this could bubble back up to the UI letting the user know they tried to update their previous (and potentially invalid) email with another, invalid email.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -445,6 +448,45 @@ private void verifyInvitationKey(String invitationKey){
...
+        "This account has already been in use, if you would like to update your contact email please login and update via the Profile page."
+      );
+    }
+  }
+
+  private void updateUser(UpdateContactEmailRequest updateRequest) {
+    com.google.api.services.admin.directory.model.User googleUser =
+      directoryService.getUser(updateRequest.getUsername());
+    String contactEmail = updateRequest.getContactEmail();
+//    try {
```
UrsaStutsman: I commented this out while I was working on the other issues. But we need this, or something like it to validate updated email addresses...

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -445,6 +448,45 @@ private void verifyInvitationKey(String invitationKey){
     }
   }
 
+  @Override
+  public ResponseEntity<Void> updateContactEmail(UpdateContactEmailRequest updateContactEmailRequest) {
+    verifyNoLogins(updateContactEmailRequest.getUsername());
+    //resend Welcome Email to new contact email... Add once Mandrill API complete
```
UrsaStutsman: was planning on using the resendWelcomeEmail method from below here (once I had that working the way I wanted)

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -445,6 +448,45 @@ private void verifyInvitationKey(String invitationKey){
...
+    com.google.api.services.admin.directory.model.User googleUser =
+      directoryService.getUser(updateRequest.getUsername());
+    String contactEmail = updateRequest.getContactEmail();
+//    try {
+//      InternetAddress email = new InternetAddress(contactEmail);
+//      email.validate();
+//    } catch (AddressException e) {
+//      throw new MessagingException("Email: " + contactEmail + " is invalid.");
+//    }
+    googleUser.setPrimaryEmail(contactEmail);
```
UrsaStutsman: This is actually WRONG! we want to set the CONTACT email on the google User. (If we go that direction). The Primary Email is the username@domain

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -35,14 +36,16 @@
   private static SecureRandom rnd = new SecureRandom();
 
   // This list must exactly match the scopes allowed via the GSuite Domain Admin page here:
-  // https://admin.google.com/AdminHome?chromeless=1#OGX:ManageOauthClients
+  // https://admin.google.com/fake-research-aou.org/AdminHome?chromeless=1#OGX:ManageOauthClients
```
UrsaStutsman: I set the particular SA referenced in this file to the new scopes on this page. We may also need to update other SA's and/or other domains for these scopes as well....

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -113,20 +118,42 @@ public boolean isUsernameTaken(String username) {
...
     String gSuiteDomain = configProvider.get().googleDirectoryService.gSuiteDomain;
     String password = randomString();
+    Alias alias = new Alias();
+    alias.setAlias(contactEmail);
     User user = new User()
       .setPrimaryEmail(username+"@"+gSuiteDomain)
       .setPassword(password)
       .setName(new UserName().setGivenName(givenName).setFamilyName(familyName))
-      .setEmails(new UserEmail().setType("custom").setAddress(contactEmail).setCustomType("contact"))
+//      .setAliases(Arrays.asList(contactEmail))
```
UrsaStutsman: This is where I'm currently struggling. I've now tried multiple ways to set the custom email field ("contact") to the user's contact email. But can NOT seem to figure it out. 
I think the alternative is to just not set this in the google account, and only use our database for that info, but not loving that approach.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -113,20 +118,42 @@ public boolean isUsernameTaken(String username) {
...
+
+  @Override
+  public void resendWelcomeEmail(String userKey) {
+    User user = getUser(userKey);
+    String password = randomString();
+    user.setPassword(password);
+//    retryHandler.run((context) -> getGoogleDirectoryService().users().update(userKey, user).execute());
+    Aliases emails = retryHandler.run((context) -> getGoogleDirectoryService().users().aliases().list(userKey).execute());
+    log.log(Level.INFO, "user: " + user.toString());
+    log.log(Level.INFO, "emails: " + emails.getAliases().toString());
```
UrsaStutsman: This is clearly still in the "figuring it out stage". The commented line above here does need to be uncommented once everything else is figured out.

rushtong:
Closing in favor of #967

PR #924, opened 2018-06-11
Bumping to pyclient-v1-11

danqrodney:


danqrodney:
@calbach is the suggestion that if the startup script fails, Leo should continue trying to set things up anyway?

(That makes sense to me... just clarifying.)

calbach:
I think it's OK for Leo to crash on startup in this case (though the error could have been clearer). We always have the option to suppress errors ourselves within the startup script if we want to. I was actually more wondering how we had a package version that failed to pip install, and whether we have a test in place to catch this going forward.

danqrodney:
Got it.

We do not have a circle test that either runs our startup script or tries to install our pyclient lib. We should probably remedy that.

PR #925, opened 2018-06-11
Commenting out pyclient install

danqrodney:


danqrodney:
Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-830 to track uncommenting this logic in future.

PR #926, opened 2018-06-12
[RW-805][risk=no] Remove mailchimp dependency

rushtong:
## Addresses
Along with https://github.com/all-of-us/workbench/pull/912 and https://github.com/all-of-us/workbench/pull/914, this PR completes https://precisionmedicineinitiative.atlassian.net/browse/RW-805

## Changes
Remove mailchimp related services, configurations, and tests.


rushtong:
@calbach The Codacy failure here is `' should be "` which I was not able to change or uncheck in Codacy for some reason, perhaps permissions. We use this check in our `tslint.json` 
```
    "quotemark": [
      true,
      "single"
    ],
```
So we do need to make this change somewhere.

calbach:
Mentioned in Slack yesterday but it probably got buried - unfortunately we still have this zombie Codacy project causing problems so we'll have to accept Codacy red'ness a while longer. I've identified the owner (@dmohs), though I suspect he's in another all-day meeting today.

rushtong:
I can dig into the codacy ownership too. That should be moved back to our realm.

calbach:
To clarify, there are multiple Codacy projects currently configured. We own one and have access to it, the "pmi-ops-org" project - this is the one we want to use. The "all-of-us" project is also integrated, and AFAICT the only person with access is @dmohs. We need to delete this Codacy project so it stops posting statuses to our PRs, as it has a divergent configuration from pmi-ops-org, and no one has access to update the style rules on it. There appears to be no way of disabling that integration from the GitHub side (killing the webhook was not sufficient).

I'll CC you on my email to dmohs, there shouldn't be anything else to dig into here.

PR #927, opened 2018-06-12
[RW-809] [Risk=no] Fix cluster reset.

blrubenstein:
This resets the retry count if you reset your cluster.


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -149,6 +149,9 @@ public void setUserProvider(Provider<User> userProvider) {
 
   @Override
   public ResponseEntity<EmptyResponse> deleteCluster(String projectName, String clusterName) {
+    User user = this.userProvider.get();
+    user.setClusterCreateRetries(0);
+    this.userDao.save(user);
```
danqrodney: Nit: it's probably better to add a method to UserService to do this, and use updateWithRetries, e.g.

https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java#L145

That way if there's another thread saving the user at the same time you don't get an exception here.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -20,10 +20,8 @@
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
-import org.pmiops.workbench.db.dao.CdrVersionDao;
-import org.pmiops.workbench.db.dao.UserDao;
-import org.pmiops.workbench.db.dao.UserService;
-import org.pmiops.workbench.db.dao.WorkspaceService;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.db.dao.*;
```
calbach: Please revert (and turn off `*` imports in your IDE) per past slack discussion.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -122,6 +126,9 @@ public void setUp() {
     when(userProvider.get()).thenReturn(user);
     clusterController.setUserProvider(userProvider);
 
+    UserService userService = new UserService(userProvider, userDao, adminActionHistoryDao, clock, fireCloudService, configProvider);
```
calbach: Yeah, this is probably the best you can do here for now. The ugliness stems from me not being able to figure out how to inject a provider with a stubbable user (it can probably be done). So the controller would initially be created using the mock `UserService` defined via `@MockBean` above, then you replace it here.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: Right, yes! This was mentioned before... I thought I'd turned it off, but will investigate,

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: Yeah, exactly. That's the thought at least.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: Reverted... sorry about that.

PR #928, opened 2018-06-12
[RW-767][risk=low] Implement a notebook cluster cleanup cron

calbach:
- Update Leo APIs
- Factor out service account access token creation
- Add an endpoint which cleans up running notebook clusters
- Add configuration for the maximum allowed lifetime for a cluster

A follow up change will add the cronjob configuration for this.

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -3,6 +3,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-test-f1-",
+    "clusterMaxAgeDays": 10,
     "debugEndpoints": false,
     "enforceRegistered": false,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-workbench-test-scripts/setup_notebook_cluster.sh",
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -3,7 +3,8 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-test-f1-",
-    "clusterMaxAgeDays": 10,
+    "clusterMaxAgeDays": 7,
+    "clusterIdleMaxAgeDays": 3,
     "debugEndpoints": false,
     "enforceRegistered": false,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-workbench-test-scripts/setup_notebook_cluster.sh",
--- a/api/config/config_stable.json
+++ b/api/config/config_stable.json
@@ -2,6 +2,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-stable-",
+    "clusterMaxAgeDays": 14,
     "debugEndpoints": false,
     "enforceRegistered": false,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-rw-stable-scripts/setup_notebook_cluster.sh",
--- a/api/config/config_stable.json
+++ b/api/config/config_stable.json
@@ -3,6 +3,7 @@
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-stable-",
     "clusterMaxAgeDays": 14,
+    "clusterIdleMaxAgeDays": 7,
     "debugEndpoints": false,
     "enforceRegistered": false,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-rw-stable-scripts/setup_notebook_cluster.sh",
--- a/api/config/config_staging.json
+++ b/api/config/config_staging.json
@@ -2,7 +2,8 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-staging-",
-    "clusterMaxAgeDays": 5,
+    "clusterMaxAgeDays": 7,
+    "clusterIdleMaxAgeDays": 3,
     "debugEndpoints": false,
     "enforceRegistered": true,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-rw-staging-scripts/setup_notebook_cluster.sh",
--- a/api/config/config_staging.json
+++ b/api/config/config_staging.json
@@ -2,6 +2,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-staging-",
+    "clusterMaxAgeDays": 5,
     "debugEndpoints": false,
     "enforceRegistered": true,
     "jupyterUserScriptUri": "gs:\/\/all-of-us-rw-staging-scripts/setup_notebook_cluster.sh",
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
@@ -1,0 +1,139 @@
+package org.pmiops.workbench.api;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.inject.Provider;
+
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.exceptions.ExceptionUtils;
+import org.pmiops.workbench.exceptions.ServerErrorException;
+import org.pmiops.workbench.model.CheckClustersResponse;
+import org.pmiops.workbench.notebooks.ApiException;
+import org.pmiops.workbench.notebooks.NotebooksConfig;
+import org.pmiops.workbench.notebooks.api.ClusterApi;
+import org.pmiops.workbench.notebooks.model.Cluster;
+import org.pmiops.workbench.notebooks.model.ClusterStatus;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+
+/**
+ * Offline cluster API. Handles cronjobs for cleanup and upgrade of older
+ * clusters. Methods here should be access restricted as, unlike
+ * ClusterController, these endpoints run as the Workbench service account.
+ */
+@RestController
+public class OfflineClusterController implements OfflineClusterApiDelegate {
+  private static final Logger log = Logger.getLogger(OfflineClusterController.class.getName());
+
+  private final Provider<ClusterApi> clusterApiProvider;
+  private final Provider<WorkbenchConfig> configProvider;
+  private final Clock clock;
+
+  @Autowired
+  OfflineClusterController(
+      @Qualifier(NotebooksConfig.SERVICE_CLUSTER_API) Provider<ClusterApi> clusterApiProvider,
+      Provider<WorkbenchConfig> configProvider,
+      Clock clock) {
+    this.clock = clock;
+    this.configProvider = configProvider;
+    this.clusterApiProvider = clusterApiProvider;
+  }
+
+  /**
+   * checkClusters deletes older clusters in order to force an upgrade on the
+   * next researcher login. This method is meant to be restricted to invocation
+   * by App Engine cron.
+   *
+   * The cluster deletion policy here aims to strike a balance between enforcing
+   * upgrades, cost savings, and minimizing user disruption. To this point, our
+   * goal is to only upgrade idle clusters if possible, as doing so gives us an
+   * assurance that a researcher is not actively using it. We delete clusters in
+   * the following cases:
+   *
+   * 1. It exceeds the max cluster age. Per environment, but O(weeks).
+   * 2. Since a cluster was last accessed, its spent over half its remaining
+   *    lifespan idle (as defined by [1]). This dynamic limit means that as the
+   *    upgrade horizon approaches, we'll be more aggressive in upgrading.
+   */
+  @Override
+  public ResponseEntity<CheckClustersResponse> checkClusters() {
+    WorkbenchConfig config = configProvider.get();
+    Instant now = clock.instant();
+    Duration maxAge = Duration.ofDays(config.firecloud.clusterMaxAgeDays);
+
+    ClusterApi clusterApi = clusterApiProvider.get();
+    List<Cluster> clusters;
+    try {
+      clusters = clusterApi.listClusters(null, false);
+    } catch (ApiException e) {
+      throw ExceptionUtils.convertNotebookException(e);
+    }
+
+    int errors = 0;
+    int activeDeletes = 0;
+    int unusedDeletes = 0;
+    for (Cluster c : clusters) {
+      String clusterId = c.getGoogleProject() + "/" + c.getClusterName();
+      if (ClusterStatus.UNKNOWN.equals(c.getStatus()) || c.getStatus() == null) {
+        log.warning(String.format("unknown cluster status for cluster '%s'", clusterId));
+        continue;
+      }
+      if (!ClusterStatus.RUNNING.equals(c.getStatus()) &&
+          !ClusterStatus.STOPPED.equals(c.getStatus())) {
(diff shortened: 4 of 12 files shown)


api/config/config_local.json
```diff
@@ -3,6 +3,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-test-f1-",
+    "clusterMaxAgeDays": 10,
```
danqrodney: Could probably be 5 like test? Any reason to make this 10?

api/config/config_stable.json
```diff
@@ -2,6 +2,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-stable-",
+    "clusterMaxAgeDays": 14,
```
danqrodney: Is this the value you'd expect to use in prod, too?

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+   * by App Engine cron.
+   *
+   * The cluster deletion policy here aims to strike a balance between enforcing
+   * upgrades, cost savings, and minimizing user disruption. To this point, our
+   * goal is to only upgrade idle clusters if possible, as doing so gives us an
+   * assurance that a researcher is not actively using it. We delete clusters in
+   * the following cases:
+   *
+   * 1. It exceeds the max cluster age. Per environment, but O(weeks).
+   * 2. Since a cluster was last accessed, its spent over half its remaining
```
danqrodney: This sorta makes sense. It's a bit more complicated to explain to researchers (and to grok for ourselves) than "your cluster will be deleted after N days once it is idle". Given that we have auto-pause, is it important to take idleness into account here?

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+  @Override
+  public ResponseEntity<CheckClustersResponse> checkClusters() {
+    WorkbenchConfig config = configProvider.get();
+    Instant now = clock.instant();
+    Duration maxAge = Duration.ofDays(config.firecloud.clusterMaxAgeDays);
+
+    ClusterApi clusterApi = clusterApiProvider.get();
+    List<Cluster> clusters;
+    try {
+      clusters = clusterApi.listClusters(null, false);
```
danqrodney: Is there any pagination or filtering available here? 

Given that we're deleting unused clusters after a while, I wouldn't expect this list to grow big enough for us to worry... just checking.

Might be good to log something here about how many clusters are currently running.

api/src/main/resources/notebooks.yaml
```diff
@@ -854,7 +860,7 @@ definitions:
       masterDiskSize:
         type: integer
         description: |
-          Optional, the size of the disk on the master node. Minimum size is 100GB. If unspecified, default size is 500GB.
+          Optional, the size in megabytes of the disk on the master node. Minimum size is 10GB. If unspecified, default size is 100GB.
```
danqrodney: Have we changed to only ask for 10 GB or 25 GB yet? Should we?

api/src/main/resources/workbench.yaml
```diff
@@ -680,6 +680,23 @@ paths:
...
             $ref: "#/definitions/AuditBigQueryResponse"
 
+  /v1/cron/checkClusters:
+    get:
+      security: []
+      tags:
+        - offlineCluster
+        - cron
+      description: >
+        Endpoint meant to be called offline to trigger cluster checks and
```
danqrodney: When do you plan on running this? Will that be a separate PR?

Seems like we should run it a couple times a day at least, so we don't wind up with clusters that never get upgraded since they are always in use at the time in question.

api/src/main/resources/workbench.yaml
```diff
@@ -680,6 +680,23 @@ paths:
...
+        cleanup. Enforces upgrades for older cluster deployments. May be slow to
+        execute. Only executable via App Engine cronjob.
```
danqrodney: What makes this only executable via cronjob?

Do we want the option of running this ourselves? (Not sure we need to, just checking.)

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+    int unusedDeletes = 0;
+    for (Cluster c : clusters) {
+      String clusterId = c.getGoogleProject() + "/" + c.getClusterName();
+      if (ClusterStatus.UNKNOWN.equals(c.getStatus()) || c.getStatus() == null) {
+        log.warning(String.format("unknown cluster status for cluster '%s'", clusterId));
+        continue;
+      }
+      if (!ClusterStatus.RUNNING.equals(c.getStatus()) &&
+          !ClusterStatus.STOPPED.equals(c.getStatus())) {
+        // For now, we only handle running or stopped (suspended) clusters.
```
danqrodney: What else will we handle in future?

Is there a status here for "paused" that's different from STOPPED? I think we'd want to delete clusters when they are paused...

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+        log.info(String.format(
+            "deleting cluster '%s', unused for %s, more than its remaining lifespan (>%s)",
+            clusterId, formatDuration(unusedDuration), formatDuration(lifeRemaining)));
+        unusedDeletes++;
+      } else {
+        // Don't delete.
+        continue;
+      }
+      try {
+        clusterApi.deleteCluster(c.getGoogleProject(), c.getClusterName());
```
danqrodney: There's a race condition here: we fetch the big list up top, start iterating and deleting clusters, and by the time we get to the deleteCluster call for a cluster, it's actually in use again. So we may want to add a check here before deleting a cluster to make sure it's still not running. (Even better would be to have deleteCluster provide the option to say "don't delete if I'm in use"... but I assume that doesn't exist.)

api/config/config_local.json
blrubenstein: I could also picture this as 7 or something, just to keep it on the same scale as stable. (Also, a week seems like a reasonable time?)

Either way, I second Dan that I think it should be about the same time as test. Although I see what you're going for, since development is likely happening locally, and thus we are likely to be reusing clusters. I suspect that five minutes every five days is acceptable cost though.

api/config/config_staging.json
```diff
@@ -2,6 +2,7 @@
   "firecloud": {
     "billingAccountId": "014D91-FCB792-33D2C0",
     "billingProjectPrefix": "aou-rw-staging-",
+    "clusterMaxAgeDays": 5,
```
blrubenstein: I'd actually say staging can probably be pretty small. Maybe even less than 5 days, just because its being used for QA and should be relatively one-off, correct?

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+   * assurance that a researcher is not actively using it. We delete clusters in
+   * the following cases:
+   *
+   * 1. It exceeds the max cluster age. Per environment, but O(weeks).
+   * 2. Since a cluster was last accessed, its spent over half its remaining
+   *    lifespan idle (as defined by [1]). This dynamic limit means that as the
+   *    upgrade horizon approaches, we'll be more aggressive in upgrading.
+   */
+  @Override
+  public ResponseEntity<CheckClustersResponse> checkClusters() {
```
danqrodney: This logic is complicated enough (and the consequences dire enough if we get things wrong) that we probably should add some tests (mocking out ClusterApi).

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
blrubenstein: Agreed. This seems to be most useful once auto pause is in place for enforcing cluster upgrades, and avoiding zombie clusters. I think it makes sense not to be more aggressive based on idleness.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
blrubenstein: Stopped is (I believe) the same as paused.

danqrodney: OK. 

It seems to me that the right policy is to feel free to delete clusters in a paused / STOPPED state whenever we think is reasonable, but deleting clusters when they are RUNNING just because they are somewhat old is not very nice... we may just happen to be running our cron job in the middle of your lengthy computation.

Perhaps we should have two day settings: # of days after we delete a paused cluster, and # of days after we delete a running cluster. The latter would be significantly bigger... this should give us several days of runs to attempt to kill the cluster when it was paused, instead of running.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
danqrodney: This comment was written assuming we would attempt to *not* delete clusters that were running, or at least as aggressively as we'd delete paused clusters... I misread the logic initially. See above.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,139 @@
...
+import org.pmiops.workbench.notebooks.model.Cluster;
+import org.pmiops.workbench.notebooks.model.ClusterStatus;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+
+/**
+ * Offline cluster API. Handles cronjobs for cleanup and upgrade of older
```
danqrodney: So this says:

https://stackoverflow.com/questions/22859423/combining-google-app-engine-cron-job-and-task-is-starting-google-tasks-as-cron

That there's a ten minute timeout for handling these requests. That should be sufficient unless we get a whole lot of clusters or deletes are super-slow... but either way I'd document that here. 

If you wanted to avoid that timeout you could run this as a background task (but lose the ability to return the info in the response) or I guess we could look into having backend instances for workbench (which we might need at some stage...) Probably not necessary.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
rushtong: The leo API doesn't have any paging options that I can see. It does have a way to filter on labels, but that's about it.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
rushtong: 👍 to not deleting running clusters.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -28,11 +32,10 @@
   private static final String END_USER_API_CLIENT = "endUserApiClient";
   private static final String ALL_OF_US_API_CLIENT = "allOfUsApiClient";
 
-  private static final String[] BILLING_SCOPES = new String[] {
+  private static final List<String> BILLING_SCOPES = ImmutableList.of(
```
rushtong: Nice!

api/src/main/resources/notebooks.yaml
calbach: It was just closed: https://github.com/DataBiosphere/leonardo/issues/388, not sure whether the docs were updated.

api/src/main/resources/workbench.yaml
calbach: Yes - follow-up, per PR description.

On the frequency, I'm not totally sure.

Either (1) run frequently, e.g. every 3h or (2) run this daily in the middle of the night (where people are least likely be to using notebooks)

api/src/main/resources/workbench.yaml
calbach: CronInterceptor picks up the "cron" tag above. We can trigger it manually via the UI. Not sure if there's a programmatic need for this anywhere.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
calbach: Added. Ultimately this is just mitigation as we can't do this transactionally (without Leo API support) - this is a matter of reducing the window from 10s of seconds -> 1s of seconds.

api/config/config_staging.json
calbach: I think it should be the same as in test. Feel free to argue for that value being lower/higher.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
calbach: Added to comment. I don't think we're going to come close to this, especially if we run this cron frequently.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
calbach: Also note: there is no autopausing yet, so I'm simulating this signal by looking at the lastAccessed time; later we can switch this to look at `STOPPED`

api/src/main/resources/workbench.yaml
danqrodney: I think with the separate settings, running this multiple times a day sounds good.

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,153 @@
...
+            "deleting cluster '%s', exceeded max lifetime @ %s (>%s)",
+            clusterId, formatDuration(age), formatDuration(maxAge)));
+        activeDeletes++;
+      } else if (isIdle && age.toMillis() > idleMaxAge.toMillis()) {
+        log.info(String.format(
+            "deleting cluster '%s', idle with age %s (>%s)",
+            clusterId, formatDuration(age), formatDuration(idleMaxAge)));
+        unusedDeletes++;
+      } else {
+        // Don't delete.
```
danqrodney: It might be useful to spit out how many clusters were idle in general, whether or not we deleted them. (Just as an extra piece of data.)

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -0,0 +1,153 @@
...
+      }
+      if (!ClusterStatus.RUNNING.equals(c.getStatus()) &&
+          !ClusterStatus.STOPPED.equals(c.getStatus())) {
+        // For now, we only handle running or stopped (suspended) clusters.
+        continue;
+      }
+
+      Instant created = Instant.parse(c.getCreatedDate());
+      Duration age = Duration.between(created, now);
+      Instant lastUsed = Instant.parse(c.getDateAccessed());
```
danqrodney: When exactly does the dateAccessed field get updated? If I open a notebook and keep the notebook open, does it periodically update this value even if nothing is running? Or is it only if I"m running cells? What if the kernel is doing a long-running operation and I'm not doing anything in the UI?

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
calbach: Done

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
calbach: My understanding is that it is purely based on the most recent web request coming through the proxy, i.e. any traffic to the Jupyter API from the notebooks UI. It's unclear to me whether a tab open in the background would trigger this, or just typing (but not executing), or if only executing cells would trigger it. I could also imagine this varying between browsers. Asked on the Leo doc for clarification.

Ultimately I believe this value is the same one that Leo is using to trigger pause, and eventually we'll just switch this code to look for the STOPPED status instead of this timestamp. In other words, if this date field doesn't match the behavior we want, then autopause may not match the behavior we want.

PR #929, opened 2018-06-12
[RW-831][Risk=Low] API refactor to remove blockscore

blrubenstein:


blrubenstein:
Hey @rushtong! The `BlockscoreIdVerificationStatus` object, and the `blockscore_verification_is_valid` values are both actually still being used by manual verification, which is why I didn't want to pull them out yet. I haven't looked at `blockscore_id`. I can check that.

My plan was to do a follow up for the DB changes, since I think that may involve reworking some of the logic on manual verification.

rushtong:
@blrubenstein - fair enough. How about we rename `BlockscoreIdVerificationStatus`? It doesn't make sense for that object to be so vendor specific. It really should just be `IdVerificationStatus` or something similar.

blrubenstein:
@rushtong I think this handles the DB changes. PTAL when you get the chance.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -524,7 +524,7 @@ private void verifyInvitationKey(String invitationKey){
   @AuthorityRequired({Authority.REVIEW_ID_VERIFICATION})
   public ResponseEntity<IdVerificationListResponse> reviewIdVerification(Long userId, IdVerificationReviewRequest review) {
     BlockscoreIdVerificationStatus status = review.getNewStatus();
-    Boolean oldVerification = userDao.findUserByUserId(userId).getBlockscoreVerificationIsValid();
+    Boolean oldVerification = userDao.findUserByUserId(userId).getIdVerificationIsValid();
```
rushtong: Nice!

api/src/main/java/org/pmiops/workbench/blockscore/BlockscoreService.java
```diff
@@ -1,15 +0,0 @@
...
-
-/**
- * Encapsulate API for interfacing with Blockscore identity verification service.
- */
-public interface BlockscoreService {
-  public PaginatedResult<Person> listPeople();
-  public Person createPerson(
-    String firstName, String lastName, Address address, String dobString,
-    String documentType, String documentNumber);
-}
```
rushtong: When we do re-implement IDV, we should try to keep the service interface named cleanly, like `IdentityVerificationService`, etc. Vendor-specific naming, IMHO, is confusing.

blrubenstein:
@rushtong, it now propegates through to the UI and Swagger, and all the test fixes are done. When you get the chance, please take another look. (Sorry about all the re-reviews)

PR #931, opened 2018-06-13
[risk=no] Drop cluster disk size from 100GB -> 20GB

calbach:
Just released in Leo per https://github.com/DataBiosphere/leonardo/issues/388


calbach:
Tested locally, confirmed this works.

PR #935, opened 2018-06-14
[RW-767][risk=no] Configure cluster cronjob for every 3h

calbach:


danqrodney:
You should probably let people know about this on #workbench-eng / #workbench-beta.

calbach:
Done - will make note of this to #workbench-beta along with the release next week.

PR #939, opened 2018-06-15
Srushti gangireddy/advanced concept search fix

SrushtiGangireddy:
1. Changed dialect in application.props to the MYSQLDialect
2. Changed the order of match in advancedconceptsearch
3. Changed the predicate a little
4. Added matchType and standardConcept to ConceptListResponse


common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -8,6 +8,10 @@
 
 public interface ConceptDao extends CrudRepository<Concept, Long> {
 
+    @Query(nativeQuery=true, value="select c.* from concept c join concept_relationship cr on c.concept_id=cr.concept_id_2 " +
+            "where cr.concept_id_1=?1 and cr.relationship_id='Maps to' ")
+    Concept findStandardConcept(long concept_id);
```
danqrodney: Is this definitely going to be limited to standard concept relationships? Is there guaranteed to only be one of these?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -74,47 +79,61 @@ public static String modifyMultipleMatchKeyword(String query){
...
+
+                    Expression<Double> matchExp = criteriaBuilder.function("match", Double.class,
+                            root.get("conceptName"), criteriaBuilder.parameter(String.class , "keyword"));
+                    conceptName_Filter.add(criteriaBuilder.greaterThan(matchExp, 0.0));
+
 
                     // Optionally filter on standard concept, vocabulary ID, domain ID
                     if (standardConceptFilter.equals(StandardConceptFilter.STANDARD_CONCEPTS)) {
-                        predicates.add(criteriaBuilder.equal(root.get("standardConcept"),
+                        conceptName_Filter.add(criteriaBuilder.equal(root.get("standardConcept"),
```
danqrodney: I'm wondering if we need to change this logic to include "C" here (and exclude it below). WDYT?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -74,47 +79,61 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                    List<Predicate> conceptName_Filter = new ArrayList<>();
+
+
+                    Expression<Double> matchExp = criteriaBuilder.function("match", Double.class,
+                            root.get("conceptName"), criteriaBuilder.parameter(String.class , "keyword"));
+                    conceptName_Filter.add(criteriaBuilder.greaterThan(matchExp, 0.0));
+
 
                     // Optionally filter on standard concept, vocabulary ID, domain ID
                     if (standardConceptFilter.equals(StandardConceptFilter.STANDARD_CONCEPTS)) {
```
danqrodney: I'm wondering if we should get rid of this code... given the logic where we return non-standard concepts that match code or ID unconditionally, it's not clear to me that filtering on standard vs. non-standard concepts really makes sense anymore. (It's weird to explicitly ask for standard concepts in the requests, and return non-standard ones anyway...)

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -74,47 +79,61 @@ public static String modifyMultipleMatchKeyword(String query){
...
                     } catch (NumberFormatException e) {
                         // Not a long, don't try to match it to a concept ID.
                     }
-                    predicates.add(criteriaBuilder.or(queryPredicates.toArray(new Predicate[0])));
+
+                    List<Predicate> conceptName_Filter = new ArrayList<>();
+
+
+                    Expression<Double> matchExp = criteriaBuilder.function("match", Double.class,
+                            root.get("conceptName"), criteriaBuilder.parameter(String.class , "keyword"));
```
danqrodney: Why the change to use a parameter instead of a literal?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -74,47 +79,61 @@ public static String modifyMultipleMatchKeyword(String query){
...
+
                     return criteriaBuilder.and(predicates.toArray(new Predicate[0]));
                 };
-
         // Return up to limit results, sorted in descending count value order.
         Pageable pageable = new PageRequest(0, limit,
                 new Sort(Direction.DESC, "countValue"));
         NoCountFindAllDao<Concept, Long> conceptDao = new NoCountFindAllDao<>(Concept.class,
                 entityManager);
+        conceptDao.keyword=keyword;
```
danqrodney: If we need to do this, let's pass it in to the constructor, and make the field final, instead of assigning it like this.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,40 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+        for(Concept con:matchedConcepts){
+            String conceptCode = con.getConceptCode();
+            Long conceptId = con.getConceptId();
+            if(conceptCode.equals(searchConceptsRequest.getQuery())) {
+
+                conceptCodeMatches.add(con);
+
+                response.setMatchType("ConceptCode");
+
+                Concept std_concept = conceptDao.findStandardConcept(con.getConceptId());
```
danqrodney: If the matched concept is a standard concept, we don't need to do this lookup, do we? It looks like you have that check below in the next if block...

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,40 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
                 conceptService.searchConcepts(searchConceptsRequest.getQuery(), convertedConceptFilter,
                         searchConceptsRequest.getVocabularyIds(), domainIds, maxResults);
         ConceptListResponse response = new ConceptListResponse();
-        response.setItems(concepts.getContent().stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
+        List<Concept> matchedConcepts = concepts.getContent();
+        List<Concept> conceptCodeMatches = new ArrayList<>();
+        for(Concept con:matchedConcepts){
```
danqrodney: Add a space after the colon

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,40 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+            String conceptCode = con.getConceptCode();
+            Long conceptId = con.getConceptId();
+            if(conceptCode.equals(searchConceptsRequest.getQuery())) {
```
danqrodney: It seems like we should be looking for a non-standard code in the results, rather than doing this if and the next if statement...

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,40 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+
+                conceptCodeMatches.add(con);
+
+                response.setMatchType("ConceptCode");
+
+                Concept std_concept = conceptDao.findStandardConcept(con.getConceptId());
+                response.setStandardConcept(TO_CLIENT_CONCEPT.apply(std_concept));
+
+                response.setItems(conceptCodeMatches.stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
```
danqrodney: So it's possible for us to have a query string that both matches a concept code or ID and *also* appears as a keyword in standard concepts. If that happens, what do we want the UX to be?

If we want to return the other standard concepts in these cases (which seems right to me), we shouldn't return a response here... we should just tack on a non-standard concept with the standard concept info filled out and continue. You could even do this logic as a part of TO_CLIENT_CONCEPT and leave the logic in here the way it was before.

SrushtiGangireddy: I thought if there is a match on concept code/id we are going to return the concept matched by code/id and throw away the rest of concept_name matches

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Done

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Seems nicer. Thanks

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: Do you mean to include "C" concepts along with standard concepts and not return them as the part of non standard concepts ?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
SrushtiGangireddy: I got to check those.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: Yeah. I'm not sure it makes sense for now... if I understand correctly, the "C" concepts are standard but because we haven't calculated descendant counts, they all have zero counts right now. So this is probably fine for the time being... eventually we may want to return C concepts with descendant counts.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
danqrodney: Yeah, I'm not sure whether this is the right behavior. It may be that some codes / concepts are short and show up in other things that researchers are looking for... might be better to always do the full search, even when there's a code / ID match. @biopete WDYT?

public-api/src/main/resources/public-api.yaml
```diff
@@ -329,8 +329,9 @@ definitions:
           type: string
           description: match column type
         standardConcept:
```
danqrodney: You might want to rename this "standardConcepts".

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -10,7 +10,7 @@
 
     @Query(nativeQuery=true, value="select c.* from concept c join concept_relationship cr on c.concept_id=cr.concept_id_2 " +
             "where cr.concept_id_1=?1 and cr.relationship_id='Maps to' and c.standard_concept='S' ")
-    Concept findStandardConcept(long concept_id);
+    List<Concept> findStandardConcept(long concept_id);
```
danqrodney: Rename to findStandardConcepts.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,41 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
                         searchConceptsRequest.getVocabularyIds(), domainIds, maxResults);
         ConceptListResponse response = new ConceptListResponse();
-        response.setItems(concepts.getContent().stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
+        List<Concept> matchedConcepts = concepts.getContent();
+        List<Concept> conceptCodeMatches = new ArrayList<>();
+        for(Concept con : matchedConcepts){
+            String conceptCode = con.getConceptCode();
+            Long conceptId = con.getConceptId();
+            if(!con.getStandardConcept().equals("S")){
+                if(conceptCode.equals(searchConceptsRequest.getQuery())) {
```
danqrodney: This code and the next code are identical apart from the "matchType" right? I think you can simplify this code and just have:

response.setMatchType(conceptCode.equals(searchConceptsRequest.getQuery()) ? "ConceptCode" : "ConceptId")

and get rid of the if/else.

public-api/src/main/resources/public-api.yaml
```diff
@@ -324,7 +325,14 @@ definitions:
           type: "array"
           items:
             $ref: "#/definitions/Concept"
-
+        matchType:
+          type: string
```
danqrodney: Consider making this an enum instead of a String.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +262,41 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+                    conceptCodeMatches.add(con);
+
+                    response.setMatchType("ConceptCode");
+
+                    List<Concept> std_concepts = conceptDao.findStandardConcept(con.getConceptId());
+                    response.setStandardConcept(std_concepts.stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
+
+                    response.setItems(conceptCodeMatches.stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
+
+                    return ResponseEntity.ok(response);
```
danqrodney: (As mentioned previously, I wouldn't necessarily return just the non-standard concept here, but I guess this is a product question.)

api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
```diff
@@ -213,6 +240,14 @@ public void testSearchConceptsConceptIdMatch() throws Exception {
             new SearchConceptsRequest().query("456")), CLIENT_CONCEPT_2);
   }
 
+  @Test
+  public void testSearchConceptsMatchOrder() throws Exception{
+    saveConcepts();
+    assertResults(
+            conceptsController.searchConcepts("ns", "name",
+                    new SearchConceptsRequest().query("conceptD")), CLIENT_CONCEPT_4 , CLIENT_CONCEPT_6);
```
danqrodney: I'm confused about how this currently passes... CLIENT_CONCEPT_6 is non-standard. Doesn't your code below only return a single concept in this case?

Either way, please add a test that asserts that standard concepts are returned with the non-standard one.

SrushtiGangireddy: I will add one.
The search matches on the code and fetches CLIENT_CONCEPT_4 and matches on the name an d fetches CLIENT_CONCEPT_5 and 6 irrespective of the standard concept filter because the query does not specify the filter.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: Done

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Makes sense.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -243,10 +252,21 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
-            maxResults = 25;
+        Integer maxResults = searchConceptsRequest.getMaxResults();
+        if(maxResults == null || maxResults == 0){
+            maxResults = Integer.MAX_VALUE;
         }
+
+        if(searchConceptsRequest.getQuery() == null || searchConceptsRequest.getQuery().isEmpty()){
+            List<Concept> concepts = conceptDao.findConceptsOrderedByCount();
+            ConceptListResponse response = new ConceptListResponse();
+            if(maxResults < concepts.size()){
```
biopete: I'm no java expert -- but this code looks to me like the limit should be put on the mysql query in findAllConceptsOrderedByCount.. See above comment on the conceptDao. Not sure if Java would return all 4million concepts and then filter that list.  Probably not.

public-api/src/main/resources/public-api.yaml
```diff
@@ -119,7 +120,7 @@ paths:
             $ref: "#/definitions/DbDomainListResponse"
 
   /v1/databrowser/advanced-concept-search:
```
biopete: Let's make sure this is the same api name/path as the private api

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: No, just get rid of the filter entirely.

(We'd start out by keeping it in the API but ignoring it... and then remove it from the API.)

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
danqrodney: Yeah, this should be a parameter to findConceptsOrderedByCount.

public-api/src/main/resources/public-api.yaml
danqrodney: So the workbench API actually has the workspace ID in the path (since we will eventually be using workspace metadata when querying for stuff). We could name it similarly, but the paths are going to differ.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
biopete: But we need to only get standard concepts . Are you saying just return standard concepts unless it is a code or concept_id match ?

public-api/src/main/resources/public-api.yaml
biopete: cool. I like things named the same that are the same.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: Yep, that's what I'm suggesting. We could add some option back in later when we have requirements for fetching non-standard concepts on something else.

If you prefer we can keep this filter around for future use, but then we ought to have requests that ask for standard concepts get *only* standard concepts, non-standard get only non-standard and match on everything, etc.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: Changed the name to searchConcepts now that the old concept-search is removed.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -34,6 +38,10 @@
     nativeQuery = true)
     List<Concept> findAllConceptsOrderedByCount(@Param("standard_concept") String standard_concept);
 
+    @Query(value = "select c.* from concept c order by c.count_value desc", nativeQuery = true)
```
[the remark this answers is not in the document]
SrushtiGangireddy: Added the optional max results parameter to the old method. Can fetch everything by passing in the max value in case the param is not present.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -243,10 +252,21 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
     @Override
     public ResponseEntity<ConceptListResponse> getAdvancedConceptSearch(SearchConceptsRequest searchConceptsRequest){
```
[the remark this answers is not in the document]
SrushtiGangireddy: Got rid of the old search concepts method and changed the name of advanced concept search to searchConcepts similar to the one in api.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: Renamed the field to standardConcepts.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
SrushtiGangireddy: Done.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Changed the method to return the non standard ones along with the standard ones they are mapped to and any name concept matches too.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
biopete: I just can't predict all the use cases , but to me , be able to specify the standard_concept flag you want in a concept search seems like not a bad thing to keep around.  Especially since we already wrote all the code for it.  I understand how we display concepts in hierarchies to users and such is up in the air .... Just saying -- if they say "Databrowser team : make a button to toggle all vs standard on the search results ... "   I can do that now. 

So i would say that we make the default "S" and allow you to pass "C" "null" .

If we want to clarify the api , we could make a option for search type:  "MatchCode && MatchConceptID&&MatchConceptName"   would specify the order and what fields you wanted to search on. "MatchConceptName" would just do the concept name search. MatchCode or Concept implies that the standard_concept flag is irrelevant because it is in these contexts.   I prefer that over deleting the S option .

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: Added the check to include C concepts as part of standard concepts check and not include them as part of non standard concepts.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: Right now it is (MatchCode) or (MatchId) or (MatchName and standardconceptfilter). So may be it is fine like this for now. We can use the filter as how we want.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,7 +280,25 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
                 conceptService.searchConcepts(searchConceptsRequest.getQuery(), convertedConceptFilter,
                         searchConceptsRequest.getVocabularyIds(), domainIds, maxResults);
         ConceptListResponse response = new ConceptListResponse();
-        response.setItems(concepts.getContent().stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList()));
+        List<Concept> matchedConcepts = concepts.getContent();
+        for(Concept con : matchedConcepts){
```
biopete: I don't think this logic is right. Fore every result that is not a standard concept , we want to get the standardConcepts and set it on the resp.standardConcepts? 
Don't we just want to do this if we matched on concept_code or concept_id and we only have one result ?

SrushtiGangireddy: Messed up a line. Added back.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -187,6 +187,14 @@ from \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.achilles_results\` r
 where r.analysis_id = 3000 and CAST(r.stratum_1 as int64) > "0" group by r.stratum_1) as r
 where r.concept_id = c.concept_id"
 
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"Update \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\` c
+set c.count_value = r.count_value,c.source_count_value=r.source_count_value
+from  (select cast(r.stratum_1 as int64) as concept_id, count_value , source_count_value
+from \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.achilles_results\` r
+where r.analysis_id in (2,4,5) and CAST(r.stratum_1 as int64) > 0 group by r.stratum_1, r.count_value, r.source_count_value) as r
```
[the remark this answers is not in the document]
SrushtiGangireddy: Changed the query.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
biopete: I was talking about adding a parameter to the SearchConceptRequest object . It would be an optional array with the match types you wanted. The array order sets the precedence. It would default to [Code, ConceptId , Name] ..   But if you  did want only standard concepts you would probably be searching on name.  That option only makes sense searching on name.  Unless a couple vocabularies use the same code and you would get more than one. It is rare in any case and there won't be enough matches to create noise imo. 

We could also do : matchType: [ {type:Code, standard_concept:false},{type:ConceptId, standard_concept:false, {type:Name, standard_concept:true} ]  -- so instead of array of scalars have array of objects and put the standard_concept filter in each match type object.

biopete: That said, its a lot of work for something we aren't using now. But given we can filter on domain, vocabulary, seems we should filter on standard_concept too just to be thorough.  That said, given that the standard_concept filter only makes sense in a concept_name search, i think we are fine as we are.  But Dan if you really don't like it , i'm fine taking it out for now until we need it.

biopete: I think the best option is to add another StandardConceptFilter.StandardOrCodeIdMatch ... Seems a good use of the StandardConceptFilter enumerator.

biopete:
I'm out tomorrow. I've tested this and functionally it works. I'm ok if api changes a little and the standard concept filter is removed or whatever.  So i will approve pending Dan's approval.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: I'm OK with that, but then for the other options (standard, non-standard) we should return only those concepts... we shouldn't do ID matching for the other kind of concepts in those cases.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -103,30 +104,62 @@ public static String modifyMultipleMatchKeyword(String query){
 
                     // Optionally filter on standard concept, vocabulary ID, domain ID
                     if (standardConceptFilter.equals(StandardConceptFilter.STANDARD_CONCEPTS)) {
+
+                        predicates.add(
```
danqrodney: It might be simpler to make a conceptNamePredicate used here and in the next if block; and only make conceptName_filter in the else case.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -103,30 +104,62 @@ public static String modifyMultipleMatchKeyword(String query){
...
                                 criteriaBuilder.literal(STANDARD_CONCEPT_CODE)));
                         standardConceptPredicates.add(criteriaBuilder.notEqual(root.get("standardConcept"),
                                 criteriaBuilder.literal(CLASSIFICATION_CONCEPT_CODE)));
+
+                        predicates.add(
+                               criteriaBuilder.or(standardConceptPredicates.toArray(new Predicate[0]))
+                        );
+
+
+                    } else if (standardConceptFilter.equals(StandardConceptFilter.STANDARD_OR_CODE_ID_MATCH)) {
```
danqrodney: So we may want to make this filter be the default behavior if no filter is specified in the request. If we don't, you should have an else block after this that adds the logic that used to be on lines 125-129... we should always apply a filter on name / ID. (And you should make sure DataBrowserController passes this filter in.)

PR #940, opened 2018-06-16
[RW-755][risk=no] Avoid n^2 regex pattern

calbach:
A mismatch at the end of the string means we have to traverse a graph with binary branching for the `.` vs `.char`. Add regression test.

Also drop `'` from valid username characters, was this included by mistake?


rushtong:
Apostrophe's are valid in gsuite usernames (https://support.google.com/a/answer/33386?hl=en) so I'm pretty sure that was intentional.

calbach:
Oops, the comment even said that apostrophes are supported. Pretty crazy character to put in an email address if you ask me, but makes sense to match gsuite for now.

Yes, this fixes 755.

PR #941, opened 2018-06-18
Retry billing project creation twice

blrubenstein:
Open question (which is somewhat rhetorical): Should we log the error message we get back from Firecloud? I assume yes, but it will involve reworking what we get back from list billing projects to include a bit more info than just the status.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -274,18 +274,22 @@ public class ProfileController implements ProfileApiDelegate {
 
       case ERROR:
         int retries = Optional.ofNullable(user.getBillingProjectRetries()).orElse(0);
-        if (retries <= 2) {
+        if (retries < workbenchConfigProvider.get().firecloud.billingRetryCount) {
           this.userService.setBillingRetryCount(retries + 1);
-          fireCloudService.removeUserFromBillingProject(user.getEmail(), user.getFreeTierBillingProjectName());
+          log.log(Level.INFO, String.format(
+              "Billing project %s failed to be created, retrying.", user.getFreeTierBillingProjectName()));
+          try {
+            fireCloudService.removeUserFromBillingProject(user.getEmail(), user.getFreeTierBillingProjectName());
+          } catch (WorkbenchException e) {
+            log.log(Level.INFO, String.format("Failed to remove user from errored billing project"));
+          }
           String billingProjectName = createFirecloudBillingProject(user);
-          user.setFreeTierBillingProjectName(billingProjectName);
-          user.setFreeTierBillingProjectStatus(BillingProjectStatus.PENDING);
-          return userDao.save(user);
+          return this.userService.setBillingProjectNameAndStatus(billingProjectName, BillingProjectStatus.PENDING);
         } else {
+          String billingProjectName = user.getFreeTierBillingProjectName();
           log.log(Level.SEVERE, String.format(
-              "free tier project %s failed to be created", user.getFreeTierBillingProjectName()));
-          user.setFreeTierBillingProjectStatus(status);
-          return userDao.save(user);
+              "free tier project %s failed to be created", billingProjectName));
+          return userService.setBillingProjectNameAndStatus(billingProjectName, status);
         }
       case READY:
         break;
@@ -319,8 +323,7 @@ public class ProfileController implements ProfileApiDelegate {
     } catch (GatewayTimeoutException e) {
       log.log(Level.WARNING, "Socket Timeout creating cluster.");
     }
-    user.setFreeTierBillingProjectStatus(BillingProjectStatus.READY);
-    return userDao.save(user);
+    return userService.setBillingProjectNameAndStatus(user.getFreeTierBillingProjectName(), BillingProjectStatus.READY);
   }
 
   private ResponseEntity<Profile> getProfileResponse(User user) {
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -150,6 +150,10 @@ public class ProfileController implements ProfileApiDelegate {
     // If the user is already registered, their profile will get updated.
     fireCloudService.registerUser(user.getContactEmail(),
         user.getGivenName(), user.getFamilyName());
+    return createFirecloudBillingProject(user);
+  }
+
+  private String createFirecloudBillingProject(User user) {
     WorkbenchConfig workbenchConfig = workbenchConfigProvider.get();
     long suffix;
     if (workbenchEnvironment.isDevelopment()) {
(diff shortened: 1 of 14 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
```diff
@@ -49,6 +49,12 @@
    */
   void addUserToBillingProject(String email, String projectName);
 
+  /**
+   * Removes the specified user from the specified billing project.
+   * Only used for errored billing projects
+   */
+  void removeUserFromBillingProject(String email, String projectName);
```
rushtong: Question: Does this result in orphaned billing accounts? And if so, will we need some way to prune them over time? I can envision this being an admin process. If this is necessary, can you add a story for that so we don't lose track?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -269,11 +273,20 @@ private User initializeUserIfNeeded() {
...
-        return userDao.save(user);
-
+        int retries = Optional.ofNullable(user.getBillingProjectRetries()).orElse(0);
+        if (retries <= 2) {
+          this.userService.setBillingRetryCount(retries + 1);
+          fireCloudService.removeUserFromBillingProject(user.getEmail(), user.getFreeTierBillingProjectName());
+          String billingProjectName = createFirecloudBillingProject(user);
+          user.setFreeTierBillingProjectName(billingProjectName);
+          user.setFreeTierBillingProjectStatus(BillingProjectStatus.PENDING);
+          return userDao.save(user);
```
danqrodney: Again, better to use UserService to save this stuff with retries, so we handle race conditions appropriately.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -306,6 +306,29 @@ public void testMe_secondCallErrorsProject() throws Exception {
...
+    Profile profile = profileController.getMe().getBody();
+    assertThat(profile.getFreeTierBillingProjectStatus()).isEqualTo(BillingProjectStatus.PENDING);
+
+    // Simulate FC "Error".
+    org.pmiops.workbench.firecloud.model.BillingProjectMembership membership =
+        new org.pmiops.workbench.firecloud.model.BillingProjectMembership();
+    membership.setCreationStatus(CreationStatusEnum.ERROR);
+    membership.setProjectName(profile.getFreeTierBillingProjectName());
+    when(fireCloudService.getBillingProjectMemberships()).thenReturn(ImmutableList.of(membership));
+    profile = profileController.getMe().getBody();
```
rushtong: Would be nice to have a comment here about why we're calling this 4 times and expecting the error case.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -269,11 +273,20 @@ private User initializeUserIfNeeded() {
...
       case ERROR:
-        log.log(Level.SEVERE, String.format(
-            "free tier project %s failed to be created", user.getFreeTierBillingProjectName()));
-        user.setFreeTierBillingProjectStatus(status);
-        return userDao.save(user);
-
+        int retries = Optional.ofNullable(user.getBillingProjectRetries()).orElse(0);
+        if (retries <= 2) {
+          this.userService.setBillingRetryCount(retries + 1);
+          fireCloudService.removeUserFromBillingProject(user.getEmail(), user.getFreeTierBillingProjectName());
```
danqrodney: Any chance this is going to fail for an errored billing project? You might want to swallow any such exception and continue...

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -269,11 +273,20 @@ private User initializeUserIfNeeded() {
...
+          String billingProjectName = createFirecloudBillingProject(user);
+          user.setFreeTierBillingProjectName(billingProjectName);
+          user.setFreeTierBillingProjectStatus(BillingProjectStatus.PENDING);
+          return userDao.save(user);
+        } else {
+          log.log(Level.SEVERE, String.format(
+              "free tier project %s failed to be created", user.getFreeTierBillingProjectName()));
+          user.setFreeTierBillingProjectStatus(status);
```
danqrodney: At this point I guess the user can't use the free tier billing project... which seems ok... the UI will indicate as such?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -269,11 +273,20 @@ private User initializeUserIfNeeded() {
...
 
       case ERROR:
-        log.log(Level.SEVERE, String.format(
-            "free tier project %s failed to be created", user.getFreeTierBillingProjectName()));
-        user.setFreeTierBillingProjectStatus(status);
-        return userDao.save(user);
-
+        int retries = Optional.ofNullable(user.getBillingProjectRetries()).orElse(0);
+        if (retries <= 2) {
+          this.userService.setBillingRetryCount(retries + 1);
```
danqrodney: Might want to log a warning here.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
rushtong: It might also be nice to iterate over the max retries value itself instead of manually calling this N times.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -269,11 +273,20 @@ private User initializeUserIfNeeded() {
         return user;
 
       case ERROR:
-        log.log(Level.SEVERE, String.format(
-            "free tier project %s failed to be created", user.getFreeTierBillingProjectName()));
-        user.setFreeTierBillingProjectStatus(status);
-        return userDao.save(user);
-
+        int retries = Optional.ofNullable(user.getBillingProjectRetries()).orElse(0);
+        if (retries <= 2) {
```
rushtong: If you moved `2` to a configuration value, we would be able to use it in the test, and perhaps adjust it more easily if this is problematic.

blrubenstein: Sure! I can do that.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: Will do.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: Its possible. I will wrap in a try/catch.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: Will do. This is what we were doing for saving user billing project stuff before. Should I move that over too?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: Yeah... I think we need some sort of UI error indicator here. I'm not sure what to do in this situation though.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
blrubenstein: Yes, it will. FWIW, our service account will be the owner of all of them. We don't really have anything we can currently do though.

PR #944, opened 2018-06-19
[RW-628][Risk=no] Fix column name

blrubenstein:


api/src/main/java/org/pmiops/workbench/db/model/User.java
```diff
@@ -299,7 +299,7 @@ public void setClusterCreateRetries(Integer clusterCreateRetries) {
     this.clusterCreateRetries = clusterCreateRetries;
   }
 
-  @Column(name = "billing_project_retries")
+  @Column(name = "billing_project_retry")
   public Integer getBillingProjectRetries() {
```
rushtong: Talking in slack about refactoring this to match the column name.

PR #946, opened 2018-06-20
[RW-871][Risk=low] Blrubenstein/fix clone registered

blrubenstein:


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -64,6 +64,16 @@ public FireCloudServiceImpl(Provider<WorkbenchConfig> configProvider,
     this.retryHandler = retryHandler;
   }
 
+  private void checkAndAddRegistered(WorkspaceIngest workspaceIngest) {
+    if (configProvider.get().firecloud.enforceRegistered) {
+      ArrayList<ManagedGroupRef> authDomain = new ArrayList<ManagedGroupRef>();
+      ManagedGroupRef registeredDomain = new ManagedGroupRef();
+      registeredDomain.setMembersGroupName(configProvider.get().firecloud.registeredDomainName);
+      authDomain.add(registeredDomain);
+      workspaceIngest.setAuthorizationDomain(authDomain);
```
calbach: nit: I would probably remove the `authDomain` variable entirely and inline an `ImmutableList.of(registeredDomain)` here

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -171,13 +181,7 @@ public void createWorkspace(String projectName, String workspaceName) {
     workspaceIngest.setName(workspaceName);
     workspaceIngest.setNamespace(projectName);
     // TODO: add concept of controlled auth domain.
```
calbach: nit: maybe move this TODO into the helper now

PR #949, opened 2018-06-22
[RW-680][RW-432][risk=low] Notebook redirect instead of modal

calbach:
- Add a new notebook page which waits for the cluster to load, sets up auth, and localizes or creates a notebook.
- Drop the notebook spinner from the workspaces page, instead opening/creating a notebook brings you to the new redirect page.
- Rough out a framework for a loading UX with more detailed messaging on this page (UX input pending).
- ~~Fix a bug in postMessaging of the clientId to Leonardo (I may split this out to get that in sooner)~~ https://github.com/all-of-us/workbench/pull/950

TODO
- ~~Unit test for the new view~~
- UX is a placeholder, getting input from Lou on this
- ~~Related: UX pending - we may not want the nav bar on this view~~

~~For future PRs~~
- ~~Integrate cluster "resume" into this new page~~

### DIFF
--- a/ui/src/app/views/notebook-redirect/component.html
+++ b/ui/src/app/views/notebook-redirect/component.html
@@ -1,0 +1,15 @@
+<h2 *ngIf="notebookName">Loading Notebook '{{notebookName}}'</h2>
+<h2 *ngIf="!notebookName">Creating a New Notebook</h2>
+<div [ngSwitch]="progress" class="spinner-container">
+  <span class="spinner spinner-lg"></span>
+  <p *ngSwitchCase="Progress.Initializing" class="spinner-text">
+    <span class="spinner-header">Initializing Notebook Server</span><br>
+    May take up to 5 minutes
+  </p>
+  <p *ngSwitchCase="Progress.Configuring" class="spinner-text">
+    Configuring Notebook Server
+  </p>
+  <p *ngSwitchCase="Progress.Redirecting" class="spinner-text">
+    Redirecting to the Notebook Server
+  </p>
+</div>
--- a/ui/src/app/views/notebook-redirect/component.html
+++ b/ui/src/app/views/notebook-redirect/component.html
@@ -3,13 +3,23 @@
 <div [ngSwitch]="progress" class="spinner-container">
   <span class="spinner spinner-lg"></span>
   <p *ngSwitchCase="Progress.Initializing" class="spinner-text">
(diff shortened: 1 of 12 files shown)


ui/src/app/views/notebook-redirect/component.html
```diff
@@ -0,0 +1,15 @@
+<h2 *ngIf="notebookName">Loading Notebook '{{notebookName}}'</h2>
+<h2 *ngIf="!notebookName">Creating a New Notebook</h2>
+<div [ngSwitch]="progress" class="spinner-container">
+  <span class="spinner spinner-lg"></span>
+  <p *ngSwitchCase="Progress.Initializing" class="spinner-text">
```
UrsaStutsman: More question than comment (still learning the nitty gritty of angular) - if this is Progress.Unknown, will the spinner not show? Or will it never be that value because you set it to one of the other 3 in OnInit?

calbach: It will be Unknown when we render initially since I only set the progress in a callback to `listClusters()`. These `ngSwitchCase` statements are on the spinner text, so I'll conditionally render one of those. The actual spinner itself is the `span` one line up.

So for progress == Unknown, we're showing a spinner here but no text underneath it.

rushtong:
This is slick - I like the new AoU tab->refresh with leo page. I think this is what Saturn is doing, which makes sense to me.

calbach:
Please take another look. Added tests and integrated auto-resume into the flow. I didn't have enough time to implement the updated UX proposed by Lou, that will have to come in a follow-up. For now, it's just a single spinner with text popping in beneath it.

This PR depends on #957.

ui/src/app/views/notebook-redirect/component.html
UrsaStutsman: 👍

rushtong:
FYI, I hit the same error that **ci/circleci: ui-build-test** hit.

calbach:
Sorry about that, unit tests didn't pick that up. Should be working now. Did some manual testing of this and also added some retries. 

Note: if you were trying to patch this and run it locally, you'll need to bring up both the API and UI server.

calbach:
FYI: rebased and fixed one minor issue I noticed during the demo (my `retry(3)` actually triggered multiple stages to retry). No material changes from the last reviewed version, so I will plan to merge once CI goes green.

PR #953, opened 2018-06-26
[RW-916][risk=no] Upgrade to Guava 25.1 to fix local dev

calbach:
This appears to fix the issue for me locally.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -248,7 +248,7 @@ dependencies {
   compile 'javax.inject:javax.inject:1'
   compile 'io.swagger:swagger-annotations:1.5.16'
   compile 'org.apache.commons:commons-lang3:3.0'
-  compile 'com.google.guava:guava:24.1.1-jre'
+  compile 'com.google.guava:guava:25.1-jre'
   compile 'org.springframework.retry:spring-retry:1.1.5.RELEASE'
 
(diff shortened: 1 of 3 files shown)


rushtong:
One thing I don't see in any of our gradle tasks is a call with `--refresh-dependencies`. I'm not sure where the best place is to do that, perhaps the `clean` task, but that might also help.

calbach:
@freemabd1 did you do a docker-clean as well?

freemabd1:
@calbach I didn't do a docker-clean... but I will try that now to confirm that it's still an issue.

calbach:
@rushtong I do not claim to have identified the root cause. I think there's a bad interaction between dependencies somewhere, and at least on my machine this change jiggled them back around to working. I observed that downgrading to 23.0 or lower seemed to fix this as well, so I just tried 25.1 on a whim and it worked for me. We'll see if it does anything for Brian though..

freemabd1:
@rushtong @calbach I have confirmed that downgrading has worked for me.

freemabd1:
@calbach after doing a docker-clean, I'm still seeing the same error

calbach:
I guess I should merge this PR? This fixes things for me, doesn't make things obviously *worse* for @freemabd1 , and also addresses the security issue.

PR #954, opened 2018-06-26
[RW-793][risk=low] Submit bug reports to JIRA

NehaBroad:


api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    fieldsDetail.setDescription(bugReport.getReproSteps());
+    fieldsDetail.setSummary(bugReport.getShortDescription());
+    fieldsDetail.setProject(projectDetails);
+
+    issueType.setName(IssueTypeEnum.Bug.name());
+    fieldsDetail.setIssuetype(issueType);
+
+    issueDetails.setFields(fieldsDetail);
+    IssueResponse response = api.createIssue(issueDetails);
+    return response.getKey();
```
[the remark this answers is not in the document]
NehaBroad: IssueResponse has 3 fields , the Key(issue number), id and self which is of the following format:
{"id":"22434","key":"RW-922","self":"https://precisionmedicineinitiative.atlassian.net/rest/api/2/issue/22434"}
I assumed id and self (Self gives information about the jira ticket in json format.) will hardly ever be used and what is most useful is just the issue key.

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+   * Attach Log file to issue
+   * @param issueKey : Issue Id number to which log file is attached
+   * @param file: Attachment
+   */
+  public void attachLogFiles(String issueKey, File file) throws ApiException {
+    api.addAttachments(issueKey, file,"nocheck");
+    try {
+      file.delete();
+    } catch(SecurityException ex){
+      log.warning("Exception while deleting temp log files");
```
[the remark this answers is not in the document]
NehaBroad: Fixed

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Attach Log file to issue
+   * @param issueKey : Issue Id number to which log file is attached
+   * @param file: Attachment
+   */
+  public void attachLogFiles(String issueKey, File file) throws ApiException {
+    api.addAttachments(issueKey, file,"nocheck");
+    try {
+      file.delete();
+    } catch(SecurityException ex){
```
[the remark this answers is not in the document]
NehaBroad: Fixed

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    issueDetails.setFields(fieldsDetail);
+    IssueResponse response = api.createIssue(issueDetails);
+    return response.getKey();
+  }
+
+  /**
+   * Attach Log file to issue
+   * @param issueKey : Issue Id number to which log file is attached
+   * @param file: Attachment
+   */
```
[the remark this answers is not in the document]
NehaBroad: Fixed

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,103 @@
...
+      log.warning(String.format("Exception while deleting temp file '%s'", attachmentName));
+    }
+  }
+
+  /**
+   * Creates temp File to be attached to jira issue
+   * @param name
+   * @param content
+   * @return temp File
+   */
```
rushtong: nit: I wouldn't include javadoc for private methods like this. I think a generic comment for explanation would suffice.

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+    issueDetails.setFields(fieldsDetail);
+    api.getApiClient().setDebugging(true);
+    IssueResponse response = api.createIssue(issueDetails);
+    return response.getKey();
+  }
+
+
+  @Override
+  public void uploadAttachment(String issueKey, String attachmentName, Object content) throws ApiException {
```
rushtong: I know you're thinking of other ways to pass the attachment content. 
One thought is to have different, overloadable methods that deal with the specific content types we support. So, for example, we could only support String content right now and in the future, support other types.

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Creates temp File to be attached to jira issue
+   * @param name
+   * @param content
+   * @return temp File
+   */
+  private File createTempFile(String name,byte[] content) {
+    try{
+      File tempFile = File.createTempFile(name, ".log");
```
rushtong: I'm not so sure we should be assuming the suffix here. The caller may have specified the full file name.

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,138 @@
...
+      required:
+        - id
+        - key
+      properties:
+        id:
+          type: string
+        key:
+          type: string
+        self:
+          type: string
```
rushtong: nit: should have a newline here.

api/src/main/java/org/pmiops/workbench/jira/JiraServiceImpl.java
rushtong: Looking more closely at the content that is returned by the jupyter api, the results are of type `JupyterContents`, from which, we get the logContent string value. It might make sense to support uploading the full object itself instead of looking at the string content.

calbach: I think `content` should definitely be a `String` at this point, where the string is the file contents of the attachment to upload. The goal with this attachment was to basically push the exact logfile as we saw it on the Jupyter server onto the Jira ticket.

The tempfile stuff is a bit messy, but I'm not sure I see a clear way around with the Swagger java codegen (I looked through the generated source). Maybe this approach will suffice for now - we'd also discussed just pushing these logs into GCS for access control purposes, so that would be another approach for bypassing this problem (and you're more than welcome to do that in this PR, if that makes life easier for you somehow - I'm thinking it probably won't).

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,40 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
                   String.format("Jupyter returned null content for '%s', continuing", fileName));
               continue;
             }
-            MimeBodyPart attachPart = new MimeBodyPart();
-            attachPart.setDataHandler(new DataHandler(logContent, "text/plain"));
-            attachPart.setFileName(fileName);
-            multipart.addBodyPart(attachPart);
+            File tempLogFile = createTempFile(fileName, logContent);
+            jiraService.uploadAttachment(issueKey, tempLogFile);
+            tempLogFile.delete();
```
rushtong: If we encounter an exception after uploading the attachment, the delete call will never be executed. This should go into a `finally` block so we're guaranteed there are no lasting side effects.

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,137 @@
...
+          schema:
+            $ref: '#/definitions/IssueRequest'
+      responses:
+        200:
+          description: File saved
+          schema:
+            $ref: '#/definitions/IssueResponse'
+        400:
+          description: Bad request, check the summary/description field
+        503:
```
rushtong: I'm confused by the 503 responses here. Where does that come from? From the docs (https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-issue-post), there's no mention of a 503 error. I think this should be a 500, along with the attachments call below.

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,137 @@
...
+         required: true
+         type: file
+       - in: header
+         name: X-Atlassian-Token
+         type: string
+         default: nocheck
+     responses:
+       200:
+         description: File saved
+       503:
```
rushtong: See above comment - is this really a 503 or a 500?

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,137 @@
...
+        - id
+        - key
+      properties:
+        id:
+          type: string
+        key:
+          type: string
+        self:
+          type: string
+
```
rushtong: The docs also mention Error responses: https://developer.atlassian.com/cloud/jira/platform/rest/#error-responses
Would it make sense to document that here and add a schema ref to it for the error cases above?

api/src/test/java/org/pmiops/workbench/jira/JiraServiceImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    );
+    doThrow(new  ApiException("Missing the required parameter 'issueKey' when calling addAttachments(Async)"))
+        .when(jiraApi).addAttachments(null, mockFile, "nocheck");
+
+    doThrow(new  ApiException("Missing the required parameter 'file' when calling addAttachments(Async)"))
+        .when(jiraApi).addAttachments("IssueKey", null, "nocheck");
+
+    service.setJiraApi(jiraApi);
+    when(jiraApi.getApiClient()).thenReturn(apiClient);
+    when(jiraApi.createIssue( any() )).thenReturn(new IssueResponse());
```
rushtong: nit: spacing. `createIssue( any() ))` should be `createIssue(any()))`

api/src/main/resources/jira.yaml
NehaBroad: I encountered this  error while i was trying to create too many issues

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
-            attachPart.setDataHandler(new DataHandler(logContent, "text/plain"));
-            attachPart.setFileName(fileName);
-            multipart.addBodyPart(attachPart);
+            File tempLogFile = createTempFile(fileName, logContent);
+            try {
+              jiraService.uploadAttachment(issueKey, tempLogFile);
+            } catch (org.pmiops.workbench.jira.ApiException e) {
+              log.severe(String.format("failed to upload attachment '%s', continuing", fileName));
+            } finally {
+              if ( tempLogFile != null) {
```
rushtong: nit: extra space in `( tempLogFile`

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
+                try {
+                  tempLogFile.delete();
+                } catch (SecurityException ex){
```
rushtong: nit: spacing, `(SecurityException ex){` should be `(SecurityException ex) {`

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
-      }
-      msg.setContent(multipart);
-      mailServiceProvider.get().send(msg);
-    } catch (MessagingException e) {
-      throw new EmailException("Error sending bug report", e);
-    } catch (UnsupportedEncodingException e) {
-      throw new EmailException("Error sending bug report", e);
+     }
+    } catch (org.pmiops.workbench.jira.ApiException e) {
+      log.severe(String.format("Error while connecting to JIRA server %s", e.getMessage() ));
```
rushtong: nit: spacing, `e.getMessage() ));` should be `e.getMessage()));`

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
     return ResponseEntity.ok(bugReport);
   }
+
+  /**
+   * Creates temp File to be attached to jira issue
+   * @param name
+   * @param content
+   * @return temp File
+   */
+  private File createTempFile(String name,String content) {
```
rushtong: nit: spacing between arguments

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
+    try{
```
rushtong: nit: spacing `try{` should be `try {`

api/src/main/java/org/pmiops/workbench/api/BugReportController.java
```diff
@@ -107,23 +73,49 @@ void setMailServiceProvider(Provider<MailService> mailServiceProvider) {
...
+      File tempFile = File.createTempFile(name, ".log");
+      FileOutputStream writer = new FileOutputStream(tempFile);
+      writer.write(content.getBytes());
+      return tempFile;
+    } catch(IOException e){
```
rushtong: nit: space before opening paren

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,141 @@
...
+          schema:
+            $ref: '#/definitions/IssueRequest'
+      responses:
+        200:
+          description: File saved
+          schema:
+            $ref: '#/definitions/IssueResponse'
+        400:
+          description: Bad request, check the summary/description field
+        503:
```
rushtong: I still think we need a 500 description here.

api/src/main/resources/workbench.yaml
```diff
@@ -120,6 +120,8 @@ paths:
           description: Success message
           schema:
             $ref: "#/definitions/BugReport"
+        500:
+          description: Cannot connect to JIRA. Please try again later
```
rushtong: I still think this message is not accurate.

api/src/main/resources/jira.yaml
```diff
@@ -0,0 +1,141 @@
...
+        - id
+        - key
+      properties:
+        id:
+          type: string
+        key:
+          type: string
+        self:
+          type: string
+
```
rushtong: I can live without the extra error response description.

api/src/test/java/org/pmiops/workbench/jira/JiraServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Test
+  public void testCreateIssue() throws Exception {
+    IssueResponse response = service.createIssue(new BugReport());
+    assertNotNull(response);
+  }
+
+  @Test
+  public void testAddAttachmentNullIssueKey() throws ApiException {
+    expectedException.expect(ApiException.class);
```
rushtong: This is a nice improvement!

PR #955, opened 2018-06-26
[RW-823][risk=no] Adding ethnicity to the demographics search in CB

freemabd1:
[RW-823][risk=no] Adding ethnicity to the demographics search in CB

This PR will allow users to search by Ethnicity in the demographics section of the CB.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/DemoQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/DemoQueryBuilder.java
@@ -43,9 +43,12 @@ public class DemoQueryBuilder extends AbstractQueryBuilder {
                     "SELECT 'x' FROM `${projectId}.${dataSetId}.death` d\n" +
                     "where d.person_id = p.person_id)\n";
 
+    private static final String DEMO_ETH =
+      "p.ethnicity_concept_id in unnest(${eth})\n";
+
     private static final String AND_TEMPLATE = "and\n";
 
-    public enum DEMOTYPE { GEN, AGE, DEC, RACE };
+    public enum DEMOTYPE { GEN, AGE, DEC, RACE, ETH };
 
     @Override
     public QueryJobConfiguration buildQueryJobConfig(QueryParameters parameters) {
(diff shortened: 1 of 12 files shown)


api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/DemoQueryBuilder.java
```diff
@@ -43,9 +43,12 @@
                     "SELECT 'x' FROM `${projectId}.${dataSetId}.death` d\n" +
                     "where d.person_id = p.person_id)\n";
 
+    private static final String DEMO_ETH =
+      "p.ethnicity_concept_id in unnest(${eth})\n";
```
rushtong: nit: indentation is inconsistent with the rest of the file.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/DemoQueryBuilder.java
```diff
@@ -85,6 +86,10 @@ public QueryJobConfiguration buildQueryJobConfig(QueryParameters parameters) {
                     throw new IllegalArgumentException("Dec must provide a value of: " + DECEASED);
                 }
 
+            } else if (key.equals(DEMOTYPE.ETH.name())) {
+                final String namedParameter = key.toLowerCase() + getUniqueNamedParameterPostfix();
+                queryParts.add(DEMO_ETH.replace("${eth}", "@" + namedParameter));
+                queryParams.put(namedParameter, QueryParameterValue.array(paramMap.get(key).stream().toArray(Long[]::new), Long.class));
```
rushtong: Perhaps not for this PR, but perhaps for a future refactoring, this series of if-else-ifs could be a switch on the `DEMOTYPE` enum. I think that would be a lot cleaner.

rushtong: Not hugely important, but you could be a little safer by streaming through a filter to ensure everything really is a Long value, then map to cast it, and finally to an array.

PR #957, opened 2018-06-26
[RW-847][risk=low] Switch local/test to FC/Leo dev 

calbach:
Note that this doubly encodes the Leo API URL on client (environment.ts) and server (config.json).

Alternative considered:
Add the leoBaseUrl to the ServerConfigResponse.
+ (+) Single definition
- (-) Higher complexity introduced by instantiating Notebook-related services in an asynchronous manner
- (-) Possible race conditions if notebook requests get made before we get back the server config response

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -1,7 +1,7 @@
 // Configuration for the local development environment.
 {
   "firecloud": {
-    "baseUrl": "https://firecloud-orchestration.dsde-dev.broadinstitute.org",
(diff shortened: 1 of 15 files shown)


rushtong:
Switching to FC dev will mean that the workspaces defined in our `Workspace` table will no longer be valid. Do we need any kind of sql cleanup for that? Outside of that, I like the cleanup.

calbach:
Yes, this will need to be followed by a wipe of the test DB (and docker-cleans locally).

PR #961, opened 2018-06-27
concept return fix [risk=no]

SrushtiGangireddy:
1. Added a check to return max results of concepts if the search query is empty/ null.
2. Added the test for the same.
3. Added the test to check if the concepts with empty count are returned in the search.


common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -14,11 +14,24 @@
...
+    List<Concept> findAllConceptsOrderedByCount(@Param("maxResults") long maxResults);
+
+    @Query(value = "select c.* from concept c " +
+            "where c.domain_id in ('Condition','Observation','Procedure', 'Measurement', 'Drug') and " +
+            "c.standard_concept='S' " +
+            "order by c.count_value desc limit :maxResults ",
+            nativeQuery = true)
+    List<Concept> findAllStandardConceptsOrderedByCount(@Param("maxResults") long maxResults);
+
+    @Query(value = "select c.* from concept c " +
```
biopete: Shouldn't these say "where count_value > 0 || source_count_value > 0 " ? 

Have you tested with the public-ui ?

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
```diff
@@ -102,7 +103,7 @@
     private static final Concept CLIENT_CONCEPT_6 = new Concept()
             .conceptId(7891L)
             .conceptName("conceptD test concept 2")
-            .standardConcept(null)
+            .standardConcept("")
```
biopete: I think we should add one for null too to test all three .

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
```diff
@@ -239,6 +254,23 @@ public void testGetDbDomains() throws Exception{
...
+
+    @Test
+    public void testConceptSearchEmptyCount() throws Exception{
+        saveData();
+        assertResults(
+                dataBrowserController.searchConcepts(new SearchConceptsRequest().query("conceptD")), CLIENT_CONCEPT_5, CLIENT_CONCEPT_6)
+        ;
+    }
+
+
```
biopete: We need a test to make sure that any concepts with count 0 are not returned

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
```diff
@@ -239,6 +284,61 @@ public void testGetDbDomains() throws Exception{
...
+        List<Concept> concepts = response.getBody().getItems().stream().map(TO_CLIENT_CONCEPT).collect(Collectors.toList());
+        assertThat(concepts)
+                .containsExactly(CONCEPT_1, CONCEPT_2, CONCEPT_3, CONCEPT_4, CONCEPT_5, CONCEPT_6)
+        ;
+    }
+
+
+    @Test
+    public void testConceptSearchEmptyCount() throws Exception{
+        // Exact concept search was not tested. Match in tests was replaced by like and like > 0.0 was not possible to test
```
biopete: Comment this better like : 
" We can't test limiting to count > 0 with a concept name search because the match function does not work in hibernate. So we make several concepts with same concept code and one with count 0.   The limit > 0 works the same weather it is code or  name  match.  "

PR #964, opened 2018-06-29
[RW-934][risk=no] Upgrade GAE-gradle and swagger gen libraries

rushtong:
Huge kudos to @biopete for finding the source of the issue. 

See [RW-934](https://precisionmedicineinitiative.atlassian.net/browse/RW-934). We started to see errors in overnight processes with swagger validation. This update seems to fix the problem.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -126,8 +126,8 @@ buildscript {    // Configuration for building
   }
   dependencies {
     classpath 'org.springframework.boot:spring-boot-gradle-plugin:1.5.3.RELEASE'
-    classpath 'com.google.cloud.tools:appengine-gradle-plugin:+'
-    classpath 'gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.9.0'
+    classpath 'com.google.cloud.tools:appengine-gradle-plugin:1.3.5'
+    classpath 'gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.12.0'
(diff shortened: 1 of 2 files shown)


api/build.gradle
```diff
@@ -126,8 +126,8 @@ buildscript {    // Configuration for building
   }
   dependencies {
     classpath 'org.springframework.boot:spring-boot-gradle-plugin:1.5.3.RELEASE'
-    classpath 'com.google.cloud.tools:appengine-gradle-plugin:+'
-    classpath 'gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.9.0'
+    classpath 'com.google.cloud.tools:appengine-gradle-plugin:1.3.5'
```
biopete: need to do same in public-api/build.gradle .

api/build.gradle
```diff
@@ -126,8 +126,8 @@ buildscript {    // Configuration for building
...
+    classpath 'gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.12.0'
```
biopete: What does this do ?

biopete: I see. Just updating to latest ...

rushtong: Yes

api/build.gradle
rushtong: Check!

rushtong:
@biopete @blrubenstein - tests are now all passing, PTAL, thank you!

PR #967, opened 2018-07-02
[RW-707][risk=low] Update Account Success Page

rushtong:
Addresses https://precisionmedicineinitiative.atlassian.net/browse/RW-707

## Changes
* Added resend email feature
* Added change contact email feature

## Notes
Both features do not use authentication because the user has not authenticated yet and needs access to both. Resend is benign and has no impact. We have a very narrow security risk with the change contact email feature. In this scenario, we are only allowing the change of contact email when the user has never signed in. This feature will not work for users that have previously signed in.

## Testing
* Create a new account in the usual manner
* Check your contact email for the welcome email
* Before using the account, stay on the success page and ...
  * Try resending the welcome email and look for it in your email
  * Try changing contact email and look for the welcome message under the new account.
* If you want to be tricky, craft a curl call to change your contact email for a user that has signed in. This fails with a 403 error.
* Before actually using this account, you can delete it in gsuite and re-use the same information to re-test.

## Welcome Message
Don't worry - that password is not valid.
screen shot 2018-07-03 at 10 54 27 am

## Resend Modal
screen shot 2018-07-03 at 9 39 50 am

## Change Contact Email Modal
screen shot 2018-07-03 at 9 40 05 am


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -433,16 +449,51 @@ private User initializeUserIfNeeded() {
...
       throw new BadRequestException(
-        "Missing or incorrect invitationKey (this API is not yet publicly launched)");
+          "Missing or incorrect invitationKey (this API is not yet publicly launched)");
+    }
+  }
+
+  /*
+   * This un-authed API method is limited such that we only allow contact email updates before the user has ever
+   * signed in. This is necessary for the new user sign-up case where the user has not yet created their account.
+   * Once they have a full account, they can change their password through the normal profile update process.
```
rushtong: TO ME: confusing language ... fix this.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -144,11 +146,11 @@ public void deleteUser(String username) {
     }
   }
 
-  private String randomString(){
+  private String randomString() {
     return IntStream.range(0, 17).boxed().
-      map(x -> ALLOWED.charAt(rnd.nextInt(ALLOWED.length()))).
-      map(Object::toString).
-      collect(Collectors.joining(""));
+        map(x -> ALLOWED.charAt(rnd.nextInt(ALLOWED.length()))).
```
blrubenstein: Nit: We've been putting the . at the start of a newline, rather than the end.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -593,6 +599,71 @@ public void testMe_removeAllInstitutionalAffiliations() throws Exception {
...
+    UpdateContactEmailRequest request = new UpdateContactEmailRequest();
+    request.setContactEmail(user.getContactEmail());
+    request.setUsername(user.getEmail());
+
+    ResponseEntity response = profileController.updateContactEmail(request);
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.NO_CONTENT);
+  }
+
+  @Test
+  public void resentWelcomeEmail_messagingException() throws Exception {
```
blrubenstein: Nit: Spelling (Should be `resend`? And below?)

ui/src/app/views/account-creation-modals/component.ts
```diff
@@ -0,0 +1,64 @@
...
+  sendAndUpdate() {
+    const request: UpdateContactEmailRequest = {
+      username: this.username + '@' + this.gsuiteDomain,
+      contactEmail: this.contactEmail
+    };
+    this.updateEmail.emit(this.contactEmail);
+    this.profileService.updateContactEmail(request).subscribe(() => {
+      this.changingEmail = false;
+    });
+    this.send();
```
blrubenstein: Does this create a race condition? I'm a little worried since it fetches the contact email from the db, which means that if the above doesn't complete before send goes off, it may not send to the right place.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
rushtong: Good catch - that was on my todo list and I missed it.

rushtong:
Modal screenshots added.

ui/src/app/views/account-creation-modals/component.ts
rushtong: Good question, and I'll look into this a little more closely.

rushtong:
Commenting to myself so we make sure to add mail content to each of the environment buckets.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -433,16 +449,51 @@ private User initializeUserIfNeeded() {
...
+  private void verifyInvitationKey(String invitationKey) {
+    if (invitationKey == null || invitationKey.equals("") || !invitationKey.equals(cloudStorageService.readInvitationKey())) {
       throw new BadRequestException(
-        "Missing or incorrect invitationKey (this API is not yet publicly launched)");
+          "Missing or incorrect invitationKey (this API is not yet publicly launched)");
+    }
+  }
+
+  /*
+   * This un-authed API method is limited such that we only allow contact email updates before the user has signed in
```
calbach: One thing we could consider doing to close the remaining gap would be to generate a nonce on account creation (stored on the user table) and require that here (and for the welcome resend).

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -433,16 +449,51 @@ private User initializeUserIfNeeded() {
...
+  public ResponseEntity<Void> updateContactEmail(UpdateContactEmailRequest updateContactEmailRequest) {
+    User user = userDao.findUserByEmail(updateContactEmailRequest.getUsername());
+    if (user.getFirstSignInTime() != null) {
+      return ResponseEntity.status(HttpStatus.FORBIDDEN).build();
+    }
+    try {
+      InternetAddress email = new InternetAddress(updateContactEmailRequest.getContactEmail());
+      email.validate();
+    } catch (AddressException e) {
+      return ResponseEntity.badRequest().build();
```
calbach: Should you log the exception here? Worth including a message like "invalid contact email" in the response?

api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -59,6 +60,7 @@
     public String adminIdVerification;
     public String supportGroup;
     public String verifiedSendingAddress;
+    public String loginPage;
```
calbach: nit: I would probably name this loginUrl

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -100,12 +108,34 @@ private MandrillMessage buildWelcomeMessage(String contactEmail, String password
...
+
+  private String buildEmailHtml(String password, User user) throws IOException {
+    CloudStorageService cloudStorageService = cloudStorageServiceProvider.get();
+    StringBuilder contentBuilder = new StringBuilder();
+    URL emailContent = Resources.getResource(WELCOME_RESOURCE);
+    Resources
+        .readLines(emailContent, StandardCharsets.UTF_8)
+        .forEach(s -> contentBuilder.append(s).append("\n"));
+    String string = contentBuilder.toString();
+    Map<String, String> replaceMap = new HashMap<>();
```
calbach: opt_nit: I'd probably use an ImmutableMap.Builder here so I could chain these puts (and then inline that in the StrSubstitutor ctor call).

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -433,16 +449,51 @@ private User initializeUserIfNeeded() {
...
+    }
+    user.setContactEmail(updateContactEmailRequest.getContactEmail());
+    userDao.save(user);
+    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+  }
+
+  @Override
+  public ResponseEntity<Void> resendWelcomeEmail(ResendWelcomeEmailRequest resendRequest) {
+    com.google.api.services.admin.directory.model.User googleUser =
```
calbach: This should have the same sign-in check. Otherwise any anonymous client could reset everyone's passwords, just knowing their username.

api/src/main/resources/emails/welcomeemail/content.html
```diff
@@ -86,31 +86,33 @@
...
   <div class="welcome-message">Welcome to the All of Us Researcher Workbench!</div>
-  <div><div class="text user-field">Your Username: </div><div class="user-info">{{USERNAME-INPUT}}</div></div>
-  <div><div class="text user-field">Your Temporary Password: </div><div class="user-info">{{PASSWORD-INPUT}}</div></div>
+  <div><div class="text user-field">Your Username: </div><div class="user-info">${USERNAME}</div></div>
+  <div><div class="text user-field">Your Temporary Password: </div><div class="user-info">${PASSWORD}</div></div>
   <br>
   <div class="text">Quick guide on how to get started:</div>
   <a class="card" href="https://myaccount.google.com/signinoptions/two-step-verification">
-    <div class="bullet">1</div>
+    <img class="bullet" src="${BULLET_1}">
```
calbach: I suggested to @blrubenstein that we just use CSS for this, as the bullets seem pretty simple. I think we should minimize images as I'd people to have them disabled in their email by default.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -593,6 +599,71 @@ public void testMe_removeAllInstitutionalAffiliations() throws Exception {
...
+  public void updateContactEmail_OK() throws Exception {
+    createUser();
+    when(fireCloudService.isRequesterEnabledInFirecloud()).thenReturn(true);
+    user.setFirstSignInTime(null);
+    UpdateContactEmailRequest request = new UpdateContactEmailRequest();
+    request.setContactEmail(user.getContactEmail());
+    request.setUsername(user.getEmail());
+
+    ResponseEntity response = profileController.updateContactEmail(request);
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.NO_CONTENT);
```
calbach: Should these 3 tests be verifying that the contact email did/ didn't change?

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -593,6 +599,71 @@ public void testMe_removeAllInstitutionalAffiliations() throws Exception {
...
+    ResponseEntity response = profileController.resendWelcomeEmail(request);
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR);
+  }
+
+  @Test
+  public void resendWelcomeEmail_OK() throws Exception {
+    createUser();
+    when(fireCloudService.isRequesterEnabledInFirecloud()).thenReturn(true);
+    when(directoryService.resetUserPassword(anyString())).thenReturn(googleUser);
+    doNothing().when(mailService).sendWelcomeEmail(any(), any(), any());
```
calbach: Should this and the above test be verifying whether the call happened?

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -593,6 +599,71 @@ public void testMe_removeAllInstitutionalAffiliations() throws Exception {
...
+    ResendWelcomeEmailRequest request = new ResendWelcomeEmailRequest();
+    request.setUsername(user.getEmail());
+
+    ResponseEntity response = profileController.resendWelcomeEmail(request);
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.NO_CONTENT);
```
calbach: Possibly also worth asserting that the password changed.

ui/src/app/views/account-creation-modals/component.html
```diff
@@ -0,0 +1,25 @@
+<div class="change_account_email">
```
calbach: We use dash-delimiters for classnames everywhere else in the app AFAIK

ui/src/app/views/account-creation-modals/component.ts
```diff
@@ -0,0 +1,64 @@
...
+    private profileService: ProfileService,
+    serverConfigService: ServerConfigService
+  ) {
+    serverConfigService.getConfig().subscribe((config) => {
+      this.gsuiteDomain = config.gsuiteDomain;
+    });
+    this.contactEmail = '';
+  }
+
+  ngOnInit() {
```
calbach: Unused, rm and unimplement the interface

ui/src/app/views/account-creation-modals/component.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="modal-body">
+            <div class="form-section">
+                <label>Contact Email:</label>
+                <input id="change-contact-email" type="text" class="input" name="contact-email"
+                       [(ngModel)]="contactEmail">
+            </div>
+        </div>
+        <div class="modal-footer">
+            <button type="button" class="btn btn-outline" (click)="changingEmail=false">Cancel</button>
+            <button id="change_email" type="button" class="btn btn-primary" (click)="sendAndUpdate()">Apply</button>
```
calbach: nit: More accurately updateAndSend

ui/src/app/views/account-creation-modals/component.ts
calbach: It definitely does, the send() should be inside the callback.

ui/src/app/views/account-creation-modals/component.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  sendAndUpdate() {
+    const request: UpdateContactEmailRequest = {
+      username: this.username + '@' + this.gsuiteDomain,
+      contactEmail: this.contactEmail
+    };
+    this.updateEmail.emit(this.contactEmail);
+    this.profileService.updateContactEmail(request).subscribe(() => {
+      this.changingEmail = false;
```
calbach: Do we want to close the modal before the send() happens?

calbach: Do we need a loading indicator while this XHR is pending? e.g. clrLoading on the send button

ui/src/app/views/account-creation-modals/component.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+    this.send();
+  }
+
+  send() {
+    const request: ResendWelcomeEmailRequest = {
```
calbach: opt_nit: I would just inline this. Here and in sendAndUpdate

ui/src/app/views/account-creation-success/component.ts
```diff
@@ -30,10 +40,17 @@ export class AccountCreationSuccessComponent {
       loginComponent.smallerBackgroundImgSrc = '/assets/images/congrats-female-standing.png';
       loginComponent.backgroundImgSrc = '/assets/images/congrats-female.png';
     }, 0);
-    this.email = account.profile.username;
+    this.username = account.profile.username;
+  }
+
+  ngOnInit () {
```
calbach: Unused, why add the interface?

ui/src/app/views/account-creation-success/component.html
```diff
@@ -7,15 +7,19 @@ <h4>Your All of Us research account has been created!</h4>
...
   <div class="row">
-    <h4>Please log in to your new account.</h4>
+    <h4>Check your contact email for instructions on getting started.</h4>
+  </div>
+  <div class="row">
+    <h4>Your contact email is: {{contactEmail}}</h4>
   </div>
   <div class="row" style="padding-top:0.5rem;">
-    <button class="btn btn-primary" (click)="signIn()">LOGIN</button>
+    <button class="button-link" (click)="accountCreationModalsComponent.resendInstructions();">Resend Instructions</button> | <button class="button-link" (click)="accountCreationModalsComponent.updateAndSendEmail();">Change contact email</button>
```
calbach: Some newlines would be nice here

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
UrsaStutsman: Hmmmm. I don't see any reason to limit the welcome resend. We just wanted to close the following security gap: If a user sets up an account, has NOT yet done the password reset (forced on first sign in), in this time someone could conceivably reset the email, and have the new password sent to that new email, blocking the initial user from the account they were trying to set up. Resending the initial email to the initial entered contact email is not a gap.

calbach: My comment here is on the updateContactEmail method.

calbach: @UrsaStutsman and I had some more follow-up discussion offline, but pasting in here for benefit of the group:

> The gap (Greg's PR also mentions it in the description under "Notes"):
> - An anonymous unauthorized user should not be able to change the contact email on an arbitrary account
> - Right now this is possible between signup and first login
>
> How would a nonce address this:
> Specifically, we would generate a random number and store on the User. We'd return this from createAccount and the client would include it on updateContactEmail (and may as well also include it on resendWelcome, agreed there's less of a security risk on that one). Server side we'd verify this number against what we stored. This allows us to verify that the request is coming from the same client session which originally created the account.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -433,16 +449,51 @@ private User initializeUserIfNeeded() {
...
+
+  /*
+   * This un-authed API method is limited such that we only allow contact email updates before the user has signed in
+   * with the newly created gsuite account. Once the user has logged in, they can change their contact email through
+   * the normal profile update process.
+   */
+  @Override
+  public ResponseEntity<Void> updateContactEmail(UpdateContactEmailRequest updateContactEmailRequest) {
+    User user = userDao.findUserByEmail(updateContactEmailRequest.getUsername());
+    if (user.getFirstSignInTime() != null) {
```
calbach: Thinking more on this, the ability to change contact email does also introduce some holes in terms of email verification. Happy to go over the various cases offline, but in general I think it would be more secure to do the following:

- If possible, check against gsuite somehow to check for whether they've already logged in, e.g. setup a new password, setup recovery options, instead of checking for workbench sign in.
- Move the password reset (and perhaps the resend of welcome email) onto the server side to guarantee that contact email and pw reset are happening together (secure ordering would be reset pw -> update contact email).

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
UrsaStutsman: Added below

api/config/config_stable.json
```diff
@@ -31,6 +32,7 @@
     "projectId": "all-of-us-rw-stable"
   },
   "admin": {
+    "loginUrl": "http:\/\/all-of-us-workbench-stable.appspot.com/login",
```
calbach: We have a prod config now as well, please add there

api/config/config_stable.json
```diff
@@ -21,6 +21,7 @@
   },
   "googleCloudStorageService": {
     "credentialsBucketName": "all-of-us-rw-stable-credentials",
+    "emailImagesBucketName": "all-of-us-workbench-test-email-images",
```
calbach: Should these be specific to the environment?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
calbach: Doesn't seem to be addressed

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
calbach: This has the same issue still, since it only checks after the password reset.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -115,13 +115,49 @@ private void sendWithRetries(MandrillMessage msg, String description) throws Mes
...
+      msg.setSubject("Your new All of Us Account");
+      msg.setFromEmail(workbenchConfigProvider.get().mandrill.fromEmail);
+      return msg;
+    } catch (IOException e) {
+      throw new MessagingException("Error reading in email");
+    }
+  }
+
+
+  private String buildEmailHtml(String password, User user) throws IOException {
```
calbach: nit: `buildWelcomeEmailHtml()`, this file is not limited to the welcome email

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -115,13 +115,49 @@ private void sendWithRetries(MandrillMessage msg, String description) throws Mes
...
+    CloudStorageService cloudStorageService = cloudStorageServiceProvider.get();
+    StringBuilder contentBuilder = new StringBuilder();
+    URL emailContent = Resources.getResource(WELCOME_RESOURCE);
+    Resources
+      .readLines(emailContent, StandardCharsets.UTF_8)
+      .forEach(s -> contentBuilder.append(s).append("\n"));
+    String string = contentBuilder.toString();
+    Map<String, String> replaceMap = new HashMap<>();
```
calbach: opt_nit: I still think the map builder would be better, not sure whether that comment was missed or just not accepted

api/src/main/resources/emails/welcomeemail/content.html
calbach: Please address or reply on the comment if you do not plan to do this.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
calbach: Unresolved

ui/src/app/views/account-creation-modals/component.html
```diff
@@ -0,0 +1,25 @@
...
+        <div class="modal-body">
+            <div class="form-section">
+                <label>Contact Email:</label>
+                <input id="change-contact-email" type="text" class="input" name="contact-email"
+                       [(ngModel)]="contactEmail">
+            </div>
+        </div>
+        <div class="modal-footer">
+            <button type="button" class="btn btn-outline" (click)="changingEmail=false">Cancel</button>
+            <button id="change_email" type="button" class="btn btn-primary" (click)="updateAndSend()">Apply</button>
```
calbach: Per my initial comments, clrLoading somewhere while this is sending?

api/config/config_stable.json
UrsaStutsman: Well, the images don't vary, so didn't think so.

api/config/config_prod.json
```diff
@@ -30,7 +30,8 @@
     "projectId": "all-of-us-rw-prod"
   },
   "admin": {
-    "adminIdVerification": "workbench-id-verifications@researchallofus.org"
+    "adminIdVerification": "workbench-id-verifications@researchallofus.org",
+    "loginUrl": "http:\/\/all-of-us-workbench.appspot.com/login"
```
UrsaStutsman: Is this the right url?

api/config/config_stable.json
UrsaStutsman: See above

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
UrsaStutsman: Believe I've resolved this

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
UrsaStutsman: Addressed below

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
UrsaStutsman: Addressed below

api/config/config_prod.json
rushtong: This is the right url: `https://all-of-us-rw-prod.appspot.com/login`

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: Will update this check to match using the gsuite user instead.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
calbach: I'm not following how this would lock them out. At this point they've yet to login to the account and currently the client code just calls updateContactEmail() -> resendWelcomeEmail(), effectively resulting in the same behavior I'm proposing. My suggestion was just to do these two things on the server rather than the client.

example:
- I own alice@gmail.com but not bob@gmail.com
- create account with contact of alice@gmail.com; password is now foo
- call updateContactEmail API directly with bob@gmail.com; password still foo
- Login using the welcome email password
- My user now has a verified email of bob@gmail.com

api/config/config_prod.json
```diff
@@ -30,7 +30,8 @@
     "projectId": "all-of-us-rw-prod"
   },
   "admin": {
-    "adminIdVerification": "workbench-id-verifications@researchallofus.org"
+    "adminIdVerification": "workbench-id-verifications@researchallofus.org",
```
calbach: Missing images bucket config

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: I better understand what you're proposing CH, I did misunderstand before. Making a combination call to do both will definitely work. My lock-out comment would only happen if we added the password reset to this method - my misunderstanding there.

In the example flow, step 3 would fail because the user had not yet signed in. This specific version of updating the contact email has the same restriction as sending the welcome email.

api/config/config_prod.json
UrsaStutsman: see above

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -420,16 +431,71 @@ private User initializeUserIfNeeded() {
...
+    String newEmail = updateContactEmailRequest.getContactEmail();
+    if (userNeverLoggedIn(googleUser)) {
+      try {
+        new InternetAddress(newEmail).validate();
+      } catch (AddressException e) {
+        log.log(Level.INFO, "Invalid email entered.");
+        return ResponseEntity.badRequest().build();
+      }
+    } else {
+      return ResponseEntity.status(HttpStatus.FORBIDDEN).build();
```
calbach: Would be good to return an error message here and below

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -420,16 +431,71 @@ private User initializeUserIfNeeded() {
...
+    return resetPasswordAndSendWelcomeEmail(username);
+  }
+
+  @Override
+  public ResponseEntity<Void> resendWelcomeEmail(ResendWelcomeEmailRequest resendRequest) {
+    String username = resendRequest.getUsername();
+    com.google.api.services.admin.directory.model.User googleUser =
+      directoryService.getUser(username);
+    User user = userDao.findUserByEmail(username);
+    if (userNeverLoggedIn(googleUser)) {
```
calbach: nit: I would probably invert this and return early, and keep the normal codepath unindented.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -420,16 +431,71 @@ private User initializeUserIfNeeded() {
...
+    }
+  }
+
+  private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.User googleUser) {
+    User user = userDao.findUserByEmail(googleUser.getPrimaryEmail());
+    return user.getFirstSignInTime() == null && googleUser.getChangePasswordAtNextLogin();
+  }
+
+  private ResponseEntity<Void> resetPasswordAndSendWelcomeEmail(String username) {
+    User user = userDao.findUserByEmail(username);
```
calbach: 3rd lookup of user. It may be cached by spring data but seems like it would be better to just pass it as a parameter.

calbach:
Also no need to block on my approval tomorrow, assuming @rushtong approves before I get on (but please address my comments).

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -420,16 +431,71 @@ private User initializeUserIfNeeded() {
...
+    user.setContactEmail(newEmail);
+    return resetPasswordAndSendWelcomeEmail(username);
+  }
+
+  @Override
+  public ResponseEntity<Void> resendWelcomeEmail(ResendWelcomeEmailRequest resendRequest) {
+    String username = resendRequest.getUsername();
+    com.google.api.services.admin.directory.model.User googleUser =
+      directoryService.getUser(username);
+    User user = userDao.findUserByEmail(username);
```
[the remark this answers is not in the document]
UrsaStutsman: I actually initially had been passing it in resetPasswordAndSendWelcomeEmail, had removed it, but now have added it back in. So leaving it in.

rushtong:
👍

PR #968, opened 2018-07-02
[risk=no] Fix leoApiUrl key in staging/stable environments

calbach:


calbach:
Submitting without api-build-test since it is unaffect by this.

calbach:
Submitting without api-build-test since it is unaffect by this.

PR #969, opened 2018-07-02
Domain Participant Count [risk=no]

SrushtiGangireddy:
1. ppi no show fix
2. added queries to add domain participant count in achilles result and concept


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,14 +262,19 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
         }
 
+        List<String> domainIds = null;
+        if (searchConceptsRequest.getDomain() != null) {
+            domainIds = DOMAIN_MAP.get(searchConceptsRequest.getDomain()).asList();
+        }
+
         if(searchConceptsRequest.getQuery() == null || searchConceptsRequest.getQuery().isEmpty()){
             List<Concept> concepts;
             if(standardConceptFilter == StandardConceptFilter.STANDARD_CONCEPTS){
```
biopete: Do we have tests for the various permutations of the searchRequest and filters?    We need tests for those. This one will probably fail if standard concept filter = STANDARD_OR_CODE_ID_MATCH  ... We will get all concepts I bet including non standard .

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
```diff
@@ -287,10 +288,11 @@ public void testGetDbDomains() throws Exception{
     @Test
     public void testConceptSearchWithEmptyQuery() throws Exception{
```
biopete: We need tests with the various standard concept filters too.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,14 +262,19 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
             standardConceptFilter = StandardConceptFilter.STANDARD_OR_CODE_ID_MATCH;
         }
 
+        List<String> domainIds = null;
+        if (searchConceptsRequest.getDomain() != null) {
+            domainIds = DOMAIN_MAP.get(searchConceptsRequest.getDomain()).asList();
+        }
+
         if(searchConceptsRequest.getQuery() == null || searchConceptsRequest.getQuery().isEmpty()){
```
biopete: One thing that may simply the logic and code here is just "If there is no query and the standard concept filter = STANDARD_OR_CODE_ID_MATCH   then set standard concept filter = STANDARD_CONCEPTS ... and the code continue.  or put an or below to catch the standard or code ...

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -790,4 +790,30 @@ group by dbd.concept_id,o.observation_source_concept_id,o.value_as_number,stratu
...
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "update \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.db_domain\`
-set concept_id=null where concept_id=0"
\ No newline at end of file
+set concept_id=null where concept_id=0"
+
+# Domain participant counts
+echo "Getting domain participant counts"
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"insert into \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.achilles_results\`
```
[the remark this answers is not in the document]
SrushtiGangireddy: We can join the db_domain table but since the fetch is from different tables (observation, measurement, drug_exposure and procedure_occurrence) i thought individual queries does the good job. Either way even if we join to the db_domain table and fetch the domain id we have to add whatever table suffix to fetch from.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -14,27 +14,30 @@
...
+            "c.vocabulary_id != 'PPI' and " +
             "c.standard_concept='' or c.standard_concept is null " +
             "and (c.count_value > 0 or c.source_count_value > 0) " +
             "order by c.count_value desc limit :maxResults ",
             nativeQuery = true)
-    List<Concept> findAllNonStandardConceptsOrderedByCount(@Param("maxResults") long maxResults);
+    List<Concept> findAllNonStandardConceptsOrderedByCount(@Param("maxResults") long maxResults,  @Param("domain") List<String> domain);
 
     @Query(value = "select c.* from concept c " +
             "where c.domain_id=:domain_id and " +
```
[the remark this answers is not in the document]
SrushtiGangireddy: This is just a single domain id passed and not the list.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Done.

public-api/src/test/java/org/pmiops/workbench/cdr/DataBrowserControllerTest.java
SrushtiGangireddy: Adding some.

api/db-cdr/generate-cdr/bq-schemas/concept.json
```diff
@@ -33,21 +33,6 @@
 	"mode": "REQUIRED", 
 	"name": "concept_code", 
 	"type": "STRING"
-	}, 
-	{
-	"mode": "REQUIRED",
```
danqrodney: Why is this being removed? 

These columns are a part of the standard OMOP schema; our code expects them to be there.

While I wouldn't do it in this PR, we should probably change our scripts to use the schema files here:

https://github.com/all-of-us/curation/blob/develop/data_steward/resources/fields

which do include these columns.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -167,11 +167,9 @@ echo "Inserting concept table data ... "
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\`
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, valid_start_date, valid_end_date, invalid_reason, count_value, prevalence)
```
danqrodney: See above, we should keep this.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -167,11 +167,9 @@ echo "Inserting concept table data ... "
...
+concept_code, count_value, prevalence, source_count_value)
```
danqrodney: See above

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -132,7 +132,8 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 (id, analysis_id, stratum_1, stratum_3, count_value, source_count_value)
 select 0, 3000 as analysis_id,
```
danqrodney: See above

api/db-cdr/vars.env
```diff
@@ -2,8 +2,8 @@ DB_CONNECTION_STRING=jdbc:mysql://db/cdr
 DB_DRIVER=com.mysql.jdbc.Driver
 DB_HOST=db
 DB_PORT=3306
-CDR_DB_NAME=cdr
-PUBLIC_DB_NAME=public
+CDR_DB_NAME=cdr20180702
```
danqrodney: I don't think this change should be checked in; our scripts are still using "cdr" and "public" by default, right?

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -262,14 +262,20 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
             List<Concept> concepts;
+            standardConceptFilter = (standardConceptFilter==StandardConceptFilter.STANDARD_OR_CODE_ID_MATCH ? StandardConceptFilter.STANDARD_CONCEPTS : standardConceptFilter);
             if(standardConceptFilter == StandardConceptFilter.STANDARD_CONCEPTS){
-                concepts = conceptDao.findAllStandardConceptsOrderedByCount(maxResults);
+                concepts = conceptDao.findAllStandardConceptsOrderedByCount(maxResults, domainIds);
```
danqrodney: Do we need to filter on anything besides domainIds here?

I'm wondering whether we could change the advanced search functionality to just make the filtering on keyword optional... then we might not need this whole if block.

public-api/src/main/java/org/pmiops/workbench/cdr/model/DbDomain.java
```diff
@@ -120,6 +121,20 @@ public DbDomain countValue(Long countValue) {
         return this;
     }
 
+    @Transient
+    public Long getParticipantCount() {
+        return participantCount;
+    }
+
+    public void setParticipantCount(Long participantCount) {
```
danqrodney: These aren't used anywhere?

api/db-cdr/generate-cdr/run-achilles-queries.sh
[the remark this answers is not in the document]
biopete: Good point. This looks good then. Keep it simple.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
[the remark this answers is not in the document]
biopete: Why is this function different than the ather findAll...that take a list ?

api/db-cdr/generate-cdr/bq-schemas/concept.json
biopete: We decided to remove them from the cloudsql tables a while back because they aren't used by workbench or data browser and it makes the cloudsql tables much smaller .  If i recall me and you brian all agreed it was ok to remove. We removed them from cloudsql a while back but forgot to pull them from the generated bq tables that we make our cloudsql from. 

Make sense ? So this is a bug fix ...

api/db-cdr/vars.env
biopete: Good catch. Thanks. I didn't see that one.

SrushtiGangireddy: Changed it. I generated a new data and was testing it.

api/db-cdr/generate-cdr/bq-schemas/concept.json
danqrodney: Removing them from Cloud SQL is different than removing them from BigQuery. And for at least BigQuery, if you don't use these columns in your query, they shouldn't impact the performance of queries on these tables. So I would not drop them here.

biopete: They still exist in the cdr .. Just not in the counts dbs we generate.  This is just in the intermediary bq tables. I thought we wanted those to be exactly like our cloudsql. And the import export relies on that fact. So if we put them back in we will have to tweak the import export csv code.   I see it as not something worth doing right now but if we need them in later add them back.  You're call though Dan. I don't mind either way. Unless its really hard to pick and choose what you export to a csv from a bq table :)

danqrodney: OK... I thought this was somehow being used as a substitute for the CDR itself. If it's just for intermediate tables, no objections here -- please disregard.

biopete: Cool. Nope. I would like to adopt the term count_cdrs or count_dbs to avoid this confusion. Or public_cdr .... It's just the data from the cdr available to public. We will probably allow them to download the csvs and mysqldump from the same public bucket for a cdr.

SrushtiGangireddy: Okay so I changed the schema for concept and concept_relationship to remove valide_start_date, valid_end_date and invalid_reason as the data import was weird without this removal. New data generation is under process.

api/db-cdr/generate-cdr/make-bq-data.sh
SrushtiGangireddy: The valid_start_date was being fetched to the count_value reason being inconsistent columns changed this query.

public-api/src/main/java/org/pmiops/workbench/cdr/model/DbDomain.java
SrushtiGangireddy: Added it to the dbdomain model. They are not used anywhere for now but would be.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: For now, the only filter is the domain ids.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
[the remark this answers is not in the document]
SrushtiGangireddy: This was the old function that was not being used, removed it.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
biopete: I don't follow. But we need to add the minCount to all of these functions right to implement that functionality ? 

This function is starting to get messy eh ? One thing we could do to simplify it is have a separate api call for findAll() .. And the uis could check for query string and call a different api to findAll.

SrushtiGangireddy: That sounds reasonable, it would make the searchConcepts api cleaner. I will do that.
Thanks.

SrushtiGangireddy:
Added new api call searchConceptsEmptyQuery.
Changed the order of standardConceptPredicates just a little.
Put in new concept and achilles_results csvs.
Added some tests to test the standardconceptfilter and mincount, maxresults, empty query.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Added.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -173,12 +174,72 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                };
+        // Return up to limit results, sorted in descending count value order.
+        Pageable pageable = new PageRequest(0, limit,
+                new Sort(Direction.DESC, "countValue"));
+        NoCountFindAllDao<Concept, Long> conceptDao = new NoCountFindAllDao<>(Concept.class,
+                entityManager);
+        return conceptDao.findAll(conceptSpecification, pageable);
+    }
+
+    public Slice<Concept> searchConcepts(StandardConceptFilter standardConceptFilter, List<String> vocabularyIds, List<String> domainIds, int limit, int minCount) {
```
danqrodney: Can't we combine the two searchConcepts methods with some conditional logic based on whether query is null?

Even if you expose two methods like this externally, it'd be nice if their implementation shared code...

SrushtiGangireddy: Removed the other function. Added if's while adding predicates to check if query is null or empty. Thanks Dan.

public-api/src/main/resources/public-api.yaml
```diff
@@ -138,6 +138,25 @@ paths:
           schema:
             $ref: "#/definitions/ConceptListResponse"
 
+  /v1/databrowser/searchConceptsEmptyQuery:
```
danqrodney: Seems like we could just have the one endpoint with an optional query parameter?

PR #973, opened 2018-07-03
[RW-781] Enforcing registered data access via FireCloud group membership [risk=low]

danqrodney:
… whenever using a CDR version.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
@@ -19,7 +19,6 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
-import org.pmiops.workbench.cdr.CdrVersionContext;
 import org.pmiops.workbench.cdr.cache.GenderRaceEthnicityConcept;
 import org.pmiops.workbench.cohortbuilder.ParticipantCounter;
 import org.pmiops.workbench.cohortbuilder.ParticipantCriteria;
@@ -84,7 +83,6 @@ public class CohortReviewController implements CohortReviewApiDelegate {
   private ParticipantCounter participantCounter;
   private ReviewTabQueryBuilder reviewTabQueryBuilder;
   private Provider<GenderRaceEthnicityConcept> genderRaceEthnicityConceptProvider;
-  private CdrVersionContext cdrVersionContext;
 
   /**
    * Converter function from backend representation (used with Hibernate) to
@@ -178,14 +176,12 @@ public class CohortReviewController implements CohortReviewApiDelegate {
                          BigQueryService bigQueryService,
                          ParticipantCounter participantCounter,
                          ReviewTabQueryBuilder reviewTabQueryBuilder,
-                         Provider<GenderRaceEthnicityConcept> genderRaceEthnicityConceptProvider,
-                         CdrVersionContext cdrVersionContext) {
+                         Provider<GenderRaceEthnicityConcept> genderRaceEthnicityConceptProvider) {
     this.cohortReviewService = cohortReviewService;
     this.bigQueryService = bigQueryService;
     this.participantCounter = participantCounter;
     this.reviewTabQueryBuilder = reviewTabQueryBuilder;
     this.genderRaceEthnicityConceptProvider = genderRaceEthnicityConceptProvider;
-    this.cdrVersionContext = cdrVersionContext;
   }
 
   /**
--- a/api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
@@ -30,7 +30,6 @@ import org.pmiops.workbench.db.model.Cohort;
 import org.pmiops.workbench.db.model.CohortReview;
 import org.pmiops.workbench.db.model.ParticipantCohortStatus;
 import org.pmiops.workbench.db.model.ParticipantCohortStatusKey;
-import org.pmiops.workbench.db.model.Workspace;
 import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.model.CohortStatus;
@@ -85,6 +84,7 @@ public class CohortReviewController implements CohortReviewApiDelegate {
   private ParticipantCounter participantCounter;
   private ReviewTabQueryBuilder reviewTabQueryBuilder;
   private Provider<GenderRaceEthnicityConcept> genderRaceEthnicityConceptProvider;
+  private CdrVersionContext cdrVersionContext;
 
   /**
    * Converter function from backend representation (used with Hibernate) to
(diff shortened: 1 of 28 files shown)


api/src/main/java/org/pmiops/workbench/cdr/CdrVersionContext.java
```diff
@@ -12,7 +17,35 @@
...
+      FireCloudService fireCloudService) {
+    this.configProvider = configProvider;
+    this.fireCloudService = fireCloudService;
+  }
+
+  public void setCdrVersion(CdrVersion version) {
+    if (configProvider.get().firecloud.enforceRegistered) {
+      // TODO: map data access level to authorization domain here (RW-943)
+      String authorizationDomain = configProvider.get().firecloud.registeredDomainName;
+      if (!fireCloudService.isUserMemberOfGroup(authorizationDomain)) {
```
calbach: I'm not sure I like encoding this business logic at this level. Up until this point, this class was pretty agnostic to where it was running and who was calling it, this brings in a bunch of deps as well.

A few other questions about this:
- How will this interact with the default CDR version? Is it possible some of those usages would sidestep auth domain enforcement?
- How will this work with DataBrowser, would they  just call the noAuthDomain variant?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -89,4 +89,14 @@ public GroupsApi groupsApi(@Qualifier(ALL_OF_US_API_CLIENT) ApiClient apiClient)
     api.setApiClient(apiClient);
     return api;
   }
+
+  @Bean(name = "user")
```
calbach: public constant

Also, if you want to match what I did for the cluster client:
https://github.com/all-of-us/workbench/blob/a9514f7773598239cbf57d7d909b20f8a5700377/api/src/main/java/org/pmiops/workbench/notebooks/NotebooksConfig.java#L21

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -89,4 +89,14 @@ public GroupsApi groupsApi(@Qualifier(ALL_OF_US_API_CLIENT) ApiClient apiClient)
     api.setApiClient(apiClient);
```
calbach: I'd thought there was a problem with having a "default" and a qualified version of the same class. In my case I had to qualify both.

I might be wrong on that, but I'd double check this is working.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -275,4 +280,19 @@ public void removeUserFromGroup(String email, String groupName) {
...
   }
+
+  @Override
+  public boolean isUserMemberOfGroup(String groupName) {
+    return retryHandler.run((context) -> {
+      // There is no endpoint in FireCloud for checking whether a user is a member of a particular
+      // group; so instead, fetch all the group memberships. (There won't be that many for our
+      // users anyway.)
+      for (ManagedGroupAccessResponse group : endUserGroupsApiProvider.get().getGroups()) {
+        if (group.getGroupName().equals(groupName) && group.getRole().equals("member")) {
```
calbach: I'd probably reverse the `.equals()` ordering here to reduce risk of NPEs (obvious for the constant, but I'd also say the API response is less trusted than the function parameter `groupName`).

calbach: What are the other roles? Presumably there are more privileged roles - should we be accepting those as well?

api/src/main/java/org/pmiops/workbench/cdr/CdrVersionContext.java
danqrodney: As discussed over Slack: I created a separate CdrVersionService which does this logic.

The default CDR version only indicates what CDR version to use when e.g. creating new workspaces; you still have to set the CDR version in CdrVersionContext before you can read from it.

Data browser doesn't presently use CdrVersionContext, but in future it can (without using CdrVersionService or FireCloud.)

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
danqrodney: Explicitly named both with constants.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
danqrodney: Done.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
danqrodney: Done.

danqrodney: I don't know what the other roles are, but we don't use them and probably shouldn't accept them... the expectation is the user had addUserToGroup() called before this gets invoked. 

Changed to use a constant here for "member".

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -210,11 +208,8 @@ public ParticipantCohortAnnotation apply(org.pmiops.workbench.db.model.Participa
 
     Cohort cohort = cohortReviewService.findCohort(cohortId);
     //this validates that the user is in the proper workspace
-    Workspace workspace = cohortReviewService.validateMatchingWorkspace(workspaceNamespace,
-      workspaceId, cohort.getWorkspaceId(), WorkspaceAccessLevel.WRITER);
-
-    CdrVersionContext.setCdrVersion(workspace.getCdrVersion());
-
+    cohortReviewService.validateMatchingWorkspace(workspaceNamespace,
```
calbach: I find setting the CDR version to be a surprising side-effect of this method name. Perhaps rename or else do not embed that logic there.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
```diff
@@ -59,6 +62,7 @@ public Userinfoplus userInfo(UserAuthentication userAuthentication) {
   @Bean
   @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
   public User user(UserAuthentication userAuthentication) {
+    logger.info("HERE = " + userAuthentication.getUser());
```
calbach: revert

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -160,6 +160,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
 
     // TODO: setup this in the context, get rid of log statement
     log.log(Level.INFO, "{0} logged in", userInfo.getEmail());
+    log.log(Level.INFO, "user = " + user);
```
calbach: revert

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
danqrodney: Done.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
danqrodney: Done.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
danqrodney: Renamed.

PR #974, opened 2018-07-03
[RW-860][RW-859][Risk=low] Blrubenstein/demos

blrubenstein:
Replace this this template with your PR description.

* **no**: None 

Please also:

* Get thumbs from reviewer(s)


blrubenstein:
This also fixes the notebook title on cards, since I was noticing that was messed up currently.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -35,6 +35,13 @@ public String readMandrillApiKey() {
     return mandrillKeys.getString("api-key");
   }
 
+  public void copyDemoNotebook(String workspaceBucket)  {
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+    BlobId demoNotebookId = storage.get(getDemosBucketName(), "demo-notebook.ipynb").getBlobId();
```
calbach: I think we should probably check this into source control, then deploy it as a GCS artifact along with our Jupyter user initialization script. (or alternatively, skip GCS and just read it server side from the server resources and write it directly to GCS).

We already have a separate "deploy-gcs-artifacts" command which allows us to push these without a full release. There's a separate question of whether it's actually desirable to push these demo notebooks out-of-band, i.e. bypassing our normal QA process (in which case, reading it as a server resource is the most hermetic option).

blrubenstein: Fritz actually asked me to make sure they could be pushed out-of-band, because I believe in the long run this is going to be written by comms and scientist folks, and not by us, so they should be able to push/update them.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -47,6 +54,10 @@ String getCredentialsBucketName() {
     return configProvider.get().googleCloudStorageService.credentialsBucketName;
   }
 
+  String getDemosBucketName() {
```
rushtong: minor nit: this and the methods above and below this can be private.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -388,6 +398,29 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    try {
+      // TODO Make this a pre-check within a transaction?
+      dbCohort = cohortDao.save(dbCohort);
+    } catch (DataIntegrityViolationException e) {
+      // TODO The exception message doesn't show up anywhere; neither logged nor returned to the
+      // client by Spring (the client gets a default reason string).
+      throw new BadRequestException(String.format(
+          "Cohort \"/%s/%s/%d\" already exists.",
+          dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getWorkspaceId(), dbCohort.getCohortId()));
+    }
```
rushtong: I do think this, and the cohort controller version, should be pulled into a service class instead of copied.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -4,13 +4,10 @@
...
 import static junit.framework.TestCase.fail;
+import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyListOf;
 import static org.mockito.Matchers.anyString;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
calbach: revert *

api/demos/demo-cohort.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "demo-cohort",
+  "description": "demo",
+  "criteria": {"includes":[{"id":"includes_twr4se71q","items":[{"id":"items_s64k36qwq","type":"ICD9","searchParameters":[{"parameterId":"param6","name":"Cholera","value":"001","type":"ICD9","subtype":"","group":true,"domain":""}],"modifiers":[]}]}],"excludes":[]}
```
calbach: If we're expecting people to hand-modify this, should format this JSON onto more than one line.

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -140,16 +145,32 @@ public void setUp() throws Exception {
...
+        cdrVersionDao, cohortDao, userDao, userProvider, fireCloudService, cloudStorageService, CLOCK,
         userService);
     stubGetWorkspace(WORKSPACE_NAMESPACE, WORKSPACE_NAME, "bob@gmail.com",
         WorkspaceAccessLevel.OWNER);
+    JSONObject demoCohort = new JSONObject();
+    demoCohort.put("name", "demo");
+    demoCohort.put("description", "demo");
+    demoCohort.put("criteria", createDemoCriteria());
+    when(cloudStorageService.readDemoCohort()).thenReturn(demoCohort);
+    when(cloudStorageService.readInvitationKey()).thenReturn("Dummy Value");
```
calbach: Why are the ApiKey read methods needed?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -388,6 +398,29 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    dbCohort.setCreator(userProvider.get());
+    dbCohort.setWorkspaceId(dbWorkspace.getWorkspaceId());
+    dbCohort.setCreationTime(now);
+    dbCohort.setLastModifiedTime(now);
+    dbCohort.setVersion(1);
+    try {
+      // TODO Make this a pre-check within a transaction?
+      dbCohort = cohortDao.save(dbCohort);
+    } catch (DataIntegrityViolationException e) {
+      // TODO The exception message doesn't show up anywhere; neither logged nor returned to the
```
calbach: What is this comment about? Is this talking about the DataIntegrity exception? Can you just log it right here?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -388,6 +398,29 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
 
     dbWorkspace = workspaceService.getDao().save(dbWorkspace);
+
+    JSONObject demoCohort = cloudStorageService.readDemoCohort();
+    // TODO: Make this a service method to avoid duplication with CohortsController
+    Cohort dbCohort = new Cohort();
+    dbCohort.setName(demoCohort.getString("name"));
+    dbCohort.setDescription(demoCohort.getString("description"));
+    dbCohort.setCriteria(demoCohort.get("criteria").toString());
+    dbCohort.setType("AoU_Discover");
```
calbach: Where did this string come from and what does it mean?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -388,6 +398,29 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    dbCohort.setCreationTime(now);
+    dbCohort.setLastModifiedTime(now);
+    dbCohort.setVersion(1);
+    try {
+      // TODO Make this a pre-check within a transaction?
+      dbCohort = cohortDao.save(dbCohort);
+    } catch (DataIntegrityViolationException e) {
+      // TODO The exception message doesn't show up anywhere; neither logged nor returned to the
+      // client by Spring (the client gets a default reason string).
+      throw new BadRequestException(String.format(
```
calbach: Hmm.. do we actually need to handle this case specifically? I'd guess there are many other exceptions that are more likely than this. The explanation given here implies someone else inserted a cohort with a conflicting column, a few milliseconds after we created the workspace. Did you observe this exception?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
blrubenstein: Doing. I don't know when or how my settings keep getting reverted, because I know for a fact last time you commented this, I changed my settings to never do this...

api/demos/demo-cohort.json
blrubenstein: I can do that. For now this is also mostly a placeholder until we get real demo cohorts... (And maybe an export as JSON functionality for a cohort definition?)

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: I am unsure. This is directly copied over from the cohortscontroller.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: Currently, nothing. I believe its for future proofing. However, I realized it may actually be useful for us here, so I should have something for it soon. (It is set hard coded in the UI right now on all cohorts)

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: Similarly copied from the cohorts controller. I did, however, get it once when I made two demo cohort files (Just pushed. Should have let you know more work ended up needing to happen), so that was useful for my development of demo files at least!

blrubenstein:
@calbach @rushtong I think I've addressed comments and this is good for another look! As an FYI, I told Greg, but for CH, this now handles multiple demos of each type. (And types the cohorts as demos, so if we want to filter demo cohorts out at some point, we can do that relatively easily)

api/libproject/devstart.rb
```diff
@@ -1039,6 +1039,7 @@ def deploy_gcs_artifacts(cmd_name, args)
   op.parse.validate
   gcc.validate
   common.run_inline %W{gsutil cp scripts/setup_notebook_cluster.sh gs://#{gcc.project}-scripts/setup_notebook_cluster.sh}
+  common.run_inline %W{gsutil cp demos/* gs://#{gcc.project}-demos/}
```
calbach: Since you switched this to multiple notebooks/cohorts, may want to do rsync -d instead to handle removals and file renames.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
calbach: Well, if neither of us knows, please just remove it.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -35,6 +39,62 @@ public String readMandrillApiKey() {
...
+  public void copyAllDemoNotebooks(String workspaceBucket)  {
+    Storage storage = StorageOptions.getDefaultInstance().getService();
+    Bucket demoBucket = storage.get(getDemosBucketName());
+    Page<Blob> allBlobs = demoBucket.list();
+    List<Blob> notebookFiles = new ArrayList<>();
+    notebookFiles.addAll(getFileTypeFromPage(allBlobs, ".ipynb"));
+    while (allBlobs.hasNextPage()) {
+      allBlobs = allBlobs.getNextPage();
+      notebookFiles.addAll(getFileTypeFromPage(allBlobs, ".ipynb"));
+    }
```
rushtong: This whole block feels like it could be much simpler in a stream. I haven't delved into it in the IDE, but will give that a shot.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -35,6 +39,62 @@ public String readMandrillApiKey() {
...
+    notebookFiles.forEach((Blob blob) -> {
+      BlobId demoNotebookId = blob.getBlobId();
+      BlobId targetLocation = BlobId.of(workspaceBucket, "notebooks/" + demoNotebookId.getName());
+      copyBlob(demoNotebookId, targetLocation);
+    });
```
rushtong: Previous comment applies this ... It feels like we should be able to `demoBucket.list() -> stream -> filter on name -> new blob -> copy blob.`

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -35,6 +39,62 @@ public String readMandrillApiKey() {
...
+    Iterable<Blob> iter = page.iterateAll();
+    iter.forEach((Blob blob) -> {
+      if (blob.getBlobId().getName().endsWith(extension)) {
+        fileList.add(blob);
+      }
+    });
+    return fileList;
+  }
+
+  private List<JSONObject> getJsonAndFilterByTypeFromPage(Page<Blob> page, String type) {
```
rushtong: `type` is always cohort here. My preference for comparisons like this is to use enums instead of raw strings. But, I'm not sure if this type is initially set on the cohort object, or if this is the same thing as the `type` field on the cohort json object (I'm guessing the latter?). If it's also set in code somewhere else, that means more refactoring, so I'm open to letting this go.

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -140,16 +146,33 @@ public void setUp() throws Exception {
...
         userService);
     stubGetWorkspace(WORKSPACE_NAMESPACE, WORKSPACE_NAME, "bob@gmail.com",
         WorkspaceAccessLevel.OWNER);
+    JSONObject demoCohort = new JSONObject();
+    demoCohort.put("name", "demo");
+    demoCohort.put("description", "demo");
+    demoCohort.put("type", "demo");
+    demoCohort.put("criteria", createDemoCriteria());
+    List<JSONObject> demoCohorts = new ArrayList<>();
+    demoCohorts.add(demoCohort);
```
rushtong: these two lines can be condensed
`List<JSONObject> demoCohorts = Collections.singletonList(demoCohort);`

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
blrubenstein: This is actually something that Fritz wanted when I talked to him so that we could add more json demos that are different things later (future proofing)

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -204,9 +209,26 @@ public void setUp() {
     cdrVersion = cdrVersionDao.save(cdrVersion);
     cdrVersionId = Long.toString(cdrVersion.getCdrVersionId());
 
+    JSONObject demoCohort = new JSONObject();
+    demoCohort.put("name", "demo");
+    demoCohort.put("description", "demo");
+    demoCohort.put("type", "demo");
+    demoCohort.put("criteria", createDemoCriteria());
+    List<JSONObject> demoCohorts = new ArrayList<>();
+    demoCohorts.add(demoCohort);
```
rushtong: Collections.singletonList is your friend here.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
rushtong: Was slacking with Bru about this. My first stab at streaming this would look like 
```
    StreamSupport
        .stream(demoBucket.list().getValues().spliterator(), false)
        .filter(blob -> blob.getBlobId().getName().endsWith(".ipynb"))
        .map(blob -> new ImmutableMap.Builder<BlobId, BlobId>()
                .put(blob.getBlobId(), BlobId.of(workspaceBucket, "notebooks/" + blob.getBlobId().getName())).build())
        .forEach(pair -> copyBlob(pair.keySet().asList().get(0), pair.values().asList().get(0)));
```
But I think that can be cleaned up some more as well. There should be something simpler than the `ImmutableMap` to use.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -35,6 +42,31 @@ public String readMandrillApiKey() {
...
+    Bucket demoBucket = storage.get(getDemosBucketName());
+
+    List<JSONObject> cohortFiles = StreamSupport
+        .stream(demoBucket.list().getValues().spliterator(), false)
+        .filter(blob -> blob.getBlobId().getName().endsWith(".json"))
+        .map(blob -> new JSONObject(new String(blob.getContent()).trim()))
+        .filter(jsonObject -> jsonObject.getString("type").equalsIgnoreCase("cohort"))
+        .map(jsonObject -> jsonObject.getJSONObject("cohort"))
+        .collect(Collectors.toList());
+    return cohortFiles;
```
rushtong: nit: you could return the result of the stream instead of assigning to `cohortFiles`.

PR #976, opened 2018-07-04
[RW-948][risk=no] Use a less confusing ID verification list name

calbach:
I realize from the PR history that using this list in all environments was intentional. That said, it's very confusing to find this in our configs, and most reasonable people would assume this was a bug. Fixing this now to avoid future developers burning time looking into this.

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -31,7 +31,7 @@
     "projectId": "all-of-us-workbench-test"
   },
   "admin": {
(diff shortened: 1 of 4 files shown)


calbach:
No one seems to have been monitoring the old one either. In most cases, people requesting ID verification are probably devs who could verify themselves, so outside of stable this isn't going to be super useful to monitor.

New list: https://groups.google.com/a/fake-research-aou.org/forum/#!forum/manual-id-verification-requests

PR #981, opened 2018-07-06
[RW-844] Fixing pagination bug in cohort materialization. [risk=no]

danqrodney:
Ideally we'd change cohort materialization logic to not mutate the incoming request, but this is a bigger change... for now, just looking to make this fix.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
+++ b/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
@@ -170,6 +170,12 @@ public class CohortMaterializationService {
             String.format("Use of pagination token %s with new parameter values", paginationToken));
       }
     }
+    // Grab the next pagination token here, because searchRequest can be mutated during query
+    // execution, which will make the hash of SearchRequest no longer match; see
+    // [RW-844].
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -170,6 +170,12 @@ public MaterializeCohortResponse materializeCohort(@Nullable CohortReview cohort
             String.format("Use of pagination token %s with new parameter values", paginationToken));
       }
     }
+    // Grab the next pagination token here, because searchRequest can be mutated during query
+    // execution, which will make the hash of SearchRequest no longer match; see
+    // [RW-844].
+    // TODO: consider pagination based on cursor / values rather than offset
+    PaginationToken newToken = PaginationToken.of(offset + pageSize, paginationParameters);
```
calbach: nit: I'd probably call this nextToken

calbach: nit: Right now this depends on the utility eagerly hashing the input parameters, else the bug wouldn't be fixed. I had to click through to verify that's what was happening.

I'd probably do the `toBase64()` here instead to remove this implementation dependency.

PR #982, opened 2018-07-06
[RW-801][risk=no] api endpoint for all Physical Measurement criteria

freemabd1:
Physical Measurements consist of the follow types:

1. Blood Pressure
2. Heart Rate
3. Heart Rhythm
4. Height
5. Weight
6. BMI
7. Waist Circumference
8. Hip Circumference
9. Pregnancy
10. Wheel Chair User


rushtong:
Question about the changelog changes. From what I remember about liquibase - it generates an ID for each change log file and skips them on subsequent runs. I'm guessing we bypass that and regen the db every time?

freemabd1:
So locally the cdr db schema changelog generates an ID and skips them on subsequent runs unless you do docker-clean. Then the all get run again. Any of the cdr db data changelog files run everytime you run run-local-data-migrations. Non of the liquibase cdr db stuff runs from circleci in test/staging/stable. In those envs the cdr db is managed by the generate-cdr goals in the readme file

rushtong:
Gotcha, thanks.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+  private static final String BMI = "BMI";
+  private static final String WC = "WC";
+  private static final String HC = "HC";
+  private static final String HR = "HR";
+  private static final String PREG = "PREG";
+  private static final String WHEEL = "WHEEL";
+  private static final String CONCEPT_ID = "conceptId";
+  private static final List<String> PM_TYPES_WITH_ATTR =
+    Arrays.asList(BLOOD_PRESSURE, HEART_RATE_DETAIL, HEIGHT, WEIGHT, BMI, WC, HC);
+  ImmutableMap<String, Object> exceptionText = ImmutableMap.<String, Object>builder()
```
rushtong: Any reason not to use `ImmutableMap.<String, String>` here?

freemabd1: Nope... changed.

api/src/main/resources/cb_search_api.yaml
```diff
@@ -239,6 +241,13 @@ definitions:
           for leaves in the tree. Parents don't have domain ids and concept
           id will be used in the case that a parent is selectable.
         type: string
+      attribute:
+        description: specifies if extra attributes are needed
```
danqrodney: What does this mean? How does this relate to predefinedAttributes?

api/src/main/java/org/pmiops/workbench/api/CohortBuilderController.java
```diff
@@ -46,6 +51,8 @@
             new Function<Criteria, org.pmiops.workbench.model.Criteria>() {
                 @Override
                 public org.pmiops.workbench.model.Criteria apply(Criteria criteria) {
+                    Type listType = new TypeToken<ArrayList<Attribute>>(){}.getType();
```
danqrodney: Are we going to need to be able to query on these attributes? 

Representing them as JSON in a string (instead of child rows) will make that difficult. If this is purely extra information returned to the client, then this may be fine.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+        + exceptionText.get(parameter.getSubtype()) + ".");
+    }
+  }
+
+  @Override
+  public FactoryKey getType() {
+    return FactoryKey.PM;
+  }
+
+  interface validationUtil extends Function<Attribute, Boolean> {
```
danqrodney: I think you can just use Predicate<Attribute>?

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+        && attribute.getOperands().size() == 1
+        && StringUtils.isNumeric(attribute.getOperands().get(0)))
+          || (attribute.getOperator() != null
+        && attribute.getOperator().equals(Operator.BETWEEN)
+        && attribute.getOperands().size() == 2
+        && StringUtils.isNumeric(attribute.getOperands().get(0)))
+        && StringUtils.isNumeric(attribute.getOperands().get(1));
+    }
+
+    default validationUtil and(validationUtil other) {
```
danqrodney: Predicates.and

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+    .put(WEIGHT, "Weight")
+    .put(BMI, "BMI")
+    .put(WC, "Waist Circumference")
+    .put(HC, "Hip Circumference")
+    .put(HR, "Heart Rate")
+    .put(PREG, "Pregnancy")
+    .put(WHEEL, "Wheel Chair User")
+    .build();
+
+  private static final String BP_INNER_SQL_TEMPLATE =
```
danqrodney: Is it possible to find every person who has a physical measurement set irrespective of range?

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+    "select person_id, measurement_date from `${projectId}.${dataSetId}.measurement`\n" +
```
danqrodney: Why does this return measurement_date where the rest don't? How does this get used when selecting participants?

api/src/main/resources/cb_search_api.yaml
```diff
@@ -388,12 +400,20 @@ definitions:
...
         type: string
+        description: the name of the attribute
+      operator:
+        $ref: "#/definitions/Operator"
       operands:
         type: "array"
         items:
           type: string
+      conceptId:
+        description: The concept id that maps to concept table.
```
danqrodney: So this is a concept ID for a concept related to the concept on the containing SearchParameter? Can you give an example in the documentation?

api/src/main/resources/cb_search_api.yaml
freemabd1: This flag signals to the UI that this criteria will be using a form to collect attribute(operator and operands) data from the user.

api/src/main/resources/cb_search_api.yaml
freemabd1: So Blood Pressure has 2 concept ids(systolic and diastolic). Therefore these 2 values can't reside in the concept_id column in the criteria table. So I've put a conceptId property on the Attribute object to account for this issue. It's not optimal, but it was the least impact to get this story complete. Let me know if you have any better ideas?

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
freemabd1: This PR will now allow for PM's that take user input values to find a match regardless of range. Makes use of an ANY Operator.

api/src/main/resources/cb_search_api.yaml
danqrodney: OK... still not quite following why this needs to be on Attribute. Seems like maybe you could just special case the BP criteria on the client and server?

api/src/main/resources/cb_search_api.yaml
danqrodney: OK. You might rename this to "attributesRequired" or something.

freemabd1: Renamed to hasAttributes

PR #984, opened 2018-07-09
[RW-859][RW-860] Separate demos from other gcs artifacts

blrubenstein:
I think that we should deploy demos to GCS separately from other artifacts, I just felt kind of weird pushing our startup script at the same time as demos. They are kind of separate things. (In that one is getting created by developers, and the other theoretically by scientists)

* **no**: None 

Please also:

* Get thumbs from reviewer(s)

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1267,6 +1267,8 @@ def deploy(cmd_name, args)
     common.status "Pushing GCS artifacts..."
     deploy_gcs_artifacts(cmd_name, %W{--project #{ctx.project}})
 
+    deploy_gcs_demos(cmd_name, %W{--project #{ctx.project}})
+
     # Keep the cloud proxy context open for the service account credentials.
     deploy_args = %W{
       --project #{gcc.project}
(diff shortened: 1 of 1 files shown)


api/libproject/devstart.rb
```diff
@@ -1053,6 +1051,24 @@ def deploy_gcs_artifacts(cmd_name, args)
...
+  op.parse.validate
+  gcc.validate
+  common.run_inline %W{gsutil rm gs://#{gcc.project}-demos/**}
+  common.run_inline %W{gsutil cp demos/* gs://#{gcc.project}-demos/}
+end
+
+Common.register_command({
+  :invocation => "deploy-gcs-demos",
+  :description => "Deploys any GCS demos associated with this environment.",
+  :fn => ->(*args) { deploy_gcs_demos("deploy-gcs-demos", args) }
```
calbach: Sounds fine to have this as a separate command, but wouldn't we want to push these automatically? Do we expect people to always manually run this in all environments any time we change a demo notebook?

blrubenstein: I think it is probably perfectly fine to push these automatically! Are they currently deployed as part of a circle workflow? (If so, have the circle files moved? I can't seem to find them)

I am not entirely certain what our plan is going to be for deployment long term though, as I think that is going to be a little bit played by ear... this is mostly temporary I think, to get the wiring set up to read demos from a google bucket. Eventually these are probably going to be stored and uploaded in/from a separate place, and we're just going to read them in.

calbach: I would just make the deploy command call this method, I don't think it's worth explicitly including in the Circle configs.

blrubenstein: Should be good now!

PR #990, opened 2018-07-10
[RW-803][risk=no] Add physical measurements search in CB

dolbeew:
Enables the Physical Measurements item in the Criteria dropdown and implements search using the wizard in CB


ui/src/app/cohort-search/constant.ts
```diff
@@ -19,3 +19,13 @@ export const DOMAIN_TYPES = [
     // { name: 'Vitals',       type: 'vitals' },
     // { name: 'Temporal',     type: 'temporal' }
 ];
```
freemabd1: Would it make sense to add a property here for each domain type that can be used for each type in the code base instead of using Strings? For example: 'PM'?

ui/src/app/cohort-search/attributes-page/attributes-page.component.html
```diff
@@ -0,0 +1,51 @@
...
+        <input type="number" [name]="'valueB'+i" [ngModel]="attr.operands[1]" required>
+        <span>{{units[node.get('subtype')]}}</span>
+      </div>
+    </div>
+    <div class="count-preview">
+      <button type="button"
+              class="btn"
+              (click)="requestPreview(attrForm)"
+              [disabled]="!attrForm.valid"
+      >
```
blrubenstein: Nit: I'd move this to the line above (Just think it might be a bit more clear)

ui/src/app/cohort-search/constant.ts
```diff
@@ -19,3 +19,13 @@ export const DOMAIN_TYPES = [
...
+
+export const PM_UNITS = {
+    HEIGHT: 'cm',
+    WEIGHT: 'kg' ,
+    BMI: '',
+    WC: 'cm',
+    HC: 'cm',
+    BP: '',
+    'HR-DETAIL': 'beats/min'
```
blrubenstein: Nit: this appears to follow a different format?

ui/src/app/cohort-search/constant.ts
```diff
@@ -19,3 +19,13 @@ export const DOMAIN_TYPES = [
     // { name: 'Vitals',       type: 'vitals' },
     // { name: 'Temporal',     type: 'temporal' }
 ];
+
+export const PM_UNITS = {
```
blrubenstein: Just wanted to check that these units will always be this? Just because I know Chris Fonnesbeck had to do some unit conversions at some point because things weren't being recorded consistently.

dolbeew: We've been told they should be consistent, but I'm sure there's always a possibility the data is using something different. AFAIK there's no way to verify the units for the data.

ui/src/app/cohort-search/constant.ts
dolbeew: The '-' caused an error, so I put HR-DETAIL in single quotes. Maybe do the same for all properties for consistency?

ui/src/app/cohort-search/constant.ts
dolbeew: Yes, that makes sense. I'll create a new story for that.

ui/src/app/cohort-search/attributes-page/attributes-page.component.html
dolbeew: Done!

ui/src/app/cohort-search/constant.ts
blrubenstein: Yeah, I think doing it consistently would be good.

ui/src/app/cohort-search/constant.ts
blrubenstein: Sounds reasonable!

ui/src/app/cohort-search/attributes-page/attributes-page.component.html
```diff
@@ -0,0 +1,51 @@
...
+        <input type="number" [name]="'valueB'+i" [ngModel]="attr.operands[1]" required>
+        <span>{{units[node.get('subtype')]}}</span>
+      </div>
+    </div>
+    <div class="count-preview">
+      <button type="button"
+              class="btn"
+              [disabled]="!attrForm.valid"
+              (click)="requestPreview(attrForm)"
+      >
```
blrubenstein: Nit: placement

ui/src/app/cohort-search/constant.ts
dolbeew: Done!

ui/src/app/cohort-search/attributes-page/attributes-page.component.html
dolbeew: Okay, I see. I moved the wrong line the first time. Should be good now!

PR #995, opened 2018-07-10
[RW-885][risk=no] Adding ability to search visit occurrence

osgsusan:
Adding ability to search visit occurrence


api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/VisitsQueryBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+        parentList.add(parameter.getConceptId());
+      } else {
+        childList.add(parameter.getConceptId());
+      }
+    }
+
+    // Collect all parent type queries to run them together
+    if (!parentList.isEmpty()) {
+      String namedParameter = "visit" + getUniqueNamedParameterPostfix();
+      queryParams.put(namedParameter, QueryParameterValue.array(parentList.toArray(new Long[0]), Long.class));
```
freemabd1: nit: I would use java 8 here: parentList.toArray(Long[]::new)

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/VisitsQueryBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+      String parentSql = VISIT_SELECT_CLAUSE_TEMPLATE +
+        VISIT_PARENT_CLAUSE_TEMPLATE.replace("${parentIds}", "@" + namedParameter);
+      queryParts.add(parentSql);
+    }
+
+    // Collect all child type queries to run them together
+    if (!childList.isEmpty()) {
+      String namedParameter = "visit" + getUniqueNamedParameterPostfix();
+      queryParams.put(namedParameter, QueryParameterValue.array(childList.toArray(new Long[0]), Long.class));
```
freemabd1: nit: I would use java 8 here: parentList.toArray(Long[]::new)

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/VisitsQueryBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+    List<String> queryParts = new ArrayList<>();
+    Map<String, QueryParameterValue> queryParams = new HashMap<>();
+
+    List<Long> parentList = new ArrayList<>();
+    List<Long> childList = new ArrayList<>();
+    for (SearchParameter parameter : inputParameters.getParameters()) {
+      if (parameter.getGroup()) {
+        parentList.add(parameter.getConceptId());
+      } else {
+        childList.add(parameter.getConceptId());
```
freemabd1: Should we have validation of some sort to check that conceptId is not null? And maybe throw a BadRequestException? If so, then maybe we should add a test case as well?

osgsusan: We should be consistent, sure. done.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/VisitsQueryBuilder.java
osgsusan: done

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/VisitsQueryBuilder.java
osgsusan: done

PR #1001, opened 2018-07-12
Turning on JupyterLab in new clusters [risk=low]

danqrodney:
Note that this does not actually change workbench to take the user into notebooks in JupyterLab... probably we'll want https://github.com/DataBiosphere/leonardo/issues/474 to be fixed before we do that.


danqrodney:
(I have confirmed that localization works with JupyterLab enabled in Leo dev now, and it should in prod, too.)

PR #1006, opened 2018-07-13
[RW-974][risk=no] reorder csv columns

freemabd1:
Specifically swap the order of **concept_id, domain_id** to **domain_id, concept_id**


api/db-cdr/csv/visit_criteria.csv
```diff
@@ -1,82 +1,82 @@
...
-316473,316411,VISIT,,,Rural Health Clinic,0,1,,8761,Visit,0,
-316474,316409,VISIT,,,School,0,1,,8537,Visit,0,
-316475,316402,VISIT,,,Skilled Nursing Facility,0,1,,8863,Visit,0,
-316476,316409,VISIT,,,Temporary Lodging,0,1,,8602,Visit,0,
-316477,316405,VISIT,,,Tribal 638 Free-standing Facility,0,1,,8941,Visit,0,
-316478,316405,VISIT,,,Tribal 638 Provider-based Facility,0,1,,8960,Visit,0,
-316479,316411,VISIT,,,Urgent Care Facility,0,1,,8782,Visit,0,
-316480,316411,VISIT,,,Walk-in Retail Health Clinic,0,1,,38003620,Visit,0,
\ No newline at end of file
+"id","parent_id","type","subtype","code","name","is_group","is_selectable","est_count","domain_id","concept_id","has_attribute","predefined_attributes"
```
osgsusan: What's the deal with all of the string quotations added?

freemabd1: When generating these csv files with sequelPro it escapes everything with ". That way if any columns have special characters they don't mess up an import. Just to be on the safe side of things.

osgsusan: Ok.

PR #1012, opened 2018-07-16
[][RISK=NO] Home page responsive

NehaBroad:
This PR contains comments from Lou for Home page and also UI for  the scroller. The logic to abstract card logic from cohort and notebook pages is not a part of this.


ui/src/app/views/homepage/component.css
```diff
@@ -230,3 +236,72 @@
...
+}
+
+@media (min-width: 1800px) {
+  .left {
+    margin-left: 13rem;
+  }
+  .single {
+    margin-left: 13rem;
+  }
+}
```
blrubenstein: Nit: Newline at end of file.

ui/src/app/views/homepage/component.html
```diff
@@ -98,15 +98,23 @@
...
+                Create a <br/> New Workspace <br/>
+                <clr-icon shape="plus-circle" class="plus-circle"></clr-icon>
+              </div>
             </div>
           </div>
+          <div class="col-md-9">
+            <div class="row links" *ngIf="card.length>0" style="margin-left:4rem;margin-bottom:-2rem;">
+              Your recent work
+            </div>
+            <app-scroller [cards]=card></app-scroller>
```
blrubenstein: Nit: Can we make array variables plural? I think it would make the code easier to understand.

ui/src/app/views/homepage/component.html
```diff
@@ -98,15 +98,23 @@
...
+
```
blrubenstein: Nit: Extra line

ui/src/app/views/homepage/component.ts
```diff
@@ -52,6 +52,7 @@ export class HomepageComponent implements OnInit, OnDestroy {
           'dolore. Mirum est notare, quam littera gothica quam nunc.',
           icon: '/assets/icons/explore.svg'
       }];
+  card: any[] = [];
```
blrubenstein: See above, I think it would be good to name arrays plural (`cards` in this case)

ui/src/app/views/card/component.html
```diff
@@ -0,0 +1,58 @@
+<div class="card item-card">
+  <div class="header-area">
+    <clr-dropdown [clrCloseMenuOnItemClick]="false">
+      <button type="button" clrDropdownTrigger>
+        <clr-icon shape="ellipsis-vertical" class="ellipsis-vertical"></clr-icon>
+      </button>
+      <clr-dropdown-menu *clrIfOpen>
```
blrubenstein: Is this the same dropdown menu for all objects? I did just see the below if statements, but I don't think we have any of the functionality for notebooks yet. 

Also, currently, when you click on a notebook title it opens it. I don't see that here? Are we removing that functionality? If not, we should probably add a `Open in Jupyter` option to the dropdown.

NehaBroad: My aim for this PR was to get the UI content up without the links to be functional since the way recent work list is to be retrieved is still to be decided and also there is a story in backlog to make notebooks event functional. Having said that i'll be taking a look at opening notebook functionality tomorrow and if its something that can be easily done will add it to this PR.

ui/src/app/app.module.ts
```diff
@@ -62,6 +62,9 @@ import {
   Configuration as LeoConfiguration,
 } from 'notebooks-generated';
 
+import {CardComponent} from './views/card/component';
+import {ScrollComponent} from './views/scroll/component';
```
rushtong: This doesn't compile for me

ui/src/app/views/card/component.html
rushtong: I'm very confused by this class. I thought the scroller content would be defined in a different story. The scope of RW-854 is just the scroller itself that can accept *something* - I think this should go away.

rushtong: Found it. I think `card/component.html` and `card/component.ts` should be removed. That is defined in https://precisionmedicineinitiative.atlassian.net/browse/RW-1009

NehaBroad:
For some reason scroll components went into ignore mode. Should be pushed now

ui/src/app/views/card/component.html
blrubenstein: I actually disagree Greg, I think that this PR is considerably more easy to implement with an underlying structure for what format the data should take when inputted. (Neha and I actually talked about this before hand) It is a bit of scope creep for the ticket, but I think it will make our life much easier in the long run.

NehaBroad: As per my understanding 1009 is to provide the list of elements as  card object as input and feed into Scroll component. By doing this they could see the cards and scroll through them ( as part of this story?)

blrubenstein:
@NehaBroad, a couple of things on trying it out:
1) I am not actually seeing the scroll component in the view. I can find it if I dig down into the tree, but can't see it on the physical page.
2) There appears to be a horizontal scroll bar? I think the margins fixing you did ended up pushing the dom wider, which created that?

ui/src/app/views/card/component.html
rushtong: @NehaBroad - that is correct - RW-1009 provides the input to the scroller. This PR defines the input also. That's what I'm concerned about - blurring the lines between what we scope in planning and what we end up implementing. If RW-854 is re-scoped to define the format of the inputs, then I'd be happier. I have a separate set of concerns about that.

rushtong: @blrubenstein If we re-scope RW-854 to cover the input format, I'll look at this with a different perspective. I think the card class needs to be much more generic if that's the case and not constrain the input to the degree that it does.

NehaBroad:
@blrubenstein I have removed the dummy data from the component  file. I can add it surrounding it with comments so that you can check it and we have the reference for the object required too

rushtong:
On a 1280x800 screen, the page boundaries are off. Was that were intended to be part of this PR (I thought that was mentioned yesterday)? https://precisionmedicineinitiative.atlassian.net/browse/RW-616 covers the right hand edge in a different context, so if this problem is better handled in that story, then I'm fine with ignoring it for this PR. 
screen shot 2018-07-17 at 9 14 26 am

rushtong:
Per the above screenshot, the link to "See all documentation" is cut off.

rushtong:
Per the above screenshot, the hover text over the links has a transparent (?) background (or is it z-index?) so the link text behind it shows through.

rushtong:
I'm fine with the `COMING SOON...` links, but I'd like that kind of information specified in the story. As a bonus, the stories that define when we change `COMING SOON...` to `Browse` and `Explore` could also be linked. As a reviewer, that's what I'm going to look at for comparison.

rushtong:
Per the above screenshot, `With` should not be capitalized.

ui/src/app/views/card/component.html
blrubenstein: Agreed. That makes sense to me.

blrubenstein:
@NehaBroad If you just send me the code I can add it manually. Rather not have dummy code potentially getting in master/remote if we can avoid it?

ui/src/app/views/homepage/component.html
```diff
@@ -98,24 +98,31 @@
...
-              <clr-icon shape="plus-circle" class="plus-circle"></clr-icon>
+          <div class="col-md-3" style="margin-bottom: 2rem;">
+            <div class="card add-card" (click)="addWorkspace()">
+              <div class="card-text">
+                Create a <br/> New Workspace <br/>
+                <clr-icon shape="plus-circle" class="plus-circle"></clr-icon>
+              </div>
             </div>
           </div>
+
```
UrsaStutsman: nit extra line

PR #1016, opened 2018-07-18
[RW-1003][Risk=low] Disable actions on review or insufficient permiss…

blrubenstein:
…ions

Replace this this template with your PR description.

* **no**: None 

Please also:

* Get thumbs from reviewer(s)

### DIFF
--- a/ui/src/app/views/cohort-list/component.ts
+++ b/ui/src/app/views/cohort-list/component.ts
@@ -77,19 +77,6 @@ export class CohortListComponent implements OnInit, OnDestroy {
     });
   }
 
-  cloneCohort(cohort: Cohort): void {
-    this.router.navigate(['build'], {relativeTo: this.route,
-      queryParams: {criteria: cohort.criteria}});
-  }
-
-  editCohort(cohortId: String): void {
-    this.router.navigate([cohortId, 'edit'], {relativeTo: this.route});
-  }
-
-  reviewCohort(cohortId: String): void {
-    this.router.navigate([cohortId, 'review'], {relativeTo: this.route});
-  }
-
   get writePermission(): boolean {
     return this.accessLevel === WorkspaceAccessLevel.OWNER
       || this.accessLevel === WorkspaceAccessLevel.WRITER;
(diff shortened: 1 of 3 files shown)


ui/src/app/views/cohort-list/component.ts
```diff
@@ -71,12 +71,37 @@ export class CohortListComponent implements OnInit, OnDestroy {
   }
 
   public deleteCohort(cohort: Cohort): void {
+    if (this.awaitingReview || !this.writePermission) {
```
freemabd1: Is this condition needed, since were checking this above for the class.disabled? Seems if the button is disabled then this function should never get called in this component?

ui/src/app/views/cohort-list/component.ts
```diff
@@ -71,12 +71,37 @@ export class CohortListComponent implements OnInit, OnDestroy {
...
     this.cohortsService.deleteCohort(this.wsNamespace, this.wsId, cohort.id).subscribe(() => {
       this.cohortList.splice(
         this.cohortList.indexOf(cohort), 1);
     });
   }
 
+  cloneCohort(cohort: Cohort): void {
+    if (this.awaitingReview || !this.writePermission) {
+      return;
+    }
```
freemabd1: If we do need these conditions in each of the 4 functions in the component, can we consolidate this into a single function?

blrubenstein: Will do!

ui/src/app/views/cohort-list/component.ts
blrubenstein: Class.disabled is actually just for display purposes, it doesn't block a click function (unfortunately). As far as I can tell there is no way to block click...

dolbeew: I believe [disabled]="condition" will grey out the button and block click

blrubenstein: Trying now. (I had some problems with that before, but it may be a built in clarity directive, and thus didn't work for other clicks?)

PR #1017, opened 2018-07-18
[RW-753][risk=low] Switch to oauth redirect

calbach:
+ This approach appears to be less dependent on iframes, cookies, local storage. Resolves a class of silent errors which prevent login that have been observed by beta users as well as myself (we get no useful failure information).
- This approach is a bit more jarring, and tacks a long string into the URL hash on redirect (which we then strip out).

The nuclear fix to these sign in issues which we may want eventually would  be to switch to a server side oauth flow. This would also involve the full page redirect, but would skip the giant URL hash fragment.

Demo: https://calbach-dot-all-of-us-workbench-test.appspot.com

^ @fknabe for product approval

### DIFF
--- a/ui/src/app/services/sign-in.service.ts
+++ b/ui/src/app/services/sign-in.service.ts
@@ -35,7 +35,11 @@ export class SignInService {
   }
 
   public signIn(): void {
-    gapi.auth2.getAuthInstance().signIn({'prompt': 'select_account'});
+    gapi.auth2.getAuthInstance().signIn({
+      'prompt': 'select_account',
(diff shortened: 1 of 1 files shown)


calbach:
Unfortunately this didn't fix sign in for Jen (though it did fix my own occurrence of this issue locally)... Probably not worth pursuing this.

calbach:
Alright, I was worried about the "from" parameter behavior, but apparently that's been broken for a while now. Unfortunately based on how the redirect whitelisting works, we'll probably need to move to server-side oauth before we can fix that after this changed. RW-1055 to track.

PR #1023, opened 2018-07-19
[risk=no] rm dead code: databrowser URL check

calbach:
Data browser is now a separate app


rushtong:
👍 It would also be nice to have a story for this.

calbach:
Note: there was no bug here AFAIK, it was just dead code. Do you still think it's worth filing tickets for trivial issues like this if I happen to notice them while working on something unrelated?

rushtong:
@calbach - I was thinking more along the lines of giving something for the security team to look at in terms of the overall intention of the PR. I can certainly live with no ticket if we're all comfortable with that approach.

calbach:
What I heard this week from security is that often commit messages aren't clear - at which point they need to dig into the content of the PR (or possibly the ticket). I fixed up the description to make this a bit easier.

My preference is for a policy where there's no hard requirement on Jira, in examples like this one it leads to scenarios where it's just generating noise in Jira and the process overhead arguably outweighs the effort involved in the PR itself.

PR #1025, opened 2018-07-19
[RW-1010][risk=low] Retry on getProfile 409s

calbach:
This is a bit of a band-aid to avoid users getting stuck in a "failed to load profile state", which will be much more likely once I add the IDV redirect page. There may be other ways to also reduce parallel calls by the client or improve the API impl.


calbach:
Per the comment in code, we throw a 409 when we try to concurrently modify the user model. Since put a ton of lazy initialization into /me when you first sign in, this has a likelihood of occurring at that point.

PR #1027, opened 2018-07-19
[RW-930][risk=no] Add unregistered view and guard to redirect to it

calbach:
This changed logic only applies when enforceRegistered is enabled. This currently affects staging/prod, but soon will also affect stable.

New logic:
- For nearly all signed in pages, guard according to data access level.
- If the user does not have registered access, redirect them to the new "unregistered" page.
- The unregistered page will submit IDV request, ToS, demographics, training, if any of the above have not already been submitted.

If a registered access user hits the unregistered page, they will be redirected according to the "from" parameter (if any), or the dashboard homepage otherwise.

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -1,6 +1,7 @@
 import {NgModule} from '@angular/core';
 import {NavigationEnd, Router, RouterModule, Routes} from '@angular/router';
 
+import {RegistrationGuard} from './guards/registration-guard.service';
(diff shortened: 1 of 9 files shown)


calbach:
Tests written for the component. Please take another quick look if you're able to.

calbach:
Thanks - those fields were stale, and as a result of my new test code now asserting the `Profile` type on that stub variable, it failed compilation (unrecognized fields for type `Profile`).

PR #1037, opened 2018-07-20
[RW-1033][RW-712][risk=low] Retry IDV failures and navigate on completed registration

calbach:
- Return conflict errors from registration-related profile methods as needed.
- For users who are already ID verified, and are missing other fields, we'll now submit the rest and redirect after.
- Retry registration API conflict errors

### DIFF
--- a/ui/src/app/views/unregistered/component.spec.ts
+++ b/ui/src/app/views/unregistered/component.spec.ts
@@ -10,6 +10,7 @@ import {
 } from '@angular/router';
 import {RouterTestingModule} from '@angular/router/testing';
 import {ClarityModule} from '@clr/angular';
+import {Observable} from 'rxjs/Observable';
 
 import {ProfileStorageService} from 'app/services/profile-storage.service';
 import {ServerConfigService} from 'app/services/server-config.service';
@@ -27,6 +28,27 @@ import {
 } from 'generated';
 
 
+class FailingProfileStub extends ProfileServiceStub {
+  constructor(private fails: number) {
+    super();
+  }
+
+  submitTermsOfService(extraHttpRequestParams?: any): Observable<Profile> {
+    if (this.fails-- > 0) {
+      throw {status: 409};
+    }
+    return super.submitTermsOfService(extraHttpRequestParams);
+  }
+}
+
+class TrainingCompletesRegistrationStub extends ProfileServiceStub {
+  completeEthicsTraining(extraHttpRequestParams?: any): Observable<Profile> {
+    const obs = super.completeEthicsTraining(extraHttpRequestParams);
+    this.profile.dataAccessLevel = DataAccessLevel.Registered;
+    return obs;
+  }
+}
+
 @Component({
   selector: 'app-test',
   template: '<router-outlet></router-outlet>'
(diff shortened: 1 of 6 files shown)


ui/src/app/views/unregistered/component.spec.ts
```diff
@@ -27,6 +28,27 @@ import {
 } from 'generated';
 
 
+class FailingProfileStub extends ProfileServiceStub {
+  constructor(private fails: number) {
+    super();
+  }
+
+  submitTermsOfService(extraHttpRequestParams?: any): Observable<Profile> {
+    if (this.fails-- > 0) {
```
blrubenstein: Nit: Spacing

ui/src/app/views/unregistered/component.spec.ts
```diff
@@ -85,6 +108,8 @@ describe('UnregisteredComponent', () => {
     router.navigateByUrl('/unregistered');
     tick();
     fixture.detectChanges();
+
+    unregisteredComponent = de.children[1].componentInstance;
```
blrubenstein: Could you add a comment on what element this is trying to grab? (Just confused by the array selection)

Is it the component as a whole?

ui/src/app/views/unregistered/component.spec.ts
```diff
@@ -101,6 +126,13 @@ describe('UnregisteredComponent', () => {
     profileStorageStub.profile.next(profile);
   };
 
+  const expectAllRegistrationSubmitted = (p: Profile) => {
```
blrubenstein: Good abstraction. 👍

ui/src/app/views/unregistered/component.spec.ts
```diff
@@ -116,11 +148,43 @@ describe('UnregisteredComponent', () => {
       requestedIdVerification: false,
     });
 
+    expectAllRegistrationSubmitted(profileStub.profile);
+  }));
+
+  it('should retry failed registration steps', fakeAsync(() => {
+    profileStub = new FailingProfileStub(2);
+    // Hacky, but unfortunately profileStub does not directly implement ProfileService.
```
blrubenstein: Should we move it to do so?

ui/src/app/views/unregistered/component.spec.ts
calbach: I think this spacing was OK, but I agree this looks weird. Split this out onto two lines instead.

ui/src/app/views/unregistered/component.spec.ts
calbach: Added comment and used a slightly less opaque method instead.

ui/src/app/views/unregistered/component.spec.ts
calbach: Done. I'd tried this in the past and gotten stuck a few times. It's not super clean because swagger codegen does not give us an interface, just an implementation, so we need to use extension. This means we can't really enforce by compilation that we'd have a stub which implements all API methods.

That said, you will get errors if those stub methods aren't implemented, or alternatively you can pass through a mock HTTP. I think this might actually be a reasonable way forward, and I've switched the profile service stub to use this approach. If we decide we like this, we can switch the other stubs to follow this pattern later. PTAL.

ui/src/app/views/unregistered/component.spec.ts
blrubenstein: It was mainly the `-- >` that felt weird. That said, that's cause I was actually reading it as an arrow function rather than a subtraction followed by a > sign. Which doesn't really make sense looking again in context, but I think separating is clearer.

calbach: Yep, completely agree. Another way might have been `if (0 < this.fails--)` but that felt a bit forced.

PR #1044, opened 2018-07-26
[RW-712][risk=no] enforceRegistered in stable

calbach:
(+) Makes staging, stable, prod all consistent w.r.t. user registration. Now exercise the same codepaths.
(-) New users in stable will need to be ID verified. For now we can have a process where users request this via slack in `#workbench-beta` (will need to communicate this change).
(-) In terms of business need, this verification is only required in production, since other environments will only contain synthetic.

Note: if users want to create many accounts without waiting on human verification, they can still use the test environment.


calbach:
This will land next release, I'll set a reminder to notify workbench-beta / Jeremy as we approach that timing.

PR #387, opened 2018-01-08
[RW-285] Clone cohorts and reviews on workflow clone; fix tests

calbach:
- Clone cohorts and reviews when cloning a workspace.
- Add a `CohortService` to later support a standalone `CloneCohort` method
- `@Autowire` the workspace controller to achieve better similarity to the server environment (this required adding `setUserProvider`, as I couldn't figure out how to bind a dynamic value for user/userProvider in the test configuration).

PR #1045, opened 2018-07-27
[RW-1022][Risk=no] Help tip/Page visit implementation

UrsaStutsman:
@fknabe @cukarthik will need final approval on these. (see below attached screenshots). 
Current logic (implemented in this PR): User only sees help tip on FIRST visit to page/feature. 

screen shot 2018-08-01 at 11 29 57 am
screen shot 2018-08-01 at 11 30 10 am


api/src/main/java/org/pmiops/workbench/db/model/User.java
```diff
@@ -170,6 +171,17 @@ public void setAuthorities(Set<Authority> newAuthorities) {
...
 
+  @ElementCollection
+  @CollectionTable(name = "PageVisit", joinColumns=@JoinColumn(name="user_id"))
+  @Column(name="page_id")
+  public Set<PageVisit> getPageVisits() {
+    return pageVisits;
+  }
+
+  public void setPageVisits(Set<PageVisit> pageVisits) {
+    this.pageVisits = pageVisits;
```
UrsaStutsman: I may be changing this a little, as I'm not positive that it update page visits the way I want it to.

api/db/changelog/db.changelog-43-add-page-visit.xml
```diff
@@ -0,0 +1,31 @@
...
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="ursas" id="changelog-43-add-user-visited-page">
+        <createTable tableName="page_visit">
+            <column name="page_visit_id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="user_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="page_id" type="varchar(40)">
```
rushtong: Throwing out some random thoughts ... 

What do you think about size here? Here's one url I have on test (length 91):
`https://all-of-us-workbench-test.appspot.com/workspaces/aou-rw-test-139-2/pendingworkspace`

Trimming that down to just the path (length 47):
`/workspaces/aou-rw-test-139-2/pendingworkspace`

Unless this ends up being a join and we normalize the path to another table for `page_id` -> `path`

api/src/main/java/org/pmiops/workbench/db/model/PageVisit.java
```diff
@@ -0,0 +1,55 @@
...
+  @Column(name = "first_visit")
+  public Timestamp getFirstVisit() { return firstVisit; }
+
+  public void setFirstVisit(Timestamp firstVisit) { this.firstVisit = firstVisit; }
+
+  @Column(name = "last_visit")
+  public Timestamp getLastVisit() { return lastVisit; }
+
+  public void setLastVisit(Timestamp lastVisit) { this.lastVisit = lastVisit; }
+}
```
rushtong: newline

api/db/changelog/db.changelog-43-add-page-visit.xml
```diff
@@ -0,0 +1,31 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="user_id"
+                                 baseTableName="page_visit"
+                                 constraintName="fk_page_visit_user"
+                                 referencedColumnNames="user_id"
+                                 referencedTableName="user"/>
+    </changeSet>
+</databaseChangeLog>
```
rushtong: newline

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -502,6 +517,44 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
+    if (shouldAdd) {
+      org.pmiops.workbench.db.model.PageVisit firstPageVisit =
+        new org.pmiops.workbench.db.model.PageVisit();
+      firstPageVisit.setPageId(newPageVisit.getPage());
+      firstPageVisit.setUser(user);
+      firstPageVisit.setFirstVisit(timestamp);
+      firstPageVisit.setLastVisit(timestamp);
+      newPageVisits.add(firstPageVisit);
+    }
+    user.setPageVisits(new HashSet<>(newPageVisits));
```
rushtong: This logic can be simplified. For example:
```
boolean exists = oldPageVisits.stream().anyMatch(v-> v.getPageId().equals(newPageVisit.getPage()));
```

Or, you could get the old page visit directly and update if it is there:
```
    Optional<org.pmiops.workbench.db.model.PageVisit> oldVisit = oldPageVisits
        .stream()
        .filter(v -> v.getPageId().equals(newPageVisit.getPage()))
        .findFirst();
```
With the optional, that gives you enough information to either update what's there or add a new one.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -88,6 +102,10 @@ public Profile getProfile(User user) {
     if (user.getAuthorities() != null) {
       profile.setAuthorities(new ArrayList<>(user.getAuthorities()));
     }
+    if (userWithPageVisits.getPageVisits() != null) {
```
rushtong: would it be wise to add a not-empty check? Saves a call to stream over an empty set.

api/src/main/java/org/pmiops/workbench/db/model/User.java
```diff
@@ -170,6 +171,18 @@ public void setAuthorities(Set<Authority> newAuthorities) {
     this.authorities = newAuthorities;
   }
 
+  @ElementCollection(fetch = FetchType.EAGER)
```
rushtong: I'm concerned about this - do we really always want to get the page visits for every user call? You have a specific dao call to get the user with visits. Also, please confirm that I'm not wrong here - my JPA is rusty.

UrsaStutsman: I initially tried to set this as FetchType.LAZY (if my understanding is correct - remember this is all new to me), but it kept failing. I can't remember the exact error, but when I dug around on it, seems like the fix was to change it to FetchType.EAGER. If someone knows how to do this differently (AND have it work), please speak up!

calbach: I banged my head against this issue for a few days a while back. See https://github.com/all-of-us/workbench/pull/387

IIRC: simply doing a size() or Hibernate.initialize() will fetch a lazy collection on our dev server or when deployed to GAE. However - this won't work in tests as we currently have them configured b/c the EntityManager gets closed after the initial fetch. In a real spring server, we're using this setting (dubiously, it's the default) `spring.jpa.open-in-view=true` which leaves the EntityManager open for the duration of the request. So I think in theory another way of doing this would be to have a service method of `@Transactional` which just calls `getPageVisits()`.

But anyways, you already added an extra method to your UserDao, which is similar to what I did. So using EAGER here shouldn't be necessary. If you were seeing exceptions I suspect the issue may be in the other line I commented on (using the wrong `user` variable).

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
UrsaStutsman: I thought that if the array was empty, no action would be taken?

rushtong: That is mostly true, i.e., nothing will be collected in the collector. In the meantime, however, you're calling stream on it, which would be unnecessary.

api/src/main/java/org/pmiops/workbench/db/model/User.java
UrsaStutsman: Thank you! this was exactly it!

api/db/changelog/db.changelog-43-add-page-visit.xml
```diff
@@ -0,0 +1,31 @@
...
+            <column name="user_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="page_id" type="varchar(40)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="first_visit" type="datetime">
+                <constraints nullable="false"/>
+            </column>
+            <column name="last_visit" type="datetime">
```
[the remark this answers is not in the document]
UrsaStutsman: So this was suggested by Dan, and it seemed like a decent idea to me. Thought it could be helpful in understanding user flow. But no, do not have an immediate use case for this column. If you're very concerned, and don't think the noise of it (it's currently tracking two pages: workspace about, notebook list) is worth having it in place, I'll remove.

ui/src/app/views/workspace/component.ts
```diff
@@ -129,12 +135,27 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
       this.researchPurposeArray.slice(
         this.leftResearchPurposes.length,
         this.researchPurposeArray.length);
-    this.showTip = true;
+    this.showTip = false;
+    this.pageId = 'workspace';
```
calbach: nit: I think a static constant would make more sense for these IDs, e.g. https://github.com/all-of-us/workbench/blob/cefa291bc43f2a3f79f83ac5ff6ba42337008c5c/ui/src/app/services/profile-storage.service.ts#L16

ui/src/app/views/workspace/component.ts
```diff
@@ -129,12 +135,27 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
...
+    this.profileService.getPageVisits().subscribe(
+      pageVisitsReceived => {
+        this.firstVisit = !pageVisitsReceived.some(v =>
+          v.page === this.pageId);
+      },
+      error => {
+        this.pageVisitsError = true;
+      },
+      () => {
+        if (this.firstVisit) { this.showTip = true; }
```
calbach: nit: AFAIK our code almost always puts newlines here and uses brackets for if statements. Personally I'm also OK with 1-liners where the brackets are excluded, but I'm not sure if that particular style issue has ever been brought up with the group.

ui/src/app/views/workspace/component.ts
```diff
@@ -240,7 +255,7 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
   }
 
   get newWorkspace(): boolean {
-    return this.cohortList.length === 0 && this.notebookList.length === 0 && this.showTip;
+    return this.showTip;
```
calbach: Seems like `newWorkspace` should be deleted and replaced with `showTip`

api/src/main/resources/workbench.yaml
```diff
@@ -277,6 +277,44 @@ paths:
           schema:
             $ref: "#/definitions/ErrorResponse"
 
+  /v1/update-page-visits:
```
calbach: Putting "update" in the name for a POST method feels a bit redundant, I think `POST /v1/page-visits` and `GET /v1/page-visits` make sense.

api/src/main/resources/workbench.yaml
```diff
@@ -277,6 +277,44 @@ paths:
...
+    get:
+      tags:
+      - profile
+      description: Gets a users page visits
+      operationId: getPageVisits
+      responses:
+        200:
+          description: OK
+          schema:
+            type: array
```
calbach: I would recommend wrapping all responses in object types `GetPageVisitsResponse` to allow for future extension (without breaking old clients).

api/src/main/resources/workbench.yaml
```diff
@@ -277,6 +277,44 @@ paths:
...
+            $ref: "#/definitions/PageVisit"
+      responses:
+        204:
+          description: Request receieved.
+        400:
+          description: Bad request
+          schema:
+            $ref: "#/definitions/ErrorResponse"
+
+  /v1/get-page-visits:
```
calbach: Do we need a separate GET method if we already have this on the profile?

ui/src/app/views/notebook-list/component.ts
```diff
@@ -81,6 +92,21 @@ export class NotebookListComponent implements OnInit, OnDestroy {
           this.cohortsError = true;
         });
     this.loadNotebookList();
+    this.profileService.getPageVisits().subscribe(
```
calbach: This is already on the profile right? Can we just read it from there to avoid the extra call?

api/src/main/java/org/pmiops/workbench/db/model/User.java
```diff
@@ -170,6 +171,18 @@ public void setAuthorities(Set<Authority> newAuthorities) {
     this.authorities = newAuthorities;
   }
 
+  @ElementCollection(fetch = FetchType.LAZY)
+  @CollectionTable(name = "page_visit", joinColumns=@JoinColumn(name="user_id"))
+  @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.EAGER, mappedBy = "user")
```
rushtong: Is there a conflict between the two fetch types here? Lazy on line 174 and eager here on 176.

api/src/main/resources/workbench.yaml
rushtong: +1, make use HTTP verbs in endpoints.

api/src/main/resources/workbench.yaml
rushtong: +1

ui/src/app/views/workspace/component.ts
```diff
@@ -4,19 +4,23 @@ import {ActivatedRoute, Router} from '@angular/router';
...
-  WorkspacesService,
+Cohort,
+CohortsService,
+ErrorResponse,
+FileDetail,
+PageVisit,
+ProfileService,
+Workspace,
+WorkspaceAccessLevel,
+WorkspacesService,
```
rushtong: Looks like the indentation was lost here.

ui/src/testing/stubs/profile-service-stub.ts
```diff
@@ -74,6 +76,25 @@ export class ProfileServiceStub extends ProfileService {
     });
   }
 
+  public getPageVisits(extraHttpRequestParams?: any): Observable<Array<PageVisit>> {
```
calbach: rm

ui/src/app/views/workspace/component.ts
```diff
@@ -129,12 +133,25 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
...
         this.researchPurposeArray.length);
-    this.showTip = true;
+    this.showTip = false;
+    this.newPageVisit.page = WorkspaceComponent.PAGE_ID;
   }
 
   ngOnInit(): void {
     this.wsNamespace = this.route.snapshot.params['ns'];
     this.wsId = this.route.snapshot.params['wsid'];
+    this.profileService.getMe().subscribe(
```
calbach: tl;dr: I would add a TODO here to switch this to use the profileStorageService.

This should probably use the profileStorageService, which caches the profile lookup. *But* unfortunately this might cause bugs in your help tip display within a single client session because the profile doesn't get updated dynamically on methods that mutate the profile. I filed a ticket to make that happen https://precisionmedicineinitiative.atlassian.net/projects/RW/issues/RW-1057

ui/src/app/views/notebook-list/component.ts
```diff
@@ -81,6 +88,16 @@ export class NotebookListComponent implements OnInit, OnDestroy {
...
         });
     this.loadNotebookList();
+    this.profileService.getMe().subscribe(
+      profile => {
+        this.firstVisit = !profile.pageVisits.some(v =>
+          v.page === NotebookListComponent.PAGE_ID);
+      },
+      error => {},
+      () => {
+        if (this.firstVisit) { this.showTip = true; }
```
calbach: same indentation as the other file

ui/src/app/views/notebook-list/component.spec.ts
```diff
@@ -81,7 +83,9 @@ const activatedRouteStub  = {
 
 describe('NotebookListComponent', () => {
   let notebookListPage: NotebookListPage;
+  let profileServiceStub: ProfileServiceStub;
```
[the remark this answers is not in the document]
calbach: Not addressed

api/src/main/resources/workbench.yaml
```diff
@@ -277,6 +277,26 @@ paths:
...
+      description: Updates a users page visits
+      operationId: updatePageVisits
+      parameters:
+        - in: body
+          name: pageVisit
+          description: the users pageVisits
+          schema:
+            $ref: "#/definitions/PageVisit"
+      responses:
+        204:
```
calbach: It is likely we will want this to return either the page visits, or the entire profile (the latter would be consistent with the TOS/demographics/idv APIs). This would be a requirement for the profile storage work I mentioned in another comment. Feel free to do that now if it's easy, otherwise it can wait.

ui/src/app/views/notebook-list/component.ts
```diff
@@ -45,7 +45,7 @@ export class NotebookListComponent implements OnInit, OnDestroy {
   showTip: boolean;
   cohortsLoading: boolean;
   cohortsError: boolean;
-  newPageVisit: PageVisit;
+  newPageVisit: PageVisit = { page: NotebookListComponent.PAGE_ID };
```
calbach: Sorry, I was not entirely clear with my comment. I meant that I would remove the `newPageVisit` variable entirely and just inline this object at the callsite below. It's not needed anywhere else.

(here and in the other file)

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -502,6 +500,26 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
+    if (shouldAdd) {
+      org.pmiops.workbench.db.model.PageVisit firstPageVisit =
+        new org.pmiops.workbench.db.model.PageVisit();
+      firstPageVisit.setPageId(newPageVisit.getPage());
+      firstPageVisit.setUser(user);
+      firstPageVisit.setFirstVisit(timestamp);
+      oldPageVisits.add(firstPageVisit);
+      user.setPageVisits(new HashSet<>(oldPageVisits));
+      userDao.save(user);
+    }
```
rushtong: Two questions here.

1. Why create the list of oldPageVisits and timestamp outside of this block? We only need them in the case of shouldAdd == true
2. I asked in a previous comment, why do we need the temp list at all? Can't we just call user.getPageVisits().add()? Does that not work with our use of JPA?

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -36,4 +30,11 @@
   @Query("SELECT DISTINCT user.freeTierBillingProjectName FROM User user\n" +
       "WHERE user.freeTierBillingProjectName IS NOT NULL")
   Set<String> getAllUserProjects();
+
+  /**
+   * Returns the user with the page visits and authorities loaded.
+   */
+
```
rushtong: nit: extra line

ui/src/app/views/notebook-list/component.ts
```diff
@@ -154,6 +172,10 @@ export class NotebookListComponent implements OnInit, OnDestroy {
     return this.awaitingReview || !this.writePermission;
   }
 
+  get helpTip(): boolean {
```
calbach: Why is this wrapper needed? Can you just reference `showTip` from the template?

ui/src/app/views/workspace/component.ts
rushtong: +1

api/src/main/java/org/pmiops/workbench/db/model/User.java
UrsaStutsman: Does not seem to be, and fails if I set the second one to LAZY, I think this goes into the issue @calbach spoke about earlier around these settings.

api/src/main/resources/workbench.yaml
```diff
@@ -950,6 +972,21 @@ definitions:
...
+    properties:
+      userId:
+        type: integer
+        format: int64
+      page:
+        type: string
+      firstVisit:
+        type: integer
+        format: int64
+      lastVisit:
```
calbach: No longer needed

api/src/main/java/org/pmiops/workbench/db/model/User.java
calbach: I'm surprised this would still be needed. I do worry that this line means we always fetch page visits on every user fetch.

I will patch this now and see if I can figure out anything.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -502,6 +499,23 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
+    if (shouldAdd) {
+      org.pmiops.workbench.db.model.PageVisit firstPageVisit =
+        new org.pmiops.workbench.db.model.PageVisit();
+      firstPageVisit.setPageId(newPageVisit.getPage());
+      firstPageVisit.setUser(user);
+      firstPageVisit.setFirstVisit(timestamp);
+      user.getPageVisits().add(firstPageVisit);
+      userDao.save(user);
+    }
+    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
```
calbach: Should now return the profile

ui/src/app/styles/cards.css
```diff
@@ -121,24 +121,30 @@
 }
 
 .help-tip {
-  height: 128px;
-  width: 293px;
+  height: 3.25rem;
```
calbach: I would use a margin top/bottom rather than a fixed height here, I noticed this issue locally when reducing the window size: 
image

api/src/main/java/org/pmiops/workbench/db/model/User.java
calbach: updatePageVisits was reading `getPageVisits()` without using the new dao method you defined, this should fix:

https://github.com/all-of-us/workbench/compare/us/rw-1022...ch/try-lazy-visits?expand=1

PR #1049, opened 2018-07-30
[RW-953][risk=no] Handle NPEs from invalid billing project memberships

rushtong:
Addresses: https://precisionmedicineinitiative.atlassian.net/browse/RW-953
Minor updates to protect against nulls in billing project memberships.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -268,6 +268,8 @@ public class ProfileController implements ProfileApiDelegate {
     BillingProjectStatus status;
     try {
       status = fireCloudService.getBillingProjectMemberships().stream()
+          .filter(m -> m.getProjectName() != null)
+          .filter(m -> m.getCreationStatus() != null)
           .filter(m -> user.getFreeTierBillingProjectName().equals(m.getProjectName()))
           .map(m -> fcToWorkbenchBillingMap.get(m.getCreationStatus()))
           // Should be at most one matching billing project; though we're not asserting this.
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -268,6 +268,8 @@ private User initializeUserIfNeeded() {
     BillingProjectStatus status;
     try {
       status = fireCloudService.getBillingProjectMemberships().stream()
+          .filter(m -> m.getProjectName() != null)
+          .filter(m -> m.getCreationStatus() != null)
```
calbach: Out of curiosity, can FC actually return null for this value? (fine to check defensively for it either way)

What line of this PR did that NPE effectively come from? While you're here, perhaps check for `fcToWorkbenchBillingMap.containsKey()` as well?

rushtong: I don't know if FC can return a null here - this is more of a defensive attempt than root cause analysis. The associated test also triggers the error case seen in stackdriver modulo the filters.

The error is triggered from the `.findFirst()` call. If the previous map function returns a null, this stream call fails.

rushtong: And 👍 on the `fcToWorkbenchBillingMap` idea.

rushtong:
FYI, the codacy check is unhappy about duplicate test code. I don't agree with this particular codacy decision in test code - refactoring the boilerplate setup is an added layer of complexity that I don't think is warranted.

calbach:
+1 to disabling that rule

PR #1054, opened 2018-07-30
Srushti gangireddy/physical measurements [risk=low]

SrushtiGangireddy:
1. Added queries to generate measurement counts data needed.
2. Added queries to generate the survey demographic data.
3. Added MeasurementAnalysis and surveydemographicanalysis apis.


api/db-cdr/csv/achilles_analysis.csv
```diff
@@ -10,18 +10,25 @@ analysis_id,"analysis_name","stratum_1_name","stratum_2_name","stratum_3_name","
...
 606,"Age distribution by procedure_concept_id","procedure_concept_id","gender_concept_id","","","","box","distribution"
-1815,"Distribution of numeric values, by measurement_concept_id and unit_concept_id","","","","","","column","counts"
+1806,"Distribution of age by measurement_concept_id","measurement_concept_id","gender_concept_id","null","null","null","box","distribution"
+1815,"Distribution of numeric values, by measurement_concept_id and unit_concept_id","measurement_concept_id","unit_concept_id","","","","box","distribution"
 1816,"Distribution of low range, by measurement_concept_id and unit_concept_id","","","","","","column","counts"
 1817,"Distribution of high range, by observation_concept_id and unit_concept_id","","","","","","column","counts"
+1900,"Measurement response gender distribution","measurement_concept_id","gender_concept_id","","value","unit_source_value","column","counts"
+1901,"Measurement response age distribution","measurement_concept_id","age_decile","","value","unit_source_value","column","counts"
+1911,"Measurement Male Distribution","measurement_concept_id","gender_concept_id",NULL,"value","unit_source_value","column","counts"
```
biopete: We don't want the male and female analysis anymore.

api/db-cdr/csv/achilles_analysis.csv
```diff
@@ -10,18 +10,25 @@ analysis_id,"analysis_name","stratum_1_name","stratum_2_name","stratum_3_name","
...
 2001,"Number of patients with at least 1 Dx and 1 Proc","","","","","","column","counts"
 2002,"Number of patients with at least 1 Meas, 1 Dx and 1 Rx","","","","","","column","counts"
 2003,"Number of patients with at least 1 Visit","","","","","","column","counts"
-3000,"Participant Count","concept_id",NULL,"domain_id",NULL,NULL,"column","counts"
+3000,"Participant Count","concept_id","bin_size","domain_id","bin_min","bin_max","column","counts"
 3004,"Number of persons with at least one concept by concept_id by calendar year by gender by age decile","concept_id","calendar year","gender_concept_id","age decile","","column","counts"
 3006,"Age distribution by gender","concept_id","gender_concept_id","","","","box","distribution"
 3101,"Gender","concept_id","ppi_sex_at_birth_concept_id","domain_id",NULL,NULL,"pie","counts"
 3102,"Age","concept_id","age_decile","domain_id","null","","column","counts"
+3103,"Race","survey_concept_id","race_concept_id","domain_id",NULL,NULL,"column","counts"
```
biopete: This should just be "concept_id" not survey_concept_id , in case we do this for more concepts

api/db-cdr/csv/achilles_analysis.csv
```diff
@@ -10,18 +10,25 @@ analysis_id,"analysis_name","stratum_1_name","stratum_2_name","stratum_3_name","
...
+3104,"Ethnicity","survey_concept_id","ethnicity_concept_id","domain_id",NULL,NULL,"column","counts"
```
biopete: same here

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,37 @@
+COUNT GENERATION
+
+This document describes what counts are generated for each of the data domains
+
+Total Person count is generated and stored with analysis id 1.
+Person count by gender is generated and stored with analysis id 2.
+Person count by year of birth is generated and stored with analysis id 3.
+Person count by race is generated and stored with analysis id 4.
+Person count by ethnicity is generated and stored with analysis id 5.
+Person count by year of birth and gender is generated and stored with analysis id id 10.
```
biopete: id id ... (too many id)

biopete: Formatting of this section in readme needs work. Use markdown like in other readmes.

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,37 @@
+COUNT GENERATION
+
+This document describes what counts are generated for each of the data domains
+
```
biopete: Please add bit about source concepts and count values. definition of count value and source count value like so: 
Most EHR data in OMOP has two concepts associated with it , the source concept and the standard concept that source maps to.  So we calculate two counts for each concept 
countValue is the number of distinct patients that had that concept or some concept that mapped to it. 
sourceCountValue is the number of distinct patients that had that concept as real ehr data.
In the case of source concepts that always map to a standard concept, the countValue would be zero. In that case we set countValue to the sourceCountValue as it is more userfriendly in apis and front end development. 
For standard concepts , the count value will be much more than the sourceCountValue in general. 

That's a start :)

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,37 @@
...
+Person count by race is generated and stored with analysis id 4.
+Person count by ethnicity is generated and stored with analysis id 5.
+Person count by year of birth and gender is generated and stored with analysis id id 10.
+Person count by race and ethnicity is generated and stored with analysis id 12.
+
+
+1. EHR DATA
+
+   Domains( VISIT, CONDITION, PROCEDURE, DRUG, OBSERVATION, MEASUREMENT):
+    a: Counts and source counts of people with one of the occurrence of the concepts in these domains are generated and stored in analysis id 3000.
```
biopete: Clear wording "Counts and source counts of distinct patients with at least one occurrence ..."

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,37 @@
...
+    Counts are the number of people that have the concept occurred and source counts are the number of people that have the same concept as source concept. In case the count value is 0
```
biopete: Peter : Todo left off here.

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,37 @@
...
+    a: Response count for each of the survey question in each module is generated and stored with analysis id 3110.
+    b: Response count by gender is generated and stored with analysis id 3111.
+    c: Response count by age decile is generated and stored with analysis id 3112.
+    d: Participant count in each of the survey module is generated.
+
+3. MEASUREMENTS
+
+    a: Box distribution of value is generated for each measurement.
+    b: Box distribution of age with gender is generated for each measurement.
+    c: Counts and source counts of different gender, age for binned values are generated. (Values are binned by 10)
```
biopete: Is this true ? I thought we were just making 20 bins based on min and max  or something like that.

api/db-cdr/csv/db_domain.csv
```diff
@@ -2,14 +2,14 @@
...
 "CPT4","CPT4","CPT or Current Procedural Terminology (CPT) is a product of the American Medical Association (AMA). CPT codes are used by health care professionals for billing of medical services and procedures to public and private health insurance programs. CPT is a list of descriptive terms and identifying numeric codes for medical services and procedures that are provided by physicians and health care professionals.","standard","cpt4",44819100,0
 "Drug","Medications","Drugs biochemical substance formulated in such a way that when administered to a Person it will exert a certain physiological or biochemical effect. The drug exposure domain concepts capture records about the utilization of a Drug when ingested or otherwise introduced into the body.","domain_filter","drug",13,0
-"Fam Med His","Family Medical History","The Family module provides information on family medical history","survey","PPI",NULL,0
+"Fam Med His","Family Medical History","The Family module provides information on family medical history","survey","PPI",0,0
```
biopete: does this break any api , going from null to 0 ?  It is fine to have it 0 as default. Just make sure our api is not looking for null.

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -549,24 +549,33 @@ where co1.observation_concept_id > 0 and (extract(year from co1.observation_date
 group by co1.observation_source_concept_id, stratum_2
 "
 
-# 1800 Measurements - Number of persons with at least one measurement occurrence, by measurement_concept_id
+# 3000 Measurements - Number of persons with at least one measurement occurrence, by measurement_concept_id
```
biopete: Please add to comment saying you are getting the max and min too

biopete: Also , should we move this to the measurement script ?

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,412 @@
+#!/bin/bash
+
+# Runs achilles queries to populate count db for cloudsql in BigQuery
+set -xeuo pipefail
+IFS=$'\n\t'
+
+USAGE="./generate-clousql-cdr/run-achilles-queries.sh --bq-project <PROJECT> --bq-dataset <DATASET> --workbench-project <PROJECT>"
```
biopete: run-measurement

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,412 @@
+#!/bin/bash
+
+# Runs achilles queries to populate count db for cloudsql in BigQuery
```
biopete: update comments and usage

api/db-cdr/csv/achilles_analysis.csv
SrushtiGangireddy: Addressed

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,412 @@
...
+min(case when p.accumulated >= .50 * o.total then count_value else round(o.max_value,2) end) as median_value,
+min(case when p.accumulated >= .10 * o.total then count_value else round(o.max_value,2) end) as p10_value,
+min(case when p.accumulated >= .25 * o.total then count_value else round(o.max_value,2) end) as p25_value,
+min(case when p.accumulated >= .75 * o.total then count_value else round(o.max_value,2) end) as p75_value,
+min(case when p.accumulated >= .90 * o.total then count_value else round(o.max_value,2) end) as p90_value
+FROM  priorstats p join overallstats o on p.stratum1_id = o.stratum1_id and p.stratum2_id = o.stratum2_id
+group by  o.stratum1_id, o.stratum2_id, o.total, o.min_value, o.max_value, o.avg_value, o.stdev_value"
+
+
+# 1900 Measurement value counts
```
biopete: Please comment on what this query is doing and how  ... it is really big.

api/db-cdr/csv/achilles_analysis.csv
SrushtiGangireddy: Done. Just added achilles result list response with list of male and female results to measurement analysis object.

api/db-cdr/csv/achilles_analysis.csv
SrushtiGangireddy: Addressed.

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,412 @@
...
+count(distinct p1.person_id) as count_value,
+count(distinct p1.person_id) as source_count_value
+FROM \`${BQ_PROJECT}.${BQ_DATASET}.measurement\` m1
+join \`${BQ_PROJECT}.${BQ_DATASET}.person\` p1 on p1.person_id = m1.person_id
+where m1.value_as_number is null and m1.value_source_value is not null
+and m1.measurement_source_concept_id > 0 and m1.measurement_source_concept_id != m1.measurement_concept_id
+group by m1.measurement_source_concept_id,m1.value_source_value,p1.gender_concept_id,m1.unit_source_value"
+
+
+# 1901 Measurement value counts
```
biopete: What is this one doing ? I thought we had value counts in 3000 ?

api/db-cdr/generate-cdr/README.md
SrushtiGangireddy: Addressed.

api/db-cdr/generate-cdr/run-measurement-queries.sh
biopete: oh it is age. Please just update comment to say what it does.

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,412 @@
...
+on p1.person_id = m1.person_id
+join \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.achilles_results\` ar on
+cast(m1.measurement_source_concept_id AS STRING)=ar.stratum_1
+where m1.measurement_source_concept_id > 0 and m1.measurement_concept_id!=m1.measurement_source_concept_id
+and m1.value_as_number is not null
+and ar.analysis_id=3000 and ar.stratum_3='Measurement'
+group by m1.measurement_source_concept_id,stratum_2,stratum_4,stratum_5"
+
+# 1900 Measurement value counts
+echo "Getting measurements value counts"
```
biopete: What are we doing here doing same thing again ? Different group of measurements ?

public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
```diff
@@ -18,6 +18,16 @@
     )
     AchillesAnalysis findConceptAnalysisResults(String concept_id,Long analysisId);
 
+    @Query(value = "select a from AchillesAnalysis a where a.analysisId=?1")
+    AchillesAnalysis findAnalysisById(Long analysisId);
+
+    @Query(value = "select distinct a from AchillesAnalysis a left join FETCH a.results as r " +
```
biopete: Please generalize this. It works for any concept id potentially. Not that we have teh data for all yet. But just rename.

public-api/src/main/java/org/pmiops/workbench/cdr/dao/AchillesAnalysisDao.java
```diff
@@ -18,6 +18,16 @@
...
+            "where r.stratum1 = ?1 and a.analysisId in (3101,3102,3103,3104) order by a.analysisId"
```
biopete: Should we make the analysis_id a list the controller can pass in ? Then this function can be findConceptAnalysesResults --- Then it is generally useful... I like that.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -5,7 +5,10 @@
 import org.pmiops.workbench.model.ConceptListResponse;
 import org.pmiops.workbench.model.ConceptAnalysisListResponse;
 import org.pmiops.workbench.model.ConceptAnalysis;
+import org.pmiops.workbench.model.MeasurementAnalysis;
```
biopete: peter todo here

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -398,6 +477,63 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
         return ResponseEntity.ok(resp);
     }
 
+    @Override
+    public ResponseEntity<MeasurementAnalysisListResponse> getMeasurementAnalysisResults(List<String> conceptIds){
```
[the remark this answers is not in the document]
biopete: we will need te rename our maleAnalysis , femailAnalysis properties to measurementValueMale measurementValueFemale ... or something ..

public-api/src/main/resources/public-api.yaml
```diff
@@ -529,6 +570,60 @@ definitions:
...
+        $ref: "#/definitions/Analysis"
+      ageAnalysis:
+        description: age analysis
+        type: object
+        $ref: "#/definitions/Analysis"
+      raceAnalysis:
+        description: race analysis
+        type: object
+        $ref: "#/definitions/Analysis"
+      ethnicityAnalysis:
```
biopete: add measurementValueMale, measurementValueFemale here and we can get rid of measurement api :)

biopete: and measurementValueAgeAnalysis ...

public-api/src/test/resources/application.properties
```diff
@@ -2,5 +2,5 @@ spring.datasource.driver-class-name=org.h2.Driver
 spring.datasource.url=jdbc:h2:mem:db;DB_CLOSE_DELAY=-1;INIT=CREATE SCHEMA IF NOT EXISTS workbench
 spring.datasource.username=sa
 spring.datasource.password=sa
-spring.jpa.hibernate.ddl-auto=create-drop
+spring.jpa.hibernate.ddl-auto=create
```
biopete: what does this change do ? What is this for?

rushtong: this tells spring to apply new schema changes if they're not in the db already. Taking away the `drop` means that spring will no longer drop tables/columns that are not referenced in the code.

SrushtiGangireddy: Peter, Greg
I seem to get an error saying table achilles_results not found after unsuccessful alter statement and tests failed.
According to this:
https://stackoverflow.com/questions/12054422/unsuccessful-alter-table-xxx-drop-constraint-yyy-in-hibernate-jpa-hsqldb-standa
hibernate seems to attempt to drop the constraints/data or whatever even from the new database and that seems to cause the issue.
Changing this to just create helped the tests to run and pass.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: Okay. Added everything to concept analysis and it seems to work fine.

api/db-cdr/csv/db_domain.csv
SrushtiGangireddy: Changed the api to fetch the domains that look out for 0 instead of null.

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -817,3 +775,104 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
...
+(id,analysis_id,stratum_1,stratum_3,count_value,source_count_value)
+SELECT 0 as id, 3000 as analysis_id,CAST(dbd.concept_id as string) as stratum_1,
+'Survey' as stratum_3,
+count(distinct o.person_id) as count_value, count(distinct o.person_id) as source_count_value
+FROM \`${BQ_PROJECT}.${BQ_DATASET}.observation\` o join \`${BQ_PROJECT}.${BQ_DATASET}.concept_relationship\` cr
+on o.observation_source_concept_id=cr.concept_id_1 and cr.relationship_id = 'Has Module'
+join \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.db_domain\` dbd on cr.concept_id_2 = dbd.concept_id
+Where (o.observation_source_concept_id > 0 and o.value_source_concept_id > 0) and (dbd.db_type = 'survey' and dbd.concept_id <> 0)
+Group by dbd.concept_id"
+
```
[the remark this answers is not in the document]
SrushtiGangireddy: Added comment.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -46,4 +46,7 @@
     @Query(value = "select c.* from concept c join db_domain d on c.concept_id=d.concept_id", nativeQuery = true)
     List<Concept> findDbDomainParticpantCounts();
 
+    @Query(value = "select distinct c.* from concept_synonym cs join concept c on c.concept_id=cs.concept_id where (cs.concept_synonym_name like ?1 and c.standard_concept='S' and c.domain_id in (?2) and c.count_value > ?3) limit ?4", nativeQuery = true)
```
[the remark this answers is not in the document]
SrushtiGangireddy: Addressed. Defined the relationship between concept and concept_synonym and fetched results by join then used the same concept search to fetch the matched concepts.

api/db-cdr/changelog-schema/db.changelog-14.xml
```diff
@@ -0,0 +1,48 @@
...
+        <indexExists indexName="concept_synonym_name_index"/>
+      </not>
+    </preConditions>
+    <sql dbms="mysql">
+      ALTER TABLE concept_synonym
+      ADD FULLTEXT INDEX concept_synonym_name_index
+      (concept_synonym_name);
+    </sql>
+
+    <createIndex tableName="concept_synonym" indexName="idx_concept_synonym_1">
```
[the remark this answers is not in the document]
SrushtiGangireddy: Oops!! changed the name.

api/db-cdr/generate-cdr/import-gcs-data.sh
```diff
@@ -37,7 +37,7 @@ REMOTE_DATA_LOC=https://storage.googleapis.com/$BUCKET
 echo "Importing data files from $REMOTE_DATA_LOC"
 
 # Add tables names of files to import here
-TABLES=(achilles_analysis achilles_results db_domain domain vocabulary criteria concept concept_relationship concept_ancestor)
+TABLES=(achilles_analysis achilles_results achilles_results_dist db_domain domain vocabulary criteria concept concept_relationship concept_ancestor)
```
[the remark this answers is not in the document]
SrushtiGangireddy: Yes, also added query to import data.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -364,36 +443,103 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+                            String analysisStratumName=ar.getAnalysisStratumName();
+                            String stratum2 = ar.getStratum2();
+                            if (analysisStratumName == null || analysisStratumName.equals("")) {
+                                ar.setAnalysisStratumName(QuestionConcept.genderStratumNameMap.get(ar.getStratum2()));
+                                if(stratum2.equals(MALE_CONCEPT_ID)){
+                                    maleResults.add(ar);
+                                }else if(stratum2.equals(FEMALE_CONCEPT_ID)){
+                                    femaleResults.add(ar);
+                                }else{
+                                    otherResults.add(ar);
```
biopete: The other results is ok for now.  Just thinking outloud here I'll have to see how it graphs ... Basically , if we had 4 other genders , we would have 40 results here for binned data, 10 for each gender, with 4 copies of each bin with different counts.  That may not graph right. May have to loop through the other results and sum up the counts for each bin so we only have one result per bin. Know what i mean ? Something like : 

unique_other_results = []
for (r of other_results) { 
    bin = r.stratum_4;
    If bin is in unique_other_results then next; // we already summed this one
    bin_count = 0;
    for (r2 of other_results) {
          if (bin === r2.stratum_4) { bin_count += r2.count_value }
    }
    Add new result for this bin to unique_other_results

I may just do something like this on front end.  Overall we may just scrap all this male / female / other and make an array for the genders and just put each gender in it's own in the array. Eventually I think we are going to show all genders ...

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -276,39 +299,59 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+            ConceptListResponse response = new ConceptListResponse();
+            List<Concept> matchedConcepts = concepts.getContent();
+            List<String> conceptSynonymNames = new ArrayList<>();
+
+            for(Concept con : matchedConcepts){
+                String conceptCode = con.getConceptCode();
+                String conceptId = String.valueOf(con.getConceptId());
+
+                if(con.getSynonyms() != null){
+                    response.setMatchType(MatchType.SYNONYM);
```
biopete: Since each concept_name is also a synonm -- we need more here to have this be accurate... basically you would have to see if the query string is actually in the concept_name ... If it is , than it is a name match , if it isn't then it is a synonm match.  I think I would just call it a name match for now.  People may not be ready to handle another match type on the front end.

api/db-cdr/generate-cdr/README.md
```diff
@@ -0,0 +1,41 @@
+COUNT GENERATION
+
+This document describes what counts are generated for each of the data domains
+
+Total Person count is generated and stored with analysis id 1.
```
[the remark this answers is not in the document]
SrushtiGangireddy: Addressed.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -237,4 +249,16 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 SELECT c.concept_id_1, c.concept_id_2, c.relationship_id
 FROM \`$BQ_PROJECT.$BQ_DATASET.concept_relationship\` c"
 
+########################
+# concept_synonym #
+########################
+echo "Inserting concept_relationship"
```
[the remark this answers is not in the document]
SrushtiGangireddy: Missed it. updated the echo statement.

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -0,0 +1,450 @@
...
+\`${BQ_PROJECT}.${BQ_DATASET}.measurement\` m1
+on p1.person_id = m1.person_id
+join \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.achilles_results\` ar on
+cast(m1.measurement_source_concept_id AS STRING)=ar.stratum_1
+where m1.measurement_source_concept_id > 0 and m1.measurement_concept_id!=m1.measurement_source_concept_id
+and m1.value_as_number is not null
+and ar.analysis_id=3000 and ar.stratum_3='Measurement' and ar.stratum_2 is not null
+group by m1.measurement_source_concept_id,stratum_2,stratum_4"
+
+# 1900 Measurement string value counts (This query generates counts, source counts of the value and gender combination. It gets bin size from joining the achilles_results)
```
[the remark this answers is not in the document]
SrushtiGangireddy: Updated the data.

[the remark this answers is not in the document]
SrushtiGangireddy: and query to add the binned/not-binned while fetching data.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -72,118 +72,131 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                                    criteriaBuilder.literal(query)));
+
+                            try {
+                                long conceptId = Long.parseLong(query);
+                                conceptCodeIDName.add(criteriaBuilder.equal(root.get("conceptId"),
+                                        criteriaBuilder.literal(conceptId)));
+                            } catch (NumberFormatException e) {
+                                // Not a long, don't try to match it to a concept ID.
+                            }
+
```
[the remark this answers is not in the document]
SrushtiGangireddy: Addressed.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
SrushtiGangireddy: Addressed.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,120 +70,96 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                            criteriaBuilder.literal(CLASSIFICATION_CONCEPT_CODE)));
+
+                    // If the concepts that match the synonym names are found, they are fetched by applying standard concept filer, domain and vocabulary id filters
+
+                    if(synonymConceptIds.size() > 0){
+                        List<Predicate> synonymConceptPredicate = new ArrayList<>();
+                        Expression<Long> conceptIdCheck = root.get("conceptId");
+                        synonymConceptPredicate.add(conceptIdCheck.in(synonymConceptIds));
+                        if (standardConceptFilter.equals(StandardConceptFilter.STANDARD_CONCEPTS) || standardConceptFilter.equals(StandardConceptFilter.STANDARD_OR_CODE_ID_MATCH)) {
+                                predicates.add(criteriaBuilder.or(synonymConceptPredicate.toArray(new Predicate[0])));
```
biopete: Just add the synonymPredicate once at the top before this if else stuff ? Looks like same line 3 times.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,120 +70,96 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                    standardConceptPredicates.add(criteriaBuilder.equal(root.get("standardConcept"),
+                            criteriaBuilder.literal(CLASSIFICATION_CONCEPT_CODE)));
+
+                    List<Predicate> nonStandardConceptPredicates = new ArrayList<>();
+                    nonStandardConceptPredicates.add(criteriaBuilder.notEqual(root.get("standardConcept"),
+                            criteriaBuilder.literal(STANDARD_CONCEPT_CODE)));
+                    nonStandardConceptPredicates.add(criteriaBuilder.notEqual(root.get("standardConcept"),
+                            criteriaBuilder.literal(CLASSIFICATION_CONCEPT_CODE)));
+
+                    // If the concepts that match the synonym names are found, they are fetched by applying standard concept filer, domain and vocabulary id filters
```
biopete: Nit: Comment is a bit misleading. The synonms don't have anything to do with vocabulary and other filters. I would just say : Apply the synomns filter if we have them or apply query/concept_code/id filters. They are mutually exclusive.  

Also, the standard concept filter could be applied generically it looks like  as you do the vocabs and domain ? Do you think ?

api/gangirs@cpm01lt.hs.it.vumc.io
```diff
@@ -0,0 +1,284 @@
+from __future__ import print_function
```
biopete: This file doesn't belong in here :)

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -92,9 +114,11 @@ public DataBrowserController(ConceptService conceptService, ConceptDao conceptDa
                             .domainId(concept.getDomainId())
                             .countValue(concept.getCountValue())
                             .sourceCountValue(concept.getSourceCountValue())
-                            .prevalence(concept.getPrevalence());
+                            .prevalence(concept.getPrevalence())
+                            .conceptSynonyms(concept.getSynonyms().stream().distinct().map(ConceptSynonym::getConceptSynonymName).collect(Collectors.toList()));
```
biopete: Can you filter out the concept_name from the synonym list, please ?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: cannot move the standard concept filter as for concept name and code/id match match had to loop it together along with the match.

api/gangirs@cpm01lt.hs.it.vumc.io
SrushtiGangireddy: oops!! removed it.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -46,7 +46,7 @@
     @Query(value = "select c.* from concept c join db_domain d on c.concept_id=d.concept_id", nativeQuery = true)
     List<Concept> findDbDomainParticpantCounts();
 
-    @Query(value = "select distinct c.conceptId from Concept c left join c.synonyms as cs where match(cs.conceptSynonymName,?1) > 0")
+    @Query(value = "select distinct c.conceptId from Concept c left join c.synonyms as cs where match(cs.conceptSynonymName,?1) > 0 or match(c.conceptName,?1) > 0")
```
biopete: This function should be named better after what it does, how about findConceptIdsByNameOrSynonymMatch 
Or if you want more general : findConceptIdsByQuery

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -93,68 +93,40 @@ public static String modifyMultipleMatchKeyword(String query){
...
-                        }
+                        predicates.add(criteriaBuilder.or(synonymConceptPredicate.toArray(new Predicate[0])));
                     }
-                    else{
-                        List<Predicate> conceptCodeIDName = new ArrayList<>();
-                        Expression<Double> matchExp = null;
-                        if(!Strings.isNullOrEmpty(query)){
-                            final String keyword = modifyMultipleMatchKeyword(query);
-                            conceptCodeIDName.add(criteriaBuilder.equal(root.get("conceptCode"),
+                    else {
```
biopete: This is all much cleaner.  One issue we may have is that if a code or conceptid happens to be in the name or synonym string, the name/synonym takes precedence here.  I think we  probably still have that problem in the master now. 

The problem is we can't tell if something is a code or concept_id until we match it.  
The logic i think needs to be : 
1) Search code / concept id  to get concept ids
2) If not code match then search name synonm to get concept ids 
3) Call the service to apply the predicate filters with list of concept ids.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -93,68 +93,40 @@ public static String modifyMultipleMatchKeyword(String query){
...
                                 // Not a long, don't try to match it to a concept ID.
                             }
-                            matchExp = criteriaBuilder.function("match", Double.class, root.get("conceptName"), criteriaBuilder.literal(keyword));
+
+                            predicates.add(criteriaBuilder.or(conceptCodeID.toArray(new Predicate[0])));
                         }
 
+
+                    }
+
```
biopete: Can you clean up indention and and add some line breaks to the really long lines in this sub and file?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
biopete: Also, the standardConcept_orCodeId filter seems to not get applied anymore.   I think the solution is to have a separate findConceptIdByCodeOrId method ... And apply that first like you do for the synonms .. Or do it inside the service searchConcepts .. You could do both inside the service searchConcepts right ? rename your variable in conceptServiceSearchConcepts to "conceptIds" instead of synonmy  ... Of course then you know the match type already before you use the service. So that will have to be handled.

Let me put in some pseudocode and commit it in this branch and you can try to implement it.

api/db-cdr/generate-cdr/import-gcs-data.sh
```diff
@@ -56,7 +56,18 @@ function mysqlimport_table () {
...
+      -u root -p${MYSQL_ROOT_PASSWORD} $db $file
+}
+
+# Function for mysqlimport concept_synonym
+function mysqlimport_concept_synonym_table () {
+   db=$1
+   file=$2
+
+   echo "Mysql importing $file into $db.$table..."
+   mysqlimport --ignore-lines=1 --fields-terminated-by=, --fields-optionally-enclosed-by='"' --columns=concept_id,concept_synonym_name,language_concept_id \
```
biopete: Why do we need this ? Wouldn't we just want to make the csv have the right columns? When i ran the command on the csv it did just fine on my computer. 

Also, Aare we sure we don't want to just add a dummy auto inc  column like we do for the other tables ?  We only did that for jpa to be happy with relationships and such.  But like achilles_results and any table that doesn't have a single pk we just made one up.  Maybe better to just have the multiple pk if we need it for relationships. 

Anway, just thoughts on that for future reference.

api/db-cdr/generate-cdr/import-gcs-data.sh
```diff
@@ -71,7 +82,12 @@ do
   #  $table0*.csv will be multiple files.
   if [ -f $local_fpath/${table}.csv ]
   then
-    mysqlimport_table $CDR_DB_NAME $local_fpath/${table}.csv
+    if [ "$table" == "concept_synonym" ];
```
biopete: Same as above? What is special about this csv that it doesn't import normally ?

biopete:
Need to resolve conflicts too. And what is up with the security error ?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,117 +70,70 @@ public static String modifyMultipleMatchKeyword(String query){
     public static final String STANDARD_CONCEPT_CODE = "S";
     public static final String CLASSIFICATION_CONCEPT_CODE = "C";
 
-    public Slice<Concept> searchConcepts(String query,
-                                         StandardConceptFilter standardConceptFilter, List<String> vocabularyIds,
-                                         List<String> domainIds, int limit, int minCount) {
+    public Slice<Concept> searchConcepts(String query, StandardConceptFilter standardConceptFilter, List<String> vocabularyIds, List<String> domainIds, int limit, int minCount, List<Long> matchedConceptIds) {
```
danqrodney: You're not actually using query in here anymore. If you're going to use this in clause approach, you should require matchedConceptIds to be set and remove the query parameter. (Though see above about why that's probably not a good idea.)

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
```diff
@@ -98,13 +104,35 @@ public ConceptsController(ConceptService conceptService, WorkspaceService worksp
...
+    conceptCodeIdMatchConcepts = conceptDao.findConceptByCode(request.getQuery());
+    try{
+      Long conceptIdQuery = Long.parseLong(request.getQuery());
+      conceptCodeIdMatchConcepts.addAll(conceptDao.findConceptById(conceptIdQuery));
+    }catch(NumberFormatException nfe){
+
+    }
+    List<Long> synonymConceptIds = new ArrayList<>();
+    if(conceptCodeIdMatchConcepts.size() == 0) {
+        synonymConceptIds = conceptDao.findConceptByNameOrSynonymName(ConceptService.modifyMultipleMatchKeyword(request.getQuery()));
```
danqrodney: Depending on what the search query is, this could return a whole bunch of concepts -- too much to fit in memory and pass into an IN clause subsequently here. Have you tried testing this with some searches for strings that appear in a whole bunch of concepts? 

I believe it should be possible to make searchConcepts join to the synonym table and match on name at the same time as filtering by vocabulary, domain, etc. and sorting by prevalence. Then the database can do the join itself, which is more efficient and scalable than us doing it this way.

common-api/src/main/java/org/pmiops/workbench/cdr/model/Concept.java
```diff
@@ -167,6 +184,21 @@ public Concept prevalence(float prevalence) {
         return this;
     }
 
+    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL, mappedBy = "concept")
```
danqrodney: Since this collection is fetched lazily, I believe by default we're going to wind up doing a separate database fetch for the list of synonyms for every concept returned in a response.... you could confirm this by turning on SQL debugging. We need to make sure these get joined in / fetched eagerly when returning concept search results to avoid this.

danqrodney: Also: what's the maximum number of synonyms for a given concept? 

We should make sure we don't try to return, say, hundreds or thousands of synonyms for a concept in a response, as that's not useful.

biopete: The one with most is  Molecular pathology procedure, Level 2. It has 204.  
Next is 94. Level 5. Then it gets down to 20. So 204 concepts have it over 20.  and 79 concepts have more than 30.   

The plan is to hide them in the Ui in an expanding text area or scrolling list. Or maybe we just show the synonyms that match the search in the databrowser. At any rate, i think that the api 
should return all of them and the ui's can do what they want.  Or maybe just return synonyms that match the search ? 

With a bunch of synonyms many are almost exactly a like. For example there are 49 synonyms for male erectile disorder i noticed in my search.  "Sexual impotence" and "Sexual impotence, Male" and  "Male Sexual impotence "  . But for the procedures like 42742371 that has 204 synonms, the synonyms look much more significantly different.   

So this is a problem we will have to tackle somehow. I think for now though just hiding or limiting the list in the ui is the best way to go since all this may change in the future.

SrushtiGangireddy: I just checked and the concept synonyms seem to be coming back.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,117 +70,70 @@ public static String modifyMultipleMatchKeyword(String query){
...
+
+        /*
+        // Peter Changes : Try something like this pseuedo code
```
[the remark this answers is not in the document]
SrushtiGangireddy: Resolved.

common-api/src/main/java/org/pmiops/workbench/cdr/model/Concept.java
SrushtiGangireddy: Changed the fetch type to lazy and used join fetch. Things seem to be working fine. JPA does not now query concept synonym table each time.

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
SrushtiGangireddy: Changed back and added concept synonym match query filter in there.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,100 +73,93 @@ public static String modifyMultipleMatchKeyword(String query){
...
                                 criteriaBuilder.or(standardConceptPredicates.toArray(new Predicate[0]))
                         );
 
-                    } else if (standardConceptFilter.equals(StandardConceptFilter.NON_STANDARD_CONCEPTS)) {
-
-                        if(!Strings.isNullOrEmpty(query)) {
+                    }else if (standardConceptFilter.equals(StandardConceptFilter.NON_STANDARD_CONCEPTS)) {
 
+                        if(keyword != null) {
                             conceptCodeIDName.add(criteriaBuilder.greaterThan(matchExp, 0.0));
```
danqrodney: Is the reason that we need this that not all concepts have synonym rows? (Just confirming.)

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -70,100 +73,93 @@ public static String modifyMultipleMatchKeyword(String query){
...
         Specification<Concept> conceptSpecification =
                 (root, criteriaQuery, criteriaBuilder) -> {
                     List<Predicate> predicates = new ArrayList<>();
+                    List<Predicate> standardConceptPredicates = new ArrayList<>();
+                    standardConceptPredicates.add(criteriaBuilder.equal(root.get("standardConcept"),
+                            criteriaBuilder.literal(STANDARD_CONCEPT_CODE)));
+                    standardConceptPredicates.add(criteriaBuilder.equal(root.get("standardConcept"),
+                            criteriaBuilder.literal(CLASSIFICATION_CONCEPT_CODE)));
+
+                    List<Predicate> nonStandardConceptPredicates = new ArrayList<>();
```
danqrodney: standardConceptPredicates and nonStandardConceptPredicates look identical... is that what you intended?

common-api/src/main/java/org/pmiops/workbench/cdr/model/Concept.java
danqrodney: Fantastic!

api/db-cdr/generate-cdr/make-bq-public-data.sh
```diff
@@ -85,7 +85,11 @@ else
   bq --project=$PUBLIC_PROJECT mk $PUBLIC_DATASET
 fi
 
+<<<<<<< HEAD
```
biopete: Git merge issue

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
biopete: Do you mean to add the match query here ? We no longer do the synonym query outside this. We got the jpa criteria to work with join. 

But it is true that not all concepts have synonyms which is why we search concept_name or concep_synonym_name.

api/db-cdr/generate-cdr/import-gcs-data.sh
SrushtiGangireddy: The import did not work fine when i tried without this.
So, added the column names.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: Yeah, that's what I was checking -- thanks.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
danqrodney: Sorry, missed the "not" part. :)

PR #1061, opened 2018-08-02
[RW-1086][risk=no] Update demo notebooks/cohorts

calbach:
Note: without this the notebooks currently in stable would have been reverted with the next push.


NehaBroad:
Do we need new lines at the EOF for json format files as well?

blrubenstein:
@NehaBroad I think it would be fine either way. I'm less worried about the demo files because the goal is for those eventually to just live in the google bucket and get pushed out by our support/data-science team

calbach:
I copied these from stable where demos are working currently, so the newline doesn't seem to be required. I'm fine to just take whatever the data-science team gives us for now and treat these as blobs.

PR #1062, opened 2018-08-03
[RW-1036][RISK=LOW] Implement storage and caching service for recently accessed user resources

NehaBroad:
This PR is only focusing on Implement storage and caching service for recently accessed user resources.


rushtong:
Nit: start to the PR title should be: `[RW-1036][risk=low]`

rushtong:
General comment: I think naming could be better. `NotebookCohortCache` is extremely implementation specific in that its only for notebooks *in combination with* cohorts. I think a better name is something like `ResourceCache` so that this concept can be expanded to other kinds of things that we might want to cache using this technique.

api/db/changelog/db.changelog-44-notebook_cohort_cache.xml
```diff
@@ -0,0 +1,35 @@
...
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="user_workspace_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cohort_id" type="bigint">
+            </column>
+            <column name="notebook_name" type="varchar(255)">
+            </column>
```
rushtong: Having one column per resource type seems inefficient. Could we have one column for "resource_identifier" and then let the calling code (service class) determine what identifier actually represents?

If you did take this approach, you'd also need to add a string column for resource type that could come from an Enum class. You'd need this in the case that two different resources happened to be identified by the same string identifier, and you'd need to disambiguate that.

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceUserRoleDao.java
```diff
@@ -0,0 +1,11 @@
...
+
+import org.pmiops.workbench.db.model.User;
+import org.pmiops.workbench.db.model.Workspace;
+import org.pmiops.workbench.db.model.WorkspaceUserRole;
+import org.springframework.data.repository.CrudRepository;
+
+public interface WorkspaceUserRoleDao extends CrudRepository<WorkspaceUserRole, Long> {
+   WorkspaceUserRole findWorkspaceUserRolesByWorkspaceAndUser(Workspace workspace, User user);
+}
+
```
rushtong: Do you need a whole new DAO for this? It seems like this query should be in `WorkspaceServiceImpl`

api/config/config_local.json
```diff
@@ -42,5 +42,8 @@
   "mandrill": {
     "fromEmail": "donotreply@fake-research-aou.org",
     "sendRetries": 3
+  },
+  "notebookCohortCacheConfig": {
+    "count":10
   }
 }
```
rushtong: Whatever changes you make here should be in all of the configs.

api/db/changelog/db.changelog-44-notebook_cohort_cache.xml
NehaBroad: I did think about this earlier, but by having cohort as fk I am gaining the functionality of safe delete as soon as Cohorts are deleted.

rushtong: Great point. Feel free to ignore this suggestion then.

api/db/changelog/db.changelog-44-notebook_cohort_cache.xml
```diff
@@ -0,0 +1,35 @@
...
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="nehasaxena" id="db.changelog-44-cacheTable">
+        <createTable tableName="notebook_cohort_cache">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="user_workspace_id" type="bigint">
```
calbach: This is an interesting choice - I'm not sure how I feel about it. Semantically I'd say this cache table is more a child to the user, and references other resources - rather than being a child table of the workspace role; but it's not too big of a stretch. My main concern is that the workspace role table is not actually authoritative and I'm not sure we should add any dependencies on it: https://docs.google.com/document/d/1n8G2x0qt5o3Y5ExXcIlC77Li6CN7Xmfv83qzM-9wfkI/edit

For these reasons I think my preference would be to just put user_id and workspace_id as seperate columns here. Let me know if you had other thinking on this.

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+  public void updateNotebook(Workspace workspace, User user, String notebookName,
+      Timestamp lastAccessDateTime) {
+    WorkspaceUserRole workspaceUserRole =
+        getworkspaceUserRoleDao().findWorkspaceUserRolesByWorkspaceAndUser(workspace, user);
+    if (workspaceUserRole == null) {
+      throw new BadRequestException(String.format(
+          "No workspace %s was found for user %s ", workspace.getName(), user.getUserId()));
+    }
+    NotebookCohortCache cache = getDao().findByUserWorkspaceIdAndNotebookName
+        (workspaceUserRole, notebookName);
```
calbach: I don't have a particular style guide to cite here but I would definitely not wrap the parens on a function call like this. Just moving the first paren up one line would work fine.

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+    cache.setLastAccessTime(lastAccessDateTime);
+    getDao().save(cache);
+  }
+
+  /**
+   * Checks if cohort is already in cache for given workspace and user
+   * if yes, update the lastmodified time only
+   * If no, check the number of cache entries for given workspace and user
+   * if it is above 10, delete the row with least lastAccessTime and add a new entry
```
calbach: nit: don't 10 hardcode in the comment since this is configurable

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+  NotebookCohortCacheDao notebookCohortCacheDao;
+  @Autowired
+  WorkspaceUserRoleDao workspaceUserRoleDao;
+
+  public NotebookCohortCacheServiceImpl(Provider<WorkbenchConfig> configProvider) {
+    this.configProvider = configProvider;
+  }
+
+  @Override
+  public NotebookCohortCacheDao getDao() {
```
calbach: WHy are these dao getters in the public interface?

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheService.java
```diff
@@ -0,0 +1,18 @@
...
+import org.pmiops.workbench.db.model.Workspace;
+
+import java.sql.Date;
+import java.sql.Timestamp;
+
+public interface NotebookCohortCacheService {
+
+  NotebookCohortCacheDao getDao();
+
+  void updateNotebook(Workspace workspace, User user, String notebookName, Timestamp lastAccessDateTime);
```
calbach: nit: I would probably use slightly different language to clarify that this refers to the cache entry rather than the notebook itself:

Maybe:

```
putNotebook()
putCohort()
removeNotebook()
```

or if you like update/delete:

```
updateNotebookEntry()
updateCohortEntry()
deleteNotebookEntry()
```

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    if (workspaceUserRole == null) {
+      throw new BadRequestException(String.format(
+          "User %s with workspace %s not found.", workspace.getName(), user.getUserId()));
+    }
+    getDao().
+        deleteNotebookCohortCacheByUserWorkspaceIdAndNotebookName(workspaceUserRole, notebookName);
+  }
+
+  //Delete the last access row for given workspace and user
+  private void handleWorkspaceUserLimitInCache(WorkspaceUserRole user_workspace) {
```
calbach: camel case

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    long count = getDao().countNotebookCohortCachesByUserWorkspaceId(user_workspace);
+    if (count == configProvider.get().notebookCohortCacheConfig.count) {
```
calbach: Please use >=.

Optionally, wrap this whole thing in a `while` to handle going over the cache. This could matter if we changed the config value from 20 -> 10, for example.

api/src/main/java/org/pmiops/workbench/db/dao/NotebookCohortCacheServiceImpl.java
```diff
@@ -0,0 +1,124 @@
+package org.pmiops.workbench.db.dao;
```
calbach: Test?

Aside from the obvious, it would also be good to cover deletion of a non-existent cache entry.

api/src/main/java/org/pmiops/workbench/db/model/NotebookCohortCache.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.persistence.ManyToMany;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+import java.sql.Timestamp;
+
+@Entity
+@Table(name = "notebook_cohort_cache")
+public class NotebookCohortCache {
```
calbach: It bothers me a bit that the model/table name is "cache", when really these are entries in a cache. Some alternative naming ideas; in particular it makes some of the service code look a bit off:

- `UserRecentResource`
- `NotebookCohortCacheEntry`
- `UserCachedResource`

rushtong: +1. I like `UserRecentResource` the best of these options

api/db/changelog/db.changelog-44-notebook_cohort_cache.xml
NehaBroad: Adding user_id and workspace _id to the table will lead to repetitive values, which is way i wanted to separate them out to another table. And while doing so i stumble upon this table and thought of re-using it.

NehaBroad:
Had a discussion with @blrubenstein and realized i was maintaining the cache limit on user workspace pair rather than just user. Hence will be changing the handleUserLimit logic.

api/db/changelog/db.changelog-44-notebook_cohort_cache.xml
calbach: Not sure I follow on repetitive values. Do you mean storing the user ID+workspace multiple times? I think this should be OK.

Anyways, since user+workspace is sort of a best-effort thing per that doc, I wouldn't try to tie our infrastructure too tightly to it (workspace+user relationships may be missing from that table).

calbach: (also if you're trying to switch to a user-based cache rather than user+workspace based cache, separating the columns will probably make this query more natural)

rushtong: It makes more sense to me that this ties to the user table. Tying it to the user-workspace table, in my opinion, overloads the intended purpose of that table.

api/db/changelog/db.changelog-44-user-recent-resource.xml
```diff
@@ -0,0 +1,45 @@
...
+            <column name="user_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="workspace_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cohort_id" type="bigint">
+                <constraints nullable="true"/>
+            </column>
+            <column name="notebook_name" type="varchar(255)">
```
calbach: The 255 is arbitrary, right? I'd probably bump this to 1024, which is the maximum GCS object name length: https://cloud.google.com/storage/docs/naming#objectnames

If we get a notebook name that long, I'm sure we'll have other problems, but 1024 at least guarantees this SQL won't be the limiting factor.

api/db/changelog/db.changelog-44-user-recent-resource.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="nehasaxena" id="db.changelog-44-cacheTable">
+        <createTable tableName="user_recent_resource">
+            <column name="id" type="bigint" autoIncrement="true">
```
calbach: I'm not a MySQL expert, but in general having a globally increasing ID synchronized across all users (which would otherwise be independent) doesn't seem like it would have ideal performance.

Reading a few threads about this online indicates that this shouldn't be a major concern so I'm probably fine going with what you have here, but I'm curious if others have experience/thoughts on this. The alternative would be to use a GUID or just a random int64

api/config/config_prod.json
```diff
@@ -41,5 +41,8 @@
   "mandrill": {
     "fromEmail": "donotreply@researchallofus.org",
     "sendRetries": 3
+  },
+  "userRecentResourceConfig": {
+    "userEntrycount":10
```
calbach: Do you think we'll ever update this or would need to update it out of band? If not, I might argue for just a constant. This config file has a lot of values that I'd argue should probably be constants.

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentResourceServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+   * @param workspaceId
+   * @param userId
+   * @param notebookName
+   */
+  @Override
+  public void deleteNotebookEntry(long workspaceId, long userId, String notebookName) {
+    getDao().deleteUserRecentResourceByUserIdAndWorkspaceIdAndNotebookName(workspaceId, userId, notebookName);
+  }
+
+  //Check number of entries in user_recent_resource for user, delete entries with least
```
calbach: I'd use a javadoc style comment in this location

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentResourceServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+      resource.setNotebookName(null);
+    }
+    resource.setLastAccessTime(lastAccessDateTime);
+    getDao().save(resource);
+  }
+
+  /**
+   * Deletes notebook entry from user_recent_resource
+   *
+   * @param workspaceId
```
calbach: Add descriptions or remove these `@param`s

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    config.userRecentResourceConfig = new WorkbenchConfig.UserRecentResourceConfig();
+    config.userRecentResourceConfig.userEntrycount = 3;
+    userRecentResourceService = new UserRecentResourceServiceImpl(Providers.of(config));
+    userRecentResourceService.setDao(notebookCohortCacheDao);
+  }
+
+
+  @Test
+  public void testInsertCohortEntry() {
+    userRecentResourceService.updateCohortEntry(workspaceId, userId, cohortId, new Timestamp(new Date().getTime()));
```
calbach: Please use a fixed test time (and offsets from it) thoughout to avoid any dependency (accidental or otherwise) on the time at which this test is run

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    assertEquals(noOfRows, 1);
+    userRecentResourceService.updateCohortEntry(workspaceId, userId, cohortId, new Timestamp(new Date().getTime()));
+    noOfRows = userRecentResourceService.getDao().count();
+    assertEquals(noOfRows, 1);
+  }
+
+  @Test
+  public void testUpdateNotebookAccessTime() {
+    userRecentResourceService.updateNotebookEntry(workspaceId, userId, "notebook1", new Timestamp(new Date(2018, 3, 3).getTime()));
+    long noOfRows = userRecentResourceService.getDao().count();
```
calbach: ultranit: `noOfRows` looks a bit awkward to me, I might go with `numRows`. Alternatively I might also just inline this in the assert below

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    userRecentResourceService.updateNotebookEntry(workspaceId, userId, "notebook1", new Timestamp(new Date(2018, 3, 3).getTime()));
+    userRecentResourceService.updateNotebookEntry(2l, userId, "notebook2", new Timestamp(new Date(2018, 4, 3).getTime()));
+    userRecentResourceService.updateCohortEntry(workspaceId, userId, cohortId, new Timestamp(new Date(2018, 5, 3).getTime()));
+    long noOfRows = userRecentResourceService.getDao().count();
+    assertEquals(noOfRows, 3);
+
+    userRecentResourceService.updateNotebookEntry(workspaceId, userId, "notebook4", new Timestamp(new Date(2018, 6, 3).getTime()));
+    noOfRows = userRecentResourceService.getDao().count();
+    assertEquals(noOfRows, 3);
+    UserRecentResource cache = userRecentResourceService.getDao().findByUserIdAndWorkspaceIdAndNotebookName(1l, 1l, "notebook1");
```
calbach: nit: perhaps use the userId, workspaceId instead of 1L here

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    assertNull(cache);
+  }
+
+  @Test
+  public void testDeleteNotebookEntry() {
```
calbach: Please add a test case which deletes something that isn't in the cache

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Autowired
+  UserDao userDao;
+  @Autowired
+  WorkspaceDao workspaceDao;
+  @Autowired
+  CohortDao cohortDao;
+  @Autowired
+  UserRecentResourceDao notebookCohortCacheDao;
+  @Mock
```
calbach: opt: you can define a `TestConfiguration` in this file and provide an instance of WorkbenchConfig to avoid using a mock here. Then you can also just Autowire in your target service instead of constructing it yourself.

Example: https://github.com/all-of-us/workbench/blob/7b992ab8fc1ccaf604326833aedafc77069a7c2d/api/src/test/java/org/pmiops/workbench/api/OfflineClusterControllerTest.java#L67

api/db/changelog/db.changelog-44-user-recent-resource.xml
rushtong: In practice, given that this table will be limited in size per user, any choice of ID datatype should be OK. In other systems with many hundreds of thousands of rows, we've seen real problems with GUIDs - indexing became a bottleneck. Random ints and auto incremented ids shouldn't be a problem.

calbach: To clarify, I'm not concerned with size or indexing overhead. Mainly my worry was around having a synchronized step when inserting two independent rows (which is highly likely to happen concurrently given the use case). I can imagine such synchronization might be a property of any MySQL row insertion anyways though, so making it autoincrement might have very little overhead.

api/src/test/java/org/pmiops/workbench/db/service/UserRecentResourceServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+    userRecentResourceService.updateNotebookEntry(workspaceId, userId, "notebook1", new Timestamp(clock.millis()));
+    long rowsCount = userRecentResourceService.getDao().count();
+    assertEquals(rowsCount, 1);
+    userRecentResourceService.deleteNotebookEntry(workspaceId, userId, "notebook1");
+    rowsCount = userRecentResourceService.getDao().count();
+    assertEquals(rowsCount, 0);
+  }
+
+  @Test
+  public void testDeleteNonExistantNotebookEntry() {
```
calbach: spelling: existent

api/db/changelog/db.changelog-44-user-recent-resource.xml
rushtong: 👍

rushtong:
@NehaBroad - Please note that there is a merge conflict.

PR #1066, opened 2018-08-06
[RW-1096][risk=no] Fix incorrect count for cloned cohorts

dolbeew:
Added hasAttributes property to search parameters of cloned cohorts so the attributes will get properly mapped


ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -505,6 +505,7 @@ export class CohortSearchActions {
    * Loads a JSONified SearchRequest into the store
    */
   loadFromJSON(json: string): void {
+    console.log(JSON.parse(json));
```
freemabd1: remove this?

dolbeew: Done.

PR #1069, opened 2018-08-07
[RW-1102][risk=no] create yaml contract for Labs api calls(autocomplete and tree sub search)

freemabd1:
create yaml contract for Labs api calls(autocomplete and tree sub search)

### DIFF
--- a/api/src/main/resources/cb_search_api.yaml
+++ b/api/src/main/resources/cb_search_api.yaml
@@ -85,6 +85,52 @@ paths:
           schema:
             $ref: "#/definitions/CriteriaListResponse"
 
+  /v1/cohortbuilder/{cdrVersionId}/criteria/{type}/search/{value}:
+    parameters:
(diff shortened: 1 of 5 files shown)


api/src/main/resources/cb_search_api.yaml
```diff
@@ -85,6 +85,52 @@ paths:
...
+          schema:
+            $ref: "#/definitions/CriteriaListResponse"
+
+  /v1/cohortbuilder/{cdrVersionId}/criteria/subtree/{id}:
+    parameters:
+      - $ref: '#/parameters/cdrVersionId'
+    get:
+      tags:
+        - cohortBuilder
+      description: Returns matches on criteria table by code or name
```
dolbeew: nit: Change description to match function

freemabd1: Done.

PR #1073, opened 2018-08-08
Add a bunch of tests

blrubenstein:
This just adds a whole bunch of test files with basic render tests. 

* **no**: None 

Please also:

* Get thumbs from reviewer(s)

### DIFF
--- a/ui/src/app/views/account-creation-modals/component.spec.ts
+++ b/ui/src/app/views/account-creation-modals/component.spec.ts
@@ -16,24 +16,8 @@ import {ServerConfigService} from '../../services/server-config.service';
 
 import {AccountCreationModalsComponent} from '../account-creation-modals/component';
 
-class AccountCreationModalsPage {
-  fixture: ComponentFixture<AccountCreationModalsComponent>;
-  component: AccountCreationModalsComponent;
-
-  constructor(testBed: typeof TestBed) {
-    this.fixture = testBed.createComponent(AccountCreationModalsComponent);
-    this.component = this.fixture.componentInstance;
-    this.readPageData();
-  }
-
-  readPageData() {
-    updateAndTick(this.fixture);
-  }
-}
-
-
 describe('AccountCreationModalsComponent', () => {
-  let page: AccountCreationModalsPage;
+  let fixture: ComponentFixture<AccountCreationModalsComponent>;
   beforeEach(fakeAsync(() => {
     TestBed.configureTestingModule({
       imports: [
@@ -53,13 +37,13 @@ describe('AccountCreationModalsComponent', () => {
           })
         }]
     }).compileComponents().then(() => {
-      page = new AccountCreationModalsPage(TestBed);
+      fixture = TestBed.createComponent(AccountCreationModalsComponent);
       tick();
     });
   }));
 
   it('should render', fakeAsync(() => {
-    page.readPageData();
-    expect(page.fixture).toBeTruthy();
+    updateAndTick(fixture);
+    expect(fixture).toBeTruthy();
   }));
 });
--- a/ui/src/app/views/account-creation-modals/component.spec.ts
+++ b/ui/src/app/views/account-creation-modals/component.spec.ts
@@ -47,8 +47,7 @@ describe('AccountCreationModalsComponent', () => {
         ClarityModule.forRoot()
       ],
       declarations: [
-        AccountCreationModalsComponent,
-        RoutingSpinnerComponent
+        AccountCreationModalsComponent
       ],
       providers: [
         { provide: ProfileService, useValue: profileServiceStub },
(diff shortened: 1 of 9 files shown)


ui/src/app/views/account-creation-modals/component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  readPageData() {
+    updateAndTick(this.fixture);
+  }
+}
+
+
+describe('AccountCreationModalsComponent', () => {
+  let profileServiceStub: ProfileServiceStub;
```
calbach: I would suggest that we only need this variable if we are in fact testing some functionality via the stub. If we're simply using it to provide a value (and not verifying anything), you can just inline the new ProfileServiceStub in the provider definition.

ui/src/app/views/account-creation-modals/component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    profileServiceStub = new ProfileServiceStub();
+    TestBed.configureTestingModule({
+      imports: [
+        RouterTestingModule,
+        FormsModule,
+        ClarityModule.forRoot()
+      ],
+      declarations: [
+        AccountCreationModalsComponent,
+        RoutingSpinnerComponent
```
calbach: Are these really needed throughout? If so, why?

ui/src/app/views/account-creation-success/component.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        FormsModule,
+        ClarityModule.forRoot()
+      ],
+      declarations: [
+        AccountCreationModalsComponent,
+        AccountCreationSuccessComponent,
+        RoutingSpinnerComponent
+      ],
+      providers: [
+        { provide: AccountCreationComponent, useValue: {profile: {
```
calbach: I'm not sure whether I like adding this test, just because I don't want to entrench the current approach. I'm really not convinced about this "inject the previous component" approach we're using here and would probably like this to be cleaned up. (as highlighted by the weird provider setup going on here).

ui/src/app/views/admin-review-workspace/component.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+} from '../../../testing/test-helpers';
+
+import {ProfileStorageService} from '../../services/profile-storage.service';
+import {ServerConfigService} from '../../services/server-config.service';
+
+import {AdminReviewWorkspaceComponent} from '../admin-review-workspace/component';
+import {BugReportComponent} from '../bug-report/component';
+import {RoutingSpinnerComponent} from '../routing-spinner/component';
+
+class AdminReviewWorkspacePage {
```
calbach: It's probably worth having another discussion about the pros/cons of this approach - I know we discussed this briefly a while back, but a comparison of the two approaches with the current team might be useful (and if there's consensus, we should switch existing tests to all match this model).

ui/src/app/views/account-creation-modals/component.spec.ts
blrubenstein: 👍 Will do

ui/src/app/views/account-creation-success/component.spec.ts
blrubenstein: I'm also not a huge fan of this test. Its probably the test I am most hesitant about. Let me check with Ursa and Greg on it and see what they think.

ui/src/app/views/admin-review-workspace/component.spec.ts
blrubenstein: The page approach, you mean?

ui/src/app/views/account-creation-modals/component.spec.ts
blrubenstein: Looks like no. Will remove.

ui/src/app/views/account-creation-success/component.spec.ts
blrubenstein: Okay! I talked to Greg, and I think it is worth having this test in order to test things where the behavior is purely scoped to the AccountCreationSuccessComponent. However, I am changing the profile value to be more of a variable/use our stub variables, rather than be hardcoded.

ui/src/app/views/admin-review-workspace/component.spec.ts
calbach: Yes.

ui/src/app/views/account-creation-success/component.spec.ts
calbach: To clarify, I'm fine with the test and how it is written. What I don't like is the component implementation itself and I think we should probably revisit it at some point.

Keeping the test in this PR sounds fine.

blrubenstein:
@calbach removed pages as per discussion on slack. PTAL when you get the chance.

PR #1074, opened 2018-08-09
[RW-1055][risk=no] Drop the login "from" feature for now

calbach:
Unfortunately we can’t fix this without doing a deeper rework of our sign in flow (switching to server side oauth) or breaking certain users' ability to login (reverting to the previous pop-up approach) because we can’t generate a dynamic redirect URI while also matchin our redirect whitelist. I tried tacking on query parameters and hash fragments, nothing makes it through, and GoogleAuth doesn't appear to have any capability to attach extras onto the sign in call.

Secondarily, this removes the navigation on user signout, as this caused a navigate/reload combo which made the page half-navigate and half-render on sign out before reloading the page.

### DIFF
--- a/ui/src/app/guards/sign-in-guard.service.ts
+++ b/ui/src/app/guards/sign-in-guard.service.ts
@@ -9,7 +9,6 @@ import {
 import {Observable} from 'rxjs/Observable';
 
 import {SignInService} from 'app/services/sign-in.service';
-import {navigateLogin} from 'app/utils';
 
 
 declare const gapi: any;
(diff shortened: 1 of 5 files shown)


calbach:
> Question: Can we rework our logic to do a direct navigate from here?

It's easy enough to do a navigate. The problem is that you cannot pass arbitrary URL parameters through the sign in flow as they all need to be explicitly whitelisted, i.e. where are you getting the "from" parameter? (without resorting to something hacky)

blrubenstein:
Got it. Makes sense.

calbach:
Thinking more, not having this "from" ability will be very annoying for link sharing, so we should aim to address the full fix (and reinstate this feature) sooner rather than later.

PR #1075, opened 2018-08-09
[risk=no] Delete apparently unused generate-cdr CSVs

calbach:
Please advise if any of the other files around here can be removed as well: https://github.com/all-of-us/workbench/tree/master/api/db-cdr/generate-cdr/csv


calbach:
Sure - but specifically this achilles_results file is not necessary - right?

freemabd1:
I faintly remember @biopete saying this is true, but he should confirm this.

biopete:
Yeah. That was an accident getting put in there.  The new pr we merge will delete it.

calbach:
Dropping then

PR #1082, opened 2018-08-13
[RW-872][risk=low] Don't store enum ordinals for WorkspaceAccessLevel

calbach:
See ticket for background. Looking for feedback on this pattern. If it looks good, will apply to all other stored enum values as well.


rushtong:
I agree with this approach over storing ordinal values.

calbach:
Correct - I'm using all the same values as the existing enum ordinals values.

PR #1083, opened 2018-08-13
[RW-1012][risk=low] Fix duplicate user sharing bug

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1012

## Changes
* Change the share workspace API to return the new/updated list of user roles
* Update the UI to rely on that list instead of managing an internal list of user roles.
* Update test stub to match the updated API

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -7,6 +7,7 @@ import com.google.common.base.Strings;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
@@ -708,6 +709,13 @@ public class WorkspacesController implements WorkspacesApiDelegate {
     dbWorkspace = workspaceService.updateUserRoles(dbWorkspace, dbUserRoles);
     ShareWorkspaceResponse resp = new ShareWorkspaceResponse();
     resp.setWorkspaceEtag(Etags.fromVersion(dbWorkspace.getVersion()));
+    List<UserRole> updatedUserRoles = dbWorkspace.getWorkspaceUserRoles()
+        .stream()
+        .map(r -> new UserRole().email(r.getUser().getEmail()).role(r.getRole()))
+        // Reverse sorting arranges the role list in a logical order - owners first, then by email.
+        .sorted(Comparator.comparing(UserRole::getRole).thenComparing(UserRole::getEmail).reversed())
+        .collect(Collectors.toList());
+    resp.setItems(updatedUserRoles);
     return ResponseEntity.ok(resp);
   }
 
--- a/ui/src/app/views/workspace-share/component.ts
+++ b/ui/src/app/views/workspace-share/component.ts
@@ -95,13 +95,17 @@ export class WorkspaceShareComponent implements OnInit {
     }
     if (!this.usersLoading) {
       this.usersLoading = true;
-      const updateList = Array.from(this.workspace.userRoles);
-      updateList.push({
-        email: this.convertToEmail(this.toShare),
-        role: this.selectedAccessLevel
-      });
-
-      this.workspacesService.shareWorkspace(
+      const email = this.convertToEmail(this.toShare);
+      const role = this.selectedAccessLevel;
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -708,6 +709,13 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     dbWorkspace = workspaceService.updateUserRoles(dbWorkspace, dbUserRoles);
     ShareWorkspaceResponse resp = new ShareWorkspaceResponse();
     resp.setWorkspaceEtag(Etags.fromVersion(dbWorkspace.getVersion()));
+    List<UserRole> updatedUserRoles = dbWorkspace.getWorkspaceUserRoles()
+        .stream()
+        .map(r -> new UserRole().email(r.getUser().getEmail()).role(r.getRole()))
+        // Reverse sorting arranges the role list in a logical order - owners first, then by email.
```
calbach: Based on the wording, this is what I'm assuming but double-checking: we aren't depending on this ordering in any way right? This is purely for display niceness?

ui/src/app/views/workspace-share/component.ts
```diff
@@ -95,13 +95,17 @@ export class WorkspaceShareComponent implements OnInit {
...
-        role: this.selectedAccessLevel
-      });
-
-      this.workspacesService.shareWorkspace(
+      const email = this.convertToEmail(this.toShare);
+      const role = this.selectedAccessLevel;
+      if (this.checkUnique(email, role)) {
+        // A user can only have one role on a workspace so we replace them in the list
+        const updateList = Array.from(this.workspace.userRoles)
+          .filter(function(r) { return r.email !== email; });
```
calbach: Should use an ES6 arrow function (and can make it shorter with that): `.filter(r => r.email !== email);`

ui/src/app/views/workspace-share/component.ts
```diff
@@ -122,7 +126,10 @@ export class WorkspaceShareComponent implements OnInit {
             }
             this.usersLoading = false;
           }
-      );
+        );
+      } else {
+        this.usersLoading = false;
```
calbach: Alternatively, it would probably fix this logic and be more readable to put `this.usersLoading = true` closer to the callback (inside that `if`)

ui/src/app/views/workspace-share/component.ts
```diff
@@ -196,4 +203,12 @@ export class WorkspaceShareComponent implements OnInit {
     this.locationService.back();
   }
 
+  // Checks for an email + role combination in the current list of user roles.
+  checkUnique(email: String, role: WorkspaceAccessLevel): Boolean {
+    return Array.from(this.workspace.userRoles)
+      .filter(function(r) { return r.email === email; })
```
calbach: Arrow functions here as well

ui/src/app/views/workspace-share/component.ts
```diff
@@ -196,4 +203,12 @@ export class WorkspaceShareComponent implements OnInit {
     this.locationService.back();
   }
 
+  // Checks for an email + role combination in the current list of user roles.
+  checkUnique(email: String, role: WorkspaceAccessLevel): Boolean {
```
calbach: primitive type seems better `boolean`

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
rushtong: That's exactly right, it's purely for display. When adding and removing users, the UI replaces the old list of roles with the new list of roles. If we don't default sort in some way (either here in the java, or in the typescript), the order bounces around between adds/removes.

ui/src/app/views/workspace-share/component.ts
rushtong: Done

ui/src/app/views/workspace-share/component.ts
rushtong: Nice catch - done.

ui/src/app/views/workspace-share/component.ts
rushtong: Done

ui/src/app/views/workspace-share/component.ts
rushtong: Done

PR #1088, opened 2018-08-14
[RW-1122][risk=low] Update create workspace conflict error message

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1122

## Changes
Added special-casing for the conflict case.

### DIFF
--- a/ui/src/app/views/workspace-edit/component.spec.ts
+++ b/ui/src/app/views/workspace-edit/component.spec.ts
@@ -93,7 +93,8 @@ describe('WorkspaceEditComponent', () => {
       ]}).compileComponents();
   }));
 
-  it('should show a conflict-specific error when creating a name conflict workspace', fakeAsync(() => {
+  it('should show a conflict-specific error when creating a name conflict workspace',
+    fakeAsync(() => {
     spyOn(TestBed.get(Router), 'navigate');
     setupComponent(WorkspaceEditMode.Create);
     testComponent.workspace.namespace = WorkspaceStubVariables.DEFAULT_WORKSPACE_NS;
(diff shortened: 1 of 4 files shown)


ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -91,6 +93,50 @@ describe('WorkspaceEditComponent', () => {
       ]}).compileComponents();
   }));
 
+  it('should show a conflict-specific error when creating a name conflict workspace',
+    fakeAsync(() => {
+    spyOn(TestBed.get(Router), 'navigate');
```
blrubenstein: This may warrant adding `.and.doNothing()`? Just to be a little more explicit about behavior?

ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -91,6 +93,50 @@ describe('WorkspaceEditComponent', () => {
...
+    const originalSize = workspacesService.workspaces.length;
+    fixture.detectChanges();
+    fixture.debugElement.query(By.css('.add-button'))
+      .triggerEventHandler('click', null);
+    updateAndTick(fixture);
+    updateAndTick(fixture);
+    expect(workspacesService.workspaces.length).toBe(originalSize);
+    const modalTitle = fixture.debugElement.query(By.css('.modal-title'));
+    const modalBody = fixture.debugElement.query(By.css('.modal-body'));
+    console.log(modalBody);
```
blrubenstein: Nit: Remove extra console logs if unnecessary.

rushtong: I hate missing those.

ui/src/app/views/workspace-edit/component.spec.ts
rushtong: So, there doesn't seem to be a `.and.doNothing()` or even just a `doNothing()`. I'll add a comment here just in case.

blrubenstein: Oh, okay! My bad. I thought there was, but may have thought wrong!

rushtong: Turns out we don't need that spy for these test cases. They are needed for other cases, but not these error ones.

PR #1091, opened 2018-08-14
[RW-872][risk=low] Don't store enum ordinals in Workbench

calbach:
Part 2 of 3. Up next: Cohort Builder models.

- Centralize the conversion logic into a single `StorageEnums` utility to minimize overhead and avoid duplication for enums shared across models.
- Create a parameterized test to ensure all enums can be mapped to storage and to reduce incremental burden of adding new enums


api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
```diff
@@ -152,11 +151,11 @@ public void setFirecloudName(String firecloudName) {
   }
 
   @Column(name = "data_access_level")
-  public DataAccessLevel getDataAccessLevel() {
+  public Short getDataAccessLevel() {
```
danqrodney: I think it'd be nice if whenever using Short like this, we added a comment indicating the enum class it corresponds to / what method to use to convert.

We *could* even add getters / setters to the model objects using the enum classes (which JPA would ignore)... the downside being we'd still have a dependency on Swagger in our model objects (which makes our build and particularly tools build messier.)

danqrodney: If it's tricky, don't bother with the extra methods; but let's document what enum value we're using here.

calbach: Several hours later, turns out my suspicions were also confirmed and Spring data dark magic of course got in the way. Apparently it will attempt to glean some meaning from public method names on your entity. I'm not sure of the exact rules or why it is doing this when it already has explicit annotations, but this code appears to be involved somehow: https://github.com/hibernate/hibernate-orm/blob/736c913f1428e16a5a3ef26b52d9ec91ff323f25/hibernate-core/src/main/java/org/hibernate/type/TypeResolver.java#L95

At least one trigger for this seems to be having a getter/setter-like syntax. So unfortunately these helpers will need to have slightly unnatural naming, I went with an "enum" prefix which dodges the problem.

I'm almost tempted to revert this because I really dislike being at the mercy of Spring's opaque automagical model handling, but I think it's probably worth going with this version since the client code is so much cleaner.

danqrodney:
I think you can annotate methods as @Transient also to avoid this.

calbach:
Nice find, thanks. Done.

api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
```diff
@@ -300,12 +307,32 @@ public void setContainsUnderservedPopulation(Boolean containsUnderservedPopulati
...
 
-  public void setUnderservedPopulationSet(Set<UnderservedPopulationEnum> newUnderservedPopulationSet) {
-    this.underservedPopulationSet = newUnderservedPopulationSet;
+  public void setUnderservedPopulationsShort(Set<Short> newUnderservedPopulations) {
+    this.underservedPopulationSet = newUnderservedPopulations;
+  }
+
+  @Transient
+  public Set<UnderservedPopulationEnum> getUnderservedPopulationsEnum() {
+    Set<Short> from = getUnderservedPopulationsShort();
```
rushtong: If this were an `Optional`, the if condition could probably be simplified.

calbach: Tried this but unfortunately I need the annotated getter to have a type that can be mapped to the stored type, else Spring Data JPA cannot map it to the SQL column. If we wanted an `Optional` version of this method, I'd need to add another method and make it `@Transient` as well which would probably not be worth it.

https://vladmihalcea.com/the-best-way-to-map-a-java-1-8-optional-entity-attribute-with-jpa-and-hibernate/

rushtong: Great article, thanks, and yeah, more effort than it saves.

api/src/main/java/org/pmiops/workbench/db/model/User.java
```diff
@@ -96,14 +97,23 @@ public void setContactEmail(String contactEmail) {
   }
 
   @Column(name = "data_access_level")
-  public DataAccessLevel getDataAccessLevel() {
+  public Short getDataAccessLevel() {
```
danqrodney: In other places you've got Short on the ends of getters and setters... want to do that in this file too?

Once you've got the Short suffix everywhere, you arguably could leave the "Enum" off the getters / setters that return enums and have them use the original method names.  (Assuming that works with JPA.) Though being explicit about enums is fine, too.

calbach: That was accidental, thanks. I reverted that approach because the naming caused problems in combination with the Dao FETCH JQL stuff. I didn't look to deeply into whether there were workarounds, but what I have now seems like the safest route.

PR #1092, opened 2018-08-15
[RW-1072][risk=no] Adding api call for Measurements(Labs)

freemabd1:
**There are 4 scenarios in terms of attributes for lab measurements:**

1. Lab has text values only - no numeric or categorical values. Because all text will be suppressed, this means these labs can’t be searched other than “lab exists”
2. Lab has numeric values only
3. Lab has categorical values only
4. Lab has both numeric and categorical values

**Scenario 1 (Text)**
- any value (lab exists)

**Scenario 2 (Numerical - can choose “any” or a numerical operator)**
- dropdown list with (any, equals, GTE, LTE or between)

**Scenario 3 (Categorical - can choose “any” or a list of predetermined values based on lab type)**
- dropdown list with (any, detected, negative, no, normal, positive, present, yes)

**Scenario 4 (Both - combination of 2 and 3)**
- dropdown list with (any, equals, GTE, LTE or between, detected, negative, no, normal, positive, present, yes)


api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/MeasurementQueryBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    Map<String, QueryParameterValue> queryParams = new HashMap<>();
+    for (SearchParameter parameter : parameters.getParameters()) {
+      validateAttributes(parameter);
+      if (parameter.getConceptId() == null) {
+        throw new BadRequestException("Please provide valid concept id for Measurements.");
+      }
+      String baseSql = MEASUREMENT_SQL_TEMPLATE.replace("${conceptId}", parameter.getConceptId().toString());
+      List<String> tempQueryParts = new ArrayList<String>();
+      for (Attribute attribute : parameter.getAttributes()) {
+        if (attribute.getOperator().equals(Operator.ANY)) {
```
danqrodney: So a couple things about the ANY operator:

1) We have logic that handles different Operator options in here:

https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/utils/OperatorUtils.java#L18

It doesn't handle ANY. That could lead to an NPE in here, for example:

https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java#L339

We face a similar problem with BETWEEN. We have hacked around this problem by stripping out the BETWEEN operator from our pyclient library here:

https://github.com/all-of-us/pyclient/blob/master/libproject/devstart.rb#L51

So perhaps we should do the same for ANY. But at some point we may need to reconsider this approach... perhaps we need different Operator enums for the client API vs. this API or something.

2) I was wondering if it's possible to express ANY in some other way; like "NOT_EQUALS" with a value of null, and have the SQL generation code smart enough to do "IS NOT NULL". That's what we're doing right now here:

https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java#L334

Or just not having an Operator or value set at all.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/MeasurementQueryBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    boolean textAttrs =
+      attrs.stream().filter(text::test).collect(Collectors.toList()).size() != 1;
+    boolean numericalAttrs =
+      attrs.stream().filter(numerical::test).collect(Collectors.toList()).size() != 1;
+    boolean categoricalAttrs =
+      attrs.stream().filter(categorical::test).collect(Collectors.toList()).size() != 1;
+    boolean bothAttrs =
+      attrs.stream().filter(both::test).collect(Collectors.toList()).size() == 0;
+    if (bothAttrs && textAttrs && numericalAttrs && categoricalAttrs) {
+      throw new BadRequestException("Please provide valid search attributes" +
```
danqrodney: At the moment we don't really expect researchers to make their own cohort specs in code, so this isn't a huge deal... but at some point we probably will, at which point it'd be nice if we returned specific validation error messages rather than this generic message.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/MeasurementQueryBuilder.java
freemabd1: Done.

api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/MeasurementQueryBuilder.java
freemabd1: I can definitely do this, but do you mind if it's in a followup PR? Will is waiting on this api call so he can complete his UI implementation of this.

danqrodney: Sure, there's no rush on this. If you want to put it off until much later, fine by me, but maybe file a ticket tracking it.

freemabd1: Thanks. Created a ticket for this: https://precisionmedicineinitiative.atlassian.net/browse/RW-1164

PR #1096, opened 2018-08-15
[RW-1132][risk=low] Don't store cohort builder enum ordinals

calbach:
PR 3/3 for converting workbench off enum ordinal value storage.


calbach:
Thought to check the bigquery integration tests, which were indeed failing - fixed that last reference.

PR #1099, opened 2018-08-16
[RISK=LOW][RW-1037]Integrate recently used resource caching

NehaBroad:
Aim of this PR is to update the table USER_RECENT_RESOURCE for the following events:

- Create Cohort
- Clone Cohort
- Review Cohort
- Open Notebook
- While retrieving the list of notebook files for workspace, it will check and delete all orphan notebooks from USER_RECENT_RESOURCE


api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -402,7 +419,9 @@ public ParticipantCohortAnnotation apply(org.pmiops.workbench.db.model.Participa
     org.pmiops.workbench.model.CohortReview responseReview = TO_CLIENT_COHORTREVIEW.apply(cohortReview, pageRequest);
     responseReview.setParticipantCohortStatuses(
       participantCohortStatuses.stream().map(TO_CLIENT_PARTICIPANT).collect(Collectors.toList()));
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
 
+    userRecentResourceService.updateCohortEntry(cohort.getWorkspaceId(), userProvider.get().getUserId(), cohortId, now );
```
freemabd1: I agree that this is the most logical place to update the cohort entry for cohort review.

api/src/main/java/org/pmiops/workbench/api/CohortsController.java
```diff
@@ -133,6 +141,7 @@ public Cohort apply(org.pmiops.workbench.db.model.Cohort cohort) {
     try {
       // TODO Make this a pre-check within a transaction?
       dbCohort = cohortDao.save(dbCohort);
+      userRecentResourceService.updateCohortEntry(workspace.getWorkspaceId(), userProvider.get().getUserId(), dbCohort.getCohortId(), now);
```
freemabd1: This is the most logical place to update the cohort entry for a clone/save cohort event.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -425,13 +429,23 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    final List<FileDetail> fileList;
     try {
       org.pmiops.workbench.firecloud.model.Workspace fireCloudWorkspace =
           fireCloudService.getWorkspace(workspaceNamespace, workspaceId)
               .getWorkspace();
       String bucketName = fireCloudWorkspace.getBucketName();
       fileList = getFilesFromNotebooks(bucketName);
+      if (fileList != null && fileList.size() > 0) {
+        long userId = userProvider.get().getUserId();
+        CompletableFuture.runAsync(() -> {
```
calbach: This is neat - is this the first time we're doing this in the codebase? I have a few questions about what's happening here:

- What thread is this running on?
- Does this run after the request returns?
- Are we guaranteed that the UserProvider will still have a reference to that previously authenticated user?
- How does this interact with unit tests? Does this get executed before the unit test completes?

+@danqrodney in case you have any Spring knowledge on the subject.

rushtong: My understanding of `runAsync` is that it does not spin off a new or different thread, so would be running in the current thread.

rushtong: Other questions are ones I have as well.

calbach: OK, I'm not entirely convinced that the request-related information is guaranteed to be properly bound in this callback, but at least it seems your callback currently doesn't require any.

I would feel a bit better if you could figure out how to test this though. One possible approach:
- Inject an Executor into this controller to pass to `runAsync`.
- In real code, bind that to `ThreadPoolTaskExecutor`, which appears to be the default recommendation from Spring: https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#scheduling
- In tests, bind that to `SyncTaskExecutor`

See also this thread for how to setup the binding in the real code: https://dzone.com/articles/spring-and-threads-taskexecutor

Alternatively, if you don't feel like figuring this out right now you could just drop the runAsync for now (though I do like the idea here).

NehaBroad: I did add the test surrounding this functionality, but after looking at @UrsaStutsman PR for notebook actions, i feel handling orphan notebooks are not going to add much (since delete and other functionality are being handled in 857 and we have another card 1195 to integrate these functions with userRecentResource). Thoughts?

NehaBroad:
I did add the test surrounding asyn functionality, but after looking at @UrsaStutsman PR for notebook actions, i feel handling orphan notebooks are not going to add much (since delete and other functionality are being handled in 857 and we have another card 1195 to integrate these functions with userRecentResource). Thoughts?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1241,6 +1249,33 @@ public void testNotebookFileList() throws Exception {
...
+    when(mockBlob2.getName()).thenReturn("notebooks/two words.ipynb");
+    when(cloudStorageService.getBlobList("bucket", "notebooks")).thenReturn(
+        ImmutableList.of(mockBlob1, mockBlob2));
+    workspacesController
+        .getNoteBookList("project", "workspace").getBody();
+
+    CompletableFuture<Void> asyncHandleOrphanNotebook = workspacesController.getAsyncHandleOrphanNotebooks();
+    Thread.sleep(5000);
+    assertEquals(asyncHandleOrphanNotebook.isDone(), true);
+    asyncHandleOrphanNotebook.thenAccept(result -> {
```
calbach: This is interesting, I'm not sure I follow why it's necessary - it seems like either the below is executing synchronously (and therefore the verification runs) - implying you don't need the callback OR the callback runs asynchronously in which case the verification wouldn't run within this test case.

NehaBroad:
Rather than this i feel it would be better to have hooks inside Jupiter api , in case users are not using the functionality we give them in All of Us and would use jupiter's instead. So i am leaning more towards removing this all together.

PR #1106, opened 2018-08-17
[RW-1149] Moving parsing of cohort spec into CohortMaterializationService. [risk=no]

danqrodney:
Hashing the spec string instead of the SearchRequest object in order to
avoid issues with non-stable hashing of enums.

(It's a little simpler to keep pagination handling logic in CohortMaterializationService from a testing perspective.)

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
+++ b/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
@@ -6,6 +6,8 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -156,15 +158,49 @@ private ParticipantCriteria getParticipantCriteria(List<CohortStatus> statusFilt
...
 
+
+  /**
+   * Materializes a cohort.
+   * @param cohortReview {@link CohortReview} representing a manual review of participants in the cohort.
+   * @param cohortSpec JSON representing the cohort criteria.
+   * @param request {@link MaterializeCohortRequest} representing the request options
+   * @return {@link MaterializeCohortResponse} containing the results of cohort materialization
+   */
+
```
calbach: nit: rm \n

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -156,15 +158,49 @@ private ParticipantCriteria getParticipantCriteria(List<CohortStatus> statusFilt
...
     FieldSet fieldSet = request.getFieldSet();
     List<CohortStatus> statusFilter = request.getStatusFilter();
     String paginationToken = request.getPageToken();
     int pageSize = request.getPageSize();
     // TODO: add CDR version ID here
-    Object[] paginationParameters = new Object[] { searchRequest, statusFilter };
+    // We require the client to specify requestHash here instead of hashing searchRequest itself;
+    // searchRequest.hashCode() includes hashing of Enum values and is thus not stable across
+    // JVMs (see [RW-1149]).
+    Object[] paginationParameters = new Object[] { requestHash, statusFilter };
```
calbach: `CohortStatus` is an enum, so shouldn't this fail with the same issue?

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Done.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Good catch! Fixed.

PR #1111, opened 2018-08-20
[RW-857][Risk=no] Notebook Actions

UrsaStutsman:
**Ready for any final reviews.**
This PR is to add the Clone, Rename and Delete actions to notebooks.
While working on this I noticed some issues with styling of the cards in the headers. I have created a JIRA ticket to address these issues since they are not limited to these cards (also on cohort and workspace cards). That ticket is: https://precisionmedicineinitiative.atlassian.net/browse/RW-1194
I have also created a JIRA ticket to refactor the confirm Delete and Rename Modals to include loading indicators to be more consistent across the site: https://precisionmedicineinitiative.atlassian.net/browse/RW-1231


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -742,6 +743,45 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     return ResponseEntity.ok(response);
   }
 
+  @Override
+  public ResponseEntity<EmptyResponse> renameNotebook(String workspace, String workspaceName, NotebookRename rename) {
```
blrubenstein: What does this do if you name a notebook the name of something that already exists in the notebooks folder? Do we want to support error messages on name collision?

ui/src/app/views/notebook-list/component.html
```diff
@@ -42,4 +72,7 @@ <h3 class="modal-title">Error:</h3>
     <button type="button" class="btn btn-primary" (click)="localizeNotebooksError = false">Ok</button>
   </div>
 </clr-modal>
-<app-bug-report></app-bug-report>
\ No newline at end of file
+<app-bug-report></app-bug-report>
+
+<app-rename-modal [resourceType]="'notebook'" [resource]="notebookInFocus" (receiveRename)="receiveRename($event)"></app-rename-modal>
+<app-confirm-delete-modal [resourceType]="'notebook'" [resource]="notebookInFocus" (receiveDelete)="receiveDelete($event)"></app-confirm-delete-modal>
```
blrubenstein: Nit: newline

ui/src/app/views/rename-modal/component.html
```diff
@@ -0,0 +1,17 @@
+<clr-modal [(clrModalOpen)]="renaming" [clrModalSize]="'md'">
+    <div class="modal-title">
+        <!--The ngIf is to avoid null pointers before the modal is opened.-->
+        <div *ngIf="renaming" class="modal-title-text">Please enter the new name for {{resource.name}}</div>
+    </div>
+    <div class="modal-body">
+        <div class="form-section">
+            <label>New Name: </label>
+            <input id="new-notebook-name" type="text" class="input"
```
blrubenstein: Nit: I would not call this `new-notebook-name` because its a generalized component. Maybe `resource-name`?

ui/src/app/views/rename-modal/component.html
```diff
@@ -0,0 +1,17 @@
...
+                   [(ngModel)]="name">
+        </div>
+    </div>
+    <div class="modal-footer">
+        <button type="button" class="btn btn-outline" (click)="renaming=false">Cancel</button>
+        <button type="button" class="btn btn-primary confirm-delete-btn" (click)="this.receiveRename.emit({name: resource.name, newName: name});">Rename {{resourceType}}</button>
+    </div>
+</clr-modal>
```
blrubenstein: Nit: Newline at end of file.

ui/src/app/views/rename-modal/component.ts
```diff
@@ -0,0 +1,23 @@
+import {Component, EventEmitter, Input, Output} from '@angular/core';
```
blrubenstein: I like this a lot. It follows the same model we have for delete-modal! :)

ui/src/app/views/notebook-list/component.html
UrsaStutsman: yup, wasn't quite ready for reviews. Sorry!

ui/src/app/views/rename-modal/component.html
UrsaStutsman: Already changed to new-name (not pushed yet - as I mentioned, not quite ready for reviews yet!)

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
UrsaStutsman: I've addressed this at the UI level.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -742,6 +743,45 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    cloudStorageService.deleteBlob(blobId);
+    return ResponseEntity.ok(new EmptyResponse());
+  }
+
+  @Override
+  public ResponseEntity<EmptyResponse> cloneNotebook(String workspace, String workspaceName, String notebookName) {
+    String fromBucket = fireCloudService.getWorkspace(workspace, workspaceName)
+      .getWorkspace()
+      .getBucketName();
+    String origPath = "notebooks/" + notebookName;
```
NehaBroad: NIT:Will it make sense to have "notebooks" as  a file const variable since its used in multiple places?

api/src/main/resources/workbench.yaml
```diff
@@ -795,6 +795,75 @@ paths:
...
+      operationId: "renameNotebook"
+      parameters:
+        - in: body
+          name: NotebookRename
+          description: new name for notebook
+          schema:
+            type: object
+            properties:
+              name:
+                type: string
```
NehaBroad: name should be a required field

ui/src/app/views/notebook-list/component.ts
```diff
@@ -158,6 +166,59 @@ export class NotebookListComponent implements OnInit, OnDestroy {
...
+    if (!(new RegExp('^.+\.ipynb$').test(newName))) {
+      newName = rename.newName + '.ipynb';
+    }
+    if (this.notebookList.filter((nb) => nb.name === newName).length > 0) {
+      this.renameModal.close();
+      this.notebookRenameConflictError = true;
+      this.duplicateName = newName;
+      return;
+    }
+   this.workspacesService.renameNotebook(this.wsNamespace, this.wsId, rename).subscribe(() => {
```
blrubenstein: Can we make this return the updated `FileDetail` for display? Just to follow the same 'update' semantics as the rest of the API?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -742,6 +743,50 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     return ResponseEntity.ok(response);
   }
 
+  @Override
+  public ResponseEntity<FileDetail> renameNotebook(String workspace, String workspaceName, NotebookRename rename) {
+    String fromBucket = fireCloudService.getWorkspace(workspace, workspaceName)
+      .getWorkspace()
+      .getBucketName();
+    String origPath = NOTEBOOKS_WORKSPACE_DIRECTORY+ "/" + rename.getName();
```
calbach: nit: space before +

api/src/main/resources/workbench.yaml
```diff
@@ -795,6 +795,76 @@ paths:
           schema:
             $ref: "#/definitions/CloneWorkspaceResponse"
 
+  # Notebooks ############################################################################
+
+  /v1/notebooks/{workspaceNamespace}/{workspaceId}/renameNotebook:
```
calbach: Notebooks are hierarchically under workspaces, so this should probably be something more like:

`/v1/workspaces/{workspaceNamespace}/{workspaceId}/notebooks/{notebookName}/rename`

e.g. see the cohort URLs

ui/src/app/views/notebook-list/component.ts
```diff
@@ -158,6 +166,58 @@ export class NotebookListComponent implements OnInit, OnDestroy {
...
+    this.renameModal.open();
+  }
+
+  receiveRename($event): void {
+    this.renameNotebook($event);
+  }
+
+  renameNotebook(rename): void {
+    let newName = rename.newName;
+    if (!(new RegExp('^.+\.ipynb$').test(newName))) {
```
calbach: We should have some protection against `/`s in the notebook name - these will not work as expected since "list-notebooks" doesn't glob into subdirectories.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -742,6 +743,50 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    return ResponseEntity.ok(fileDetail);
+  }
+
+  @Override
+  public ResponseEntity<EmptyResponse> cloneNotebook(String workspace, String workspaceName, String notebookName) {
+    String fromBucket = fireCloudService.getWorkspace(workspace, workspaceName)
+      .getWorkspace()
+      .getBucketName();
+    String origPath = NOTEBOOKS_WORKSPACE_DIRECTORY + "/" + notebookName;
+    String newPath = NOTEBOOKS_WORKSPACE_DIRECTORY + "/" + notebookName.replaceAll("\\.ipynb", " ") + "Clone.ipynb";
```
calbach: For workspaces, we name them "Clone of X" by default. May make sense to use the same convention here.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -742,6 +743,50 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    fileDetail.setName(newName);
+    fileDetail.setPath("gs://" + fromBucket + "/" + NOTEBOOKS_WORKSPACE_DIRECTORY + "/" + newName);
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    fileDetail.setLastModifiedTime(now.getTime());
+    return ResponseEntity.ok(fileDetail);
+  }
+
+  @Override
+  public ResponseEntity<EmptyResponse> cloneNotebook(String workspace, String workspaceName, String notebookName) {
+    String fromBucket = fireCloudService.getWorkspace(workspace, workspaceName)
```
calbach: nit: I would probably call this just `bucket` here and above. I used `from` and `to` prefixes in similar methods just because the source and destination were different.

ui/src/app/views/notebook-list/component.ts
```diff
@@ -158,6 +166,58 @@ export class NotebookListComponent implements OnInit, OnDestroy {
...
+  renameThis(notebook: FileDetail): void {
+    this.notebookInFocus = notebook;
+    this.renameModal.open();
+  }
+
+  receiveRename($event): void {
+    this.renameNotebook($event);
+  }
+
+  renameNotebook(rename): void {
```
calbach: These parameters should all have types

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
UrsaStutsman: Mohs and I had a conversation about this... We thought in the context of Cloning the notebooks in this list, that the user may want the clone to appear next to the cloned, which is achieved by Name Clone. If you really think we should use the other convention I will change it.

calbach: OK, sounds fine for now if we're sorting on notebook name.

Separately, I don't know whether that was actually the intended product behavior or that just happened to be how it was implemented (the mocks are ordered arbitrarily). I could imagine we may want to sort this page (by default) by last access / modification time, in which case the name matters less.

ui/src/app/views/notebook-list/component.html
```diff
@@ -18,7 +18,40 @@ <h3 class="notebooks-header">Notebooks</h3>
...
+          <button class="action-item"
+                  id="rename-{{notebook.name.replace('.ipynb', '')}}"
+                  (click)="renameThis(notebook)"
+                  [class.disabled]="actionsDisabled"
+                  [disabled]="actionsDisabled">
+            <clr-icon shape="pencil"></clr-icon>
+            Rename
+          </button>
+          <button class="action-item"
+                  id="clone-{{notebook.name.replace('.ipynb', '')}}"
```
calbach: I'm not sure about this pattern, there can be pretty arbitrary characters in notebook names including spaces which are not generally valid as ids.

ui/src/app/views/notebook-list/component.html
```diff
@@ -18,7 +18,40 @@ <h3 class="notebooks-header">Notebooks</h3>
...
+                  (click)="cloneThis(notebook.name)"
+                  [class.disabled]="actionsDisabled"
+                  [disabled]="actionsDisabled">
+            <clr-icon shape="copy"></clr-icon>
+            Clone
```
calbach: Do clone/delete have any loading indicator?

ui/src/app/views/notebook-list/component.ts
```diff
@@ -158,6 +168,64 @@ export class NotebookListComponent implements OnInit, OnDestroy {
...
+   this.workspacesService.renameNotebook(this.wsNamespace, this.wsId, rename).subscribe((newNb) => {
+      this.notebookList.splice(
+        this.notebookList.indexOf(this.notebookInFocus), 1, newNb
+      );
+     this.renameModal.close();
+   });
+  }
+
+  cloneThis(notebook: string): void {
+    this.workspacesService.cloneNotebook(this.wsNamespace, this.wsId, notebook)
```
calbach: We might want this method to return the newly cloned notebook (that would be consistent with the rename). Though in general I also like the approach of refreshing the full list.

ui/src/app/views/rename-modal/component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  @Input() resourceType: string;
+  @Output() receiveRename = new EventEmitter<object>();
+  @Input() resource: any;
+
+  open(): void {
+    this.renaming = true;
+  }
+
+  close(): void {
```
calbach: I think the pattern we've typically used is "submit modal" -> "load+spin" -> "close modal"

ui/src/app/views/notebook-list/component.ts
```diff
@@ -158,6 +168,64 @@ export class NotebookListComponent implements OnInit, OnDestroy {
     this.openNotebook();
   }
 
+  renameThis(notebook: FileDetail): void {
+    this.notebookInFocus = notebook;
+    this.renameModal.open();
+  }
+
+  receiveRename($event: NotebookRename): void {
```
calbach: nit: There doesn't seem to be a purpose for this method, can't you just call renameNotebook directly?

ui/src/app/views/rename-modal/component.ts
calbach: OK, thanks - definitely leave as-is then.

As a follow-up we may want to address that behavior across the board, I'm not sure we want to close the dialog before the action is completed as things will shuffle around on the screen some indeterminate time after the modal has been closed. Alternatively we could have some indication that something is still loading somewhere else on the page.

api/src/main/resources/workbench.yaml
```diff
@@ -795,6 +795,76 @@ paths:
           schema:
             $ref: "#/definitions/CloneWorkspaceResponse"
 
+  # Notebooks ############################################################################
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/notebooks/renameNotebook:
```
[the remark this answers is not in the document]
UrsaStutsman: I'm going to push back a little on this. I prefer the object NotebookRename being used in the body for a couple reason. And putting both the old and new names in the path I think makes for a messy url. How bout notebooks/rename?

ui/src/app/views/notebook-list/component.html
UrsaStutsman: I'd agree, but this id is for testing. When we have control over the names being used.

ui/src/app/views/notebook-list/component.ts
UrsaStutsman: I thought of "rename" as essentially a "update" so as Bru suggested I returned a new FileDetail. I prefer this for clone.

ui/src/app/views/rename-modal/component.ts
UrsaStutsman: I used the same pattern as the ConfirmDeleteModal.... Maybe we should consider refactoring both? (Another ticket?)

ui/src/app/views/notebook-list/component.html
calbach: Since this in the real template, I'm not sure I'm comfortable chalking it up to only needing to work for test purposes. Does this not cause errors when you have a notebook with a space in the name?

Also do you actually need this for your test - can you not just grab the appropriate button via class name after opening the menu? Are the menus for all notebooks rendered but hidden for some reason?

api/src/main/resources/workbench.yaml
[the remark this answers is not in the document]
calbach: I was definitely not suggesting putting the new name in the URL, just the old name. That way all the methods that act on a single notebook follow the same pattern. I still think it's slightly inconsistent, but if you feel strongly then I'll take /notebooks/rename over the current.

ui/src/app/views/notebook-list/component.ts
calbach: Could you expand "this". Do you mean you prefer what it currently returns (nothing)? To be clear I was suggesting it return a FileDetail. In general if you have a method that creates something it returns that thing.

Anyways, it's not super important since I'm not sure it will be used but it does seem like a better/more consistent API - see CloneWorkspace, CloneCohort which return the new thing.

ui/src/app/views/notebook-list/component.html
UrsaStutsman: So no errors when you have a notebook with spaces (or other special characters that I've tested). I'll try rewriting the test without this pattern.

ui/src/app/views/notebook-list/component.ts
UrsaStutsman: It's not used at all. "this" being how I have it currently set up.

ui/src/app/views/rename-modal/component.ts
UrsaStutsman: https://precisionmedicineinitiative.atlassian.net/browse/RW-1231

ui/src/app/views/notebook-list/component.ts
UrsaStutsman: But I will change it... That said I feel like this adds a bunch of code, for not a extremely strong reason.

ui/src/app/views/notebook-list/component.html
UrsaStutsman: I've now also tried removing the pattern and grabbing the button via class name, and that does not seem to work. I'll continue playing around with it. But I'm leaning heavily towards leaving this as is.

calbach: OK, I don't feel strongly enough to push further on this.

PR #1114, opened 2018-08-21
[RW-1189] Adding concept_set, concept_set_concept_id tables to database schema. [risk=no]

danqrodney:

### DIFF
--- a/api/db/changelog/db.changelog-45-concept-set.xml
+++ b/api/db/changelog/db.changelog-45-concept-set.xml
@@ -72,3 +72,4 @@
         referencedTableName="concept_set"/>
 
   </changeSet>
+</databaseChangeLog>
--- a/api/db/changelog/db.changelog-45-concept-set.xml
+++ b/api/db/changelog/db.changelog-45-concept-set.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="danrodney" id="changelog-45">
+
+    <createTable tableName="concept_set">
+      <column name="concept_set_id" type="bigint" autoIncrement="true">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="workspace_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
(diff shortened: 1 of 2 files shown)


api/db/changelog/db.changelog-45-concept-set.xml
```diff
@@ -0,0 +1,75 @@
...
+      <column name="workspace_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
+      <column name="name" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="description" type="clob">
+        <constraints nullable="true"/>
+      </column>
+      <column name="domain" type="tinyint">
```
calbach: So a concept set is homogeneous for a single domain - that sounds reasonable to me though I don't know off-hand whether or not that's a hard requirement, perhaps worth adding a sentence with rationale in your design doc.

api/db/changelog/db.changelog-45-concept-set.xml
```diff
@@ -0,0 +1,75 @@
...
+  <changeSet author="danrodney" id="changelog-45">
+
+    <createTable tableName="concept_set">
+      <column name="concept_set_id" type="bigint" autoIncrement="true">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="workspace_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
+      <column name="name" type="varchar(80)">
```
calbach: I realize this would be a departure from other tables, but I'm curious about the decision here to use an internal ID vs using a composite primary key of workspace ID + name

danqrodney: Name is mutable, so it can't be the primary key on a concept set.

api/db/changelog/db.changelog-45-concept-set.xml
danqrodney: The idea is a single concept set can be used to retrieve data from a single OMOP table during cohort materialization, so it's important that all the concepts are from the same domain. I will update the doc.

api/db/changelog/db.changelog-45-concept-set.xml
```diff
@@ -0,0 +1,75 @@
...
+        constraintName="fk_concept_set_workspace_id"
+        onDelete="CASCADE"
+        referencedColumnNames="workspace_id"
+        referencedTableName="workspace"/>
+    <createIndex
+        indexName="idx_concept_set_workspace_id_name"
+        tableName="concept_set"
+        unique="true">
+      <column name="workspace_id"/>
+      <column name="name"/>
```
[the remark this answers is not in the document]
danqrodney: Yes.

api/db/changelog/db.changelog-45-concept-set.xml
danqrodney: (Also, the query for joining concept_set to concept_set_concept_id is going to be more efficient when using a numeric ID.)

PR #1118, opened 2018-08-22
[RISK=LOW][RW 1038] Implement API to fetch a user's recently accessed resources

NehaBroad:
This PR is for "GET" operation. It will return back the list of accessed resource for the logged in user in order of descending date .
NOTE: There will be a separate PR for POST : update/rename Notebook and DELETE notebook.


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,92 @@
...
+        response.setModifiedTime(userRecentResource.getLastAccessDate().toString());
+        if (userRecentResource.getCohortId() == null) {
+          response.setName(userRecentResource.getNotebookName());
+          response.setType("notebook");
+          response.setDescription("");
+        } else {
+          response.setType("cohort");
+        }
+        response.setWorkspaceId(userRecentResource.getWorkspaceId());
+        Workspace workspace = workspaceService.findByWorkspaceId(response.getWorkspaceId());
```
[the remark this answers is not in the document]
danqrodney: It's also potentially 10 DB calls to fetch cohort names, right? That's somewhat faster than 10 FC calls, but still slow.

Better would be to have either the cohort names, etc. denormalized into the resource entry (though this raises interesting questions about renames); or to do a fetch join in the query that retrieves the resource entries, so that the cohorts come back in the single SQL statement.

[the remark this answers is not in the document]
calbach: We link to cohorts via cohort ID, but definitely agree that this would ideally be doing a batch cohort fetch instead.

api/src/main/resources/workbench.yaml
```diff
@@ -1669,3 +1681,30 @@ definitions:
...
         description: >
           Number of clusters deleted during the check.
+
+
+  RecentResourceResponse:
+    type: array
+    items:
+      $ref: "#/definitions/RecentResource"
+
+  RecentResource:
```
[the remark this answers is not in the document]
NehaBroad: I did not use that since we need permission information for each cohort/notebook  to display the options on each card appropriately.

api/src/main/resources/workbench.yaml
```diff
@@ -872,6 +872,18 @@ paths:
           schema:
             $ref: '#/definitions/EmptyResponse'
 
+  /v1/workspaces/user-recent-resources:
+    get:
+      tags:
+        - userMetrics
+      description: Returns the resources accessed by user order by access time desc
+      operationId: getUserMetrics
```
blrubenstein: If the intent is for this to have more than just recentResources, I would call this function something different than `getUserMetrics`. Maybe `getRecentResources`?

api/src/main/resources/workbench.yaml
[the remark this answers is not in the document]
blrubenstein: Could we have two properties: object and permission?

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -190,6 +182,15 @@ public void setUserService(UserService userService) {
...
     // in workspace IDs.
     String gcsNotebooksDir = "gs://" + fcWorkspace.getBucketName() + "/notebooks";
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    long workspaceId = workspaceService
+        .getRequired(body.getWorkspaceNamespace(), body.getWorkspaceId())
+        .getWorkspaceId();
+
+    body.getNotebookNames().forEach(
+        notebook ->
+            userRecentResourceService.updateNotebookEntry(workspaceId, userProvider.get().getUserId(),gcsNotebooksDir+"/" + notebook, now)
```
calbach: Please add some spaces

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+@RestController
+public class UserMetricsController implements UserMetricsApiDelegate {
+   Provider<User> userProvider;
```
calbach: unindent

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,116 @@
...
+  UserRecentResourceService userRecentResourceService;
+  WorkspaceService workspaceService;
+  FireCloudService fireCloudService;
+  Map<Long, String> workspaceAccessMap = new HashMap<Long, String>();
+  private static final String NOTEBOOKS_WORKSPACE_DIRECTORY = "notebooks/";
```
calbach: This is already defined in ClusterController, please reference that constant or else pull that to a shared location.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+@RestController
+public class UserMetricsController implements UserMetricsApiDelegate {
+   Provider<User> userProvider;
+  UserRecentResourceService userRecentResourceService;
+  WorkspaceService workspaceService;
+  FireCloudService fireCloudService;
+  Map<Long, String> workspaceAccessMap = new HashMap<Long, String>();
```
calbach: A controller instance is shared across requests, there should never be any request-scoped variables stored at this level, i.e. it may contain data for a completely different user or be concurrently modified on access.

api/src/main/resources/workbench.yaml
```diff
@@ -1925,3 +1936,20 @@ definitions:
...
+  RecentResource:
+    type: object
+    properties:
+      permission:
+        type: string
+      cohort:
+        $ref: "#/definitions/Cohort"
+      notebook:
+        $ref: '#/definitions/FileDetail'
+      modifiedTime:
```
calbach: lastAccessedTime, this is different from last modified which we are tracking separately.

I also think we should consider making this a field on the notebooks / cohorts.

NehaBroad: Do you mean inside notebook or cohort object? lastAccesstime or modified time is user based, that can be different from the actual notebook or cohort last modified time

calbach: Good point - this is a good argument for keeping it outside. My concern was that we might need the same data for the existing notebooks/cohorts pages (and on the workspace about page). But perhaps you hvae another plan for getting that into there.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,114 @@
...
+    List<UserRecentResource> userRecentResourceList = userRecentResourceService.findAllResourcesByUser(userId);
+    List<Long> workspaceIdList = userRecentResourceList
+        .stream()
+        .map(UserRecentResource::getWorkspaceId)
+        .distinct()
+        .collect(Collectors.toList());
+
+    Map<Long, String>  workspaceAccessMap = workspaceIdList.stream().collect(Collectors.toMap(id -> id, id -> {
+      Workspace workspace = workspaceService.findByWorkspaceId((long) id);
+      WorkspaceResponse workspaceResponse = fireCloudService
```
rushtong: I don't like that we're making external API calls in this loop. I don't have a better suggestion at the moment, but there should  be a FC API call that gets this information for a batch.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,114 @@
...
+      resource.setPermission(workspaceAccessMap.get(userRecentResource.getWorkspaceId()));
+      recentResponse.add(resource);
+    });
+    return ResponseEntity.ok(recentResponse);
+  }
+
+  //Converts DB model to client Model
+  private final Function<UserRecentResource, RecentResource> TO_CLIENT =
+      userRecentResource -> {
+        RecentResource response = new RecentResource();
```
rushtong: nit naming: this should be more indicative of the type, like 'recentResource` or just `resource`

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
rushtong: After looking at FC's apis, I think the other option is to make a single call to get all workspaces from FC, and then pull out the access levels from the ones you care about. The tradeoff here is that the all-workspaces call can be a large (and slower) response, but it's just one call compared to the possibility of N calls for a single workspace. 

Side note - I don't see it, but FC should have an endpoint to get all permissions the user has across all sorts of things. Right now, they're scattered across different endpoints.

rushtong: I can be persuaded that this is an acceptable option given the current realities - the list workspaces API *is* slow and the FC UI team has also struggled with that. I think what I'd like to see at some point is some pushback on FC to make available a batch permissions endpoint that gives us what we need. I can write that story up and see about getting it prioritized.

calbach: Sounds good to me - thanks 👍

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -0,0 +1,118 @@
...
+      Workspace workspace = workspaceService.findByWorkspaceId((long) id);
+      WorkspaceResponse workspaceResponse = fireCloudService
+          .getWorkspace(workspace.getWorkspaceNamespace(),
+              workspace.getFirecloudName());
+      return workspaceResponse.getAccessLevel();
+    }));
+
+    userRecentResourceList.stream().forEach(userRecentResource -> {
+      RecentResource resource = TO_CLIENT.apply(userRecentResource);
+      resource.setPermission(workspaceAccessMap.get(userRecentResource.getWorkspaceId()));
```
calbach: This needs to handle the fact that this workspace might not be present in the map (if so, you want to skip it)

NehaBroad: Ahhh sorry to miss this.

NehaBroad: @rushtong @jscherdin will it be better to use
 workspaceIdList.stream().parallel().collect(<permission>)

than restricting to 5 distinct workspace permission?

PR #1126, opened 2018-08-23
[RW-968][risk=no] Compile integration tests in Circle

calbach:
This turned out to be pretty easy, I was just very confused about what the "build" task was doing in gradle.

For posterity, here is what I was originally trying out:

```
gradle build integration -x test
```

i.e. "build the integration target but skip the tests".

This actually just tells gradle to run the "build" task, followed by the "integration" task, while excluding the "test" task. This caused compilation issues because "integration" depends on the "test" compilation target in order to get some of it's deps, like the "common-api". By not building that, I was getting failures importing those classes into test classes.

`build` is actually just a default task installed by the Java plugin as outlined here: https://docs.gradle.org/current/userguide/java_plugin.html#_lifecycle_tasks , it is associated with the main compile and test targets.

Instead I was able to just run a task to build everything that happens before the test target, which are the compile tasks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,11 @@ jobs:
           working_directory: ~/workbench/api
           command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
       - run:
+          name: Integration tests compile
+          working_directory: ~/workbench/api
+          command: ./project.rb gradle compileBigQuery compileIntegration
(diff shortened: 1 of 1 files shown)


calbach:
Don't worry, the confusion was entirely self-induced on account of faulty assumptions and lacking gradle docs.

PR #1130, opened 2018-08-24
[RW-1100][risk=no] API call for participant charts per domain

freemabd1:
API call for participant charts per domain


api/src/main/java/org/pmiops/workbench/cohortreview/ReviewQueryBuilder.java
```diff
@@ -107,6 +107,19 @@
...
 
+  private static final String CHART_DATA_TEMPLATE =
+    "select distinct a.standard_name as standardName, a.standard_vocabulary as standardVocabulary, " +
+      "DATE(a.start_datetime) as startDate, a.age_at_event as ageAtEvent, rnk as rank\n" +
+      "from `${projectId}.${dataSetId}.%s` a\n" +
+      "left join (select standard_code, RANK() OVER(ORDER BY COUNT(*) DESC) as rnk\n" +
+      "from `${projectId}.${dataSetId}.%s`\n" +
+      "where person_id = @" + NAMED_PARTICIPANTID_PARAM + "\n" +
+      "group by standard_code\n" +
+      "LIMIT 5) b on a.standard_code = b.standard_code\n" +
```
jscherdin: Should we move the 5 out as a variable since it shows up in 2 places here and it might change over time? Or even allow the UI to specify how many in a param?

freemabd1: I can let the UI specify.. should we put a min/max on limit?

jscherdin: Probably. how about Min = 1 and Max = 20

freemabd1: done.

PR #1133, opened 2018-08-27
[RW-1231][Risk=no] Loading spinner on delete and rename modals

blrubenstein:

### DIFF
--- a/ui/src/app/views/confirm-delete-modal/component.ts
+++ b/ui/src/app/views/confirm-delete-modal/component.ts
@@ -25,7 +25,7 @@ export class ConfirmDeleteModalComponent {
   }
 
   emitDelete(resource: any): void {
-    if (this.loading === false) {
+    if (!this.loading) {
       this.loading = true;
       this.receiveDelete.emit(resource);
     }
--- a/ui/src/app/views/confirm-delete-modal/component.ts
+++ b/ui/src/app/views/confirm-delete-modal/component.ts
@@ -13,11 +13,21 @@ export class ConfirmDeleteModalComponent {
   @Output() receiveDelete = new EventEmitter<any>();
   @Input() resource: any;
 
+  loading = false;
+
   open(): void {
     this.deleting = true;
+    this.loading = false;
   }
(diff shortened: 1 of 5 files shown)


ui/src/app/views/rename-modal/component.ts
```diff
@@ -14,11 +14,21 @@ export class RenameModalComponent {
...
     this.renaming = true;
+    this.loading = false;
   }
 
   close(): void {
     this.renaming = false;
   }
+
+  emitRename(resource: any): void {
+    if (this.loading === false) {
```
calbach: `!this.loading`

ui/src/app/views/confirm-delete-modal/component.ts
```diff
@@ -13,11 +13,21 @@ export class ConfirmDeleteModalComponent {
...
     this.deleting = true;
+    this.loading = false;
   }
 
   close(): void {
     this.deleting = false;
   }
+
+  emitDelete(resource: any): void {
+    if (this.loading === false) {
```
calbach: `!this.loading`

ui/src/app/views/notebook-list/component.ts
```diff
@@ -212,13 +212,13 @@ export class NotebookListComponent implements OnInit, OnDestroy {
 
   receiveDelete($event: FileDetail): void {
     this.deleteNotebook($event);
-    this.deleteModal.close();
   }
 
   public deleteNotebook(notebook: FileDetail): void {
```
calbach: Why is this a public method? Looks like it may be a historical artifact as it doesn't seem to be referenced anywhere else. I would just inline this method into the `receiveDelete` above; as it stands it's a bit messy that `deleteNotebook` would have the logic for closing the dialog (alternatively, if this method is still needed for some reason, return the observable here and add the deleteModal callback within `receiveDelete`).

blrubenstein: Yeah, I think its a historical artifact. Will change.

PR #1134, opened 2018-08-27
[RW-1201][risk=no] Update Drugs tree and search

dolbeew:
Change drugs tree to lazy load children and use path column for search


ui/src/app/cohort-search/node-info/node-info.component.css
```diff
@@ -52,3 +52,9 @@ selection-icon-blue {
   white-space: nowrap;
   align-self: baseline;
 }
+
+/*.search-keyword {*/
+  /*color: #659F3D;*/
+  /*font-weight: bolder;*/
+  /*background-color: rgba(101,159,61,0.2);*/
+/*}*/
```
freemabd1: Can we remove this since it's commented out?

dolbeew: Done. Also removed the unused class from the html file

PR #1136, opened 2018-08-27
[RW-1237][risk=no]Bug - search page "save cohort" button not disabled when it should be

smazumd:
Bug - search page "save cohort" button not disabled when it should be


ui/src/app/cohort-search/overview/overview.component.html
```diff
@@ -9,11 +9,12 @@ <h2 id="total-count">
         {{ total$ | async | number }}
       </span>
     </h2>
-    <div>
+    <div *ngIf="chartData$ | async as chartData">
       <button
         type="button"
         class="btn btn-primary"
         [style.margin-left]="'0'"
+        [disabled]="!chartData.size || (total$ | async | number) === 0"
```
dolbeew: Do we need the number pipe here?

smazumd: Its working without number pipe. so I will remove number pipe.

PR #1142, opened 2018-08-27
[RW-1142][Risk=no] Make cohortEdit conform with rest of app

blrubenstein:


ui/src/app/views/cohort-list/component.spec.ts
```diff
@@ -15,6 +15,7 @@ import {
   updateAndTick
 } from 'testing/test-helpers';
 
+import {CohortEditModalComponent} from '../cohort-edit-modal/component';
```
UrsaStutsman: not sure - so guess this is more a question than comment... But do we want relative paths? Or absolute ones? I've always preferred absolute ones, but figure we should be consistent.

blrubenstein: Agreed. Absolute ones seem better. Will swap!

blrubenstein:
Will add test!

PR #1143, opened 2018-08-27
Api fix and concept_synonym index addition [risk=no]

SrushtiGangireddy:
1. Added unit to the measurement analysis
2. Missed adding back the indexes on concept synonym table.


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -504,19 +508,34 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+            }
 
-            for(AchillesAnalysis aa:analysisList){
-                    if(aa.getAnalysisId() == GENDER_ANALYSIS_ID){
+            conceptAnalysis.setConceptId(conceptId);
+            Iterator it = analysisHashMap.entrySet().iterator();
+            while(it.hasNext()) {
+                    Map.Entry pair = (Map.Entry)it.next();
+                    Long analysisId = (Long)pair.getKey();
+                    AchillesAnalysis aa = (AchillesAnalysis)pair.getValue();
```
biopete: Just set unitName on every analysis right here. It doesn't matter if it is null. And make your copy constructor use it.

SrushtiGangireddy: Done.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -521,12 +521,15 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
                 }
             }
 
+            System.out.println(unitName);
```
biopete: Remove this.

api/hs_err_pid68.log
```diff
@@ -0,0 +1,906 @@
+#
```
danqrodney: I don't think you meant to add this file?

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
```diff
@@ -103,6 +115,14 @@ public ConceptsController(ConceptService conceptService, WorkspaceService worksp
               request.getVocabularyIds(), domainIds, maxResults, minCount);
 
 
+    List<Long> conceptIds =concepts.getContent().stream().map(org.pmiops.workbench.cdr.model.Concept::getConceptId).collect(Collectors.toList());
+    Multimap<Long,ConceptSynonym> synonymMap = Multimaps.index(conceptSynonymDao.findByConceptIdIn(conceptIds),ConceptSynonym::getConceptId);
```
danqrodney: Given that we do this logic in both places where we use ConceptService.searchConcepts, we should probably just put this logic inside that method?

public-api/src/main/resources/public-api.yaml
```diff
@@ -353,6 +353,26 @@ definitions:
             $ref: "#/definitions/Concept"
           description: standard concepts associated with the matched concept
 
+  ConceptSynonym:
```
danqrodney: Not seeing where this gets used?

Do we want full objects representing the synonyms on the client, or just an array of strings like above?

api/hs_err_pid68.log
SrushtiGangireddy: Yup for sure. Removed it.

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
SrushtiGangireddy: Yeah, put it in the concept service.

public-api/src/main/resources/public-api.yaml
SrushtiGangireddy: We just need the array of strings with the concept. This is just the intermediary object needed for the tests.

danqrodney:
I think we're going to need to do something similar regarding synonym fetching when fetching concepts in a concept set... this is what's causing the new test failures. Let me know if I can help with that.

Otherwise, looks great!

SrushtiGangireddy:
I added a multi map just like what we do in concept service after find all from concept dao and that fixed it. Thanks Dan.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -3,8 +3,8 @@
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Ordering;
+import com.google.common.collect.*;
```
danqrodney: Nit: We don't usually use wildcard imports (as it makes it hard to tell where things are coming from.)

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -130,9 +134,15 @@ public void setUserProvider(Provider<User> userProvider) {
   private ConceptSet toClientConceptSet(org.pmiops.workbench.db.model.ConceptSet conceptSet) {
     ConceptSet result = TO_CLIENT_CONCEPT_SET.apply(conceptSet);
     if (!conceptSet.getConceptIds().isEmpty()) {
-      result.setConcepts(ImmutableList.copyOf(
-          conceptDao.findAll(conceptSet.getConceptIds()))
-              .stream()
+      Iterable<org.pmiops.workbench.cdr.model.Concept> concepts = conceptDao.findAll(conceptSet.getConceptIds());
```
danqrodney: We could (if we wanted to centralize the multimap logic) add a findByIds() method to ConceptService that does this, too.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -128,26 +128,28 @@
...
-      .standardConcept(true)
-      .conceptCode("conceptA")
-      .conceptClassId("classId")
-      .vocabularyId("V1")
-      .domainId("Condition")
-      .countValue(123L)
-      .prevalence(0.2F);
+          .conceptId(123L)
+          .conceptName("a concept")
+          .standardConcept(true)
```
biopete: Curious why this isn't an "S" or "C" or null ?

SrushtiGangireddy: True is converted to S and false to C and null is left as it is in the make concept below.

PR #1148, opened 2018-08-28
[RW-1195][Risk=no] Integrate notebook actions with Users Recently Used Cache

UrsaStutsman:
Added in the UserRecentResource calls, added tests to make sure they are called correctly, did a little refactoring in Controller.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -22,6 +22,7 @@ import org.pmiops.workbench.annotations.AuthorityRequired;
 import org.pmiops.workbench.db.dao.CdrVersionDao;
 import org.pmiops.workbench.db.dao.CohortDao;
 import org.pmiops.workbench.db.dao.UserDao;
+import org.pmiops.workbench.db.dao.UserRecentResourceService;
 import org.pmiops.workbench.db.dao.UserService;
 import org.pmiops.workbench.db.dao.WorkspaceService;
 import org.pmiops.workbench.db.model.CdrVersion;
@@ -80,6 +81,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
   private final CloudStorageService cloudStorageService;
   private final Clock clock;
   private final UserService userService;
+  private final UserRecentResourceService userRecentResourceService;
 
   @Autowired
   WorkspacesController(
@@ -91,7 +93,8 @@ public class WorkspacesController implements WorkspacesApiDelegate {
       FireCloudService fireCloudService,
       CloudStorageService cloudStorageService,
       Clock clock,
-      UserService userService) {
+      UserService userService,
+      UserRecentResourceService userRecentResourceService) {
     this.workspaceService = workspaceService;
     this.cdrVersionDao = cdrVersionDao;
     this.cohortDao = cohortDao;
@@ -101,6 +104,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
     this.cloudStorageService = cloudStorageService;
     this.clock = clock;
     this.userService = userService;
+    this.userRecentResourceService = userRecentResourceService;
   }
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -744,54 +748,60 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
     return ResponseEntity.ok(fileDetail);
   }
 
   @Override
   public ResponseEntity<EmptyResponse> deleteNotebook(String workspace, String workspaceName, String notebookName) {
+    performNotebookOperation("delete", workspace, workspaceName, notebookName, "");
+    return ResponseEntity.ok(new EmptyResponse());
+  }
+
+  private FileDetail performNotebookOperation(String operation, String workspace, String workspaceName, String notebookName, String newName) {
```
blrubenstein: I like this a lot. Good refactor. (Always happy to see extra code go away)

An alternative way to do this could also be a 'notebookCloneOperation' and a 'notebookDeleteOperation'? Then rename would call Clone and then Delete, clone would just call Clone, and delete just delete?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1390,12 +1393,15 @@ public void testRenameNotebookinWorkspace() throws Exception {
...
     String newPath = "notebooks/nb2.ipynb";
+    String fullPath = "gs://workspace-bucket/" + newPath;
     NotebookRename rename = new NotebookRename();
     rename.setName("nb1.ipynb");
     rename.setNewName(newName);
     workspacesController.renameNotebook(workspace.getNamespace(), workspace.getId(), rename);
     verify(cloudStorageService).copyBlob(BlobId.of(BUCKET_NAME, nb1), BlobId.of(BUCKET_NAME, newPath));
     verify(cloudStorageService).deleteBlob(BlobId.of(BUCKET_NAME, nb1));
+    verify(userRecentResourceService).updateNotebookEntry(1, 1, fullPath, Timestamp.from(NOW));
+    verify(userRecentResourceService).deleteNotebookEntry(1, 1, fullPath);
```
blrubenstein: I like these tests a lot. 👍

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
UrsaStutsman: Ooooh hm, I actually like that even better. Yup! Ok going to do that!

UrsaStutsman: Hmmm, or maybe variation of that... a lot of the setup for both operations is the same....

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -744,54 +750,69 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-    BlobId blobId = BlobId.of(bucket, origPath);
-    cloudStorageService.copyBlob(blobId, BlobId.of(bucket, newPath));
-    cloudStorageService.deleteBlob(blobId);
-    FileDetail fileDetail = new FileDetail();
-    fileDetail.setName(newName);
-    fileDetail.setPath("gs://" + bucket + "/" + NOTEBOOKS_WORKSPACE_DIRECTORY + "/" + newName);
-    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
-    fileDetail.setLastModifiedTime(now.getTime());
+    FileDetail fileDetail = notebookCloneOperation(workspace, workspaceName, rename.getName(), newName);
+    notebookDeleteOperation(workspace, workspaceName, rename.getName());
```
rushtong: This is a lot cleaner!

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -744,54 +750,69 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    Map<String, Object> map = notebookOpSetup(workspace, workspaceName, notebookName, "");
+    cloudStorageService.deleteBlob((BlobId) map.get("blobId"));
+    userRecentResourceService.deleteNotebookEntry(objToLong(map.get("workspaceId")), objToLong(map.get("userId")), map.get("fullPath").toString());
+  }
+
+  private long objToLong(Object object) {
+    return Long.parseLong(String.valueOf(object));
+  }
+
+  private Map<String, Object> notebookOpSetup(String workspace, String workspaceName, String notebookName, String newName) {
```
rushtong: I can see what you're trying to do here with `Map<String, Object>` but that leads to extra code, ie. `objToLong` that you then need to de-convolute what's in the map. In my opinion, it would be cleaner to just make a new DTO class that holds the variables you want with the types that you want so you don't have to munge-unmunge data.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1390,12 +1393,16 @@ public void testRenameNotebookinWorkspace() throws Exception {
...
+    String fullPath = "gs://workspace-bucket/" + newPath;
+    String origFullPath = "gs://workspace-bucket/" + nb1;
     NotebookRename rename = new NotebookRename();
     rename.setName("nb1.ipynb");
     rename.setNewName(newName);
     workspacesController.renameNotebook(workspace.getNamespace(), workspace.getId(), rename);
     verify(cloudStorageService).copyBlob(BlobId.of(BUCKET_NAME, nb1), BlobId.of(BUCKET_NAME, newPath));
     verify(cloudStorageService).deleteBlob(BlobId.of(BUCKET_NAME, nb1));
+    verify(userRecentResourceService).updateNotebookEntry(1, 1, fullPath, Timestamp.from(NOW));
+    verify(userRecentResourceService).deleteNotebookEntry(1, 1, origFullPath);
```
rushtong: I just made this same comment in one of Neha's PRs. Can we use variables instead of literals in the tests (here and below)? I think it's easier to understand the test with names. Looking at these methods quickly, I'll have to do some hunting to understand the `1` values here to see if the test makes sense.

PR #1149, opened 2018-08-29
[RISK=LOW][RW-1038]delete rename notebook api

NehaBroad:
Endpoints enabling deleting and renaming notebook entries from recent resource.


api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,54 @@ paths:
...
+          name: notebook
+          required: true
+          type: string
+      responses:
+        200:
+          description: Successfully deleted
+          schema:
+            $ref: "#/definitions/EmptyResponse"
+
+  /v1/recentResource/{workspaceId}/{workspaceName}/rename-notebook:
```
calbach: This feels to me like it might be overly tuned to the use case. I would recommend just providing a delete and update method, I think the Jupyter server extension can make do with just that.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -52,26 +54,30 @@
   private UserMetricsController userMetricsController;
   private static final Instant NOW = Instant.now();
   private FakeClock clock = new FakeClock(NOW);
+  private long workspace_1_Id = 1l;
```
calbach: Please don't use underscores in java variable names (exception: constants in all caps)

NehaBroad:
i messed something up while merging fixing it now

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,54 @@ paths:
...
+
+  /v1/recentResource/{workspaceId}/{workspaceName}/{notebook}/delete-notebook:
+    parameters:
+      - $ref: '#/parameters/workspaceNamespace'
+      - $ref: '#/parameters/workspaceId'
+    delete:
+      tags:
+        - userMetrics
+      description: Deletes Notebook entry from recent resource
+      operationId: deleteNotebookEntry
```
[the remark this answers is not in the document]
NehaBroad: I missed this comment. will work on it now

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,48 @@ paths:
           description: A list of resources order by last access time desc order
           schema:
             $ref: "#/definitions/RecentResourceResponse"
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/recent-resource/{notebook}/update-notebook:
```
calbach: Not sure if you already addressed my other comment that you said you'd missed, but per that comment I would change this to just "update", and put the notebookname into the request

calbach: I think I would go with user-recent-resources to match the other URLs

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,48 @@ paths:
...
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/user-recent-resource/updateEntry:
+    parameters:
+      - $ref: '#/parameters/workspaceNamespace'
+      - $ref: '#/parameters/workspaceId'
+    post:
+      tags:
+        - userMetrics
+      description: Add/update Notebook entry to recent resource
+      operationId: updateNotebookEntry
```
calbach: updateRecentResource

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,48 @@ paths:
           description: A list of resources order by last access time desc order
           schema:
             $ref: "#/definitions/RecentResourceResponse"
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/user-recent-resource/updateEntry:
```
calbach: I would add an s: user-recent-resources here and below to match the above GET API

calbach: I don't think you need the "entry" part, but if you decide to leave it, should be hyphenated - otherwise you're mixing camelCase and hyphenation in the same URL

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,48 @@ paths:
...
+      - $ref: '#/parameters/workspaceNamespace'
+      - $ref: '#/parameters/workspaceId'
+    post:
+      tags:
+        - userMetrics
+      description: Add/update Notebook entry to recent resource
+      operationId: updateNotebookEntry
+      parameters:
+        - in: body
+          name: notebook
```
calbach: I'm not sure what this looks like in JSON. To match other methods I would probably recommend putting a UpdateRecentResourceRequest object here. I recall @UrsaStutsman  hitting some weird swagger codegen bugs when doing something similar to what you have here.

api/src/main/resources/workbench.yaml
```diff
@@ -960,6 +960,48 @@ paths:
...
+          schema:
+            type: string
+      responses:
+        200:
+          description: Successfully added
+          schema:
+            $ref: "#/definitions/RecentResource"
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/user-recent-resource/deleteEntry:
```
calbach: same here (delete or delete-entry)

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -52,26 +54,32 @@
   private UserMetricsController userMetricsController;
   private static final Instant NOW = Instant.now();
   private FakeClock clock = new FakeClock(NOW);
+  private static final long WORKSPACE_1_ID = 1l;
+  private static final long WORKSPACE_2_ID = 2l;
+  private static final long COHORT_ID = 1l;
+  private static final long USER_ID = 123l;
+  private final String WORKSPACE_NAMESPACE = "workspaceNamespace";
```
calbach: static?

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -158,5 +170,36 @@ public void testWorkspaceLimit() {
...
+  public void testUpdateNotebookEntry() {
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+
+    UserRecentResource mockUserRecentResource = new UserRecentResource();
+    mockUserRecentResource.setCohort(null);
+    mockUserRecentResource.setWorkspaceId(WORKSPACE_2_ID);
+    mockUserRecentResource.setUserId(USER_ID);
+    mockUserRecentResource.setNotebookName("gs://newBucket/notebooks/notebook.ipynb");
+    mockUserRecentResource.setLastAccessDate(now);
+    when(userRecentResourceService.updateNotebookEntry(WORKSPACE_2_ID, USER_ID,"gs://newBucket/notebooks/notebook.ipynb", now ))
```
calbach: spacing

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -158,5 +170,36 @@ public void testWorkspaceLimit() {
...
+    .thenReturn(mockUserRecentResource);
```
calbach: indent

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -158,5 +170,36 @@ public void testWorkspaceLimit() {
...
+    RecentResource addedEntry = userMetricsController
+        .updateNotebookEntry(WORKSPACE_NAMESPACE, FIRECLOUD_WORKSPACE_ID, "gs://newBucket/notebooks/notebook.ipynb")
+        .getBody();
+    assertNotNull(addedEntry);
+    assertEquals((long)addedEntry.getWorkspaceId(), WORKSPACE_2_ID);
```
calbach: space after the case

PR #1155, opened 2018-08-29
Biopete/ppi measurements [risk=low]

biopete:
This adds:

- initial Physical Measurements Page to the databrowser. It is pending further design.
- Concept synonyms to search and results
- Histogram charts for measurements
- Adds DBConstantsService for constants used throughout business logic such as concept id, analysis id, color constants.


public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -478,6 +483,32 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
             }
 
             List<AchillesAnalysis> analyses = achillesAnalysisDao.findSurveyAnalysisResults(surveyConceptId, qlist);
+            for(AchillesAnalysis analysis: analyses){
```
calbach: nit: please use consistent spacing throughout this file, specifically most of the codebase uses
```
for (Foo f : flist) {
...
Foo f = 1;
if (true) {
...
```

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -478,6 +483,32 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
...
+                if(analysis.getAnalysisId() == SURVEY_AGE_ANALYSIS_ID){
+                    Set<String> uniqueAgeDeciles = new TreeSet<String>();
+                    for(AchillesResult ar: analysis.getResults()){
+                        String analysisStratumName=ar.getAnalysisStratumName();
+                        uniqueAgeDeciles.add(ar.getStratum2());
+                        if (analysisStratumName == null || analysisStratumName.equals("")) {
+                            ar.setAnalysisStratumName(QuestionConcept.ageStratumNameMap.get(ar.getStratum2()));
+                        }
+                    }
+                    if(uniqueAgeDeciles.size() < 7){
```
calbach: This entire codeblock appears to be duplicated at 559

public-ui/src/app/data-browser/chart/chart.component.ts
```diff
@@ -47,208 +13,201 @@ const DATA_LABEL_STYLE = {
...
-      return {
-          chart: options.chart,
-          credits: {
+    const options = this.makeChartOptions();
+    // Override title if they passed one
+    if (this.chartTitle) {
+      options.title.text = this.chartTitle;
+    }
+    // Override chart type if we have it
+    // todo  -- check we don't need this. too hacky
```
calbach: remove?

public-ui/src/app/views/ehr-view/ehr-view.component.html
```diff
@@ -127,19 +127,21 @@ <h5 class="col-xs-12 box-title">
...
 
         </div>
         <div class="row expandable-grid-row"
              *ngFor="let r of items.concat(standardConcepts), let i = index"
              [ngClass]="{'expanded-grid-row': r.expanded}" >
-          <div class="col-md-4 flex-xs-middle" innerHtml="{{i+1}}. {{r.conceptName | highlightSearch : prevSearchText}}"> </div>
-          <div class="col-md-4 flex-xs-middle">
-            {{r.countValue | number }} out of {{totalParticipants | number}} participants
-
+          <div class="col-md-2 flex-xs-middle" innerHtml="{{i+1}}. {{r.conceptName | highlightSearch : prevSearchText}}
```
calbach: Do you really need innerHtml? It should generally be avoided on account of risk of injection.

public-ui/src/app/views/pm/pm.component.css
```diff
@@ -0,0 +1,16 @@
...
+    width: 336px;
+}
+
+/* Style the active sub buttons */
+button.active.btn-link {
+    font-weight: bolder;
+}
+
+.row.page-align {
+    margin-left: 0px;
```
calbach: nit: no need to specify units for 0s in CSS

public-ui/src/app/views/pm/pm.component.html
```diff
@@ -1,7 +1,72 @@
...
+        <div *ngIf="selectedConcept.analyses && selectedConcept.analyses.ageAnalysis" class="col-md-4 col-xs-12 chart-wrapper" >
+          <app-chart [analysis]="selectedConcept.analyses.ageAnalysis" [pointWidth]="20" [backgroundColor]="backgroundColor"> </app-chart>
+        </div>
+
+        <!-- Todo wait on design for further dev -->
+
+      </div>
+    </div>
+
+    <!-- Don't want to deploy this link
```
calbach: remove? This doesn't seem like it should be in the code

public-ui/src/app/views/pm/pm.component.ts
```diff
@@ -1,15 +1,228 @@
...
+        for (const g of this.genderAnalysis.results) {
+          if (g.stratum1 === this.dbc.FEMALE_GENDER_ID) {
+            this.femaleCount = g.countValue;
+          } else if (g.stratum1 === this.dbc.MALE_GENDER_ID) {
+            this.maleCount = g.countValue;
+          } else {
+            this.otherCount += g.countValue;
+          }
+        }
+        this.loadingStack.pop();
```
calbach: This is a nice pattern - should you be popping the stack if the request fails as well?

public-ui/src/app/views/pm/pm.component.ts
```diff
@@ -1,15 +1,228 @@
...
+      // Add not pregnant result to the female value results
+      // because this concept is just a one value Yes
+      const pregnantResult  = concept.analyses.measurementValueFemaleAnalysis.results[0];
+
+      const notPregnantResult: AchillesResult = {
+        analysisId: pregnantResult.analysisId,
+        stratum1: pregnantResult.stratum1,
+        stratum2: pregnantResult.stratum2,
+        stratum3: pregnantResult.stratum3,
+        stratum4: 'Not Pregnant',
```
calbach: WHy is this one hardcoded? Comment?

public-ui/src/app/data-browser/chart/chart.component.ts
```diff
@@ -47,208 +13,201 @@ const DATA_LABEL_STYLE = {
...
+        series: {
+          animation: {
+            duration: 100,
           },
-          zAxis: {
+          pointWidth: options.pointWidth ? options.pointWidth : null,
+          minPointLength: 3,
+          events: {
+            click: function (event) {
+              console.log('plot options clicked ', event.point);
```
calbach: These kinds of debugging statements should probably not be checked in

public-ui/src/app/utils/db-constants.service.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class DbConstantsService {
```
calbach: Does this need to be a service, or could you just be putting these values into some static utility package and referencing them directly? The only reason I can think you might want this to be a service would be if you expect you'll have a strong need to stub out these constant values for testing. If so, that's fine

biopete: I was thinking more like if we wanted them to be dynamic constants from a database or change if this is used in workbench etc. And we may add more variables that are dynamic things from the database.

public-ui/src/app/data-browser/chart/chart.component.ts
biopete: Yeah. I am just going to have to add them back when we get into drilling down by clicking plots. That is why i left them. More like a todo. We will want to capture analytics too when people click a plot.

public-ui/src/app/views/pm/pm.component.ts
biopete: I could use a constant. We don't get any result for not pregnant. It is an analysis with one result with the count of people that are pregnant. But that doesn't make for a good graph. So i make a not pregnant result here. This needs a little more design about how to graph concepts like "pregnant" and "wheel chair use" that just give us the count for those things.  I just threw this in there for a option for now.

public-ui/src/app/views/pm/pm.component.html
biopete: I can remove this. It is just handy and this screen will be redesigned.

public-ui/src/app/views/pm/pm.component.ts
biopete: Good catch. Yes. I will do that.

public-ui/src/app/views/pm/pm.component.css
biopete: Yeah.

public-ui/src/app/views/ehr-view/ehr-view.component.html
biopete: Yes. It's been a while since i wrote this. For any pipe to change the markup and add the <mark/> tag as this does  it needs innerhtml.  Angular sanitizes the return from pipes and just displays the markup as text in the variable.   I never found a way around it.

public-ui/src/app/data-browser/chart/chart.component.ts
biopete: Yes.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
biopete: oops. This file shouldn't even be in the pr. I made no intentional changes here.  I put in the version from master.

biopete:
FYI I still can't figure out the codacy thing. Now it is going to Neha's codacy again.  I'll fix your suggestions and let you know.

calbach:
As mentioned in slack - it's going to be non-deterministic on each push until we delete all of the shadow Codacy projects. Looks like Neha has deleted hers so perhaps we're in the clear now.

biopete:
Great. It looks like it is going to workbench codacy now. So i have removed any api changes , linted. Hopefully everything passes.

public-ui/src/app/views/ehr-view/ehr-view.component.html
```diff
@@ -102,9 +58,6 @@ <h5 class="col-xs-12 box-title">
 
     <div class="col-md-8 col-sm-6 col-xs-12 chart-wrapper flex-xs-middle">
       <app-chart [concepts]="top10Results" [chartTitle]="'Top 10 by particpant count'" [pointWidth]="15" [backgroundColor]="'#D9E4EA'"> </app-chart>
```
SrushtiGangireddy: Participant?

public-ui/src/app/views/ehr-view/ehr-view.component.html
```diff
@@ -127,19 +80,24 @@ <h5 class="col-xs-12 box-title">
           </div>
         </div>
         <div class="row">
-          <div class="col-md-4 grid-header flex-xs-middle"> {{dbDomain.domainDisplay}}</div>
-          <div class="col-md-4 grid-header flex-xs-middle"> Number of particpants </div>
-          <div class="col-md-3 grid-header flex-xs-middle info-text"> <clr-icon class="" shape="info-standard" class="is-solid"></clr-icon> What is SNOMED? </div>
+          <div class="col-md-2 grid-header flex-xs-middle"> {{dbDomain.domainDisplay}}</div>
+          <div class="col-md-4 grid-header flex-xs-middle"> Synonyms </div>
+          <div class="col-md-2 grid-header flex-xs-middle"> Particpants of <b> {{totalParticipants | number}}</b> </div>
```
SrushtiGangireddy: Same.

public-ui/src/app/views/ehr-view/ehr-view.component.html
```diff
@@ -127,19 +80,24 @@ <h5 class="col-xs-12 box-title">
...
+          <div class="col-md-3 grid-header flex-xs-middle info-text">
```
SrushtiGangireddy: Some procedure in concepts seem to have lot of synonyms in there and it looks weird. 
May be we could make vocabulary code 2 ? 
Just a thought. Did not try how that looks for rest of the concepts.
Do not bother if you have tried that out already.

public-ui/src/app/views/ehr-view/ehr-view.component.html
biopete: Good catch. Thanks.

public-ui/src/app/views/ehr-view/ehr-view.component.html
biopete: Thanks.

public-ui/src/app/views/ehr-view/ehr-view.component.html
biopete: Good idea. I did that and made name and synonym take up rest of width. We may add something to hide huge lists of synonyms behind button in future.

public-ui/src/app/views/pm/pm.component.ts
```diff
@@ -1,15 +1,251 @@
...
+  ethnicityAnalysis: Analysis = null;
+
+  // Chart titles for our gender histograms.
+  maleGenderChartTitle =  '';
+  femaleGenderChartTitle = '';
+  otherGenderChartTitle = '';
+
+  conceptGroups = [
+    { group: 'blood-pressure', groupName: 'Mean Blood Pressure', concepts: [
+      {conceptId: '903118', conceptName: 'Systolic ', analyses: null, chartType: this.chartType,
```
calbach: There's a lot of repeated code here, I would probably make a helper or possibly a base concept object:

```
baseConcept = {
  chartType: this.chartType,
  maleCount: 0
}

{
  ...baseConcept,
  conceptId: '',
  conceptName: '',
}
```

public-ui/src/app/utils/db-constants.service.ts
calbach: We have a similar concept in workbench which we call  the "server config". Optional: could call this DbConfigService to match that.

PR #1157, opened 2018-08-29
[RW-1219][risk=low] API for user autocomplete

rushtong:
## Addresses 
https://precisionmedicineinitiative.atlassian.net/browse/RW-1219

## Changes
This PR adds a new controller for performing autocomplete on the user table. The new thing we're doing here is using spring-data's [`Pageable`](https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#repositories.core-concepts) interface so I'd like some opinions on that approach. 

**Note:** I chose not to allow for the API to specify the sorting field. This is intentional as I think it's more than what's required for the story, but I'm open to feedback on that. It would require some (minor) additional complexity to the rest interface. 

This PR does not complete the story. It needs a UI component as well.


api/src/main/resources/workbench.yaml
```diff
@@ -425,6 +425,41 @@ paths:
...
+      description: Returns user information matching search terms
+      operationId: user
+      responses:
+        200:
+          description: The user information matching supplied search terms
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/UserResponse"
+      parameters:
```
calbach: We have an existing paginated API defined here: https://github.com/all-of-us/workbench/blob/master/api/src/main/resources/client_api.yaml#L127 which follows the Google Cloud API style guide for pagination: https://cloud.google.com/apis/design/design_patterns#list_pagination

My preference would be to follow that pattern here - happy to talk at length about the merits of that approach. I'm also OK using an offset as the implementation for now to keep similarity with what you've already done (though I do think that approach is not ideal).

api/src/main/resources/workbench.yaml
```diff
@@ -425,6 +425,41 @@ paths:
...
+        - in: path
+          name: term
+          description: String to find in user's name or email address
```
calbach: case sensitive? substring match?

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+    Sort.Direction direction = Optional
+        .ofNullable(Sort.Direction.fromStringOrNull(sort))
+        .orElse(Sort.Direction.ASC);
+    Pageable pageable = new PageRequest(
+        Optional.ofNullable(page).orElse(DEFAULT_PAGE),
+        Optional.ofNullable(size).orElse(DEFAULT_SIZE),
+        new Sort(direction, DEFAULT_SORT_FIELD));
+    List<UserResponse> responses = userDao
+        .findUsersBySearchString(term, pageable)
```
calbach: I'm actually thinking we should restrict this search to ID-verified users (possibly relaxing this constraint in dev - maybe via dataAccessLevel as a function of workbenchConfig.enforceRegistered?). In the event that spammers create accounts with text ads or expletives in the first/last name, at least things are a bit more thoroughly vetted before we push this into other people's autocompletes.

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -36,4 +37,11 @@
    */
   @Query("Select user FROM User user LEFT JOIN FETCH user.authorities LEFT JOIN FETCH user.pageVisits WHERE user.userId = :id")
   User findUserWithAuthoritiesAndPageVisits(@Param("id") long id);
+
+  /**
+   * Find users matching the user's name or email
+   */
+  @Query("Select user FROM User user WHERE lower(user.email) LIKE lower(concat('%', :term, '%')) OR lower(user.familyName) LIKE lower(concat('%', :term, '%')) OR lower(user.givenName) LIKE lower(concat('%', :term, '%'))")
```
calbach: nit: SELECT to match the other keyword spacing?

api/src/main/resources/workbench.yaml
rushtong: Thank you - will look at that.

api/src/main/java/org/pmiops/workbench/api/UserController.java
rushtong: I like it - will do.

api/src/main/resources/workbench.yaml
rushtong: Done

api/src/main/java/org/pmiops/workbench/api/UserController.java
rushtong: Done

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return ResponseEntity.ok(response);
+  }
+
+  private int getPageToken(String pageToken) {
+    if (pageToken == null) {
+      return DEFAULT_PAGE;
+    }
+    try {
+      return Integer.valueOf(pageToken);
```
calbach: Alternatively, consider using this helper: https://github.com/all-of-us/workbench/blob/2831ab87b423c146a107b2e5b4bea2c3c6fb5b7f/api/src/main/java/org/pmiops/workbench/utils/PaginationToken.java

Given the use case, it's very unlikely to matter, but my general preference for using a b64 encoded token is that it forces clients to use treat it as an opaque token, rather than depending on the details of your implementation (e.g. assuming a number encoded as a string works). This matters if you want to change the pagination impl without breaking clients.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -242,4 +248,23 @@ private void logAdminAction(Long targetUserId, Long targetWorkspaceId, String ta
...
   }
+
+  /**
+   * Find users matching the user's name or email
+   */
+  public List<User> findUsersBySearchString(String term, Sort sort) {
+    List<Short> dataAccessLevels;
+    if (configProvider.get().firecloud.enforceRegistered) {
+      dataAccessLevels = Collections
+          .singletonList(StorageEnums.dataAccessLevelToStorage(DataAccessLevel.REGISTERED));
```
calbach: CONTROLLED also should be accepted, basically >= registered

api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
```diff
@@ -0,0 +1,207 @@
...
+    // Now reverse one and assert both in the same order
+    List<org.pmiops.workbench.model.User> descendingReversed = Lists.reverse(robinsonsDesc.getUsers());
+    assertThat(robinsonsAsc.getUsers()).containsAllIn(descendingReversed).inOrder();
+
+    // Test that JPA sorting is really what we expected it to be by re-sorting one into a new list
+    List<org.pmiops.workbench.model.User> newAscending = Lists.newArrayList(robinsonsAsc.getUsers());
+    newAscending.sort(Comparator.comparing(org.pmiops.workbench.model.User::getEmail));
+    assertThat(robinsonsAsc.getUsers()).containsAllIn(newAscending).inOrder();
+  }
+
```
calbach: Consider adding some test cases to cover invalid page tokens, including well-formed, but negative offset values.

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -0,0 +1,91 @@
...
+    } catch (NumberFormatException nfe) {
+      log.info(String.format("Unable to parse invalid page token: %s", pageToken));
```
calbach: This should result in a BadRequestException in my opinion. Also would be good to capture negative offsets here.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
rushtong: The four levels we have now are: 
```
  UNREGISTERED("unregistered"),
  
  REGISTERED("registered"),
  
  PROTECTED("protected"),
  
  REVOKED("revoked");
```
I think you mean protected?

calbach: Yes, sorry

api/src/main/java/org/pmiops/workbench/api/UserController.java
rushtong: Took me a bit to get my head around this ... but I like it. Done.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
rushtong: Done.

api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
rushtong: Done.

api/src/main/java/org/pmiops/workbench/api/UserController.java
rushtong: Done.

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -0,0 +1,92 @@
...
+
+    if (null == term || term.isEmpty()) {
+      return ResponseEntity.ok(response);
+    }
+
+    PaginationToken paginationToken;
+    try {
+      paginationToken = getPaginationTokenFromPageToken(pageToken);
+    } catch (BadRequestException e) {
+      return ResponseEntity.badRequest().body(response);
```
calbach: I believe this should get converted automatically

api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
```diff
@@ -0,0 +1,220 @@
...
+  @Test
+  public void testUserNoUsersResponse() {
+    UserResponse response = userController.user("Smith", null, null, null).getBody();
+    assertThat(response.getUsers()).hasSize(0);
+  }
+
+  @Test(expected=IllegalArgumentException.class)
+  public void testInvalidPageTokenCharacters() {
+    userController.user("Robinson", "Inv@l!dT0k3n#", null, null);
```
rushtong: This should really be another bad request ... not an outright exception.

PR #1165, opened 2018-08-31
Attempt to change api tests to run in circle only if master or api ch…

blrubenstein:
…anges

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -372,18 +372,18 @@ Common.register_command({
 
 def test_api_changes(branch_name)
   common = Common.new
-  common_commit = common.capture_stdout(%W{git merge-base master blrubenstein/test-circle}).split("\n")[0]
+  common_commit = common.capture_stdout(%W{git merge-base master #{branch_name}}).split("\n")[0]
   api_lines = common.capture_stdout(%W{git diff --name-only #{common_commit}}).split("\n")
-  return api_lines.any? { |s| s.include?('api') }
+  api_lines.any? { |s| s.include?('api') }
 end
 
 def run_all_tests(cmd_name, args)
-  common = Common.new
   circle_branch = ENV["CIRCLE_BRANCH"]
   if circle_branch.nil? || circle_branch.empty? || circle_branch == "master"
     run_tests = true
   else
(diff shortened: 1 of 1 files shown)


api/libproject/devstart.rb
```diff
@@ -370,10 +370,26 @@ def run_public_api_tests(cmd_name, args)
   :fn => ->(*args) { run_public_api_tests("test-public-api", args) }
 })
 
+def test_api_changes(branch_name)
+  common = Common.new
+  common_commit = common.capture_stdout(%W{git merge-base master blrubenstein/test-circle}).split("\n")[0]
```
calbach: Use branch_name variable

api/libproject/devstart.rb
```diff
@@ -370,10 +370,26 @@ def run_public_api_tests(cmd_name, args)
...
+  api_lines = common.capture_stdout(%W{git diff --name-only #{common_commit}}).split("\n")
+  return api_lines.any? { |s| s.include?('api') }
```
calbach: This substring check feels like it should be tightened up to match a directory specifically, e.g. via startswith or including "/"s. Note that specifically in this case we care about "api" and "common-api".

api/libproject/devstart.rb
```diff
@@ -370,10 +370,26 @@ def run_public_api_tests(cmd_name, args)
...
+end
 
 def run_all_tests(cmd_name, args)
-  run_api_tests(cmd_name, args)
-  run_public_api_tests(cmd_name, args)
+  common = Common.new
+  circle_branch = ENV["CIRCLE_BRANCH"]
```
calbach: I still feel it would be cleaner to move this logic into a circle-specific shell script and invoke it from the circle config. Ideally project.rb would not be aware that it's running within Circle or even git.

api/libproject/devstart.rb
```diff
@@ -370,10 +370,26 @@ def run_public_api_tests(cmd_name, args)
...
+  if circle_branch.nil? || circle_branch.empty? || circle_branch == "master"
+    run_tests = true
+  else
+    run_tests = test_api_changes(circle_branch)
+  end
+
+  if run_tests
```
calbach: We definitely should log that we skipped the tests and why

PR #1166, opened 2018-08-31
[RW-1258][risk=no]PM - selecting one heart rhythm item selects them both

smazumd:
PM - selecting one heart rhythm item selects them both

### DIFF
--- a/ui/src/app/cohort-search/attributes-page/attributes-page.component.ts
+++ b/ui/src/app/cohort-search/attributes-page/attributes-page.component.ts
@@ -40,6 +40,8 @@ export class AttributesPageComponent implements OnDestroy, OnInit {
   rangeAlert = false;
   loading: boolean;
   selectedCode: any;
+  sysOption: any;
+  diaOption: any;
   options = [
     {value: 'EQUAL', name: 'Equals', code: '01'},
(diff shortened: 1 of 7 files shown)


ui/src/app/cohort-search/attributes-page/attributes-page.component.ts
```diff
@@ -39,11 +39,12 @@ export class AttributesPageComponent implements OnDestroy, OnInit {
   subscription: Subscription;
   rangeAlert = false;
   loading: boolean;
+  selectedCode: any;
```
dolbeew: This needs to have a value assigned to it in ngOnInit for PM and in radioChange function for measurements. Otherwise param id can get set to 'paramNaN' because it's undefined, which will also allow adding 2 selections for 'Any' in PM

ui/src/app/cohort-search/attributes-page/attributes-page.component.ts
```diff
@@ -106,6 +107,7 @@ export class AttributesPageComponent implements OnDestroy, OnInit {
   }
 
   selectChange(index: number, option: any) {
+    this.selectedCode = option.code;
```
dolbeew: If the attribute is for Blood Pressure Detail (subtype is 'BP') this should get the code for both select options and add them together

PR #1169, opened 2018-08-31
[No Ticket][Risk=low] Abort running API tests if no api changes and not master

blrubenstein:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,11 +23,11 @@ jobs:
       - run:
           command: git submodule update --init --recursive
       - run:
-        command: |
-          if [git diff --name-only $(git merge-base master ${CIRCLE_BRANCH}) | grep api/ | wc -l] && ${CIRCLE_BRANCH != "master"}; then
-            echo No relevant changes on non-master branch, skipping
-            circleci step halt
-          fi
+          command: |
+            if [git diff --name-only $(git merge-base master ${CIRCLE_BRANCH}) | grep api/ | wc -l] && ${CIRCLE_BRANCH != "master"}; then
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -22,6 +22,13 @@ jobs:
       - checkout
       - run:
           command: git submodule update --init --recursive
+      - run:
+        #Circle is bad about setting up your remotes for you, you must use origin/
+          command: |
+            if [ $(git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep api/ | wc -l | xargs) == 0 ] && [ ${CIRCLE_BRANCH} != "master" ]; then
```
calbach: I would reverse the order of these predicates. I would also include the check for empty CIRCLE_BRANCH (I believe this happens when we tag a release - not 100%)

.circleci/config.yml
```diff
@@ -22,6 +22,13 @@ jobs:
       - checkout
       - run:
           command: git submodule update --init --recursive
+      - run:
+        #Circle is bad about setting up your remotes for you, you must use origin/
```
calbach: nit: +2 indent

.circleci/config.yml
blrubenstein: Will CIRCLE_BRANCH being empty be caught by != master?

blrubenstein: Will do!

calbach:
We want to run tests when we do a release

blrubenstein:
Yup, I realized right after commenting what you were actually saying... Should be good now @calbach.

PR #1177, opened 2018-09-05
[RW-1261][risk=no] Add basic tests for cohort review components

dolbeew:
Add basic render tests for all components under the cohort-review directory


dolbeew:
Yep, I meant to get all of those. Fixed now.

PR #1178, opened 2018-09-05
[RW-1250][risk=no] When parent is selected, gray out children

dolbeew:
When adding a parent to the right side, all the children will also be grayed out since they are technically selected


ui/src/app/cohort-search/node-info/node-info.component.ts
```diff
@@ -198,19 +205,27 @@ export class NodeInfoComponent implements OnInit, OnDestroy, AfterViewInit {
...
     return this.node.get('type') === TreeType[TreeType.PM];
   }
 
-  isDrug() {
+  get isDrug() {
     return this.node.get('type') === TreeType[TreeType.DRUG];
   }
 
-  isMeas() {
+  get isMeas() {
```
freemabd1: Would it make sense to create a function(consolidate isPM(), isDrug(), isMeas()) and some variables like so:
```
private pm = TreeType[TreeType.PM];
private drug = TreeType[TreeType.DRUG];
private meas = TreeType[TreeType.MEAS];
private icd9 = TreeType[TreeType.ICD9];
private icd10 = TreeType[TreeType.ICD10];
```

```
get isNodeType(... args) {
     return args.includes(this.node.get('type'));
}
```

And you can call it like so: 
`this.isNodeType(pm); `
or 
`this.isNodeType(drug, icd9, icd10);`

ui/src/app/cohort-search/node-info/node-info.component.ts
```diff
@@ -123,13 +119,20 @@ export class NodeInfoComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      || this.isPM) {
       return this.node.get('name', '');
     }
     return this.node.get('code', '');
   }
 
+  get selectAllChildren() {
+    return (this.isDrug
+      || this.node.get('type') === TreeType.ICD9
+      || this.node.get('type') === TreeType.ICD10)
```
freemabd1: Can consolidate this into one function. Please see comment on line 220.

ui/src/app/cohort-search/node-info/node-info.component.ts
freemabd1: Also, should we move this into a utils component since it seems to be a re-occurring theme in many components?

dolbeew: Yes, I think that's a good idea. Can do a follow-up pr soon to address here and other components as well

PR #1180, opened 2018-09-06
[RW-48] Adding API for retrieving CDR versions [risk=low]

danqrodney:


api/src/main/java/org/pmiops/workbench/api/CdrVersionsController.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (defaultCdrVersionId == null) {
+      // This shouldn't happen.
+      throw new ServerErrorException("User does not have access to default CDR version");
+    }
+    CdrVersionListResponse response = new CdrVersionListResponse();
+    response.setItems(cdrVersions.stream()
+        .map(TO_CLIENT_CDR_VERSION)
+        .collect(Collectors.toList()));
+    response.setDefaultCdrVersionId(defaultCdrVersionId);
```
freemabd1: Could this be the CdrVersion object(seems useful to access more info about the cdr)? I guess you could always look it up from the collection of CdrVersions.

api/src/main/java/org/pmiops/workbench/cdr/CdrVersionService.java
```diff
@@ -39,4 +70,25 @@ public void setCdrVersion(CdrVersion version) {
...
+   */
+  public List<CdrVersion> findAuthorizedCdrVersions(DataAccessLevel dataAccessLevel) {
+    ImmutableSet<Short> visibleValues = DATA_ACCESS_LEVEL_TO_VISIBLE_VALUES.get(dataAccessLevel);
+    if (visibleValues == null) {
+      return ImmutableList.of();
+    }
+    return cdrVersionDao.findByDataAccessLevelIn(visibleValues)
+        .stream()
+        .sorted(ORDER_BY_DESCENDING_CREATION_TIME_AND_ACCESS_LEVEL)
+        .collect(Collectors.toList());
```
freemabd1: So why do we sort the stream instead of letting jpa do the sort?

api/src/main/resources/workbench.yaml
```diff
@@ -1253,6 +1266,40 @@ definitions:
...
+  CdrVersion:
+    type: object
+    required:
+      - cdrVersionId
+      - name
+      - dataAccessLevel
+      - creationTime
+    properties:
+      cdrVersionId:
+        type: string
```
freemabd1: Should this be a long, since it's a long in the db?
type: integer
format: int64

api/src/main/resources/workbench.yaml
```diff
@@ -1253,6 +1266,40 @@ definitions:
...
+    required:
+      - items
+      - defaultCdrVersionId
+    properties:
+      items:
+        type: "array"
+        items:
+          $ref: "#/definitions/CdrVersion"
+      defaultCdrVersionId:
+        type: "string"
```
freemabd1: Should this be a long, since it's a long in the db?
type: integer
format: int64

api/src/main/java/org/pmiops/workbench/cdr/CdrVersionService.java
danqrodney: I went back and forth on this. It won't matter performance-wise. The code is sorta simpler with JPA doing the sorting, but it means a really long ugly method name. :) 

Anyway, I switched it to use JPA for the sort.

PR #1186, opened 2018-09-07
[RW-1177][Risk=no] Add modal to let us pick name and kernel for new notebooks

blrubenstein:


ui/src/app/app-routing.module.ts
```diff
@@ -170,16 +170,18 @@ const routes: Routes = [
         // The notebook redirect pages are interstitial pages, so we want to
         // give them special chrome treatment - we therefore put them outside
         // the normal /workspaces hierarchy.
-        path: 'workspaces/:ns/:wsid/notebooks/create',
+        path: 'workspaces/:ns/:wsid/notebooks/create/:nbName/:kernelType',
```
calbach: These feel like query parameters to me. Both could be optional.

ui/src/app/views/new-notebook-modal/component.html
```diff
@@ -0,0 +1,27 @@
+<clr-modal [(clrModalOpen)]="creatingNotebook" [clrModalSize]="'md'">
+  <div class="modal-title">
+    <!--The ngIf is to avoid null pointers before the modal is opened.-->
```
calbach: This doesn't seem necessary here.

ui/src/app/views/new-notebook-modal/component.html
```diff
@@ -0,0 +1,27 @@
...
+  </div>
+  <div class="modal-body">
+    <div class="form-section">
+      <label class="modal-label">New Name</label>
+      <input id="new-name" type="text" class="input"
+             [(ngModel)]="newName">
+      <label class="modal-label">Python or R?</label>
+      <div class="select">
+        <select [(ngModel)]="kernelType">
+          <option [ngValue]="Kernels.Python3">{{Kernels[Kernels.Python3]}}</option>
```
calbach: Will one of these be the default?

ui/src/app/views/new-notebook-modal/component.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        ClarityModule.forRoot(),
+        FormsModule
+      ],
+      declarations: [
+        NewNotebookModalComponent
+      ],
+      providers: [
+        {
+          provide: ServerConfigService,
+          useValue: new ServerConfigServiceStub({
```
calbach: This doesn't appear to be needed

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -0,0 +1,78 @@
+import {Component, Input, OnDestroy} from '@angular/core';
+
+import {SignInService} from 'app/services/sign-in.service';
+import {Kernels} from 'app/views/notebook-redirect/component';
```
calbach: This dependency doesn't seem ideal, would prefer pulling this enum out to a shared location, e.g. utils

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -30,6 +30,25 @@ enum Progress {
   Redirecting
 }
 
+export enum Kernels {
+  R,
+  Python3,
+}
+
+const rNotebookFileContent = '{ "cells": [ { "cell_type": "code", "execution_count": null, \
```
calbach: How did you generate this? Is it minimal? I think we should try to minimize any information being encoded here so we don't go out of sync with Leo / the Jupyter server.

For example, what happens if you just reduced this to `{kernelspec: {name: "ir"}}` or `{language_info: {name: "R"}}`

calbach: Please define this as an object, then use JSON.stringify(...) to turn it into a string. Else you're needlessly skipping validation of this JSON content.

calbach: Another idea which might be viable but which I have not thoroughly investigated is whether we can keep the creation logic the same, but modify the redirection URI to specify a different kernel type. On first save, this would stick.

Things to check on that approach:
- Whether a newly created notebook currently has a type and kernel type, or it is typeless and the Jupyter server is picking  adefault kernel
- Whether we might need to determine this information immediately after creation in order to display the notebook type on the list page (this might be an argument to continue with your current approach). 

Actually trying this out on Jupyter, I'm not sure there's a way to encode this in the URL so it may be a dead-end.

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.loading = false;
+  }
+
+  close(): void {
+    this.creatingNotebook = false;
+  }
+
+  submitNewNotebook(): void {
+    const nbUrl = `/workspaces/${this.workspace.namespace} \
+        ${this.workspace.id}/notebooks/create/${this.newName}/${this.kernelType}`;
```
calbach: You'll need to encode the notebook name: https://github.com/all-of-us/workbench/blob/1956c20e0319bffe1a7e0c7386d6542daffba233/ui/src/app/views/notebook-list/component.ts#L133

ui/src/app/views/new-notebook-modal/component.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        },
+        {provide: SignInService, useValue: new SignInServiceStub()}
+      ]
+    }).compileComponents().then(() => {
+      fixture = TestBed.createComponent(NewNotebookModalComponent);
+      tick();
+    });
+  }));
+
+  it('should render', fakeAsync(() => {
```
calbach: Would be good to have a more functional test here.

calbach: Note: you can inject window via WINDOW_REF for testing purposes.

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -0,0 +1,78 @@
...
+@Component({
+  selector: 'app-new-notebook-modal',
+  styleUrls: ['./component.css',
+    '../../styles/buttons.css',
+    '../../styles/inputs.css'],
+  templateUrl: './component.html',
+})
+export class NewNotebookModalComponent implements OnDestroy {
+  public creatingNotebook = false;
+  public newName = '';
```
calbach: Is the new name required, or not? Note that the Jupyter UI does not require specification of a new name so we may want to match that behavior.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -148,10 +182,12 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
       // API handles notebook name collisions and matches the behavior of
       // clicking "new notebook" in the Jupyter UI.
       const workspaceDir = localDir.replace(/^workspaces\//, '');
-      return this.jupyterService.postContents(
+      return this.jupyterService.putContents(
         this.cluster.clusterNamespace, this.cluster.clusterName,
-        workspaceDir, {
-          'type': 'notebook'
+        workspaceDir, this.notebookName + '.ipynb', {
```
calbach: Currently if they leave newName blank I believe you'll just get a file called ".ipynb".

ui/src/app/app-routing.module.ts
blrubenstein: I don't think we want them to be optional? I think part of the goal is that you need to pick a name and kernel when you create a new notebook. When would we not need them?

ui/src/app/views/new-notebook-modal/component.html
blrubenstein: Py3 is I believe. Will confirm

ui/src/app/views/new-notebook-modal/component.ts
blrubenstein: Will do!

ui/src/app/views/new-notebook-modal/component.ts
blrubenstein: It is required in this case, because we want to be able to encourage naming and renaming from our UI, to add to the cache. That is probably a Product question though. Will follow up with Karthik

ui/src/app/views/new-notebook-modal/component.ts
blrubenstein: Will do.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: I will check if we can do this, but my understanding is that it will break if we provide any less than that... 

I actually very much wish we could just pass the kernel in... but I couldn't figure out how to do that.

blrubenstein: Will do!

blrubenstein: Part of the goal here is also to do a majority of the work through our UI, rather than Jupyter's, because it means things get preserved in our cache.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: Correct. I can have it check in case and if it is empty make it untitled. That said, Karthik asked me to check if Saturn requires a name, and mirror what they do. So I will check on that

ui/src/app/app-routing.module.ts
calbach: Whether we want the name to be optional can be a product question. Whether we want the kernel to be optional, I'm not sure if we'll have other links to create a new notebook, but having a default value as we currently do might make sense in some of those cases.

I still think these don't feel correct as path parameters. Name/kernel type don't seem natural in a REST URL, rather than referring to an existing resource, they are more part of the creation payload (we're referencing a thing that doesn't yet exist). It feels like we could have a potentially growing list of values here (e.g. if we added more options to your modal). Hence why query parameters feel more natural to me.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: I generated this by downloading an empty notebook and copying the format. The issue is that we can't use notebook as the type and pass in content. I can show you in person tomorrow or Wednesday if you like.

ui/src/app/views/notebook-redirect/component.ts
calbach: Please use "put" if we go this route. Using "untitled" scales to exactly 1 usage, which is the reason I integrated with the Jupyter contents API in the first place.

ui/src/app/app-routing.module.ts
blrubenstein: That makes sense. I can convert them to query parameters.

blrubenstein:
Should cover all PR comments except more tests. Figured I would put up for another round of review while I thought about tests. @calbach PTAL when you get the chance.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: This ends up being obsoleted by it blocking '' as a name option.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -146,12 +228,14 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     return this.localizeNotebooks([]).flatMap((localDir) => {
       // Use the Jupyter Server API directly to create a new notebook. This
       // API handles notebook name collisions and matches the behavior of
-      // clicking "new notebook" in the Jupyter UI.
+      // clicking 'new notebook' in the Jupyter UI.
       const workspaceDir = localDir.replace(/^workspaces\//, '');
-      return this.jupyterService.postContents(
+      return this.jupyterService.putContents(
```
calbach: What happens if you choose an existing notebook name here?

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -146,12 +228,14 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
...
         this.cluster.clusterNamespace, this.cluster.clusterName,
-        workspaceDir, {
-          'type': 'notebook'
+        workspaceDir, this.notebookName + '.ipynb', {
+          'type': 'file',
```
calbach: Wanted to confirm that what you were trying included passing format "json" instead of "text". 

```
{
  'type': 'notebook',
  'format': 'json',
  'content': { ... } 
}
```

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -30,6 +31,70 @@ enum Progress {
   Redirecting
 }
 
+
+
+const rNotebookFileContent = {
+  'cells': [
+    {
+      'cell_type': 'code',
+      'execution_count': null,
```
calbach: We can go over this in person, but the root of my question is "how much of this object can you remove and having things still work as expected?". I'm concerned about this drifting from the default behavior for a new Jupyter notebook as we get new versions of Jupyter and Leo.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -61,7 +132,18 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this.wsNamespace = this.route.snapshot.params['ns'];
     this.wsId = this.route.snapshot.params['wsid'];
-    this.notebookName = this.route.snapshot.params['nbName'];
+    this.creating = this.route.snapshot.data.creating;
+
+    if (this.creating) {
+      this.notebookName = this.route.snapshot.queryParamMap.get('notebook-name');
+      if (this.route.snapshot.queryParamMap.get('kernel-type') === Kernels.R.toString()) {
+        this.fileContent = JSON.stringify(rNotebookFileContent);
```
calbach: nit: I think it would be cleaner to assign `this.kernelType` instead, then move this file content switch right before the putContents call. It feels to me like `kernelType` is a property of the component while `fileContent` is a downstream implementation detail to be isolated to a local variable if possible.

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.creatingNotebook = true;
+    this.loading = false;
+  }
+
+  close(): void {
+    this.creatingNotebook = false;
+  }
+
+  submitNewNotebook(): void {
+    const nbUrl = `/workspaces/` + encodeURIComponent(this.workspace.namespace) + '/' +
```
calbach: You shouldn't need it for these other parts, just the notebook name.

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -0,0 +1,89 @@
...
+    const notebook = window.open(nbUrl, '_blank');
+
+    // TODO(RW-474): Remove the authHandler integration. This is messy,
+    // non-standard, and currently will break in the following situation:
+    // - User opens a new notebook tab.
+    // - While that tab is loading, user immediately navigates away from this
+    //   page.
+    // This is not easily fixed without leaking listeners outside the lifespan
+    // of the workspace component.
+    const authHandler = (e: MessageEvent) => {
```
calbach: I'd complain about duplication here, but this code should be gone in the next couple of sprints so I don't think it's worth doing extra work.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: I don't know, but we should probably make that give an error anyway :) 
Will add an error message there.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: Yeah, I am also very worried about that... I will try to remove what I can, but my suspicion is that most of it is necessary for it to operate in Leo/Jupyter as a notebook.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: Yeah, just tried passing notebook as type, with format JSON, and got: 
`{"message": "Unexpected error while saving file: workspaces/test/Test NB.ipynb 'unicode' object has no attribute 'nbformat'", "reason": null}` again. (The name clash thing that I mentioned.)

blrubenstein: And type text does not accept json as a legal field.

blrubenstein:
@calbach when you get the chance, should be good for another round of reviews.

ui/src/app/views/new-notebook-modal/component.spec.ts
```diff
@@ -33,11 +39,60 @@ describe('NewNotebookModalComponent', () => {
...
+
+  it('errors if name exists', fakeAsync(() => {
+    updateAndTick(fixture);
+    simulateInput(fixture, fixture.debugElement.query(By.css('#new-name')), 'mockFile');
+    updateAndTick(fixture);
+    simulateClick(fixture, fixture.debugElement.query(By.css('.confirm-name-btn')));
+    expect(fixture.debugElement.query(By.css('.error'))).toBeDefined();
+  }));
+
+  it('does not allow blank names', fakeAsync(() => {
```
calbach: revert, or implement?

ui/src/app/views/new-notebook-modal/component.spec.ts
```diff
@@ -33,11 +39,60 @@ describe('NewNotebookModalComponent', () => {
...
+
+  // TODO: Get this to work. for some reason nothing would register
+  //       as an actual successful change of the select element.
+  it('allows creation of R notebooks', fakeAsync(() => {
+    spyOn(window, 'open');
+    const name = 'new-name-r';
+    updateAndTick(fixture);
+    simulateInput(fixture, fixture.debugElement.query(By.css('#new-name')), name);
+    updateAndTick(fixture);
+    simulateInput(fixture, fixture.debugElement.query(By.css('#new-name')), name);
```
calbach: Is this needed? Duplicates above

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -44,10 +46,16 @@ export class NewNotebookModalComponent implements OnDestroy {
...
 
   submitNewNotebook(): void {
-    const nbUrl = `/workspaces/` + encodeURIComponent(this.workspace.namespace) + '/' +
-        encodeURIComponent(this.workspace.id) + '/notebooks/create/?notebook-name=' +
-        encodeURIComponent(this.newName) + '&kernel-type=' +
-        encodeURIComponent(this.kernelType.toString());
+    const existingNotebook =
+      this.existingNotebooks.find((currentNotebook) =>
+        currentNotebook.name === this.newName + '.ipynb');
+    if (existingNotebook !== undefined) {
```
calbach: nit: I think `!existingNotebook` is more readable

ui/src/app/views/notebook-redirect/component.ts
calbach: That error makes it sound like you may have passed a serialized JSON string. I don't feel strongly, but would be slightly more natural with type "json" passing json.

api/config/config_local.json
```diff
@@ -3,7 +3,7 @@
   "firecloud": {
     "baseUrl": "https:\/\/firecloud-orchestration.dsde-dev.broadinstitute.org",
     "billingAccountId": "014D91-FCB792-33D2C0",
-    "billingProjectPrefix": "aou-rw-local1-",
+    "billingProjectPrefix": "aou-rw-local2-",
```
calbach: Revert?

ui/src/app/views/new-notebook-modal/component.ts
blrubenstein: (I think you mean `existingNotebook`, but agreed.

api/config/config_local.json
blrubenstein: I actually ended up with a broken local... so I wanted to change it because this fixes it. (I think in the flurry of billing project fixes and breakages, my local got messed up somehow. Happy to revert back but I want to log that local2 has been used if we do, so there's not weird errors if we need to change it in the future.

ui/src/app/views/notebook-redirect/component.ts
blrubenstein: Content currently just accepts string arguments, not anything, so it doesn't really allow passing in as an object. I think type is a specific enum on the JupyterApi side.

On the error, we actually did a bunch of googling trying to figure it out, and it sounded like it was an explicit name collision that caused it, and the error message was just really painfully unclear...

blrubenstein:
Per recommendation from Jeremy, I am moving to radio buttons here.

PR #1188, opened 2018-09-07
Fixing path to CDR version config file on deploy [risk=no]

danqrodney:


danqrodney:
(Submitted TBR to fix the build.)

PR #1189, opened 2018-09-10
[RW-854][Risk=low] Adding Recent Work Scroll to Homepage

UrsaStutsman:
This is everything sans tests. Writing tests now. Broken Tests will be fixed with new tests.
UPDATE 9/12: Refactoring of ResourceCard, still working on related tests. 
Closed due to some sort of mess-up when rebasing master. Work continued at https://github.com/all-of-us/workbench/pull/1212


ui/src/app/app.module.ts
```diff
@@ -57,14 +59,14 @@ import {CohortCommonModule} from './cohort-common/module';
 import {IconsModule} from './icons/icons.module';
 
 import {
-  ApiModule,
-  ConfigService,
-  Configuration,
+ApiModule,
+ConfigService,
+Configuration,
```
rushtong: nit: please revert the spacing change

ui/src/app/app.module.ts
```diff
@@ -57,14 +59,14 @@ import {CohortCommonModule} from './cohort-common/module';
...
 } from 'generated';
 
 import {
-  ApiModule as LeoApiModule,
-  Configuration as LeoConfiguration,
+ApiModule as LeoApiModule,
+Configuration as LeoConfiguration,
```
rushtong: nit: please revert the spacing change

ui/src/app/icons/right-scroll/component.css
```diff
@@ -0,0 +1,3 @@
+:hover {
+    fill: #72B9E2;
+}
```
rushtong: nit: newline

api/src/main/resources/workbench.yaml
```diff
@@ -2056,6 +2056,10 @@ definitions:
       workspaceId:
         type: integer
         format: int64
+      workspaceNamespace:
+        type: string
+      workspaceFirecloudName:
```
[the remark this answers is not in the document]
UrsaStutsman: I totally agree. BUT there's a bunch of logic done in previous work that counts on the DB ID here. SO. I'm going to submit a card, and add comments around the relevant code (that I know of, there maybe other repercussions that I haven't found) that reference the card. Let me know if this isn't satisfactory.

ui/src/app/views/card/component.ts
```diff
@@ -0,0 +1,185 @@
...
+    '../../styles/template.css',
+    './component.css'],
+  templateUrl: './component.html'
+})
+
+export class CardComponent implements OnInit {
+  actionList = [];
+  type: string;
+  @Input('card')
+  card: RecentResource;
```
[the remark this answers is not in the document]
UrsaStutsman: Not sure what this nit is about?

ui/src/app/utils/resourceActions.ts
```diff
@@ -0,0 +1,48 @@
+export class NotebookActions {
```
calbach: nit: you shouldn't need the class wrapper here,  you could just export the lists.

ui/src/app/views/recent-work/component.html
```diff
@@ -0,0 +1,18 @@
+<div *ngIf="elementVisible()" class="recent-work" style="display: inline-block">
+    <div *ngIf="leftScrollVisible()" class="scroll-indicator" id="left-scroll" (click)="moveDownList()">
```
calbach: nit: we use 2 line indentation in other html templates

ui/src/app/views/recent-work/component.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      .toEqual(['mockFile4.ipynb', 'mockFile3.ipynb', 'mockFile2.ipynb']);
+    expect(rightScroll).toBeDefined();
+    expect(leftScroll).toBeNull();
+  }));
+});
+
+function stubRecentResources(numberOfResources: number) {
+  const currentCache = [];
+  while (numberOfResources > 0) {
+    const currentResource = {
```
calbach: nit: I would probably just inline this below, I don't think the local var is helping readability

ui/src/app/views/recent-work/component.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+import {RenameModalComponent} from 'app/views/rename-modal/component';
+
+import {LeftScrollComponent} from 'app/icons/left-scroll/component';
+import {RightScrollComponent} from 'app/icons/right-scroll/component';
+
+describe('RecentWorkComponent', () => {
+  let fixture: ComponentFixture<RecentWorkComponent>;
+  let userMetricsSpy: jasmine.SpyObj<UserMetricsService>;
+  beforeEach(fakeAsync(() => {
+    const spy = jasmine.createSpyObj('UserMetricsService', ['getUserRecentResources']);
```
calbach: In general I'd prefer fakes to mocks, which is what we're doing elsewhere for the other services.

ui/src/app/utils/resourceActions.ts
```diff
@@ -0,0 +1,44 @@
+  export const NotebookActionList = [
```
calbach: nit: would lowercase these since they aren't classes anymore. We do this elsewhere for exported constants

ui/src/app/views/recent-work/component.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  // test that component scrolls correctly
+  //    moves down list on right scroll click
+  //    right scroll disappears and left appears
+  //    moves up list on left scroll click
+  //    left scroll disappears and right appears
+  it('should scroll correctly', fakeAsync(() => {
+    userMetricsSpy.getUserRecentResources.and.returnValue(Observable.of(stubRecentResources(4)));
+    updateAndTick(fixture);
+    const de = fixture.debugElement;
+    const rightScroll = () => de.query(By.css('#right-scroll'));
```
calbach: nice

ui/src/app/views/resource-card/component.ts
```diff
@@ -0,0 +1,214 @@
...
+export class ResourceCardComponent implements OnInit, OnDestroy {
+  resourceType: ResourceType;
+  @Input('resourceCard')
+  resourceCard: RecentResource;
+  @Output() onUpdate: EventEmitter<void> = new EventEmitter();
+  actions = [];
+  notebookRenameError = false;
+  wsNamespace: string;
+  wsId: string;
+  resource: any;
```
calbach: I believe this is no longer used

ui/src/app/views/resource-card/component.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    permission: 'Owner',
+    modifiedTime: Date.now().toString(),
+    notebook: {
+      'name': 'mockFile.ipynb',
+      'path': 'gs://bucket/notebooks/mockFile.ipynb',
+      'lastModifiedTime': 100
+    },
+  };
+  // call OnInit here because that's where ResourceCard reads the resourceType
+  component.ngOnInit();
```
calbach: FYI: I recalled having solved this issue during review, but I just happened upon the code. The downside is that every test will need to call this initially (or *maybe* you could call it once in beforeEach, and optionally a second time here...). The upside is that you don't need to hand-replicate the angular lifecycle and injection:
https://github.com/all-of-us/workbench/blob/876114ab08ff32f4c197874882b1bcb49c102bcd/ui/src/app/views/workspace-edit/component.spec.ts#L34

UrsaStutsman: hmmm, so not sure what you want me to change here. The example you gave is not exactly the same. I am still foggy from this cold, so sorry if I'm being dim. Leaving as is for now.

PR #1192, opened 2018-09-10
[DB-29] Start moving CDR version stuff into common-api. [risk=low]

danqrodney:
(This isn't working yet.)


api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -39,8 +39,8 @@
       {
         DataAccessLevel.class.getSimpleName(),
         DataAccessLevel.values(),
-        (Function<Short, DataAccessLevel>) StorageEnums::dataAccessLevelFromStorage,
-        (Function<DataAccessLevel, Short>) StorageEnums::dataAccessLevelToStorage
+        (Function<Short, DataAccessLevel>) CommonStorageEnums::dataAccessLevelFromStorage,
```
calbach: Should fork the test instead

common-api/build.gradle
```diff
@@ -35,13 +95,28 @@ dependencies {
...
+  compile 'com.google.code.gson:gson:2.8.5'
 
   testCompile 'junit:junit:4.12'
   testCompile 'org.mockito:mockito-core:1.10.19'
   testCompile 'com.google.truth:truth:0.42'
   testCompile 'com.h2database:h2:1.4.194'
   testCompile 'org.liquibase:liquibase-core:3.5.3'
   testCompile 'org.bitbucket.radistao.test:before-after-spring-test-runner:0.1.0'
+
+  // For Swagger generation. These should include dependencies found in the Swagger codegen
```
calbach: Should we remove these deps from api/build.gradle, then?

public-api/build.gradle
```diff
@@ -144,12 +150,13 @@ dependencies {
   generatedCompile 'com.google.code.gson:gson:2.8.5'
   generatedCompile 'joda-time:joda-time:2.10'
   generatedCompile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.9.6'
+  generatedCompile project(":common-api")
```
calbach: why?

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
```diff
@@ -0,0 +1,75 @@
...
+import org.pmiops.workbench.db.model.CommonStorageEnums;
+import org.pmiops.workbench.exceptions.ServerErrorException;
+import org.pmiops.workbench.model.CdrVersionListResponse;
+import org.pmiops.workbench.model.DataAccessLevel;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class CdrVersionsController implements CdrVersionsApiDelegate {
```
calbach: Is it not possible for this to live in common-api?

public-api/src/main/webapp/WEB-INF/appengine-web.xml.template
```diff
@@ -1,7 +1,8 @@
 <appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
-  <service>public-api</service>
```
calbach: I believe this change would cause workbench and db APIs to deploy to the same GAE service.

common-api/build.gradle
danqrodney: I don't think so? We are generating code from Swagger specs in both places. If we stopped generating Swagger in common-api, we'd still have these deps in api... so normally I'd say we should spell out the direct dependencies everywhere.

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
danqrodney: The implementation is slightly different; this one doesn't depend on an authenticated user (we don't have one). I could try to refactor out common code if you'd like into a base class in common-api... given how simple the logic is, I wasn't sure it was worth the trouble.

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
```diff
@@ -0,0 +1,75 @@
...
+    this.workbenchConfigProvider = workbenchConfigProvider;
+  }
+
+
+  @Override
+  public ResponseEntity<CdrVersionListResponse> getCdrVersions() {
+    // We return CDR versions for both registered and controlled CDR versions in the public API,
+    // since the metadata for them is public.
+    List<CdrVersion> cdrVersions = cdrVersionDao
+        .findByDataAccessLevelInOrderByCreationTimeDescDataAccessLevelDesc(REGISTERED_AND_PROTECTED);
```
danqrodney: Per discussion with @biopete and Kayla, changing this to just REGISTERED.

public-api/src/main/webapp/WEB-INF/appengine-web.xml.template
danqrodney: Oops! Good catch. Fixed.

.gitignore
```diff
@@ -51,7 +53,10 @@ public-api/wrapper
 public-api/.swagger-codegen/*
 public-api/src/generated/
 public-api/src/main/webapp/WEB-INF/appengine-web.xml
-
+public-api/src/main/resources/merged.yaml
+public-api/gradle
```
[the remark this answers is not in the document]
danqrodney: Intellij generated these gradle wrapper files in my public-api dir. (I'm not sure if my settings were different whether this would have happened, but we don't want to include these files in the repo.)

public-api/build.gradle
danqrodney: Removed.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
danqrodney: I'm working on this. (At the moment, we don't have a test dir in common-api.)

danqrodney: (Added common-api test.)

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
biopete: Yeah. Not worth the trouble now.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -307,6 +335,7 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
 
     @Override
     public ResponseEntity<DbDomainListResponse> getSurveyList() {
+        CdrVersionContext.setCdrVersionNoCheckAuthDomain(defaultCdrVersionProvider.get());
```
biopete: So the idea is with every api call you just call this first?   If this api call gets moved to common with this same code just work?

danqrodney: I don't think this controller should moved to common; controllers in workbench generally require that the user is signed in, and use the user's credentials to determine what data they have access to. That said, yes, if you moved this to common it should still work I believe.

All of these calls should be replaced by a call that will accept a CDR version ID parameter, look up that CDR version, and call CdrVersionContext.setCdrVersionNoCheckAuthDomain with it. That will allow for CDR version switching. But I didn't want to do that in this PR.

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
biopete: We will be able to have a common controller though for common endpoints that are exactly the same right ? I can see a lot of the stuff in public being common eventually

danqrodney: If they were exactly identical, sure, it could live in common. And when they differ slightly (as in this case), we have the option of making a base controller class in common-api with common code, and subclasses in api and public-api.

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
biopete: Cool. Yeah, sorry I just meant like on an individual endpoint basis ... like some endpoints may be exactly the same in workbench or public. So the common controller would be good that could be merged with the individual controllers. Not really sure how that would be implemented.

common-api/build.gradle
calbach: On further thought, that is definitely preferred.

common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
```diff
@@ -74,9 +74,7 @@ public CdrDataSource(CdrVersionDao cdrVersionDao,
           cdrVersionId = cdrVersion.getCdrVersionId();
         }
 
-        // TODO: Reconcile this with the public-api DB configs; this should
-        // share code and use the same approach, but attach to the public DBs.
-        String dbName = cdrVersion.getCdrDbName();
+        String dbName = "public".equals(dbUser) ? cdrVersion.getPublicDbName() : cdrVersion.getCdrDbName();
```
biopete: NIT: our dbUser always has to be named public. That is fine . Too bad there isn't another property on dataSource that says if it's public or private. No biggie.

danqrodney: I could make this a config setting if you prefer. But this seemed OK for now; I don't have any reason to expect our DB user to change.

danqrodney:
@calbach @biopete Do you have any further comments?

biopete:
How do we change the public cdr version now ? Do we just set it in the workbench db ?

danqrodney:
As mentioned offline: you can change the contents of cdr_version by populating the relevant CDR version config file, e.g.

https://github.com/all-of-us/workbench/blob/master/api/config/cdr_versions_local.json

And you can change what the "default" CDR version is by changing workbench config:

https://github.com/all-of-us/workbench/blob/master/api/config/config_local.json#L20

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+
+  @Override
+  public ResponseEntity<CdrVersionListResponse> getCdrVersions() {
+    // We return CDR versions for just registered CDR versions; controlled CDR data is currently
+    // out of scope for the data browser.
+    List<CdrVersion> cdrVersions = cdrVersionDao
+        .findByDataAccessLevelInOrderByCreationTimeDescDataAccessLevelDesc(
+            ImmutableSet.of(CommonStorageEnums.dataAccessLevelToStorage(DataAccessLevel.REGISTERED)));
```
calbach: Wouldn't we want UNREGISTERED as well?

common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
calbach: Some kind of injected confg property does seem like it would be safer if we already have something. I suppose even if this logic breaks the public user will fail hard trying to access the private CDR index

public-api/src/main/java/org/pmiops/workbench/publicapi/CdrVersionsController.java
danqrodney: So we're using DataAccessLevel for CDR versions and *also* for users. For CDR versions, my assumption was the only valid values are REGISTERED and CONTROLLED -- no CDR versions are accessible to people who haven't gotten REGISTERED data access. For users, we use UNREGISTERED to represent users who haven't gotten data access yet.

(We could consider having two separate enums here to make this clearer.)

But anyway, I think REGISTERED is all we want here.

common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
danqrodney: Correct -- only "workbench" has access to the CDR DB. So until I added this check, attempts to access CDR DB info as the public DB user failed. 

We don't already have a config setting for this, but I'm happy to add one if you think it's worthwhile.

PR #1196, opened 2018-09-13
[RW-1286] Adding getDataTableQuery and getCohortAnnotations methods [risk=no]

danqrodney:
Implementation of each will come later.

For the moment, I'm not proposing we deprecate materializeCohort... there may be clients that can't easily query BigQuery themselves, and could still use it. But we won't use it anymore in our pyclient library once this gets implemented.

### DIFF
--- a/api/src/main/resources/client_api.yaml
+++ b/api/src/main/resources/client_api.yaml
@@ -188,11 +188,8 @@ definitions:
       name:
         description: the name of the parameter
         type: string
-      value:
-        description: A string value for this parameter.
-        type: string
       values:
-        description: An array of string values for this parameter.
+        description: An array of string values for this parameter. (It may contain just one value.)
         type: array
         items:
           type: string
@@ -202,11 +199,8 @@ definitions:
       valueDatetime:
         description: A datetime (yyyy-MM-dd HH:mm:ss zzz) value for this parameter
         type: string
-      valueNumber:
-        description: A number value for this parameter.
-        type: number
       valueNumbers:
-        description: An array of number values for this parameter
+        description: An array of number values for this parameter. (It may contain just one value.)
         type: array
         items:
           type: number
@@ -222,7 +216,7 @@ definitions:
         type: string
       statusFilter:
         description: >
-          An array of status values; participants with these statuses will be have their annotations
+          An array of status values; participants with these statuses will have their annotations
           retrieved. Defaults to [NOT_REVIEWED, INCLUDED, NEEDS_FURTHER_REVIEW] --
           everything but EXCLUDED.
         type: array
--- a/api/src/main/resources/client_api.yaml
+++ b/api/src/main/resources/client_api.yaml
@@ -53,28 +53,6 @@ parameters:
 
 paths:
 
-  /v1/workspaces/{workspaceNamespace}/{workspaceId}/materializeCohort:
-    parameters:
-      - $ref: '#/parameters/workspaceNamespace'
-      - $ref: '#/parameters/workspaceId'
-    post:
-      tags:
-        - cohorts
-      description: >
-        Materializes a cohort for a given CDR version to specified output.
-      operationId: "materializeCohort"
-      parameters:
-       - in: body
-         name: request
-         description: cohort materialization request
-         schema:
-           $ref: "#/definitions/MaterializeCohortRequest"
-      responses:
-        200:
-          description: The results of materializing the cohort
-          schema:
-            $ref: "#/definitions/MaterializeCohortResponse"
-
   /v1/workspaces/{workspaceNamespace}/{workspaceId}/getDataTableQuery:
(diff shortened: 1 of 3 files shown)


danqrodney:
(I could imagine moving materializeCohort out of client_api.yaml and into workbench.yaml, so we no longer generate code for it in our client libs... let me know what you think about that.)

calbach:
Seems better to remove the API entirely if we don't have any planned usages for it (once it's been phased out of the client lib anyways). We can always add code back in if we decide we need it. If you think it will be needed by clients, then I'd leave it in client_api.

danqrodney:
I don't think it's needed by current clients, certainly. So I'll move it out of client_api.yaml for now. (We can talk about getting rid of it entirely later, once it's no longer in use by client libs.)

danqrodney:
(Moved to workbench.yaml.)

api/src/main/resources/client_api.yaml
```diff
@@ -165,6 +210,139 @@ definitions:
...
+        type: array
+        items:
+          $ref: "#/definitions/CohortStatus"
+      cdrVersionName:
+        description: >
+          The name of a CDR version to use when retrieving annotations; if none is specified,
+          the CDR version currently associated with the workspace will be used
+        type: string
+      annotationQuery:
+        description: >
```
freemabd1: Would it make sense to make statusFilter above become part of the annotationQuery? Seems like these would be used together in implementation?

api/src/main/resources/client_api.yaml
```diff
@@ -124,64 +150,111 @@ definitions:
...
+      cdrVersionName:
+        description: >
+          The name of a CDR version to use when retrieving annotations; if none is specified,
+          the CDR version currently associated with the workspace will be used
+        type: string
       annotationQuery:
         description: >
-          A query specifying how to retrieve annotation values created about participants in a
-          cohort during cohort review. Either this or tableQuery should be set (not both.)
+          Specification defining what annotations to retrieve.
```
freemabd1: Would it make sense to add statusFilter from above to annotationQuery? Seems like they will be used together in implementation?

api/src/main/resources/client_api.yaml
```diff
@@ -165,6 +210,139 @@ definitions:
...
+      - name
+    description: A query parameter and value. One (and only one) of the value fields should be set.
+    properties:
+      name:
+        description: the name of the parameter
+        type: string
+      value:
+        description: A string value for this parameter.
+        type: string
+      values:
```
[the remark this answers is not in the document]
danqrodney: That's fine. Ultimately I think we want to generate different parameters when calling BQ for the one vs. many case, but we could have logic based on the length of the array.

api/src/main/resources/client_api.yaml
```diff
@@ -165,6 +210,139 @@ definitions:
...
+      valueDate:
+        description: A date (yyyy-MM-dd) value for this parameter
+        type: string
+      valueDatetime:
+        description: A datetime (yyyy-MM-dd HH:mm:ss zzz) value for this parameter
+        type: string
+      valueNumber:
+        description: A number value for this parameter.
+        type: number
+      valueNumbers:
```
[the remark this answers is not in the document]
danqrodney: Done.

api/src/main/resources/client_api.yaml
```diff
@@ -165,6 +210,139 @@ definitions:
...
+    required:
+      - cohortName
+    properties:
+      cohortName:
+        description: >
+         The name of a cohort that annotations should be retrieved for.
+        type: string
+      statusFilter:
+        description: >
+          An array of status values; participants with these statuses will be have their annotations
```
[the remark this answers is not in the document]
danqrodney: Done.

api/src/main/resources/client_api.yaml
danqrodney: So the current code has a separation between TableQuery / AnnotationQuery (which indicate what data to get about a person) and MaterializeCohortRequest / DataTableSpecification / CohortAnnotationRequest (which indicate what *people* to get data for, as well as including what data to get about them.)

I like this separation, since you can totally imagine re-using TableQuery or AnnotationQuery in other contexts in future (e.g. "get me the data for just person X"). So I think I want to keep statusFilter separate here.

(Also FWIW: we need to make sure we maintain backwards compatibility for materialize cohort logic until we get rid of it.)

api/src/main/resources/client_api.yaml
danqrodney: See above.

freemabd1: sorry for the dup question. For some reason when i submitted my review I couldn't see that question anymore and wasn't sure if I forgot to click button to add the comment to the review. Anyway looks good.

PR #1198, opened 2018-09-13
[RISK=NO] Unable to edit data set on edit workspace, if the Fill later option is selected

NehaBroad:


ui/src/app/views/workspace-edit/component.html
```diff
@@ -71,7 +71,7 @@ <h4 class="section-header">Please select all data use categories that apply for
                   name="disease-focused-research"
                   id="disease-focused-research"
                   [(ngModel)]="workspace.researchPurpose.diseaseFocusedResearch"
-                  [clrDisabled]="mode == Mode.Edit">
+                  [clrDisabled]=!canEditDataSet>
```
blrubenstein: Nit: Can we change this to `canEditResearchPurpose` or something similar? This just reads to me like we are editing the actual CDR for the workspace.

PR #1200, opened 2018-09-14
[RW-574][risk=no] Add Custom App ID Header for FireCloud

rushtong:
## Addresses:
https://precisionmedicineinitiative.atlassian.net/browse/RW-574

## Notes
* Upgraded all of the okhttp libs to latest. Setting clients to debug was throwing an error with the older/mis-matched versions previously
* Added the custom header to the default FC client as well as the custom Status client
* Added configs to the integration source path so we can read in those configs if necessary
* I tested locally with debug enabled and was able to see the correct headers pass through to FC.


api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -41,17 +50,18 @@
...
-        assertThat(fireCloudService.getFirecloudStatus()).isTrue();
-    }
-
-    private static WorkbenchConfig createConfig() {
-        WorkbenchConfig config = new WorkbenchConfig();
-        config.firecloud = new WorkbenchConfig.FireCloudConfig();
-        config.firecloud.debugEndpoints = true;
-        return config;
-    }
+  private static WorkbenchConfig createConfig() throws Exception {
```
danqrodney: How does using config_local.json affect the functioning of this test (if at all?)

(It doesn't necessarily hurt anything... I'm just curious what prompted the change here.)

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -102,4 +104,20 @@ public GroupsApi groupApi(@Qualifier(END_USER_API_CLIENT) ApiClient apiClient) {
     return api;
   }
 
+  @Bean
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
+  public StatusApi statusApi(WorkbenchConfig workbenchConfig) {
+    StatusApi statusApi = new StatusApi();
+    ApiClient apiClient = new ApiClient();
```
danqrodney: So is the deal that we don't need to be auth'd to call the status endpoint, and thus want to make a separate ApiClient here?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -102,4 +104,20 @@ public GroupsApi groupApi(@Qualifier(END_USER_API_CLIENT) ApiClient apiClient) {
...
+    apiClient.setBasePath(workbenchConfig.firecloud.baseUrl);
+    apiClient.setDebugging(workbenchConfig.firecloud.debugEndpoints);
+    addFireCloudDefaultHeader(apiClient);
+    statusApi.setApiClient(apiClient);
+    return statusApi;
+  }
+
+  private void addFireCloudDefaultHeader(ApiClient apiClient) {
+    apiClient.addDefaultHeader("X-App-ID", "AoU-RW");
```
danqrodney: Do we need to have a different app ID for different environments?

Could define constant or config setting for this....

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -82,7 +78,7 @@ private void checkAndAddRegistered(WorkspaceIngest workspaceIngest) {
   @Override
   public boolean getFirecloudStatus() {
     try {
-      new StatusApi().status();
+      new FireCloudConfig().statusApi(configProvider.get()).status();
```
danqrodney: We should instead inject Provider&lt;StatusApi&gt; into the constructor and use it here, I think.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
rushtong: The only reason to use the config is to get the base firecloud url that the status call hits. I didn't want to hard-code that into the test itself

rushtong: It was an intentional change ... but I'm rethinking it now. This is an integration test so maybe this should be the production url. Yeah ... the more I think about it, the more I lean towards hitting prod, not dev.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
danqrodney: OK. Let me know if I can help.

rushtong:
@danqrodney - PTAL, feedback applied. Thank you!

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
rushtong: I was having trouble with that. For autowired, the app code needs a config to do the population of the client. The test needs to also provide a config, any config, but I could not get the autowired to work so resorted to a manual status api creation.

danqrodney: So previously, createConfig() just constructed a WorkbenchConfig that didn't read in any JSON file... the only thing that got populated was debugEndpoints. I was wondering why that didn't continue working the way it was.

There is a way of doing all this with @Autowired too, FWIW, but I don't know that it's worth bothering in this case.

Anyway, this is fine, too.

rushtong: The previous way of creating a config didn't work due to not having a base url. Since we use that in the construction in `FirecloudConfig` we need a real value to populate it with.

PR #1210, opened 2018-09-18
[RW-1222][risk=low] Implement CDR version toggle on create/clone

calbach:
- Make the CdrVersions endpoint work for dev/test (!enforceRegistered)
- Add a cache service for CdrVersions, as these won't change normally.

image

### DIFF
--- a/ui/src/app/views/workspace-edit/component.html
+++ b/ui/src/app/views/workspace-edit/component.html
@@ -14,11 +14,28 @@
       </ng-container>
       <div class="required-text">(Required)</div>
     </div>
-    <div class="name-area required-field" [class.validation-error]="nameNotEntered">
+    <div class="name-area required-field" [class.validation-error]="nameNotEntered"
+          style="display: inline-block;">
       <input type="text" class="input name" name="workspace-name" autofocus
              [(ngModel)]="workspace.name" [disabled]="!hasPermission"
              placeholder="Workspace Name">
     </div>
+    <clr-dropdown style="display: inline-block;">
+      <button [disabled]="mode == Mode.Edit"
+          class="btn btn-dropdown cdr-version-dropdown" type="button" clrDropdownTrigger>
+        {{getSelectedCdrName()}}
+        <clr-icon shape="caret down"></clr-icon>
+      </button>
+      <clr-dropdown-menu *clrIfOpen>
+        <ng-container *ngFor="let version of cdrVersions">
+          <button type="button" clrDropdownItem class="cdr-version-item"
+                  [class.active]="workspace.cdrVersionId === version.cdrVersionId"
+                  (click)="workspace.cdrVersionId = version.cdrVersionId">
+            {{version.name}}
+          </button>
+        </ng-container>
+      </clr-dropdown-menu>
+    </clr-dropdown>
   </div>
   <div>
 
--- a/ui/src/app/views/workspace-edit/component.html
+++ b/ui/src/app/views/workspace-edit/component.html
@@ -15,17 +15,24 @@
       <div class="required-text">(Required)</div>
     </div>
     <div class="name-area required-field" [class.validation-error]="nameNotEntered"
-          style="display: inline-block;">
+          style="display: inline-block; vertical-align: middle;">
       <input type="text" class="input name" name="workspace-name" autofocus
              [(ngModel)]="workspace.name" [disabled]="!hasPermission"
              placeholder="Workspace Name">
     </div>
-    <clr-dropdown style="display: inline-block;">
-      <button [disabled]="mode == Mode.Edit"
-          class="btn btn-dropdown cdr-version-dropdown" type="button" clrDropdownTrigger>
-        {{getSelectedCdrName()}}
-        <clr-icon shape="caret down"></clr-icon>
-      </button>
+    <clr-dropdown style="display: inline-block; vertical-align: middle;">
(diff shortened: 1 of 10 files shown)


calbach:
@danqrodney please review the changes to the CdrVersions controller
@blrubenstein please review the rest of the PR

Thanks

ui/src/app/views/workspace-edit/component.ts
```diff
@@ -369,6 +379,14 @@ export class WorkspaceEditComponent implements OnInit {
         });
   }
 
+  getSelectedCdrName(): string {
```
blrubenstein: Can we make this a `get` function?

ui/src/app/views/workspace-edit/component.html
```diff
@@ -14,11 +14,28 @@ <h2 class="page-header">Clone workspace "{{oldWorkspaceName}}"</h2>
...
     </div>
-    <div class="name-area required-field" [class.validation-error]="nameNotEntered">
+    <div class="name-area required-field" [class.validation-error]="nameNotEntered"
+          style="display: inline-block;">
       <input type="text" class="input name" name="workspace-name" autofocus
              [(ngModel)]="workspace.name" [disabled]="!hasPermission"
              placeholder="Workspace Name">
     </div>
+    <clr-dropdown style="display: inline-block;">
+      <button [disabled]="mode == Mode.Edit"
```
blrubenstein: I think it would be good to have a tooltip to explain why you can't edit it.

ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -87,13 +91,32 @@ describe('WorkspaceEditComponent', () => {
         // for testing.
         { provide: ActivatedRoute, useFactory: () => activatedRouteStub },
         { provide: ProfileStorageService, useValue: new ProfileStorageServiceStub() },
+        {
+          provide: CdrVersionStorageService,
+          useValue: new CdrVersionStorageServiceStub({
```
blrubenstein: Do we think this could be used more places than here? If so, can we make stub variables rather than use hardcoded values here? (Put the variables in the stub file as static values/exportable constants.)

ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -173,30 +196,64 @@ describe('WorkspaceEditComponent', () => {
...
-      fixture.detectChanges();
-      tick();
-      expect(workspacesService.workspaces.length).toBe(2);
-      const got = workspacesService.workspaces.find(w => w.name === testComponent.workspace.name);
-      expect(got).not.toBeNull();
-      expect(got.namespace).toBe(
-        ProfileStubVariables.PROFILE_STUB.freeTierBillingProjectName);
-      expect(spy).toHaveBeenCalled();
-    })));
+  fit('should support cloning a workspace with cdr version change', fakeAsync(() => {
```
blrubenstein: Remove fit.

ui/src/app/views/workspace-edit/component.html
calbach: Done

ui/src/app/views/workspace-edit/component.spec.ts
calbach: Maybe, I think we should abstract it there once there is more than one test using it. There is a price to moving the definition of a model into a utility file somewhere else, while your test is checking for some specific relationship in the data, e.g. in this test I depend on "2" being the default, and I depend on 2 mapping to name "cdr2". This would be very opaque if that were defined elsewhere.

ui/src/app/views/workspace-edit/component.spec.ts
calbach: Oops, thanks. Done

ui/src/app/views/workspace-edit/component.ts
calbach: Done

ui/src/app/views/workspace-edit/component.html
```diff
@@ -14,11 +14,35 @@ <h2 class="page-header">Clone workspace "{{oldWorkspaceName}}"</h2>
...
+                  class="btn btn-dropdown cdr-version-dropdown" type="button" clrDropdownTrigger>
+            {{selectedCdrName}}
+            <clr-icon shape="caret down"></clr-icon>
+          </button>
+          <div *ngIf="mode == Mode.Edit" class="tooltip-text">
+            To use a different dataset version, clone or create a new workspace.
+          </div>
+        </div>
+      </div>
+      <clr-dropdown-menu *clrIfOpen>
```
danqrodney: What CDR version is getting selected by default? (It should be the default returned in the CDR versions endpoint...)

We should possibly have a message that appears such that when you pick an older CDR version, we warn you that it's only OK to do this if continuing with an existing research purpose or reproducing someone else's work.

calbach: The default CDR version is selected by default.

Filed RW-1327 to track warning the user about the policy. I don't think we have the details/language right now.

calbach:
PTAL

ui/src/app/views/workspace-edit/component.spec.ts
blrubenstein: Perfectly happy with that. Sounds good.

PR #1211, opened 2018-09-19
[RW-1276] Experimental: replace DbDomain with DomainInfo, make survey modules config based

danqrodney:
This is nowhere near to a working PR, but shows how we might go about:

1) replacing DbDomain with DomainInfo (with just the fields we intend to use, and no more vocab stuff)
2) using a JSON-based config for representing survey modules, topics, etc. rather than new DB tables or using DbDomain

I wanted to chat with you about the approach @biopete before proceeding further.


api/db-cdr/csv/domain_to_domain_id.csv
```diff
@@ -0,0 +1,15 @@
+"domain","domain_id"
+0,"Condition"
```
biopete: I don't understand how these Condition/Observation type domains are used in workbench or whatever. There are no concepts in concept table that have Condition/Observation domain id. But we do have counts like for concept id :  141693  that came from Condition table and the Observation table. Do : select * from cdr20180913.achilles_results where stratum_1 = '141693' 
on test db . 

If you want to know what bq tables to search when you pick a concept id -- you should look at that the achilles_results analysis_id 3000 for that concept_id and it will tell you what tables you will find rows in.

danqrodney: After discussion offline: we're going to ignore the domain IDs with '/' in them for now. So I'm going to get rid of this file, and the logic in concept search that uses mappings like these.

danqrodney:
I'm going to punt on #2 for now (the story for how we represent questionnaire modules, etc. is still being worked out) and #3. Will update this PR when ready.

danqrodney:
@calbach and @biopete I updated the PR. I still need to update tests, UI, and SQL scripts (for populating concept counts on domain_info and question counts on survey_module). But PTAL and let me know if the general setup looks good to you.

biopete:
Looks good to me. Definitely an improvement. Thanks. If you want when it is ready Srushti or I can go in to this branch and update the ui to handle the new new response and update the generate cdr scripts.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
```diff
@@ -0,0 +1,51 @@
...
+
+  // TODO: consider not using @Query since it doesn't let us re-use shared SQL expressions for these methods?
+
+  /**
+   * Returns domain metadata and concept counts for domains, matching standard concepts by name
+   * and all concepts by code or concept ID.
+   */
+  @Query(nativeQuery=true,value="select d.domain, d.name, d.description,\n" +
+      "d.concept_id, COUNT(DISTINCT c.concept_id) as all_concept_count,\n" +
+      "SUM(c.standard_concept IN ('S', 'C')) as standard_concept_count,\n" +
```
calbach: This is a basic question, but what is the format of this column and where is it documented? I was sort of expecting that 'S' and 'C' were the only legal values, but then I'm not sure this select statement makes sense.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
```diff
@@ -0,0 +1,51 @@
...
+import org.pmiops.workbench.cdr.model.DomainInfo;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface DomainInfoDao extends CrudRepository<DomainInfo, Long> {
+
+  // TODO: consider not using @Query since it doesn't let us re-use shared SQL expressions for these methods?
+
+  /**
+   * Returns domain metadata and concept counts for domains, matching standard concepts by name
```
calbach: Possibly existing issue, but I can't tell what the parameters are controlling given the vague naming, could you add docs?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
danqrodney: 'S' = standard. 'C'=  standard classification... meaning it's like a standard higher-level concept, which may or may not have any rows in the data directly... (instead it may just have descendant concepts based on relationships with rows in the data.) We want to match either of those.

Not set is the other valid value for this column, and it means non-standard concept.

Documentation for this is here:

http://www.ohdsi.org/web/wiki/doku.php?id=documentation:cdm:concept

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
danqrodney: I updated the parameter name and docs here and in SurveyModuleDao.

danqrodney:
I attempted updating all the BQ scripts to populate and use domain_info and survey_module instead of db_domain. I'm testing them out now. @biopete PTAL.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -209,33 +209,80 @@ case when count_value > 0 then round(count_value/$person_count, 2)
...
+set sm.participant_count=c.count_value from
+`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.concept\` c
+where c.concept_id=sm.concept_id"
+
+# Set the question count on the survey_module row
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"update \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.survey_module\` sm
+set sm.question_count=num_questions from
+(select count(distinct cr.concept_id_1) num_questions, cr.concept_id_2 as survey_concept_id from
+`${BQ_PROJECT}.${BQ_DATASET}.concept_relationship\` cr
```
biopete: Should this be Workbench_Project and dataset? This code would have to run after concept and concept_relationship is imported.  May not matter.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -209,33 +209,80 @@ case when count_value > 0 then round(count_value/$person_count, 2)
...
+  join `${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.survey_module\` sm2
+    on cr.concept_id_2 = sm2.concept_id
+where cr.relationship_id = 'Has Module'
```
biopete: I'm looking how we solved this before but the question counts you end up with are inflated due to data. We need to see what the final data  is but joining on question concept where count_value > 0 helps some. I think add that here and we figure out later on prod data what we need to fix 

select count(distinct cr.concept_id_1) num_questions, cr.concept_id_2 as survey_concept_id from
`all-of-us-workbench-test.cdr20180921.concept_relationship` cr
  join `all-of-us-workbench-test.cdr20180921.survey_module` sm2 
    on cr.concept_id_2 = sm2.concept_id 
  join `all-of-us-workbench-test.cdr20180921.concept` qc 
    on cr.concept_id_1 = qc.concept_id 
where cr.relationship_id = 'Has Module' and (qc.count_value > 0 ) 
  group by survey_concept_id

biopete: Also, this query would need to move to after get the question counts below.

api/db-cdr/changelog-schema/db.changelog-17.xml
```diff
@@ -0,0 +1,54 @@
...
+            <column name="name" type="VARCHAR(30)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="description" type="VARCHAR(4000)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="all_concept_count" type="INTEGER">
+                <constraints nullable="true"/>
+            </column>
+            <column name="standard_concept_count" type="INTEGER">
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/changelog-schema/db.changelog-17.xml
```diff
@@ -0,0 +1,54 @@
...
+            </column>
+            <column name="participant_count" type="INTEGER">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+        <createTable tableName="survey_module">
+            <column name="concept_id" type="INTEGER">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="name" type="VARCHAR(30)">
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/changelog-schema/db.changelog-17.xml
```diff
@@ -0,0 +1,54 @@
...
+                <constraints nullable="true"/>
+            </column>
+            <column name="description" type="VARCHAR(4000)">
+                <constraints nullable="true"/>
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/changelog-schema/db.changelog-17.xml
```diff
@@ -0,0 +1,54 @@
...
+            </column>
+            <column name="question_count" type="INTEGER">
+                <constraints nullable="true"/>
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/generate-cdr/bq-schemas/domain_info.json
```diff
@@ -0,0 +1,42 @@
...
+    "name": "name",
+    "type": "STRING"
+  },
+  {
+    "mode": "REQUIRED",
+    "name": "description",
+    "type": "STRING"
+  },
+  {
+    "mode":"NULLABLE",
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/generate-cdr/csv/domain_info.csv
```diff
@@ -0,0 +1,5 @@
+"concept_id","domain","domain_id","name","description","all_concept_count","standard_concept_count","participant_count"
+19,0,"Condition","Conditions","Conditions are records of a Person suggesting the presence of a disease or medical condition stated as a diagnosis, a sign or a symptom, which is either observed by a Provider or reported by the patient.",0,0,0
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: I don't think it matters for this.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: Done.

api/src/test/resources/application.properties
```diff
@@ -4,3 +4,10 @@ spring.datasource.username=sa
 spring.datasource.password=sa
 spring.jpa.hibernate.ddl-auto=create-drop
 spring.jpa.properties.hibernate.dialect=org.pmiops.workbench.cdr.CommonTestDialect
+
+# Uncomment the following to turn on full SQL debugging
+logging.level.org.hibernate.SQL=DEBUG
```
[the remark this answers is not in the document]
danqrodney: No, reverted.

common-api/src/main/java/org/pmiops/workbench/config/SurveyModulesConfig.java
```diff
@@ -0,0 +1,30 @@
+package org.pmiops.workbench.config;
+
+import java.util.List;
+
+public class SurveyModulesConfig {
```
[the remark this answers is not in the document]
danqrodney: I wrote this when I was experimenting with having survey module info represented as JSON configuration instead of rows in dedicated tables. I abandoned that in this PR (mostly because I want to be able to update the counts for survey modules using SQL.) Anyway, we can revisit this later when we have a better story around PPI; for now, deleting this.

common-api/src/main/resources/common_api.yaml
```diff
@@ -71,6 +76,66 @@ definitions:
         format: int64
         description: Milliseconds since the UNIX epoch.
 
+  DomainInfo:
```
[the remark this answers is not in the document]
danqrodney: It is not. Everything in common-api gets merged into workbench.yaml at build time, but it's not duplicated there by design.

api/db-cdr/generate-cdr/make-bq-data.sh
biopete: Cool. Yeah i think this is good. Our api code does some stuff i think when it getsSurveyQuestions ... We just show the array size as the qustion count. Ill see how to make this number accurate later.

danqrodney:
I updated cdr_versions_test.json so it points at cdr20180921 and public20180921. And I copied over domain_info and survey_module into cdr20180831 and public20180920 so that the other CDR version should still work. I'm going to go ahead and merge.

Thanks for the review!

PR #1212, opened 2018-09-19
[RW-854][Risk=low] Adding recent work scroll component to homepage

UrsaStutsman:
This is a redone branch of 854, due to me doing something funky apparently during a rebase. Please look at https://github.com/all-of-us/workbench/pull/1189 to see original PR


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -141,7 +144,10 @@ public void setDistinctWorkspaceLimit(int limit) {
         })
         .forEach(userRecentResource -> {
           RecentResource resource = TO_CLIENT.apply(userRecentResource);
-          resource.setPermission(workspaceAccessMap.get(userRecentResource.getWorkspaceId()));
+          WorkspaceResponse workspaceDetails = workspaceAccessMap.get(userRecentResource.getWorkspaceId());
+          resource.setPermission(workspaceDetails.getAccessLevel());
+          resource.setWorkspaceNamespace(workspaceDetails.getWorkspace().getNamespace());
+          resource.setWorkspaceFirecloudName(workspaceDetails.getWorkspace().getName());
```
rushtong: Another way to do this would be to have another `Function` object that takes a `WorkspaceResponse` and also returns a `RecentResource`, then you could chain the two together with `andThen` instead of doing all of the resource.set* calls inside the loop. No need to implement that for this PR, just wanted to point out an alternative approach.

ui/src/app/views/recent-work/component.html
```diff
@@ -0,0 +1,20 @@
+<div *ngIf="elementVisible()" class="recent-work" style="display: inline-block">
+  <div *ngIf="leftScrollVisible()" class="scroll-indicator" id="left-scroll"
+       (click)="moveDownList()">
```
rushtong: Nit: The naming here threw me. the div is all about left scrolling, but the click handler is not named similarly. `scrollLeft()` seems like a better handler name to me.

Same feedback for the name of `moveUpList()`

UrsaStutsman:
screen shot 2018-09-19 at 11 59 04 am
screen shot 2018-09-19 at 11 59 15 am
screen shot 2018-09-19 at 11 59 27 am
screen shot 2018-09-19 at 11 59 36 am
screen shot 2018-09-19 at 11 59 46 am

UrsaStutsman:
@lstrano-broad @cukarthik Attached screenshots of what this looks like. Be aware if user has no cache, nothing will display, if cache is <=3 neither scroll arrow will appear. Any questions don't hesitate.

PR #1214, opened 2018-09-19
[RW-1416][RW-1355][risk=low] Rename CDRs, fix prod CDR config

calbach:
- Fix prod CDR configs to point to an actual CDR that's available in prod (it's currently broken)
- Rename to "Dataset" to avoid the opaque "CDR" acronym
- Require an "id" in CDR JSONs; use this as the primary key
-> I prefer this to introducing a new ID type

### DIFF
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "Test Registered CDR",
+    "name": "Synthetic Dataset v1",
     "dataAccessLevel": 1,
     "bigqueryProject": "all-of-us-ehr-dev",
     "bigqueryDataset": "test_merge_dec26",
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -7,7 +7,7 @@
     "creationTime": "2017-12-26 00:00:00Z",
     "releaseNumber": 1,
     "numParticipants": 946237,
-    "cdrDbName": "cdr20180831",
-    "publicDbName": "public20180831"
+    "cdrDbName": "cdr20180817",
+    "publicDbName": "public20180817"
   }
 ]
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -1,5 +1,6 @@
 [
   {
+    "cdrVersionId": 1,
     "name": "Synthetic Dataset v1",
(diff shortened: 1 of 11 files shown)


danqrodney:
I have no objections to the naming convention.

We will need to hack our database if we want to do a rename of the CDR version without breaking things, though.

danqrodney:
Note: I could imagine us adding some other external ID to CDR versions so that we can rename them in future without hacking the DB.

calbach:
When I sent this PR that was the plan, but since we got feedback during the demo it sounded like that was less solid. In any case I'm now tracking this via https://precisionmedicineinitiative.atlassian.net/browse/RW-1355

api/config/cdr_versions_stable.json
```diff
@@ -11,7 +11,7 @@
     "publicDbName": "public20180831"
   },
   {
-    "name": "Test Registered CDR 20180920",
+    "name": "Synthetic Dataset v2",
```
calbach: @danqrodney this one isn't "releaseNumber=2", should it be? I don't have context on this field.

danqrodney: Release number is there as an ordering of releases; in future we expect to have both registered and controlled CDR versions with the same release number, released around the same time. For now it doesn't mean a whole lot.

calbach: Got it. Is there a reason test is different from local/staging/stable here? (test sets releaseNumber=2)

api/config/cdr_versions_prod.json
```diff
@@ -1,13 +1,13 @@
...
+    "name": "Synthetic Dataset v1",
     "dataAccessLevel": 1,
     "bigqueryProject": "all-of-us-ehr-dev",
     "bigqueryDataset": "test_merge_dec26",
     "creationTime": "2017-12-26 00:00:00Z",
     "releaseNumber": 1,
     "numParticipants": 946237,
-    "cdrDbName": "cdr20180831",
-    "publicDbName": "public20180831"
+    "cdrDbName": "cdr20180817",
```
calbach: 20180831 doesn't exist in prod

calbach:
Updated: I deviated from your recommendation by just using the primary DB ID here. I don't see a good reason to introduce another ID concept since we're effectively just mirroring this JSON in MySQL anyways. Let me know if you have opinions on this.

PR #1216, opened 2018-09-20
[RW-1299][Risk=no] Refactoring NotebookList and Cohort List to use ResourceCard

UrsaStutsman:
Ready for PR Review


rushtong:
Seems reasonable to me at this point - no real feedback on approach.

ui/src/app/utils/resourceActions.ts
```diff
@@ -41,4 +54,40 @@ export const cohortActionList = [
 
 export const resourceActionList =  notebookActionList.concat(cohortActionList);
 
+export function convertToResources(list: any[], component): RecentResource[] {
+  const resourceList = [];
+  if (component.resourceType === ResourceType.NOTEBOOK) {
```
blrubenstein: Do you access anything from component other than `resourceType`? Can we just pass in resource type?

ui/src/app/views/rename-modal/component.html
```diff
@@ -14,7 +14,7 @@
         <button type="button" class="btn btn-outline" (click)="renaming=false">Cancel</button>
         <button type="button" class="btn btn-primary confirm-rename-btn" id="rename"
                 (click)="this.emitRename(resource);" [clrLoading]="loading" [disabled]="loading">
-            Rename {{resourceType}}
+            Rename Notebook
```
blrubenstein: If we are making this notebook specific, we should probably rename the directory and component.

ui/src/app/utils/resourceActions.ts
UrsaStutsman: yup this.workspaceName, this.workspace, etc, in convertToResource...

ui/src/app/utils/resourceActions.ts
```diff
@@ -41,4 +54,40 @@ export const cohortActionList = [
...
+  if (resource.lastModifiedTime === undefined) {
+    mTime = new Date().toDateString();
+  } else {
+    mTime = resource.lastModifiedTime.toString();
+  }
+  const newResource: RecentResource = {
+    workspaceNamespace: component.wsNamespace,
+    workspaceFirecloudName: component.wsId,
+    permission: WorkspaceAccessLevel[component.accessLevel],
+    modifiedTime: mTime
```
calbach: Sidenote: Not for this PR but we really need to rename this property

ui/src/app/utils/resourceActions.ts
```diff
@@ -41,4 +54,40 @@ export const cohortActionList = [
 
 export const resourceActionList =  notebookActionList.concat(cohortActionList);
 
+export function convertToResources(list: any[], component): RecentResource[] {
```
calbach: types

calbach: I'm guessing you might sort this list later in the components, but this function would be better if it preserved the order of the input list

ui/src/app/utils/resourceActions.ts
calbach: opt: might be cleaner as a `return list.map(...)`

ui/src/app/utils/resourceActions.ts
```diff
@@ -41,4 +54,40 @@ export const cohortActionList = [
...
+    return resourceList;
+  } else if (component.resourceType === ResourceType.COHORT) {
+    for (const cohort of list) {
+      resourceList.push(this.convertToResource(cohort, component));
+    }
+    return resourceList;
+  }
+}
+
+export function convertToResource(resource: any, component): RecentResource {
```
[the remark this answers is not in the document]
UrsaStutsman: I tried this and it would not compile because when assigning resource to notebook or cohort it complained. I'll keep trying.

ui/src/app/views/cohort-list/component.html
```diff
@@ -13,58 +13,9 @@ <h3 class="cohorts-header">Cohorts</h3>
...
-        </div>
-      </div>
-      <div class="description">{{cohort.description}}</div>
-    </div>
-    <div class="footer">
-      <div class="last-modified">Last Updated <br/>{{cohort.lastModifiedTime | date}}</div>
-      <div class="item-type cohort"><div>CH</div></div>
-    </div>
+  <div *ngFor="let cohort of resourceList">
+    <app-resource-card [resourceCard]="cohort" [forList]="'for-list'"
```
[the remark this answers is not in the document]
UrsaStutsman: css

ui/src/testing/stubs/workspace-service-stub.ts
```diff
@@ -57,6 +60,7 @@ export class WorkspacesServiceStub {
     }
   ];
   notebookList: FileDetail[];
+  resourceList: RecentResource[];
```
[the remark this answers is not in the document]
UrsaStutsman: Yup L331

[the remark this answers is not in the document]
calbach: I saw that you were setting it in this file - but I still don't see why it would be needed in WorkspaceServiceStub nor do I see it being read anywhere

ui/src/app/views/notebook-list/component.spec.ts
```diff
@@ -137,27 +139,28 @@ describe('NotebookListComponent', () => {
...
-    simulateClick(fixture, de.query(By.css('.notebook-menu')));
-    simulateClick(fixture, de.query(By.css('#rename-mockFile')));
+    simulateClick(fixture, de.query(By.css('.resource-menu')));
+    simulateClick(fixture, de.query(By.css('.pencil')));
     simulateInput(fixture, de.query(By.css('#new-name')), 'mockFile');
     simulateClick(fixture, de.query(By.css('button#rename')));
     updateAndTick(fixture);
     tick();
     const errorMessage = de.queryAll(By.css('.modal-title'));
+    console.log(errorMessage);
```
blrubenstein: Nit: Remove console log message.

ui/src/app/views/cohort-list/component.html
```diff
@@ -13,58 +13,9 @@ <h3 class="cohorts-header">Cohorts</h3>
...
-        </div>
-      </div>
-      <div class="description">{{cohort.description}}</div>
-    </div>
-    <div class="footer">
-      <div class="last-modified">Last Updated <br/>{{cohort.lastModifiedTime | date}}</div>
-      <div class="item-type cohort"><div>CH</div></div>
-    </div>
+  <div *ngFor="let cohort of resourceList">
+    <app-resource-card [resourceCard]="cohort" [cssClass]="'for-list'"
```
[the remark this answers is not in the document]
calbach: To summarize offline Slack discussion, this parameter was to allow control for different styling on the homepage vs notebook list. I suggested to follow-up with Lou and verify these differences were intentional, since the card styling look pretty nearly identical to me in the latest version of the mocks. If they are meant to be the same, then we don't need this parameter.

If the styling is intentionally different, I would suggest switching to something more self-documenting, e.g. `[compactStyle]="true"`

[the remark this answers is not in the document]
UrsaStutsman: So the css I'm using matches what was already there. Thus needing the two different styles. The update to match current mocks is out of scope for this card. I can change the name.

PR #1217, opened 2018-09-20
[RW-1289][risk=no] Fix z-index loading screen rendering problems

calbach:
Ultimately I think this may have come down to a browser rendering issue, AFAICT the HTML/CSS stacking order defined did not match what was being rendered in the browser - my interpretation is that a correct rendering would have placed the spinner on top of everything (including the top-nav). Toggling various CSS properties on/off (and returning to the original CSS state) in chrome inspector resulted in things reaching the correct state. I remain perplexed as to why the browser rendered the top-nav on top of the loading screen at all, it seems by sheer coincidence this happened to also be the desired behavior.

Ultimately to solve this I made two changes:
- Switch from opacity-based visibility to "display", this seems to be more obvious signal for the browser to redraw (rather than scaling the opacity + z-index change).
- Move the spinner into the signed in page - shifting the spinner below the top-nav. AFAIK we didn't have a use case for this spinner in the signed out state.

### DIFF
--- a/ui/src/app/views/app/component.html
+++ b/ui/src/app/views/app/component.html
@@ -2,4 +2,3 @@
 <div *ngIf="overriddenUrl" style="position: absolute; top: 0; left: 1rem;">
   <span style="font-size: 80%; color: darkred;">API URL: {{this.overriddenUrl}}</span>
 </div>
-<app-routing-spinner></app-routing-spinner>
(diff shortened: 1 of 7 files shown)


blrubenstein:
Tests should be fixed by adding the routing spinner to the `SignedInComponent` test.

PR #1221, opened 2018-09-24
Cdr version swtich changes [risk=no]

SrushtiGangireddy:
1. Changed the cdr version with the new database built.

### DIFF
--- a/api/config/cdr_versions_stable.json
+++ b/api/config/cdr_versions_stable.json
@@ -9,5 +9,16 @@
     "numParticipants": 946237,
     "cdrDbName": "cdr20180831",
     "publicDbName": "public20180831"
+  },
+  {
+    "name": "Test Registered CDR 20180920",
+    "dataAccessLevel": 1,
+    "bigqueryProject": "all-of-us-ehr-dev",
+    "bigqueryDataset": "test_merge_dec26",
+    "creationTime": "2017-12-26 00:00:00Z",
+    "releaseNumber": 1,
+    "numParticipants": 946237,
+    "cdrDbName": "cdr20180920",
+    "publicDbName": "public20180920"
   }
 ]
--- a/api/config/cdr_versions_test.json
+++ b/api/config/cdr_versions_test.json
@@ -11,7 +11,7 @@
     "publicDbName": "public20180920"
   },
   {
-    "name": "Test Registered CDR 20180920",
+    "name": "New Test Registered CDR",
     "dataAccessLevel": 1,
     "bigqueryProject": "all-of-us-ehr-dev",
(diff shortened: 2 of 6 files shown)


api/config/cdr_versions_test.json
```diff
@@ -11,7 +11,7 @@
     "publicDbName": "public20180920"
   },
   {
-    "name": "New Test Registered CDR",
```
danqrodney: Note that doing this should cause the old CDR version row with this name to get deleted... please confirm that works after the push.

api/config/cdr_versions_stable.json
```diff
@@ -9,5 +9,16 @@
     "numParticipants": 946237,
     "cdrDbName": "cdr20180831",
     "publicDbName": "public20180831"
+  },
+  {
+    "name": "Test Registered CDR 20180920",
```
biopete: This name is a key. It's a better name than the old one .. But question is What happens when we switch the name ? Basically what you did is delete the cdr that was being used since friday in workbench and add this new one. 

@danqrodney What will happen to my workspace in test that i made with that old cdr on friday in workbench? 

We will find out. It is a good thing to test and see what happens. So i approve this name change .

api/config/cdr_versions_test.json
SrushtiGangireddy: Changed it back. Just thought it would be good if it is consistent.

api/config/cdr_versions_stable.json
danqrodney: I believe it will fail with an error, here:

https://github.com/all-of-us/workbench/blob/master/api/tools/src/main/java/org/pmiops/workbench/tools/UpdateCdrVersions.java#L80

Which I guess will probably stop the deploy. Given that we know this is going to happen, perhaps we should not remove the old CDR version yet and change the code to be robust to this happening.

(Not deleting CDR versions with existing workspaces seems correct, but the tool should probably just log a warning and continue.)

api/config/cdr_versions_test.json
biopete: Darn. I wanted to see what broke :)

danqrodney: OK. We don't support renaming CDR versions right now. For the time being, I think it's OK to have this inconsistency.

danqrodney: (BTW, it's easy enough to hack the DB and then change the file to match in a follow-up.)

api/config/cdr_versions_stable.json
SrushtiGangireddy: @danqrodney  Seems reasonable not to let deploy fail when we know its going to happen. So, we probably can let the test one be the same.

PR #1222, opened 2018-09-24
[risk=no] Fix deploy typo

calbach:


calbach:
Tests don't cover this file, so not going to wait on them

PR #1229, opened 2018-09-25
Biopete/measurement graphs [risk=no]

biopete:
tslint ignore publicGenerated in ui .angular-cli.json

### DIFF
--- a/public-ui/.angular-cli.json
+++ b/public-ui/.angular-cli.json
@@ -45,6 +45,7 @@
       "project": "src/tsconfig.app.json",
       "exclude": [
         "**/src/generated/**/*.ts",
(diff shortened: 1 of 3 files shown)


biopete:
I had no trouble linting pushing after devup in ui. Deleting my generated and testing final.

calbach:
Would prefer people just remove these stale generated directories: `rm -rf ui/src/publicGenerated`

The current codebase is already self-consistent: doesn't generate publicGenerated and doesn't reference it in .angular-cli.json (likewise for public-ui and notebooks-generated). The transition was just slightly awkward.

biopete:
Gotcha. Thanks CH.

PR #1232, opened 2018-09-25
[RW-1344][risk=low] Support deployment dry runs, fix prod deploy

calbach:

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -49,7 +49,8 @@ ENVIRONMENTS = {
 }
 
 def run_inline_or_log(dry_run, args)
-  common.run_inline(DRY_RUN_CMD + args)
+  cmd_prefix = dry_run ? DRY_RUN_CMD : []
+  Common.new.run_inline(cmd_prefix + args)
 end
 
 def get_config(project)
@@ -1293,7 +1294,7 @@ def migrate_database(dry_run = false)
   common = Common.new
   common.status "Migrating main database..."
   Dir.chdir("db") do
-    run_inline_or_log(op.opts.dry_run, %W{gradle --info update -PrunList=main})
+    run_inline_or_log(dry_run, %W{gradle --info update -PrunList=main})
   end
 end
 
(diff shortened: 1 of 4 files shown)


api/libproject/devstart.rb
```diff
@@ -1120,14 +1121,27 @@ def deploy_gcs_artifacts(cmd_name, args)
...
+    "--dry-run",
+    ->(opts, _) { opts.dry_run = true},
+    "Don't actually push, just log the command lines which would be " +
+    "executed on a real invocation."
+  )
   gcc = GcloudContextV2.new(op)
   op.parse.validate
   gcc.validate
-  common.run_inline %W{gsutil cp scripts/setup_notebook_cluster.sh gs://#{gcc.project}-scripts/setup_notebook_cluster.sh}
+  cmd_prefix = op.opts.dry_run ? DRY_RUN_CMD : []
```
danqrodney: I'm wondering if it would be cleaner to have a method like common.run_inline that takes a command and dry_run as a parameter... and just call that everywhere instead of repeating this?

deploy/README.md
```diff
@@ -51,3 +51,21 @@ up volume file permissions for the circleci UID (specified in our Dockerfile).
 **Note**: Local changes to deploy.rb or devstart.rb are **not** currently picked
 up by the deployment process; it uses the tools code from the commit version
 specified for deployment.
+
+# Manual testing
+
+Note: See above for an explanation of why this process is special right now.
+This could be improved in the future by supporting builds via dirtly clients.
```
danqrodney: dirtly -> dirty?

deploy/README.md
```diff
@@ -51,3 +51,21 @@ up volume file permissions for the circleci UID (specified in our Dockerfile).
...
+
+Use the --dry-run flag to test changes. To ensure your local changes are picked
+up during the deploy process, follow this process:
+
+1. Commit your changes to the deploy scripts
+1. Push your branch to GitHub, e.g. 'ch/deploy-fix'
+1. Use --dry-run and --git-version to build using your branch:
```
danqrodney: So the point of dry-run is just to check to make sure your script is issuing the commands you expected?

I can see how that could be useful for quick iteration on script changes, but you probably won't have confidence that everything works as intended until you actually run it.

deploy/libproject/deploy.rb
```diff
@@ -181,10 +191,18 @@ def deploy(cmd_name, args)
 
   op.parse.validate
 
+  if op.opts.account == "#{TEST_PROJECT}@appspot.gserviceaccount.com"
+    raise ArgumentError.new(
```
danqrodney: Why is this account incompatible?

deploy/README.md
calbach: Good point, I've added instructions here as well for if you want to do a full test deployment (--no-promote)

deploy/libproject/deploy.rb
calbach: Added a comment. tl;dr: fixable but probably not worth bothering

api/libproject/devstart.rb
calbach: Done

deploy/README.md
calbach: Done

PR #1234, opened 2018-09-26
[RW-1312][risk=moderate] API support for optionally cloning with collaborators

calbach:
Unfortunately this requires adding another phase to the clone - a call to FC to update the ACL. This is eventually something we should request as a feature from the workspace service.

- Also, fix changing CDR version on clone.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -584,8 +584,8 @@ public class WorkspacesController implements WorkspacesApiDelegate {
           .getWorkspace()
           .getBucketName();
 
-    org.pmiops.workbench.db.model.Workspace fromWorkspace = workspaceService.getRequiredWithCohorts(
-        workspaceNamespace, workspaceId);
+    org.pmiops.workbench.db.model.Workspace fromWorkspace =
+        workspaceService.getRequiredWithCohorts(workspaceNamespace, workspaceId);
     if (fromWorkspace == null) {
       throw new NotFoundException(String.format(
           "Workspace %s/%s not found", workspaceNamespace, workspaceId));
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -655,17 +655,31 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    WorkspaceUserRole ownerRole = new WorkspaceUserRole();
+    ownerRole.setRoleEnum(WorkspaceAccessLevel.OWNER);
+    ownerRole.setWorkspace(dbWorkspace);
+    ownerRole.setUser(user);
+
+    dbWorkspace.addWorkspaceUserRole(ownerRole);
+    org.pmiops.workbench.db.model.Workspace savedWorkspace =
+        workspaceService.saveAndCloneCohortsAndConceptSets(fromWorkspace, dbWorkspace);
+
+    if (Boolean.TRUE.equals(body.getIncludeUserRoles())) {
```
blrubenstein: Can this just be: `if (body.getIncludeUserRoles()) {`

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1024,6 +1034,51 @@ public void testCloneWorkspaceDifferentOwner() throws Exception {
     assertThat(workspace2.getUserRoles().get(0).getEmail()).isEqualTo(cloner.getEmail());
   }
 
+  @Test
+  public void testCloneWorkspaceIncludeUserRoles() throws Exception {
+    User cloner = createUser("cloner@gmail.com");
+    createUser("reader@gmail.com");
```
blrubenstein: Can we abstract these strings to be reusable variables rather than hardcoding the string multiple times.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
calbach: We need to tolerate null. Other options include:

- body.getIncludeUserRoles() != null && body.getIncludeUserRoles()
- Optional.ofNullabel(body.getIncludeUserRoles()).orElse(false)

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
calbach: Done

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: Lets do the `Optional.ofNullable()` then? Just to make it more clear why we're doing it this way?

calbach: Done

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -655,17 +659,31 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     dbWorkspace.setCdrVersion(fromWorkspace.getCdrVersion());
```
danqrodney: I thought you were able to override the CDR version of the workspace when cloning -- you said you'd tested this -- but this logic doesn't seem to allow for it. Am I missing the part where you change it?

Note that if you *do* change the CDR version to a version with a more restrictive data access level, ideally we would kick out any collaborators that don't have that data access level... they won't be able to do anything with it. (We should ideally warn the user about this if they request to clone collaborators and this happens.)

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -260,19 +261,21 @@ private static String generateRandomChars(String candidateChars, int length) {
...
-          throw new BadRequestException(
-              String.format("CDR version with ID %s not found", cdrVersionId));
-        }
-        dbWorkspace.setCdrVersion(cdrVersion);
-      } catch (NumberFormatException e) {
-        throw new BadRequestException(String.format(
-            "Invalid cdr version ID: %s", cdrVersionId));
+  private CdrVersion setCdrVersionId(org.pmiops.workbench.db.model.Workspace dbWorkspace, String cdrVersionId) {
+    if (Strings.isNullOrEmpty(cdrVersionId)) {
+      return null;
```
danqrodney: This should throw an exception, yes?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -645,27 +648,47 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-    resp.setWorkspace(TO_SINGLE_CLIENT_WORKSPACE_FROM_FC_AND_DB.apply(dbWorkspace, toFcWorkspace));
-    return ResponseEntity.ok(resp);
+    String reqCdrVersionId = body.getWorkspace().getCdrVersionId();
+    if (Strings.isNullOrEmpty(reqCdrVersionId) ||
+        reqCdrVersionId.equals(Long.toString(fromWorkspace.getCdrVersion().getCdrVersionId()))) {
+      dbWorkspace.setCdrVersion(fromWorkspace.getCdrVersion());
+      dbWorkspace.setDataAccessLevel(fromWorkspace.getDataAccessLevel());
+    } else {
+      CdrVersion reqCdrVersion = setCdrVersionId(dbWorkspace, reqCdrVersionId);
+      dbWorkspace.setDataAccessLevelEnum(reqCdrVersion.getDataAccessLevelEnum());
```
danqrodney: Seems like setCdrVersionId() could do this and not return anything... OTOH, as you noted, I'm not sure what the point of denormalizing this is.... maybe a slight performance win?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -645,27 +648,47 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    }
+
+    WorkspaceUserRole ownerRole = new WorkspaceUserRole();
+    ownerRole.setRoleEnum(WorkspaceAccessLevel.OWNER);
+    ownerRole.setWorkspace(dbWorkspace);
+    ownerRole.setUser(user);
+
+    dbWorkspace.addWorkspaceUserRole(ownerRole);
+    org.pmiops.workbench.db.model.Workspace savedWorkspace =
+        workspaceService.saveAndCloneCohortsAndConceptSets(fromWorkspace, dbWorkspace);
```
danqrodney: It'd be nice if this could be done in a Transactional block with the workspace creation, just to make sure everything gets committed at once... but I can't remember if that actually doesn't work for some reason. Anyway, not a huge deal.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
danqrodney: We should probably file an issue on the second point then so we don't forget about it.

PR #1235, opened 2018-09-26
[RW-1312][risk=no] Add checkbox for cloning collaborators

calbach:
Depends on https://github.com/all-of-us/workbench/pull/1234

Note: due to the 60% page-width consistency on the clone page, this will wrap to the next line on most screens. Not sure if that affects the UX and whether we want more spacing beneath

Screenshots:
image
image

### DIFF
--- a/ui/src/app/views/workspace-edit/component.css
+++ b/ui/src/app/views/workspace-edit/component.css
@@ -34,8 +34,9 @@
   width: 100%;
 }
 
-.name-area {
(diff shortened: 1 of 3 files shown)


calbach:
@lstrano-broad PTAL at the wrapping above and let me know if you have suggestions on handling it differently.

blrubenstein:
Has Lou gotten back to you on this yet?

PR #1236, opened 2018-09-26
[RW-1280][RW-1278] Blrubenstein/concept search homepage 

blrubenstein:
Tests are in progress right now.


ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,83 @@
...
+  </div>
+  <ng-template #chartBox>
+    <div class="chart-box lower-box">
+      <div class="chart-selector-tabs">
+        <div *ngFor="let domain of conceptDomainList" class="chart-selector-tab">
+          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading">
+            <div class="chart-selector-header">
+              {{domain.name}}
+            </div>
+            <div>{{domain.allConceptCount}}</div>
```
biopete: Does this need  to be based on standrd check box: 
{{ standardConceptsOnly === true ? domain.standardConceptCount : domain.allConceptCount }}

ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,83 @@
...
+          </button>
+        </div>
+      </div>
+      <div class="filter-options">
+        Filter by vocabulary:
+        <clr-checkbox class="filter-checkbox" [clrChecked]="standardConceptsOnly">
```
biopete: I see this is stub out code here. Maybe put  TODO in a comment.

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -0,0 +1,152 @@
...
+    this.selectedDomain = domain.domain;
+    this.searchConcepts();
+  }
+
+  searchConcepts() {
+    console.log(this.standardConceptsOnly);
+    this.searching = true;
+    this.searchLoading = true;
+    let standardConceptFilter: StandardConceptFilter;
+    if (this.standardConceptsOnly) {
```
biopete: I think at this point in development,  i might search all concepts all the time and filter on standard concepts in the ui  . Like in setConcepts() below , filter on your checkbox there.

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  get getSearchDisabled() {
+    return this.searchLoading || this.loadingDomains || this.searchTermNotLongEnough;
+  }
+
+  get searchTermNotLongEnough() {
+    return this.searchTerm === undefined || this.searchTerm.length < 3;
+  }
+
+  setStandardOnly() {
```
biopete: Is this used? Stub out code?  I can see how you would use it to reload the search results from cache or launch a new search

ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,83 @@
...
+    <span role="tooltip" aria-haspopup="true" class="tooltip tooltip-md tooltip-top-right btn-search-container">
+      <button class="btn btn-secondary btn-search" (click)="searchButton()" [disabled]="getSearchDisabled">
+        Search
+        <span style="padding-left: 0.5rem; display: flex;" *ngIf="searchLoading"><span class="spinner spinner-sm"></span></span>
+      </button>
+      <span *ngIf="searchTermNotLongEnough" class="tooltip-content">Search must be at least three letters long</span>
+    </span>
+    <clr-checkbox class="standard-concepts-checkbox" [(clrChecked)]="standardConceptsOnly">
+      Standard concepts only
+    </clr-checkbox>
```
rushtong: I have a UX philosophical objection to using a checkbox as a search button which is what this effectively does (see bullet 2 in "When to Use Which Widget" here https://www.nngroup.com/articles/checkboxes-vs-radio-buttons/). However, I wouldn't hold this PR up on that. I think we should consider revisiting this at some point.

ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,83 @@
...
+      <span class="spinner spinner-lg"></span>
+    </div>
+    <ng-template #domains>
+      <div class="card item-card" *ngFor="let domain of conceptDomainList">
+        <div class="concept-name">
+          {{domain.name}}
+        </div>
+        <div class="concept-text">
+          <div>
+            <span class="concept-count-text" *ngIf="!standardConceptsOnly else standardCount">{{domain.allConceptCount}}</span>
```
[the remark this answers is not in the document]
blrubenstein: Good catch, that totally makes sense. Will do!

ui/src/app/views/concept-homepage/component.html
blrubenstein: This is a placeholder, will add a todo though. (Should be counts on the search, not total counts.

ui/src/app/views/concept-homepage/component.html
blrubenstein: Will add! (Or maybe just remove this for now)

ui/src/app/views/concept-homepage/component.ts
blrubenstein: I don't want to do that because we cap the search at 100 results, so I don't want it to only get non-standard concepts, and then the user is left seeing no results.

ui/src/app/views/concept-homepage/component.ts
blrubenstein: This is old from previous work. Can remove.

ui/src/app/views/concept-homepage/component.html
blrubenstein: Talked offline, this is going to be a followup with Karthik and Lou. Currently not searching on select/deselect though because it adds complexity.

ui/src/app/views/concept-homepage/component.html
blrubenstein: That will change the numbers on the cards. These only show up if you have searched anyway

blrubenstein:
Per conversation yesterday, I am going to merge this in shortly, but follow up with tests. I am going to comment out the nav button before merging though, so we don't accidentally push an incomplete feature to stable before UX testing.

ui/src/app/views/concept-add-modal/component.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.modalOpen = true;
+    this.loading = false;
+  }
+
+  close(): void {
+    this.modalOpen = false;
+  }
+
+  createConceptSet(): void {
+    // TODO: Implement
```
rushtong: Perhaps reference https://precisionmedicineinitiative.atlassian.net/browse/RW-1282 in either the PR description or this TODO.

ui/src/app/views/concept-homepage/component.html
rushtong: 👍

ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,81 @@
...
+  </div>
+  <ng-template #chartBox>
+    <div class="chart-box lower-box">
+      <div class="chart-selector-tabs">
+        <div *ngFor="let domain of conceptDomainList" class="chart-selector-tab">
+          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading">
+            <div class="chart-selector-header">
+              {{domain.name}}
+            </div>
+            <!-- TODO: Make this the counts returned from search. -->
```
rushtong: Is this https://precisionmedicineinitiative.atlassian.net/browse/RW-1284 ? If so, please reference in the PR or the TODO.

ui/src/app/views/concept-homepage/component.html
```diff
@@ -0,0 +1,81 @@
...
+        <!--<clr-checkbox class="filter-checkbox" [clrChecked]="standardConceptsOnly">-->
+          <!--SNOMED (4011)-->
+        <!--</clr-checkbox>-->
+        <!--<clr-checkbox class="filter-checkbox" [clrChecked]="standardConceptsOnly">-->
+          <!--CPT (2951)-->
+        <!--</clr-checkbox>-->
+        <!--<clr-checkbox class="filter-checkbox" [clrChecked]="standardConceptsOnly">-->
+          <!--ICD10 (2112)-->
+        <!--</clr-checkbox>-->
+      <!--</div>-->
```
rushtong: Remove?

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      ],
+      providers: [
+        { provide: ConceptsService, useValue: new ConceptsServiceStub() },
+        { provide: ConceptSetsService, useValue: new ConceptSetsServiceStub() },
+        { provide: ActivatedRoute, useValue: activatedRouteStub }
+      ]}).compileComponents().then(() => {
+        fixture = TestBed.createComponent(ConceptHomepageComponent);
+        tick();
+        tick();
+        tick();
```
rushtong: This looks annoying. Is there a programmatic need for 3 ticks or is it that this seems to work this way?

ui/src/app/views/concept-table/component.html
```diff
@@ -0,0 +1,15 @@
+<clr-datagrid [(clrDgSelected)]="selectedConcepts" [clrDgLoading]="loading">
+  <clr-dg-column>Name</clr-dg-column>
+  <clr-dg-column>Synonyms</clr-dg-column>
+  <clr-dg-column>Code</clr-dg-column>
+  <clr-dg-column>Vocabulary</clr-dg-column>
+  <clr-dg-column>Count</clr-dg-column>
+  <div class="break"></div>
+  <clr-dg-row *ngFor="let concept of concepts" [clrDgItem]="concept">
+    <clr-dg-cell>{{concept.conceptName}}</clr-dg-cell>
+    <clr-dg-cell>{{concept.conceptSynonyms.join(', ')}}</clr-dg-cell>
```
rushtong: This looks interesting in that we're assuming a particular form of concept synonym, i.e. they won't end in commas or have commas as part of the synonym. I think it's fine for this PR, but we should be conscious that synonym format might be problematic and we might need a better way to join them.

ui/src/app/views/concept-homepage/component.html
blrubenstein: Its actually this: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-1277

Will add.

ui/src/app/views/concept-homepage/component.html
blrubenstein: I want this because it has all the styling and functionality worked out? Just non-functional yet.

ui/src/app/views/concept-table/component.html
blrubenstein: 👍

PR #1246, opened 2018-09-27
[RW-1305][risk=no] Update prod domains

calbach:

### DIFF
--- a/api/config/config_prod.json
+++ b/api/config/config_prod.json
@@ -33,7 +33,7 @@
   },
   "admin": {
     "adminIdVerification": "workbench-id-verifications@researchallofus.org",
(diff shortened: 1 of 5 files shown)


api/config/config_prod.json
```diff
@@ -33,7 +33,7 @@
   },
   "admin": {
     "adminIdVerification": "workbench-id-verifications@researchallofus.org",
-    "loginUrl": "http:\/\/all-of-us-rw-prod.appspot.com/login"
+    "loginUrl": "http:\/\/workbench.researchallofus.org/login"
```
rushtong: https? the http link doesn't work for me.

calbach: Nice catch - seems like all of them should be using https (except local). Made this change across the board.

PR #1248, opened 2018-09-28
[RW-1325][Risk=no] Adding Scroll Component to Workspace About page

UrsaStutsman:
Adds the scroll component to workspace about page


ui/src/app/views/recent-work/component.ts
```diff
@@ -14,21 +23,57 @@ import {
...
+        .getCohortsInWorkspace(this.workspace.namespace, this.workspace.id);
+      Observable.forkJoin(notebookCall, cohortCall)
+        .subscribe(([notebooks, cohorts]) => {
+          const notebookResources = convertToResources(notebooks, this.workspace.namespace,
+            this.workspace.id, this.accessLevel, ResourceType.NOTEBOOK);
+          const cohortResources = convertToResources(cohorts.items, this.workspace.namespace,
+            this.workspace.id, this.accessLevel, ResourceType.COHORT);
+          this.fullList = notebookResources.concat(cohortResources);
+          this.fullList.sort((leftSide, rightSide): number => {
+            if (leftSide.modifiedTime < rightSide.modifiedTime) { return 1; }
```
calbach: Is there a ticket to rename this to lastAccessTime or lastAccessedTime? I remember commenting about it but I forgot the resolution.

ui/src/app/views/recent-work/component.ts
```diff
@@ -14,21 +23,57 @@ import {
...
+            if (leftSide.modifiedTime > rightSide.modifiedTime) { return -1; }
```
calbach: opt_nit: Personally I prefer either braces OR 1-line if statements without braces. Optional since we haven't really decided on a team style convention on this point.

ui/src/app/views/recent-work/component.ts
```diff
@@ -14,21 +23,57 @@ import {
...
     this.updateList();
   }
 
   updateList(): void {
-    this.userMetricsService.getUserRecentResources().subscribe((resources) => {
-      this.fullList = resources;
-      this.resourceList = this.fullList.slice(this.startIndex, this.startIndex + this.size);
-    });
+    if (this.workspace) {
+      this.cssClass = 'for-list';
```
calbach: I still don't think this style divergence is intentional in the mocks or necessary, do you mind filing a ticket to track?

ui/src/app/views/recent-work/component.ts
```diff
@@ -14,21 +23,57 @@ import {
...
+        .getNoteBookList(this.workspace.namespace, this.workspace.id);
+      const cohortCall = this.cohortsService
+        .getCohortsInWorkspace(this.workspace.namespace, this.workspace.id);
+      Observable.forkJoin(notebookCall, cohortCall)
+        .subscribe(([notebooks, cohorts]) => {
+          const notebookResources = convertToResources(notebooks, this.workspace.namespace,
+            this.workspace.id, this.accessLevel, ResourceType.NOTEBOOK);
+          const cohortResources = convertToResources(cohorts.items, this.workspace.namespace,
+            this.workspace.id, this.accessLevel, ResourceType.COHORT);
+          this.fullList = notebookResources.concat(cohortResources);
```
calbach: Confirming: did we want the scroll component to contain ALL resources, or just the most recent N?

ui/src/app/views/workspace/component.html
```diff
@@ -5,26 +5,34 @@
...
+          <div class="card add-card" [ngClass]="{'disabled' : !writePermission }" style="margin-top: 13px;" (click)="!writePermission || newNotebook()">
+            <div class="create-text">
+              Create a<br/>New Notebook
+              <clr-icon shape="plus-circle" class="plus-circle"></clr-icon>
+            </div>
           </div>
         </div>
       </div>
+      <div style="padding-left: 10%">
+        <app-recent-work [workspace]="workspace"></app-recent-work>
```
calbach: FYI(no action needed): It almost feels like the it would be cleaner if the input were a list of RecentResource, but since this scrolling component could potentially lazily fetch more data in the future, I think the approach you have is probably reasonable. Alternatively it could be some generic interface which gives you a list of RecentResources and allows you to request more.

ui/src/app/views/recent-work/component.ts
```diff
@@ -14,21 +23,57 @@ import {
 export class RecentWorkComponent implements OnInit {
   resourceList: RecentResource[];
   fullList: RecentResource[] = [];
+  @Input('workspace')
+  workspace: Workspace;
```
calbach: There should probably be a test case that hits this new workspace-input codepath.

ui/src/app/views/recent-work/component.ts
UrsaStutsman: So not sure we did ever resolve... But I'll enter a ticket for resolution (and implementation if resolution involves changing anything)

ui/src/app/views/recent-work/component.ts
UrsaStutsman: Yeah at some point when we're not quite so under the gun, I think we should have a dev meeting (or two) on style conventions and do up a style guide... I'll put it on my list of things to think about after Algood?

ui/src/app/views/recent-work/component.ts
UrsaStutsman: Sure, I'm going to tie it to the other tickets that I think will resolve this as a side effect (resizing issues, redux of homepage), so that we can resolve it immediately if I'm right...

ui/src/app/views/recent-work/component.ts
UrsaStutsman: I'd assume all (for the workspace). Can't think of any reason not to? In the recent cache, (on the homepage) it makes sense to limit for a number of reasons, but here for workspace specific....

ui/src/app/views/workspace/component.html
```diff
@@ -5,26 +5,34 @@
...
-    <div style="width: 200px">
-      <div role="tooltip" aria-haspopup="true" class="tooltip tooltip-md tooltip-top-right">
-        <span *ngIf="!writePermission" class="tooltip-content">Write permission required to create cohorts</span>
-        <div class="card add-card" [ngClass]="{'disabled' : !writePermission}" (click)="!writePermission || buildCohort()">
-          <div class="create-text">
-            Create a <br/>New Cohort
-            <clr-icon shape="plus-circle" class="plus-circle"></clr-icon>
+    <div style="display: flex; flex-direction: row">
+      <h2 class="greeting" style="width: 200px">Actions</h2>
+      <h2 class="greeting" style="padding-left: 10%; margin-left:2rem">Recently Modified Items</h2>
```
calbach: Per recent Invision comments, drop the "Items", it's cleaner: https://projects.invisionapp.com/d/#/console/12820961/303198606/comments

ui/src/app/views/recent-work/component.ts
calbach: I recall Karthik saying we didn't want the full list here but that may have been on an earlier iteration of this page. Anyways, this is good. It's easy enough to tune this later if that comes up again.

calbach:
Post-merge LGTM - looks like my comments were addressed

PR #1250, opened 2018-09-28
[RISK=no] Cleaning up unused stuff in our build file

rushtong:
No functional changes - just removing some unused code.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -5,9 +5,9 @@ import org.pmiops.workbench.tooling.IncrementalHotSwapTask
 
 def swaggerTemplateDir = 'src/main/resources'
 def mergedApiSourceFile = 'src/main/resources/merged.yaml'
-def clientApiSourceFile = 'src/main/resources/client_api.yaml'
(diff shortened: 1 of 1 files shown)


rushtong:
@calbach - PTAL, thank you.

api/build.gradle
```diff
@@ -309,30 +306,6 @@ dependencies {
   generatedCompile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.9.6'
 }
 
-swaggerSources {
```
calbach: Could you verify that validateSwagger still validates swagger files after this change? I'm not 100% sure, but I vaguely recall setting up some non-obvious hand-off to the swagger plugin to feed into that validation.

rushtong: Thank you, you're correct. This is required. The validation code runs, but doesn't see the files so doesn't fail. This is very non-obvious - I'll close this and look for a better way to make it clear that these are used.

calbach: Sorry, I probably should have left a big comment here explaining why we had this.

PR #1253, opened 2018-09-28
[RISK=no] Swagger Config Comment

rushtong:
No functional changes, just a comment to point users to a required swagger config.


rushtong:
@calbach - Follow on from my last PR, PTAL if you have a chance.

PR #1255, opened 2018-09-28
[Risk=low] Concept synonym column and domain vocabulary info table addition

SrushtiGangireddy:
1. Update the synonym column in concept table to quicken the search.
2. Fill in counts of domain vocabulary info table.
3. Round up the counts to bins of 20 instead of to nearest integer in public data.


api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -177,9 +177,13 @@ echo "Inserting concept table data ... "
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\`
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, count_value, prevalence, source_count_value)
+concept_code, count_value, prevalence, source_count_value, synonyms)
+with concept_synonyms as
+(select c.concept_id as concept, concat(cast(c.concept_id as string),'|',c.concept_code,'|',c.concept_name,'|',string_agg(cs.concept_synonym_name,'|')) as synonyms
```
danqrodney: We should consider escaping concept_name and concept_synonym_name so that | becomes ||. Otherwise if this character *does* appear in a name / synonym, we won't be able to parse it correctly later.

If we think we can pick some delimiter char that we're confident is never going to show up in OMOP names / synonyms, then I guess this isn't necessary.

Also: per discussion on Slack, we may want to add vocabulary_id here too.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -177,9 +177,13 @@ echo "Inserting concept table data ... "
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\`
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, count_value, prevalence, source_count_value)
+concept_code, count_value, prevalence, source_count_value, synonyms)
+with concept_synonyms as
```
danqrodney: Any reason not to just inline this in the SELECT subquery below?

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -303,7 +306,25 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
...
-
+###########################
+# Domain_Vocabulary_Info #
+###########################
+echo "Updating all concept count in domain_vocabulary_info"
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"insert into \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.domain_vocabulary_info\`
+(concept_id,vocabulary_id,all_concept_count,standard_concept_count)
+with all_concept_count as
+(select d.domain_concept_id as concept_id,c.vocabulary_id as vocabulary_id,count(distinct c.concept_id) as all_concept_count
```
danqrodney: So I would do this by joining to domain_info, with a single select, similar to:

https://github.com/all-of-us/workbench/pull/1255/files#diff-f8104808aff893885bceb08ea8f80fd3R219

insert into...

select d.concept_id as concept_id, c.vocabulary_id as vocabulary_id, count(distinct c.concept_id) as all_concept_count, SUM(CASE WHEN c.standard_concept IN ('S', 'C') THEN 1 ELSE 0 END) as standard_concept_count from
...domain_info d JOIN ... concept

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -177,9 +177,13 @@ echo "Inserting concept table data ... "
...
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, count_value, prevalence, source_count_value)
+concept_code, count_value, prevalence, source_count_value, synonyms)
+with concept_synonyms as
+(select c.concept_id as concept, concat(cast(c.concept_id as string),'|',c.concept_code,'|',c.concept_name,'|',string_agg(cs.concept_synonym_name,'|')) as synonyms
+from \`${BQ_PROJECT}.${BQ_DATASET}.concept\` c join \`${BQ_PROJECT}.${BQ_DATASET}.concept_synonym\` cs
+on c.concept_id=cs.concept_id group by c.concept_id,c.concept_code,c.concept_name)
 SELECT c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
-0 as count_value , 0.0 as prevalence, 0 as source_count_value
+0 as count_value , 0.0 as prevalence, 0 as source_count_value,(select synonyms from concept_synonyms where concept=c.concept_id)
```
biopete: Is it tricky to just do the synonym concat in this select rather than making the temp table ?  .. I also like joins over inner selects .. i think they perform better ... not sure though.  I would just move the join above and synonym generation here .

api/db-cdr/generate-cdr/make-bq-data.sh
biopete: Ahh, concept_id is for domain ...  I thought that was the vocabulary_concept_id .. I would use domain_id as the key here. The concept table has domain_id and vocabulary_id ... If we do keep concept_id here it should be the vocabulary_concept_id ... or we could have both. But they are both in the vocabulary and domain tables when we want them anyway.

biopete: The pk here should be domain_id, vocabulary_id in plain english :)

api/db-cdr/generate-cdr/bq-schemas/domain_vocabulary_info.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
```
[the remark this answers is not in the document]
biopete: Update, so we should use domain_id as key for domain as that is what concept uses. 

Also, no participantCount here ?

api/db-cdr/generate-cdr/make-bq-public-data.sh
```diff
@@ -109,13 +109,24 @@ set count_value =
     case when count_value < ${BIN_SIZE}
         then ${BIN_SIZE}
     else
-        cast(ROUND(count_value / ${BIN_SIZE}) * ${BIN_SIZE} as int64)
+        cast(CEIL(count_value / ${BIN_SIZE}) * ${BIN_SIZE} as int64)
```
biopete: Can you update the comment that says how we use ceil instead of rounding now ? Its above this somewhere.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: So if you want to change that, you'll need to change the primary key of domain_info (which means switching to a new table.) Right now its primary key is concept ID.

We could change this to domain_id in a new table, but I'm not sure it really matters?

api/db-cdr/generate-cdr/bq-schemas/domain_vocabulary_info.json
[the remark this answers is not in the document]
danqrodney: See above; concept_id is the primary key of domain_info right now and thus we just need concept ID here.

We could have participant counts here, too, but there's no UX where we display that right now, so we could just add that later if it becomes useful.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: If this works, not sure why we'd want a temp table? 

We do need to make sure this query returns just one row for each concept, so a join to concept synonym outside of a subselect probably doesn't work.

api/db-cdr/generate-cdr/make-bq-data.sh
biopete: Sorry i missed this one in review. domain_id and vocab_id should be the pks because concept_id's can change in omop updates. . They are the pk's of the domain and vocabulary table ..  I think we should be consistent.  

Why does it mean switching to new table?

api/db-cdr/generate-cdr/bq-schemas/domain_vocabulary_info.json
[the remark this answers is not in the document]
biopete: Yeah. We can always query that if we want it.

api/db-cdr/generate-cdr/make-bq-data.sh
biopete: Sorry. I call the " with concept_synonyms as..." thing above a temp table.  'm not sure why we just don't do it all in one query and  left join to concept_synonm table  and put hat concat statement where the select statement is.

api/db-cdr/generate-cdr/make-bq-data.sh
SrushtiGangireddy: I was thinking of some concepts which do not have the synonyms listed.

biopete:
I notice db_domain is still getting created in our schemas . Can that get taken out ... ? 
I think the proper way is that we add a new changelog to drop it in db_schema

api/db-cdr/generate-cdr/make-bq-public-data.sh
SrushtiGangireddy: Added it just beside round comment.

api/db-cdr/generate-cdr/make-bq-data.sh
SrushtiGangireddy: Added concat in the same query.

api/db-cdr/generate-cdr/make-bq-data.sh
SrushtiGangireddy: Escaped the pipe delimiter.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: You can't change the primary key of a table in MySQL AFAIK. We'd need to drop and recreate it.

danqrodney: I'll also note that if the concept ID did change, we'd have to change the CDR metadata in domain_info to match anyway. So while I agree it may be more consistent, not sure it really matters?

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: Not sure why that matters? Anyway, so long as the query works, this is fine.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -177,10 +177,11 @@ echo "Inserting concept table data ... "
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\`
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, count_value, prevalence, source_count_value)
-SELECT c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
-0 as count_value , 0.0 as prevalence, 0 as source_count_value
-from \`$BQ_PROJECT.$BQ_DATASET.concept\` c"
+concept_code, count_value, prevalence, source_count_value, synonyms)
+select c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
+0 as count_value , 0.0 as prevalence, 0 as source_count_value,concat(cast(c.concept_id as string),'\\|',c.vocabulary_id,':',c.concept_code,'\\|',c.concept_name,'\\|',string_agg(cs.concept_synonym_name,'\\|')) as synonyms
```
danqrodney: So it looks like the delimiter is now \| instead of |?

That's OK, and this definitely reduces the chance of the delimiter appearing in a concept name or synonym, though it's still technically possible. If we wanted to avoid any possibility of this happening, what you'd do is escape occurrences of the delimiter in concept names or synonyms... so you could (for example) use | as the delimiter but replace all occurrences of | with || or whatever.

Anyway... I think this combination of characters is sufficiently unlikely to occur that we should be OK.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: So the one performance downside to include concept code and name in the synonyms column is it means if we read concept code and name and synonyms, we're getting the same data twice from the database... so slightly more memory overhead. (I'm not really worried about the disk.)

If we want to instead do what Peter suggests and add a full text index on name, code, synonyms, and update our JPA code to have match_concept() / match_3() or whatever, we'd avoid that. We should verify that searching across multiple columns performs the same and behaves the same way as if they were all mashed together.

I don't mind whether or not we try that in this PR. We could do that as a follow-up later. @biopete WDYT?

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: So you'd do a left join and a group by?

That sounds fine; I'm not sure whether it makes much difference.

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: @biopete do you still feel like we need to address this?

api/db-cdr/generate-cdr/make-bq-data.sh
biopete: We are seeing performance wise. Just taking a while for db to add ft indexes. Anyway, we should fix it in this pr. It will be fine. From my research nobody combines text fields in the same row just to index. It should not effect performance have 3 columns in the index. Much cleaner ..  Innodb indexes are done by tokens in separate index tables so basically every token is taken out of every field and put in an index table. So it doesn't matter what columns the tokens are in. 

I'm still puzzled why doing the join on the full text index to synonms takes so much longer than having synonms on concept ... I'm seeing if it is just a query thing.

api/db-cdr/generate-cdr/bq-schemas/domain_vocabulary_info.json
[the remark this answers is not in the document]
biopete: We need to update this table  to have domain_id and have concept_id be the vocabulary concept_id.  vocabulary_id, domain_id is the pk. The idea is that we have some domain vocab info ... and if we want more info on the vocab , we can look in concept table or concept relationship table or vocabulary table, etc.   This is consistent with how we use domain_info and survey_modules tables.

danqrodney:
OK, didn't realize you could drop and add new primary keys to existing tables in MySQL. 

So it sounds like in this PR, we'd like to:

- Change domain_info's PK to be domain_id (note this means changes to DomainInfo.java)
- Change domain_vocabulary_info to have domain_id rather than concept ID, and make that the foreign key to the parent.
- Add a full text index on concept.name, concept_code, vocabulary ID, and synonyms; change the population code to not bother populating name, concept code, and vocabulary ID into the synonyms column.

In a follow up PR we can add a SQLFunction to use the new compound full text index.

That all sounds fine to me.

biopete:
Sounds good. I guess we don't need the vocabular_concept_id in the table. It would be consistent with our other tables but we can always grab it. I don't care either way.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -300,10 +300,23 @@ echo "Inserting concept_synonym"
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept_synonym\`
  (id, concept_id, concept_synonym_name)
-SELECT 0, c.concept_id, c.concept_synonym_name
+SELECT 0, c.concept_id, REPLACE(c.concept_synonym_name,"|","||") as concept_synonym_name
```
danqrodney: The idea isn't to do the REPLACE here; you can remove this statement. It's to do the REPLACE in the string_agg above.

If you decide to do this, you don't need to put "\|" between synonyms; just | will suffice.

api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -642,30 +642,32 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
...
 FROM \`${BQ_PROJECT}.${BQ_DATASET}.measurement\` m1
 join \`${BQ_PROJECT}.${BQ_DATASET}.person\` p1 on p1.person_id = m1.person_id
 join \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.achilles_results\` ar on cast(m1.measurement_source_concept_id as string)=ar.stratum_1
-where m1.value_as_number is null and m1.value_source_value is not null
+join \`${BQ_PROJECT}.${BQ_DATASET}.concept\` c2 on m1.value_as_concept_id=c2.concept_id
+where m1.value_as_number is null and m1.value_as_concept_id != 0
 and ar.analysis_id = 3000 and ar.stratum_3='Measurement' and ar.stratum_4 != 'unknown'
 and m1.measurement_source_concept_id > 0 and m1.measurement_source_concept_id != m1.measurement_concept_id
-group by m1.measurement_source_concept_id,m1.value_source_value,p1.gender_concept_id"
+group by m1.measurement_source_concept_id,c2.concept_name,p1.gender_concept_id"
```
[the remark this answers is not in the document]
biopete: here: https://github.com/all-of-us/workbench/blob/dd5d9f30bf24dc9952df56f29f70b14db406473a/api/db-cdr/generate-cdr/run-measurement-queries.sh#L653
add the value_as_concept_id to stratum 5 and take out the value_as_number is null for these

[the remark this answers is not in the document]
biopete: here: https://github.com/all-of-us/workbench/blob/dd5d9f30bf24dc9952df56f29f70b14db406473a/api/db-cdr/generate-cdr/run-measurement-queries.sh#L948    change to value_as_concept id for the age deciles

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -177,10 +177,11 @@ echo "Inserting concept table data ... "
 bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 "INSERT INTO \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\`
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
-concept_code, count_value, prevalence, source_count_value)
-SELECT c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
-0 as count_value , 0.0 as prevalence, 0 as source_count_value
-from \`$BQ_PROJECT.$BQ_DATASET.concept\` c"
+concept_code, count_value, prevalence, source_count_value, synonyms)
+select c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
+0 as count_value , 0.0 as prevalence, 0 as source_count_value,concat(cast(c.concept_id as string),'\\|',string_agg(cs.concept_synonym_name,'\\|')) as synonyms
```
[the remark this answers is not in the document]
danqrodney: Did you mean replace '|' with '||'? Yes, that's what I meant.

api/db-cdr/changelog-schema/db.changelog-19.xml
```diff
@@ -0,0 +1,56 @@
...
+      ALTER TABLE concept
+      DROP INDEX idx_concept_name;
+    </sql>
+    <sql dbms="mysql">
+      ALTER TABLE concept
+      DROP INDEX idx_concept_code;
+    </sql>
+    <sql dbms="mysql">
+      ALTER TABLE concept
+      DROP INDEX idx_vocabulary_id;
```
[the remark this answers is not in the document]
SrushtiGangireddy: Yeah, i am only dropping the full text indexes on concept name and code we have btree index in there.

api/db-cdr/changelog-bigdata/db.changelog-1.xml
```diff
@@ -118,6 +118,18 @@
...
+    </changeSet><changeSet id="add concept_code_index full_text" author="biopete" runAlways="true">
+        <validCheckSum>ANY</validCheckSum>
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="concept_code_index"/>
+            </not>
+        </preConditions>
+        <sql dbms="mysql">
+            ALTER TABLE concept
+            ADD FULLTEXT INDEX concept_code_index
```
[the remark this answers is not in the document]
SrushtiGangireddy: This is a normal index on concept_code just in case if we want to search concept_code = ? anytime.

api/db-cdr/changelog-schema/db.changelog-19.xml
```diff
@@ -0,0 +1,56 @@
...
+
+  <changeSet id="add vocabulary_id_index full_text" author="srushtigangireddy" runAlways="true">
+    <validCheckSum>ANY</validCheckSum>
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <indexExists indexName="vocabulary_id_index"/>
+      </not>
+    </preConditions>
+    <sql dbms="mysql">
+      ALTER TABLE concept
```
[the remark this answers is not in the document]
SrushtiGangireddy: dropped full text indexes.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -303,7 +303,20 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
...
+# Domain_Vocabulary_Info #
+###########################
+echo "Updating all concept count in domain_vocabulary_info"
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"insert into \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.domain_vocabulary_info\`
+(domain_id,vocabulary_id,all_concept_count,standard_concept_count)
+select d2.domain_id as domain_id,c.vocabulary_id as vocabulary_id, COUNT(DISTINCT c.concept_id) as all_concept_count,
+SUM(CASE WHEN c.standard_concept IN ('S', 'C') THEN 1 ELSE 0 END) as standard_concept_count from
+\`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.concept\` c
+join \`$WORKBENCH_PROJECT.$WORKBENCH_DATASET.domain\` d2
```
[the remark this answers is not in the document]
SrushtiGangireddy: It it not too many of rows in case we need the other ppi counts anytime.

api/db-cdr/generate-cdr/make-bq-public-data.sh
```diff
@@ -86,15 +86,15 @@ else
...
+copy_tables=(achilles_analysis achilles_results achilles_results_dist concept concept_relationship criteria criteria_attribute domain_info survey_module domain vocabulary concept_synonym domain_vocabulary_info)
 
 for t in "${copy_tables[@]}"
 do
   bq --project=$WORKBENCH_PROJECT rm -f $PUBLIC_PROJECT:$PUBLIC_DATASET.$t
   bq --project=$WORKBENCH_PROJECT --nosync cp $WORKBENCH_PROJECT:$WORKBENCH_DATASET.$t $PUBLIC_PROJECT:$PUBLIC_DATASET.$t
 done
 
-# Round counts for public dataset
+# Round counts for public dataset (The counts are rounded up using ceil. For example 4 to 20, 21 to 40, 43 to 60)
```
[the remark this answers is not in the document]
SrushtiGangireddy: Done.

api/db-cdr/generate-cdr/run-measurement-queries.sh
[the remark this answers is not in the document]
SrushtiGangireddy: Worked on these. Will work on remodeling these queries in next pr.

api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -179,7 +179,7 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
 (concept_id, concept_name, domain_id, vocabulary_id, concept_class_id, standard_concept,
 concept_code, count_value, prevalence, source_count_value, synonyms)
 select c.concept_id, c.concept_name, c.domain_id, c.vocabulary_id, c.concept_class_id, c.standard_concept, c.concept_code,
-0 as count_value , 0.0 as prevalence, 0 as source_count_value,concat(cast(c.concept_id as string),'\\|',string_agg(cs.concept_synonym_name,'\\|')) as synonyms
+0 as count_value , 0.0 as prevalence, 0 as source_count_value,concat(cast(c.concept_id as string),'|',string_agg(replace(cs.concept_synonym_name,'|','||'),'|')) as synonyms
```
biopete: This string_agg should be \| instead of ||

SrushtiGangireddy: string_agg is '|'. It is the replace that is replacing any | in concept_synonym_name with ||.

api/db-cdr/changelog-bigdata/db.changelog-1.xml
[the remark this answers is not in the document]
danqrodney: This isn't a normal index, it's a full text index, right?

(Having a normal index for exact matching on code sounds fine if we don't already have one, but I wouldn't add this.)

api/db-cdr/generate-cdr/make-bq-data.sh
danqrodney: The string_agg is using |... it's just replacing | with || in the synonyms before doing that. Which sounds correct.

biopete: So then you would just split on '|' but not '||' ?

danqrodney: You'd need to have parsing logic which splits on '|' but replaces '||' with '|' to unescape things. Haven't quite worked out what the right code for this will be, but there's a way. :)

PR #1256, opened 2018-09-28
[RW-1278][RW-1280][Risk=no] Add tests to concept search

blrubenstein:
I found this actually tested most of the behavior of the datagrid within the parent component! If I should add datagrid tests as well, please let me know! :)


ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -71,12 +72,89 @@ describe('ConceptHomepageComponent', () => {
...
+    updateAndTick(fixture);
+
+    DomainStubVariables.STUB_DOMAINS.forEach((domain) => {
+      const expectedRequest = {
+        query: searchTerm,
+        // Tests that it searches only standard concepts.
+        standardConceptFilter: StandardConceptFilter.STANDARDCONCEPTS,
+        domain: domain.domain
+      };
+      expect(spy).toHaveBeenCalledWith(
```
calbach: opt: I'm on the fence for these verifications. It does seem that a slightly better test would look at the rendered output and assert that it matches expectations w.r.t. STANDARDCONCEPTS. However, this might make the test case a lot less readable because the stubbed concepts are defined in another file and you shouldn't have to go read that file to understand the test. This is why I'm not super bullish on shared test models (exception: utilities / base models / cases where you just need any valid model).

A middle ground might be to have exported named CONCEPTs on the StubVariablesClass, then reference them here. e.g. check standardconcepts, then assert that row1 contains STANDARD_CONDITION_1.name.
 
Anyways - I think what you have right now is not bad so I don't object to keeping it as is. Mostly food for thought.

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -71,12 +72,89 @@ describe('ConceptHomepageComponent', () => {
...
+        WorkspaceStubVariables.DEFAULT_WORKSPACE_NS,
+        WorkspaceStubVariables.DEFAULT_WORKSPACE_ID,
+        expectedRequest);
+    });
+    // Tests that it makes a call for each domain.
+    expect(spy).toHaveBeenCalledTimes(DomainStubVariables.STUB_DOMAINS.length);
```
calbach: I like this verification. I think this is a good usage of a spy e.g. ensure we didn't make an excess number of calls.

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -71,12 +72,89 @@ describe('ConceptHomepageComponent', () => {
...
+    const searchTerm = 'test';
+    simulateClick(fixture, fixture.debugElement
+      .query(By.css('.standard-concepts-checkbox')).children[0]);
+
+    simulateInput(fixture, fixture.debugElement.query(By.css('#concept-search-input')), searchTerm);
+    simulateClick(fixture, fixture.debugElement.query(By.css('.btn-search')));
+    updateAndTick(fixture);
+
+    DomainStubVariables.STUB_DOMAINS.forEach((domain) => {
+      const expectedRequest = {
```
calbach: opt_nit: I would probably inline this below

ui/src/app/views/concept-homepage/component.spec.ts
blrubenstein: That makes sense. I actually think I disagree though, because we don't really want to test the API behavior here, just that when the UI makes calls, it does so as the API expects it to be made, and then I think we want to validate that the API returns the correct values back. Otherwise I think we should add more checks, rather than removing this.

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -71,12 +72,89 @@ describe('ConceptHomepageComponent', () => {
...
+        expectedRequest);
+    });
+    // Tests that it makes a call for each domain.
+    expect(spy).toHaveBeenCalledTimes(DomainStubVariables.STUB_DOMAINS.length);
+
+    // Tests that it switches to the datagrid view.
+    expect(fixture.debugElement.query(By.css('clr-datagrid'))).toBeTruthy();
+    expect(fixture.debugElement.queryAll(By.css('.concept-row')).length).toBe(1);
+    const firstDomainRowName =
+      fixture.debugElement.queryAll(By.css('.concept-name'))[0].properties.innerHTML;
```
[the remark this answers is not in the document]
calbach: Unresolved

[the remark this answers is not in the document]
blrubenstein: Should be good now.

PR #1257, opened 2018-09-28
[RW-1341][risk=no] Implement selected tab styling

calbach:
Notes:
- Put activated contents inside the button, required dropping the borders/padding 
- Required removing the intermediate button borders for now; implementing them with the clarity button group is tricky - possibly a follow-up (note: the mocks show the dividers spanning only about 75% of the button heights, whereas the current implementation spans 100%)

 Demo: https://calbach-dot-all-of-us-workbench-test.appspot.com/

### DIFF
--- a/ui/src/app/views/workspace-nav-bar/component.css
+++ b/ui/src/app/views/workspace-nav-bar/component.css
@@ -6,11 +6,9 @@
   height: 100%;
   width: 8rem;
   background-color: transparent;
-  border-left: 1px solid rgba(255, 255, 255, 0.15);
-  border-right: 1px solid rgba(255, 255, 255, 0.15);
+  border: none;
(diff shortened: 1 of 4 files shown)


calbach:
Codacy flaked, the tests should be fixed at this point. Repushed so Codacy will rerun. Modulo PR feedback the work has already been done.

I agree we should probably refactor the navigation bar. The number of ng-deep's does not give me confidence about upgrading our Clarity dependency in the future. This solution didn't require much in the way of CSS change though; the most complex piece was probably just figuring out while tab is selected, which should carry over if move off the clarity nav here.

calbach:
Filed as https://precisionmedicineinitiative.atlassian.net/browse/RW-1400

PR #1258, opened 2018-10-01
[RW - 1393] [risk = no] Review - change text Measurement to Drug

smazumd:


ui/src/app/cohort-review/overview-page/overview-page.ts
```diff
@@ -89,8 +89,13 @@ export class OverviewPage implements OnInit, OnDestroy {
...
     this.fetchChartsData(names);
-    this.title = typeToTitle(names);
-    return this.title;
+    if (DomainType[DomainType.DRUG] === names) {
+      return this.title = 'Drugs';
+    } else {
+      this.title = typeToTitle(names);
+      return this.title;
+    }
+
```
dolbeew: I think we should add a case for drugs in the typeToTitle function in utils.ts instead of hardcoding it here

smazumd: As per our conversation I made changes in utils.ts and removed hardcoded value.

PR #1264, opened 2018-10-01
[RW-1284] Adding participant count to concept sets. [risk=no]

danqrodney:
Populating it when updating concepts in a concept set, or cloning a workspace.

Fixing API spec to have conceptCount instead of concept_count.


api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -178,74 +174,33 @@ private ParticipantCriteria getParticipantCriteria(List<CohortStatus> statusFilt
     }
   }
 
+
   private void addFilterOnConcepts(TableQuery tableQuery, Set<Long> conceptIds,
       TableConfig tableConfig) {
-    String standardConceptColumn = null;
     String sourceConceptColumn = null;
     for (ColumnConfig columnConfig : tableConfig.columns) {
-      if (DOMAIN_CONCEPT_STANDARD.equals(columnConfig.domainConcept)) {
```
danqrodney: Note: this would have us looking in just source concept ID column when matching concepts.... which is probably not right for standard concepts. We may need an or here (without necessarily bothering with looking up the concepts in the database as we were before.) Discussing with Peter now.

api/db/changelog/db.changelog-46-add-concept-set-count.xml
```diff
@@ -0,0 +1,12 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="danrodney" id="changelog-46">
+    <addColumn tableName="concept_set">
+      <column name="participant_count" type="INTEGER"/>
+    </addColumn>
+  </changeSet>
+</databaseChangeLog>
```
blrubenstein: Nit: Newline at end of file

api/src/main/java/org/pmiops/workbench/cdr/ConceptBigQueryService.java
```diff
@@ -0,0 +1,67 @@
...
+    innerSql.append(" in unnest(@conceptIds)");
+
+    QueryJobConfiguration jobConfiguration = QueryJobConfiguration
+        .newBuilder(innerSql.toString())
+        .setNamedParameters(ImmutableMap.of("conceptIds",
+            QueryParameterValue.array(conceptIds.toArray(new Long[0]), Long.class)))
+        .setUseLegacySql(false)
+        .build();
+    QueryResult result = bigQueryService.executeQuery(
+        bigQueryService.filterBigQueryConfig(jobConfiguration));
```
blrubenstein: I really like this. This is cool. Kudos.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: After further discussion, reverted the functional changes, but refactored out the logic for getting concept columns and classified concept IDs for use in ConceptBigQueryService.

PR #1266, opened 2018-10-01
[RW-1278][RW-1280][Risk=no] Blrubenstein/domain counts

blrubenstein:
This adds real domain counts, only disables tabs that haven't loaded yet, and adds filtering by vocabulary. We make a new call on filtering in order to properly support pagination once that is added.

### DIFF
--- a/ui/src/app/views/concept-homepage/component.html
+++ b/ui/src/app/views/concept-homepage/component.html
@@ -48,15 +48,20 @@
   <ng-template #chartBox>
     <div class="chart-box lower-box">
       <div class="chart-selector-tabs">
-        <div *ngFor="let domain of conceptDomainList" class="chart-selector-tab">
-          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading">
+        <div *ngFor="let domain of conceptDomainCounts" class="chart-selector-tab">
+          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading || !completedDomainSearches.includes(domain.domain)">
             <div class="chart-selector-header">
               {{domain.name}}
             </div>
-            <!-- TODO (RW-1277): Make this the counts returned from search. -->
-            <div>{{domain.allConceptCount}}</div>
+            <div>{{domain.concept_count}}</div>
           </button>
         </div>
+      </div>
+      <div class="filter-options">
+        Filter by vocabulary:
+        <clr-checkbox class="filter-checkbox" *ngFor="let vocabulary of vocabularies" (click)="filterList()" [(clrChecked)]="vocabulary.selected">
+          {{vocabulary.vocabularyId}} ({{vocabulary.concept_count}})
+        </clr-checkbox>
       </div>
       <app-concept-table [concepts]="concepts" [selectedDomain]="selectedDomain" [loading]="searchLoading" [searchTerm]="currentSearchString"></app-concept-table>
       <div (mouseenter)="addTextHovering=true" (mouseleave)="addTextHovering=false" (click)="openAddModal()" class="add-section">
--- a/ui/src/app/views/concept-homepage/component.html
+++ b/ui/src/app/views/concept-homepage/component.html
@@ -59,7 +59,7 @@
       </div>
       <div class="filter-options">
         Filter by vocabulary:
-        <clr-checkbox class="filter-checkbox" *ngFor="let vocabulary of vocabularies" (click)="filterList()" [(clrChecked)]="vocabulary.selected">
+        <clr-checkbox class="filter-checkbox" *ngFor="let vocabulary of vocabularies" (click)="filterList($event)" [(clrChecked)]="vocabulary.selected">
           {{vocabulary.vocabularyId}} ({{vocabulary.concept_count}})
         </clr-checkbox>
       </div>
--- a/ui/src/app/views/concept-homepage/component.ts
+++ b/ui/src/app/views/concept-homepage/component.ts
@@ -1,6 +1,6 @@
 import {Component, OnInit, ViewChild} from '@angular/core';
 import {ActivatedRoute} from '@angular/router';
-import {BehaviorSubject} from 'rxjs/BehaviorSubject';
+import {Subject} from 'rxjs/Subject';
 
 import {ConceptAddModalComponent} from 'app/views/concept-add-modal/component';
 import {ConceptTableComponent} from 'app/views/concept-table/component';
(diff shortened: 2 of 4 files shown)


ui/src/app/views/concept-homepage/component.html
```diff
@@ -48,16 +48,21 @@
...
-        <div *ngFor="let domain of conceptDomainList" class="chart-selector-tab">
-          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading">
+        <div *ngFor="let domain of conceptDomainCounts" class="chart-selector-tab">
+          <button (click)="selectDomain(domain.domain)" class="domain-selector-button btn btn-link nav-link nav-item" [class.active]="domain.domain === selectedDomain" [disabled]="searchLoading || !completedDomainSearches.includes(domain.domain)">
             <div class="chart-selector-header">
               {{domain.name}}
             </div>
-            <!-- TODO (RW-1277): Make this the counts returned from search. -->
-            <div>{{domain.allConceptCount}}</div>
+            <div>{{domain.concept_count}}</div>
```
calbach: Hmm, we should fix this - I suspect the underscore was an accident @danqrodney ?

ui/src/app/views/concept-homepage/component.html
```diff
@@ -48,16 +48,21 @@
...
           </button>
         </div>
       </div>
+      <div class="filter-options">
+        Filter by vocabulary:
+        <clr-checkbox class="filter-checkbox" *ngFor="let vocabulary of vocabularies" (click)="filterList()" [(clrChecked)]="vocabulary.selected">
```
calbach: Huh, what's the significance of `[(clrChecked)]` ? vs just `[]` or just `()`

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -46,13 +54,23 @@ export class ConceptHomepageComponent implements OnInit {
...
+  completedDomainSearches: Array<Domain> = [];
+
+  vocabularies: Array<VocabularyCountSelected> = [];
+
   wsNamespace: string;
   wsId: string;
 
+  // For some reason clr checkboxes trigger click events twice on click. This
+  // is a workaround to not allow multiple filter events to get triggered.
+  blockMultipleSearchFromFilter = true;
```
calbach: Have you tried `e.stopPropagation()` in your click callback??

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -105,36 +129,51 @@ export class ConceptHomepageComponent implements OnInit {
...
 
-  setConcepts() {
-    this.concepts = this.conceptsCache.find(
-      conceptDomain => conceptDomain.domain === this.selectedDomain).items;
+  setConceptsAndVocabularies() {
+    const cacheItem = this.conceptsCache.find(
+      conceptDomain => conceptDomain.domain === this.selectedDomain);
+    this.concepts = cacheItem.items;
+    this.vocabularies = [];
+    cacheItem.vocabularyList.forEach((vocabulary) => {
```
calbach: opt: map instead of forEach `this.vocatularies = ...map()`

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -144,4 +183,33 @@ export class ConceptHomepageComponent implements OnInit {
   get searchTermNotLongEnough() {
     return this.searchTerm === undefined || this.searchTerm.length < 3;
   }
+
```
calbach: nit: excessive number of newlines

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -144,4 +183,33 @@ export class ConceptHomepageComponent implements OnInit {
...
+    }
+    this.searchLoading = true;
+    const request = {
+      query: this.currentSearchString,
+      standardConceptFilter: standardConceptFilter,
+      domain: this.selectedDomain,
+      vocabularyIds: this.vocabularies
+        .filter(vocabulary => vocabulary.selected).map(vocabulary => vocabulary.vocabularyId),
+    };
+    this.conceptsService.searchConcepts(this.wsNamespace, this.wsId, request)
```
calbach: Error callback?

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -105,36 +129,51 @@ export class ConceptHomepageComponent implements OnInit {
...
-        domain: conceptDomain.domain
+        domain: conceptDomain.domain,
+        includeDomainCounts: activeTabSearch,
+        includeVocabularyCounts: true
       };
       this.conceptsService.searchConcepts(this.wsNamespace, this.wsId, request)
         .subscribe((response) => {
           numCalls += 1;
-          numCallsSubject.next(numCalls);
+          numCallsSubject.next(conceptDomain.domain);
```
calbach: Not really part of the PR but I'm not sure I'm following this `numCallsSubject` approach, I assume it has to do with tracking multiple parallel concept requests. Did you consider using `forkJoin`?

ui/src/app/views/concept-homepage/component.html
blrubenstein: It was, I have already told Dan and he is fixing it. If he gets a fix in before this goes in, I will rebase and change, otherwise I shall send him the two lines to change.

ui/src/app/views/concept-homepage/component.html
blrubenstein: Ooooh! Teaching time, cause I just learned this recently-ish! In angular `[]` is an input binding, and `()` is an output binding. So `[(clrChecked)]` is a two way binding to the variable, such that it defaults to the state passed in, and then changes to the state of the checkbox component are propegated back out to the model variable. Similar to why we do `[(ngModel)]` vs `[inputVariable]` or `(outputHandler)`

ui/src/app/views/concept-homepage/component.ts
blrubenstein: Will try. Let me look into that.

ui/src/app/views/concept-homepage/component.ts
blrubenstein: It is handling multiple concept requests. I think `forkJoin` would have worked before, but now it is changed to enable tabs as the data comes back, so I am not blocking on all of them. I can add a comment if you would like?

ui/src/app/views/concept-homepage/component.ts
blrubenstein: Haven't implemented errors here yet, but I can do that.

ui/src/app/views/concept-homepage/component.ts
blrubenstein: Tried it, it blocked changing the values too. I think Clarity needs the propagation, but for some reason that triggers multiple events... :(

ui/src/app/views/concept-homepage/component.html
danqrodney: Yeah, I'm fixing this. I'll update the PR to also include the corresponding UI change.

blrubenstein: There's also a test file that will need to change. (But only when this goes in) I can send you a list of lines of code after this goes in.

PR #1267, opened 2018-10-02
[RW-1285][risk=low] Factor out a sliding FAB component

calbach:

### DIFF
--- a/ui/src/app/views/concept-homepage/component.html
+++ b/ui/src/app/views/concept-homepage/component.html
@@ -59,10 +59,9 @@
         </div>
       </div>
       <app-concept-table [concepts]="concepts" [selectedDomain]="selectedDomain" [loading]="searchLoading" [searchTerm]="currentSearchString"></app-concept-table>
-      <div (mouseenter)="addTextHovering=true" (mouseleave)="addTextHovering=false" (click)="openAddModal()" class="add-section">
-        <div class="add-text" [class.hovering]="addTextHovering">Add to set</div>
+      <app-sliding-fab [expanded]="'Add to set'" (submit)="openAddModal()">
         <clr-icon class="add-icon" shape="plus"></clr-icon>
-      </div>
+      </app-sliding-fab>
     </div>
   </ng-template>
 </app-top-box>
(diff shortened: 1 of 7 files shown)


calbach:
Tagging in @NehaBroad since @blrubenstein may be out tomorrow (or maybe just thurs/fri? - don't remember)

calbach:
Nevermind, @blrubenstein will be here tomorrow

ui/src/app/views/sliding-fab/component.css
```diff
@@ -0,0 +1,34 @@
+.sliding-button {
+	box-shadow: 0 2px 5px 0 rgba(0,0,0,0.26), 0 2px 10px 0 rgba(0,0,0,0.16);
```
blrubenstein: Nit: Spacing

blrubenstein: Also, where is the box shadow coming from? If its coming from Invision, I would check with Lou, because I've noticed random box shadows sometimes showing up, where, when I talk to Lou, they shouldn't actually be there.

ui/src/app/views/concept-homepage/component.html
```diff
@@ -59,10 +59,9 @@
         </div>
       </div>
       <app-concept-table [concepts]="concepts" [selectedDomain]="selectedDomain" [loading]="searchLoading" [searchTerm]="currentSearchString"></app-concept-table>
-      <div (mouseenter)="addTextHovering=true" (mouseleave)="addTextHovering=false" (click)="openAddModal()" class="add-section">
-        <div class="add-text" [class.hovering]="addTextHovering">Add to set</div>
+      <app-sliding-fab [expanded]="'Add to set'" (submit)="openAddModal()">
         <clr-icon class="add-icon" shape="plus"></clr-icon>
```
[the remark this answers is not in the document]
calbach: I would have liked to be able to do this, unfortunately I think we're going to need to control the detailed styling of the icon - even in the case of the add vs delete icon I've had to size them differently to fit into this button.

ui/src/app/views/sliding-fab/component.css
calbach: Done - and double-checking with @lstrano-broad on the box-shadow. I would definitely expect a shadow on a floating button personally but I wasn't aware of a recurring problem of unexpected box shadows in invision.

ui/src/app/views/concept-homepage/component.html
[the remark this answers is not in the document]
blrubenstein: Got it. In that case, 👍

ui/src/app/views/sliding-fab/component.css
calbach: Verified with Lou that the shadow is intentional.

PR #1271, opened 2018-10-02
[risk=no] Update BigQuery test MySQL dialect

calbach:
Fixes bigquery integration tests.


calbach:
Submitting since the normal CI doesn't come into play on this PR.

PR #1272, opened 2018-10-02
[RW-1225][risk=no] add permissions level labels

jacmrob:

### DIFF
--- a/ui/src/app/utils/workspace-permissions.ts
+++ b/ui/src/app/utils/workspace-permissions.ts
@@ -37,5 +37,4 @@ export class WorkspacePermissions {
     return this.workspace.researchPurpose.reviewRequested === true &&
       this.workspace.researchPurpose.approved === false;
   }
-
 }
--- a/ui/src/app/utils/workspace-permissions.ts
+++ b/ui/src/app/utils/workspace-permissions.ts
@@ -38,8 +38,4 @@ export class WorkspacePermissions {
       this.workspace.researchPurpose.approved === false;
   }
 
-  get isReadWrite() {
-    return this.accessLevel === WorkspaceAccessLevel.WRITER &&
-        this.isOwner === false;
-  }
 }
--- a/ui/src/app/utils/workspace-permissions.ts
+++ b/ui/src/app/utils/workspace-permissions.ts
@@ -40,6 +40,6 @@ export class WorkspacePermissions {
 
   get isReadWrite() {
     return this.accessLevel === WorkspaceAccessLevel.WRITER &&
-        this.isOwner == false;
+        this.isOwner === false;
   }
 }
--- a/ui/src/app/utils/workspace-permissions.ts
+++ b/ui/src/app/utils/workspace-permissions.ts
@@ -37,4 +37,9 @@ export class WorkspacePermissions {
     return this.workspace.researchPurpose.reviewRequested === true &&
       this.workspace.researchPurpose.approved === false;
   }
+
+  get isReadWrite() {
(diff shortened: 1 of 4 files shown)


ui/src/app/views/workspace-list/component.css
```diff
@@ -108,6 +108,34 @@
   color: #216FB4;
 }
 
+.card .owner-permission {
```
jacmrob: is it possible to abstract the permission box into its own class that inherits from `.card`, and then have each permission type inherit from from that?  (`.card .permission-box .owner-permission` or something)  Not sure how to deal with grandparents; new to CSS and frontend coding in general

calbach: There's not really inheritance per-se in vanilla CSS; the typical thing to do is apply multiple CSS classes to an element as needed. When there are conflicts there are rules for which rule wins, but in general more specific selectors / later in the file wins.

So in this case, I'd pull out all these common attributes onto a `.permission` class, then anything specific to each access level I'd put in separate rules:

```
.permission {
 ...
}

.owner-permission {
  background-color: #4996A2;
}

.reader-permission {
...
```

ui/src/app/views/workspace-list/component.html
```diff
@@ -100,17 +100,26 @@ <h3 class="table-header-descriptor">Your Workspaces</h3>
...
+            <div class="date-text">
+              Last Changed:<br>
+              {{workspacePermissions.workspace.lastModifiedTime | date:'short'}}
+            </div>
+
+            <div class="owner-permission" *ngIf="workspacePermissions.isOwner">
+              {{workspacePermissions.accessLevel}}
+            </div>
+            <div class="reader-permission" *ngIf="workspacePermissions.isReadOnly">
+            {{workspacePermissions.accessLevel}}
```
calbach: nit: +2 indent

ui/src/app/utils/workspace-permissions.ts
```diff
@@ -37,4 +37,9 @@ export class WorkspacePermissions {
     return this.workspace.researchPurpose.reviewRequested === true &&
       this.workspace.researchPurpose.approved === false;
   }
+
+  get isReadWrite() {
+    return this.accessLevel === WorkspaceAccessLevel.WRITER &&
+        this.isOwner == false;
```
calbach: Why this second check? I think either the method should be named differently or this part should be removed. I'm not sure what the actual desired behavior is.

rushtong: Ditto, I'm pretty sure that if the access level is Writer, then owner will have to be false. Also, if you did keep that check, `===` instead of `==`

rushtong:
For the PR title, please prefix it with `[RW-X][risk=no]` where X is the number on the ticket.

ui/src/app/views/workspace-list/component.html
```diff
@@ -100,17 +100,26 @@ <h3 class="table-header-descriptor">Your Workspaces</h3>
...
+            </div>
+            <div class="writer-permission" *ngIf="workspacePermissions.isReadWrite">
+              {{workspacePermissions.accessLevel}}
+            </div>
```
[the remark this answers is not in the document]
blrubenstein: I think you can actually do this. I can help with the implementation, cause there's some angular stuff involved in getting that to function cleanly.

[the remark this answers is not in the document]
jacmrob: I think I got it actually!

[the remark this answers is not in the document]
rushtong: Pretty slick!

PR #1274, opened 2018-10-03
[RW-1285][risk=no] Implement initial concept set details page

calbach:
- Fully styled
- Missing most of the interesting functionality:
  - action menu functionality
  - "remove from set" functionality
  - edit mode
- Minor adjustments to the concepts table to support my usage

### DIFF
--- a/ui/src/app/views/breadcrumb/component.ts
+++ b/ui/src/app/views/breadcrumb/component.ts
@@ -85,6 +85,9 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
       if (label === 'Param: Cohort Name') {
         label = child.snapshot.data['cohort'].name;
       }
+      if (label === 'Param: Concept Set Name') {
+        label = child.snapshot.data['conceptSet'].name;
+      }
       // Prevent processing children with duplicate urls
       if (!breadcrumbs.some(b => b.url === url)) {
         const breadcrumb = BreadcrumbComponent.makeBreadcrumb(label, url, child);
--- a/ui/src/app/views/concept-set-details/component.html
+++ b/ui/src/app/views/concept-set-details/component.html
@@ -1,0 +1,51 @@
+<app-top-box>
+  <div class="concept-set-header">
+    <!-- TODO(calbach): Implement the action menu. -->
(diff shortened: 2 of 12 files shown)


ui/src/app/views/concept-set-details/component.html
```diff
@@ -0,0 +1,51 @@
...
+        <h2>{{conceptSet.name}}</h2>
+        <!-- TODO(calbach): Implement edit. -->
+        <button (mouseover)="editHover=true" (mouseleave)="editHover=false"
+            class="btn btn-icon edit-button" [disabled]="true">
+          <app-edit-icon [editHover]="editHover"></app-edit-icon>
+        </button>
+      </div>
+      <p>{{conceptSet.description}}</p>
+      <div class="concept-set-tags">
+        <!-- TODO(calbach): Once counts are returned. -->
```
danqrodney: Counts should be returned now. (I just merged my changes for that.)

ui/src/app/views/breadcrumb/component.ts
```diff
@@ -85,6 +85,9 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
       if (label === 'Param: Cohort Name') {
         label = child.snapshot.data['cohort'].name;
       }
+      if (label === 'Param: Concept Set Name') {
```
UrsaStutsman: Hmmm, we should maybe coordinate on this, because I was using the same param for the concept sets list page, but there was some funky stuff I had to do to prevent dup breadcrumbs... You can look at our PR for Concept Set Creation and list page for details

calbach: Per offline discussion Ursa is checking in with @lstrano-broad on this. The mocks are not consistent right now.

For consistency with the rest of the app as currently implemented, I'd expect:

- List page: Workspaces > workspaceName > Concept Sets
- Details page: Workspaces > workspaceName > Concept Sets > conceptSetName

calbach:
Snuck in a few more minor styling while rebasing against the application-wide font switch. I don't think any should be too controversial so I won't wait for re-review.

PR #1275, opened 2018-10-03
[risk=no] Hide initial spinner on initial error

calbach:


blrubenstein:
Just to check, would it be good to do some tests here?

calbach:
I was looking at this, but it felt like the test setup would probably be too contrived. I think this might be a good place for an eventual UI integration test since it's so dependent on the app wireup. Open to alternative opinions on this if you have them.

To be clear, this wasn't really a regression, but rather a case where I failed to identify the desired behavior.

PR #1277, opened 2018-10-03
[RW-1414][risk=no] Handle notebook paths better

rushtong:
## Addresses
Bug fix for https://precisionmedicineinitiative.atlassian.net/browse/RW-1414

## Changes
Updated the approach to parsing out paths and filenames from notebook gs uris

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
@@ -19,20 +19,27 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
 import javax.inject.Provider;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 @RestController
 public class UserMetricsController implements UserMetricsApiDelegate {
+  private static final Logger log = Logger.getLogger(UserMetricsController.class.getName());
   private Provider<User> userProvider;
   private UserRecentResourceService userRecentResourceService;
   private WorkspaceService workspaceService;
   private FireCloudService fireCloudService;
-  private static final String NOTEBOOKS_WORKSPACE_DIRECTORY = "notebooks/";
   private int distinctWorkspacelimit = 5;
   private Clock clock;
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -41,12 +48,19 @@
...
+        if (notebookName != null && !notebookName.isEmpty()) {
+          try {
+            URI notebookUri = new URI(notebookName);
+            Path path = Paths.get(notebookUri.getPath());
+            String fileName = path.getFileName().toString();
+            String filePath = notebookName.replace(fileName, "");
+            FileDetail fileDetail = new FileDetail().name(fileName).path(filePath);
+            resource.setNotebook(fileDetail);
+          } catch (InvalidPathException | URISyntaxException e) {
+            log.log(Level.WARNING,
```
calbach: I would say this is severe, i.e. bad data in the DB. Debatable whether we'd want to return the resource in this state as it would likely cause errors in the UI - I might prefer to just throw the ServerError (or log severe and don't return this one, but that might complicate the code beyond a reasonable level).

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -155,6 +155,67 @@ public void setUp() {
...
+    RecentResourceResponse recentResources = userMetricsController
+        .getUserRecentResources().getBody();
+    assertNotNull(recentResources);
+    assertEquals(recentResources.get(0).getNotebook().getPath(), "gs://bucketFile/");
+    assertEquals(recentResources.get(0).getNotebook().getName(), "notebook.ipynb");
+  }
+
+  /**
+   * Note about this test vs other tests. This is intended to be run
+   * independently from other tests, therefore we're doing resource
```
calbach: Wording/semantic nit: All (good) tests are meant to be run independently of other tests. I'm guessing the intent here was to express that the setup is different than the standard test setup in this file.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -41,12 +48,19 @@
...
-          fileDetail.setPath(notebookDetails[0] + NOTEBOOKS_WORKSPACE_DIRECTORY);
-          fileDetail.setName(notebookDetails[1]);
-          resource.setNotebook(fileDetail);
+        String notebookName = userRecentResource.getNotebookName();
+        if (notebookName != null && !notebookName.isEmpty()) {
+          try {
+            URI notebookUri = new URI(notebookName);
+            Path path = Paths.get(notebookUri.getPath());
+            String fileName = path.getFileName().toString();
+            String filePath = notebookName.replace(fileName, "");
```
calbach: nit: To future proof, I'd make sure we only swap the last instacne of fileName here. e.g. if we wind up supporting subdirectories for notebooks or filenames without ipynb, 

gs://bkt/notebooks/asdf/asdf
gs://bkt/notebooks/foo.ipynb/foo.ipynb

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
rushtong: Yes, that is true. I'd prefer that the test setup didn't do what it does, but I think that's a separate refactor (I'll likely take that on soon).

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
rushtong: Severe also works for me ... but I would not like to throw a server error to the user if we did something wrong when storing the path information. If it's bad, I think it best to log it so we can deal with it and move along because the user won't be able to do anything about it and even if it were returned, wouldn't be able to do anything with it.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
rushtong: Gotcha - good catch.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
rushtong: Done

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
rushtong: Removed these comments and added a TODO before the setup. Will tackle that refactor separately.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
rushtong: Done

rushtong:
@UrsaStutsman @calbach - Feedback addressed, PTAL, thank you!

PR #1278, opened 2018-10-03
[RW-1422][Risk=low] Change font to montserrat

blrubenstein:

### DIFF
--- a/api/src/main/resources/emails/welcomeemail/content.html
+++ b/api/src/main/resources/emails/welcomeemail/content.html
@@ -1,6 +1,7 @@
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link href="https://fonts.googleapis.com/css?family=Montserrat" rel="stylesheet">
   <style>
     body {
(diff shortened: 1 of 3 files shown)


calbach:
Do you mind bringing up a demo server? We also are using Gotham HTF pretty widely but most of Lou's mocks now have Montserrat instead.

blrubenstein:
Yeah, I can bring one up. Will do so when I have a moment.

calbach:
Also, wasn't Gotham HTF included in that cloudtypography link? If so I would expect all of those font rules to break

blrubenstein:
Yes, but this should remove all gotham htf.

blrubenstein:
I couldn't find any instances of `Gotham` in the UI directory after this PR.

calbach:
Ah - the others I saw were in public-ui (which I think probably never worked).

Also one more in the welcome email (which I think never worked since we didn't load the CSS URL) - probably can just remove that one entirely

blrubenstein:
Added email references as well, I want to add a separate ticket for public-ui

PR #1279, opened 2018-10-03
[RW-1423][risk=no] Update autocomplete trigger and limit

dolbeew:
Set autocomplete trigger to 2 characters for all trees and set results limit to 100 in api call

### DIFF
--- a/ui/src/app/cohort-search/search-bar/search-bar.component.ts
+++ b/ui/src/app/cohort-search/search-bar/search-bar.component.ts
@@ -138,7 +138,7 @@ export class SearchBarComponent implements OnInit, OnDestroy {
   inputChange(newVal: string) {
     this.typedTerm = newVal;
     if (this._type === TreeType[TreeType.VISIT] || this._type === TreeType[TreeType.PM]) {
-      if (newVal.length > 2) {
+      if (newVal.length >= 2) {
         this.actions.setCriteriaSearchTerms([newVal]);
       } else {
         this.actions.setCriteriaSearchTerms([]);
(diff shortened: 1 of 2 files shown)


ui/src/app/cohort-search/search-bar/search-bar.component.ts
```diff
@@ -68,7 +68,7 @@ export class SearchBarComponent implements OnInit, OnDestroy {
     const optionsSub = this.ngRedux
       .select(autocompleteOptions())
       .subscribe(options => {
-        if (this.searchTerm.length >= 4) {
+        if (this.searchTerm.length >= 2) {
```
freemabd1: Should we make this a constant? or maybe create a function that does this check to consolidate this logic since it's in 3 locations?

dolbeew: Good idea, done.

ui/src/app/cohort-search/search-bar/search-bar.component.ts
```diff
@@ -135,11 +137,11 @@ export class SearchBarComponent implements OnInit, OnDestroy {
     this.subscription.unsubscribe();
   }
 
-  inputChange(newVal: string) {
-    this.typedTerm = newVal;
+  inputChange() {
```
calbach: Driveby comment: I happened to notice this behavior yesterday when I was doing the release QA. Maybe you're already planning this or its happening elsewhere, but I would recommend debouncing these queries i.e. if I quickly type in "heart", we ideally don't want to dispatch requests for "he", "hea", "hear", "heart". Debounce throttles inputs if they come in too quickly and ignores intermediates.

Example:
https://github.com/all-of-us/workbench/blob/fc988fc05b1e321b9fb0839abef3f041d9dc301d/public-ui/src/app/views/quick-search/quick-search.component.ts#L97

rushtong: Another example here: https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/workspace-share/component.ts#L64

dolbeew: Nice catch. I switched the input to use reactive forms and added debounce.

PR #1281, opened 2018-10-03
[risk=no] change ws permission label font size

jacmrob:


jacmrob:
fix the permission label font size to match the mockup

PR #1283, opened 2018-10-04
[RW-1285][risk=no] Implement concept set details edit functionality

calbach:
Screenshots (I took some liberties with the designs, based on other screens):

image
image

### DIFF
--- a/ui/src/app/views/concept-set-details/component.ts
+++ b/ui/src/app/views/concept-set-details/component.ts
@@ -12,6 +12,7 @@ import {
   Domain,
   DomainInfo,
   StandardConceptFilter,
+  WorkspaceAccessLevel,
 } from 'generated';
 
 @Component({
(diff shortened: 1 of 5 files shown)


ui/src/app/views/concept-set-details/component.ts
```diff
@@ -26,20 +27,56 @@ import {
...
+      this.editSubmitting = false;
+    });
   }
 
   openRemoveModal() {
     // TODO(calbach): Implement.
   }
+
+  get canEdit(): boolean {
+    return this.accessLevel === WorkspaceAccessLevel.OWNER
```
NehaBroad: Nit: Will it make more sense just to have !this.accessLevel === WorkspaceAccessLevel.Reader?

rushtong: Given the name, `canEdit`, having a positive conditional check makes the most sense to me. Mixing a truthy method with negative conditions is something I try to avoid unless necessary.

PR #1289, opened 2018-10-04
[risk=no] fixing bug with multiple cdr datasources.

freemabd1:
Using the same PoolConfiguration for datasources with different db urls was overriding the db urls to the last cdr db name. Seems that all datasources in the same PoolConfiguration must have the same db url.

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
@@ -93,15 +93,16 @@ public class CdrDbConfig {
           // explicitly initialize the pool parameters here. We override the primary connection
           // info, as the autowired PoolConfiguration is initialized from the same set of properties
           // as the workbench DB.
-          basePoolConfig.setUsername(dbUser);
-          basePoolConfig.setPassword(dbPassword);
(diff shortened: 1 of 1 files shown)


common-api/src/main/java/org/pmiops/workbench/cdr/CdrDbConfig.java
```diff
@@ -93,15 +93,16 @@ public CdrDataSource(CdrVersionDao cdrVersionDao,
           // explicitly initialize the pool parameters here. We override the primary connection
           // info, as the autowired PoolConfiguration is initialized from the same set of properties
           // as the workbench DB.
-          basePoolConfig.setUsername(dbUser);
```
danqrodney: So I think you need to copy the configuration from basePoolConfig somehow... @calbach  to confirm?

freemabd1: this code worked fine locally with 2 cdr versions.

calbach: Yikes - this is a bad bug - my fault most likely. Have to remember that Beans are singleton by default... Thanks for debugging and catching.

I think this is fine, as long as we're copying all of the properties from that pool config.

PR #1291, opened 2018-10-04
[RW-1285][risk=no] Implement concept set details page action menu

calbach:
Depends on #1283 

Screenshots:
image
image

### DIFF
--- a/ui/src/app/views/concept-set-details/component.html
+++ b/ui/src/app/views/concept-set-details/component.html
@@ -1,6 +1,7 @@
 <app-top-box>
   <div class="concept-set-header">
     <clr-dropdown class="action-menu-dropdown">
+      <!-- TODO: Once clone is added to the menu, never disable the dropdown. -->
       <button type="button" class="resource-menu" [disabled]="!canEdit" clrDropdownTrigger>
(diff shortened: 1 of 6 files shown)


calbach:
@lstrano-broad FYI see above screenshots

lstrano-broad:
looks good - thx

ui/src/app/views/concept-set-details/component.html
```diff
@@ -1,11 +1,25 @@
...
+        <button clrDropdownItem class="action-item action-edit"
+            (click)="editing=true"
+            [class.disabled]="!canEdit"
+            [disabled]="!canEdit">
+          <clr-icon shape="pencil"></clr-icon>
+          Edit
+        </button>
+        <button clrDropdownItem class="action-item action-delete"
+            (click)="deleteModal.open()"
+            [class.disabled]="!canEdit"
```
NehaBroad: Since the resource menu will already be disabled, i am not sure if we need [disabled] options for action items

calbach: Added a comment - the disabled dropdown is temporary and we'll permanently enable it once we have clone, so I'm just preemptively disabling the buttons as well - does this sound OK?

NehaBroad: ah ok

PR #1292, opened 2018-10-05
[RW-1285][risk=no] Implement "remove concepts" from concept set support

calbach:
Also adds support in the concept sets stub to optionally add a concept service stub dependency for cross-integration (e.g. adding concepts by ID).

Screenshot:
image

### DIFF
--- a/ui/src/app/views/concept-set-details/component.html
+++ b/ui/src/app/views/concept-set-details/component.html
@@ -70,10 +70,20 @@
   <div *ngIf="conceptSet.concepts?.length" class="concept-table-wrap">
     <app-concept-table [concepts]="conceptSet.concepts"></app-concept-table>
   </div>
-  <app-sliding-fab *ngIf="canEdit" [expanded]="'Remove from set'" (submit)="openRemoveModal()">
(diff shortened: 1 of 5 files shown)


ui/src/app/views/concept-set-details/component.html
```diff
@@ -70,10 +70,20 @@ <h2 *ngIf="!editing">{{conceptSet.name}}</h2>
...
     <clr-icon shape="trash" class="remove-icon"></clr-icon>
   </app-sliding-fab>
 </app-top-box>
 
 <app-confirm-delete-modal [resourceType]="'Concept Set'" [resource]=conceptSet (receiveDelete)=receiveDelete()>
 </app-confirm-delete-modal>
+
+<clr-modal [(clrModalOpen)]="removing" [clrModalSize]="'md'">
+  <div class="modal-title">
+    <div *ngIf="removing" class="modal-title-text">Are you sure you want to remove {{selectedConceptsCount}} concepts from this set?</div>
```
dolbeew: nit: Would it make sense to make the 's' on 'concepts' conditional in the event only 1 concept is being removed?

calbach: Good idea - done.

PR #1297, opened 2018-10-05
[No ticket][Risk=no] open concept set details

UrsaStutsman:
Plugging concept set details in


ui/src/app/views/resource-card/component.ts
```diff
@@ -127,7 +127,7 @@ export class ResourceCardComponent implements OnInit, OnDestroy {
       case ResourceType.COHORT: {
         const url =
           '/workspaces/' + this.wsNamespace + '/' + this.wsId + '/cohorts/build?cohortId=';
-        this.route.navigateByUrl(url + resource.cohort.id);
+        this.router.navigateByUrl(url + resource.cohort.id);
```
blrubenstein: We could also probably use `router.navigate()` here as well, but that may be out of scope for this PR. Leave that refactor up to you.

PR #1300, opened 2018-10-05
[RW-1286] Implementing getDataTableQuery [risk=no]

danqrodney:
Changing API to return a generic "configuration" dictionary, rather than a list of query parameter values. (This is what pandas.read_gbq is going to use, is a bit more flexible for changes in future, and it's simpler than converting QueryParameterValue to the previous parameter objects.)

Working on tests now.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
@@ -619,7 +619,14 @@ public class FieldSetQueryBuilder {
     return outerSql.toString();
   }
 
-  public QueryConfiguration buildQuery(ParticipantCriteria participantCriteria,
+  public QueryJobConfiguration getQueryJobConfiguration(ParticipantCriteria participantCriteria,
+      TableQueryAndConfig tableQueryAndConfig, Long resultSize) {
+    QueryConfiguration queryConfiguration = buildQuery(participantCriteria, tableQueryAndConfig,
+        resultSize, 0L);
+    return bigQueryService.filterBigQueryConfig(queryConfiguration.getQueryJobConfiguration());
+  }
+
+  private QueryConfiguration buildQuery(ParticipantCriteria participantCriteria,
       TableQueryAndConfig tableQueryAndConfig, Long resultSize, long offset) {
     QueryState queryState = new QueryState();
     queryState.schemaConfig = tableQueryAndConfig.getConfig();
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
@@ -262,6 +262,9 @@ public class FieldSetQueryBuilder {
         SelectedColumn selectedColumn = new SelectedColumn();
         selectedColumn.columnInfo = new ColumnInfo(columnName, columnConfig);
         selectedColumn.tableAlias = tableNameAndAlias.alias;
+        // Separate table aliases from column aliases with "__" in the results. These can be
+        // replaced with "." to produce dot notation column names in the output. (BigQuery does not
+        // allow column aliases to contain dots, so we can't do that here directly.)
         selectedColumn.columnAlias = String.format("%s__%s", tableNameAndAlias.alias, columnEnd);
         selectColumns.add(selectedColumn);
       }
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
@@ -688,7 +688,7 @@ public class FieldSetQueryBuilder {
    */
   public Iterable<Map<String, Object>> materializeTableQuery(TableQueryAndConfig tableQueryAndConfig,
       ParticipantCriteria criteria, int limit, long offset) {
-    QueryConfiguration queryConfiguration = buildQuery(criteria, tableQueryAndConfig, limit, offset);
+    QueryConfiguration queryConfiguration = buildQuery(criteria, tableQueryAndConfig, (long) limit, offset);
     QueryResult result;
     QueryJobConfiguration jobConfiguration = queryConfiguration.getQueryJobConfiguration();
     result = bigQueryService.executeQuery(bigQueryService.filterBigQueryConfig(jobConfiguration));
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
@@ -494,7 +494,7 @@ public class FieldSetQueryBuilder {
 
   private String buildSql(ParticipantCriteria participantCriteria, QueryState queryState,
       ImmutableList<SelectedColumn> selectColumns, String whereSql,
-      ImmutableList<OrderByColumn> orderByColumns, long resultSize, long offset) {
+      ImmutableList<OrderByColumn> orderByColumns, Long resultSize, long offset) {
 
     // Joining to tables after applying the LIMIT performs better in BigQuery than
     // joining to them before. Figure out if there are any tables that can be joined to after
@@ -567,8 +567,13 @@ public class FieldSetQueryBuilder {
         }
       }
     }
-    StringBuilder limitOffsetSql = new StringBuilder("\nlimit ");
-    limitOffsetSql.append(resultSize);
+    StringBuilder limitOffsetSql;
+    if (resultSize == null) {
+      limitOffsetSql = new StringBuilder("\n");
+    } else {
+      limitOffsetSql = new StringBuilder("\nlimit ");
+      limitOffsetSql.append(resultSize);
+    }
     if (offset > 0) {
       limitOffsetSql.append(" offset ");
       limitOffsetSql.append(offset);
@@ -614,8 +619,8 @@ public class FieldSetQueryBuilder {
     return outerSql.toString();
   }
 
-  private QueryConfiguration buildQuery(ParticipantCriteria participantCriteria,
-      TableQueryAndConfig tableQueryAndConfig, long resultSize, long offset) {
+  public QueryConfiguration buildQuery(ParticipantCriteria participantCriteria,
+      TableQueryAndConfig tableQueryAndConfig, Long resultSize, long offset) {
     QueryState queryState = new QueryState();
     queryState.schemaConfig = tableQueryAndConfig.getConfig();
     TableQuery tableQuery = tableQueryAndConfig.getTableQuery();
--- a/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
@@ -262,7 +262,7 @@ public class FieldSetQueryBuilder {
         SelectedColumn selectedColumn = new SelectedColumn();
         selectedColumn.columnInfo = new ColumnInfo(columnName, columnConfig);
         selectedColumn.tableAlias = tableNameAndAlias.alias;
-        selectedColumn.columnAlias = String.format("%s_%s", tableNameAndAlias.alias, columnEnd);
+        selectedColumn.columnAlias = String.format("%s__%s", tableNameAndAlias.alias, columnEnd);
         selectColumns.add(selectedColumn);
       }
     }
--- a/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
+++ b/api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
@@ -10,12 +10,22 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.pmiops.workbench.cdr.CdrVersionContext;
 import org.pmiops.workbench.cdr.dao.ConceptService;
 import org.pmiops.workbench.cdr.dao.ConceptService.ConceptIds;
 import org.pmiops.workbench.cohortbuilder.FieldSetQueryBuilder;
 import org.pmiops.workbench.cohortbuilder.ParticipantCriteria;
-import org.pmiops.workbench.cohortbuilder.QueryConfiguration;
 import org.pmiops.workbench.cohortbuilder.TableQueryAndConfig;
 import org.pmiops.workbench.cohortreview.AnnotationQueryBuilder;
 import org.pmiops.workbench.config.CdrBigQuerySchemaConfig;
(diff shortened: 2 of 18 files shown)


api/src/main/resources/client_api.yaml
```diff
@@ -155,56 +155,36 @@ definitions:
...
+          Google SQL to use when querying the CDR. If empty, it means no participants can possibly
+          match the data table specification, and an empty data table should be returned.
         type: string
-      parameters:
-        description: named query parameters used when executing the above SQL
-        type: array
-        items:
-          $ref: "#/definitions/CdrQueryParameter"
+      configuration:
+        description: configuration for the BigQuery job (includes named parameters)
```
calbach: Please expand on this, this is not enough information to know what to do with this as a client.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -47,11 +45,48 @@
...
+
 @Service
 public class CohortMaterializationService {
 
-  private static final String DOMAIN_CONCEPT_STANDARD = "standard";
-  private static final String DOMAIN_CONCEPT_SOURCE = "source";
+  // See https://cloud.google.com/bigquery/docs/parameterized-queries for JSON configuration of query parameters
+  private static final Function<Map.Entry<String, QueryParameterValue>, Map<String, Object>> TO_QUERY_PARAMETER_MAP =
+      (entry) -> {
+        Map<String, Object> result = new HashMap<>();
```
calbach: opt_nit: I think it would be slightly more readable to put this at the end like:

```
return ImmutableMap.newBuilder()
  .put("name", entry.getKey())
  .put(...)
  .build();
```

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -47,11 +45,48 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
 @Service
 public class CohortMaterializationService {
 
-  private static final String DOMAIN_CONCEPT_STANDARD = "standard";
-  private static final String DOMAIN_CONCEPT_SOURCE = "source";
+  // See https://cloud.google.com/bigquery/docs/parameterized-queries for JSON configuration of query parameters
+  private static final Function<Map.Entry<String, QueryParameterValue>, Map<String, Object>> TO_QUERY_PARAMETER_MAP =
```
calbach: What is the expected format of the input to this function? Is this just the representation of a BigQuery parameter as used by the BigQuery Java lib?

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -107,19 +142,13 @@ private ColumnConfig findPrimaryKey(TableConfig tableConfig) {
     throw new IllegalStateException("Table lacks primary key!");
   }
 
-  private TableQueryAndConfig getTableQueryAndConfig(FieldSet fieldSet,
+  private TableQueryAndConfig getTableQueryAndConfig(@Nullable TableQuery tableQuery,
```
calbach: I don't follow the relationship between this PR and the TableQuery becoming nullable.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -223,6 +252,50 @@ private void addFilterOnConcepts(TableQuery tableQuery, Set<Long> conceptIds,
     }
   }
 
+  @VisibleForTesting
+  CdrQuery getCdrQuery(@Nullable CohortReview cohortReview, SearchRequest searchRequest,
```
calbach: opt_nit: in general (barring inconsistency with other related methods) I find sorting parameters in descending order of importance to be the most readable, so usually I'd expect to find nullables towards the end.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -223,6 +252,50 @@ private void addFilterOnConcepts(TableQuery tableQuery, Set<Long> conceptIds,
...
+      @Nullable Set<Long> conceptIds, DataTableSpecification dataTableSpecification) {
+    CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
+    CdrQuery cdrQuery = new CdrQuery();
```
calbach: nit: can use chained setters to initialize this

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -223,6 +252,50 @@ private void addFilterOnConcepts(TableQuery tableQuery, Set<Long> conceptIds,
...
+    cdrQuery.setBigqueryProject(cdrVersion.getBigqueryProject());
+    List<CohortStatus> statusFilter = dataTableSpecification.getStatusFilter();
+    if (statusFilter == null) {
+      statusFilter = NOT_EXCLUDED;
+    }
+    ParticipantCriteria criteria = getParticipantCriteria(statusFilter, cohortReview,
+        searchRequest);
+    if (criteria.getParticipantIdsToInclude() != null
+        && criteria.getParticipantIdsToInclude().isEmpty()) {
+      // There is no cohort review, or no participants matching the status filter;
```
calbach: Using an empty list as a sentinel I think is a bit brittle - instead of `participantIdsToInclude` can you just check for null'ness of `cohortReview` on L268?

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
```diff
@@ -223,6 +252,50 @@ private void addFilterOnConcepts(TableQuery tableQuery, Set<Long> conceptIds,
...
+      // return a query with no SQL, indicating there should be no results.
+      return cdrQuery;
+    }
+    TableQueryAndConfig tableQueryAndConfig = getTableQueryAndConfig(
+        dataTableSpecification.getTableQuery(), conceptIds);
+    QueryConfiguration queryConfiguration = fieldSetQueryBuilder.buildQuery(criteria, tableQueryAndConfig,
+        dataTableSpecification.getMaxResults(),0L);
```
calbach: nit: space before 0

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,225 @@
...
+    cohortReview.setReviewedCount(2);
+    cohortReview.setReviewSize(3);
+    cohortReviewDao.save(cohortReview);
+
+    participantCohortStatusDao.save(makeStatus(cohortReview.getCohortReviewId(), 1L, CohortStatus.INCLUDED));
+    participantCohortStatusDao.save(makeStatus(cohortReview.getCohortReviewId(), 2L, CohortStatus.EXCLUDED));
+
+    ConceptService conceptService = new ConceptService(entityManager, conceptDao, conceptSynonymDao);
+
+    this.cohortMaterializationService = new CohortMaterializationService(fieldSetQueryBuilder,
```
calbach: Is there a reason you can't just `@Autowire` the materialization service rather than explicitly passing all its deps? `EntityManager` seems like the only potential stumbling block here, though I bet if you provided it in your test configuration...

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,225 @@
...
+            null, 0, makeRequest(fieldSet, 1000));
+    ImmutableMap<String, Object> p1Map = ImmutableMap.of("person_id", 1L, "review_status", "INCLUDED");
+    assertResults(response, p1Map);
+    assertThat(response.getNextPageToken()).isNull();
+  }
+
+  @Test
+  public void testMaterializeAnnotationQueryWithPagination() {
+    FieldSet fieldSet = new FieldSet();
+    fieldSet.setAnnotationQuery(new AnnotationQuery());
```
calbach: nit: can use chaining throughout here to initialize these models

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,225 @@
...
+    MaterializeCohortResponse response =
+        cohortMaterializationService.materializeCohort(cohortReview, SearchRequests.allGenders(), null, 0, request);
+    ImmutableMap<String, Object> p1Map = ImmutableMap.of("person_id", 1L, "review_status", "INCLUDED");
+    assertResults(response, p1Map);
+    assertThat(response.getNextPageToken()).isNotNull();
+
+    request.setPageToken(response.getNextPageToken());
+    MaterializeCohortResponse response2 =
+        cohortMaterializationService.materializeCohort(cohortReview, SearchRequests.allGenders(), null, 0, request);
+    ImmutableMap<String, Object> p2Map = ImmutableMap.of("person_id", 2L, "review_status", "EXCLUDED");
```
calbach: I think I'm missing something - is this test hitting BigQuery? If not - how is it working?

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,225 @@
...
+    fieldSet.setAnnotationQuery(new AnnotationQuery());
+    MaterializeCohortRequest request = makeRequest(fieldSet, 1);
+    request.setStatusFilter(ImmutableList.of(CohortStatus.INCLUDED, CohortStatus.EXCLUDED));
+    MaterializeCohortResponse response =
+        cohortMaterializationService.materializeCohort(cohortReview, SearchRequests.allGenders(), null, 0, request);
+    ImmutableMap<String, Object> p1Map = ImmutableMap.of("person_id", 1L, "review_status", "INCLUDED");
+    assertResults(response, p1Map);
+    assertThat(response.getNextPageToken()).isNotNull();
+
+    request.setPageToken(response.getNextPageToken());
```
calbach: Awesome - having pagination tests will be super useful.

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,225 @@
...
+    ParticipantCohortStatusKey key = new ParticipantCohortStatusKey();
+    key.setCohortReviewId(cohortReviewId);
+    key.setParticipantId(participantId);
+    ParticipantCohortStatus result = new ParticipantCohortStatus();
+    result.setStatusEnum(status);
+    result.setParticipantKey(key);
+    return result;
+  }
+
+  private void assertResults(MaterializeCohortResponse response, ImmutableMap<String, Object>... results) {
```
calbach: Could you name these `actual` and `expected` ?

calbach:
Also IMO anything touching the cohort materialization code should probably be at least a `low` risk - even if this PR doesn't directly touch the MySQL query building pieces.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Switched to use builder everywhere in here. Not sure if this is actually more readable, but at least it's immutable. :)

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Before we passed in fieldSet, which was nullable but we hadn't annotated it that way. Now we pass in TableQuery, which is also nullable.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Done.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: So it's possible to have no cohort review, but it's also possible to have a cohort review that doesn't actually include anybody based off the status filter. This handles both of those.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: This code is gone now.

api/src/main/java/org/pmiops/workbench/cohorts/CohortMaterializationService.java
danqrodney: Will follow up on this.

api/src/main/resources/client_api.yaml
danqrodney: Updated.

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
danqrodney: Will come back to this. The trouble is that ConceptService injects @PersistenceContext(unitName = "cdr")... i need to see how to set that up in test config.

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
danqrodney: Annotations are stored in Cloud SQL. So it's not hitting BigQuery here.

api/src/main/java/org/pmiops/workbench/cohortbuilder/FieldSetQueryBuilder.java
```diff
@@ -262,7 +262,7 @@ private TableNameAndAlias getTableNameAndAlias(List<String> columnParts, QuerySt
         SelectedColumn selectedColumn = new SelectedColumn();
         selectedColumn.columnInfo = new ColumnInfo(columnName, columnConfig);
         selectedColumn.tableAlias = tableNameAndAlias.alias;
-        selectedColumn.columnAlias = String.format("%s_%s", tableNameAndAlias.alias, columnEnd);
+        selectedColumn.columnAlias = String.format("%s__%s", tableNameAndAlias.alias, columnEnd);
```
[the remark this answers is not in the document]
danqrodney: Added a comment.

What I'd like to do is have BigQuery allow column aliases like 'foo.bar.baz'... that's how our API refers to columns on related tables. But column aliases can only contain letters, numbers, and underscores. So we're returning double underscores for separators between table aliases here, which can be replaced with dots on the client.

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    DemoQueryBuilder.class, QueryBuilderFactory.class, BigQueryService.class,
+    CohortMaterializationService.class, ConceptService.class, TestJpaConfig.class})
+@MockBean({BigQuery.class})
+public class CohortMaterializationServiceTest {
+
+  private static final String DATA_SET_ID = "data_set_id";
+  private static final String PROJECT_ID = "project_id";
+
+  @Autowired
+  private FieldSetQueryBuilder fieldSetQueryBuilder;
```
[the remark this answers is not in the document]
danqrodney: Removed.

danqrodney:
I'm going to update this PR so that BQ parameter names are generated in a nicer-to-read deterministic form, which will also make writing tests easier. Stay tuned.

danqrodney:
OK, I changed the cohort builder code so that query parameter names are now assigned deterministically; it should work with the code in FieldSetQueryBuilder too. This should make the SQL generated more readable and allow us to write better tests. PTAL.

api/src/test/java/org/pmiops/workbench/cohorts/CohortMaterializationServiceTest.java
```diff
@@ -0,0 +1,275 @@
...
+    participantCohortStatusDao.save(makeStatus(cohortReview.getCohortReviewId(), 2L, CohortStatus.EXCLUDED));
+  }
+
+  @Test
+  public void testGetCdrQueryNoTableQuery() {
+    CdrQuery cdrQuery = cohortMaterializationService.getCdrQuery(
+        SearchRequests.allGenders(), new DataTableSpecification(), cohortReview, null);
+    assertThat(cdrQuery.getBigqueryDataset()).isEqualTo(DATA_SET_ID);
+    assertThat(cdrQuery.getBigqueryProject()).isEqualTo(PROJECT_ID);
+    // TODO: consider making parameter names deterministic, check the entire query
```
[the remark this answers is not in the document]
danqrodney: Done.

PR #1312, opened 2018-10-09
Fixing annotation DB reading of enum values. [risk=no]

danqrodney:
Adding columns to CdrQuery, CohortAnnotationsResponse; these will be used to give the names of columns in resulting data frames (whether or not there are any results.)

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohortreview/AnnotationQueryBuilder.java
+++ b/api/src/main/java/org/pmiops/workbench/cohortreview/AnnotationQueryBuilder.java
@@ -4,14 +4,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.TimeZone;
-import java.util.stream.Collectors;
 import org.pmiops.workbench.db.dao.CohortAnnotationDefinitionDao;
 import org.pmiops.workbench.db.model.CohortAnnotationDefinition;
 import org.pmiops.workbench.db.model.CohortReview;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/cohortreview/AnnotationQueryBuilder.java
```diff
@@ -239,7 +258,7 @@ private String getSql(CohortReview cohortReview, List<CohortStatus> statusFilter
           if (obj != null) {
             String column = columns.get(i);
             if (column.equals(REVIEW_STATUS_COLUMN)) {
-              result.put(column, StorageEnums.cohortStatusFromStorage((Short) obj).name());
+              result.put(column, StorageEnums.cohortStatusFromStorage(((Number) obj).shortValue()).name());
```
calbach: Regression test?

api/src/main/resources/client_api.yaml
```diff
@@ -168,13 +168,22 @@ definitions:
...
 
     properties:
       sql:
         description: >
           Google SQL to use when querying the CDR. If empty, it means no participants can possibly
           match the data table specification, and an empty data table should be returned.
         type: string
+      columns:
+        description: >
+          An array of names to be used for the columns being returned by the query.
```
calbach: How is this different from the values in the select query? Presumably gbq uses the SELECT column names by default for its column labels?

Is it possible to instead implement this with aliases in the SELECT statement? I worry about customizing the logic too far away from standard gbq usage.

api/src/main/java/org/pmiops/workbench/cohortreview/AnnotationQueryBuilder.java
danqrodney: We already have tests in AnnotationQueryBuilder that test this logic... but they use H2 / in-memory which works differently than MySQL apparently, because they pass (whereas enum annotations weren't working for MySQL in test.)

The only way of testing this would be to spin up MySQL, which seems like more trouble than it's worth.

api/src/main/resources/client_api.yaml
danqrodney: Two things:

1) BigQuery doesn't support using dots (e.g. "foo.bar") in column aliases... you can only use letters, numbers, and underscores. That's much uglier / harder to read / doesn't match our way of constructing column names in the client library... so we'd prefer to give dot notation names in the final results. This will allow us to do that.

2) In some cases we don't actually have any matching participants and don't run any SQL (e.g. cohort status = INCLUDED but you didn't mark any participants as INCLUDED in a cohort review.) We still want to return a data frame with the correct columns in that context -- see #workbench-beta about this.

So that's why I'm adding these.

PR #1313, opened 2018-10-09
[RW-1355][risk=no] Set CDR default via cdr_versions.json

calbach:
Moves the default'ness config from config.json to cdr_versions.json (ultimately in MySQL).

### DIFF
--- a/api/config/cdr_versions_local.json
+++ b/api/config/cdr_versions_local.json
@@ -1,6 +1,7 @@
 [
   {
     "cdrVersionId": 1,
+    "isDefault": true,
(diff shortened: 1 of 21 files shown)


danqrodney:
Looks OK; please look into broken tests.

We had discussed an alternative to this wherein there's a column on cdr_version indicating which one is the default... did you decide you'd rather not do that in the end?

calbach:
I was going to punt on this, then I gave it some more thought after your comment and decided it would make sense to just bite the bullet right now while I'm in the thick of this.

calbach:
Playbook updated.

PR #1315, opened 2018-10-10
[RW-1455][risk=low] Handle spaces in path

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1455

## Changes
Handle spaces in the recent paths.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
@@ -51,7 +51,7 @@ public class UserMetricsController implements UserMetricsApiDelegate {
         String notebookName = userRecentResource.getNotebookName();
         if (notebookName != null && !notebookName.isEmpty()) {
           try {
-            URI notebookUri = new URI(notebookName);
+            URI notebookUri = convertToURI(notebookName);
             Path path = Paths.get(notebookUri.getPath());
             String fileName = path.getFileName().toString();
             String filePath = notebookName.replaceFirst(fileName + "$", "");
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -172,6 +172,29 @@ private long getWorkspaceId(String workspaceNamespace, String workspaceId) {
...
-}
 
+  /**
+   * Utility method to parse a URI from a string that may or may not have spaces in the path.
+   *
+   * @param str The string to convert
+   * @return The converted URI
+   * @throws URISyntaxException The URISyntaxException
+   */
+  private URI convertToURI(String str) throws URISyntaxException {
```
blrubenstein: Would it be valuable to put this in a utils file rather than as a function within the component? Maybe here: https://github.com/all-of-us/workbench/blob/master/ui/src/app/utils/index.ts

rushtong: I don't think this is a very generic kind of thing - it's a very narrow use case. Looking on the java side of things, we don't really have a great home for this kind of thing. I'd prefer to punt on this.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -172,6 +172,29 @@ private long getWorkspaceId(String workspaceNamespace, String workspaceId) {
...
+    try {
+      return new URI(str);
+    } catch (URISyntaxException e) {
+      // Handle edge case of path with spaces
+      String newStr = str.replace(" ", "%20");
```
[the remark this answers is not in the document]
calbach: We currently allow any valid GCS path to be created via our own UI. If this isn't exhaustive then it's going to fail on other notebook names. We've already seen `%`s in notebook names in stable which have caused bugs. I'd prefer to just not validate these as true URIs - unless you want to take on a larger task of validating/escaping GCS URIs as they come in and only storing valid URIs (probably will require clearing the recent user cache as well).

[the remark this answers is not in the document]
rushtong: @calbach PTAL - I moved away from URI and back to string parsing for a more generic approach to destructuring a string into a `FileDetail`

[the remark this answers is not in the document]
rushtong: ... however, I just thought of another test case.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -172,6 +155,16 @@ private long getWorkspaceId(String workspaceNamespace, String workspaceId) {
     Workspace dbWorkspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
     return dbWorkspace.getWorkspaceId();
   }
-}
 
+  private FileDetail convertStringToFileDetail(String str) {
+    if (str != null && str.startsWith("gs://")) {
+      int pos = str.lastIndexOf('/') + 1;
+      String fileName = str.substring(pos);
+      String filePath = str.replaceFirst(fileName + "$", "");
```
calbach: Perhaps use `Matcher.QuoteReplacement` on fileName here. Could be problems if I decided to name my notebook `.*.ipynb`

rushtong: 👍

rushtong: Done.

PR #1316, opened 2018-10-10
[RW-1365][Risk=moderate] Blrubenstein/id verification email

blrubenstein:


api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -157,6 +171,48 @@ private String buildWelcomeEmailHtml(String password, User user) throws IOExcept
...
+    try {
+      String msgHtml = buildIdVerificationCompleteHtml(status, username);
+      msg.setHtml(msgHtml);
+      msg.setSubject("All of Us ID Verification Complete");
+      msg.setFromEmail(workbenchConfigProvider.get().mandrill.fromEmail);
+      return msg;
+    } catch (IOException e) {
+      throw new MessagingException("Error reading in email");
+    }
+  }
```
rushtong: optional nit: You could use the chained builder pattern for `MandrillMessage` construction, it might read better. Partial example:
```
msg.html(msgHtml)
    .subject("All of Us ID Verification Complete")
    .fromEmail(workbenchConfigProvider.get().mandrill.fromEmail)
```

blrubenstein: I will try that! (And refactor the other build methods as well)

blrubenstein:
@rushtong should be good for another look

PR #1320, opened 2018-10-10
[RW-1420][risk=no]Review - overview charts data call gets executed too many times [P3]

smazumd:
Review - overview charts data call gets executed too many times [P3]
Disabled button untill get response

### DIFF
--- a/ui/src/app/cohort-review/overview-page/overview-page.ts
+++ b/ui/src/app/cohort-review/overview-page/overview-page.ts
@@ -32,7 +32,6 @@ export class OverviewPage implements OnInit, OnDestroy {
   showTitle = false;
   loading: any;
   domainItems = [];
-  spinner = false;
   selectedCohortName: string;
   review: CohortReview;
   totalParticipantCount: number;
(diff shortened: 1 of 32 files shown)


ui/src/app/cohort-review/overview-page/overview-page.ts
```diff
@@ -51,6 +52,7 @@ export class OverviewPage implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit() {
+    // this.buttonsDisableFlag = true;
```
dolbeew: Remove comment

ui/src/app/cohort-review/overview-page/overview-page.ts
```diff
@@ -39,6 +39,7 @@ export class OverviewPage implements OnInit, OnDestroy {
   isCancelTimerInitiated: any = false;
   domainTitle: '';
   trackClickedDomains = false;
+  buttonsDisableFlag = false;
```
dolbeew: This seems to always be the same as the 'spinner' variable. If so, we can remove this and use 'spinner' instead

smazumd: Removed the spinner flag.

PR #1321, opened 2018-10-11
[risk=no] Fix integration test SQL schema

calbach:


calbach:
Oops - I was too trigger happy here. Meant to do some local validation this morning -this is now showing a different error. Will send another PR.

On a related point, perhaps is_default should be non-nullable.

PR #1322, opened 2018-10-11
Swap to create cluster V2

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
@@ -63,7 +63,7 @@ public class NotebooksServiceImpl implements NotebooksService {
   public Cluster createCluster(String googleProject, String clusterName, String userEmail) {
     ClusterApi clusterApi = clusterApiProvider.get();
     return retryHandler.run((context) ->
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -63,7 +63,7 @@ private ClusterRequest createFirecloudClusterRequest(String userEmail) {
   public Cluster createCluster(String googleProject, String clusterName, String userEmail) {
     ClusterApi clusterApi = clusterApiProvider.get();
     return retryHandler.run((context) ->
-        clusterApi.createCluster(googleProject, clusterName, createFirecloudClusterRequest(userEmail)));
+        clusterApi.createClusterV2(googleProject, clusterName, createFirecloudClusterRequest(userEmail)));
```
calbach: We also want to pass our oauth client ID as the `defaultClientID` when creating the cluster, but that could be a follow-up.

blrubenstein: Looked into this, and saw it is a bit more complicated than just adding it to the `createFirecloudClusterRequest` function, so punting for now.

PR #1325, opened 2018-10-11
[RW-1405][Risk=no] create a recent resource when notebook created

UrsaStutsman:


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -97,7 +97,16 @@ public void setDistinctWorkspaceLimit(int limit) {
   public ResponseEntity<RecentResource> updateRecentResource(String workspaceNamespace, String workspaceId, RecentResourceRequest recentResourceRequest) {
     Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     long wId = getWorkspaceId(workspaceNamespace, workspaceId);
-    UserRecentResource recentResource = userRecentResourceService.updateNotebookEntry(wId, userProvider.get().getUserId(), recentResourceRequest.getNotebookName(), now);
+    String notebookPath;
+    if (recentResourceRequest.getNotebookName().startsWith("gs://")) {
```
blrubenstein: Does this need a null check? Could this be called with a recent resource that is not a notebook?

UrsaStutsman: nope. No null check needed.

PR #1327, opened 2018-10-11
Quick search count display fix

SrushtiGangireddy:
1. Fix that displays standard concept count from domain_info table instead of all concept count.
2. Adding back full text index on concept name if not already exists as it is being used to calculate the domain counts.


api/db-cdr/changelog-schema/db.changelog-20.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
biopete: We don't need this file. We can just remove the add index part from this folder.

api/db-cdr/changelog-schema/db.changelog-master.xml
```diff
@@ -26,4 +26,5 @@
   <include file="changelog-schema/db.changelog-17.xml"/>
   <include file="changelog-schema/db.changelog-18.xml"/>
   <include file="changelog-schema/db.changelog-19.xml"/>
+  <include file="changelog-schema/db.changelog-20.xml"/>
```
biopete: won't need this.

public-ui/src/app/views/quick-search/quick-search.component.html
```diff
@@ -34,14 +34,9 @@ <h5 class="box-heading" style="display:block" *ngIf="showDataType(EHR_DATATYPE)"
             <h5 class="box-title">
               {{r.name}}
             </h5>
-            <div class="card-stat" *ngIf="searchText.value; else standard_count">
-              {{r.allConceptCount | number}}
+            <div class="card-stat">
+              {{r.standardConceptCount | number}}
```
biopete: Good. thanks.

api/db-cdr/changelog-schema/db.changelog-20.xml
```diff
@@ -0,0 +1,23 @@
...
+  <changeSet id="add synonyms_index full_text" author="srushtigangireddy" runAlways="true" onValidationFail="MARK_RAN">
+    <validCheckSum>ANY</validCheckSum>
+    <preConditions onFail="MARK_RAN">
+      <not>
+      <indexExists tableName="concept" indexName="concept_name_index"></indexExists>
+      </not>
+    </preConditions>
+    <sql dbms="mysql">
+      ALTER TABLE concept
+      ADD FULLTEXT INDEX concept_name_index
```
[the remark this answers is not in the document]
danqrodney: OK. I get why we need it now. We can if we want to drop this index later once we've switched the code and data over to use the new index.

PR #1329, opened 2018-10-11
[RW-1348][Risk=no] Blrubenstein/create workspace character limit

blrubenstein:


blrubenstein:
Agreed. I am going to make an investigation ticket in our backlog for that.

PR #1330, opened 2018-10-11
[RW-1438] Make concept search, counting use new synonyms index. [risk=low]

danqrodney:
(This speeds everything up dramatically.)

I'm currently working on parsing synonyms out of synonyms column instead of having to do a second query to fetch those in ConceptService. (Though even with that second query things are still pretty snappy.)


common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -127,107 +125,45 @@ public static String modifyMultipleMatchKeyword(String query){
...
-                            matchFilter.add(root.get("conceptId").in(subquery));
-                            conceptNameFilter.add(criteriaBuilder.or(matchFilter.toArray(new Predicate[0])));
-                            conceptNameFilter.add(criteriaBuilder.or(standardConceptPredicates.toArray(new Predicate[0])));
-
-                            predicates.add(
-                                    criteriaBuilder.or(
-                                            criteriaBuilder.or(conceptCodeIDName.toArray(new Predicate[0])),
-                                            criteriaBuilder.and(conceptNameFilter.toArray(new Predicate[0]))
-                                    ));
+                          List<Predicate> standardOrCodeOrIdMatch = new ArrayList<>();
```
biopete: Is the plan to eventually take this out since we have this in the the ft index now and just check the results to see if it's a code match or concept_id match?

common-api/src/main/java/org/pmiops/workbench/cdr/model/Concept.java
```diff
@@ -184,6 +186,19 @@ public Concept prevalence(float prevalence) {
...
+    }
+
+    public void setSynonymsStr(String synonymsStr) { this.synonymsStr = synonymsStr; }
+
+    public Concept synonymsStr(String synonymsStr) {
+        this.synonymsStr = synonymsStr;
+        return this;
+    }
+
+
```
biopete: What is the plan for the synonm relationship? Remove it eventually?

biopete: Duh. I see. Nice . You use this and set the synonms array and made it transient. Looks good.

SrushtiGangireddy: Looks good

calbach:
I had to stare at this for a while and dig through old PRs but I think I get it now, please check my understanding:
- Previously (and currently, I guess) synonyms are stored in the concept_synonym table
- A recent PR also added a column `concept.synonym` which will instead contain a `|` concatenated list of synonyms
- This PR changes synonym queries to use that column rather than joining against the concept_synonym table.

-> The concept_synonym table may be removed in the future(?)

danqrodney:
@calbach Yep that's all correct. (At the moment, CriteriaDao still uses concept_synonym, so I haven't removed it yet.)

danqrodney:
I'm going to hold off on merging this PR until we get the synthetic data fixed up (as it's missing drugs, etc. right now)

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -44,59 +44,44 @@
...
-        "c.conceptId=?3 or c.conceptCode=?2) and\n" +
+        "matchConcept(c.conceptName, c.conceptCode, c.vocabularyId, c.synonymsStr, ?1) > 0 and\n" +
         "c.standardConcept IN ('S', 'C') and\n" +
-        "c.domainId = ?4\n" +
+        "c.domainId = ?2\n" +
         "group by c.vocabularyId\n" +
         "order by c.vocabularyId\n")
-    List<VocabularyCount> findVocabularyStandardConceptCounts(String matchExp, String query,
-        Long conceptId, String domainId);
+    List<VocabularyCount> findVocabularyStandardConceptCounts(String matchExp, String domainId);
```
biopete: OPtimization thought for later:  Speed wise it is same to not limit by domain_id and just fetch all the domain vocab info in one query when you search.

biopete: I'll add pr when i get a chance as i want to use some of this in databrowser. But basically this in jpa to return domain info objects that have vocabulary_count array in them. : 
select d.domain_id, c.vocabulary_id as vocabulary_id, count(distinct c.concept_id) as conceptCount from domain_info d left join concept c 
  on c.domain_id = d.domain_id
  where (c.count_value > 0 or c.source_count_value > 0) and
  match(c.concept_name, c.concept_code, c.vocabulary_id, c.synonyms) against ('diabetes' in boolean mode) and
  c.standard_concept IN ('S', 'C')
  group by vocabulary_id , d.domain_id

Rather than querying for each domain separately.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
```diff
@@ -38,46 +36,36 @@
...
    *                https://dev.mysql.com/doc/refman/5.7/en/fulltext-boolean.html
-   * @param query the exact query that the user entered
-   * @param conceptId the converted ID value for the query, or null
    */
   @Query(value="select new org.pmiops.workbench.cdr.model.DomainInfo(\n" +
       "d.domain, d.domainId, d.name, d.description,\n" +
-      "d.conceptId, 0L, COUNT(DISTINCT c.conceptId), 0L)\n" +
+      "d.conceptId, 0L, COUNT(*), 0L)\n" +
       "from DomainInfo d\n" +
       "join Concept c ON d.domainId = c.domainId\n" +
```
biopete: Left join here so we get all domainInfos with counts 0 ?

common-api/src/test/java/org/pmiops/workbench/cdr/model/ConceptTest.java
```diff
@@ -0,0 +1,32 @@
+package org.pmiops.workbench.cdr.model;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.Test;
+
+public class ConceptTest {
+
+  @Test
```
biopete: NICE tests. Thank you.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
danqrodney: The requests for each domain / tab come in separately. This is by design, as we don't want to wait for the longest domain to show the tab you're in.

We can consider changing this now that concept search is faster, but I think the way we've got stuff now is fine.

PR #1331, opened 2018-10-11
[RW-654][risk=no] Show CDR name instead of version

calbach:
Demo: https://calbach-dot-all-of-us-workbench-test.appspot.com/workspaces

Note: there may be a brief delay before the name is displayed - during this time it will appear blank.

image

### DIFF
--- a/ui/src/app/views/workspace/component.spec.ts
+++ b/ui/src/app/views/workspace/component.spec.ts
@@ -139,14 +139,6 @@ describe('WorkspaceComponent', () => {
       });
   }));
 
-  it('loads notebooks', fakeAsync(() => {
-    // Mock notebook service in workspace stub will be called as part of ngInit
-    const app = fixture.debugElement.componentInstance;
-    expect(app.notebookList.length).toEqual(1);
-    expect(app.notebookList[0].name).toEqual('mockFile.ipynb');
-    expect(app.notebookList[0].path).toEqual('gs://bucket/notebooks/mockFile.ipynb');
-  }));
-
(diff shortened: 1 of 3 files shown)


ui/src/app/views/workspace/component.spec.ts
```diff
@@ -149,26 +115,37 @@ describe('WorkspaceComponent', () => {
...
         }
       ]}).compileComponents().then(() => {
-        workspacePage = new WorkspacePage(TestBed);
+        fixture = TestBed.createComponent(WorkspaceComponent);
+        updateAndTick(fixture);
       });
-      tick();
   }));
 
   it('displays correct notebook information', fakeAsync(() => {
```
blrubenstein: Is this test old? Are we still displaying the notebooks on the workspace about page?

calbach: It's very old. We are still loading notebooks here though so we can put them in the scroll component.

This test is pretty useless though, I'm just going to drop this test case and leave only the research purpose check as a basic render verification.

calbach:
I'll deploy a demo and check with @lstrano-broad on this.

calbach:
@lstrano-broad please check out the "about workspace" page on my demo server: https://calbach-dot-all-of-us-workbench-test.appspot.com/workspaces

calbach:
Per offline discussion with Lou and Karthik, just going with a simple "Loading..." text for now. We think a spinner here would probably be too distracting, and it loads in fairly quickly anyways.

PR #1334, opened 2018-10-12
[No Ticket][Risk=no] Move to match comps more

blrubenstein:
This adds to the screens the little top box gradient things, and also swaps the order of notebooks and cohorts tabs to match the comps.


ui/src/app/views/cohort-list/component.html
```diff
@@ -1,21 +1,23 @@
...
+      </app-resource-card>
     </div>
   </div>
-  <div *ngFor="let cohort of resourceList">
-    <app-resource-card [resourceCard]="cohort" [cssClass]="'for-list'"
-                       (onUpdate)="reloadCohorts()">
-    </app-resource-card>
-  </div>
-</div>
+</app-top-box>
```
NehaBroad: nit:newline

ui/src/app/views/top-box/component.css
```diff
@@ -22,4 +22,5 @@
   border-left: 1px solid;
   border-right: 1px solid;
   overflow-y: visible;
+  padding: 0 1rem;
```
calbach: This may be totally unrelated and I'm not sure whether you've fixed it here, but one thing that's been bugging me is that the left and right spacing don't match outside the top box; the left margin is bigger. I think it's actually the padding on the app component.

blrubenstein: I don't think I have fixed it here, I can take a look and see if I can figure it out.

blrubenstein: This should fix it now. (Just made the padding on the app-component even. It was in fact that)

PR #1335, opened 2018-10-12
[RW-436][risk=no] Add env-specific GA ids

rushtong:
## Addresses 
https://precisionmedicineinitiative.atlassian.net/browse/RW-436

## Changes
Moves the GA tracking id from the index page into the app so we can make use of different IDs per environment. The ids provided have been created in GA.

## Notes
I copied from two places for the bulk of the code here: 
* https://developers.google.com/analytics/devguides/collection/gtagjs/
* https://github.com/angular/angular-cli/issues/4451#issuecomment-384992203

### DIFF
--- a/ui/src/app/views/app/component.ts
+++ b/ui/src/app/views/app/component.ts
@@ -107,7 +107,8 @@ export class AppComponent implements OnInit {
 
   /**
    * Setting the Google Analytics ID here.
-   * This first injects Google's gtag script via iife, then secondarily defines the global gtag function.
+   * This first injects Google's gtag script via iife, then secondarily defines
+   * the global gtag function.
    */
   private setGTagManager() {
     const s = this.doc.createElement('script');
--- a/ui/src/app/views/app/component.ts
+++ b/ui/src/app/views/app/component.ts
@@ -1,5 +1,5 @@
-import {Location} from '@angular/common';
-import {Component, OnInit} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+import {Component, Inject, OnInit} from '@angular/core';
 import {Title} from '@angular/platform-browser';
 import {
   ActivatedRoute,
@@ -30,6 +30,7 @@ export class AppComponent implements OnInit {
   private overriddenPublicUrl: string = null;
 
   constructor(
+    @Inject(DOCUMENT) private doc: any,
     private activatedRoute: ActivatedRoute,
     private router: Router,
     private titleService: Title
@@ -83,6 +84,8 @@ export class AppComponent implements OnInit {
         this.initialSpinner = false;
       }
     });
+
+    this.setGTagManager();
   }
 
   /**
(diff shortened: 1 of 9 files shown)


ui/src/app/views/app/component.ts
```diff
@@ -102,4 +105,30 @@ export class AppComponent implements OnInit {
     }
   }
 
+  /**
+   * Setting the Google Analytics ID here.
+   * This first injects Google's gtag script via iife, then secondarily defines
```
danqrodney: What is iife?

ui/src/app/views/app/component.ts
```diff
@@ -102,4 +105,30 @@ export class AppComponent implements OnInit {
...
+      '(function(w,d,s,l,i){' +
+        'w[l]=w[l]||[];' +
+        'var f=d.getElementsByTagName(s)[0];' +
+        'var j=d.createElement(s);' +
+        'var dl=l!=\'dataLayer\'?\'&l=\'+l:\'\';' +
+        'j.async=true;' +
+        'j.src=\'https://www.googletagmanager.com/gtag/js?id=\'+i+dl;' +
+        'f.parentNode.insertBefore(j,f);' +
+      '})' +
+      '(window, document, \'script\', \'dataLayer\', \'' + environment.gaId + '\');' +
```
danqrodney: Why not just inline the function above, as we only call it here, once? (Not sure it matters, just curious.)

ui/src/environments/environment.ts
```diff
@@ -3,5 +3,6 @@ import {testEnvironmentBase} from 'environments/test-env-base';
 export const environment = {
   ...testEnvironmentBase,
   displayTag: 'Local->Test',
-  debug: true
+  debug: true,
+  gaId: 'UA-112406425-1'
```
danqrodney: Wondering if we should have a separate web property for local vs. test. Not sure it really matters...

danqrodney:
(Please make sure all the relevant people get access to all the web properties you created.)

ui/src/app/views/app/component.ts
rushtong: I did try that, but it doesn't as the function is then scoped in the iife (will expand on that in other comment). It would have made the code a bit shorter if it worked.

ui/src/environments/environment.ts
rushtong: True - that's easy enough to do.

ui/src/app/views/app/component.ts
rushtong: Immediately Invoked Function Expression
http://benalman.com/news/2010/11/immediately-invoked-function-expression/

rushtong:
Thanks @danqrodney - one additional note on permissions. The new properties are all within the same account and all account users have the same admin-like permissions for that account. I didn't add everyone specifically to the new properties, but it looks like everyone has the same inherited access to all of them.

danqrodney:
OK, cool. At some point we might need to change prod so it's in an account managed by Neil Norman or something... that's what we've done WRT all of our other prod resources. For now this should be fine.

ui/src/app/views/app/component.ts
rushtong: And for those of you around at the time of that article, I was at the jQuery conference (that year) where Ben had a session on this subject. Great speaker.

ui/src/environments/environment.ts
rushtong: Done.

ui/src/app/views/app/component.ts
```diff
@@ -102,4 +105,30 @@ export class AppComponent implements OnInit {
...
+  /**
+   * Setting the Google Analytics ID here.
+   * This first injects Google's gtag script via iife, then secondarily defines
+   * the global gtag function.
+   */
+  private setGTagManager() {
+    const s = this.doc.createElement('script');
+    s.type = 'text/javascript';
+    s.innerHTML =
+      '(function(w,d,s,l,i){' +
```
blrubenstein: What are the various variables here? (Just not really sure what everything is. Is this copied from somewhere?)

rushtong: This function is invoked with the variables provided on line 126. This is mostly copied code from the angular boards, but it's using a javascript construct called "iife" which defines a function, then immediately invokes it. 

In this obscure block of code, the iife is inserting a script tag into the head of the document that looks like this:
```
<script async src="https://www.googletagmanager.com/gtag/js?id=[ga id variable here]"></script>
```

ui/src/environments/environment.ts
calbach: Would declare this either in test-env-base or here, but not both.

PR #1336, opened 2018-10-12
[RW-1416][risk=no] Upgrade prod to newly imported synthetic CDR indexes

calbach:


calbach:
Yeah - Dan and I were discussing on Slack, will post an update there. Dropping this for now.

calbach:
Per offline discussion, this PR will likely be moot, but moving ahead with it in case #1330 doesn't make the next release.

PR #1339, opened 2018-10-12
[risk=no]Us/rw 1318

jacmrob:


ui/src/app/views/homepage/component.html
```diff
@@ -1,139 +1,67 @@
...
+  <div class="row" style="display:flex; justify-content:center">
+    <div class="card single">
+      <div class="row" style="flex-direction: row; height: 17.47%; margin-bottom: 0.5rem; justify-content: flex-start; flex-wrap: nowrap;">
+        <div class="col main-heading">Researcher Workbench</div>
+        <div class="heading-links" style="display: flex; flex-wrap: nowrap; overflow: hidden;">
+          <a class="links" (click)="addWorkspace()">Create a
+            Workspace</a>
+          <a class="links" (click)="listWorkspaces()">See All
+            Workspaces</a>
+          <a class="links">Quick Tour</a>
```
blrubenstein: Is this functional yet? Otherwise I would probably remove it and let the person who adds quick tour add this.

ui/src/app/views/homepage/component.ts
```diff
@@ -162,9 +147,6 @@ export class HomepageComponent implements OnInit, OnDestroy {
     }
     // Don't show the banner after 1 week as their account would
     // have been disabled had they not enabled 2-factor auth.
-    if (new Date().getTime() - this.firstSignIn.getTime() > 1 * 7 * 24 * 60 * 60 * 1000) {
-      return false;
-    }
-    return true;
+    return (new Date().getTime() - this.firstSignIn.getTime() > 7 * 24 * 60 * 60 * 1000);
```
blrubenstein: I think this flipped the logic, so this should be not-ed or the sign should be flipped.

ui/src/app/views/recent-work/component.css
```diff
@@ -1,7 +1,9 @@
 .card-area {
-  display: inline-block;
-  width: 40%;
-  margin-left: 1%;
+    margin-left: 2.3rem;
```
blrubenstein: Nit: Spacing. Should be two spaces rather than 4. (If you would like I can help you configure intellij for that)

ui/src/app/views/recent-work/component.spec.ts
```diff
@@ -86,16 +88,17 @@ describe('RecentWorkComponent', () => {
...
   it('should display recent work', fakeAsync(() => {
-    userMetricsSpy.getUserRecentResources.and.returnValue(Observable.of(stubRecentResources(4)));
+    userMetricsSpy.getUserRecentResources.and.returnValue(Observable.of(stubRecentResources(5)));
     updateAndTick(fixture);
     const de = fixture.debugElement;
     const cardsOnPage = de.queryAll(By.css('.card'));
     const cardNames = de.queryAll(By.css('.name')).map((card) => card.nativeElement.innerText);
-    expect(cardsOnPage.length).toEqual(3);
+    expect(cardsOnPage.length).toEqual(4);
     // should match LAST 3, and NOT include the "oldest"
```
blrubenstein: I think this is stale now.

ui/src/app/views/recent-work/component.ts
```diff
@@ -22,27 +22,51 @@ import {convertToResources, ResourceType} from 'app/utils/resourceActions';
...
   constructor(
     private userMetricsService: UserMetricsService,
     private workspacesService: WorkspacesService,
     private cohortsService: CohortsService,
     private route: ActivatedRoute,
   ) {}
   index: Number;
-  size = 3;
+  @ViewChild('recentWork')
+  eMainFrame: ElementRef;
```
blrubenstein: Nit: What is `eMainFrame`? Can we call this recentWork or something?

ui/src/app/views/homepage/component.html
jacmrob: it's not.  i can comment it out with a note

ui/src/app/views/recent-work/component.ts
jacmrob: sure! it's basically just the whole component element

ui/src/app/views/recent-work/component.css
jacmrob: got it 👍

ui/src/app/views/recent-work/component.css
```diff
@@ -16,15 +18,14 @@
 }
 
 .scroll-indicator {
-  cursor: pointer;
```
calbach: My `cursor: pointer;` change seems to have been reverted. Was this intentional or just a bad merge?

jacmrob: probably a bad merge.  i will fix with 1486

PR #1342, opened 2018-10-15
[RW-1477] Adding concepts to concept set on creation. [risk=no]

danqrodney:
Changing create concept set endpoint to take a request with concept set,
added IDs.

### DIFF
--- a/ui/src/app/views/concept-add-modal/component.ts
+++ b/ui/src/app/views/concept-add-modal/component.ts
@@ -6,6 +6,7 @@ import {
   ConceptSet,
   ConceptSetsService,
   ConceptsService,
+  CreateConceptSetRequest,
   Domain,
   UpdateConceptSetRequest
 } from 'generated';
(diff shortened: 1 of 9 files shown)


ui/src/app/views/concept-add-modal/component.ts
```diff
@@ -114,12 +115,17 @@ export class ConceptAddModalComponent {
...
+      name: this.name,
+      description: this.description,
+      domain: this.selectDomain
+    };
+    const request: CreateConceptSetRequest = {
+      conceptSet: conceptSet,
+      addedIds: conceptIds
+    };
+
+    this.conceptSetsService.createConceptSet(this.wsNamespace, this.wsId, request)
```
NehaBroad: Error handling:  The error message for bad request is set to conflicting name, however that will not be true always. So we might want to set the value of variable error message as that of  the message attribute of error object

danqrodney: The two new cases for throwing BadRequestException in the code -- domain mismatch or too many concepts -- should never occur in the UI as written.  So I'm actually not sure we need to worry about this.

PR #1344, opened 2018-10-15
[RW-1399][Risk=no] Update messaging if no selected vocabularies or no…

blrubenstein:
… concepts found

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
@@ -166,7 +166,7 @@ public class ConceptsController implements ConceptsApiDelegate {
       minCount = 1;
     }
     if (request.getVocabularyIds() != null && request.getVocabularyIds().size() == 0) {
-      throw new BadRequestException("Zero vocabulary options selected");
+      throw new BadRequestException("No vocabulary options selected");
     }
     StandardConceptFilter standardConceptFilter = request.getStandardConceptFilter();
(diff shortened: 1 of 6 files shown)


ui/src/app/views/concept-homepage/component.ts
```diff
@@ -223,4 +242,8 @@ export class ConceptHomepageComponent implements OnInit {
       this.conceptSelected = true;
     }
   }
+
+  get noConceptsConstant() {
+    return 'No concepts found for domain \'' + this.selectedDomain.name + '\' this search.';
```
jacmrob: can this be done more concisely with double quotes?

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
```diff
@@ -165,6 +165,9 @@ private void addVocabularyCounts(SearchConceptsRequest request, ConceptListRespo
     if(minCount == null){
       minCount = 1;
     }
+    if (request.getVocabularyIds() != null && request.getVocabularyIds().size() == 0) {
+      throw new BadRequestException("Zero vocabulary options selected");
```
NehaBroad: Nit: Do you think No vocabulary options selected will make more sense?

ui/src/app/views/concept-homepage/component.ts
blrubenstein: I think it could, but our linter doesn't like double quotes, and sematically I think it makes more sense to have single quotes in the display.

PR #1345, opened 2018-10-15
[No Ticket][Risk=no] Fix the fonts!

blrubenstein:
Lou and I caught a pretty scary bug that meant none of the bold text on the page was actually bold... this should fix that. (Loading the font only loaded the standard formatting.)


calbach:
Nice find, I was wondering why they didn't seem very bold, but I figured maybe it was browser / screen-size dependent.

PR #1348, opened 2018-10-15
[RW-1479][risk=no] Refactor apiHostName -> config.apiBaseUrl to support custom domains

calbach:
In prod/stable, we will not be using appspot domains.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
@@ -73,6 +73,8 @@ public class ClusterControllerTest {
   private static final String BUCKET_NAME = "workspace-bucket";
   private static final Instant NOW = Instant.now();
   private static final FakeClock CLOCK = new FakeClock(NOW, ZoneId.systemDefault());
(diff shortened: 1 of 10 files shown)


api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -87,9 +87,11 @@
   })
   static class Configuration {
     @Bean
-    @Qualifier("apiHostName")
-    String apiHostName() {
-      return "https://api.blah.com";
+    public WorkbenchConfig workbenchConfig() {
+      WorkbenchConfig config = new WorkbenchConfig();
+      config.server = new WorkbenchConfig.ServerConfig();
+      config.server.apiBaseUrl = "https://api.stable.fake-research-aou.org";
```
danqrodney: You could update testLocalize to make sure this value gets written?

calbach:
I realized I'd inadvertently changed the config to use a base URL (including scheme) rather than a host name. Switched back to using host name so I don't break the client library.

Decided to leave it as baseURL in the config to match the other API baseURLs we have defined.

PR #1352, opened 2018-10-16
[RW-1485] Fixing concept fetch used by cohort review.

danqrodney:
Changing pyclient lib used when setting up new clusters to
pyclient-v1-15.


danqrodney:
(Note: we do have a unit tests for the DAO method that wasn't working before, but for some reason H2 doesn't fail here when MySQL does... so there's no simple way of testing this fix.)

PR #1353, opened 2018-10-16
[risk=no][RW-1486] sticky footer on bottom of homepage

jacmrob:

### DIFF
--- a/ui/src/app/views/homepage/component.html
+++ b/ui/src/app/views/homepage/component.html
@@ -23,7 +23,7 @@
       </div>
     </div>
   </div>
-  <div #fullFooter>
+  <div>
     <div class="footer">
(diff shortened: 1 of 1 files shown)


ui/src/app/views/homepage/component.html
```diff
@@ -23,43 +23,45 @@
...
   </div>
-  <div class="footer">
-    <div class="row" style="display: flex; flex-direction:column; margin-left:5%; margin-right: 5%;">
-      <div class="footerTitle">How to Use the All of Us Researcher Workbench</div>
-      <div class="footerRtLink">
-        <div class="tooltip">
-          <a href="#" class="footerLink">
-            See all documentation</a>
-          <span class="tooltiptext">Coming Soon</span>
+  <div #fullFooter>
```
calbach: opt_nit: may not need this ID here since it doesn't seem to be getting used anywhere.

PR #1355, opened 2018-10-16
[RW-1439] Using domain_vocabulary_info for returning vocab counts [risk=no]

danqrodney:
(when there's no query string.)

### DIFF
--- a/api/db-cdr/changelog-data/db.changelog-9.xml
+++ b/api/db-cdr/changelog-data/db.changelog-9.xml
@@ -18,12 +18,24 @@
       <column name="domain_id" type="STRING"/>
       <column name="name" type="STRING"/>
       <column name="description" type="STRING"/>
+      <column name="all_concept_count" type="NUMERIC"/>
+      <column name="standard_concept_count" type="NUMERIC"/>
+      <column name="participant_count" type="NUMERIC"/>
+    </loadData>
+
+    <loadData tableName="domain_vocabulary_info" file="csv/domain_vocabulary_info.csv" encoding="UTF-8" quotchar='"'>
+      <column name="domain_id" type="STRING"/>
+      <column name="vocabulary_id" type="STRING"/>
+      <column name="all_concept_count" type="NUMERIC"/>
+      <column name="standard_concept_count" type="NUMERIC"/>
     </loadData>
 
     <loadData tableName="survey_module" file="csv/survey_module.csv" encoding="UTF-8" quotchar='"'>
       <column name="concept_id" type="NUMERIC"/>
       <column name="name" type="STRING"/>
       <column name="description" type="STRING"/>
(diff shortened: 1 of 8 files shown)


api/db-cdr/csv/domain_vocabulary_info.csv
```diff
@@ -0,0 +1,36 @@
+"domain_id","vocabulary_id","all_concept_count","standard_concept_count"
```
calbach: Hmm.. Doesn't there need to be some script or process for regenerating this file? I assume this exists for the other CSVs here.

api/db-cdr/changelog-data/db.changelog-9.xml
```diff
@@ -18,12 +18,24 @@
       <column name="domain_id" type="STRING"/>
       <column name="name" type="STRING"/>
       <column name="description" type="STRING"/>
+      <column name="all_concept_count" type="NUMERIC"/>
```
calbach: Are these changes to the other tables relevant to the PR? Wouldn't you also need to update the relevant CSV files?

common-api/src/main/java/org/pmiops/workbench/cdr/model/DomainInfo.java
```diff
@@ -188,7 +188,8 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(domain, name, description, conceptId, allConceptCount, standardConceptCount, participantCount);
+    return Objects.hash(domain, name, description, conceptId, allConceptCount, standardConceptCount,
```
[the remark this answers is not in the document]
danqrodney: Done.

api/db-cdr/csv/domain_vocabulary_info.csv
danqrodney: There is a script for generating this data when producing data in Cloud SQL. However, for the purposes of setting up a database locally, we have CSV files in github that we load in directly. See other CSVs in this dir.

This data originally was pulled from Cloud SQL. We don't at present have scripts for grabbing this CSV data out -- I did it manually -- but that's not super-important; we don't really need to make sure these counts for local development match CDR versions we use in production. (And if we were going to try to script it, that would be a different PR.)

api/db-cdr/changelog-data/db.changelog-9.xml
danqrodney: The CSV files already had these columns... the fact that they weren't in here before was an oversight.

api/db-cdr/csv/domain_vocabulary_info.csv
calbach: Thanks, I'm confused every time I look at these CSV files - filed https://precisionmedicineinitiative.atlassian.net/browse/RW-1493 to clarify the roles.

PR #1356, opened 2018-10-16
[RW-1482] Changing cluster localization to write CDR_BILLING_CLOUD_PROJECT [risk=no]

danqrodney:
Value is set to cluster project. Will be used to bill BQ usage to the user billing project, rather than the
billing project for the workspace.


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -51,6 +51,8 @@
   private static final String BUCKET_NAME_KEY = "BUCKET_NAME";
   private static final String CDR_VERSION_CLOUD_PROJECT = "CDR_VERSION_CLOUD_PROJECT";
   private static final String CDR_VERSION_BIGQUERY_DATASET = "CDR_VERSION_BIGQUERY_DATASET";
+  // The billing project to use when querying the CDR.
+  private static final String CDR_BILLING_CLOUD_PROJECT = "CDR_BILLING_CLOUD_PROJECT";
```
calbach: nit: I think just BILLING_CLOUD_PROJECT would make sense. Our current use case is just querying of the CDR, but if they used any other Google services here we might want them to use this same billing project. i.e. the project to bill any analysis to.

PR #1358, opened 2018-10-16
[RW 1352][risk=no]Creating annotation with really long name causes 500, endless spinner

smazumd:
Creating annotation with really long name causes 500, endless spinner

### DIFF
--- a/ui/src/app/cohort-review/set-annotation-create/set-annotation-create.component.html
+++ b/ui/src/app/cohort-review/set-annotation-create/set-annotation-create.component.html
@@ -7,11 +7,12 @@
         class="tooltip tooltip-validation tooltip-sm tooltip-bottom-left"
         [class.invalid]="name.invalid && (name.dirty || name.touched)">
         <input id="name"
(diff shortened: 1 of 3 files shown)


ui/src/app/cohort-review/set-annotation-create/set-annotation-create.component.html
```diff
@@ -7,11 +7,12 @@
...
         [class.invalid]="name.invalid && (name.dirty || name.touched)">
         <input id="name"
-               class="annotation-name-input"
+          class="annotation-name-input"
           name="name"
           formControlName="name"
           type="text"
-               placeholder="ANNOTATION NAME"
+          placeholder="ANNOTATION NAME"
+          maxlength="225"
```
danqrodney: Why 225? I think the limit in the DB is technically 255...

smazumd: I will make it as 255 ... Thanks

PR #1359, opened 2018-10-16
[risk=no][RW-1454] Breadcrumbs refactor

jacmrob:


ui/src/app/views/breadcrumb/component.ts
```diff
@@ -86,13 +86,23 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
         label = child.snapshot.data['cohort'].name;
       }
       if (label === 'Param: Concept Sets Name') {
-        label = child.snapshot.data['workspace'].name + ' Concepts';
+        label = child.snapshot.data['workspace'].name;
```
calbach: I'd drop this clause entirely, the original motivation to dig into the breadcrumbs is that the concept sets list breadcrumb logic didn't match the rest of the application, should drop the "param" piece for the list page in app-routing-module.ts

ui/src/app/views/breadcrumb/component.ts
```diff
@@ -86,13 +86,23 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
...
         if (breadcrumbs.length > 2) {
           breadcrumbs = breadcrumbs.filter(b => !b.url.endsWith('/concepts'));
         }
       }
       if (label === 'Param: Concept Set Name') {
         label = child.snapshot.data['conceptSet'].name;
+        // For the most part, we don't want to append the current label to the breadcrumbs
+        //      since we want the label to stop at the parent.
+        // In this case, we want the breadcrumb header to be the Concept Set Name
+        const conceptSetBreadcrumb = BreadcrumbComponent.makeBreadcrumb(label, url, child);
```
calbach: The existing route hierarchy of concepts defined in app-routing-module.ts doesn't necessarily make sense and I think can freely change in this PR, instead of adding logic here.

Probably what would make sense is:
- workspacePage
  - concepts search
  - concept sets list
    - concept set details

ui/src/app/views/breadcrumb/component.ts
```diff
@@ -86,13 +86,23 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
...
+        breadcrumbs.push(conceptSetBreadcrumb);
+      }
+      // Return if  child is a leaf, or if child's child has same breadcrumb as its parent
```
calbach: Per slack discussion, I think something like an `intermediateBreadcrumb` parameter might make sense to address these cases. Reading more, I'm not actually sure I follow the logic here; why would the child have the same breadcrumb as the parent?

ui/src/app/views/breadcrumb/component.ts
```diff
@@ -65,43 +70,60 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
...
     }
+  }
 
+  /**
+   * Filters an array of Breadcrumbs so that the last element is never an intermediateBreadcrumb
+   * This ensures that breadcrumb headers are displayed correctly while still tracking
+   * intermediate pages.
+   */
+  private filterBreadcrumbs(breadcrumbs: Breadcrumb[]): Array<Breadcrumb> {
+    let last = breadcrumbs[breadcrumbs.length - 1];
```
calbach: nit: should probably handle the 0 length breadcrumbs edge case

ui/src/app/views/breadcrumb/component.ts
```diff
@@ -65,43 +70,60 @@ export class BreadcrumbComponent implements OnInit, OnDestroy {
...
-        const breadcrumb = BreadcrumbComponent.makeBreadcrumb(label, url, child);
+        const breadcrumb = BreadcrumbComponent.makeBreadcrumb(label, breadcrumbType, url, child);
         breadcrumbs.push(breadcrumb);
       }
       return this.buildBreadcrumbs(child, url, breadcrumbs);
     }
+  }
 
+  /**
+   * Filters an array of Breadcrumbs so that the last element is never an intermediateBreadcrumb
```
calbach: Nice, thanks for the generalized solution and detailed comment.

ui/src/app/views/breadcrumb/component.spec.ts
```diff
@@ -131,4 +170,26 @@ describe('BreadcrumbComponent', () => {
       'Breadcrumb label should be "Workspace Name"');
   }));
 
+  it('should show intermediateBreadcrumbs when they have children', fakeAsync( () => {
+    router.navigate(['intermediateParent', 'intermediateChild', 'intermediateGrandchild']);
+    tick();
+    expect(testComponent.breadcrumbs.length).toBe(3);
+    expect(testComponent.breadcrumbs[0].label).toBe('intermediateParent');
+    expect(testComponent.breadcrumbs[1].label).toBe('intermediateChild');
+    expect(testComponent.breadcrumbs[2].label).toBe('intermediateGrandchild');
```
calbach: I think this should just be "grandchild" for this test to make sense (looks like it is indeed a normal `breadcrumb`, not an `intermediateBreadcrumb`)

ui/src/app/views/breadcrumb/component.spec.ts
```diff
@@ -131,4 +170,26 @@ describe('BreadcrumbComponent', () => {
...
+  }));
+
+  it('should not show intermediateBreadcrumbs when they are the last child', fakeAsync( () => {
+    router.navigate(['intermediateParent', 'intermediateChild']);
+    tick();
+    expect(testComponent.breadcrumbs.length).toBe(1);
+    expect(testComponent.breadcrumbs.pop().label).toBe('intermediateParent');
```
calbach: Likewise here, should just be "parent", I think

ui/src/app/views/breadcrumb/component.ts
jacmrob: the empty list?

ui/src/app/app-routing.module.ts
```diff
@@ -150,7 +156,8 @@ const routes: Routes = [
             {
               path: 'concepts',
               data: {
-                breadcrumb: 'Concepts'
+                intermediateBreadcrumb: 'Concepts',
```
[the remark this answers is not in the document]
jacmrob: I think we do want it if we want the breadcrumbs to allow us to redirect back to the Concepts page from an open concepts set (`Workspace -> ws-name -> Concepts -> concept-set-name`.  Unless we want it to redirect us back to the Concept sets page... (`Workspace -> ws-name -> Concept Sets -> concept-set-name`)

ui/src/app/app-routing.module.ts
```diff
@@ -89,36 +91,40 @@ const routes: Routes = [
               component: WorkspaceComponent,
               data: {
                 title: 'View Workspace Details',
-                breadcrumb: 'View Workspace Details'
+                intermediateBreadcrumb: 'View Workspace Details'
```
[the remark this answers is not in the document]
jacmrob: removing it doesn't break anything 👍

ui/src/app/app-routing.module.ts
```diff
@@ -77,7 +82,10 @@ const routes: Routes = [
           component: WorkspaceNavBarComponent,
           data: {
             title: 'View Workspace Details',
-            breadcrumb: 'Param: Workspace Name'
+            breadcrumb: {
+              value: 'Param: Workspace Name',
+              intermediate: false
```
calbach: opt_nit: I would probably also just omit the value instead of false just to make this a bit more readable (code checking `!breadcrumb.intermediate` will work either way, or I would use `!!breadcrumb.intermediate` if you need to coerce to boolean)

ui/src/app/app-routing.module.ts
```diff
@@ -89,36 +97,49 @@ const routes: Routes = [
...
+                }
               }
             },
             {
               path: 'cohorts',
-              data: { breadcrumb: 'Cohorts' },
+              data: {
+                breadcrumb: {
+                  value: 'Cohorts',
+                    intermediate: true
```
calbach: nit: spacing

PR #1366, opened 2018-10-17
[RW-1490][risk=no] no longer using csv files for criteria and criteria_attribute

freemabd1:
no longer using csv files for criteria and criteria_attribute


api/db-cdr/generate-cdr/bq-schemas/criteria.json
```diff
@@ -42,7 +42,7 @@
       {
         "mode": "NULLABLE", 
         "name": "est_count", 
-        "type": "INTEGER"
+        "type": "STRING"
```
biopete: curious why this is a string ?

PR #1368, opened 2018-10-17
[DB-89][Risk=low] Add login guard

blrubenstein:
This adds a login guard to databrowser, requiring you to sign in with an account for the domain which you are on. If you are not signed in, it routes you to a login page that has the option to sign in, and the option to go to the workbench to create an account.


public-ui/src/environments/environment.test.ts
```diff
@@ -3,5 +3,6 @@ import {testEnvironmentBase} from 'environments/test-env-base';
 export const environment = {
   ...testEnvironmentBase,
   displayTag: 'Test',
+  workbenchUrl: 'https://all-of-us-rw-stable.appspot.com',
```
biopete: Do you really want this to be stable?

public-ui/src/environments/environment.stable.ts
```diff
@@ -1,5 +1,7 @@
 export const environment = {
   displayTag: 'Stable',
-    publicApiUrl: 'https://public-api-dot-all-of-us-rw-stable.appspot.com',
+  publicApiUrl: 'https://public-api-dot-all-of-us-rw-stable.appspot.com',
+  workbenchUrl: 'https://all-of-us-rw-stable.appspot.com',
+  clients: '56507752110-ovdus1lkreopsfhlovejvfgmsosveda6.apps.googleusercontent.com',
```
calbach: clientId

PR #1379, opened 2018-10-19
[RW-1270][risk=low] Fix bug report bugs, add "CDR" bug type

calbach:
- API support for an alternate bug type "Data Quality" or "Workbench Application" (default is "application")
- "data" sends tickets to the CDR Jira component
- UI for dropdown (checking UX guidance with @lstrano-broad )
- Minor bug report UX fixes
  - Use a loading spinner and wait for submit before closing the modal
  - Clear error state on the dialog before reopening

image

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -1854,6 +1854,8 @@ definitions:
       - shortDescription
       - contactEmail
     properties:
+      bugType:
+        $ref: "#/definitions/BugReportType"
       shortDescription:
         description: A short description of the bug
         type: string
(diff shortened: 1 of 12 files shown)


calbach:
Got feedback from Lou - updated the UX to better match our other modals and improved checkbox alignment (still not perfect - there's some mystery padding coming from somewhere on the checkbox)

api/src/main/resources/workbench.yaml
```diff
@@ -1869,6 +1871,11 @@ definitions:
         description: The email with which to contact the bug reporter.
         type: string
 
+  BugReportType:
+    type: string
+    description: Possible value for bug reports
+    enum: [APPLICATION, DATA]
```
UrsaStutsman: Do we want to also add an attribute for environment reported from?  either here or above? Think it would be a good pc of information to have included.

PR #1385, opened 2018-10-23
[RW-474][risk=low] Set a default cluster oauth client ID on all Leo clusters

calbach:
Previously we depended on inter-window communication between Workbench and Leo to communicate the Oauth client ID to use for credential refreshes. This was brittle and would fail in particular basic modes, e.g. copy-paste a link or closing the original parent tab.

This inter-tab communication also somehow broken within the past few releases so currently we don't seems to plumb through a client ID at all, i.e. tabs die after 1h.

I've confirmed locally that this properly plumbs down the client id.

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -31,7 +31,8 @@
     "apiBaseUrl": "https:\/\/api-dot-all-of-us-workbench-test.appspot.com",
     "publicApiKeyForErrorReports": "",
     "projectId": "all-of-us-workbench-test",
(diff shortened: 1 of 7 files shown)


calbach:
Actually, looks like #1216 just moved the code elsewhere, so I've not confirmed that was the cause of the regression.

calbach:
Update: looks like it was actually a Leo regression and not #1216 - sent fix: https://github.com/DataBiosphere/leonardo/pull/644

PR #1386, opened 2018-10-23
[RW-751][Risk=no] adding length constraints to user fields

UrsaStutsman:
Adding length constraint to given name, family name, and area of research fields in profile page


rushtong:
Please add the jira ticket and risk to the PR title.

PR #1388, opened 2018-10-23
[RW-1317][Risk=moderate] Use IFrame to display notebooks also updating redirect page (RW-931)

UrsaStutsman:
This PR is passing and received approval. HOLD ON MERGE for Leo(FC) release. Current ETA 4pm(EST) today (11/1)


ui/src/app/views/notebook-frame/component.ts
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      this.jupyterUrl = `/workspaces/${this.workspace.namespace}/${this.workspace.id}/` +
+        `notebooks/create/?notebook-name=` +
+        encodeURIComponent(this.route.snapshot.queryParams['notebook-name']) +
+        `&kernel-type=${this.route.snapshot.queryParams['kernelType']}`;
+    }
+  }
+
+  get iFrameSrc() {
+    return this.sanitizer.bypassSecurityTrustResourceUrl(this.jupyterUrl);
```
calbach: Why is this needed? I also don't think this meets the requirements of being a trusted value, parts of this came from the URL which a malicious user can control by either sending a link they constructed, or by using a particular notebook name.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -121,7 +121,8 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     this.creating = this.route.snapshot.data.creating;
 
     if (this.creating) {
-      this.notebookName = this.route.snapshot.queryParamMap.get('notebook-name');
+      this.notebookName =
+        decodeURIComponent(this.route.snapshot.queryParamMap.get('notebook-name'));
```
calbach: Why would this change, was this a display bug?

ui/src/app/views/signed-in/component.html
```diff
@@ -1,4 +1,4 @@
-<div class="page-container">
+<div *ngIf="!inNotebookFrame" class="page-container">
```
calbach: Could you add a high-level description of the approach you are using to the PR description? I think I see it now, but I wasn't initially understanding why you'd need to do this - are you loading our application inside the iframe initially?

An alternative would be to take the existing notebook redirect page, and instead of redirecting at the end show the iframe. Do we plan to continue supporting the full page redirect -> Jupyter after we have iframe support?

ui/src/app/views/notebook-frame/component.ts
```diff
@@ -0,0 +1,44 @@
...
+import {WorkspaceData} from 'app/resolvers/workspace';
+import {Workspace} from 'generated';
+
+@Component({
+  selector: 'app-notebook-frame',
+  styleUrls: ['./component.css'],
+  templateUrl: './component.html'
+})
+
+export class NotebookFrameComponent implements OnInit {
```
rushtong: nit: naming ... would this be better named `NotebookIFrameComponent`? I was halfway through the PR before I understood what a frame was.

ui/src/app/views/notebook-frame/component.ts
UrsaStutsman: Angular doesn't trust iframe sources from what I've read. If there's another way of doing this I'm all up for doing it, just couldn't figure out what/how.

ui/src/app/views/notebook-frame/component.ts
UrsaStutsman: I can rename

ui/src/app/views/notebook-redirect/component.ts
UrsaStutsman: Um, think that's what it was I'd have to check again locally. Will do now.

ui/src/app/views/signed-in/component.html
UrsaStutsman: Maybe we should chat, as I'm a little confused by your last question. I could implement differently as far as the notebook redirect page, but since the notebookframecomponent is still in our app, it would render with the our header, which is what I'm attempting to avoid here.

rushtong:
Functional note: clicking on the navbar icon causes the frame to reload which leads to strange behavior. The menu with workspace actions shows up while the iframe reloads underneath it, even though it was previously loaded.

ui/src/app/views/notebook-frame/component.ts
```diff
@@ -0,0 +1,44 @@
...
+  ) {
+    const wsData: WorkspaceData = this.route.snapshot.data.workspace;
+    this.workspace = wsData;
+  }
+
+  ngOnInit(): void {
+    this.notebookName = this.route.snapshot.params['nbName'];
+    if (this.notebookName) {
+      this.jupyterUrl = '/workspaces/' + this.workspace.namespace + '/' +
+        this.workspace.id + '/notebooks/' + this.notebookName;
```
rushtong: I know we've run into this in the past - there are old cases where notebooks aren't actually in a `/notebooks/` directory. This feels brittle to me, but I don't have a better solution.

ui/src/app/views/notebook-redirect/component.ts
UrsaStutsman: Yup, for sure. It displays the name with URI encoding in the Loading message

ui/src/app/views/new-notebook-modal/component.ts
```diff
@@ -51,11 +50,14 @@ export class NewNotebookModalComponent {
...
-        `&kernel-type=${this.kernelType}`;
-
-    window.open(nbUrl, '_blank');
+    this.route.navigate(['workspaces', this.workspace.namespace, this.workspace.id,
+        'notebooks', 'create'], {
+      queryParams: {
+        'notebook-name': encodeURIComponent(this.newName),
+        'kernelType': this.kernelType
+      },
+      relativeTo: null
```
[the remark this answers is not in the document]
UrsaStutsman: yup needs this. Not sure what it defaults to, but errors otherwise.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -176,7 +179,9 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
       })
       .subscribe((nbName) => {
         this.progress = Progress.Redirecting;
-        this.window.location.href = this.notebookUrl(this.cluster, nbName);
+        this.notebookLoaded = true;
```
calbach: Note: It would be nicer if we waited for some loading status from the iframe before showing it. Not for this PR but possible future improvement.

ui/src/app/app-routing.module.ts
```diff
@@ -128,8 +128,27 @@ const routes: Routes = [
                   intermediate: true
                 }
               }
-            },
-            {
+            }, {
+              path: 'notebooks/create',
```
calbach: We're probably going to want to change the behavior here. The create should actually create the notebook, then navigate to the "/notebooks/name" page. Alternatively we change "?create" to be a query parameter and put the name in the URL normally.

The issue with the current approach is that the top bar is going to be "Creating a new Notebook" instead of the notebook name while you're looking at your newly created notebook.

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -220,25 +213,22 @@ describe('NotebookRedirectComponent', () => {
...
-
-    fixture.componentInstance.notebookName = '1%2B1.ipynb';
-    blockingClusterStub.cluster.status = ClusterStatus.Running;
-    tick(10000);
     fixture.detectChanges();
-
-    expect(fakeWindow.location.href).toContain('/1%252B1.ipynb');
+    expect(fixture.debugElement.query(By.css('.title')).nativeElement.textContent)
+      .toMatch('blah blah');
+    fixture.destroy();
```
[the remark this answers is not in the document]
UrsaStutsman: "blah blah" is the name of the test notebook. I changed to "blah blah" from "blah" to capture having a space in there.

ui/src/app/app-routing.module.ts
UrsaStutsman: When you say "top bar" do you mean in the browser's navigation bar?  Because the "top bar" of our app, shows the breadcrumbs regardless of whether it's a create or not, nothing referring to creating. The redirect page WILL say "creating new notebook" (not in top bar, but at top of page, under top bar) but, that makes sense to me. Can you clarify before I make this change?

calbach: Huh, you're right that breadcrumbs aren't being shown. I think that may actually just be a bug. Don't need to worry about it for this PR then. Filed RW-1564

ui/src/app/views/notebook-redirect/component.spec.ts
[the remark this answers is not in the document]
calbach: My question was specifically about this line I commented on: `fixture.destroy()` I don't understand why that was added.

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -220,25 +213,22 @@ describe('NotebookRedirectComponent', () => {
...
     blockingClusterStub.cluster.status = ClusterStatus.Running;
     updateAndTick(fixture);
     fixture.detectChanges();
-    expect(spinnerText()).toContain('Redirecting');
+    expect(fixture.debugElement.query(By.css('.i-frame'))).toBeTruthy();
   }));
 
 
-  it('should escape notebooks names', fakeAsync(() => {
+  it('should properly display notebooks names', fakeAsync(() => {
```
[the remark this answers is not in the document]
calbach: It was testing whether the outgoing notebook name is being encoded - the analog in this PR would be to check the iframe src URL. The bug we saw previously was that a notebook name with URL escape sequences could would be improperly handled when we redirected to the Jupyter UI.

 The display bug is probably worth having a test for but it's different.

ui/src/app/app-routing.module.ts
UrsaStutsman: Hey CH. Closed that bug. Since the notebookRedirectComponent now includes the display of the notebook on redirect, I already addressed breadcrumbs being shown. I think you misunderstood my original comment. The notebook redirect page WILL now show breadcrumbs (see img I attached to bug before closing).

ui/src/app/views/notebook-redirect/component.html
```diff
@@ -1,25 +1,43 @@
...
-  <p *ngSwitchCase="Progress.Creating" class="spinner-text">
-    Creating the new notebook
-  </p>
-  <p *ngSwitchCase="Progress.Redirecting" class="spinner-text">
-    Redirecting to the notebook server
-  </p>
-</div>
+<app-top-box>
+  <div *ngIf="progress !== Progress.Loaded">
+    <div style="display: flex; flex-direction: column; margin-left: 3rem; width: 780px; padding-top: 1rem">
```
calbach: No action needed: We should probably make a codebase decision on this, but personally I'm not really a fan of inline styles (with some exceptions). They are less performant in the browser and it feels more consistent having styles all defined in the same place.

ui/src/app/views/notebook-redirect/component.html
```diff
@@ -1,25 +1,43 @@
...
+      <div style="display: flex; flex-direction: row; justify-content: space-between; flex-grow: 1;">
+        <h2 class="notebooks-header" *ngIf="!creating">Loading Notebook '{{notebookName}}'</h2>
+        <h2 class="notebooks-header" *ngIf="creating">Creating New Notebook '{{notebookName}}'</h2>
+        <button id="cancelButton" class="btn btn-secondary" style="margin-right: 0;" (click)="navigateBack()">Cancel</button>
+      </div>
+      <div class="progress-container">
+        <div id="Initializing" class="progress-card" [class.progress-card-current]="[Progress.Initializing, Progress.Resuming, Progress.Unknown].includes(progress)">
```
calbach: nit: generally our id names are lowercased

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -175,8 +179,19 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
         return localizeObs.retry(3);
       })
       .subscribe((nbName) => {
-        this.progress = Progress.Redirecting;
-        this.window.location.href = this.notebookUrl(this.cluster, nbName);
+        this.incrementProgress(Progress.Redirecting);
+        if (this.creating) {
+          window.history.replaceState({}, '', 'workspaces/' + this.wsNamespace +
+          '/' + this.wsId + '/notebooks/' + this.fullNotebookName);
```
calbach: encodeUriComponent

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -237,4 +267,21 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
       })
       .map(resp => resp.clusterLocalDirectory);
   }
+
+    navigateBack(): void {
+        this.locationService.back();
```
calbach: -2 spaces

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -175,8 +179,19 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
         return localizeObs.retry(3);
       })
       .subscribe((nbName) => {
-        this.progress = Progress.Redirecting;
-        this.window.location.href = this.notebookUrl(this.cluster, nbName);
+        this.incrementProgress(Progress.Redirecting);
+        if (this.creating) {
+          window.history.replaceState({}, '', 'workspaces/' + this.wsNamespace +
```
calbach: Does this clear the window title? (that should be the second parameter)

ui/src/app/views/recent-work/component.spec.ts
```diff
@@ -102,7 +102,7 @@ describe('RecentWorkComponent', () => {
     expect(cardsOnPage.length).toEqual(4);
     // should match LAST 4, and NOT include the "oldest"
     expect(cardNames).toEqual(
-        ['mockFile5.ipynb', 'mockFile4.ipynb', 'mockFile3.ipynb', 'mockFile2.ipynb']);
+        ['mockFile5', 'mockFile4', 'mockFile3', 'mockFile2']);
```
calbach: I'm not sure whether this was intentional, but if so could you please split this change out? THis feels like an entirely separable 3rd PR in one.

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -105,140 +104,234 @@ describe('NotebookRedirectComponent', () => {
...
+    expect(fixture.componentInstance.leoUrl).toMatch(environment.leoApiUrl);
   }));
 
-  it('should display "Initializing" until ready', fakeAsync(() => {
+  it('should set spinner "Initializing" until ready', fakeAsync(() => {
     updateAndTick(fixture);
     fixture.detectChanges();
-    expect(spinnerText()).toContain('Initializing');
+    const initBox = fixture.debugElement.queryAll(By.css('#Initializing'))[0];
+    expect(initBox.children[0].nativeElement === spinner());
```
calbach: These spinner checks feel way too dependent on specific HTML structure.  Just query for existence of `"#initializing .spinner"`

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -105,140 +104,234 @@ describe('NotebookRedirectComponent', () => {
...
+        .children[0].nativeElement).toBe(spinner());
+
   }));
 
+  it('should escape notebook names', fakeAsync(() => {
+    blockingClusterStub.cluster.status = ClusterStatus.Running;
+    updateAndTick(fixture);
+    fixture.detectChanges();
+    tick(10000);
+    expect(fixture.componentInstance.leoUrl).toMatch('blah%20blah');
```
calbach: This still isn't reproducing the original bug. The initial value needs to contain an escape sequence. Please just use my original test value which was taken from an actual bug we saw: `1%2B1.ipynb`

Also, is it not possible to inspect the iframe URL? THat would be preferred to digging into the component properties.

ui/src/app/views/notebook-redirect/component.ts
UrsaStutsman: It has not in my testing but I will try doing that.

ui/src/app/views/notebook-redirect/component.ts
UrsaStutsman: So I think the replaceState function handles that (since it has in all manual and automated testing). But will add to triple make sure.

ui/src/app/views/recent-work/component.spec.ts
UrsaStutsman: When doing this ticket I noticed that the mocks and what was on the site differed, I checked with Karthik and this is the way he wants it. Since I was doing a lot of tweaking/fixing of how we were dealing with notebook names these tests were failing. So I fixed them.

ui/src/app/views/notebook-redirect/component.spec.ts
UrsaStutsman: So I've tried that strategy before, and had my browser break. I think that was the problem Jackie was running into. Which is why she set it up like this. I just tried again, and ran into the same problem. So leaving as is until we figure out why Jasmine can't seem to handle multiple css identifiers

ui/src/app/views/notebook-redirect/component.ts
calbach: I suspect this does automatically encode the URI if required just as setting the window.location also does that. This is not sufficient to cover the scenario in my test case in the test case I've been commenting on. If that comment thread still doesn't make sense, we can discuss more on slack.

ui/src/app/views/notebook-redirect/component.spec.ts
UrsaStutsman: Ok. So in manual testing if I set a new notebook to the name '1%2B1.ipynb'. A new notebook is created with the name '1+1.ipynb' and our url shows '/1%252B1.ipynb', the iframe source shows '1+1.ipynb', and the breadcrumb shows '1%2B1.ipynb'.... The only issue I see with any of that is possibly the breadcrumb. I'm going to update the test to look at the iframe src but expect '1+1.ipynb'.

ui/src/app/views/recent-work/component.spec.ts
calbach: I guessed as much. I don't think it has any real relationship to this PR other than you happened to be working on these things at the same time, so it would be cleaner to just send a separate PR for it.

ui/src/app/views/notebook-redirect/component.ts
calbach: To be clear, I don't care very much whether it updates the window title from "creating" - I just wanted to verify it wasn't setting it to the empty string.

ui/src/app/views/notebook-redirect/component.spec.ts
calbach: Sounds a lot like this issue: https://precisionmedicineinitiative.atlassian.net/browse/RW-1424 - have you already tried the workaround I listed there?

ui/src/app/views/notebook-redirect/component.spec.ts
calbach: OK, that doesn't look to be reproducing the conditions of the bug. I think the behavior of the new notebook modal is probably fine, but to see the bug you need to have a notebook that actually has a name (in GCS) with a URL encoded escape sequence in it, i.e. your actual notebook should have name `1%2B1.ipynb`.

You can probably get there by specifying `1%252B1.ipynb` as the new notebook name in Workbench.

It's an edge case, but we have actually hit this bug before. Likely originally it was generated by uploading a file into Jupyter, or creating/renaming via the Jupyter UI itself.

ui/src/app/views/notebook-redirect/component.spec.ts
UrsaStutsman: Yes

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -105,140 +105,289 @@ describe('NotebookRedirectComponent', () => {
...
+    });
+  }));
+
+
+  it('should escape notebook names', fakeAsync(() => {
+    blockingClusterStub.cluster.status = ClusterStatus.Running;
+    updateAndTick(fixture);
+    fixture.detectChanges();
+    tick(10000);
+    expect(fixture.debugElement.queryAll(By.css('#leo-iframe'))[0]
```
calbach: nit: `query()`  instead of `queryAll()[0]`

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -105,140 +105,289 @@ describe('NotebookRedirectComponent', () => {
...
+      .properties['src']).toMatch(/1\+1.ipynb/);
```
calbach: If this assertion is actually passing then I'm perplexed, I would have expected the doubly encoded form. Yet from the manual testing you described it sounds like things are working properly.

PR #1400, opened 2018-10-25
[RW-1498][risk = no] [P3] Empty search results in autocomplete doesn't hide when search input is empty

smazumd:
[P3] Empty search results in autocomplete doesn't hide when search input is empty


ui/src/app/cohort-search/search-bar/search-bar.component.html
```diff
@@ -12,7 +12,8 @@
      (keyup.arrowdown)="moveDown()"
      [formControl]="searchTerm"/>
   </div>
-  <div *ngIf="noResults" class="dropdown-menu">
+
+  <div *ngIf="noResults && searchTerm.value.length !== 0" class="dropdown-menu">
```
dolbeew: This covers most scenarios, but it still shows the message in some cases when searchTerm.value has a length of 1. Seems to work better in search-bar.component.ts in the else statement ~line 132:
`this.noResults = false;`

PR #1402, opened 2018-10-25
[DB-89][Risk=high] Blrubenstein/data browser block auth

blrubenstein:
This is almost certainly not final, I'm sure there's lots of shortcuts that I took getting here, but wanted to throw what I have up for review. This is functional auth interceptor for the data browser.


public-api/src/main/java/org/pmiops/workbench/config/WorkbenchEnvironment.java
```diff
@@ -0,0 +1,29 @@
+package org.pmiops.workbench.config;
+
+import com.google.appengine.api.utils.SystemProperty;
+import com.google.appengine.api.utils.SystemProperty.Environment.Value;
+
+public class WorkbenchEnvironment {
```
calbach: Is this file needed?

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+    if (authorizationHeader == null || !authorizationHeader.startsWith("Bearer ")) {
+      log.warning("No bearer token found in request");
+      response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
+      return false;
+    }
+
+    String token = authorizationHeader.substring("Bearer".length()).trim();
+
+    if (token.equals("null")) {
+      throw new RuntimeException();
```
calbach: Is this a real situation?

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    Profile profile = privateWorkbenchService.getMe(token);
+    if (configProvider.get().firecloud.enforceRegistered) {
+      if (profile.getIdVerificationStatus() != IdVerificationStatus.VERIFIED) {
+        throw new RuntimeException();
```
calbach: Probably meant as a placeholder, but should return UnauthorizedException with a recognizable error message we can pick out on the client if needed.

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+   * using Spring's SecurityContext.
+   * @param handler The Swagger-generated ApiController. It contains our handler as a private
+   *     delegate.
+   */
+  @Override
+  public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler)
+      throws Exception {
+    // Clear the security context before we start, to make sure we're not using authentication
+    // from a previous request.
+    SecurityContextHolder.clearContext();
```
calbach: This can be removed probably since we're not really using it in data browser

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.web.method.HandlerMethod;
+import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+
+/**
+ * Intercepts all non-OPTIONS API requests to ensure they have an appropriate auth token.
+ *
+ * Checks handler methods for annotations like
```
calbach: Not applicable to Data Browser

public-api/src/main/java/org/pmiops/workbench/privateWorkbench/PrivateWorkbenchConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private static final String END_USER_API_CLIENT = "endUserApiClient";
+
+  @Bean(name=END_USER_API_CLIENT)
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
+  public ApiClient workbenchApiClient(WorkbenchConfig workbenchConfig) {
+    ApiClient apiClient = new ApiClient();
+    apiClient.setBasePath(workbenchConfig.server.apiBaseUrl);
+    apiClient.setDebugging(workbenchConfig.firecloud.debugEndpoints);
+    apiClient.setConnectTimeout(10000);
```
calbach: Did this do anything? I thought 10s might have actually been the default anyways

public-api/src/main/java/org/pmiops/workbench/privateWorkbench/PrivateWorkbenchConfig.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+@org.springframework.context.annotation.Configuration
+public class PrivateWorkbenchConfig {
+
+  private static final String END_USER_API_CLIENT = "endUserApiClient";
+
+  @Bean(name=END_USER_API_CLIENT)
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
+  public ApiClient workbenchApiClient(WorkbenchConfig workbenchConfig) {
```
calbach: Did you check whether the initial approach you were using worked (injecting request here)? I do like that one best as it most closely resembles our similar usages in Workbench.

public-api/src/main/resources/private-workbench.yaml
```diff
@@ -0,0 +1,93 @@
+# This file is for endpoints we do NOT intend to generate client libraries for
```
calbach: I would put a disclaimer about the intended lifespan on this swagger file and trim some of the fat in these comments. If we wanted something longer term we'd probably want to avoid duplicating our own swagger yaml

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
blrubenstein: So, I agree with adding a recognizeable error message. There is currently no structure in place for custom exceptions in the data browser, so if we want to do that there is some pretty large scope creep. I am not against that, but we should be aware of that if we want to do it.

public-api/src/main/java/org/pmiops/workbench/privateWorkbench/PrivateWorkbenchConfig.java
blrubenstein: Will try it.

public-api/src/main/java/org/pmiops/workbench/privateWorkbench/PrivateWorkbenchConfig.java
blrubenstein: Not sure. I can leave it out and if there are a lot of issues, we can revisit later.

public-api/src/main/resources/private-workbench.yaml
blrubenstein: Yeah, I'll throw some more comments in there.

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
blrubenstein: Talked to Karthik. I opened this ticket to handle exceptions in reasonable ways:
`https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=22&projectKey=DB&view=planning&selectedIssue=DB-120`

calbach: Should be easy enough to copy the approach used on line 75 of this file. Didn't mean to fixate on specifically using a custom exception type which might not be setup in Data Browser, I just don't want to be returning 500 errors in an expected/common case.

blrubenstein:
Looking for review from @calbach on approach.

Looking for review from @SrushtiGangireddy to make sure she has understanding and knows what follow ups there are.

public-api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
blrubenstein: Oh! I missed that. Will do.

PR #1404, opened 2018-10-25
[RW-1530][RW-1533][risk=no] Concept set bug fixes

calbach:
- Concept set deletion now redirects to concept sets list page
- Concept removal uses disabled FAB UX
- Fix disablement of "add to set" FAB and selected count (+some refactoring)
- Minor wording tweak on "add to set" modal

### DIFF
--- a/ui/src/app/views/concept-homepage/component.spec.ts
+++ b/ui/src/app/views/concept-homepage/component.spec.ts
@@ -220,7 +220,7 @@ describe('ConceptHomepageComponent', () => {
     expect(buttonText).toBe('Add (1) to set');
   }));
 
-  fit('should clear selected count after adding', fakeAsync(() => {
+  it('should clear selected count after adding', fakeAsync(() => {
     const de = fixture.debugElement;
 
     simulateClick(fixture, de.query(By.css('.standard-concepts-checkbox')).children[0]);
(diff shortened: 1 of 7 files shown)


calbach:
I realized I should probably have added a regression test here, at least for the bug. Added.

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -224,4 +219,30 @@ describe('ConceptHomepageComponent', () => {
...
+    updateAndTick(fixture);
+    const dataRow = de.queryAll(By.css('.concept-row'));
+    const checkBox = dataRow[0].queryAll(By.css('.datagrid-select'))[0]
+        .query(By.css('.checkbox')).children;
+    simulateClick(fixture, checkBox[0]);
+    updateAndTick(fixture);
+
+    simulateClick(fixture, de.query(By.css('.sliding-button')));
+    updateAndTick(fixture);
+
```
NehaBroad: Nit: I understand the  button "btn-primary" is for the add concept set to concept model with the option of 'existing concept set'. Do you think it will be informative to add comment here. 
My only concern is since it is indirectly relaying on the stubConceptSets from ConceptSetsServiceStub and in case anyone make any changes to it or change domain (i know they shouldnt) it will not be very easy to debug why this test failed because of that.

calbach: Good catch, thanks. I didn't actually think about that when I was writing the test.

I decided to just create a new concept set instead to sidestep this issue.

PR #1415, opened 2018-10-29
[RW-1538][Risk=no] fix for concept set recent resource cards

UrsaStutsman:
fix for concept set recent resource cards not displaying correctly

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
@@ -87,7 +87,6 @@ public class UserMetricsController implements UserMetricsApiDelegate {
           .etag(Etags.fromVersion(conceptSet.getVersion()))
           .lastModifiedTime(conceptSet.getLastModifiedTime().getTime())
           .creationTime(conceptSet.getCreationTime().getTime())
-//          .domain(conceptSet.getDomain())
           .description(conceptSet.getDescription())
           .id(conceptSet.getConceptSetId())
           .name(conceptSet.getName());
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -74,6 +76,25 @@ public Cohort apply(org.pmiops.workbench.db.model.Cohort cohort) {
...
+      @Override
+      public ConceptSet apply(org.pmiops.workbench.db.model.ConceptSet conceptSet) {
+        if (conceptSet == null) {
+          return null;
+        }
+        ConceptSet result = new ConceptSet()
+          .etag(Etags.fromVersion(conceptSet.getVersion()))
+          .lastModifiedTime(conceptSet.getLastModifiedTime().getTime())
+          .creationTime(conceptSet.getCreationTime().getTime())
+//          .domain(conceptSet.getDomain())
```
calbach: Is this intentionally commented out? Probably should be including the domain if we can. If it's not working for some reason then please add a TODO for now.

UrsaStutsman: Yeah, not working, but also on reflection not sure what good it would do since this is only used for user recent resource, and we don't store the domain (or need to) for that.

calbach: We don't currently use it, but it is a bit weird to return this resource with missing information. Anyways, I think it's probably fine for now but probably deserves a comment

PR #1422, opened 2018-10-29
[RW-1180][Risk=no] Update emails

blrubenstein:

### DIFF
--- a/api/src/main/resources/emails/welcomeemail/content.html
+++ b/api/src/main/resources/emails/welcomeemail/content.html
@@ -96,7 +96,7 @@
       width: 20%;
     }
     th.long {
-      width: 80%
+      width: 60%
     }
   </style>
 </head>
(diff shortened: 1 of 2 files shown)


api/src/main/resources/emails/welcomeemail/content.html
```diff
@@ -95,23 +92,79 @@
...
+      <td>
+        <div class="user-info">${PASSWORD}</div>
+      </td>
+    </tr>
+    <tr>
+      <td colspan="3">
+        <div class="text">Quick guide on how to get started:</div>
+      </td>
+    </tr>
+    <tr class="card" href="https://myaccount.google.com/signinoptions/two-step-verification">
```
calbach: I was imagining that just this section (bullets + steps) would be in the table. If you had things this way, then perhaps the table autosizing would work properly and avoid the padding?

blrubenstein: You think just that section? I can try two tables as well, one for the bullets, and one for username/password.

blrubenstein:
Will do on ticket.

PR #1423, opened 2018-10-30
[RW-1447][risk=no] Collect More User Info in Profile & Account Creation

rushtong:
## Addresses 
Jira: https://precisionmedicineinitiative.atlassian.net/browse/RW-1447
Profile Mocks: https://projects.invisionapp.com/d/#/console/12820961/286774501/preview
Create Account Mocks: https://projects.invisionapp.com/d/#/console/12820961/324620184/preview

## Changes
* Added two new fields (current position, organization) to the user profile and account creation page
* Relabeled _Given Name_ to _First Name_
* Relabeled _Family Name_ to _Last Name_
* Relabeled _Area of Research_ to _Current Research Work_ and added to the account creation page
* Added validation for the new/changed fields
* Refactored `AccountCreationComponent` to be more canonical, i.e. see `ProfilePageComponent`

### Blank Form
create_account_1

### Normal Input
create_account_2_2

### Error Input
create_account_3_2

### Normal Input
profile_1_3

### Error Input
profile_2_2


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -393,9 +396,15 @@ private User initializeUserIfNeeded() {
     // profile, since it can be edited in our UI as well as the Google UI,  and we're fine with
     // that; the expectation is their profile in AofU will be managed in AofU, not in Google.
 
-    userService.createUser(request.getProfile().getGivenName(),
+    userService.createUser(
+        request.getProfile().getGivenName(),
```
blrubenstein: On speaking with Lou, we are doing a field change to first name and last name, rather than given and family.

ui/src/app/views/profile-page/component.css
```diff
@@ -140,3 +140,17 @@ hr {
...
   justify-content: space-between;
 }
+
+.field-wrapper-aligned {
+  display: flex;
+  align-items: flex-start;
+}
+
+.error-margin {
+  margin-bottom: 16px;
```
blrubenstein: I would do a `margin-bottom: calc(total-margin - error-height)` to be a little more self documenting on what this is doing

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: Updated code and screenshots to cover the new labeling. I did not rename the objects themselves, just the labeling.

rushtong:
@blrubenstein - Lots of small stuff added, PTAL, thank you!

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: Okay, that works. Do you mind making a tech debt ticket or something to update the fields to match?

ui/src/app/views/account-creation/component.spec.ts
```diff
@@ -69,60 +50,92 @@ describe('AccountCreationComponent', () => {
...
-    simulateInput(page.fixture, page.usernameField, '.username');
+    const usernameField = fixture.debugElement.query(By.css('#username'));
+    simulateEvent(fixture, usernameField, 'focus');
+    simulateInput(fixture, usernameField, '.username');
     tick(300);
-    updateAndTick(page.fixture);
-    expect(page.fixture.debugElement.query(By.css('#username-invalid-error'))).toBeTruthy();
-    expect(page.usernameField.classes.unsuccessfulInput).toBeTruthy();
+    updateAndTick(fixture);
+    expect(fixture.debugElement.query(By.css('#username-invalid-error'))).toBeTruthy();
```
blrubenstein: This is one of the 'this could crash jasmine' query expects. Since this was just refactoring out the page object, probably okay to leave, but just wanted to show you what it looked like. And also it may make sense to throw a tech debt ticket for this in the backlog as well.

ui/src/app/views/profile-page/component.spec.ts
```diff
@@ -76,11 +78,46 @@ describe('ProfilePageComponent', () => {
...
     expect(fixture).toBeTruthy();
   }));
+
+  it('handles long given name errors', fakeAsync(() => {
+    console.log(fixture);
+    simulateInput(
+      fixture, fixture.debugElement.query(By.css('#givenName')), randomString(81));
+    tick(300);
+    updateAndTick(fixture);
+    expect(fixture.componentInstance.givenNameValid).toBeFalsy();
```
blrubenstein: If there is a way, I would probably recommend making sure the error element shows up on the page, rather than just reading the component variable. (From a test philosophy standpoint)

Same comment for the following tests.

ui/src/app/views/account-creation/component.spec.ts
rushtong: Ah ... I think this was one of the ones I refactored ... will update.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: Can do.

ui/src/app/views/account-creation/component.spec.ts
rushtong: Done.

ui/src/app/views/profile-page/component.spec.ts
rushtong: Done.

rushtong:
@blrubenstein - Updated tests, PTAL, thank you!

PR #1424, opened 2018-10-30
[RW-1549][risk=no] Open Data Browser in new tab

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1549

### DIFF
--- a/ui/src/app/views/signed-in/component.html
+++ b/ui/src/app/views/signed-in/component.html
@@ -39,9 +39,8 @@
           [class.active]="homeActive"
           routerLink="/"
           (click)="sidenavToggle = false">Home</button>
-      <!--Data browser won't show the sidenav, so active is irrelevant-->
-      <a class="btn btn-primary nav-link"
-          [attr.href]="publicUiUrl">Browse Data</a>
+      <button class="btn btn-primary nav-link"
+          (click)="openPublicBrowser()">Browse Data</button>
       <button class="btn btn-primary nav-link"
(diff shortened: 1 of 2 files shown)


ui/src/app/views/signed-in/component.html
```diff
@@ -39,9 +39,8 @@ <h6 style="padding: 0.5rem;">
           [class.active]="homeActive"
           routerLink="/"
           (click)="sidenavToggle = false">Home</button>
-      <!--Data browser won't show the sidenav, so active is irrelevant-->
-      <a class="btn btn-primary nav-link"
-          [attr.href]="publicUiUrl">Browse Data</a>
```
NehaBroad: <a target="_blank"> will also open the url in a new tab

rushtong: A few things about why I didn't choose that approach. I did try it, but ended up going a different way.

1. The anchor tag in this menu is the only one - all others are buttons. I like consistency.
2. The anchor tag uses angular syntax to set the href, and the same treatment for target does not open the link in a new tab.
3. This code *does* work as you suggest, but it adds a layer of inconsistency that I don't like:
```<a class="btn btn-primary nav-link" target="_blank" [attr.href]="publicUiUrl">Browse Data</a>```
4. We use `window.open(some_url, '_blank')` in quite a few other places in the code.

I'd like to keep this as it is now if that' OK with you.

PR #1426, opened 2018-10-30
[risk=no] public database version change

SrushtiGangireddy:
Changed the public db version to use the latest version.

### DIFF
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -10,6 +10,6 @@
     "releaseNumber": 1,
     "numParticipants": 946237,
     "cdrDbName": "cdr20181023",
-    "publicDbName": "public20181015"
+    "publicDbName": "public20181026"
   }
 ]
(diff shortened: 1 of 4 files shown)


api/config/cdr_versions_prod.json
```diff
@@ -10,6 +10,6 @@
     "releaseNumber": 1,
     "numParticipants": 946237,
     "cdrDbName": "cdr20181023",
-    "publicDbName": "public20181015"
+    "publicDbName": "public20181026"
```
calbach: This doesn't exist in prod yet. Are these changes breaking, or can prod stay on the old one for now?

api/config/cdr_versions_test.json
```diff
@@ -22,6 +22,6 @@
     "releaseNumber": 2,
     "numParticipants": 946237,
     "cdrDbName": "cdr20181023",
-    "publicDbName": "public20181022"
+    "publicDbName": "public20181026"
```
calbach: What about staging?

SrushtiGangireddy: Updated and pushed the change.

api/config/cdr_versions_prod.json
SrushtiGangireddy: I guess i would leave it as it is for now since databrowser is not functional in prod anyway.

PR #1427, opened 2018-10-30
[risk=no][RW-1548] update demo notebooks and cohorts

jacmrob:


jacmrob:
I believe this has to merge before running the `./api/project.rb deploy-gcs-demos --project=all-of-us-workbench-test` to put these new demo notebooks on test.  And then we will have to run that command for each environment as we release

api/demos/UsingCohortsAndConcepts.ipynb
```diff
@@ -0,0 +1,812 @@
+{
```
calbach: I realize this was just a download and push from the data science workspace and I didn't actually review the content, but I would put spaces in this notebook name to match the others.

api/demos/diabetes-cases.json
```diff
@@ -1,50 +1,41 @@
...
+              "modifiers": []
             }
           ]
         }
       ],
       "excludes": []
     }
   }
-}
+}
```
rushtong: nit: newline

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -416,6 +422,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
             dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getWorkspaceId(), dbCohort.getCohortId()));
       }
     }
+
+    List<JSONObject> demoConceptSets = cloudStorageService.readAllDemoConceptSets();
+    for (JSONObject conceptSet: demoConceptSets) {
+      ConceptSet dbConceptSet = new ConceptSet();
+      JSONArray conceptIdsJSON = conceptSet.getJSONArray("concept_ids");
```
rushtong: This can throw `JSONException` ... should catch this

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -416,6 +422,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+      Set<Long> conceptIds = new HashSet<>();
+      for (int i = 0; i < conceptIdsJSON.length(); i++) {
```
rushtong: Whenever I see a for loop, I want to convert it to a streaming lambda. It generally reads a lot cleaner and saves intermediary collection objects. Something like (untested) `conceptIdsJSON.toStream().map(id -> Long.valueOf(id)).collect(Collectors.toSet());`

api/demos/bloodpressure.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "concept_set",
+  "concept_set": {
+    "name": "bloodpressure",
+    "description": "demo",
+    "type": "demo",
```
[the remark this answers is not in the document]
jacmrob: do you know what the domain should be set to?

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -26,6 +26,19 @@
 
   final Provider<WorkbenchConfig> configProvider;
 
+  private List<JSONObject> readJSONObject(String filterType, String filterField) {
```
rushtong: nit: name, I think this should be plural.

api/demos/bloodpressure.json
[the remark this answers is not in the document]
jacmrob: i see that these concepts are Measurements but the value should be a short... any idea where those domains are defined?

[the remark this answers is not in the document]
calbach: (resolved on slack)

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
jacmrob: it seems like (from what i'm reading) this is the most concise way to cast a jsonarray to a set, since there is no way to do the lambda function without error on something that could be any type

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
jacmrob: how does that work?  am confused about syntax

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
rushtong: This compiles for me:
```
      List<Object> conceptIdsList = conceptIdsJSON.toList();
      Set<Long> conceptIds = conceptIdsList
          .stream()
          .map(Object::toString)
          .map(Long::valueOf)
          .collect(Collectors.toSet());
```

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
rushtong: Something like this
```suggestion
      try {
        JSONArray conceptIdsJSON = conceptSet.getJSONArray("concept_ids");
      } catch (JSONException e) {
        // log exception
        // return an internal server error 
      }
```

PR #1428, opened 2018-10-30
[RW-1552][risk=no] Prevent extra api calls for create review page

dolbeew:
Move create review modal to show on cohort list page instead of cohort review to prevent unnecessary api calls


dolbeew:
Closing since this approach doesn't work from the recent resources list on the homepage without significant changes to the api. Will keep the modal in cohort review and clean up api calls from there.

PR #1434, opened 2018-10-31
[RW-1560][risk=no] refactor denormalized tables and cohort chart queries

freemabd1:
refactor denormalized tables and cohort chart queries


dolbeew:
Verbal approval given

PR #1438, opened 2018-11-01
[RW-1519][risk=low] Add playground Jupyter extension, push with GCS script

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1519
Design doc: https://docs.google.com/document/d/1-JDFypXveP8e5Npvj9nDnpSpHqNBMdQPYepzqQrWAq4/edit

- Implement a small Jupyter UI extension for playground mode: https://projects.invisionapp.com/d/#/console/12820961/328089475/preview
- Include the extension on all cluster creation requests
- Push the new extension to GCS along side the initialization script
- Rename the "-scripts" bucket to a more generic "-cluster-resources"

Before merging: I will create the new GCS buckets in all environments
After this PR is live for 2 releases: I will destroy the old "-scripts" buckets


rushtong:
@calbach - FYI, I'm in meetings all day and can't get to testing this in depth until Friday.

calbach:
Thanks for the heads up - I will switch over to @blrubenstein in this case.

calbach:
FYI: I will also fix up the Codacy issues, though the JS style issues we're triggering here are highly questionable.

api/cluster-resources/playground-extension.js
```diff
@@ -0,0 +1,43 @@
...
+    // This prefix must be kept in sync with the Workbench localization API,
+    // see https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+    const nbPath = Jupyter.notebook.notebook_path;
+    if (!nbPath.startsWith('workspaces:playground/')) {
+      return;
+    }
+
+    // Disable UI controls/notifications relating to saving.
+
+    // "notbook" is an intentional typo to match Jupyter UI HTML.
```
blrubenstein: This isn't a comment on your PR but... really? Wow.

blrubenstein:
Is there any way for me to get around this to open a read-only workspace not in playground mode? We should make sure to be clear with restricting access levels.

calbach:
RE when do we run this: We will always load it for new clusters. We only change the UI if the path matches, as you said.

RE `workspaces:playground`: this is what is outlined in the design document: https://docs.google.com/document/d/1-JDFypXveP8e5Npvj9nDnpSpHqNBMdQPYepzqQrWAq4/edit.

RE getting around this: Yes, technically someone could manipulate the notebook page URL (or bookmark a link and then be demoted to READ access). If we wanted to be very careful we could put a check at the UI or API layer for this, but I don't think it's too bad if it happens. To the user, it will look like they have opened the notebook normally, but all delocalization will fail (consistent with what will happen if they suddenly lose access in the middle of their work).

api/cluster-resources/playground-extension.js
calbach: Yep. They'd probably break people at this point if they changed, given how extensive some of the UI extensions probably are.

PR #1440, opened 2018-11-01
[No ticket][Risk=no] Fixing notebook display names

UrsaStutsman:
Fixing notebook display names

### DIFF
--- a/ui/src/app/views/resource-card/component.ts
+++ b/ui/src/app/views/resource-card/component.ts
@@ -154,17 +154,26 @@ export class ResourceCardComponent implements OnInit {
     switch (this.resourceType) {
       case ResourceType.NOTEBOOK: {
         this.workspacesService.deleteNotebook(this.wsNamespace, this.wsId, $event.name)
-          .subscribe(() => this.onUpdate.emit());
+          .subscribe(() => {
+            this.onUpdate.emit();
+            this.deleteModal.close();
+          });
         break;
       }
       case ResourceType.COHORT: {
         this.cohortsService.deleteCohort(this.wsNamespace, this.wsId, $event.id)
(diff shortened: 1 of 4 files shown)


ui/src/app/views/resource-card/component.ts
```diff
@@ -190,9 +182,8 @@ export class ResourceCardComponent implements OnInit {
         break;
       }
       case ResourceType.NOTEBOOK: {
-        const nbUrl = '/workspaces/' + this.wsNamespace + '/' + this.wsId + '/notebooks/'
-          + encodeURIComponent(this.resourceCard.notebook.name);
-        window.open(nbUrl, '_blank');
+        this.route.navigate(['workspaces', this.wsNamespace, this.wsId, 'notebooks',
```
calbach: I think this line can be reverted and included in your other PR (we still want to open a new tab for now)

ui/src/app/views/resource-card/component.ts
```diff
@@ -154,28 +154,20 @@ export class ResourceCardComponent implements OnInit {
...
       case ResourceType.CONCEPT_SET: {
         this.conceptSetsService.deleteConceptSet(this.wsNamespace, this.wsId, $event.id)
-          .subscribe(() => {
-            this.onUpdate.emit();
-            this.deleteModal.close();
-          });
+          .subscribe(() => this.onUpdate.emit());
       }
     }
+    this.deleteModal.close();
```
calbach: Was this change intentional? Looks like it may have reverted something that Bru did

ui/src/app/views/resource-card/component.ts
UrsaStutsman: I swear I included this change in my other PR. I'll double check now and add it back in if it got left out somehow.

PR #1448, opened 2018-11-02
[RW-1518][Risk=no] Introduce Playground Mode from Read only notebook access

UrsaStutsman:
Rough draft, for CH to help troubleshoot.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -202,20 +202,30 @@ public class ClusterController implements ClusterApiDelegate {
     if (!projectName.equals(body.getWorkspaceNamespace())) {
       workspacePath = body.getWorkspaceNamespace() + ":" + body.getWorkspaceId();
     }
+    //if playgroundMode = true this should be "workspaces:playground/
     String apiDir = "workspaces/" + workspacePath;
+    if (body.getPlaygroundMode()) {
+      apiDir = "workspaces:playground/" + workspacePath;
+    }
+    log.info(String.format("apiDir: %s", apiDir));
     String localDir = "~/" + apiDir;
 
     // Always localize config files; usually a no-op after the first call.
     Map<String, String> localizeMap = new HashMap<>();
-    localizeMap.put(
+    //skip this if playgroundMode = true
+    log.info(String.format("before if: %s", body.getPlaygroundMode()));
+    if (!body.getPlaygroundMode()) {
+      log.info(String.format("in if"));
+      localizeMap.put(
         localDir + "/" + DELOCALIZE_CONFIG_FILENAME,
         jsonToDataUri(new JSONObject()
-            .put("destination", gcsNotebooksDir)
-            .put("pattern", "\\.ipynb$")));
+          .put("destination", gcsNotebooksDir)
+          .put("pattern", "\\.ipynb$")));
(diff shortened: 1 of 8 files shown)


ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -119,10 +124,17 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     private leoNotebooksService: NotebooksService,
     private jupyterService: JupyterService,
     private sanitizer: DomSanitizer
-  ) {}
+  ) {
+    const wsData: WorkspaceData = this.route.snapshot.data.workspace;
+    this.accessLevel = wsData.accessLevel;
```
calbach: I realize this is still WIP, but would save us work down the road if this were instead a queryParam per the ticket description. Eventually we want to support launching **any** notebook in playground mode (as an option; it would only be the default behavior for read-only notebooks) , so ideally the playground mode decision should not happen here.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -202,20 +202,30 @@ public void setUserService(UserService userService) {
     if (!projectName.equals(body.getWorkspaceNamespace())) {
       workspacePath = body.getWorkspaceNamespace() + ":" + body.getWorkspaceId();
     }
+    //if playgroundMode = true this should be "workspaces:playground/
```
calbach: nit: would clean up this comment or just rm, pretty much restates the code

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -202,20 +202,30 @@ public void setUserService(UserService userService) {
...
     String apiDir = "workspaces/" + workspacePath;
+    if (body.getPlaygroundMode()) {
+      apiDir = "workspaces:playground/" + workspacePath;
+    }
+    log.info(String.format("apiDir: %s", apiDir));
```
calbach: nit: rm debug loggers

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -122,10 +127,10 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit(): void {
-    this.initializeProgressMap();
     this.wsNamespace = this.route.snapshot.params['ns'];
     this.wsId = this.route.snapshot.params['wsid'];
     this.creating = this.route.snapshot.queryParams['creating'] || false;
+    this.playground = this.route.snapshot.queryParams['playgroundMode'] || false;
```
calbach: I commented on my own PR: I'm worried about this actually being a string which might mess things up downstream, please double check with manual testing.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -165,8 +170,9 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
         let localizeObs: Observable<string>;
         // This will contain the Jupyter-local path to the localized notebook.
         if (!this.creating) {
-          this.incrementProgress(Progress.Copying);
-          localizeObs = this.localizeNotebooks([this.notebookName])
+          this.progress = Progress.Copying;
```
calbach: incrementProgress

PR #1450, opened 2018-11-02
Playground branch suggestions

calbach:

### DIFF
--- a/ui/src/app/views/notebook-redirect/component.ts
+++ b/ui/src/app/views/notebook-redirect/component.ts
@@ -124,20 +124,13 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     private leoNotebooksService: NotebooksService,
     private jupyterService: JupyterService,
     private sanitizer: DomSanitizer
-  ) {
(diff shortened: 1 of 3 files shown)


ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -124,20 +124,13 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
...
+  ) {}
 
   ngOnInit(): void {
-    if (this.accessLevel == WorkspaceAccessLevel.READER) {
-      this.playground = true;
-    }
     this.wsNamespace = this.route.snapshot.params['ns'];
     this.wsId = this.route.snapshot.params['wsid'];
     this.creating = this.route.snapshot.queryParams['creating'] || false;
+    this.playground = this.route.snapshot.queryParams['playgroundMode'] || false;
```
calbach: @UrsaStutsman  I would double check this works locally - I'm not confident of the type of the queryParam here. If it's a string, this line might instead need to be:

`this.route.snapshot.queryParams['playgroundMode'] === 'true'`

PR #1457, opened 2018-11-05
[RW-1578][Risk=no] Fix users not signed in showing up

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserController.java
@@ -60,9 +60,10 @@ public class UserController implements UserApiDelegate {
         .ofNullable(Sort.Direction.fromStringOrNull(sortOrder))
         .orElse(Sort.Direction.ASC);
     Sort sort = new Sort(new Sort.Order(direction, DEFAULT_SORT_FIELD));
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -60,7 +60,10 @@ public UserController(UserService userService) {
         .ofNullable(Sort.Direction.fromStringOrNull(sortOrder))
         .orElse(Sort.Direction.ASC);
     Sort sort = new Sort(new Sort.Order(direction, DEFAULT_SORT_FIELD));
-    List<User> users = userService.findUsersBySearchString(term, sort);
+    List<User> users = userService.findUsersBySearchString(term, sort)
+        .stream()
+        .filter(user -> user.getFirstSignInTime() != null)
```
calbach: Please add a comment, it is not obvious that we'd be using first sign in time as a proxy for Firecloud user creation.

calbach: nit: Checking freeTierBillingProjectName (!= `null`) or freeTierBillingProjectStatus (!= `NONE`) is also imperfect, but might be semantically closer to what we're actually looking for here (FC initialization).

blrubenstein: Will change.

PR #1458, opened 2018-11-06
[RW-1233][risk=low] Implement an account creation nonce to secure resend/updateContactEmail

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1233

- Store a random int64 on account creation and return it
- Require that unauthenticated "resendWelcomeEmail" and "updateContactEmail" calls provide this value, effectively limiting those calls to the same account creation client session.
- Implement a `FakeLongRandom` class for testing purposes, faking out the whole Random implementation was too onerous.


rushtong:
Overall, I like this approach. Will let Ursa do the technical review.

PR #1465, opened 2018-11-06
[RW-1561][RISK-NO]Select concept

NehaBroad:


ui/src/app/views/concept-homepage/component.ts
```diff
@@ -240,18 +267,37 @@ export class ConceptHomepageComponent implements OnInit {
...
+        });
+        this.conceptTable.selectedConcepts = [];
       });
   }
 
   selectConcept(concepts) {
-    this.selectedConcept = concepts;
+    if (concepts && concepts.length > 1 ) {
+      concepts = concepts.filter((value) => !(value === undefined || value === null));
+      this.selectedConcept = concepts.filter(
```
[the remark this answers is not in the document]
NehaBroad: clr-dataGrid is sending duplicated values with lots of empty array elements as well. This is happening after using select attribute of row.

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -197,7 +212,19 @@ export class ConceptHomepageComponent implements OnInit {
   setConceptsAndVocabularies() {
     const cacheItem = this.conceptsCache.find(
         conceptDomain => conceptDomain.domain === this.selectedDomain.domain);
-    this.concepts = cacheItem.items;
+    this.concepts = cacheItem.items.map((concept) => {
```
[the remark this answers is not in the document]
NehaBroad: Filtering of concepts have been moved to helper function this.filterConceptSelection(this.concepts);

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -312,4 +347,14 @@ export class ConceptHomepageComponent implements OnInit {
   domainLoading(domain) {
     return this.searchLoading || !this.completedDomainSearches.includes(domain.domain);
   }
+
+  filterConceptSelection(concepts) {
+    concepts.forEach((concept) => {
+      const exist = this.selectConceptMap.get(concept.conceptId);
+      if (exist) {
+        concept.selected = true;
```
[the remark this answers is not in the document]
NehaBroad: It has been set to False by default in both the instance its called.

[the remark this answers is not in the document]
calbach: I saw that - I was suggesting it might be cleaner to handle both things here rather than using a two-step process. I may be missing something else though

calbach:
FYI: @UrsaStutsman I don't receive any notification when you update the PR description, please use a comment on the PR.

ui/src/app/views/concept-table/component.ts
```diff
@@ -19,6 +19,7 @@ export class ConceptTableComponent {
   selectedConcepts: Array<any> = [];
 
   onSelectedChanged($event) {
+    $event = $event.filter((el, i, a) => i === a.indexOf(el));
```
calbach: What is this doing and is it actually needed? Please remove or add a comment.

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -351,4 +350,21 @@ export class ConceptHomepageComponent implements OnInit {
...
+      return {
+        ...concept,
+        selected: false
+      };
+    });
+    this.filterConceptSelection(conceptInfos);
+    return conceptInfos;
+  }
+
+  filterConceptSelection(concepts: ConceptInfo[]) {
```
calbach: private

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -69,13 +70,15 @@ export class ConceptHomepageComponent implements OnInit {
...
-  concepts: Array<Concept> = [];
+  concepts: Array<ConceptInfo> = [];
   conceptsCache: Array<ConceptCacheSet> = [];
   completedDomainSearches: Array<Domain> = [];
   placeholderValue = '';
   vocabularies: Array<VocabularyCountSelected> = [];
   wsNamespace: string;
   wsId: string;
+  selectConceptMap = new Map<number, ConceptInfo>();
+  conceptsToAdd: Concept[];
```
calbach: It would be better to dynamically compute this than risk the state going out of sync. If that's too much effort and we're still planning to move off clarity data grid, fine to leave for now.

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -69,13 +70,15 @@ export class ConceptHomepageComponent implements OnInit {
...
   selectedConceptDomainMap: Map<String, number>  = new Map<string, number>();
-  concepts: Array<Concept> = [];
+  concepts: Array<ConceptInfo> = [];
   conceptsCache: Array<ConceptCacheSet> = [];
   completedDomainSearches: Array<Domain> = [];
   placeholderValue = '';
   vocabularies: Array<VocabularyCountSelected> = [];
   wsNamespace: string;
   wsId: string;
+  selectConceptMap = new Map<number, ConceptInfo>();
```
calbach: nit: this is only used to check for presence, so a Set of IDs would have sufficed.

calbach: L361 is n^2 now so would prefer a set. Feel free to build it locally at L360 if you don't need it elsewhere (one of my other comments suggested this)

ui/src/app/views/concept-homepage/component.ts
```diff
@@ -358,8 +358,8 @@ export class ConceptHomepageComponent implements OnInit {
 
   private filterConceptSelection(concepts: ConceptInfo[]) {
     concepts.forEach((concept) => {
-      concept.selected = this.conceptsToAdd.map(c => c.conceptId)
-        .indexOf(concept.conceptId) > -1;
+      const conceptSet = new Set(this.conceptsToAdd.map(c => c.conceptId));
```
calbach: This is still n^2 if you build the set inside the for loop

ui/src/app/views/concept-homepage/component.ts
UrsaStutsman: Ok, think I got it.

ui/src/app/views/concept-homepage/component.ts
calbach: (move outside this callback)

UrsaStutsman: Haha - I knew that too. Yeah sorry about that. Fixed now

PR #1469, opened 2018-11-07
[RW-1528][risk=low] Verify notebooks exist before returning recent resources

calbach:
- Take the first 5 notebooks in the cache and test that they actually exist
- Test via GCS getBlob, this is generally faster than list


calbach:
FYI, filed:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1601
https://precisionmedicineinitiative.atlassian.net/browse/RW-1602

PR #1471, opened 2018-11-07
[RW-1590][risk=no]correct language for research purpose

jacmrob:


rushtong:
@jacmrob  Did you get this from Lou? I'm pretty sure this is not the correct text. See the profile page for the correct version.

rushtong:
I think this PR is meant for https://precisionmedicineinitiative.atlassian.net/browse/RW-1590. Please update the title with [RW-1590][risk=no] so it gets picked up correctly via the release process. 

Whatever the final language is, it should be on both the profile page and the account creation page.

jacmrob:
@rushtong I got this from Karthik - will check with lou

PR #1472, opened 2018-11-07
[risk=no] Safe null check

rushtong:


rushtong:
The latest failing codacy check is a flake.

blrubenstein:
Yeah, I saw that. Seems fine to me, although I've been curious why those flakes are happening.

PR #1475, opened 2018-11-08
[RW-1595][risk =no]Number of participants change in Demographics when add more criterion from dropdown.

smazumd:
Number of participants change in Demographics when add more criterion from dropdown.

### DIFF
--- a/ui/src/app/cohort-search/demographics/demographics.component.ts
+++ b/ui/src/app/cohort-search/demographics/demographics.component.ts
@@ -14,7 +14,6 @@ import {
 } from '../redux';
 
 import {Attribute, CohortBuilderService, Operator, TreeSubType, TreeType} from 'generated';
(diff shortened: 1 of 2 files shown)


ui/src/app/cohort-search/demographics/demographics.component.ts
```diff
@@ -143,10 +145,13 @@ export class DemographicsComponent implements OnInit, OnChanges, OnDestroy {
             .subscribe(sel => this.noSelection = sel)
         );
 
-        this.subscription.add(this.ngRedux
-            .select(totalCount)
-            .subscribe(participantCount => this.count = participantCount)
-        );
+
+      this.subscription = this.count$
+        .subscribe(item => {
```
dolbeew: nit: 'item' doesn't make sense here, maybe name it 'count' or something similar to avoid confusion

PR #1477, opened 2018-11-08
[RW-1604][RW-1075][risk=low] Upgrade GCS and BigQuery out of the stone age

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1075
https://precisionmedicineinitiative.atlassian.net/browse/RW-1604


rushtong:
Do we need to worry about https://github.com/broadinstitute/firecloud-tools/blob/c1033a565514c4f2b88f54e33411bfdbd5fe3beb/install.sh#L11 ?

calbach:
My understanding is that we might execute that script as part of some of the CLI invocations (for some of the Firecloud manual API calls). I don't know whether that dependency is actually necessary for AoU, or if there are other dependencies on that requirement. I don't foresee it conflicting with this change so I think it could probably be updated separately by someone who is more familiar with the usage.

If that's used by AoU exclusively, let me know and I can look at sending a PR or removing that.

calbach:
@rushtong please let me know if you still have concerns around that script per my above comment

rushtong:
@calbach - No, no concerns, 👍

calbach:
Confirmed integration tests pass as well.

PR #1479, opened 2018-11-08
[RW-1424][Risk=no] Updating Jasmine config to prevent breaking on null DebugElement

UrsaStutsman:
This is the needed change to the Jasmine config... Should I also go through specs and change them back where we had changed them to follow the queryAll patter....? Thoughts


rushtong:
My 2 cents - I don't think it's worth going back through tests and rewriting them. This seems like a reasonable config change in and of itself.

PR #1480, opened 2018-11-08
[RW-1404][Risk=no] Fix weird extra whitespace to make CH happy! :)

blrubenstein:

### DIFF
--- a/ui/src/app/views/resource-card/component.css
+++ b/ui/src/app/views/resource-card/component.css
@@ -82,7 +82,8 @@
 
 .card .footer {
   position: absolute;
-  bottom: -0.05rem;
+  /* This deals with clarity's built in bottom border for cards... */
+  bottom: -1px;
 }
 
 .last-modified {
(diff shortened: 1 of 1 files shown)


ui/src/app/views/resource-card/component.css
```diff
@@ -82,7 +82,7 @@
 
 .card .footer {
   position: absolute;
-  bottom: .01rem;
+  bottom: -0.05rem;
```
calbach: I was so confused as to where the card border comes from. After digging a bit it runs out it's actually being applied via Clarity's CSS, where it puts a `1px` border-bottom on to `.card`s. So I'd use -1px here instead to match that, and put a comment as well - please.

https://github.com/vmware/clarity/blob/56d3f500ed6db1e5a3ffc3442b65ad4a8337997f/src/clr-angular/layout/_card.clarity.scss#L45

blrubenstein: Done! PTAL

PR #1481, opened 2018-11-08
[risk=no][RW-1523] profile edit client side input validation

jacmrob:


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -160,19 +160,25 @@ private String createFirecloudUserAndBillingProject(User user) {
...
-  private User saveUserWithConflictHandling(User user) {
-    if (user.getGivenName().length() > 80) {
-      throw new BadRequestException("Given Name length exceeds character limit. (80)");
-    }
-    if (user.getFamilyName().length() > 80) {
-      throw new BadRequestException("Family Name length exceeds character limit. (80)");
-    }
-    if (user.getCurrentPosition() != null && user.getCurrentPosition().length() > 255) {
-      throw new BadRequestException("Current Position length exceeds character limit. (255)");
+  private void validateStringLength(String field, String fieldName, Integer max, Integer min) {
```
calbach: nit: `int` preferred, unless you need the boxed type for some specific reason (doesn't look like it here).

ui/src/app/views/profile-page/component.ts
```diff
@@ -86,33 +100,62 @@ export class ProfilePageComponent implements OnInit {
 
   get givenNameValid(): boolean {
     if (this.workingProfile) {
-      return this.valid(this.workingProfile.givenName, 80);
+      return (ProfilePageComponent.valid(this.workingProfile.givenName, 80));
```
calbach: nit: would drop these extra parens throughout

ui/src/app/views/profile-page/component.ts
```diff
@@ -86,33 +100,62 @@ export class ProfilePageComponent implements OnInit {
...
+    }
+  }
+
+  get givenNameNotEmpty(): boolean {
```
calbach: existing issue but the function doesn't really match this return type. Should `return false;` at the end. If you let execution reach the end of a JS function without returning, that's the same as `return undefined;`

ui/src/app/views/profile-page/component.ts
```diff
@@ -55,6 +55,20 @@ export class ProfilePageComponent implements OnInit {
   @ViewChild(BugReportComponent)
   bugReportComponent: BugReportComponent;
 
+  static valid(val: string, len: number): boolean {
```
calbach: nit: existing issue but I'd probably make the function name more specific, e.g. `validLength`

ui/src/app/views/profile-page/component.spec.ts
```diff
@@ -124,4 +124,49 @@ describe('ProfilePageComponent', () => {
...
+    expect(fixture.debugElement.queryAll(By.css('#organizationEmpty')).length).toBeTruthy();
+  }));
+
+  it('handles empty organization field', fakeAsync( () => {
+    simulateInput(
+        fixture, fixture.debugElement.query(By.css('#areaOfResearch')), randomString(0));
+    tick(300);
+    updateAndTick(fixture);
+    expect(fixture.componentInstance.currentResearchNotEmpty).toBeFalsy();
+    expect(fixture.debugElement.queryAll(By.css('#currentResearchEmpty')).length).toBeTruthy();
```
calbach: You should be able to go back to the old way of doing this with is slightly more readable - i.e. `expect(de.query(...)).toBeTruthy()`; @UrsaStutsman submitted a fix that should prevent your browser from crashing on failed debugElement expectations.

ui/src/app/views/profile-page/component.spec.ts
```diff
@@ -124,4 +124,49 @@ describe('ProfilePageComponent', () => {
...
+  it('handles empty organization field', fakeAsync( () => {
+    simulateInput(
+        fixture, fixture.debugElement.query(By.css('#organization')), randomString(0));
+    tick(300);
+    updateAndTick(fixture);
+    expect(fixture.componentInstance.organizationNotEmpty).toBeFalsy();
+    expect(fixture.debugElement.queryAll(By.css('#organizationEmpty')).length).toBeTruthy();
+  }));
+
+  it('handles empty organization field', fakeAsync( () => {
```
calbach: update

ui/src/app/views/profile-page/component.spec.ts
```diff
@@ -124,4 +124,49 @@ describe('ProfilePageComponent', () => {
       fixture.debugElement.queryAll(By.css('#currentPositionInvalid')).length).toBeTruthy();
   }));
 
+  it('handles empty givenName field', fakeAsync( () => {
+      simulateInput(
+          fixture, fixture.debugElement.query(By.css('#givenName')), randomString(0));
```
calbach: nit: `s/randomString(0)/''/` ?

ui/src/app/views/profile-page/component.spec.ts
```diff
@@ -124,4 +124,49 @@ describe('ProfilePageComponent', () => {
       fixture.debugElement.queryAll(By.css('#currentPositionInvalid')).length).toBeTruthy();
   }));
 
+  it('handles empty givenName field', fakeAsync( () => {
```
calbach: nit: rm leading space before the callback  `fakeAsync(() =>`

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jacmrob: I used the boxed type because I needed to use the `equals` method

calbach: Why not just use `==` ?

ui/src/app/views/profile-page/component.spec.ts
calbach: Not addressed, if you disagree or do not understand please comment (else I must assume you just missed it)

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jacmrob: that just checks null/not null i believe, not actually checking equality

ui/src/app/views/profile-page/component.spec.ts
jacmrob: sorry i missed this!  not sure what you mean though?

rushtong: `randomString(0)` is the same thing as the empty string.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -160,19 +160,29 @@ private String createFirecloudUserAndBillingProject(User user) {
...
     }
-    if (user.getCurrentPosition() != null && user.getCurrentPosition().length() > 255) {
-      throw new BadRequestException("Current Position length exceeds character limit. (255)");
+    if (field.length() > max) {
+      throw new BadRequestException(String.format("%s length exceeds character limit. (%d)", fieldName, max));
     }
-    if (user.getOrganization() != null && user.getOrganization().length() > 255) {
-      throw new BadRequestException("Organization length exceeds character limit. (255)");
+    if (field.length() < min) {
+      if (min == 1) {
```
calbach: Shouldn't this be 0?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: CH is right - you could use the primitive `int min`, use `==`, which will do the same thing as `equals` does on full objects.

ui/src/app/views/profile-page/component.spec.ts
calbach: Thanks Greg, Jackie and I just synced on Slack as well on this and the boxed type thread.

For other observers, my above syntax was a sed replacement, i.e. suggesting substitution of `randomString(0)` with `''` throughout.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jacmrob: no bc 1 character is the minimum character length.  we don't want to allow blank fields (0 characters)

calbach: OK, I wasn't following the logic initially. This does seem pretty round-about though; could you instead just check `field.length() == 0` ?

jacmrob: true.  i wanted it to be robust to handle different mins in case we decided to implement them down the road, but i thought it would be nice to have a custom message for the empty case.  i can get rid of that though - what do you think?

calbach: I can't really imagine cases other than the two you have here, but I don't really feel strongly about this, I'll leave it up to your judgement.

PR #1484, opened 2018-11-12
[risk=no][RW-1651][RW-1523] Revert "profile edit client side input validation"

calbach:
Reverts all-of-us/workbench#1481 - see comments on there as well as https://precisionmedicineinitiative.atlassian.net/browse/RW-1651

@jacmrob FYI


calbach:
Just reverting for now to avoid any release delays tomorrow, I think the original approach needs to be adjusted slightly and Broad is out today. This will be easy enough to re-revert and fix forward.

PR #1490, opened 2018-11-13
[RW-1601][risk=low] Switch to GCS batch GET, increase max notebooks

calbach:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
@@ -1,7 +1,6 @@
 package org.pmiops.workbench.api;
 
 import com.google.cloud.storage.BlobId;
-import com.google.cloud.storage.StorageException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import java.sql.Timestamp;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -180,23 +179,14 @@ public void setDistinctWorkspaceLimit(int limit) {
...
 
     // Check for existence of recent notebooks. Notebooks reside in GCS so they may be arbitrarily
-    // deleted or renamed without notification to the Workbench. This makes serial GCS requests so
-    // we limit to the first N notebooks.
-    // TODO: Dispatch these requests in batch or in parallel.
+    // deleted or renamed without notification to the Workbench. This makes a batch of GCS requests
+    // so it will scale fairly well, but limit to the first N notebooks to avoid excess GCS traffic.
     // TODO: If we find a non-existent notebook, expunge from the cache.
-    Set<BlobId> foundNotebooks = workspaceFilteredResources.stream()
+    Set<BlobId> foundNotebooks = cloudStorageService.blobsExist(workspaceFilteredResources.stream()
```
blrubenstein: Since we are using `toList()` below, should this be a `List<BlobId>`?

calbach: `toList()` is called on the method argument (not the return value). I changed the spacing here to hopefully make this more obvious.

PR #1503, opened 2018-11-15
[RW-1523][risk=no][P2] Profile validation

jacmrob:
- no empty fields client-side validation in create profile
- no empty fields client-side validation in update profile
- no empty fields server-side validation in create profile
- no empty fields server-side validation in update profile
- enforce character cap on research purpose

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -160,19 +160,29 @@ public class ProfileController implements ProfileApiDelegate {
     return createFirecloudBillingProject(user);
   }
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -420,6 +422,7 @@ private User initializeUserIfNeeded() {
         request.getProfile().getOrganization(),
         request.getProfile().getAreaOfResearch()
     );
+    validateProfileFields(user);
```
calbach: I think this should be an easy change:

- Switch validation from `User` -> `Profile`
- Move the validation ahead of the user creation

This would be nice because it would prevent creation of an invalid user.

PR #1504, opened 2018-11-15
[risk=no] DB-HighlightSearchTextIssueFix

SrushtiGangireddy:
1. Fixed the regex expression to be highlighted.
2. Fixed the way multiple keywords were modified in concept service.

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
@@ -89,7 +89,7 @@ public class ConceptService {
                 }
                 else{
                     if(key.length() < 3){
-                        temp.add(new String("\"" + key + "\""));
+                        temp.add(new String(key));
                     }else{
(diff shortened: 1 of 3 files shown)


common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -69,29 +71,38 @@ public static String modifyMultipleMatchKeyword(String query){
...
+                if(si-1 >= 0 && si-1 <= query.length()){
+                    String pre = query.substring(si-1,si);
+                    if(pre.equals("+")) {
+                        temp.add(new String(pre+key));
+                        continue;
+                    }
+                }
+                if(key.contains("-") && !temp.contains(key)){
+                    temp.add(key);
+                }else if(key.contains("*") && key.length() > 1){
```
blrubenstein: Nit: Some of the spacing here is wacky, actually. Do you mind addressing some of that? (Space after close curly braces and before open curly braces. And probably before and after operators.

PR #1507, opened 2018-11-15
[RW-1671][risk=no] Save Research Purpose Update Workspace 

rushtong:
## Addresses: 
https://precisionmedicineinitiative.atlassian.net/browse/RW-1671

## Changes
Previously, update workspace ignored any changes to the research purpose that were submitted from the UI, but handled it correctly on create and clone. This change adds the same save-research-purpose functionality back into update workspace.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
@@ -379,8 +379,8 @@ public class WorkspacesControllerTest {
     researchPurpose.setPopulation(true);
     researchPurpose.setPopulationOfFocus("population");
     researchPurpose.setAdditionalNotes("additional notes");
-    researchPurpose.setTimeRequested(new Long(1000));
-    researchPurpose.setTimeReviewed(new Long(1500));
+    researchPurpose.setTimeRequested(1000L);
+    researchPurpose.setTimeReviewed(1500L);
     researchPurpose.setReviewRequested(true);
     researchPurpose.setApproved(false);
     Workspace workspace = new Workspace();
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -536,6 +536,44 @@ public void testUpdateWorkspace() throws Exception {
...
+            .reviewRequested(false);
+    ws.setResearchPurpose(rp);
+    UpdateWorkspaceRequest request = new UpdateWorkspaceRequest();
+    request.setWorkspace(ws);
+    ResearchPurpose updatedRp = workspacesController
+            .updateWorkspace(ws.getNamespace(), ws.getId(), request)
+            .getBody()
+            .getResearchPurpose();
+
+    assertThat(updatedRp.getDiseaseFocusedResearch()).isFalse();
```
jacmrob: if most of these things default to false (i assume) should we try this test setting some of them to true?

rushtong: They are all set to true/truthy in the create default workspace method, hence my setting them to false here.

rushtong: You know what I can do though ... is to reset them again in a mixed fashion and ensure they still line up.

jacmrob: ah that makes sense.  no need to change then unless you want to!

rushtong: Ok, thanks ... was having issues making multiple update calls inside the same test. Will go with this and we can always add more tests.

PR #1520, opened 2018-11-19
[RW-1679][Risk=no] Fix grammar

blrubenstein:

### DIFF
--- a/ui/src/app/views/quick-tour-modal/component.ts
+++ b/ui/src/app/views/quick-tour-modal/component.ts
@@ -69,10 +69,10 @@ export class QuickTourModalComponent implements OnInit {
       'Notebooks'];
     this.panels.set(this.panelTitles[0],
       'Welcome to the All of Us Research Workbench!\n\nAll workbench analyses ' +
-      'happen in a “Workspace”. Within a workspace you can select participants ' +
-      'using the “Cohort Builder” tool.  Another tool, the “Concept Set Builder”, ' +
+      'happen in a “Workspace.” Within a Workspace you can select participants ' +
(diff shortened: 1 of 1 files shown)


ui/src/app/views/quick-tour-modal/component.ts
```diff
@@ -69,10 +69,10 @@ export class QuickTourModalComponent implements OnInit {
       'Notebooks'];
     this.panels.set(this.panelTitles[0],
       'Welcome to the All of Us Research Workbench!\n\nAll workbench analyses ' +
-      'happen in a “Workspace”. Within a workspace you can select participants ' +
-      'using the “Cohort Builder” tool.  Another tool, the “Concept Set Builder”, ' +
+      'happen in a “Workspace.” Within a Workspace you can select participants ' +
+      'using the “Cohort Builder” tool.  Another tool, the “Concept Set Builder,” ' +
       'allows you to select data types for analysis.  The cohorts and concept sets ' +
```
UrsaStutsman: More a question than a comment, but do we also want to capitalize these instances of "cohorts" and "concept sets"

blrubenstein: I was not entirely sure, to be honest... I checked with Jen Ayala, who filed the ticket, and it came from the bug bash, with the person who filed being out. I figured I would submit this, which at least makes it consistent within plurals and singulars, and then if they want pluralized versions capitalized they can file a new ticket.

blrubenstein: I'm going to file a product ticket about it.

PR #1524, opened 2018-11-19
[RW-1702][Risk=no] Fix display of concept modal

blrubenstein:
This fixes the mismatch of display from text area and input on the create concept, and makes it so if there are no concept sets in the domain, we keep the UX consistent and display a tooltip to the user regarding the lack of concept sets.


blrubenstein:
I totally agree, although, I tried and couldn't get the tooltip to appear. I think it is because it goes off the modal? I will see what I can do.

PR #1533, opened 2018-11-21
[RW-1667][RISK=LOW][P3] Notebook name with special chars

NehaBroad:


ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -210,6 +211,9 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
   setNotebookNames(): void {
     this.notebookName =
       decodeURIComponent(this.route.snapshot.params['nbName']);
+    this.notebookName = this.notebookName.replace('#', '\\#');
```
blrubenstein: Did you figure out why these two characters in particular were causing a problem? If so, can you throw a comment here?

NehaBroad: I suspect its because it is a # fragment identifier. But i did not get a chance to read more. I'll confirm and add comments

ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -387,7 +387,7 @@ describe('NotebookRedirectComponent', () => {
     fixture.detectChanges();
     tick(10000);
     expect(fixture.debugElement.query(By.css('#leo-iframe'))
-      .properties['src']).toMatch(/1\+1.ipynb/);
+      .properties['src']).toMatch('1%2B1.ipynb');
```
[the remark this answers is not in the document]
calbach: This test was guarding against a specific regression which was reintroduced by this PR. I'm reverting this commit for now, but when fixing forward please leave this test case unmodified.

PR #1542, opened 2018-11-28
[RW-1647][risk=no] [P4] CB - Demo modal Finish button disable while getting calculated number

smazumd:
[P4] CB - Demo modal Finish button disable while getting calculated number

### DIFF
--- a/ui/src/app/cohort-search/modal/modal.component.ts
+++ b/ui/src/app/cohort-search/modal/modal.component.ts
@@ -132,7 +132,6 @@ export class ModalComponent implements OnInit, OnDestroy {
         this.subtype = subtype;
       })
     );
-
     this.originalNode = this.rootNode;
   }
   addSelectionToGroup(selection: any) {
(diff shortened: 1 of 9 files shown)


ui/src/app/cohort-search/modal/modal.component.ts
```diff
@@ -1,4 +1,4 @@
-import {select} from '@angular-redux/store';
+import {NgRedux, select} from '@angular-redux/store';
```
dolbeew: Doesn't look like we're using NgRedux anywhere

ui/src/app/cohort-search/modal/modal.component.ts
```diff
@@ -12,7 +12,7 @@ import {
   activeItem,
   activeParameterList,
   CohortSearchActions,
-  nodeAttributes,
+  nodeAttributes, previewStatus,
```
dolbeew: Move previewStatus to new line

PR #1543, opened 2018-11-28
[RW-1371][Risk=no] Adding sorting/filtering to Admin ID Verification Page

UrsaStutsman:
Quick PR to add sorting filtering to the Admin ID verification page


blrubenstein:
So, I have one concern, which is what the default sorting behavior is? Does it follow the behavior in the ts file first, and then this just gives more granular control after the page loads? Or does the data table override that default behavior?

PR #1550, opened 2018-11-28
[RW-1754][RW-1667][RISK=LOW][P3] Revert "Notebook name with special chars (#1533)"

calbach:
This reverts commit ce6ac784bbb9517d53107dc4eb2ebc086384dddd.

The original PR may have fixed issues with certain notebook names, but broke our URI encoding model. This includes notebooks with spaces in the name, e.g. all of our demo notebooks.


calbach:
Thanks @osgsusan

PR #1551, opened 2018-11-29
[RW-1756][risk=low] Get all docker images onto the same gcloud version, update build image

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1756

CC: @blrubenstein

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 defaults: &defaults
   docker:
-    - image: allofustest/workbench:buildimage-0.0.12
+    - image: allofustest/workbench:buildimage-0.0.13
   working_directory: ~/workbench
 java_defaults: &java_defaults
   <<: *defaults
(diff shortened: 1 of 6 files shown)


calbach:
Grr.. will check out the failure tomorrow - doesn't seem directly related but I may have upgraded something inadvertently in the new image. Let me know if anything obvious jumps out at you.

```
#!/bin/bash -eo pipefail
./project.rb start-local-api && ./project.rb run-local-api-tests && ./project.rb stop-local-api
Starting API server...
+ gradle appengineStart
...

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':findMainClass'.
> org.gradle.api.tasks.SourceSetOutput.getClassesDir()Ljava/io/File;
```

rushtong:
The only thing that looks unusual to me is downgrading cloud sdk from 217 to 215. There was an update to java sdk from 216 to 217 (https://cloud.google.com/sdk/docs/release-notes) but that doesn't seem like it would cause this failure.

I also can't run the command that fails `./project.rb start-local-api`. It fails for me while generating the app engine xml which requires having `envsubst` on the system - I'm guessing it's only available in the docker image circle uses.

calbach:
Solved the mystery. `apt-get install gradle` recently changed behavior from 4.0 -> 5.0. I've pinned the gradle version for now (eventually we can consider upgrading across the board).

I also switched back to 217 - seems to work just as well.

PR #1559, opened 2018-11-30
[risk=no] DB-142 (Adding source concept check during count generation)

SrushtiGangireddy:
1. Standard concepts which are also source concepts were being counted as two different concepts (even after the standard concepts having source count in source count value)
2. This was causing two rows of data and disrupting graphs.

### DIFF
--- a/api/db-cdr/generate-cdr/run-achilles-queries.sh
+++ b/api/db-cdr/generate-cdr/run-achilles-queries.sh
@@ -155,7 +155,7 @@ union all
 select 0 as id,3000 as analysis_id,CAST(vo1.visit_source_concept_id AS STRING) as stratum_1,'Visit' as stratum_3,
 COUNT(distinct vo1.PERSON_ID) as count_value,COUNT(distinct vo1.PERSON_ID) as source_count_value
 from \`${BQ_PROJECT}.${BQ_DATASET}.visit_occurrence\` vo1
-where vo1.visit_concept_id != vo1.visit_source_concept_id
+where vo1.visit_source_concept_id not in (select distinct visit_concept_id from \`${BQ_PROJECT}.${BQ_DATASET}.visit_occurrence\`)
 group by vo1.visit_source_concept_id"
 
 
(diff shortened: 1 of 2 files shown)


api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -573,6 +583,7 @@ COUNT(distinct p1.PERSON_ID) as source_count_value from \`${BQ_PROJECT}.${BQ_DAT
 on p1.person_id = co1.person_id
 where (extract(year from current_date()) - p1.year_of_birth) >= 18 and
 (extract(year from current_date()) - p1.year_of_birth) < 30
+and co1.procedure_source_concept_id not in (select distinct procedure_concept_id from \`${BQ_PROJECT}.${BQ_DATASET}.procedure_occurrence\`)
```
calbach: Some of these look slightly different than the others as it didn't already have `procedure_concept_id != prcedure_source_concept_id` - please double check

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -614,6 +625,7 @@ COUNT(distinct p1.PERSON_ID) as count_value,COUNT(distinct p1.PERSON_ID) as sour
 from \`${BQ_PROJECT}.${BQ_DATASET}.person\` p1 inner join \`${BQ_PROJECT}.${BQ_DATASET}.drug_exposure\` co1
 on p1.person_id = co1.person_id
 where co1.drug_concept_id != co1.drug_source_concept_id
+and co1.drug_source_concept_id not in (select distinct drug_concept_id from \`${BQ_PROJECT}.${BQ_DATASET}.drug_exposure\`)
```
calbach: I believe you can remove the above line e.g. `co1.drug_concept_id != co1.drug_source_concept_id` from most of these queries, it should be redundant with the new check you've added.

api/db-cdr/generate-cdr/run-achilles-queries.sh
```diff
@@ -155,7 +155,7 @@ union all
 select 0 as id,3000 as analysis_id,CAST(vo1.visit_source_concept_id AS STRING) as stratum_1,'Visit' as stratum_3,
 COUNT(distinct vo1.PERSON_ID) as count_value,COUNT(distinct vo1.PERSON_ID) as source_count_value
 from \`${BQ_PROJECT}.${BQ_DATASET}.visit_occurrence\` vo1
-where vo1.visit_concept_id != vo1.visit_source_concept_id
+where vo1.visit_source_concept_id not in (select distinct visit_concept_id from \`${BQ_PROJECT}.${BQ_DATASET}.visit_occurrence\`)
```
calbach: You hadn't modified these queries on the previous pass - double checking this change was intentional

SrushtiGangireddy: Yeah. Just thought the new check would do better job than the previous one
(in case the source concept is not equal to concept in this row but it is standard concept in some other row)

PR #1571, opened 2018-12-03
[risk=no] Fixing highlighting issue

SrushtiGangireddy:
1. Split the word and highlighted only parts of the text that match the search string.


public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -23,4 +24,25 @@ export class HighlightSearchComponent implements OnChanges {
     }
     this.words = this.text.split(' ');
   }
+  public highlight(word) {
+    const splitWords = word.split(' ');
```
calbach: I'm confused - the original `words` array was generated by splitting on `' '`. YOu then iterate through that list, and once again split on space here. Shouldn't this list always be length 1?

public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -23,4 +24,25 @@ export class HighlightSearchComponent implements OnChanges {
...
+    const temp = [];
+    for (let i = 0; i < splitWords.length; i++) {
```
calbach: Took me a while to understand this, I'm not sure whether this would be clearer, but what do you think about something more like:

```
const matches = word.match(new RegExp(this.reString, 'g'));
const splits = word.split(new RegExp(this.reString));

// Interleave matches/splits.
for (let i = 0; i < matches.length; i++) {
  temp.push(splits[i], matches[i]);
}
temp.push(splits[splits.length-1]);
```

(also requires the `reString` is changed to be a string - current var name is a bit misleading)

public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -14,6 +14,7 @@ export class HighlightSearchComponent implements OnChanges {
   @Input() indNum: number;
   words: string[];
   reString: RegExp;
+  toBeHighlightedWords: string[];
```
calbach: unused?

calbach:
Also if you can add a unit test that covers this error case, that would be great.

public-ui/src/app/highlight-search/highlight-search.component.ts
SrushtiGangireddy: Fixed.

public-ui/src/app/highlight-search/highlight-search.component.ts
SrushtiGangireddy: Fixed.

ui/src/app/highlight-search/highlight-search.component.spec.ts
```diff
@@ -5,21 +5,25 @@ import { HighlightSearchComponent } from './highlight-search.component';
...
   beforeEach(() => {
     fixture = TestBed.createComponent(HighlightSearchComponent);
     component = fixture.componentInstance;
     fixture.detectChanges();
   });
-
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+  it('highlight_split_test', () => {
```
calbach: Some suggestions here: https://github.com/all-of-us/workbench/compare/SrushtiGangireddy/RW-1757...ch/hilight-test?expand=1

At a high level the changes are:
- Actually render the component (use a test wrapper component to achieve this)
- Try to make assertions about the actually rendered HTML, rather then probing internal methods

I'm a bit confused by the output though. Seems like there's a lot of extra spans being produced here. I'm wondering whether you need this two-level behavior at all in the component. Could you just split the whole "searchWords" string by using the searchTerm matchString regexp?

public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -21,20 +21,26 @@ export class HighlightSearchComponent implements OnChanges {
...
-    const splits = word.split(new RegExp(this.matchString));
-    const temp = [];
-    if (matches) {
+    const matches = this.text.match(new RegExp(this.matchString, 'gi'));
+    const splits = this.text.split(new RegExp(this.matchString, 'i'));
+    if (matches && this.searchTerm) {
       for (let i = 0; i < matches.length; i++) {
-        temp.push(splits[i], matches[i]);
+        // did not want to highlight whole of disorder of (, when searched by disorder
+        const tempSplitwords = splits[i].split(' ');
```
calbach:
```
this.words.push(...tempSplitWords);
```

or

```
this.words.push(...splits[i].split(' '));
```

public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -21,20 +21,26 @@ export class HighlightSearchComponent implements OnChanges {
...
-    const matches = word.match(new RegExp(this.matchString, 'g'));
-    const splits = word.split(new RegExp(this.matchString));
-    const temp = [];
-    if (matches) {
+    const matches = this.text.match(new RegExp(this.matchString, 'gi'));
+    const splits = this.text.split(new RegExp(this.matchString, 'i'));
+    if (matches && this.searchTerm) {
       for (let i = 0; i < matches.length; i++) {
-        temp.push(splits[i], matches[i]);
+        // did not want to highlight whole of disorder of (, when searched by disorder
```
calbach: I'm confused - why would a matched value contain spaces unless the search term also contained spaces? I don't see why it's necessary to split on space at this point.

public-ui/src/app/highlight-search/highlight-search.component.ts
```diff
@@ -25,20 +25,10 @@ export class HighlightSearchComponent implements OnChanges {
...
-        }
-        const tempMatchWords = matches[i].split(' ');
-        for (let j = 0; j < tempMatchWords.length; j++) {
-          this.words.push(tempMatchWords[j]);
-        }
-      }
-      const tempSplitWords = splits[splits.length - 1].split(' ');
-      for (let j = 0; j < tempSplitWords.length; j++) {
-        this.words.push(tempSplitWords[j]);
+        this.words.push(splits[i]);
```
calbach: nit: can combine `this.words.push(splits[i], matches[i]);`

PR #1572, opened 2018-12-03
Convert quick tour to React

jacmrob:


ui/angular.json
```diff
@@ -0,0 +1,182 @@
+{
```
calbach: Did you need to upgrade Angular as part of this change? If that's a pre-requisite, would be nice to split out this task and have someone tackle it independent of the React stuff.

ui/angular.json
```diff
@@ -0,0 +1,182 @@
...
+            }
+          }
+        },
+        "serve": {
+          "builder": "@angular-devkit/build-angular:dev-server",
+          "options": {
+            "browserTarget": "aou-workbench-ui:build"
+          },
+          "configurations": {
+            "dev": {
```
calbach: "dev" and "production" aren't real environments AFAIK

ui/src/app/views/homepage/component.ts
```diff
@@ -68,13 +68,17 @@ export class HomepageComponent implements OnInit, OnDestroy {
...
 
   constructor(
     private profileService: ProfileService,
     private profileStorageService: ProfileStorageService,
     private route: ActivatedRoute,
     private router: Router,
-  ) {}
+  ) {
+    // create bound methods to use as callbacks
+    this.closeQuickTour = this.closeQuickTour.bind(this);
```
calbach: This should happen by virtue of this being an instance method on the class. Is this actually necessary?

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,222 @@
...
+          Cohort(s) and Concept Set(s) into the Notebook and can assist with basic analyses.
+          <br/><br/>
+         For example, you can launch a Notebook
+          to import your “diabetes cases” Cohort and then select your “biometrics” Concept Set, to
+          get biometrics data for the participants in your Cohort. You can then analyze the data to
+          study correlation between hypertension and diabetes.</div>,
+      image: '/assets/images/notebooks.png'
+    }];
+
+class QuickTourReact extends React.Component<any, any> {
```
calbach: What do the generics here correspond to?

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,222 @@
...
+  checkImg = '/assets/images/check.svg';
+  expandIcon = '/assets/icons/expand.svg';
+
+  constructor(props: Object) {
+    super(props);
+    this.state = {selected: 0, fullImage: false, numPanels: 5};
+  }
+
+  previous(): void {
+    this.setState({selected: this.state.selected - 1});
```
calbach: Based on my limited understanding of React, when you reference `state` you should instead use the callback invocation form as this update may be applied asynchronously:

```
this.setState(state => {
  return {selected: state.selected - 1};
});
```

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,222 @@
...
+  lastButtonText(): string {
+    return (this.state.selected === 4 ? 'Close' : 'Next');
+  }
+
+  toggleImage(): void {
+    this.setState({fullImage: !this.state.fullImage});
+  }
+
+  render() {
+      return  <>
```
calbach: What is this for? I haven't seen it in other React code (I haven't looked at much)

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  constructor() {}
+
+  ngOnInit(): void {
+    ReactDOM.render(React.createElement(QuickTourReact,
+        {learning: this.learning, closeFunction: this.onClose}),
+        document.getElementById('quick-tour'));
+  }
+
+  ngDoCheck(): void {
```
calbach: Interesting .. this would probably get invoked a bunch. Is this necessary? Wouldn't any state updates necessarily go through `setState()` which would then trigger an update?

ui/src/app/views/quick-tour-modal/component.tsx
rushtong: While not strictly necessary - react will shallow merge state when called directly - I think the functional form is [preferred](https://reactjs.org/docs/react-component.html#setstate). 

To take this a step further, if you extracted the functional call into a separate function, you can gain some unit-testability.
```
function decrementSelected(state, props) { // setState takes two args, current state and props
  return {selected: state.selected - 1}
}
... 
this.setState(decrementSelected)
```

ui/src/app/views/homepage/component.ts
rushtong: I've seen the need for this in react to preserve the context of `this` when the bound method is passed down to child components - I'm interested in knowing if this is necessary too.

rushtong:
I'm confused about the larger strategy here. This PR wraps a react component inside an angular component which is a nice step toward full conversion. Do we have some kind of tech doc that describes how we eventually integrate the wrapped components into a full react app?

ui/angular.json
blrubenstein: Yeah, so angular doesn't pick up TSX files in version 5 of the CLI, so it needs to be upgraded to support that. Agreed for splitting it out if we can.

ui/angular.json
blrubenstein: Dev is run when you run dev up as a default value. It should be fine to exclude, but that takes a little more investigation into how the config file works to be sure.

ui/src/app/views/homepage/component.ts
jacmrob: Brett will be able to explain better than me, but I know it had something to do with scoping.  I can ping him for a better explanation?

ui/src/app/views/quick-tour-modal/component.tsx
jacmrob: the state and props.  not sure if we should type them or not; was just trying to get my intellij to display less errors

ui/src/app/views/quick-tour-modal/component.tsx
jacmrob: again, Brett will know better than me, but it was some mechanism for returning multiple divs that are not all children of each other

blrubenstein:
FWIW, I don't think we are planning to merge this soon, this is mostly as a proof of concept? Is that right @jacmrob?

jacmrob:
@blrubenstein yes this is just a proof of concept!  Although i am not opposed to really cleaning it up into something that is mergable.  i think that may warrant more in-depth conversations about style and conventions

ui/src/app/views/homepage/component.ts
jacmrob: From brett: 

> In js, if you do this: `const funReference = this.fn; funReference()`, it loses the `this` binding inside the function. So if we want to pass a reference to the bound function, we have to explicitly create it. That line is effectively replacing the unbound method with a bound version, so we can pass along a reference

ui/src/app/views/quick-tour-modal/component.tsx
jacmrob: from brett: 

> The empty tag is called a Fragment, and it's basically a way to bundle up several DOM nodes together without wrapping them in an actual element. It's nice so we don't end up with extraneous wrappers in the DOM

ui/src/app/views/quick-tour-modal/component.tsx
jacmrob: i think you're right. let me test that.

jacmrob: @calbach so it seems like we do need to call the react component through the `ngDoCheck` method, otherwise we initialize it once on `ngOnInit` but then can't actually open it, since the click to access the quick tour is an angular action.   i think lots of things may have to be set up this way as we do half-angular, half-react, but like the angular wrapper, it's a stopgap during the transition process

ui/src/app/views/quick-tour-modal/component.tsx
panentheos: The function form is indeed preferred, especially in the case of multiple simultaneous updates. In practice, I find that scenario doesn't happen much, so the object form is usually fine.

ui/src/app/views/quick-tour-modal/component.tsx
panentheos: There are two kinds of updates that cause re-renders in react: `state`, which lives and changes inside a component, and `props`, which are passed in from outside. Very often, the state of one component gets passed down as props to another, so the component will see changes to those props over time. This mechanism is basically a systematic way to pass 'state' from angular into the react component as props.

There may be another lifecycle hook that's more appropriate, but the outcome is very much intended: every time the relevant data changes, refresh (e.g. re-render) the react component with updated props.

rushtong:
@panentheos - Agreed, thanks for the explanation!

ui/src/app/views/homepage/component.spec.ts
```diff
@@ -38,9 +38,13 @@ import {RightScrollLightComponent} from 'app/icons/right-scroll-light/component'
 import {RightScrollComponent} from 'app/icons/right-scroll/component';
 import {ShrinkComponent} from 'app/icons/shrink/component';
 
+import * as React from 'react';
+import * as ReactTestUtils from 'react-dom/test-utils'
```
rushtong: We generally avoid * imports in java - I would think we should keep that tradition here for the same reason.

ui/src/app/views/homepage/component.ts
```diff
@@ -110,6 +110,7 @@ export class HomepageComponent implements OnInit, OnDestroy {
   }
 
   public closeQuickTour(): void {
+    console.log("homepage: closing aquick tour");
```
rushtong: nit: remove

ui/src/app/views/homepage/component.spec.ts
jacmrob: so we found that this was the only way to get react to properly import for tests to be able to run.  when we import just as "import React from 'react'" we get this error:
```
ERROR in src/app/views/quick-tour-modal/component.tsx(2,8): error TS1192: Module '"/w/ui/node_modules/@types/react/index"' has no default export.
src/app/views/quick-tour-modal/component.tsx(3,8): error TS1192: Module '"/w/ui/node_modules/@types/react-dom/index"' has no default export.
```

and in the test suite:
```
Uncaught TypeError: Cannot read property 'createElement' of undefined
```

ui/angular.json
```diff
@@ -0,0 +1,182 @@
...
+            "staging": {
+              "browserTarget": "aou-workbench-ui:build:staging"
+            },
+            "stable": {
+              "browserTarget": "aou-workbench-ui:build:stable"
+            },
+            "prod": {
+              "browserTarget": "aou-workbench-ui:build:prod"
+            },
+            "production": {
```
blrubenstein: You should be able to remove production here, we use prod for that.

ui/src/app/views/quick-tour-modal/component.css
```diff
@@ -9,7 +9,7 @@
   z-index: 1040;
 }
 
-.main {
+:host::ng-deep .main {
```
blrubenstein: We can do this as a follow up, but I propose that we move this to what I am proposing, where we use JS/TS objects, rather than CSS classes, which dodges the need for the :host::ng-deep.

ui/src/app/views/quick-tour-modal/component.spec.ts
```diff
@@ -2,22 +2,19 @@ import {DebugElement} from '@angular/core';
...
-import {ShrinkComponent} from 'app/icons/shrink/component';
+import {QuickTourModalComponent, QuickTourReact} from './component';
 
 import {simulateClick, updateAndTick} from 'testing/test-helpers';
 
+/*
 describe('QuickTourModalComponent', () => {
   let fixture: ComponentFixture<QuickTourModalComponent>;
   let de: DebugElement;
+  const panelTitles = ['Introduction', 'Workspaces', 'Cohorts', 'Concepts', 'Notebooks'];
```
blrubenstein: Can we pull this information from the tsx file? If possible I would prefer to avoid the brittleness of hardcoded values here.

ui/src/app/views/quick-tour-modal/component.css
jacmrob: am totally on board, but let's do it as a follow up PR. i can make a ticket for it.

ui/src/app/views/quick-tour-modal/component.spec.ts
jacmrob: good idea.  will make that the case when I write the jest test suite

ui/src/app/views/quick-tour-modal/component.spec.ts
```diff
@@ -2,22 +2,19 @@ import {DebugElement} from '@angular/core';
...
 
-import {QuickTourModalComponent} from './component';
-
-import {ExpandComponent} from 'app/icons/expand/component';
-import {ShrinkComponent} from 'app/icons/shrink/component';
+import {QuickTourModalComponent, QuickTourReact} from './component';
 
 import {simulateClick, updateAndTick} from 'testing/test-helpers';
 
+/*
```
calbach: Should uncomment or delete

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,243 @@
...
+         For example, you can launch a Notebook
+          to import your “diabetes cases” Cohort and then select your “biometrics” Concept Set, to
+          get biometrics data for the participants in your Cohort. You can then analyze the data to
+          study correlation between hypertension and diabetes.</div>,
+      image: '/assets/images/notebooks.png'
+    }];
+
+export class QuickTourReact extends React.Component<any, any> {
+  state: {
+      selected: number,
```
calbach: nit: 2sp indent

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  previous(): void {
+    this.setState(state => {
+      return {selected: state.selected - 1};
+    });
+  }
+
+  next(): void {
+    if (this.state.selected === 4) {
```
calbach: nit: `panels.length` ?

ui/src/app/views/quick-tour-modal/component.tsx
```diff
@@ -0,0 +1,243 @@
...
+      closeFunction: Function
+  };
+
+  checkImg = '/assets/images/check.svg';
+  expandIcon = '/assets/icons/expand.svg';
+  shrinkIcon = '/assets/icons/shrink.svg';
+
+  constructor(props: Object) {
+    super(props);
+    this.state = {selected: 0, fullImage: false, numPanels: 5};
```
calbach: Based on my limited understanding of React, I think state is meant to be scoped to mutable internal state to the component. `numPanels` is static so I think it would be better to just reference panels.length statically throughout.

An argument could be made for `props` for this kind of static state alternatively, but I'm not clear on whether a component should be initializing/modifying their own props.

ui/src/app/views/homepage/component.spec.ts
```diff
@@ -91,6 +94,12 @@ describe('HomepageComponent', () => {
     };
   };
 
+  // From https://stackoverflow.com/questions/36434002/
+  //        new-compilation-errors-with-react-addons-test-utils
+  function renderIntoDocument (reactEl: React.ReactElement<{}>) {
```
calbach: nit: rm extra space before `(`

ui/src/app/views/homepage/component.spec.ts
```diff
@@ -100,31 +109,38 @@ describe('HomepageComponent', () => {
   it('should display quick tour when clicked', fakeAsync(() =>  {
     loadProfileWithPageVisits({pageVisits: [{page: 'homepage'}]});
     updateAndTick(fixture);
+    expect(ReactTestUtils.scryRenderedDOMComponentsWithClass(
+        renderIntoDocument(React.createElement(
```
calbach: Sorry if I missed it, but is there background around why this element would need to be manually rendered like this? Shouldn't that just naturally be triggered by Angular?

ui/angular.json
```diff
@@ -0,0 +1,179 @@
...
+              "node_modules/nouislider/distribute/nouislider.min.css",
+              "src/styles.css"
+            ],
+            "scripts": [
+              "node_modules/mutationobserver-shim/dist/mutationobserver.min.js",
+              "node_modules/@webcomponents/custom-elements/custom-elements.min.js",
+              "node_modules/@clr/icons/clr-icons.min.js"
+            ]
+          },
+          "configurations": {
```
calbach: Please verify you can still run all the compilation variants:

```
ui$ ./project.rb build --env local
ui$ ./project.rb build --env test
...
ui$ ./project.rb build --env prod
```

See comment here about the expected behavior, i.e. we expect optimizations in staging,stable,prod:
https://github.com/all-of-us/workbench/blob/47bfb716834517cfd415153c7429be0d66ebe10b/common-ui/common-ui-devstart.rb#L46

ui/src/app/views/quick-tour-modal/component.tsx
calbach: Thanks @panentheos , good to know.

ui/angular.json
blrubenstein: Testing now...

ui/src/app/views/homepage/component.spec.ts
blrubenstein: The jasmine tests do some weird abstractions and optimizations. I think we ended up figuring out that it blocked the procedural generation like that? Not entirely sure. I can poke around some to figure out the why.

calbach: Doesn't necessarily need to block this PR, but I'm not sure how effective the tests will be if we need to manually render React elements before the parent component can be tested. This might be changing anyways if we're looking at Jest, though.

ui/angular.json
blrubenstein: Okay, this all works!

ui/src/app/views/homepage/component.spec.ts
calbach: FYI: on that particular error message, `import {React} from 'react'` *might* get you further (no promises).

PR #1578, opened 2018-12-06
[risk=no] DB-144 Stackdriver error logging config for db

SrushtiGangireddy:
1. Added error reported to public-ui.

### DIFF
--- a/public-ui/src/app/services/error-reporter.service.ts
+++ b/public-ui/src/app/services/error-reporter.service.ts
@@ -1,0 +1,57 @@
+import {ErrorHandler, Injectable} from '@angular/core';
+import {StackdriverErrorReporter} from 'stackdriver-errors-js';
+
+import {ServerConfigService} from 'app/services/server-config.service';
+import {environment} from 'environments/environment';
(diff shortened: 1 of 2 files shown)


public-ui/src/app/services/error-reporter.service.ts
```diff
@@ -0,0 +1,57 @@
+import {ErrorHandler, Injectable} from '@angular/core';
+import {StackdriverErrorReporter} from 'stackdriver-errors-js';
+
+import {ServerConfigService} from 'app/services/server-config.service';
+import {environment} from 'environments/environment';
+import {ConfigResponse} from 'publicGenerated';
+
+@Injectable()
+export class ErrorReporterService extends ErrorHandler {
```
calbach: You'll also need this: https://github.com/all-of-us/workbench/blob/8893ff7a1b32b25b925013c804b8bc41f5757322/ui/src/app/app.module.ts#L195

SrushtiGangireddy: Added it.

PR #1579, opened 2018-12-06
[risk=no] DB-143 Fetching missing measurement data

SrushtiGangireddy:
1. Change that do not display empty space when gender value measurement data is not present.
2. Query change to fetch missing measurement data.


api/db-cdr/generate-cdr/run-measurement-queries.sh
```diff
@@ -601,9 +612,9 @@ from \`${BQ_PROJECT}.${BQ_DATASET}.measurement\` m1 join \`${BQ_PROJECT}.${BQ_DA
 join measurement_quartile_data on m1.measurement_concept_id=concept
 join \`${BQ_PROJECT}.${BQ_DATASET}.concept\` c1 on m1.measurement_concept_id=c1.concept_id
 join \`${WORKBENCH_PROJECT}.${WORKBENCH_DATASET}.unit_map\` um on m1.unit_source_value=um.unit_source_value
-where m1.measurement_concept_id > 0 and measurement_source_concept_id = measurement_concept_id
+where m1.measurement_concept_id > 0
```
calbach: I don't follow this change (and the other similar ones), can you explain?

SrushtiGangireddy: So, the idea was to generate the count value and the source count value of histogram data of values for the measurement concepts. Me and peter agreed upon not calculating the source counts for measurements for now as the inner query of calculating the source counts gets complicated. I do not remember when this check came up here. Irrespective of this check, the query above the union should generate the counts of standard concepts and the query below should generate the counts of source concepts. This check was only calculating the counts of standard concepts that are also source concepts in particular row. So removed this.

SrushtiGangireddy: Eg. If a concept a has occurred 10 times as a concept and has not occurred as the source concept id in that particular 10 rows then it was not being concept as concept_id != source_concept_id (source concept id in this case is 0).

calbach: Thanks

PR #1590, opened 2018-12-11
[risk=no] Db 149 Hiding the synonyms 

SrushtiGangireddy:
1. Displaying only part of the synonyms unless user wants to view all.

### DIFF
--- a/public-ui/src/app/data-browser/chart/chart.component.ts
+++ b/public-ui/src/app/data-browser/chart/chart.component.ts
@@ -101,7 +101,12 @@ export class ChartComponent implements OnChanges {
             style: this.isGenderIdentityAnalysis()
                 ? this.dbc.GI_DATA_LABEL_STYLE : this.dbc.DATA_LABEL_STYLE,
             distance: this.isGenderIdentityAnalysis() ? 3 : -30,
-            format: '{point.name} {point.percentage:.0f}%',
+            formatter: function () {
+              if (this.percentage < 1) {
+                return this.point.name + ' ' + Number(this.percentage).toFixed(1) + '%';
+              }
+              return this.point.name + ' ' + Number(this.percentage).toFixed(0) + '%';
+            }
           }
         },
         column: {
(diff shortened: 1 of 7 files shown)


public-ui/src/app/data-browser/chart/chart.component.ts
```diff
@@ -101,7 +101,12 @@ export class ChartComponent implements OnChanges {
             style: this.isGenderIdentityAnalysis()
                 ? this.dbc.GI_DATA_LABEL_STYLE : this.dbc.DATA_LABEL_STYLE,
             distance: this.isGenderIdentityAnalysis() ? 3 : -30,
-            format: '{point.name} {point.percentage:.0f}%',
+            formatter: function () {
```
calbach: nit: arrow function is preferred

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -195,5 +190,11 @@ export class EhrViewComponent implements OnInit, OnDestroy {
       this.showGenderIdentity = false;
     }
   }
-
+  public toggleSynonyms(conceptId) {
+    if (!this.showMoreSynonyms[conceptId]) {
```
calbach: nit: would probably just do a `this.showMore[..] = !this.showMore[..]`

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -38,6 +38,7 @@ export class EhrViewComponent implements OnInit, OnDestroy {
   private searchRequest: SearchConceptsRequest;
   private subscriptions: ISubscription[] = [];
   private initSearchSubscription: ISubscription = null;
+  showMoreSynonyms = {};
```
calbach: nit: would add a comment here describing what this property does

public-ui/src/app/views/ehr-view/ehr-view.component.ts
SrushtiGangireddy: Done.

public-ui/src/app/views/ehr-view/ehr-view.component.ts
SrushtiGangireddy: Done thanks.

public-ui/src/app/data-browser/chart/chart.component.ts
SrushtiGangireddy: This unfortunately seems to be tough as the arrow function gets the whole context of the document and unlike chart's context in this function. So, leaving it as it is for now.

calbach: Ah I see - thanks for checking.

PR #1597, opened 2018-12-14
Update Account Creation Success Page to react

blrubenstein:


blrubenstein:
@gjuggler this is the proposal for JS object styles.

ui/src/app/common/common.tsx
```diff
@@ -0,0 +1,34 @@
...
+    fontWeight: 800,
+    fontSize: '1rem',
+  },
+  h4: {
+    fontWeight: 300,
+    color: '#000',
+  }
+};
+
+export const BolderHeader = ({children, style = {}}) =>
```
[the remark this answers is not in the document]
blrubenstein: Isaac was saying something to me about needing a package to do the merging if we did that? Is that correct? I agree with you that if this is doable, I totally prefer this way of doing it.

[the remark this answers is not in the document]
panentheos: No special packages needed. JSX supports that spread syntax for doing basic prop merging. If you need to do something fancy, you can use e.g. lodash to manipulate data structures before doing that, but that should work as-is (try it out).

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+  ngDoCheck(): void {
+    this.renderReactComponent();
+  }
+  renderReactComponent(): void {
+    ReactDOM.render(React.createElement(AccountCreationSuccessReact,
+      {
+        contactEmail: this.contactEmail,
+        account: this.account,
+        signInService: this.signInService,
```
UrsaStutsman: So I'd like to move this and the related logic out of angular (and into react). I've submitted a PR: https://github.com/all-of-us/workbench/pull/1607, that adds the code gen for the fetch stuff which we can use in React. Example of how it is used can be found in this PR: https://github.com/all-of-us/workbench/pull/1598/files in the file /ui/src/app/views/account-creation-modals/component.tsx starting around line 60.

blrubenstein:
@jacmrob you mentioned being interested in this when it was ready.
@panentheos @UrsaStutsman this should be good for another review.

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,134 @@
...
+    account: AccountCreationComponent,
+    accountCreationModalsComponent: AccountCreationModalsComponent,
+  };
+
+  constructor(props: Object) {
+    super(props);
+  }
+
+  resendInstructions(): void {
+    this.props.accountCreationModalsComponent.resendInstructions();
```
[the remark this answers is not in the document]
panentheos: I think that sounds like the right order, yes.

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const buttonLinkStyling = {
+  border: 'none',
+  cursor: 'pointer',
+  outlineColor: 'transparent',
+  color: '#2691D0',
+  backgroundColor: 'transparent',
+};
+
+export class AccountCreationSuccessReact extends React.Component<any, any> {
```
jacmrob: should define the type of the state and props and pass it in here instead of using any

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,134 @@
...
+  state: {};
+  props: {
```
jacmrob: see above; use type here

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,134 @@
...
+import {
+  BolderHeader,
+  Header,
+  SmallHeader
+} from 'app/common/common';
+import {AccountCreationModalsComponent} from 'app/views/account-creation-modals/component';
+import {AccountCreationComponent} from 'app/views/account-creation/component';
+import {LoginComponent} from 'app/views/login/component';
+
+const buttonLinkStyling = {
```
jacmrob: should this be within a `styles` block?  is that how we want to define the local styles or are we not doing that anymore?

ui/src/app/views/account-creation-success/component.tsx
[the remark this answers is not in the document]
blrubenstein: That sounds good to me.

ui/src/app/views/account-creation-success/component.tsx
blrubenstein: I thought I'd pushed that up... might just be local. Will push after other changes have been addressed.

ui/src/app/views/account-creation-success/component.tsx
blrubenstein: Probably should! Will do.

blrubenstein:
@UrsaStutsman would like a final look and new approval, now that your account creation modal work has been integrated.

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,166 @@
...
+interface AccountCreationSuccessState {
+  resendModal: boolean;
+  updateModal: boolean;
+  contactEmail: string;
+}
+
+export class AccountCreationSuccessReact
+    extends React.Component<AccountCreationSuccessProps, AccountCreationSuccessState> {
+  state: AccountCreationSuccessState;
+  props: AccountCreationSuccessProps;
```
panentheos: Specifying types in the angle brackets above will propagate those types onto `props` and `state`, so we don't need to repeat them here. Which means they can then be inlined, rather than split out and named.

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <AccountCreationResendModalReact
+        username={this.props.username}
+        creationNonce={this.props.creationNonce}
+        resend={this.state.resendModal}
+        closeFunction={() => this.closeModals()}
+      />
+      <AccountCreationUpdateModalReact
+        username={this.props.username}
+        creationNonce={this.props.creationNonce}
+        passNewEmail={(newEmail: string) => this.updateEmail(newEmail)}
```
panentheos: We could inline `updateEmail` here, since it's small and only used once.

ui/src/app/views/account-creation-success/component.tsx
```diff
@@ -0,0 +1,166 @@
...
+    this.state = {
+      contactEmail: this.props.contactEmailOnCreation,
+      resendModal: false,
+      updateModal: false,
+    };
+  }
+
+  public closeModals(): void {
+    this.setState({resendModal: false});
+    this.setState({updateModal: false});
```
panentheos: In practice, only one of these will be in effect at a time. We could break this apart and inline each one into its relevant spot below.

PR #1598, opened 2018-12-14
[RW-1851][Risk=low] Account Creation Modals React conversion 

UrsaStutsman:
May have jumped the gun requesting reviews - though Brett's is super helpful, so maybe not. Anyways - this PR will be changing as Scout and I decided this will be the "counter-proposal" to their "styles as js objects" proposal and I'll be incorporating SCSS instead. We think this is a good idea particularly since I've never used SCSS before, and it will give us a good idea as to how "easy" it is for previous non-users to pick up.


ui/package.json
```diff
@@ -12,6 +12,7 @@
     "codegen": "yarn run merge-swagger && yarn run codegen-clean && yarn run codegen-swagger && yarn run codegen-swagger-fetch && yarn run post-codegen && yarn run notebooks-codegen-swagger && yarn run notebooks-post-codegen",
     "merge-swagger": "../api/build.rb merge-yaml",
     "codegen-clean": "rm -rf src/generated src/notebooks-generated",
+    "codegen-swagger-fetch": "java -jar ../aou-utils/swagger-codegen-cli.jar generate --lang typescript-fetch --input-spec ../api/src/main/resources/merged.yaml --output src/generated/fetch",
```
panentheos: This looks like a duplicate line.

ui/src/app/react-components/modals.tsx
```diff
@@ -0,0 +1,83 @@
...
+    justifyContent: 'flex-end' as 'flex-end',
+    marginBottom: '.4rem',
+    marginRight: '.4rem',
+    marginTop: '.4rem'
+  }
+};
+
+export const Modal = ({children, style = {}, ...props}) => {
+  return <div><div style={{...styles.modalBackdrop}}></div>
+    <div {...props} style={{...styles.modalMain, ...style}}>{children}</div></div>;
```
panentheos: This is a good starting point, but as a follow-on we should probably bring in `react-modal`, which handles a lot of the fiddly bits, including things like keyboard shortcuts, focus trapping, etc. We'll still be able to completely customize the appearance and animations.

ui/src/app/react-components/modals.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+export const ModalTitle = ({children, style = {}, ...props}) =>
+  <div {...props} style={{...styles.modalTitle, ...style}}>{children}</div>;
```
panentheos: FYI: Although children has a special syntax in JSX, it otherwise acts like a normal prop, so you could omit it entirely here if you wanted, and it would get passed down implicitly via the `...props`

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+function isBlank(s: string) {
+  return (!s || /^\s*$/.test(s));
+}
+
+class AccountCreationModalsReact extends React.Component<any, any> {
+
+  state: {
+    changingEmail: boolean,
+    resendingEmail: boolean,
```
panentheos: These two pieces of state look like they are never used. Can they be removed?

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+      return;
+    }
+    this.props.passNewEmail(this.state.contactEmail);
+    const request: UpdateContactEmailRequest = {
+      username: this.props.username,
+      contactEmail: this.state.contactEmail,
+      creationNonce: this.props.creationNonce
+    };
+    const args: FetchArgs = ProfileApiFetchParamCreator().updateContactEmail(request);
+    fetch(this.fullUrl(args.url), args.options).then(() => {
```
panentheos: This `fetch` call might fail, so we should have some error handling for that.

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+                  }
+                </td>
+              </tr>
+              </tbody>
+            </table>
+          </ModalBody>
+          <ModalFooter>
+            <button type={'button'} className='btn btn-outline'
+                    onClick={() => this.close()}>Cancel</button>
+            <button id={'change_email'} type={'button'}
```
panentheos: FYI, you don't need braces if the value is a simple string, e.g. `id='change_email' type='button'`

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    className={'btn btn-primary'}
+                    onClick={() => this.updateAndSend()}>Apply</button>
+          </ModalFooter>
+        </Modal>}
+      {this.props.resend &&
+      <Modal id={'resend-instructions>'}>
```
panentheos: That angle bracket in the id looks unintentional.

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+        <ModalTitle>Resend Instructions</ModalTitle>
+        <ModalFooter>
+          <button type={'button'} className='btn btn-outline'
+                  onClick={() => this.close()}>Cancel</button>
+          <button type={'button'} id={'resend_instructions'}
+                  className={'btn btn-primary'}
+                  onClick={() => this.send()}>Send</button>
+        </ModalFooter>
+      </Modal>}
+    </React.Fragment>;
```
panentheos: Bundling the two modals together works ok, but generally I think it would be preferably to split these out into separate components. Particularly because they're independent flows, and they have their own state.

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,206 @@
...
+    };
+    const args: FetchArgs = ProfileApiFetchParamCreator().resendWelcomeEmail(request);
+    fetch(this.fullUrl(args.url), args.options).then(() => {
+      this.setState({resending: false});
+    });
+    this.props.closeFunction();
+  }
+
+  leaveFocusEmail(evt): void {
+    this.setState({emailOffFocus: true, contactEmail: evt.target.value});
```
panentheos: Rather than setting `contactEmail` here, it's more typical to use an `onChange` handler on the input, which will fire changes on every keystroke. You can still track the focus state for the purpose of toggling the error message.

ui/angular.json
```diff
@@ -24,7 +24,8 @@
               "node_modules/@clr/icons/clr-icons.min.css",
               "node_modules/@clr/ui/clr-ui.min.css",
               "node_modules/nouislider/distribute/nouislider.min.css",
-              "src/styles.css"
+              "src/styles.css",
+              "src/assets/scss/styles.scss"
```
panentheos: If we're going to be using scss, we should configure and use CSS Modules so we maintain local scoping of class names.

https://github.com/webpack-contrib/css-loader#modules

This will probably require some tinkering with the webpack config, so it probably deserves its own PR, but we should tackle that before we accrue too many global class names.

UrsaStutsman: Yup - this was actually my exact worry with scss, because I couldn't figure out the webpack config tinkering required. We will definitely be discussing further at our tech talk the day we come back from break. But if you have any ideas about how to get this setup correctly that would be great! (Feel free to commit to this branch, as I'm now treating this branch as the scss demo)

UrsaStutsman:
Closing because we are not currently going in this direction.

PR #1603, opened 2018-12-17
[risk=low] DB-14[5-8] Data Browser search title, survey graph text updates, survey answer table updates

SrushtiGangireddy:
1. Change in the search title.
2. Displaying survey question answers on the graph and removed concept id from the answer text.


public-ui/src/app/views/survey-view/survey-view.component.html
```diff
@@ -64,9 +64,8 @@ <h5 class="box-title">
     </div>
   </div>
   -->
-
+  <div class="info-text"> Note: The survey questions appear in the order in which people actualy took the survey.</div>
```
calbach: actualy -> actually

PR #1605, opened 2018-12-17
[risk=no] adding syn cdr for prod

freemabd1:
adding syn cdr for prod


freemabd1:
Yes, about an hour and 50 mins to export/import

PR #1606, opened 2018-12-17
[risk=no] add popups and tooltips

panentheos:
This is a port of the Terra tooltip and popup library. It handles arbitrary content, and uses smart dynamic positioning to keep the boxes visible. Usage examples:

```
<TooltipTrigger content="hi there">
  <button>Button with tooltip</button>
</TooltipTrigger>
```

```
<PopupTrigger content={<div>Hi there</div>}>
  <button>Button that opens popup</button>
</PopupTrigger>
```


blrubenstein:
Taking this over now (As an FYI so we don't end up with duplicate work)

blrubenstein:
This should be ready for re-review. @jacmrob? Can you review or should we tag someone else in?

ui/src/app/components/popups.tsx
```diff
@@ -0,0 +1,285 @@
...
+        style={{
+          transform: `translate(${position.left}px, ${position.top}px)`,
+          ...styles.popup} as React.CSSProperties}
+      >{children}</div>
+    </PopupPortal>;
+  }
+});
+
+export class PopupTrigger extends React.Component {
+  static defaultProps = {
```
calbach: nit: `static readonly` would probably be better

ui/src/app/components/popups.tsx
```diff
@@ -0,0 +1,285 @@
...
+      onClick
+    } = this.props;
+    const { position } = computePopupPosition({ side, target, element, viewport, gap: 10 });
+    return <PopupPortal>
+      <div
+        onClick={onClick}
+        ref={elementRef}
+        style={{
+          transform: `translate(${position.left}px, ${position.top}px)`,
+          ...styles.popup} as React.CSSProperties}
```
calbach: Note: Having to do this cast everywhere is pretty rough. We should think about how we can avoid making this a standard pattern in the codebase when decide on a convention for our CSS.

ui/src/app/components/popups.tsx
```diff
@@ -0,0 +1,285 @@
...
+        <svg viewBox='0 0 2 1' style={{...getNotchPosition(), ...styles.notch}}>
+          <path d='M0,1l1,-1l1,1Z' />
+        </svg>
+      </div>
+    </PopupPortal>;
+  }
+});
+
+export class TooltipTrigger extends React.Component {
+  props: any;
```
calbach: We should consider whether we want typedefs for props/state. If we decide to continue with Typescript, we should likely be typing these as they are pretty fundamental to coding a react component.

ui/src/app/components/popups.tsx
```diff
@@ -0,0 +1,285 @@
...
+  id: string;
+
+  constructor(props) {
+    super(props);
+    this.state = { open: false };
+    this.id = `tooltip-trigger-${uniqueId('')}`;
+  }
+
+  render() {
+    const { children, content, ...props } = this.props;
```
calbach: Ah.. very cool - I did not realize you could do a LHS splay in es6.

ui/src/app/utils/index.ts
```diff
@@ -1,5 +1,8 @@
 import {Router} from '@angular/router';
 import {fromJS} from 'immutable';
+import {
+  find
```
calbach: Do we really need lodash find? I would generally favor using standard Javascript APIs (i.e. `[].find()` in this case) unless there's a major deficiency there.

ui/src/app/components/popups.tsx
```diff
@@ -0,0 +1,285 @@
+import {
+  clamp,
+  flow,
+  isEqual,
+  pick,
+  uniqueId
+} from 'lodash/fp';
```
calbach: Note: we likely can configure our build to just have a `_` global. I'm not sure whether I support doing that, but something to consider.

ui/src/app/utils/index.ts
panentheos: The builtin function would be totally suitable in this case. The `lodash/fp` versions do come with some niceties (e.g. predicate shorthands, auto-currying) that can be leveraged to make very compact, expressive code in some cases.

ui/src/app/components/popups.tsx
blrubenstein: I actually argued against that, and moved to this, because I didn't like the convention for the sake of understandability to people not familiar with Lodash. I might be okay with having a `lodash` global? But `_` I don't think is very clear to people without the background.

ui/src/app/components/popups.tsx
blrubenstein: So... I just removed the cast and it worked fine and linted fine. Not sure why it was complaining before, but looks like it might not even be needed?

blrubenstein: Nevermind, strict test compilation fails without it...

ui/src/app/components/popups.tsx
blrubenstein: Correct me if I am wrong @panentheos, but it looks like (from the defaultProps) thing, that these are pretty flexible/mutable? So it might be hard to pin down a typedef

panentheos: React itself doesn't care what you put in props, but any well-designed component should be expecting a particular structure, so you should be able to specify types if you want to. Same goes for state.

blrubenstein:
I have added the active communication topics to the next tech talk, and am going to merge this in.

PR #1610, opened 2018-12-18
[RW-1867][risk=no] Fix matching/highlighting for autocomplete

dolbeew:
Prevent api call for autocomplete results from returning duplicates and fix highlighting of nodes with special characters in the name


dolbeew:
This seems to be caused by the 'distinctUntilChanged' operator used when debouncing user input. Also only happens when you paste in the same text that was in the input before an option was selected.

PR #1612, opened 2018-12-19
[No ticket][Risk=low] update aou-utils

UrsaStutsman:
With the new Swagger (2.4). We need to make sure to add the parameter consumes - application/json to post/put/patch endpoints that have params in their bodies. If the endpoint has no params, or only params in their path, do NOT add this parameter!


calbach:
Did you mean to unrevert the swagger file changes I reverted yesterday? Isn't that needed along with the swagger CLI update?

https://github.com/all-of-us/workbench/pull/1609

PR #1613, opened 2018-12-20
[RW-1853][risk=no] Add common validator functions for forms

dolbeew:
Create custom form validator functions and error message component to use across all cohort builder forms. Will do a follow-up PR for attributes forms since we need to switch them to reactive forms first.


dolbeew:
1. Looks like annotation names are already limited to 255
2. Done
3. Set input max at 999999999 for 9 character limit (number inputs don't have a 'maxlength' attribute)

PR #1614, opened 2018-12-20
[risk=no] Display only valid age deciles in graph

SrushtiGangireddy:
The prod data has weird age deciles like 11 (110-120), 98 and such. 
Filtering away those results to make the bar graph look clean.

### DIFF
--- a/public-api/src/main/java/org/pmiops/workbench/cdr/model/QuestionConcept.java
+++ b/public-api/src/main/java/org/pmiops/workbench/cdr/model/QuestionConcept.java
@@ -169,7 +169,7 @@ public class QuestionConcept {
                     if (validAgeDeciles.contains(r.getStratum5())) {
                         questionAnalysis.addResult(r);
                     }
-                } else{
+                } else {
                     questionAnalysis.addResult(r);
                 }
                 String rStratum5Name = r.getAnalysisStratumName();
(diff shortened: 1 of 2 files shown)


public-api/src/main/java/org/pmiops/workbench/cdr/model/QuestionConcept.java
```diff
@@ -165,7 +165,13 @@ public static void mapAnalysesToQuestions(List<QuestionConcept> questions, List<
                     q.setAnalysis(new AchillesAnalysis(analysis));
                 }
                 AchillesAnalysis questionAnalysis = q.getAnalysis(analysis.getAnalysisId());
-                questionAnalysis.addResult(r);
+                if (analysis.getAnalysisId() == SURVEY_AGE_ANALYSIS_ID) {
+                    if (validAgeDeciles.contains(r.getStratum5())) {
+                        questionAnalysis.addResult(r);
+                    }
+                } else{
```
calbach: nit: space after `else`

public-api/src/main/java/org/pmiops/workbench/publicapi/DataBrowserController.java
```diff
@@ -468,7 +468,6 @@ public ConceptAnalysis apply(ConceptAnalysis ca) {
             List<AchillesAnalysis> analyses = achillesAnalysisDao.findSurveyAnalysisResults(surveyConceptId, qlist);
             QuestionConcept.mapAnalysesToQuestions(questions, analyses);
         }
-
```
calbach: nit: would probably revert all changes to this file since it's a whitespace-only change - it keep the per-file commit history/blame cleaner

PR #1616, opened 2018-12-22
[risk=no] adding application type and modifying search type

SrushtiGangireddy:
1. Added application type to search concepts api call.
2. Modified search word to search with abc*.

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
@@ -82,7 +82,7 @@ public class ConceptService {
                     if (key.length() < 3) {
                         temp.add(key);
                     } else {
-                        temp.add(new String("+" + key));
+                        temp.add(new String("+" + key + "*"));
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


public-ui/src/app/views/quick-search/quick-search.component.ts
```diff
@@ -62,6 +62,8 @@ export class QuickSearchComponent implements OnInit, OnDestroy {
     }
 
     ngOnInit() {
+      // Scroll to the top of the page on component refresh/ route change
+      this.container.nativeElement.parentElement.parentElement.scrollTop = 0;
```
calbach: This scares me, what is this specifically trying to address? Moving back to the top after a navigate?

I think perhaps a better first step might be to remove the scrolling container around the app. I've noticed that it causes pretty painful sizing issues on certain screen sizes; ideally just the window would scroll, rather than a div inside the window. window.scrollTop is at least a small improvement over this.

Possibly handle this piece in a different PR?

SrushtiGangireddy: Hey ch, yeah. I think this pr is all messed up. Would open seperate prs for different things. Thanks. Somehow window.scrollTop here did not seem to work. I will try removing scrolling container.

PR #1617, opened 2019-01-02
[RW-1824][risk=no] Show message for non-js enabled browsers

rushtong:
## Addresses 
https://precisionmedicineinitiative.atlassian.net/browse/RW-1824

Have showed these screens to Lou and I have his 👍 

## Changes:
* Show message when the user has javascript disabled
* Show message when the user has cookies blocked

## Screenshots

Screenshot in FF with javascript disabled: 

screen shot 2019-01-03 at 7 45 46 am

Screenshot in FF with cookies disabled: 

screen shot 2019-01-03 at 9 02 22 am


rushtong:
@blrubenstein - Yeah - I have two things I want to be more clear on. I want to be 100% sure on the links out to the two different sites (one for js enabling, one for cookie enabling) and I do need to run both screens by Lou/Karthik

rushtong:
@blrubenstein - Tagging for a re-review! Got Lou's approval in person.

ui/src/app/app.module.ts
```diff
@@ -88,7 +88,12 @@ import {ToolTipComponent} from './views/tooltip/component';
 (<any>window).StackTrace = StackTrace;
 
 function getBasePath() {
-  return localStorage.getItem(overriddenUrlKey) || environment.allOfUsApiUrl;
+  try {
+    return localStorage.getItem(overriddenUrlKey) || environment.allOfUsApiUrl;
+  } catch (err) {
+      console.log('Error accessing local storage: ' + err);
+      return environment.allOfUsApiUrl;
+  }
```
rushtong: I needed to catch this because the service constructor is called before the App component loads.

ui/src/app/views/app/component.ts
```diff
@@ -37,37 +38,43 @@ export class AppComponent implements OnInit {
...
-    console.log('To override the API URLs, try:\n' +
-      'setAllOfUsApiUrl(\'https://host.example.com:1234\')\n' +
-      'setPublicApiUrl(\'https://host.example.com:5678\')');
+        window.location.reload();
+      };
+      console.log('To override the API URLs, try:\n' +
+        'setAllOfUsApiUrl(\'https://host.example.com:1234\')\n' +
+        'setPublicApiUrl(\'https://host.example.com:5678\')');
+    } catch (err) {
+      this.localStorageDisabled = true;
```
calbach: browser local storage is not the same thing as cookies. Disabling local storage may also break the site, but if so the error messaging should be different.

rushtong: In my testing, disabling cookies has the effect of disabling local storage. In effect, trying to use local storage is what breaks in the app when cookies are disabled in the browser.

calbach: Got it - they may be grouped under the same browser setting in Chrome. I'm not sure this will be generally true in all browsers, perhaps just modifying the messaging to instead say "local storage and cookies" would be more accurate.

Alternatively, these lines of code are for debug purposes only and are not necessary for the application to function. Is this the only thing that breaks in the app (I'm guessing not)? We could also just log an error here instead or remove this functionality entirely if no one's using it (I think having this debug backdoor in the production build is fairly questionable anyways).

rushtong: I did not test all browsers - just firefox and chrome. Both have the same effect on local storage when cookies are disabled, i.e. local storage is also disabled. 

There are only places where *we* refer to local storage, but it is used in some libraries. I think the main one we care about is angular-2-local-storage (not sure how that's pulled in, or if it's just part of angular). I'm not opposed to removing that functionality altogether ... but it does seem like a separate ticket to me. 

Let me do a little bit of rework and see if I can find a more angular-idiomatic way of getting the right info. That way we can keep the cookie-specific user-messaging simple.

calbach: FWIW: looks like that package is only used in data browser, so could likely be removed from the workbench package.json. And don't bother spending too much time on this on my account, though I would request that if you want to go with this approach that you expand on the above comment to make it clear as to why there would be any linkage between cookies and local storage.

rushtong:
Thanks @calbach - appreciate the re-review. Merged from master and re-tested.

PR #1620, opened 2019-01-02
[RW-1851][Risk=no] 1851 with ts objects

UrsaStutsman:
Ok, version of RW-1851 but with styles as JS objects (the other branch is now for demoing scss). Addressed comments from other PR though @panentheos.


ui/src/app/react-components/modals.tsx
```diff
@@ -0,0 +1,83 @@
...
+    marginTop: '.4rem'
+  }
+};
+
+export const Modal = ({style = {}, ...props}) => {
+  return <div><div style={{...styles.modalBackdrop}}></div>
+    <div {...props} style={{...styles.modalMain, ...style}}>{props.children}</div></div>;
+};
+export const ModalTitle = ({style = {}, ...props}) =>
+  <div {...props} style={{...styles.modalTitle, ...style}}>{props.children}</div>;
```
panentheos: nit: you could skip explicitly passing `children` here since they're included in the `...props`

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,231 @@
...
+            <tr>
+              <td><label>Contact Email:</label></td>
+              <td style={{width: '70%'}}><FieldInput
+                id='change-contact-email'
+                style={this.showEmailValidationError() ? styles.unsuccessfulInput : {}}
+                onChange={(e) => this.setState({contactEmail: e.target.value})}
+                onBlur={() => this.leaveFocusEmail()}
+                onFocus={() => this.enterFocusEmail()}
+              >
+              </FieldInput>
```
panentheos: FYI, if an element is empty, you can use a self-closing tag syntax (works for any element, even custom ones)

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  constructor(
+  ) {}
+
+  renderModal(): void {
+    let component;
+    const props = {username: this.userName, creationNonce: this.creationNonce,
+                   closeFunction: this.close};
+    if (this.resend) {
+      component = AccountCreationResendModalReact;
```
panentheos: TOL: Normally we'd have two separate wrapper components here, but that would be overly verbose, so this is fine for now. Once the wrapper is abstracted out we can tighten this up a bit.

ui/src/app/views/account-creation-modals/component.tsx
```diff
@@ -0,0 +1,231 @@
...
+} from 'app/react-components/modals';
+
+function isBlank(s: string) {
+  return (!s || /^\s*$/.test(s));
+}
+
+class AccountCreationResendModalReact extends React.Component<any, any> {
+
+  state: {};
+  props: {
```
blrubenstein: Nit: Per discussion yesterday, this should be defined above and typed in the React.Component typings.

ui/src/app/views/account-creation-modals/component.tsx
UrsaStutsman: That's exactly what I was thinking

blrubenstein:
@panentheos should be good for another look.

PR #1624, opened 2019-01-02
[risk=no][RW-1841] Testing framework for React components (quick tour)

jacmrob:


ui/src/app/views/quick-tour-modal/component.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  it('should not show the previous button when we are on the first slide', () => {
+    const wrapper = component();
+    expect(wrapper.exists('#previous')).toBeFalsy();
+  });
+
+  it('should go to the next slide when we click next', () => {
+    const wrapper = component();
+    const firstSlide = wrapper.state().selected;
+    wrapper.find('#next').simulate('click');
+    expect(wrapper.state().selected).toBe(firstSlide + 1);
```
UrsaStutsman: So been thinking a bunch about this since yesterday, and just bounced my ideas off of Neha. And here's my conclusion. We should be testing the html here (export the array/list thing of text). Reason being, if some day in the future we change the implementation, we still want the html to behave the same way (but may not care about state.selected). Also, with the idea you mentioned yesterday of exporting the array, if we change text there, it should still pass here.

UrsaStutsman: And more I reflect on it yeah, we should be trying to test the html as CH said as much as possible.

jacmrob: okay good thoughts! I'll make that change using the panels array

PR #1625, opened 2019-01-03
[risk=no] add table components

panentheos:
This adds some basic components for tables, pagination, and sortable headers. The style is currently modeled on the Clarity table, but that can be changed later. The current implementation does a basic flex layout, which works best for small tables with limited rows and columns, but it could be replaced with a "virtual scroll" implementation later, if desired.

### DIFF
--- a/ui/src/app/components/tables.tsx
+++ b/ui/src/app/components/tables.tsx
@@ -4,7 +4,7 @@ import Pagination from 'react-paginating';
 import { nextSort } from '../utils/index';
 import { ClrIcon } from './icons';
 
-const rowHeight = 36;
+const rowHeight = '1.5rem';
 
 const styles = {
   flexCell: ({ basis = 0, grow = 1, shrink = 1, min: minWidth = 0, max } = {} as any) => ({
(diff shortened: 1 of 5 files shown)


blrubenstein:
I would ignore the codacy failure, it looks like it just isn't analyzing the PR, which happens sometimes. Reviewing now.

ui/src/app/components/tables.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }}
+    {...{ onClick, children }}
+  />;
+};
+
+export const SimplePagination = ({ total, limit, currentPage, onPageChange }) => {
+  return <Pagination {...{ total, limit, currentPage }} pageCount={5}>
+    {({ pages, hasPreviousPage, previousPage, hasNextPage, nextPage, totalPages }) => {
+      return <div style={styles.paginationRow}>
+        <div style={{ marginLeft: 'auto', marginRight: 24 }}>
```
blrubenstein: Nit: units

ui/src/app/components/tables.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }, toPairs(columns))}
+      </div>;
+    }, range(0, rowCount))}
+  </div>;
+};
+
+const PageButton = ({ active = false, onClick, children }) => {
+  return <div
+    style={{
+      cursor: 'pointer', padding: 4,
```
blrubenstein: Nit: units

PR #1626, opened 2019-01-03
[RW-1840][Risk=low] Adding Moodle api and offline sync

UrsaStutsman:
Not ready for review. This is just the db and api changes so far. Still need clarification from Moodle about field data to make sure logic is correctly implemented.


api/db/changelog/db.changelog-master.xml
```diff
@@ -61,4 +61,5 @@
   <include file="changelog/db.changelog-51-user-profile-updates.xml"/>
   <include file="changelog/db.changelog-52-user-nonce.xml"/>
   <include file="changelog/db.changelog-53-research-purpose-type.xml"/>
+  <include file="changelog/db.changelog-54-moodle-id-to-user.xml"/>
```
rushtong: Note that this doesn't match the actual filename `db.changelog-54-add-moodle-id-to-user.xml` which is why the test is failing.

PR #1633, opened 2019-01-07
[risk=no] add buttons

panentheos:
This adds a basic button component with the existing style variants. Usage is like this:
`<Button onClick={...}>Save</Button>`
`<Button type='secondary' onClick={...}>Cancel</Button>`

There's a lightweight wrapper for tracking hover state. We can decide later whether it's worth bringing in a heavier library like `react-interactive` later.

The lower level `Clickable` component can serve as a useful abstraction of basic button-like behavior (specifically disabled state), without any styling baked in. It can be used as a base for one-off icon buttons, menu items, etc.

Note: This is currently using a div element without any accessibility attributes, due to some unfortunate browser behavior around disabled buttons. So it does not score well from a screen reader or keyboard perspective. Looking at the existing code, it didn't seem like this was a priority, so I left it alone for now, but this could be improved later if desired.

### DIFF
--- a/ui/src/app/components/buttons.tsx
+++ b/ui/src/app/components/buttons.tsx
@@ -1,0 +1,76 @@
+import * as React from 'react';
+
+const styles = {
+  base: ({ disabled }) => ({
+    display: 'inline-flex', justifyContent: 'space-around', alignItems: 'center',
+    height: '1.5rem', minWidth: '3rem', maxWidth: '15rem',
+    fontWeight: 500, fontSize: 12, letterSpacing: '0.02rem', textTransform: 'uppercase',
+    overflow: 'hidden', whiteSpace: 'nowrap', textOverflow: 'ellipsis',
+    cursor: disabled ? undefined : 'pointer', userSelect: 'none',
+    margin: 0, padding: '0rem 0.77rem',
+  }),
+  primary: ({ hover, disabled }) => ({
+    border: 'none',
+    backgroundColor: disabled ? '#c3c3c3' : (hover ? '#4356A7' : '#262262'),
+    borderRadius: '0.3rem',
+    color: '#fff',
+  }),
+  secondary: ({ hover, disabled }) => ({
+    border: '2px solid',
+    borderColor: disabled ? '#c3c3c3' : '#262262',
+    backgroundColor: disabled ? '#f1f2f2' : (hover ? '#262262' : '#ffffff'),
+    borderRadius: '0.2rem',
+    color: disabled ? '#c3c3c3' : (hover ? '#ffffff' : '#262262'),
+  })
+};
+
+class HoverContainer extends React.Component<
+  { children: ({ hover: boolean, trackHover: Function }) => React.ReactNode },
+  { hover: boolean }
+> {
+  constructor(props) {
+    super(props);
+    this.state = { hover: false };
+  }
+
+  trackHover = el => {
+    return React.cloneElement(el, {
+      onMouseEnter: (...args) => {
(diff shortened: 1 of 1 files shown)


ui/src/app/components/buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+class HoverContainer extends React.Component<
+  { children: ({ hover: boolean, trackHover: Function }) => React.ReactNode },
+  { hover: boolean }
+> {
+  constructor(props) {
+    super(props);
+    this.state = { hover: false };
+  }
+
+  trackHover = el => {
```
jacmrob: could you explain a bit more about how this method works?  can a react component not have a render method?  what is this returning/how does it get threaded through the parent component?

ui/src/app/components/buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const { children } = this.props;
+    const { hover } = this.state;
+    return children({ hover, trackHover: this.trackHover });
+  }
+}
+
+export const Clickable = ({ disabled = false, onClick = null, ...props }) => {
+  return <div
+    {...props}
+    onClick={(...args) => onClick && !disabled && onClick(...args)}
```
jacmrob: here are you returning the result of the onClick function or the function itself?

ui/src/app/components/buttons.tsx
```diff
@@ -0,0 +1,76 @@
...
+  />;
+};
+
+export const Button = ({ type = 'primary', style = {}, disabled = false, ...props }) => {
```
jacmrob: so if we have a unique hover style do we pass that through the styles prop too with a conditional?

ui/src/app/components/buttons.tsx
panentheos: This component is using a technique called 'function-as-child', which is a way to abstract out state management while giving the consumer full control of rendering.

This component expects its 'children' prop to be a function with a specific signature, and it delegates its rendering through that function. In this case, we pass `hover`, the hover state, and `trackHover`, which is a function that modifies a DOM element to wire up the hover state tracking. The consumer can render whatever it wants, but should call `trackHover` on the DOM element it's interested in tracking.

Inside `trackHover`, we call `React.cloneElement`, which clones the original element, adding and overriding the specified props. Here, we're adding mouse handlers and wiring those up to the local state. That state ultimately gets passed back down to the consumer via the child function, and used for styling.

The end result is that we can abstract out just the state management piece into a transparent container, and allow the consumer full control of how it's used.

ui/src/app/components/buttons.tsx
panentheos: It's returning the result of the passed `onClick` handler, but that's kind of moot because the return value is meaningless here.

ui/src/app/components/buttons.tsx
gjuggler: Heh, I was also scratching my head while trying to read through this part of the code. Thanks for the well-written clarifications.

To help future readers, it might be worth including a light comment or two calling out that the 'children' prop is being used in a non-standard way (e.g. a singular function rather than an array of child components). I wouldn't have been able to easily grok what's going on without your description above.

ui/src/app/components/buttons.tsx
panentheos: You'd need to add a new `type` value that implements those specific styles in the same way as the other ones. For now, that would be the preferred way of doing things. We can always tweak the underlying implementation if it doesn't work out, but the API should be solid.

PR #1634, opened 2019-01-07
[RW-1824][risk=no] More handling of disabled cookies.

rushtong:
Follow-up to https://github.com/all-of-us/workbench/pull/1617 that addresses https://precisionmedicineinitiative.atlassian.net/browse/RW-1824

* Only try to set urls when cookies are enabled
* Only load tcell js (which relies on local storage) when cookies are enabled.

### DIFF
--- a/ui/src/app/views/app/component.ts
+++ b/ui/src/app/views/app/component.ts
@@ -96,7 +96,10 @@ export class AppComponent implements OnInit {
     });
 
     this.setGTagManager();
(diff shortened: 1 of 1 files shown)


rushtong:
@calbach - Yes, this is really just for that. There is one more error that I was not able to fix, `zone.js` is getting called by something in angular and relies on local storage too. The page still functions OK - just trying to plug the error.

PR #1635, opened 2019-01-07
[RW-1829] React - login page]page template and invitation key

NehaBroad:
Dependent on Account Summary https://github.com/all-of-us/workbench/pull/1648


ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+interface PageTemplateState {
+  currentStep: number;
+  invitationKey: string;
+}
+
+@withWindowSize()
+export class SignPageTemplateReact extends React.Component<any, PageTemplateState> {
+  state: PageTemplateState;
```
panentheos: BTW, since you already declared the `PageTemplateState` in the angle brackets above, there's no need to also declare it here.

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  render() {
+    return <SignedIn>
+      <div style={{width: '100%', display: 'flex', flexDirection: 'column'}}>
+        <Template images={this.pageImages[this.state.currentStep]}
+                  windowsize={this.props.windowSize}>
+            <Header src={this.headerImg}/>
+            <Content>
+              {this.nextDirective(this.state.currentStep)}
```
panentheos: nit: since `nextDirective` is a method in this component, it already has access to state and could read out `currentStep` itself, rather than getting it as a parameter.

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  setInvitationKey(invitationKey) {
+    this.state.invitationKey = invitationKey;
+  }
+
+  render() {
+    return <SignedIn>
+      <div style={{width: '100%', display: 'flex', flexDirection: 'column'}}>
+        <Template images={this.pageImages[this.state.currentStep]}
```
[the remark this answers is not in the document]
NehaBroad: I am not sure if i understand this. Are you talking about <template>?

NehaBroad:
Have updated the PR with all of @panentheos  comments

calbach:
This PR says it's dependent on #1648. #1648 says it's dependent on this PR. 🤔 

I believe this one needs to go in first, so I think the PR description here needs to be fixed.

NehaBroad:
This PR as well as  #1648 needs to go together in order to prevent breaking registration flow

ui/src/app/views/invitation-key/component.tsx
```diff
@@ -0,0 +1,109 @@
...
+      this.inputElement.current.focus();
+      return;
+    }
+
+    const request: InvitationVerificationRequest = {
+      invitationKey: this.state.invitationKey
+    };
+
+    const args: FetchArgs = ProfileApiFetchParamCreator().invitationKeyVerification(request);
+    fetch(fullUrl(args.url), args.options)
```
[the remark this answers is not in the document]
UrsaStutsman: @calbach yup, and I did that because when I did the fetch set up I didn't bother figuring out Configuration. Which now I am trying to figure out.

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  constructor(private signInService: SignInService, private router: Router) {
+    super(SignPageTemplateReact, ['signIn']);
+    this.signIn = this.signIn.bind(this);
+  }
+
+  ngOnInit(): void  {
+    document.body.style.backgroundColor = '#e2e3e5';
+
+    this.signInService.isSignedIn$.subscribe((signedIn) => {
```
panentheos: Since these wrappers are temporary, we should avoid putting logic in them if at all possible. This subscription logic could go into lifecycle methods in the React component. Subscribe in `componentDidMount` and unsubscribe in `componentWillUnmount`.

ui/src/app/views/invitation-key/component.tsx
```diff
@@ -0,0 +1,109 @@
...
+import {FormInput} from 'app/components/inputs';
+
+import {FetchArgs,
+  InvitationVerificationRequest,
+  ProfileApiFetchParamCreator
+} from 'generated/fetch/api';
+
+import * as React from 'react';
+
+function isBlank(s: string) {
```
panentheos: Instead of this function we could do `!someString.trim()`

blrubenstein: The problem with that is that it doesn't handle when the string is null or undefined. We were actually originally doing that and it caused issues.

NehaBroad:
Closing this as to avoid duplication and confusion, the latest code for login/registration is in the PR https://github.com/all-of-us/workbench/pull/1675 along with account creation

PR #1638, opened 2019-01-08
[risk=no] add react-modal

panentheos:
This introduces the `react-modal` library for backing our modal implementation. Main benefits:
* The modal is rendered in a 'portal', meaning the markup appears at the end of the body element. This ensures that it always appears over the main content without worrying about z-index.
* It does focus management, ensuring that a user cannot tab out and interact with the underlying page while the modal is open.

It also has hooks for doing animations, if we want to wire that up at some point.

### DIFF
--- a/ui/src/app/components/modals.tsx
+++ b/ui/src/app/components/modals.tsx
@@ -1,57 +1,48 @@
 import * as React from 'react';
-export const styles = {
-  modalMain: {
-    // tricky little bit required to get certain CSSProperties
-    // to work properly in this setup
-    position: 'fixed' as 'fixed',
-    background: 'white',
-    borderRadius: '8px',
-    width: '30%',
-    height: 'auto',
-    top: '50%',
-    left: '50%',
-    transform: 'translate(-50%,-50%)',
-    zIndex: 1050
+import ReactModal from 'react-modal';
+import {ReactComponent} from '../utils/index';
+
+const styles = {
+  modal: {
+    borderRadius: 8, position: 'relative',
+    padding: '1rem', outline: 'none',
+    backgroundColor: 'white', boxShadow: '0 1px 2px 2px rgba(0,0,0,.2)'
   },
 
-  modalBackdrop: {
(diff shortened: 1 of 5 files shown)


panentheos:
I did verify that those modals continue to work as expected.

Not sure why the tests are timing out though.

ui/src/app/components/modals.tsx
```diff
@@ -1,57 +1,48 @@
...
-    height: 'auto',
-    top: '50%',
-    left: '50%',
-    transform: 'translate(-50%,-50%)',
-    zIndex: 1050
+import ReactModal from 'react-modal';
+import {ReactComponent} from '../utils/index';
+
+const styles = {
+  modal: {
```
calbach: I'm working on the rename notebook modal so I patched this locally. I noticed the positioning is top/middle with this PR vs middle/middle for the existing modal. Adding this to the modal style appears to do the trick:

```
    top: '50%',
    transform: 'translateY(-50%)'
```

panentheos: That was a deliberate change which I forgot to call out, sorry. We found in Terra that having it centered meant that the whole modal would shift if the content changed, which was a jarring UX. Having it top-aligned still allows it to expand downwards if needed, but the window doesn't jump around.

If you prefer it centered though, I can switch it.

calbach: I see the issue. That said - I'm not sure we have this kind of scenario in AoU right now, our modals are generally pretty simple. I'm fine with the change if @lstrano-broad is suggesting we make this style rework across all AoU modals - otherwise I would leave the styling as-is since our mocks are still showing centered modals.

For example, this looks weird to me: image

panentheos: Sounds good, let's center it so things stay consistent during the refactor, and we can always swap the implementation later if we decide that's preferable. Thanks!

lstrano-broad:
Since our AoU models are simple and don’t involve deeper interaction from users - would like to keep the modals centered vs. pinned to top

PR #1640, opened 2019-01-08
[RW-1897][risk=no] Switch resource rename modal to React

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1897

Notes:
- Moves "show/hide" logic to the parent `*ngIf`. This felt better to me than the existing pattern where the modal would handle its own visibility state.
- Import Typescript bindings for react-modal
- Add helpers for interacting with a React child in an Angular test
- Add helper for react-modal test setup (requires global element)

Future work:
- Decision on loading UX in a modal; previously we used clr-loading on a button
- Consider supporting a close button, which the previous modal has (but isn't strictly necessary)


ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+export class RenameModalComponent extends React.Component<RenameModalProps, RenameModalState> {
+  constructor(props: RenameModalProps) {
+    super(props);
+    this.state = {
+      loading: false,
+      newName: ''
+    };
+  }
+
+  private emitRename(): void {
```
panentheos: nit: I realize this was copied from the previous code, but since we're not technically emitting an event any more, what about a simpler name such as `save` or `rename`?

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (this.state.loading) {
```
panentheos: nit: This check should be unnecessary, since the button's `disabled` prop will prevent this from being called.

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+  ModalBody,
+  ModalFooter,
+  ModalTitle,
+} from 'app/components/modals';
+import {ReactComponent} from 'app/utils';
+
+
+interface RenameModalProps {
+  resource: any;
+  resourceType: string;
```
panentheos: This looks like it isn't used.

calbach: Good catch, removed.

ui/src/app/views/rename-modal/component.tsx
calbach: Agreed - done.

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+    this.props.onRename({
+      name: this.props.resource.name,
+      newName: this.state.newName
+    });
+  }
+
+  render() {
+    return <React.Fragment>
+      <Modal>
+        <ModalTitle>Please enter the new name for {this.props.resource.name}</ModalTitle>
```
calbach: @panentheos A lot of React code I've seen pulls props out to local constants in the render() method. Do you recommend this pattern and would you apply it here (i.e. for the resource name)? (I assume it's purely for code readability).

ui/src/app/views/rename-modal/component.tsx
calbach: I wasn't sure about a possible race here, but should probably be fine - done.

ui/src/app/views/resource-card/component.html
```diff
@@ -43,9 +43,11 @@
                           (receiveDelete)="receiveDelete($event)"></app-confirm-delete-modal>
 
 
-<app-rename-modal [resource]="resourceCard.notebook"
+<app-rename-modal *ngIf="renaming && resourceCard.notebook"
+                  [resource]="resourceCard.notebook"
                   [resourceType]="'Notebook'"
-                  (receiveRename)="receiveNotebookRename($event)"></app-rename-modal>
+                  [onCancel]="cancelRename.bind(this)"
```
calbach: Note: @panentheos I diverged slightly from what we had in prior PRs - where we bound methods in the constructor and overwrote them: https://github.com/all-of-us/workbench/blob/228888948f45ab8e28ea3160877175775c277caf/ui/src/app/views/homepage/component.ts#L83 Any thoughts on this?

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+  resourceType: string;
+  onRename: Function;
+  onCancel: Function;
+}
+interface RenameModalState {
+  loading: boolean;
+  newName: string;
+}
+
+@ReactComponent({
```
calbach: @panentheos is there a symbol I can use with this approach to render the original RenameModal directly in React? Or would the idea for migration instead be to remove `@ReactComponent` and compose into other React components in one atomic change.

calbach:
Alright, the test fixes resulted in several changes. Please take another look when you have a chance. Also @jacmrob per your testing doc (I will also comment there).

calbach:
Hm - also stuck on this compilation failure. Full detail on https://github.com/all-of-us/workbench/pull/1642

ui/src/app/views/resource-card/component.html
panentheos: That looks good to me, I didn't know you could explicitly reference `this` in the template expression.

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -0,0 +1,69 @@
...
+  selector: 'app-rename-modal',
+  propNames: ['resource', 'onRename', 'onCancel']
```
blrubenstein: It would be great to have a utility that read the key names from an interface and spat out an array of strings. I think I found some external libraries that do that? Or we could roll our own pretty easily. (To avoid the duplication here)

calbach: Good suggestion, but apparently non-trivial in typescript with either the object type or interface type since they don't exist at runtime. https://github.com/Microsoft/TypeScript/issues/13267

There's apparently a custom typescript plugin which lets you do that, but looks questionable and probably not worth bringing in.

Something like this would also work ([per SO](https://stackoverflow.com/questions/45670705/typescript-iterate-interface-properties)), but probably adds more complexity than it saves in boilerplate - especially since this is transitional anyways:

```
const RenameModalProps = {
  resource: {name: ""};
  onRename: ()=>{};
  onCancel: ()=>{};
};
type RenameModalProps = typeof RenameModalProps;
...

{
  propNames: Object.keys(RenameModalProps)
```

ui/src/app/views/resource-card/component.html
jacmrob: i like this, more concise

calbach:
This PR now depends on https://github.com/all-of-us/workbench/pull/1642

PR #1643, opened 2019-01-09
[RW-1900][risk=no] Convert edit icon to react

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1900


ui/src/app/icons/edit/component.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  }));
+
+  it('should render', fakeAsync(() => {
+    updateAndTick(fixture);
+    expect(fixture).toBeTruthy();
+  }));
+
+  // TODO: Figure out how to get the style out of the hovered element.
+  it('should change style on hover', fakeAsync(() => {
+    simulateEvent(fixture, fixture.debugElement, 'onmouseover');
```
calbach: FYI: for event simulation check out the stuff I did in https://github.com/all-of-us/workbench/pull/1640 , could likely do something similar here.

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,73 @@
+import {Component, OnInit} from '@angular/core';
+
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+
+interface EditComponentState { style: object; }
+
+class EditComponentReact extends React.Component<{}, EditComponentState> {
```
calbach: opt: I suspect this may be simple enough to use the functional form of a react component to minimize boilerplate: https://reactjs.org/docs/components-and-props.html

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </g>
+        </g>
+      </svg>
+    );
+  }
+}
+
+@Component({
+  selector: 'app-edit-icon',
+  templateUrl: './component.html',
```
calbach: No need for a separate file, can just inline a template here. Also could use this base class once merged: https://github.com/all-of-us/workbench/pull/1642

calbach: Specfifically `template: '<div #root></div>'` (or whatever you want to call it)

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,73 @@
...
+})
+
+export class EditComponent implements OnInit {
+
+  constructor() {}
+
+  ngOnInit(): void {
+    ReactDOM.render(
+        React.createElement(EditComponentReact, {}),
+        document.getElementById('app-edit-icon'));
```
calbach: Should use ViewChild rather than document get. there may be multiple of these pencils rendered into the document

rushtong: Good point, will do.

ui/src/app/icons/edit/component.tsx
rushtong: Thanks, yes - as I've been trimming this down, it's becoming clearer how much can be pruned away.

ui/src/app/icons/edit/component.tsx
rushtong: I had considered that, and can easily be persuaded to do that. I was trying to be consistent with our typing of props and state in other conversions - will give it some more thought.

ui/src/app/icons/edit/component.spec.ts
rushtong: Thank you!

ui/src/app/icons/edit/component.tsx
rushtong: I ended up not doing this - while it compiled and ran fine, the IDE complains about typing when trying to access state variables.

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,83 @@
+import {AfterViewInit, Component, ElementRef, ViewChild} from '@angular/core';
+
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+
+interface EditComponentState { style: object; }
+
+const style = {
+  height: 14,
```
calbach: units?

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,83 @@
...
+const fillStyle = {
+  fill: '#4D4CA5'
+};
+
+class EditComponentReact extends React.Component<{}, EditComponentState> {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+        style: style
```
calbach: nit: generally we do two-space block indents (we should actually encode this somewhere in our linter..)

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </svg>
+    );
+  }
+}
+
+@Component({
+  selector: 'app-edit-icon',
+  template: '<div #editIcon></div>',
+})
+export class EditComponent implements AfterViewInit {
```
calbach: It won't help much in this particular case, but to make the boilerplate more consistent - consider switching this to https://github.com/all-of-us/workbench/blob/d7d5f247616103a83f1da10fb35cbec3c8690208/ui/src/app/views/rename-modal/component.tsx#L73

ui/src/app/icons/edit/component.tsx
rushtong: I was following the PR advice here: https://github.com/all-of-us/workbench/pull/1644#discussion_r246774697

> FYI: px is assumed as the default, so this can be written as `fontSize: 21`

I'm not opposed to either approach - would prefer to be consistent in whatever we choose.

ui/src/app/icons/edit/component.tsx
rushtong: Oof ... my IDE is weirdly messed up. The default indent is set 2 spaces but it still messes up sometimes. Will fix.

rushtong: Done.

ui/src/app/icons/edit/component.tsx
rushtong: Done

rushtong:
@calbach , @jacmrob - feedback addressed/commented - PTAL, thanks.

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,77 @@
...
+const fillStyle = {
+  fill: '#4D4CA5'
+};
+
+class EditComponentReact extends React.Component<{}, EditComponentState> {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      style: style
```
panentheos: While we do need state to track the hover, we could minimize it a bit by just storing the hover state as a boolean, and computing the final style during render. (However, see below)

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  render() {
+    return (
+      <svg style={this.state.style}
+           viewBox='0 0 14 14'
+           version='1.1'
+           xmlns='http://www.w3.org/2000/svg'
+           xmlnsXlink='http://www.w3.org/1999/xlink'
+           onMouseOver={() => this.mouseOver()}
```
panentheos: FYI, I'm about to add `react-interactive` which will allow us to add hover styles much more easily, without tracking the state by hand. So don't fret too much about the implementation of this.

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,77 @@
...
+           onMouseLeave={() => this.mouseLeave()}>
+        <title>Edit</title>
+        <desc>Created with Sketch.</desc>
+        <defs></defs>
```
panentheos: It might be nice to run this through SVGO to strip out some of the cruft.

ui/src/app/icons/edit/component.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface EditComponentState { style: object; }
+
+const style = {
+  height: 14,
+  width: 14,
+};
+
+const fillStyle = {
+  fill: '#4D4CA5'
```
panentheos: FYI, a nice trick for coloring SVGs is to set `fill='currentColor`' on the svg, which makes it inherit the current `color` value from wherever it's rendered. That way you can style it from outside by e.g. setting the color on the container.

ui/src/app/icons/edit/component.tsx
rushtong: This is a straight copy (modulo the react/lint changes) from the current `svg` - not sure how I feel about making that kind of change in a while-you're-at-it phase. We do have several of these, so maybe it makes sense to make that specific task.

panentheos:
Practical question: If the current implementation of the svg fill color isn't actually doing anything, can we just strip out that code for now until we figure out how we want it to behave? Perhaps we should keep the code simple until we decide we need something else.

ui/src/app/icons/edit/component.tsx
rushtong: Got it - yes, that makes sense.

ui/src/app/icons/edit/component.tsx
rushtong: This specific style is only supposed to take effect on hover as per the original code, even doesn't seem to do anything visually. I still need to run this by Lou to make sure something wasn't lost at some point.

rushtong:
@panentheos - I might end up doing exactly that. My first goal was to get the react-conversion to do exactly what the current implementation does. 

> Practical question: If the current implementation of the svg fill color isn't actually doing anything, can we just strip out that code for now until we figure out how we want it to behave? Perhaps we should keep the code simple until we decide we need something else.

rushtong:
@calbach, @jacmrob - More updates based on Lou's feedback - PTAL, thank you.

ui/src/app/icons/edit/component.spec.ts
```diff
@@ -0,0 +1,64 @@
+import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import * as ReactTestUtils from 'react-dom/test-utils';
+import {
+  updateAndTick
+} from 'testing/test-helpers';
+import {EditComponent} from './component';
+
+// Note that the description is slightly different from the component for easier test filtering.
+describe('EditIconComponent', () => {
```
jacmrob: should these be jest tests and the file be `.spec.tsx`?

rushtong: I don't know - is that part of the conversion process?

jacmrob: yes- see (this spec)[https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/quick-tour-modal/component.spec.tsx] for example

rushtong: Done - thank you!

rushtong:
@jacmrob - feedback addressed, PTAL

PR #1644, opened 2019-01-09
[RW-1855][risk=no] CSS for quick tour

jacmrob:
Note: this is currently not including changing the color of the links on "Workspaces" change of the quick tour. I believe this will have to involve using some alternative library (maybe Interactive?).  I haven't been able to find any examples of how to do this either in code or online (i tried something with radium but couldn't get it to work).  If anyone has thoughts, I would love to hear them, but I don't want them to block this story for the time being.  

This also fixes the [RW-1930](https://precisionmedicineinitiative.atlassian.net/browse/RW-1930) link spacing issue.


ui/src/app/components/buttons.tsx
```diff
@@ -21,6 +21,18 @@ const styles = {
     backgroundColor: disabled ? '#f1f2f2' : (hover ? '#262262' : '#ffffff'),
     borderRadius: '0.2rem',
     color: disabled ? '#c3c3c3' : (hover ? '#ffffff' : '#262262'),
+  }),
+  darklingPrimary: ({ hover, disabled }) => ({
```
jacmrob: currently i don't believe we use these buttons anywhere else in the site, but making this general in case we decide to have more modals that use a dark background

ui/src/app/views/quick-tour-modal/component.css
```diff
@@ -1,223 +1,3 @@
...
-
-:host::ng-deep .resize-icon {
-  cursor: pointer;
-}
-
-:host::ng-deep .invisible {
-  display: none;
-}
-
 :host::ng-deep .link, .link:visited {
```
jacmrob: keeping this for now so that the link coloring works properly

PR #1646, opened 2019-01-10
Db 183

SrushtiGangireddy:
1. Get the expanded chart div into view when the concept at the end of the page is expanded.
2. Also close the previously expanded graphs.

### DIFF
--- a/public-ui/src/app/views/ehr-view/ehr-view.component.ts
+++ b/public-ui/src/app/views/ehr-view/ehr-view.component.ts
@@ -190,10 +190,7 @@ export class EhrViewComponent implements OnInit, OnDestroy {
   public selectGraph(g) {
     document.getElementById('chartElement').scrollIntoView(
       { behavior: 'smooth', block: 'nearest', inline: 'start' });
-    this.showGenderGraph = false;
-    this.showAgeGraph = false;
-    this.showSourcesGraph = false;
-    this.showMeasurementGenderBins = false;
+    this.resetSelectedGraphs();
     if (g === 'Gender') {
       this.showGenderGraph = true;
     } else if (g === 'Age') {
@@ -220,6 +217,12 @@ export class EhrViewComponent implements OnInit, OnDestroy {
       return this.sourcesChartHelpText;
     }
   }
+  public resetSelectedGraphs() {
+    this.showGenderGraph = false;
+    this.showAgeGraph = false;
+    this.showSourcesGraph = false;
+    this.showMeasurementGenderBins = false;
+  }
   public expandRow(concepts, r) {
     const index = concepts.findIndex(x => x.conceptId === r.conceptId);
     const prevConcepts = concepts.slice(0, index);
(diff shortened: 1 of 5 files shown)


public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -93,7 +93,6 @@ export class EhrViewComponent implements OnInit, OnDestroy {
       this.domainId = params.id;
     });
   }
-
```
calbach: nit: I'd revert this, generally there should be newline between method definitions

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -189,6 +188,8 @@ export class EhrViewComponent implements OnInit, OnDestroy {
     }
   }
   public selectGraph(g) {
+    document.getElementById('chartElement').scrollIntoView(
```
calbach: This breaks component encapsulation. This would be better:

html:
```
<div #chartElement>
```

ts:
```
@ViewChild('chartElement') chartEl: ElementRef;
```

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -219,4 +220,20 @@ export class EhrViewComponent implements OnInit, OnDestroy {
       return this.sourcesChartHelpText;
     }
   }
+  public expandRow(concepts, r) {
```
calbach: Please use types

public-ui/src/app/views/ehr-view/ehr-view.component.ts
```diff
@@ -219,4 +220,20 @@ export class EhrViewComponent implements OnInit, OnDestroy {
...
+    const index = concepts.findIndex(x => x.conceptId === r.conceptId);
+    const prevConcepts = concepts.slice(0, index);
```
calbach: nit: the prev/next thing seems a little verbose, maybe something like this?

```
if (r.expanded) {
  r.expanded = false;
  return;
}
concepts.forEach(c => c.expanded = false);
r.expanded = true;
```

SrushtiGangireddy: Done.

public-ui/src/app/views/ehr-view/ehr-view.component.ts
SrushtiGangireddy: Concept does not have expanded prop in model so using any.

public-ui/src/app/views/ehr-view/ehr-view.component.ts
SrushtiGangireddy: Done.

public-ui/src/app/views/ehr-view/ehr-view.component.ts
SrushtiGangireddy: my tslint does not allow to push code with the trailing whitespace somehow.

PR #1648, opened 2019-01-10
[RW-1883] account creation conversion/refactor

UrsaStutsman:
Dependent on RW-1829 (https://github.com/all-of-us/workbench/pull/1635)


ui/src/app/views/account-creation/component.tsx
```diff
@@ -0,0 +1,316 @@
...
+                Username is already taken.
+                </Error>}
+              {this.usernameInvalidError() &&
+                <Error>
+                Username is not a valid username.
+                </Error>}
+            </div>
+          </FormSection>
+          <FormSection>
+            <button type="submit"
```
NehaBroad: Can we use <Button>

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -47,23 +62,27 @@ export class SignPageTemplateReact extends React.Component<any, PageTemplateStat
...
       {
-        backgroundImgSrc: '/assets/images/congrats-female.png',
-        smallerBackgroundImgSrc: '/assets/images/congrats-female-standing.png'
+        backgroundImgSrc: 'assets/images/congrats-female.png',
+        smallerBackgroundImgSrc: 'assets/images/congrats-female-standing.png'
       }];
     this.updateNext = this.updateNext.bind(this);
     this.setInvitationKey = this.setInvitationKey.bind(this);
+    this.setProfile = this.setProfile.bind(this);
+    this.invitationKey = '';
```
NehaBroad: Question: Why do we need invitationKey

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -19,19 +20,33 @@ interface ImagesSource {
...
   pageImages: Array<ImagesSource>;
   headerImg = '/assets/images/logo-registration-non-signed-in.svg';
+  invitationKey: string;
 
   constructor(props: object) {
     super(props);
     this.state = {
-      currentStep: 0,
-      invitationKey: ''
+      currentStep: 1,
```
NehaBroad: Just a note for other reviewers.(@UrsaStutsman  correct me if i am wrong) currentStep will be 0 (or rather some string as per review by @panentheos ) this is 1 right now because of testing by ursa.

ui/src/app/components/inputs.tsx
```diff
@@ -24,14 +37,37 @@ export const styles = {
...
   <input {...props} style={{...styles.input, ...style}} />;
 export const FormInput = ({style = {}, ...props}) =>
     <input {...props} style={{...styles.formInput, ...style}} ref={props.inputref}/>;
+export const LongInput = ({style = {}, ...props}) =>
+  <input {...props} style={{...styles.formInput, ...styles.longInput, ...style}} />;
 export const Error = ({style = {}, ...props}) =>
   <div {...props} style={{...styles.error, ...style}} />;
-
+export const FormSection = ({style = {}, ...props}) =>
+  <div {...props} style={{...styles.formSection, ...style}} />;
```
NehaBroad: So from the class name i assumed the styles will be for inputs and not form. I think we should change the name of the file or movie Formsection error message to a different file

UrsaStutsman: Yeah I almost did, but currently this is the only "form" element, so felt silly to do another whole file, plus this would only be used in conjunction with inputs... If anyone really has strong feeling about this I'll move it.

ui/src/app/views/account-creation/component.tsx
```diff
@@ -0,0 +1,326 @@
...
+          <FormSection>
+            <LongInput type='text' id='givenName' name='givenName' autoFocus
+                    placeholder='First Name'
+                    value={this.state.profile.givenName}
+                    style={this.errorMap['givenName'] ?
+                      inputStyles.unsuccessfulInput : inputStyles.successfulInput}
+                    onChange={this.updateField}/>
+              {this.errorMap['givenName'] &&
+                <ErrorMessage>
+                First Name must be 80 characters or less.
```
NehaBroad: nit: tabbing for error message First Name

ui/src/app/views/account-creation/component.tsx
```diff
@@ -0,0 +1,312 @@
...
+          console.log(error);
+          this.setState({usernameCheckInProgress: false});
+        });
+    }, 300);
+  }
+
+  updateProfile(attribute: string, value: string) {
+    const newProfile = this.state.profile;
+    newProfile[attribute] = value;
+    this.setState({profile: newProfile});
```
panentheos: It's generally a good idea to avoid mutating state if possible. You could use the function form of `setState` along with a `_.set` to implement this function as:
```
this.setState(({profile}) => ({profile: _.set(attribute, value, profile)}));
```

ui/src/app/views/account-creation/component.tsx
```diff
@@ -0,0 +1,312 @@
...
+  usernameCheckInProgress: boolean;
+  usernameConflictError: boolean;
+  creatingAccount: boolean;
+  showAllFieldsRequiredError: boolean;
+}
+
+export class AccountCreationReact extends
+    React.Component<AccountCreationProps, AccountCreationState> {
+  usernameCheckTimeout: NodeJS.Timer;
+  accountCreated = false;
```
panentheos: Placing this boolean directly on `this` makes it 'stateful', but React can't see it and so it won't cause re-renders on change. Since the parent will end up unmounting this component on completion anyway, we can probably remove this.

ui/src/app/views/sign-in-flow-template/component.tsx
```diff
@@ -49,7 +54,19 @@ const SignPageTemplateReact = withWindowSize()(
...
+          // profile creation, this field is populated with the full email address.
+          username: '',
+          dataAccessLevel: DataAccessLevel.Unregistered,
+          givenName: '',
+          familyName: '',
+          contactEmail: '',
+          currentPosition: '',
+          organization: '',
+          areaOfResearch: ''
+        }
```
panentheos: Is the profile needed here? It doesn't look like it's used anywhere.

ui/src/app/views/account-creation/component.tsx
```diff
@@ -0,0 +1,312 @@
...
+    this.setState({creatingAccount: true});
+    const args: FetchArgs = ProfileApiFetchParamCreator().createAccount(request);
+    fetch(fullUrl(args.url), args.options)
+      .then(handleErrors)
+      .then((response) => response.json)
+      .then((profile) => {
+          this.setState({profile: profile, creatingAccount: false});
+          this.accountCreated = true;
+          setProfile(profile);
+          onAccountCreation();
```
panentheos: Since these two callbacks are called together, they could be combined. E.g. we could just pass the profile to `onAccountCreation`, and the parent could save it and change the step at the same time. (Assuming the parent really needs the profile, see other comment)

ui/src/app/views/sign-in-flow-template/component.tsx
UrsaStutsman: not yet, but by the time we finish the full flow.

ui/src/app/views/account-creation/component.tsx
UrsaStutsman: So I tried this out, and typescript compilation is complaining on 'attribute' because it's expecting an object but I'm passing a string?

UrsaStutsman: nvm. figured it out (order of args is actually: profile, attribute, value)

PR #1649, opened 2019-01-11
DB-190 [Wildcard* search causing issue on source vocabulary search]

SrushtiGangireddy:
1. Modified concept search to fetch counts by searching with wildcard* and to fetch actual concepts by the exact match.
2. Added order number ordering in survey results fetch.

Detail:
Bug: This line https://github.com/all-of-us/workbench/blob/master/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java#L85 for wildcard * search 
is causing DB-190 issue.

For Eg: Search on 507 code fetches all the concepts that have 507 in concept code along with the source concept that is exact match and the standard concept of the particular source concept.
Eg matched codes: 50792001, 50749006, 50799005, 507, 50711007.

So fixed this line to add wildcard* only in the case of fetching survey counts.

Reason to add this when fetching survey counts is the survey angular view fetches all the results before hand and filters the results that CONTAINS the search text (not the exact match with the words like mysql match).

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
@@ -74,6 +74,7 @@ public class ConceptService {
         List<String> temp = new ArrayList<>();
         for (String key: keywords) {
             String tempKey;
+            // This is to exact match concept codes like 100.0, 507.01. Without this mysql was matching 100*, 507*.
             if (key.contains(".")) {
                 tempKey = "\"" + key + "\"";
             } else {
@@ -90,6 +91,11 @@ public class ConceptService {
                     if (key.length() < 3) {
                         temp.add(key);
                     } else {
+                        // Only in the case of calling this method from getDomainSearchResults to fetch survey counts add wildcard* search.
+                        // The survey view angular code fetches all the results of each survey module and then checks if the search text is present in the concept name / stratum4 of achilles results using regex test.
+                        // Without this the number of results for search smoke would be 6 while also the actual results would be 12 as smoking, smoked (smoke*) are considered.
+                        // Changing this would address the search count discrepancy for survey results. If * wildcard is added to all search types, source vocabulary code match on 507 fetches all the results matching 507*. (which is not desired to show the source / standard code mapping)
+                        // So added different search type for each purpose
                         if (searchType == SearchType.SURVEY_COUNTS) {
                             temp.add(new String("+" + key + "*"));
                         } else {
@@ -158,6 +164,7 @@ public class ConceptService {
                           try {
                             long conceptId = Long.parseLong(query);
                             standardOrCodeOrIdMatch.add(criteriaBuilder.equal(root.get("conceptId"),
+
                                 criteriaBuilder.literal(conceptId)));
                           } catch (NumberFormatException e) {
                             // Not a long, don't try to match it to a concept ID.
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
@@ -64,7 +64,7 @@ public class ConceptService {
         this.conceptDao = conceptDao;
     }
 
-    public static String modifyMultipleMatchKeyword(String query) {
+    public static String modifyMultipleMatchKeyword(String query, String purpose) {
         // This function modifies the keyword to match all the words if multiple words are present(by adding + before each word to indicate match that matching each word is essential)
         if (query == null || query.trim().isEmpty()) {
             return null;
@@ -82,7 +82,11 @@ public class ConceptService {
                     if (key.length() < 3) {
                         temp.add(key);
                     } else {
-                        temp.add(new String("+" + key + "*"));
+                        if (purpose.equalsIgnoreCase("survey_counts")) {
+                            temp.add(new String("+" + key + "*"));
+                        } else {
+                            temp.add(new String("+" + key));
+                        }
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -64,7 +64,7 @@ public ConceptService(EntityManager entityManager, ConceptDao conceptDao) {
         this.conceptDao = conceptDao;
     }
 
-    public static String modifyMultipleMatchKeyword(String query) {
+    public static String modifyMultipleMatchKeyword(String query, String purpose) {
```
calbach: Please use an enum for this purpose

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -82,7 +82,11 @@ public static String modifyMultipleMatchKeyword(String query) {
                     if (key.length() < 3) {
                         temp.add(key);
                     } else {
-                        temp.add(new String("+" + key + "*"));
+                        if (purpose.equalsIgnoreCase("survey_counts")) {
+                            temp.add(new String("+" + key + "*"));
```
calbach: `new STring()` should be redundant here

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -82,7 +82,11 @@ public static String modifyMultipleMatchKeyword(String query) {
                     if (key.length() < 3) {
                         temp.add(key);
                     } else {
-                        temp.add(new String("+" + key + "*"));
+                        if (purpose.equalsIgnoreCase("survey_counts")) {
```
calbach: Please add a comment - I don't understand why we would only glob for survey counts

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: Done.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
SrushtiGangireddy: Added detail in the description.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -64,25 +65,36 @@ public ConceptService(EntityManager entityManager, ConceptDao conceptDao) {
...
         List<String> temp = new ArrayList<>();
         for (String key: keywords) {
-            if (!key.isEmpty()) {
-                if (key.contains("-") && !temp.contains(key)) {
-                    temp.add(key);
-                } else if (key.contains("*") && key.length() > 1) {
-                    temp.add(new String("+" + key));
+            String tempKey;
+            if (key.contains(".")) {
+                tempKey = "\"" + key + "\"";
```
SrushtiGangireddy: Found an issue when matching 100.0. By default matches with only 100 somehow. 
So matching on "100.0" (exact match).

common-api/src/main/java/org/pmiops/workbench/cdr/dao/SurveyModuleDao.java
```diff
@@ -16,17 +16,17 @@
...
-      "0 participant_count\n" +
+      "0 participant_count, m.order_number \n" +
       "from survey_module m\n" +
       "join achilles_results r on m.concept_id = r.stratum_1\n" +
       "join concept c on r.stratum_2 = c.concept_id\n" +
       "where r.analysis_id = 3110\n" +
       // Because we're using a native query we use MySQL match() here directly instead of matchConcept()
       // TODO: add AchillesResults entity, replace this with JQL
-      "and (match(c.concept_name, c.concept_code, c.vocabulary_id, c.synonyms) against (?1 in boolean mode) > 0 or\n" +
+      "and (match(c.concept_name) against (?1 in boolean mode) > 0 or\n" +
```
SrushtiGangireddy: The angular view filters only on the concept_name / stratum 4. So changed this to avoid difference in the counts and the result count.

SrushtiGangireddy: https://github.com/all-of-us/workbench/blob/master/public-ui/src/app/views/survey-view/survey-view.component.ts#L166

common-api/src/main/resources/common_api.yaml
```diff
@@ -43,6 +43,11 @@ definitions:
     description: a domain for concepts corresponding to a table in the OMOP schema
     enum: &DOMAIN [OBSERVATION, PROCEDURE, DRUG, CONDITION, MEASUREMENT, DEVICE, DEATH, VISIT]
 
+  SearchType:
```
calbach: Is this used int he API? If not - I'd suggest just using a normal Java enum instead in ConceptService

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
calbach: Could you add a comment to the code? I'm still not clear on how this relates to the SearchType

calbach: Actually, I read the PR description and this makes some more sense (thanks for adding this) - but a comment in the code here would still be good.

common-api/src/main/resources/common_api.yaml
SrushtiGangireddy: This is used in the api conceptscontroller too.

calbach: Sure - but it's not actually used in the Swagger API (e.g. a client would never need to reference this) - it's just an internal implementation detail so I'd  recommend just leaving it in Java.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptService.java
```diff
@@ -18,11 +18,15 @@
 import org.springframework.data.domain.Sort.Direction;
 import org.springframework.data.jpa.domain.Specification;
 import org.springframework.stereotype.Service;
-import org.pmiops.workbench.model.SearchType;
 
 @Service
 public class ConceptService {
 
+    public static enum SearchType
+    {
```
calbach: nit: normally this `{` would go on the previous line, as with a class. We should have a more obvious style guide so this isn't so subjective though..

PR #1650, opened 2019-01-11
[risk=no] add react-interactive

panentheos:
This adds `react-interactive`, a library that makes it a bit easier to apply inline styling for hover, focus, etc. It cleans up the button implementation a bit (API remains the same). It's also available to be used for other elements besides buttons, as needed.


ui/src/app/components/buttons.tsx
```diff
@@ -1,93 +1,77 @@
...
   }
+};
 
-  render() {
-    const { children } = this.props;
-    const { hover } = this.state;
-    return children({ hover, trackHover: this.trackHover });
-  }
-}
+const computeStyle = ({style, hover, disabledStyle}, {disabled}) => {
```
rushtong: Lou's comment in [slack](https://pmi-engteam.slack.com/archives/C66DX9FD0/p1547221639134600) indicates that we might need a key-pressed state, but that's likely separate from this PR. Just food for thought.

PR #1652, opened 2019-01-11
[RW-1896][risk=no] Confirm delete modal in react

jacmrob:


ui/src/app/utils/index.ts
```diff
@@ -217,3 +217,17 @@ export class ReactWrapperBase implements DoCheck, OnInit, OnDestroy {
...
+  return str
+      .replace(/([a-z\d])([A-Z])/g, '$1' + separator + '$2')
+      .replace(/([A-Z]+)([A-Z][a-z\d]+)/g, '$1' + separator + '$2')
+      .toLowerCase();
+}
+
+export function capitalize(s: string) {
+  return s.toLowerCase().replace( /\b./g,
+      function(a) { return a.toUpperCase(); } );
+}
```
panentheos: `lodash` has several string functions that might suffice here. Specifically see `lowerCase` and `capitalize`

ui/src/app/views/confirm-delete-modal/component.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export interface ConfirmDeleteModalState {
+  loading: boolean;
+}
+
+export class ConfirmDeleteModal
+    extends React.Component<ConfirmDeleteModalProps, ConfirmDeleteModalState> {
+  state: ConfirmDeleteModalState;
+  props: ConfirmDeleteModalProps;
```
panentheos: You don't need to declare `state` and `props` types separately here. They're automatically inferred from the declarations in the angle brackets above.

PR #1655, opened 2019-01-14
[DO NOT MERGE] Srubenst/settings page

blrubenstein:
This PR is pending a fix from Ursa to pass bearer token up with fetch requests. Other than that it should be good to review.


blrubenstein:
Actually, I want to fix the tests first. Please wait! Looking at that now.

blrubenstein:
Okay! Tests are green. @calbach if you would like to do preliminary review, that would be great! Otherwise, we can wait until Ursa has the auth figured out.

PR #1657, opened 2019-01-15
[risk=no] convert resource card menu to React

panentheos:
Implements the popup menu on resource cards using the new popup library.

### DIFF
--- a/ui/src/app/utils/index.ts
+++ b/ui/src/app/utils/index.ts
@@ -188,8 +188,7 @@ export function reactStyles<T extends {[key: string]: React.CSSProperties}>(t: T
 export class ReactWrapperBase implements OnChanges, OnInit, OnDestroy {
   @ViewChild('root') rootElement: ElementRef;
 
-  constructor(private wrapped: (new (...args: any[]) => React.Component)|React.FunctionComponent,
-              private propNames: string[]) {}
+  constructor(private wrapped: React.ComponentType, private propNames: string[]) {}
 
   ngOnInit(): void {
     this.renderComponent();
--- a/ui/src/app/views/resource-card/component.spec.ts
+++ b/ui/src/app/views/resource-card/component.spec.ts
@@ -16,7 +16,7 @@ import {CohortsService} from 'generated/api/cohorts.service';
 import {ConceptSetsService} from 'generated/api/conceptSets.service';
 import {WorkspacesService} from 'generated/api/workspaces.service';
 
-import {ResourceCardComponent} from './component';
+import {ResourceCardComponent, ResourceCardMenuComponent} from './component';
 
 import {ConfirmDeleteModalComponent} from 'app/views/confirm-delete-modal/component';
 import {EditModalComponent} from 'app/views/edit-modal/component';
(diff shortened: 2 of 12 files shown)


ui/src/app/views/resource-card/component.tsx
```diff
@@ -15,6 +20,84 @@ import {environment} from 'environments/environment';
...
+        ['conceptSet', () => {
+          return <React.Fragment>
+            <MenuItem icon='pencil' onClick={onEditConceptSet}>Edit</MenuItem>
+            <MenuItem icon='trash' onClick={onDeleteResource}>Delete</MenuItem>
+          </React.Fragment>;
+        }]
+      )
+    }
+  >
+    <Clickable disabled={disabled} data-test-id='resource-menu'>
```
panentheos: This `data-test-id` is a technique for identifying specific elements for testing purposes. Using a data attribute makes its meaning clear and avoids overloading CSS classes for that purpose.

ui/src/app/views/resource-card/component.spec.ts
```diff
@@ -88,30 +89,6 @@ describe('ResourceCardComponent', () => {
...
-  it('should have correct options in menu when it is a notebook', fakeAsync(() => {
-    setNotebookResource(fixture, component);
-    const de = fixture.debugElement;
-    simulateClick(fixture, de.query(By.css('.dropdown-toggle')));
-    const actionItems = de.queryAll(By.css('.action-item'))
-      .map((item) => item.nativeElement.innerText.trim());
-    ['Rename', 'Clone', 'Delete'].forEach((action) => {
-      expect(actionItems).toContain(action);
-    });
-  }));
```
panentheos: NB: Deleted this because this logic is now straightforward to verify by inspecting the code, and the other tests already do more thorough testing of the underlying behavior.

ui/src/app/views/resource-card/component.tsx
```diff
@@ -15,6 +20,84 @@ import {environment} from 'environments/environment';
 import {ConfirmDeleteModalComponent} from 'app/views/confirm-delete-modal/component';
 import {EditModalComponent} from 'app/views/edit-modal/component';
 
+const MenuItem = ({ icon, children, ...props }) => {
```
calbach: Should MenuItem be in a more central location, or is the styling specific to the resource menu? We have at least one other such menu in the application (workspace-nav-bar dropdown menu).

ui/src/app/utils/index.ts
```diff
@@ -192,8 +192,7 @@ export function reactStyles<T extends {[key: string]: React.CSSProperties}>(t: T
 export class ReactWrapperBase implements DoCheck, OnInit, OnDestroy {
   @ViewChild('root') rootElement: ElementRef;
 
-  constructor(private wrapped: (new (...args: any[]) => React.Component)|React.FunctionComponent,
-              private propNames: string[]) {}
+  constructor(private wrapped: React.ComponentType, private propNames: string[]) {}
```
calbach: Ah.. nice find.

ui/src/app/views/resource-card/component.tsx
panentheos: It will probably get used again, at which point we'll move it to a more central location. I tend to delay that decision until we have two concrete examples, because it's much easier to get the abstraction right.

calbach: Sounds good.

PR #1659, opened 2019-01-15
[risk=no] clean up account creation success modals

panentheos:
Cleans up and standardizes the account success modals.
* removes the unused angular wrapper
* converts render tests to jest
* renames and consolidates some props, e.g. `onDone`
* disables the 'Apply' button in the update email modal if the address is invalid


UrsaStutsman:
So this all looks good. But is tightly connected to the monster PR neha and I trying to keep organized. what I'm worried about particularly is merge conflicts... ... Trying to think of best way to manage this.

panentheos:
Per offline conversation, we're going to merge this first and then rebase the big branch onto it.

PR #1662, opened 2019-01-16
[RW-1972][risk=low] Add a swagger codegen fetch module

calbach:
This will allow for injection of fetch-based API stubs with proper credential plumbing and without explicit inclusion of the base URL. See the module comment for details.

I believe we'll likely want these kinds of services available via `props` once we're fully on React. @panentheos to sanity check that assertion.

CC: @UrsaStutsman 

Example usage:
- https://github.com/all-of-us/workbench/pull/1663
- https://github.com/all-of-us/workbench/compare/srubenst/settings-page...ch/fetch-fix?expand=1

### DIFF
--- a/ui/src/app/app.module.ts
+++ b/ui/src/app/app.module.ts
@@ -6,6 +6,7 @@ import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 import {ClarityModule} from '@clr/angular';
 import {NgxChartsModule} from '@swimlane/ngx-charts';
 import {environment} from 'environments/environment';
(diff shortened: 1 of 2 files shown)


panentheos:
It won't need to go in props, because it's not related to the rendering of components. It's sufficient to just import from a module and use it.

UrsaStutsman:
Awesome CH! Thank you so much!

calbach:
@panentheos I want to ensure we have a way of referencing these services in a non-static manner so we can easily stub out fakes for testing. Props may not be the right answer. Can discuss more on #1663 where this gets used.

PR #1667, opened 2019-01-16
[P1][RW-1971][risk=no] set --environment flag properly for public-ui deploy

jacmrob:

### DIFF
--- a/common-ui/common-ui-devstart.rb
+++ b/common-ui/common-ui-devstart.rb
@@ -143,7 +143,7 @@ class CommonUiDevStart
     Common.register_command({
                                 :invocation => "build",
                                 :description => "Builds the UI for the given environment.",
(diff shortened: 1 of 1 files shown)


common-ui/common-ui-devstart.rb
```diff
@@ -36,7 +36,7 @@ def swagger_regen(cmd_name)
   common.run_inline %W{yarn run codegen}
 end
 
-def build(cmd_name, args)
+def build(cmd_name, ui_name, args)
```
calbach: not provided by callers?

jacmrob: what does that mean?

calbach: Nevermind - I thought this PR introduced `@ui_name`

PR #1668, opened 2019-01-16
[RW-1971][P1][risk=no] Databrowser hotfix and deploy --environment flag

jacmrob:


common-api/src/main/java/org/pmiops/workbench/cdr/dao/SurveyModuleDao.java
```diff
@@ -16,7 +16,7 @@
   @Query(nativeQuery=true,value="select m.name, m.description,\n" +
       "m.concept_id, COUNT(DISTINCT c.concept_id) as question_count,\n" +
       // We don't show participant counts when filtering by keyword, and don't have a way of computing them easily; return 0.
-      "0 participant_count\n" +
+      "0 participant_count, m.order_number\n" +
```
calbach: Did you rebase already? You should have gotten merge conflicts with this file and workspaces controller

jacmrob: resolved

PR #1669, opened 2019-01-16
[RW-1898][risk=no] Edit Modal React

jacmrob:


ui/src/app/views/resource-card/component.ts
```diff
@@ -69,6 +67,7 @@ export class ResourceCardComponent implements OnInit {
       }
     }
     this.actions = this.actionList.filter(elem =>  elem.type === this.resourceType);
+    console.log(this.actions);
```
UrsaStutsman: nit: did you mean to leave this in?

jacmrob: nope! will remove

PR #1670, opened 2019-01-16
[1889][Risk=no] Srubenst/settings page

blrubenstein:


blrubenstein:
The rebase appeared to mess with the previous PR, so I closed it and reopened with a new copy of the branch on gh

ui/src/app/views/settings/component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+describe('SettingsComponent', () => {
+  let props: {};
+
+  const component = () => {
+    return mount<SettingsReact, {}, SettingsState>
+    (<SettingsReact {...props}/>);
+  };
+
+  beforeAll(() => {
```
jacmrob: why do you do this before tests?

ui/src/app/views/settings/component.tsx
```diff
@@ -0,0 +1,154 @@
...
+      });
+  }
+
+  private pollCluster(): void {
+    const repoll = () => {
+      this.pollClusterTimer = setTimeout(() => this.pollCluster(), 15000);
+    };
+    const args: FetchArgs = ClusterApiFetchParamCreator()
+      .listClusters();
+    console.log(args);
```
jacmrob: debug line?  or do we want this written to the console?

ui/src/app/views/settings/component.tsx
```diff
@@ -0,0 +1,154 @@
...
+        this.setState({cluster: cluster});
+      })
+      .catch(() => {
+        // Also retry on errors
+        repoll();
+      });
+  }
+}
+
+@Component({
```
jacmrob: do we need a `selector` in this?

ui/src/app/views/settings/component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {SettingsReact, SettingsState} from './component';
+
+import {Cluster, ClusterStatus} from 'generated/fetch/api';
+
+describe('SettingsComponent', () => {
+  let props: {};
+
+  const component = () => {
+    return mount<SettingsReact, {}, SettingsState>
+    (<SettingsReact {...props}/>);
```
calbach: rm newline? Breaking between the method name and the paren of the method invocation is not very readable

ui/src/app/views/settings/component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+    popupRoot.setAttribute('id', 'popup-root');
+    document.body.appendChild(popupRoot);
+  });
+
+  afterAll(() => {
+    document.removeChild(document.getElementById('popup-root'));
+  });
+
+  beforeEach(() => {
+    props = {};
```
calbach: Doesn't seem like there's much purpose to this - or are you setting up the convention / boilerplate for future tests?

ui/src/app/views/settings/component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+  it('should not open the cluster reset modal when no cluster', () => {
+    const wrapper = component();
+    expect(wrapper.find('Modal[data-test-id="reset-notebook-modal"]').length).toBe(0);
+    wrapper.find('[data-test-id="reset-notebook-button"]').at(0).simulate('click');
+    expect(wrapper.find('Modal[data-test-id="reset-notebook-modal"]').length).toBe(0);
+  });
+
+  it('should open the cluster reset modal when there is cluster', () => {
+    const wrapper = component();
+    wrapper.setState({cluster: {clusterName: 'testing',
```
calbach: I realize you'd be blocked on my PR to do this, but a better test would be to stub a cluster into the clusterApi stub rather than explicitly changing the state here.

ui/src/app/views/settings/component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+      clusterNamespace: 'testing',
+      status: ClusterStatus.Running}});
+    expect(wrapper.find('Modal[data-test-id="reset-notebook-modal"]').length).toBe(0);
+    wrapper.find('[data-test-id="reset-notebook-button"]').at(0).simulate('click');
+    expect(wrapper.find('Modal[data-test-id="reset-notebook-modal"]').length).toBe(1);
+  });
+});
```
calbach: One thing lost in translation is the test which actually clicks reset and checks that the delete happened.

ui/src/app/views/settings/component.tsx
blrubenstein: Removing

ui/src/app/views/settings/component.tsx
blrubenstein: Nope! This is actually a top level component, so it is invoked only by the router.

ui/src/app/views/settings/component.spec.tsx
blrubenstein: This is necessary to provide the modals a place to live. This should actually be done as part of a global setup for all tests, but I am not entirely sure where that is. If you know, let me know and I will move it?

ui/src/app/views/settings/component.tsx
```diff
@@ -0,0 +1,154 @@
...
+  componentDidMount() {
+    this.pollCluster();
+  }
+
+  componentWillUnmount() {
+    if (this.pollClusterTimer) {
+      clearTimeout(this.pollClusterTimer);
+    }
+  }
+
```
calbach: ultranit: rm extra \n

ui/src/app/views/settings/component.spec.tsx
calbach: I made an angular helper for this: https://github.com/all-of-us/workbench/blob/bddb11471e19d06cef7a7e46a54581d17f68605d/ui/src/testing/test-helpers.ts#L8

But I think global test setup might be a better place for it going forward with jest. I'd add it there if it's easy to do.

blrubenstein: I am not sure if that should go here:
https://github.com/all-of-us/workbench/blob/master/ui/test-setup.js#L10
Or here: 
https://github.com/all-of-us/workbench/blob/master/ui/test-shim.js#L10

I think in the setup file? But I don't know if that actually works. I can't find anything that actually does what we want here.

ui/src/app/views/settings/component.spec.tsx
blrubenstein: Setting up the boilerplate, but I don't think its actually important to build that out. Will remove.

ui/src/app/views/settings/component.tsx
blrubenstein: Will remove!

ui/src/app/views/settings/component.spec.tsx
calbach: I think test-setup is probably an OK place. If that doesn't work, you could alternatively try to use a new file with this different jest config property: https://jestjs.io/docs/en/configuration.html#setuptestframeworkscriptfile-string

ui/src/app/views/settings/component.spec.tsx
calbach: merging is complete

blrubenstein:
@calbach when you get a chance, I have figured out the global setup thing, and it should be good for final review!

ui/test-setup-script.js
```diff
@@ -0,0 +1,9 @@
+global.beforeEach(() => {
```
calbach: Nice

ui/src/testing/react-test-helpers.ts
```diff
@@ -0,0 +1,9 @@
+import {ReactWrapper} from 'enzyme';
+
+
+// This file is necessary because angular imports complain if there
```
calbach: What is the exact error you were seeing? AFAICT your `.spec.tsx` test should not have any dependency on Angular (though I guess there are some related imports in the file). Could you expand this comment and include when you intend a client to use this?

Also I definitely don't understand why there are two identical promise invocations below, a comment there would be good.

blrubenstein: It complained because there was no definition of `Zone`, which meant the imports failed.

blrubenstein: If I imported from the main test file.

PR #1674, opened 2019-01-17
[RW-1883][Risk=low] Account Creation React conversion and Account Creation Success cleanup

UrsaStutsman:


UrsaStutsman:
Merged as part of larger Pr

PR #1676, opened 2019-01-17
[RW-1972][risk=no] ts fetch swagger API clients attempt 2

calbach:
Per discussion today and on #1663. This is an alternate approach which places a registry layer between consumers and the binding of API clients, enabling React components to make direct reference to various API client layers. This allows for easy stubbing using our existing fake patterns.

Example usage now:

```
import {clusterApi} from 'app/services/swagger-fetch-clients';

clusterApi().listClusters();
```

Example test usage:
```
beforeEach(() => {
  registerApiClient(ProfileApi, new ProfileApiStub());
});
```

My prior was to use something more similar to dependency injection, but looks like that's a big lift that's not naturally support out of the box. This approach is a more hardcoded and direct answer specifically for API clients, which is the primary interface that we care about stubbing for tests. For other global data (e.g. isSignedIn), we can consider using something similar to the `atom()` pattern that @panentheos shared with me from Saturn.

### DIFF
--- a/ui/src/app/views/account-creation-modals/component.spec.tsx
+++ b/ui/src/app/views/account-creation-modals/component.spec.tsx
@@ -1,7 +1,18 @@
 import {shallow} from 'enzyme';
 import * as React from 'react';
 
+import {registerApiClient} from 'app/services/tsfetch';
+import {ProfileApiStub} from 'testing/stubs/profile-service-stub';
+
 import {AccountCreationResendModal, AccountCreationUpdateModal} from './component';
+
+import {
+  ProfileApi
+} from 'generated/fetch';
+
+beforeEach(() => {
+  registerApiClient(ProfileApi, new ProfileApiStub());
+});
 
 describe('AccountCreationResendModal', () => {
   it('should render', () => {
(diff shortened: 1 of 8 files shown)


calbach:
@UrsaStutsman @NehaBroad FYI

ui/src/app/views/account-creation-modals/component.spec.tsx
```diff
@@ -1,8 +1,19 @@
...
+import {ProfileApiStub} from 'testing/stubs/profile-service-stub';
+
 import {AccountCreationResendModal, AccountCreationUpdateModal} from './component';
 
+import {
+  ProfileApi
+} from 'generated/fetch';
+
+beforeEach(() => {
+  registerApiClient(ProfileApi, new ProfileApiStub());
```
blrubenstein: You might be able to do this as `beforeAll`, FYI.

ui/src/testing/stubs/profile-service-stub.ts
```diff
@@ -26,6 +31,12 @@ export class ProfileStubVariables {
   };
 }
 
+export class ProfileApiStub extends ProfileApi {
```
gjuggler: More naming nits:

When reading through the test code that uses this class, I found it pretty confusing that a file named profile-service-stub exports a class named ProfileApiStub. It feels like this class is somewhat shoved in here, especially since there *is* an exported class that matches the filename.

Will some of this go away once Angular is gone?

Or maybe making the filename more generic would help, e.g. "testing/stubs/profile.ts" (since "stub" is implied by the path, and this file now exports both an Api stub and a Service stub for the profile API).

PR #1681, opened 2019-01-17
[no ticket][Risk=no] Account creation follow up using new fetch, little cleanup

UrsaStutsman:

### DIFF
--- a/ui/src/app/views/account-creation/component.tsx
+++ b/ui/src/app/views/account-creation/component.tsx
@@ -3,14 +3,9 @@ import * as ReactDOM from 'react-dom';
 
 import * as fp from 'lodash/fp';
 
-import {fullUrl, handleErrors} from 'app/utils/fetch';
-
 import {
-  CreateAccountRequest,
   DataAccessLevel,
-  FetchArgs,
   Profile,
-  ProfileApiFetchParamCreator,
 } from 'generated/fetch/api';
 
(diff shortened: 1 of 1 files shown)


ui/src/app/views/account-creation/component.tsx
```diff
@@ -78,30 +79,24 @@ export class AccountCreation extends
 
   createAccount(): void {
     const {invitationKey, setProfile} = this.props;
+    const profile = this.state.profile;
+    const {givenName, familyName, username, contactEmail,
+      currentPosition, organization, areaOfResearch} = this.state.profile;
```
calbach: nit: `profile`. Also arguably I'm not sure I would redeclare these as they're only used once `profile.givenName` etc seems fine.

ui/src/app/views/account-creation/component.tsx
```diff
@@ -32,6 +32,7 @@ import {
 import {Button} from 'app/components/buttons';
 import { FormSection } from 'app/components/forms';
 import {BoldHeader} from 'app/components/headers';
+import {profileApi} from 'app/services/swagger-fetch-clients';
```
calbach: There should now be a bunch of unused imports from fetch and utils/fetch. If they aren't unused, then please replace other usages in this file as they shouldn't be needed.

PR #1682, opened 2019-01-17
[RW-1829][RISK=LOW]React - login page Test 

NehaBroad:
Test for sign in, login and invitation key


ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {ProfileApi} from 'generated/fetch';
+import {ProfileApiStub} from 'testing/stubs/profile-api-stub';
+
+describe('InvitationKeyComponent', () => {
+  let props: InvitationKeyProps;
+  const onInvitationVerify = jest.fn();
+
+
+  const component = () => {
+    return mount<InvitationKeyReact>
```
calbach: Please remove the newline, possibly this is something that can be automated by our linter, but we really should avoid putting a newline between method name and paren to invoke that method.

Also is the generic type actually necessary here?

ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    (<InvitationKeyReact {...props}/>);
+  };
+
+  beforeEach(() => {
+    props = {onInvitationKeyVerify: onInvitationVerify};
```
calbach: (1) why is the test property named differently?
(2) if they had the same name, this could be simplified to just: `props = {onInvitationKeyVerify}`

ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const wrapper = component();
+    const nextButton = wrapper.find(Button);
+    nextButton.simulate('click');
+    const error = wrapper.find(AlertDanger);
+    expect(error.text()).toBe('Invitation Key is required.');
+  });
+
+  it('should display error message if Invitation key is not valid', async() => {
+    profileApi().invitationKeyVerification = jest.fn().mockRejectedValue(() => {
+          throw new Error('test error inside');
```
calbach: nit: 2sp indent

ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    input.simulate('change', {target: {value: 'notValid '}});
+    nextButton.simulate('click');
+    await new Promise(setImmediate).then(() => wrapper.update());
+    const error = wrapper.find(AlertDanger);
+    expect(error.text()).toBe('Invitation Key is not Valid.');
+  });
+
+  it('should call props onInvitationKeyVerify function on entering correct invitation key',
+      async() => {
```
calbach: (again, maybe something we can lint) : please put a space between async and the `()`, this makes it look like a method invocation

ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+        profileApi()
```
calbach: Alternatively, don't mock here and instead just add a successful fake implementation to the profileApi stub

ui/src/app/views/invitation-key/component.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+            .invitationKeyVerification = jest.fn()
```
calbach: Please improve line wrapping, e.g. `profileApi().invitationKeyVerification =`

ui/src/app/views/login/component.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+describe('LoginComponent', () => {
+  let props: {signIn: Function,onCreateAccount:Function};
+
+  const signIn = jest.fn();
+  const onCreateAccount = jest.fn();
+
+  const component = (props) => {
+    return mount
+    (<LoginReactComponent {...props}/>);
```
calbach: line wrapping - see other comment

ui/src/app/views/sign-in/component.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const signIn = jest.fn();
+
+  const component = () => {
+    return mount(<SignInReact {...props}/>);
+  };
+
+  beforeEach(() => {
+    props = {
+      onInit: jest.fn(),
```
calbach: If you don't use this, I'd just give an empty arrow function rather than creating a mock

ui/src/app/views/invitation-key/component.tsx
```diff
@@ -68,7 +68,7 @@ export class InvitationKey extends React.Component<InvitationKeyProps, Invitatio
   }
 
   render() {
-    return <div style={{padding: '3rem 3rem 0 3rem'}}>
+    return <div data-test-id='invitationKey' style={{padding: '3rem 3rem 0 3rem'}}>
```
panentheos: This should probably match the 'invitation' value used in the tests (currently obscured by `exists` issue)

ui/src/app/views/sign-in/component.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  it('should display small background image when window width is between 900 and 1300', () => {
+    props.windowSize.width = 999;
+    const wrapper = component();
+    const templateImage = wrapper.find('[data-test-id="template"]');
+    const backgroundImage = templateImage.prop('style').backgroundImage;
+
+    expect(backgroundImage)
+        .toBe('url(\'' + pageImages.login.smallerBackgroundImgSrc + '\')');
+    expect(wrapper.find('[data-test-id="invitation"]')).toBeTruthy();
```
panentheos: What code is this particular assertion trying to exercise? (note: value is incorrect, but obscured by the `exists` issue)

ui/src/app/views/invitation-key/component.spec.tsx
panentheos: FYI: the linter currently demands no space here, because it seems to read this as a function declaration.

PR #1683, opened 2019-01-18
[RW-1829][RISK=LOW] Replace fetch logic for invitation key verification

NehaBroad:


ui/src/app/views/invitation-key/component.tsx
```diff
@@ -62,8 +58,7 @@ export class InvitationKey extends React.Component<InvitationKeyProps, Invitatio
       invitationKey: this.state.invitationKey
     };
 
-    const args: FetchArgs = ProfileApiFetchParamCreator().invitationKeyVerification(request);
-    fetch(fullUrl(args.url), args.options)
+    profileApi().invitationKeyVerification(request)
```
calbach: nit: I would just inline the request object from above - not much value in this local var (also means you can avoid importing the type).

NehaBroad: Updated PR as per the comment

PR #1684, opened 2019-01-18
[RW-1902][RISK=NO] React - icons/left

NehaBroad:


ui/src/app/icons/left-scroll-light/component.tsx
```diff
@@ -0,0 +1,67 @@
...
+      fillColor: '#5FAEE0'
+    };
+  }
+
+  mouseOver(): void {
+    this.setState({fillColor: '#1892E0'});
+  }
+
+  mouseLeave(): void {
+    this.setState({fillColor: '#5FAEE0'});
```
blrubenstein: Could this color be a variable?

ui/src/app/icons/left-scroll/component.tsx
```diff
@@ -0,0 +1,64 @@
...
+      fillColor: '#2691D0'
+    };
+  }
+
+  mouseOver(): void {
+    this.setState({fillColor: '#72B9E2'});
+  }
+
+  mouseLeave(): void {
+    this.setState({fillColor: '#2691D0'});
```
blrubenstein: Likewise with this color, since its used in two places.

ui/src/app/icons/scroll/component.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface ScrollProps {
+  direction: string;
+}
+
+export const ScrollReact: React.FunctionComponent<{direction: string}> = ({direction}) => {
+     return  <div>
+       {
+         (() => {
+           switch (direction) {
+             case 'left':
```
blrubenstein: Do you know what the default behavior here is?

rushtong:
Closing this. The solution in https://github.com/all-of-us/workbench/pull/1684 will cover the requirements for RW-1901

rushtong:
Oops ... closed the wrong PR.

ui/src/app/icons/scroll/component.tsx
```diff
@@ -0,0 +1,50 @@
...
+        opacity = '0.54';
+        width = '40';
+        height = '41';
+      }
+
+      return  <div>
+        <Interactive as={wrapper}
+          style={{color: color, opacity: opacity, transform: transform, ...style}}
+          hover={{color: hoverColor}} width= {width} height={height}/>
+      </div>;
```
panentheos: Since there are a lot of differences between the two cases, it might be cleaner to express this as a single conditional, and then inline the relevant props/style. That would allow you to avoid using `let` above (which should be avoided unless you really need mutation).

E.g.
```
kind === 'one' ?
  <Foo a='1' b='2' c='3' /> :
  <Foo a='7' b='8' c='9' />
```

ui/src/app/icons/scroll/component.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expect(darkShadeSvgProp.style).not.toBe(lightShadeSvgProp.style);
+    expect(lightShadeSvgProp.style.color).toBe('#5FAEE0');
+    expect(darkShadeSvgProp.style.color).toBe('#2691D0');
+    expect(lightShadeSvgProp.style.opacity).toBe('0.54');
+    expect(darkShadeSvgProp.style.opacity).toBe('1');
+    expect(lightShadeSvgProp.width).toBe('40');
+    expect(darkShadeSvgProp.width).toBe('47');
+    expect(lightShadeSvgProp.height).toBe('41');
+    expect(darkShadeSvgProp.height).toBe('48');
+  });
```
panentheos: TOL: This seems like an awful lot of test code for what amounts to a single conditional operator. What are we expecting to fail here?

ui/src/app/views/recent-work/component.ts
```diff
@@ -22,6 +22,8 @@ import {convertToResources, ResourceType} from 'app/utils/resourceActions';
 })
 
 export class RecentWorkComponent implements OnInit {
+  leftLightIcon = 'leftLight';
+  leftIcon = 'left';
```
panentheos: Are these used anywhere?

ui/src/app/icons/scroll/component.tsx
```diff
@@ -0,0 +1,50 @@
...
+import * as React from 'react';
+import * as Interactive from 'react-interactive';
+
+const FillColorId = '#2691D0';
+const HoverFillColorId = '#72B9E2';
+const LightFillColorId = '#5FAEE0';
+const LightHoverFillColorId = '#1892E0';
+
+const wrapper = ScrollIconReact;
+export const ScrollReact: React.FunctionComponent<{icon: string, shade: string, style: object}> =
```
panentheos: Let's stop putting 'React' in the name, since that seems somewhat redundant. Going forward I think we should follow the pattern of naming react components like `FooBar`, and the corresponding angular wrapper like `FooBarComponent`. Thoughts?

ui/src/app/views/recent-work/component.ts
NehaBroad: This has been removed

ui/src/app/icons/scroll/component.tsx
NehaBroad: I agree, have updated the file accordingly.

ui/src/app/icons/scroll/component.spec.tsx
NehaBroad: I went too deep. Have replaced the test with NOT to be i.e The light shade icons width/opacity and color will not be same as that of dark shade icon

ui/src/app/icons/scroll/component.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const darkShadeWrapper = component();
+    const darkShadeSvgProp = darkShadeWrapper.find('svg').props();
+    props.shade = 'Light';
+
+    const lightShadeWrapper = component();
+    const lightShadeSvgProp = lightShadeWrapper.find('svg').props();
+
+    expect(darkShadeSvgProp.style).not.toBe(lightShadeSvgProp.style);
+    expect(lightShadeSvgProp.style.color).toBe('#5FAEE0');
+    expect(darkShadeSvgProp.style.color).toBe('#2691D0');
```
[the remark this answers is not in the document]
NehaBroad: updated the test to validate that the light shade icons opacity color and width are not same as that of dark shade

ui/src/app/icons/scroll/component.tsx
```diff
@@ -0,0 +1,50 @@
...
+import {ReactWrapperBase} from 'app/utils';
+
+import {ScrollIconReact} from 'app/icons/scroll-icon/component';
+import * as React from 'react';
+import * as Interactive from 'react-interactive';
+
+const FillColorId = '#2691D0';
+const HoverFillColorId = '#72B9E2';
+const LightFillColorId = '#5FAEE0';
+const LightHoverFillColorId = '#1892E0';
```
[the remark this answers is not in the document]
NehaBroad: updated

PR #1687, opened 2019-01-18
[risk=no] expose navigation for react components

panentheos:
This exposes the `navigate` and `navigateByUrl` methods via a module, so they can be used by React components without injecting.

Usage:
```
import {navigate} from 'app/utils/navigation';

<Button onClick={() => navigate(['somewhere'])}>
```

Note: Angular routing seems to be extremely fragile, which is why the code is formulated the way it is. Specifically, the actual call to `navigate` must originate from within a Component class, so we need to store callback functions to make that happen.

### DIFF
--- a/ui/src/app/utils/navigation.tsx
+++ b/ui/src/app/utils/navigation.tsx
@@ -1,0 +1,14 @@
+export const NavStore = {
+  navigate: undefined,
+  navigateByUrl: undefined
+};
+
+// NOTE: Because these are wired up directly to the router component,
+// all navigation done from here will effectively use absolute paths.
+export const navigate = (...args) => {
+  return NavStore.navigate(...args);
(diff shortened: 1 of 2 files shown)


ui/src/app/utils/navigation.tsx
```diff
@@ -0,0 +1,14 @@
+export const NavStore = {
+  navigate: undefined,
+  navigateByUrl: undefined
+};
+
+// NOTE: Because these are wired up directly to the router component,
+// all navigation done from here will effectively use absolute paths.
+export const navigate = (...args) => {
```
calbach: Is it possible to just attach the `Router` here instead or are there issues? It may be needed in a few places e.g. for event subscription and it would keep type safety on the methods.

PR #1692, opened 2019-01-18
[risk=no] add withStyle

panentheos:
Adds a little HOC to abstract the pattern of applying styles to a base element, while still allowing explicit overrides. Forwards `ref` to the wrapped element.

### DIFF
--- a/ui/src/app/utils/index.tsx
+++ b/ui/src/app/utils/index.tsx
@@ -219,3 +219,11 @@ export function decamelize(str: string, separator: string) {
       .replace(/([A-Z]+)([A-Z][a-z\d]+)/g, '$1' + separator + '$2')
       .toLowerCase();
 }
+
+export const withStyle = styleObj => WrappedComponent => {
+  const Wrapper = React.forwardRef(({ style = {}, ...props }: any, ref) => {
+    return <WrappedComponent ref={ref} style={{...styleObj, ...style}} {...props} />;
+  });
+  Wrapper.displayName = 'withStyle';
+  return Wrapper;
+};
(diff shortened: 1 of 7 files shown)


ui/src/app/utils/index.tsx
```diff
@@ -219,3 +219,11 @@ export function decamelize(str: string, separator: string) {
       .replace(/([A-Z]+)([A-Z][a-z\d]+)/g, '$1' + separator + '$2')
       .toLowerCase();
 }
+
+export const withStyle = styleObj => WrappedComponent => {
```
gjuggler: (This is a naming question mostly unrelated to this PR per se, so feel free to ignore and/or ask me to bring up in a future tech talk.)

What's the motivation behind putting this HOC / helper function into a file named "index"? Is there a logical association between this and other functions in the file, or has index.tsx become a dumping ground of sorts?

If it's the latter... I'm wondering if we can take some small action now to prevent this from exploding over time.

Taking a quick look through the file, it looks like we have a few categories:
 - String utils (--> utils/strings.ts?)
 - DOM utils (--> utils/dom.ts?)
 - React utils (--> utils/react.tsx?)

Just a thought!

panentheos: The common theme in that file is small, generic, reusable utilities. While it's certainly possible to  apply some human-oriented taxonomy to these, there will inevitably be some dependencies among them, which can turn into circular dependencies if the files are split up. Webpack can resolve certain kinds of circular dependencies, but not all, so it's worth being careful of.

If we decide that's something we want to explore, let's do it in a separate PR.

PR #1693, opened 2019-01-19
[RW-1980] [risk=low] code cleanup

zarsky-broad:
I'm not sure who else to tag on this, so feel free to add anyone.

Functional changes:
- QuickTourModalComponent rendering now controlled via `*ngIf`
- AlertDanger now implements Alert
- Consolidated inline styles in `index.html` to css

~Dev changes:~ all now #1702
- ~Yarn scripts `dev-up` and `test-react` added~
- ~Pre-push lint will now run without docker if yarn is installed locally~
- ~Swagger codegen can now happen without docker, since ruby is already presume to be installed locally~
- ~Git-secrets scripts should hopefully be a bit more robust now?~
- ~De-duplicated packages & upgraded tslint~

Nonfunctional changes:
- Added a few lint rules in consultation with @blrubenstein 
- Linted files (spacing normalized, removed unused imports, alphabetized imports, made all import paths absolute)
- ~Extensively optimized svg's by running them through [svgomg](https://jakearchibald.github.io/svgomg/) (and in two cases converting them to png's)~ now #1700


gjuggler:
We had a timely discussion today in our retrospective about valuing small PRs :)

I wouldn't want to unnecessarily stall progress on these obvious wins... but maybe there's an opportunity to quickly break this into a few thematically-grouped PRs? At the very least, it could spread the review load a bit.

The three headings in your description (functional / non-functional / dev changes) seem like a good place to start. The linting results would also be an easy / mechanical review if it were separated out.

That said, I'm happy to defer if I'm outnumbered, and/or if someone jumps in with a heroic quick review :P

ui/src/app/views/homepage/component.css
```diff
@@ -377,7 +377,7 @@
 }
 
 .image {
-  height: 136px;
+  max-height: 136px;
```
calbach: Is this related to the svg changes?

hooks/pre-push
```diff
@@ -30,7 +30,11 @@ do
     pushd "${dir}" > /dev/null
     if [ -n "`echo ${files} | grep ^${dir}\/.*\.ts$`" ]; then
       echo "Linting ${dir}..."
-      ./project.rb lint
+      if [ -x "$(command -v yarn)" ]; then
+        yarn run lint
```
calbach: Out of curiosity, what difference in runtime did you see from this?

common-ui/tslint.json
```diff
@@ -146,8 +143,6 @@
       "Layout"
     ],
     "directive-class-suffix": true,
-    "no-access-missing-member": true,
```
calbach: I believe these were stale and don't work on the latest Angular - can you confirm that is why these 3 were removed?

common-ui/tslint.json
```diff
@@ -60,6 +61,7 @@
       true,
       "ignore-params"
     ],
+    "no-irregular-whitespace": true,
```
calbach: Note: while you're here, a style inconsistency that has been grating on me is our mixture of absolute and relative module imports. I haven't tried it out, but I found `no-relative-imports` which might alleviate this. Anyways - if that rule is controversial or doesn't work as expected, no need to touch it in this PR.

UrsaStutsman: +1 to adding that rule

ui/src/app/views/homepage/component.css
zarsky-broad: Yup, to prevent png's from getting stretched.

hooks/pre-push
zarsky-broad: Everything is to enable running without docker. I'm working on swagger gen happening without docker, too.

common-ui/tslint.json
zarsky-broad: They're not even in the current tslint.

gjuggler:
Thanks for splitting things up! 👍

zarsky-broad:
@calbach want to re-review? or am I good?

common-ui/tslint.json
calbach: Hmm I don't see the new rule here (but I did see you updated all the import paths - thanks!). Am I missing something?

zarsky-broad: Because it requires an external import, and there's no `package.json` here, I put the rule in `ui/tslint.json`, where the import actually happens.

calbach: Ah - I see it now. Thanks.

PR #1696, opened 2019-01-22
[risk=no] standardize text inputs

panentheos:
Introduces `TextInput` and `TextArea` components to standardize the basic text input appearance. The exact design may be tweaked with e.g. nicer focus outlines and success/error states, but I wanted to get the abstraction in place so people can start using it.

n.b. Width defaults to `100%`, but can be overridden via `style`

### DIFF
--- a/ui/src/app/views/account-creation/component.tsx
+++ b/ui/src/app/views/account-creation/component.tsx
@@ -8,12 +8,7 @@ import {
   Profile,
 } from 'generated/fetch/api';
 
-import {
-  Error,
-  ErrorMessage,
-  LongInput,
-  styles as inputStyles
-} from 'app/components/inputs';
+import { Error, ErrorMessage, styles as inputStyles, TextArea, TextInput } from 'app/components/inputs';
 
 import {
   InfoIcon,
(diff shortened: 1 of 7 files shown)


ui/src/app/views/account-creation/component.tsx
```diff
@@ -163,69 +158,48 @@ export class AccountCreation extends
...
-                       style={(givenName.length > 80) ?
-                         inputStyles.unsuccessfulInput : inputStyles.successfulInput}
-                       onChange={e => this.updateProfile('givenName', e.target.value)}/>
-            {givenName.length > 80 &&
-            <ErrorMessage id='givenNameError'>
-              First Name must be 80 characters or less.
-            </ErrorMessage>}
+                       style={{width: '16rem'}}
+                       maxLength={80}
+                       onChange={v => this.updateProfile('givenName', v)}/>
```
UrsaStutsman: Did we get this change approved? Meaning losing the error messages

ui/src/app/views/account-creation/component.tsx
```diff
@@ -163,69 +158,48 @@ export class AccountCreation extends
...
-                       style={(currentPosition.length > 255) ?
-                         inputStyles.unsuccessfulInput : inputStyles.successfulInput}
-                       onChange={e => this.updateProfile('currentPosition', e.target.value)}/>
-            {currentPosition.length > 255 &&
-            <ErrorMessage id='currentPositionError'>
-              Current Position must be 255 characters or less.
-            </ErrorMessage>}
+                       maxLength={255}
+                       style={{width: '16rem'}}
+                       onChange={v => this.updateProfile('currentPosition', v)}/>
```
UrsaStutsman: Same question here,  has loss of this visual cue been approved?

ui/src/app/views/account-creation/component.tsx
panentheos: I chatted informally with @blrubenstein about this. The thinking was that it would be just as effective to prevent the error entirely with an enforced `maxLength`.

UrsaStutsman: Ok. I just know there has been conversation around this sort of UI question before and the thinking was that to have the form "just stop working" and not indicate to the user what the problem is can be confusing. Probably more of an edge case, but just putting out what has been brought up before.

panentheos: FYI I backed out the validation change in the interest of getting this merged.

PR #1697, opened 2019-01-22
Db 201 [risk=no] Removing participant photos from db page

SrushtiGangireddy:
1. Removing participant photos from db page.


public-ui/src/app/views/pm/pm.component.html
```diff
@@ -55,6 +55,5 @@ <h5 class="box-title"> {{selectedGroup.groupName}}</h5>
 
   </div>
   <div class="col-xs-12 col-sm-12 col-md-3">
-    <img [src]="pageImage" />
```
calbach: pageImage is a property somewhere, right? Unused now? Also can this image resource be deleted now?

SrushtiGangireddy: Removed page-image property and unused resources.

PR #1702, opened 2019-01-23
[risk=no]easy dev without docker

zarsky-broad:
- Yarn scripts `dev-up` and `test-react` added
- Pre-push lint will now run without docker if yarn is installed locally
- Swagger codegen can now happen without docker, since ruby is already presume to be installed locally
- Git-secrets scripts should hopefully be a bit more robust now?
- De-duplicated packages & upgraded tslint

Broken out from #1693 as requested


hooks/pre-commit
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-./hooks/apply-git-secrets.sh
+eval "$( cd "$(dirname "$0")" ; pwd -P )/apply-git-secrets.sh"
```
calbach: FYI my usual way of getting the git root is `git rev-parse --show-toplevel`, might be slightly simpler here.

PR #1703, opened 2019-01-23
[RW-1892][risk=no] convert new notebook modal to react

panentheos:
See story for details. Introduces the `validate.js` library to enable cleaner validation of forms.


ui/src/app/components/modals.tsx
```diff
@@ -12,8 +12,7 @@ const styles = {
 
   overlay: {
     backgroundColor: 'rgba(49, 49, 49, 0.85)', padding: '1rem', display: 'flex',
-    position: 'fixed', left: 0, right: 0, top: 0, bottom: 0, overflowY: 'auto',
-    zIndex: 1040
```
blrubenstein: Could you verify that removing this doesn't have side effects on anything else?

panentheos: Good call, needed to create a stacking context for the recent work left icon.

jacmrob: sorry didn't finish reviewing this before it got merged, but that was going to be my one comment - that we need that for the menu modals to stack properly

panentheos: FYI, since modals and popups are rendered in a 'portal' at the end of the document, we can now rely on element order for correct stacking in most cases. I fixed the left icon by creating a new stacking context around it at z-index 0 to maintain that baseline, which avoids the 'arms race' of ever increasing numbers.

PR #1704, opened 2019-01-23
[RW-1695][risk=low] Call out to FC for NIH linked username info

rushtong:
## Addresses 
[RW-1695](https://precisionmedicineinitiative.atlassian.net/browse/RW-1695)

This PR adds a new service call to FireCloud to get/store NIH information. It also updates the user's profile call to populate required fields in the user profile. I have not implemented any kind of update to the user's NIH information, that should be handled in [RW-1698](https://precisionmedicineinitiative.atlassian.net/browse/RW-1698)

## Testing
You can test this by:
* First examining the profile json returned from the profile call. It should normally show a null value for the linked nih username. 
* Next, log into https://firecloud.dsde-dev.broadinstitute.org/ with your aou account.
* Go to the profile page
* Follow the NIH login link - this takes you to a testing page that's not real, but does update FC dev.
* Enter `mbemis` into the username field and click the `Link to FireCloud` link
* Go back to AoU and refresh the profile page
* Examine the profile json and now the value should show `mbemis`


dmohs:
I appreciate how mall this bite is. Quick glance over looks good to me.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -277,4 +283,30 @@ public boolean isUserMemberOfGroup(String groupName) {
...
+      }
+    });
+  }
+
+  @Override
+  public void postNihCallback(JWTWrapper wrapper) {
+    NihApi nihApi = nihApiProvider.get();
+    retryHandler.run((context) -> {
+      nihApi.nihCallback(wrapper);
+      return null;
```
NehaBroad: Since the return type is void, just return should be fine.

api/src/test/java/org/pmiops/workbench/firecloud/FireCloudServiceImplTest.java
```diff
@@ -118,4 +126,36 @@ public void testIsUserMemberOfGroup_match() throws Exception {
...
+
+  @Test
+  public void testNihCallback() throws Exception {
+    JWTWrapper wrapper = new JWTWrapper().jwt("random");
+    doNothing().when(nihApi).nihCallback(wrapper);
+    try {
+      service.postNihCallback(wrapper);
+    } catch (Exception e) {
+      fail();
+    }
```
NehaBroad: This test just confirms that service.postNihCallBack does not fail? is there anything that can be tested in this like if nihApi.nihCallback(wrapper) was called?

rushtong: I will flesh the tests out for this method in RW-1698, which will have code that actually uses it. For this PR (which doesn't use it), this is just a basic test and I was intentionally leaving out anything more complicated.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
rushtong: Unfortunately, that doesn't compile. It has to return something from inside the retry handler. You can see this same pattern in other void methods in this class.

rushtong:
@NehaBroad @dmohs  Comments addressed, PTAL, thanks!

rushtong:
Thank you!

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -366,7 +367,13 @@ private User initializeUserIfNeeded() {
   }
 
   private ResponseEntity<Profile> getProfileResponse(User user) {
-    return ResponseEntity.ok(profileService.getProfile(user));
+    Profile profile = profileService.getProfile(user);
```
calbach: This breaks account creation:

```
org.pmiops.workbench.exceptions.ExceptionAdvice serverError: org.springframework.beans.factory.BeanCreationException (ExceptionAdvice.java:45)
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'nihApi' defined in class path resource [org/pmiops/workbench/firecloud/FireCloudConfig.class]: Unsatisfied dependency expressed through method 'nihApi' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'endUserApiClient' defined in class path resource [org/pmiops/workbench/firecloud/FireCloudConfig.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.pmiops.workbench.firecloud.ApiClient]: Factory method 'fireCloudApiClient' threw exception; nested exception is java.lang.NullPointerException
```

Translation: there is no FC "end user client" (which NihApi requires) at this point during account creation, since the user is not logged in at that point.

PR #1706, opened 2019-01-24
[risk=no] improve notebook rename flow

panentheos:
Cleans up the code and UX of the notebook rename modal. Notes:
* The modal now handles the rename itself, rather than delegating out to the parent container, so it should work properly and consistently in all locations.
* Performs inline validation to give the user feedback sooner. Pre-loads the notebook list when the modal mounts in order to provide duplicate name validation.
* Cleans up unused code in notebook list and concept set list.

Note: The two commented error paths should be entirely unexpected/unknown errors, so I think just logging is ok for the moment, but we should come up with a more robust solution for those kinds of error cases soon.

### DIFF
--- a/ui/src/app/views/concept-set-list/component.html
+++ b/ui/src/app/views/concept-set-list/component.html
@@ -37,8 +37,7 @@
     <div style="display: flex; flex-wrap: wrap;">
       <div *ngFor="let conceptSet of resourceList">
         <app-resource-card [resourceCard]="conceptSet" [cssClass]="'for-list'"
-                           (onUpdate)="loadConceptSets()"
-                           (duplicateNameError)="duplicateNameError($event)">
+                           (onUpdate)="loadConceptSets()">
         </app-resource-card>
       </div>
     </div>
@@ -46,7 +45,3 @@
 </app-top-box>
 
 <app-create-concept-modal (onUpdate)="loadConceptSets()"></app-create-concept-modal>
-<clr-modal [(clrModalOpen)]="nameConflictError">
-  <h3 class="modal-title">Error:</h3>
-  <div class="modal-body">You already have a concept set named {{duplicateName}}. Please choose another name.</div>
-</clr-modal>
--- a/ui/src/app/views/notebook-list/component.html
+++ b/ui/src/app/views/notebook-list/component.html
@@ -23,9 +23,7 @@
     <div style="display: flex; flex-wrap: wrap; justify-content: flex-start;">
       <div *ngFor="let notebook of resourceList">
         <app-resource-card [resourceCard]="notebook" [cssClass]="'for-list'"
-                           (onUpdate)="updateList($event)"
-                           (duplicateNameError)="duplicateNameError($event)"
-                           (invalidNameError)="invalidNameError($event)">
+                           (onUpdate)="updateList($event)">
         </app-resource-card>
       </div>
     </div>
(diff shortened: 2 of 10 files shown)


ui/src/app/views/concept-set-list/component.html
```diff
@@ -37,16 +37,11 @@
     <div style="display: flex; flex-wrap: wrap;">
       <div *ngFor="let conceptSet of resourceList">
         <app-resource-card [resourceCard]="conceptSet" [cssClass]="'for-list'"
-                           (onUpdate)="loadConceptSets()"
-                           (duplicateNameError)="duplicateNameError($event)">
+                           (onUpdate)="loadConceptSets()">
```
jacmrob: should `onUpdate` be in square braces instead of parens?  also do you need to bind loadConceptSets()? (`loadConceptSets.bind(this)`)

ui/src/app/views/notebook-list/component.html
```diff
@@ -23,9 +23,7 @@ <h3 class="notebooks-header">
     <div style="display: flex; flex-wrap: wrap; justify-content: flex-start;">
       <div *ngFor="let notebook of resourceList">
         <app-resource-card [resourceCard]="notebook" [cssClass]="'for-list'"
-                           (onUpdate)="updateList($event)"
-                           (duplicateNameError)="duplicateNameError($event)"
-                           (invalidNameError)="invalidNameError($event)">
+                           (onUpdate)="updateList($event)">
```
jacmrob: ditto to above comment

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -1,72 +1,117 @@
...
 import * as React from 'react';
 
 
+const fullName = name => {
+  return !name || /^.+\.ipynb$/.test(name) ? name : `${name}.ipynb`;
+};
+
 interface RenameModalProps {
-  resource: { name: string };
+  resource: {name: string};
```
jacmrob: i am kind of against passing in this pseudo-resource-- i was finding some difficulty with type stuff trying to give it a "full" resource (either notebook, cohort, or concept set).  if all we're using is the name, let's just pass `resourceName: string` as the prop.  I made this change for the confirm delete modal, it reduced some of the complexity in the modal itself

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  WorkspacesApi
+} from 'generated/fetch';
+
+export class WorkspacesApiStub extends WorkspacesApi {
```
jacmrob: why do we need this?

ui/src/app/views/rename-modal/component.tsx
panentheos: Good point, it would be cleaner that way.

ui/src/app/views/concept-set-list/component.html
panentheos: Note, I didn't add this line, it's just preserving what's there. Since the resource card refactor hasn't happened yet, these are still angular-style bindings.

ui/src/testing/stubs/workspaces-api-stub.ts
panentheos: Mounting the component in tests calls the `componentDidMount` lifecycle hook, which tries to make an ajax request, so we need to stub it out.

ui/src/app/views/concept-set-list/component.html
jacmrob: oh riiiight sorry i am working on the resource card refactor now and am used to seeing it that way.

PR #1707, opened 2019-01-25
[RW-2000][risk=no] Enable prod mode for non-local builds

calbach:
See https://precisionmedicineinitiative.atlassian.net/browse/RW-2000

`enableProdMode()` docs: https://angular.io/api/core/enableProdMode


calbach:
Tested via manual deploy.

PR #1708, opened 2019-01-28
[risk=no] add withCurrentWorkspace

panentheos:
This adds an HOC to enable components to access the current workspace from the route. It injects a prop `workspace` of type `WorkspaceData`. Usage is similar to other HOCs:

```const MyComponent = withCurrentWorkspace()(BaseComponent)```


blrubenstein:
(I am going to wander over to talk faces, because we already have something in Angular that does something like this with the router, and I want to make sure you have seen that as well.)

PR #1709, opened 2019-01-28
[RW-1385][risk=no] Add undo/suppress functions for search groups and items 

dolbeew:
Add options to undo delete and suppress from total count to search groups and search items on the overview page of cohort builder search. Also added snowman menus and updated styles to match new design.


ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -178,62 +178,101 @@ export class CohortSearchActions {
...
+      } else {
+        // todo clear total count
+        this.clearTotalCount();
+      }
     }
   }
 
-  removeGroupItem(role: keyof SearchRequest, groupId: string, itemId: string): void {
+  removeGroupItem(
+    role: keyof SearchRequest,
```
freemabd1: Seems we have a bug here when removing the last search group item the status is set to 'deleted' causing a 500 on group count and total count never getting called.

ui/src/app/cohort-search/redux/actions/creators.ts
```diff
@@ -291,6 +300,21 @@ export const clearWizardFocus =
   (): ActionTypes[typeof CLEAR_WIZARD_FOCUS] =>
   ({type: CLEAR_WIZARD_FOCUS});
 
+export const hideGroup =
+  (groupId: string, status: string
+  ): ActionTypes[typeof HIDE_GROUP] =>
+    ({type: HIDE_GROUP, groupId, status});
```
freemabd1: when hiding a search group, it seems to show a count of 0 or empty when another group is present.

ui/src/app/cohort-search/redux/reducer.ts
```diff
@@ -385,14 +395,33 @@ export const rootReducer: Reducer<CohortSearchState> =
...
+            return item.get('id') !== action.itemId && item.get('status') === 'active';
+          });
+        if (!activeItems.size) {
+          state = state.setIn(['entities', 'groups', action.groupId, 'count'], null);
+        }
+        return state.setIn(['entities', 'items', action.itemId, 'status'], action.status);
+      }
+
+      case HIDE_GROUP:
+        return state.setIn(['entities', 'groups', action.groupId, 'status'], action.status);
```
freemabd1: When hiding multiple AND'ed search groups, seems that the cancel isn't called between each hide group call. Total count is calculating wrong.

ui/src/app/cohort-search/redux/actions/creators.ts
dolbeew: Changed all to empty, may change again based on feedback

ui/src/app/cohort-search/redux/actions/service.ts
dolbeew: Fixed by pausing the delete timer when the wizard is open, but that may change depending on feedback

ui/src/app/cohort-search/redux/reducer.ts
dolbeew: Should be fixed

ui/src/app/cohort-search/redux/actions/creators.ts
dolbeew: Changed total count back to 0 based on CB team recommendation

ui/src/app/cohort-search/redux/actions/service.ts
dolbeew: Removed pausing of delete timer. Will now keep empty groups when all items are deleted per recommendation of CB team

PR #1710, opened 2019-01-28
[RW-1698][risk=low] NIH Callback Endpoint

rushtong:
## Addresses
[RW-1698](https://precisionmedicineinitiative.atlassian.net/browse/RW-1698)

## Notes
Adds a callback url, `/nih-callback`, that is intended to be used as a UI endpoint to update a user's NIH login status. Links out to the NIH (or eRA Commons) site will be redirected with an extra token added to the url. The UI part of this parses that string from the query parameter, posts it to FireCloud, then redirects the user back to the home page.

Requires the provided callback url to be in the form of: `/nih-callback?token={token}` where `{token}` is replaced by Shibboleth with a valid, encoded token string representing the user's NIH username.

## Local Testing
You can test a local instance using these examples:
* Success: http://localhost:4200/nih-callback?token=eyJhbGciOiJIUzI1NiJ9.bWJlbWlz.9ps0wCd2BwT3QD1xHLbJdrBEdMWXz4TvjtDMlcguwaY
* Error: http://localhost:4200/nih-callback?token=bad-token
* Error: http://localhost:4200/nih-callback?token=
* Error: http://localhost:4200/nih-callback

The success case is an encoded string that works for the dev environment. It isn't sensitive as it corresponds to dummy data.

## Success Case
In the event that linking succeeds, the user first sees a spinner and is then redirected back to the home page with a new link status saved to the background. Note `linkExpireTime` and `linkNihUsername` in the profile response.
screen shot 2019-01-31 at 8 20 21 am

## Error Cases
We can distinguish two error cases, no token provided or error saving provided token. **Please note** that these screenshots are not the final views. We don't have the UI to make the call out to link to NIH yet, so once that UI is in place, the current error screens can be re-purposed and displayed within the NIH linking area on the page. See [RW-1697](https://precisionmedicineinitiative.atlassian.net/browse/RW-1697) for that.

### No Token
screen shot 2019-01-31 at 8 31 05 am

### Error Saving Token
screen shot 2019-01-31 at 8 31 59 am


api/src/main/resources/workbench.yaml
```diff
@@ -294,6 +294,28 @@ paths:
...
+        - application/json
+      description: Updates a users NIH token
+      operationId: updateNihToken
+      parameters:
+        - in: body
+          name: token
+          description: the token retrieved from NIH
+          schema:
+            $ref: "#/definitions/NihToken"
+      responses:
```
calbach: Should this return the updated profile? Or do we not expect strong consistency on the profile being updated?

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+    super(props);
+    this.state = {
+      error: false,
+      errorMessage: ''
+    };
+  }
+
+  async componentDidMount() {
+    // Assumes callback url has format of `/nih-callback?token=XYZ`
+    const token = window.location.search.replace('?token=', '');
```
calbach: This is brittle as it assumes no other query parameters. Should use a library to extract the query params instead, e.g.

```
(new URL(window.location)).searchParams.get('token');
```

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+        this.navigateHome();
+      } catch (e) {
+        this.setErrorStatus('Error saving NIH Authentication status');
+      }
+    } else {
+      this.setErrorStatus('An NIH Authentication token is required.');
+    }
+  }
+
+  setErrorStatus(message: string) {
```
calbach: nit: could these two methods be private?

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+    this.setState({error: true, errorMessage: message});
+  }
+
+  navigateHome() {
+    navigateByUrl('/');
+  }
+
+  render() {
+    const error = <Error>Error Linking NIH Username: {this.state.errorMessage}
+      <div onClick={this.navigateHome} style={{cursor: 'pointer'}}>Please try linking again.</div>
```
calbach: nit: this should probably be styled like a link. I'd suggest making it an `<a>` (that goes nowhere). Will have some routing discussion later today about how we can fix these links in the future.

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+    </Error>;
+
+    return (this.state.error ? error : spinner);
+  }
+
+}
+
+/**
+ * The general flow for this component is:
```
calbach: Thanks - nice summary.

api/src/main/resources/workbench.yaml
rushtong: There's no reason **not** to return the full profile, but I was not envisioning that being a necessity either, i.e. it isn't necessary for the usage in this PR. Would be happy to make that a return if we think it will be helpful in the future.

ui/src/app/views/nih-callback/component.tsx
rushtong: Got it - will do.

ui/src/app/views/nih-callback/component.tsx
rushtong: I was getting weird lint errors when I tried it - will dig into it a little more.

ui/src/app/views/nih-callback/component.tsx
rushtong: Great point - will do.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -669,4 +671,19 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
     );
     return getIdVerificationsForReview();
   }
+
+  @Override
+  public ResponseEntity<Void> updateNihToken(NihToken token) {
+    if (token == null || token.getJwt() == null) {
```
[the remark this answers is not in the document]
rushtong: I think that might be true due to the generated js client. I was trying to be thorough on the backend and trying not to assume front validation would be sufficient.

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+  navigateHome() {
+    navigateByUrl('/');
+  }
+
+  render() {
+    const error = <Error>Error Linking NIH Username: {this.state.errorMessage}
+      <div onClick={this.navigateHome} style={{cursor: 'pointer'}}>Please try linking again.</div>
+    </Error>;
+
+    return (this.state.error ? error : spinner);
```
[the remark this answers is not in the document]
rushtong: Great points - will do.

ui/src/app/views/nih-callback/component.tsx
calbach: OK - don't worry about it then. I'm curious what errors you saw though.

api/src/main/resources/workbench.yaml
calbach: Sounds fine.

ui/src/app/views/nih-callback/component.tsx
rushtong: My bad - the errors aren't with private, they were with static. Will do.

ui/src/app/views/nih-callback/component.tsx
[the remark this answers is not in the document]
rushtong: Done

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+  error: boolean;
+  errorMessage: string;
+}
+
+export class NihCallback extends React.Component<{}, NihCallbackState> {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      error: false,
```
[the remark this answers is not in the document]
rushtong: Done

ui/src/app/views/nih-callback/component.tsx
rushtong: Done

ui/src/app/views/nih-callback/component.tsx
rushtong: Done

api/src/main/resources/workbench.yaml
rushtong: Added

ui/src/app/views/nih-callback/component.tsx
rushtong: Done

rushtong:
@calbach @dmohs @NehaBroad - PTAL, PR is updated with feedback addressed. I added some more content to the PR description as well. Thanks all for the great feedback - really good stuff.

rushtong:
Looks like I broke a unit test ... will try to get to that today.

ui/src/app/views/nih-callback/component.tsx
```diff
@@ -0,0 +1,91 @@
...
+    super(props);
+    this.state = {
+      errorMessage: ''
+    };
+  }
+
+  async componentDidMount() {
+    // Assumes callback url has format of `/nih-callback?token=XYZ`
+    const token = (new URL(window.location.href)).searchParams.get('token');
+    if (token && token.length > 1) {
```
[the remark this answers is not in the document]
rushtong: It should be sufficient ... left-over logic from the previous version. Will double check and update if necessary.

ui/src/testing/stubs/profile-api-stub.ts
```diff
@@ -1,10 +1,33 @@
...
+      dataAccessLevel: DataAccessLevel.Registered,
+      givenName: 'Tester!@#$%^&*()><script>alert("hello");</script>',
+      familyName: 'MacTesterson!@#$%^&*()><script>alert("hello");</script>',
+      phoneNumber: '999-999-9999',
+      pageVisits: [{page: 'test'}],
+    };
   }
+
+  public updateNihToken(token?: NihToken, options?: any) {
+    return new Promise<Profile>(resolve => { resolve(this.profile); });
```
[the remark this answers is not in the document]
rushtong: Thank you - I was letting the IDE help me out with the syntax here.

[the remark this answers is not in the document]
rushtong: Ick ... weird TS compilation error here. Will punt on this for now. 
```
TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/testing/stubs/profile-api-stub.ts:30:12 - error TS2350: Only a void function can be called with the 'new' keyword.

    30     return new Promise.resolve(this.profile);
```

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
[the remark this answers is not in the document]
rushtong: Thank you, and I tend to agree here. The main reason I have used `required` is not for code generation, but for the UI that swagger-ui generates, which we don't actually do in AoU.

ui/src/testing/stubs/profile-api-stub.ts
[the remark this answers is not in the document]
calbach: Drop the `new` - `Promise.resolve` is a static helper function.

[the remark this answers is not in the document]
rushtong: Ah ... thank you!

PR #1711, opened 2019-01-28
[RW-1987][risk=no] Autocomplete search - show only first matching node in tree

dolbeew:
When selecting an option from the autocomplete dropdown, expand and scroll only to the first match instead of expanding all duplicate nodes as well


ui/src/app/cohort-search/node/node.component.ts
```diff
@@ -128,12 +125,16 @@ export class NodeComponent implements OnInit, OnDestroy {
       const subtreeSelectSub = this.selected$
         .filter(selectedIds => !!selectedIds)
         .subscribe(selectedIds => {
-          this.subMatches = selectedIds.length;
           if (parentId !== 0) {
+            if (selectedIds.includes(parentId)) {
+            }
```
smazumd: Looks Good.. remove this if statement if we don't need it.

PR #1712, opened 2019-01-28
[risk=no] add spinners

panentheos:
Adds a standard spinner overlay component. Can be used to provide feedback that an action is in progress, as well as block the underlying UI.

Typical usage is to set `position: relative` on the container you want to overlay, then conditionally render the `SpinnerOverlay` inside that container (at the end, to make sure it's on top of other elements).

See the included example (note that modals already `position: relative` on them)

### DIFF
--- a/ui/src/app/components/spinners.tsx
+++ b/ui/src/app/components/spinners.tsx
@@ -1,0 +1,29 @@
+import {reactStyles} from 'app/utils/index';
+import * as React from 'react';
+
+const styles = reactStyles({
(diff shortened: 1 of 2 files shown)


rushtong:
Also a 👍 for me - I actually need this for my current nih PR so am hoping this merges soon.

panentheos:
There isn't a way to change the size for now, but we could certainly add one if it's needed. We should probably do a check-in with design regarding the visuals anyway.

PR #1713, opened 2019-01-29
[RW-1943][risk=moderate] Add an interactive shell command for mysqlbinlog usage

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-1943

Corresponding playbook entry:
https://docs.google.com/document/d/1hbYNWzdxuRybTfW3M70hWZfr_TXRVd5rOMR9kZtwAFY/edit#heading=h.1bl49nkg5ekf

Example usage:

```
./project.rb connect-to-cloud-db-binlog --project all-of-us-workbench-test
...
********************************************************************************
Listing available journal files: 
+ echo 'SHOW BINARY LOGS;' | mysql --host=127.0.0.1 --port=3307 --user=root --database=workbench --password=********************
2019/01/29 00:37:35 New connection for "all-of-us-workbench-test:us-central1:workbenchmaindb"
Log_name	File_size
mysql-bin.002951	104857882
mysql-bin.002952	1076501425
mysql-bin.002953	104857740
mysql-bin.002954	45435992
********************************************************************************

********************************************************************************
mysql login has been configured. Pick a journal file from above and run commands like: 
  mysqlbinlog -R mysql-bin.xxxxxx

See the Workbench playbook for more details.
********************************************************************************
+ export MYSQL_HOME=$(with-mysql-login.sh root ********************); cd /tmp; /bin/bash
bash-4.4$
```

`mysqlbinlog` can then be repeatedly executed within the shell as follows:

```
bash-4.4$ mysqlbinlog -R -s mysql-bin.002951 > out.txt
```

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1204,6 +1204,47 @@ Common.register_command({
   :fn => ->(*args) { connect_to_cloud_db("connect-to-cloud-db", *args) }
 })
 
+def connect_to_cloud_db_binlog(cmd_name, *args)
+  ensure_docker cmd_name, args
+  common = Common.new
+  op = WbOptionsParser.new(cmd_name, args)
+  gcc = GcloudContextV2.new(op)
+  op.parse.validate
+  gcc.validate
+  env = read_db_vars(gcc)
+  CloudSqlProxyContext.new(gcc.project).run do
+    common.status "\n" + "*" * 80
+    common.status "Listing available journal files: "
+
+    # "root" is required for binlog access.
+    password = env["MYSQL_ROOT_PASSWORD"]
+    run_with_redirects(
+      "echo 'SHOW BINARY LOGS;' | " +
+      "mysql --host=127.0.0.1 --port=3307 --user=root " +
+      "--database=#{env['DB_NAME']} --password=#{password}", password)
+    common.status "*" * 80
+
+    common.status "\n" + "*" * 80
+    common.status "mysql login has been configured. Pick a journal file from " +
+                  "above and run commands like: "
+    common.status "  mysqlbinlog -R mysql-bin.xxxxxx\n"
+    common.status "See the Workbench playbook for more details."
+    common.status "*" * 80
+
+    # Work out of /tmp for easy local file redirection. We don't want binlogs
+    # winding up back in Workbench source control accidentally.
+    run_with_redirects(
+      "export MYSQL_HOME=$(with-mysql-login.sh root #{password}); " +
+      "cd /tmp; /bin/bash", password)
+  end
+end
+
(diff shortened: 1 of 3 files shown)


rushtong:
From a code perspective, this all looks reasonable. 👍 I ran into trouble trying to use it though.

The command `./project.rb connect-to-cloud-db-binlog --project all-of-us-workbench-test` works great, and I can see the various logs listed. However, running `mysqlbinlog` in the shell consistently fails
```
bash: mysqlbinlog: command not found
```
I see the new addition to the dockerfile, `mysql` - I did try a `./project.rb docker-clean` but have now gotten my dev env into a very broken state with `Failure: 'getpwuid(): uid not found: 1992018429'`. Will dig into this when I can get back to a working state.

api/libproject/devstart.rb
```diff
@@ -1204,6 +1204,47 @@ def connect_to_cloud_db(cmd_name, *args)
...
+  gcc = GcloudContextV2.new(op)
+  op.parse.validate
+  gcc.validate
+  env = read_db_vars(gcc)
+  CloudSqlProxyContext.new(gcc.project).run do
+    common.status "\n" + "*" * 80
+    common.status "Listing available journal files: "
+
+    # "root" is required for binlog access.
+    password = env["MYSQL_ROOT_PASSWORD"]
```
gjuggler: Is the storage & use of mysql passwords documented anywhere? L1244 doesn't mention it, nor does the runbook. 

Looking around, it's scattered all throughout this file, so I suppose devs are familiar with it... but the runbook should probably have some pointer on where / how to find the root password. Do we have some standard place to store team secrets like this?

calbach: w.r.t. these tools, it's an implementation detail. The credentials themselves are stored in GCS. Most of these tools read that file dynamically to pull necessary credential information automatically (this happens in `read_db_vars()` above).

There's no expected prerequisite of environment variables etc for running these commands, which is probably what this code looks like.

gjuggler: Thanks – my mistake for not reading carefully, for some reason I'd thought 'env' was the shell environment, not a local variable :) Makes total sense.

rushtong:
@calbach - I don't think I hit a specific gsutil issue. I ended up rebuilding the image and restarting docker. Those things got me past the failures I did hit.

calbach:
Ah - I omitted some information. I found that in this context `Failure: 'getpwuid(): uid not found: 1992018429'` originates from gsutil (only specific versions have this issue).

calbach:
More information: https://github.com/GoogleCloudPlatform/gsutil/issues/575

rushtong:
Thank you - that one was a mystery.

PR #1715, opened 2019-01-29
[RW-1903][Risk=low] Workspace Share conversion - READY FOR REVIEW

UrsaStutsman:
Almost ready, think I just have a couple more tests to change...
Since the last set of reviews changes:
1) Added React Select component as the standard select was presenting lots of issues
2) Added more tests
3) Changed Styling (worked with @lstrano-broad on this) - Will be adding screenshots of this below


ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,412 @@
...
+      });
+  }
+
+  removeCollaborator(user: UserRole): void {
+    if (!this.state.usersLoading) {
+      this.setState({usersLoading: true});
+      const position = this.state.userRolesList.findIndex((userRole) => {
+        return user.email === userRole.email;
+      });
+      this.state.userRolesList.splice(position, 1);
```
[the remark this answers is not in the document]
panentheos: Correct, it's strongly preferred to avoid mutation. There are several non-mutating array methods and lodash functions available for cases like these. In this case, looking up at the calling context, since we know the index of the element we clicked on, this removal could be done as `fp.pullAt(i, userRolesList)`

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,412 @@
...
+    this.setState({userRolesList: this.state.userRolesList});
+  }
+
+  searchTermChangedEvent($event: string) {
+    this.userSearch($event);
+  }
+
+  userSearch(value: string): void {
+    this.setState({autocompleteLoading: true, autocompleteUsers: [], searchTerm: value});
+    if (!this.state.searchTerm.trim()) {
```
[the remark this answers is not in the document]
panentheos: Correct. In this case, the simplest change would be to just use `value` to perform the search, since that's the latest value.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,412 @@
...
+  save(): void {
+    if (this.state.usersLoading) {
+      return;
+    }
+    this.setState({usersLoading: true, workspaceShareError: false});
+    workspacesApi().shareWorkspace(this.state.workspace.namespace,
+      this.state.workspace.id,
+      {workspaceEtag: this.state.workspace.etag, items: this.state.workspace.userRoles})
+      .then((resp: ShareWorkspaceResponse) => {
+        this.setState({usersLoading: false, toShare: '', searchTerm: ''});
```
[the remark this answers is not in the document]
panentheos: Agreed. In general, it's a good idea to collapse `setState` calls when possible. You can always use conditional logic in the values themselves, for example as a way to make a conditional change:
```
this.setState(({ bar }) => ({
  foo: 'a',
  bar: updateBar ? newBar : bar
}))
```

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+  tooltipPoint: {
+    listStylePosition: 'outside',
+    marginLeft: '0.6rem'
+  },
+
+  collabUser: {
+    fontSize: '.66667rem',
+    marginTop: '1rem',
+    letterSpacing: '.01em',
```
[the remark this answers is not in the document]
UrsaStutsman: I based this on the previous styling and 'em' was used (em is a unit).

ui/src/app/views/workspace-share/component.spec.tsx
```diff
@@ -0,0 +1,4 @@
+// Displays correct information (setup a default ACL)
```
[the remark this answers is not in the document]
UrsaStutsman: Yup writing those now

ui/src/app/views/workspace-share/component.tsx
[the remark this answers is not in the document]
jacmrob: gotcha

ui/src/app/views/workspace/component.ts
```diff
@@ -84,6 +84,7 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
   @ViewChild(ToolTipComponent) toolTip: ToolTipComponent;
   @ViewChild(WorkspaceShareComponent)
   shareModal: WorkspaceShareComponent;
```
calbach: Can this reference be removed now?

ui/src/app/views/workspace-list/component.spec.ts
```diff
@@ -151,8 +151,11 @@ describe('WorkspaceListComponent', () => {
     simulateClick(workspaceListPage.fixture, de.query(By.css('.delete-item')));
     updateAndTick(workspaceListPage.fixture);
     expect(workspaceListPage.fixture.componentInstance.confirmDeleting).toBeTruthy();
-    expect(workspaceListPage.fixture.componentInstance.resource)
-      .toEqual(firstWorkspace);
+    // TODO: 1/29/2019 review these expects, is this where we want to test this?
+    // regardless breaking during React conversion, so taking out for now
```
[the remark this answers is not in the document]
UrsaStutsman: Yes, the test really broke.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,410 @@
...
+    if (this.state.usersLoading) {
+      return;
+    }
+    this.setState({usersLoading: true, workspaceShareError: false});
+    workspacesApi().shareWorkspace(this.state.workspace.namespace,
+      this.state.workspace.id,
+      {workspaceEtag: this.state.workspace.etag, items: this.state.workspace.userRoles})
+      .then((resp: ShareWorkspaceResponse) => {
+        const updatedWorkspace = this.state.workspace;
+        fp.set('etag', resp.workspaceEtag, updatedWorkspace);
```
calbach: `fp.set` does not have any side-effects so you have to use the results - these two statements are currently no-ops.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,409 @@
...
+        } else {
+          this.setState({workspaceShareError: true});
+        }
+        this.setState({usersLoading: false});
+      });
+  }
+
+  removeCollaborator(user: UserRole): void {
+    if (!this.state.usersLoading) {
+      const position = this.state.userRolesList.findIndex((userRole) => {
```
calbach: This should be happening inside the updater function below; userRolesList might have changed between here and the callback below.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,409 @@
...
+        return user.email === userRole.email;
+      });
+      this.setState(({userRolesList}) => (
+        {userRolesList: fp.pullAt(position, userRolesList)}));
```
calbach: `fp.remove(({email}) => user.email === email, userRolesList)` is probably easier than the two step thing

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,409 @@
...
+
+  userSearch(value: string): void {
+    this.setState({autocompleteLoading: true, autocompleteUsers: [], searchTerm: value});
+    if (!value.trim()) {
+      this.setState({autocompleteLoading: false});
+      return;
+    }
+    userApi().user(this.state.searchTerm)
+      .then((response) => {
+        this.setState({autocompleteLoading: false});
```
calbach: Should just include this in the `setState()` below

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,409 @@
...
+        response.users = fp.differenceWith((a, b) => {
+          return (a.email === b.email && a.givenName === b.givenName
```
calbach: Existing issue, but email should really be sufficient here.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,409 @@
...
+  userSearch(value: string): void {
+    this.setState({autocompleteLoading: true, autocompleteUsers: [], searchTerm: value});
+    if (!value.trim()) {
+      this.setState({autocompleteLoading: false});
+      return;
+    }
+    userApi().user(this.state.searchTerm)
+      .then((response) => {
+        this.setState({autocompleteLoading: false});
+        response.users = fp.differenceWith((a, b) => {
```
calbach: `differenceWith` is probably not correct here as it's going to pull in elements from the user list as well. I'd throw the `userRolesList` emails into a `Set` then filter against that.

UrsaStutsman: I'm a little confused here. Isn't the behaviour we want is that the autocomplete returns users matching search that are NOT already in the userRolesList? and isn't that what differenceWith does? Feel free to DM me if I'm around.

ui/package.json
```diff
@@ -89,6 +89,7 @@
     "highcharts-react-official": "^2.0.0",
     "immutable": "^4.0.0-rc.9",
     "lodash": "^4.17.11",
+    "lolex": "^3.0.0",
```
calbach: If this is only used in tests, it can go into devDependencies instead, I believe.

ui/src/testing/stubs/user-api-stub.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  user(searchTerm: string): Promise<UserResponse> {
+    return new Promise<UserResponse>(resolve => {
+      let usersToReturn: User[] = [];
+      if (this.existingUsers) {
+        usersToReturn = this.existingUsers.filter((userRole) => {
+          return fp.includes(searchTerm, fp.values(userRole).join(' '));
+        });
+      } else {
+        usersToReturn.push(<User>{
```
calbach: nit: `as` syntax is preferred in typescript

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,405 @@
...
+    };
+    this.searchTermChangedEvent = fp.debounce(300, this.userSearch);
+  }
+
+  save(): void {
+    if (this.state.usersLoading) {
+      return;
+    }
+    this.setState({usersLoading: true, workspaceShareError: false});
+    workspacesApi().shareWorkspace(this.state.workspace.namespace,
```
calbach: The stub doesn't have an implementation for `shraeWorkspace()` - so how are the tests working? Am I missing something?

ui/src/app/views/workspace-share/component.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {UserApiStub} from 'testing/stubs/user-api-stub'
+
+describe('WorkspaceShareComponent', () => {
+  let props: WorkspaceShareProps;
+
+  const component = () => {
+    return mount<WorkspaceShare, WorkspaceShareProps, WorkspaceShareState>
+    (<WorkspaceShare {...props}/>);
+  };
+  const harry: User = {givenName: 'Harry', familyName: 'Potter', email: 'harry.potter@hogwarts.edu'};
```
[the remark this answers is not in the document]
UrsaStutsman: :)

ui/src/app/views/workspace-share/component.tsx
UrsaStutsman: Oooh. Good point, I actually test for all the functionality of the modal (reactive to user additions/removals, etc) but not the save function! I'll add that now.

UrsaStutsman: tests added

UrsaStutsman:
Upon first opening modal, note owner cannot change their own role, or delete themselves:
screen shot 2019-02-15 at 12 51 18 pm

UrsaStutsman:
When searching:
screen shot 2019-02-15 at 12 49 31 pm

UrsaStutsman:
When changing role: 
screen shot 2019-02-15 at 12 50 03 pm

calbach:
Sorry - I didn't realize this was ready for re-review. Apparently PR title updates do not result in an email notification. Will take a look ASAP.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+        } else {
+          this.setState({workspaceShareError: true});
+        }
+        this.setState({usersLoading: false});
+      });
+  }
+
+  removeCollaborator(user: UserRole): void {
+    this.setState(({userRolesList, workspace}) => (
+      {userRolesList: fp.remove(({email}) => user.email === email, userRolesList),
```
calbach: nit: My eyes cannot follow this indentation. I think something like this would be better:

```
... => ({
  userRolesList: ...
  workspace: {
   ...
```

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+          this.setState({workspaceFound: false});
+        }
+      });
+  }
+
+  addCollaborator(user: User): void {
+    const userRole: UserRole = {givenName: user.givenName, familyName: user.familyName,
+      email: user.email, role: WorkspaceAccessLevel.READER};
+    this.setState(({userRolesList, workspace}) => (
+      {searchTerm: '', autocompleteLoading: false, autocompleteUsers: [], dropDown: false,
```
calbach: nit: sort of the same readability here, I feel like putting this bracket on the previous line is nicer

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+  handleClickOutsideSearch = (e) => {
+    if (this.searchingNode) {
+      if (this.searchingNode.contains(e.target)) {
+        return;
+      }
+      this.closeDropdown();
+    }
+  }
+
+  cleanClassNameForSelect(value: string): string {
```
calbach: This is just for testing right? What happens if this produces a bad value? Does it result in an error? I'm not sure this regex is bullet-proof for all possible email characters so I want to understand what risk there is here.

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -50,4 +114,48 @@ export class WorkspacesApiStub extends WorkspacesApi {
     });
   }
 
+
+  getWorkspaces(options?: any): Promise<WorkspaceResponseListResponse> {
+    return new Promise<WorkspaceResponseListResponse>(resolve => {
```
calbach: nit: Probably more readable to just get the value then do a `return Promise.resolve(value)`.

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -1,14 +1,35 @@
...
 export class WorkspacesApiStub extends WorkspacesApi {
+  workspaces: Workspace[];
+  workspaceAccess: Map<string, WorkspaceAccessLevel>;
   notebookList: FileDetail[];
 
-  constructor() {
+  constructor(workspaces?: Workspace[]) {
     super(undefined, undefined, (..._: any[]) => { throw Error('cannot fetch in tests'); });
-
+    this.workspaces = fp.defaults([WorkspacesApiStub.stubWorkspace()], workspaces);
```
calbach: I don't think `defaults` is the right fit here - looks like it's intended purpose is to fill in missing property values; I'm not sure what the expected behavior in this case of arrays is, but `[]` should be a legal input and it's handling that incorrectly here.

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -50,4 +114,48 @@ export class WorkspacesApiStub extends WorkspacesApi {
...
+          };
+        })});
+    });
+  }
+
+  shareWorkspace(workspaceNamespace: string, workspaceId: string,
+    body?: ShareWorkspaceRequest, options?: any): Promise<ShareWorkspaceResponse> {
+    return new Promise<ShareWorkspaceResponse>(resolve => {
+      const newEtag = fp.defaults(2, (body.workspaceEtag + 1));
+      const newItems = fp.defaults({}, body.items);
```
calbach: `{}` --> `[]`

ui/src/testing/stubs/user-api-stub.ts
```diff
@@ -0,0 +1,33 @@
+import {User, UserApi, UserResponse, UserRole} from 'generated/fetch';
+
+import * as fp from 'lodash/fp';
+
+export class UserApiStub extends UserApi {
+  existingUsers: UserRole[];
+  constructor(existingUsers?: UserRole[]) {
+    super(undefined, undefined, (..._: any[]) => { throw Error('cannot fetch in tests'); });
+    if (existingUsers) { this.existingUsers = existingUsers; }
```
calbach: opt_nit: We don't have an official rule on this, but I would either do a one line without a brackets, or multiline with, but not a one-line with brackets.

ui/src/app/views/workspace/component.ts
```diff
@@ -225,7 +225,18 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
...
+
+  closeShare(): void {
+    this.sharing = false;
+    // only done until this page refactored ot React
+    window.location.reload();
+  }
+
+  updateAclList(userRoleList: UserRole[]): void {
+    this.workspace.userRoles = userRoleList;
+    console.log(this.workspace);
```
calbach: rm

ui/src/app/views/workspace/component.ts
```diff
@@ -225,7 +225,18 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
   }
 
   share(): void {
-    this.shareModal.open();
+    this.sharing = true;
+  }
+
+  closeShare(): void {
+    this.sharing = false;
+    // only done until this page refactored ot React
```
calbach: Please reference the conversion ticket in this comment and put a TODO.

ui/src/app/views/workspace-share/component.spec.tsx
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  it('does not allow self role change', () => {
+    const wrapper = component();
+    expect(wrapper.find('[data-test-id="harry.potter@hogwarts.edu-user-role"]').first()
+      .props()['isDisabled']).toBe(true);
+  });
+
+  it('saves acl correctly after changes made', async() => {
+    let clock = Lolex.install({shouldAdvanceTime: true});
```
calbach: Mostly for my understanding: any reason not to put this in setup/teardown? Are there downsides to using this in a test?

ui/src/app/views/workspace-share/component.spec.tsx
```diff
@@ -0,0 +1,140 @@
...
+    // change hermione's access to owner
+    let selectComponent = wrapper.find('[data-test-id="hermione.granger@hogwarts.edu-user-role"]')
+      .find(Select);
+    selectComponent.instance().setState({menuIsOpen: true});
+    wrapper.update();
+    wrapper.find('.hermionegrangerhogwartsedu-user-role__option')
+      .findWhere(n => n.text() === fp.capitalize(WorkspaceAccessLevel[WorkspaceAccessLevel.OWNER]))
+      .simulate('click');
+    wrapper.update();
+    const spy = jest.spyOn(workspacesApi(), 'shareWorkspace');
```
calbach: nit: it would be better to just check the UI after, rather than spying on this method, since you went through the effort of implementing the stub as well. Or does that part not work (this is the reload thing?)

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -50,4 +114,48 @@ export class WorkspacesApiStub extends WorkspacesApi {
...
+            workspace: {...workspace},
+            accessLevel: accessLevel
+          };
+        })});
+    });
+  }
+
+  shareWorkspace(workspaceNamespace: string, workspaceId: string,
+    body?: ShareWorkspaceRequest, options?: any): Promise<ShareWorkspaceResponse> {
+    return new Promise<ShareWorkspaceResponse>(resolve => {
```
calbach: opt: actually set these roles on the workspace

ui/src/app/views/workspace-share/component.spec.tsx
UrsaStutsman: Not specifically that I am aware of. This is just the only test where it's needed so I thought keeping it  here would make troubleshooting issues later easier.

ui/src/app/views/workspace-share/component.spec.tsx
UrsaStutsman: The reload thing definitely an issue on the workspace page. And on the workspace list page there would be nothing to check on the UI (since the page doesn't display workspaces ACLs). So I thought best way to test was to make sure this method got called with the correct request.

ui/src/app/views/workspace-share/component.tsx
UrsaStutsman: So not really sure what you mean by "bad value". This is just for testing. So the rest of the time we don't really care what it produces (as long as it doesn't actually break anything). I replaced it with bunch of random things, and it doesn't cause any problems as far as I can tell (tested with values like "k* 0").

ui/src/testing/stubs/workspaces-api-stub.ts
UrsaStutsman: Hmmm, defaultTo then? I think that may be actually what I meant to use.

ui/src/app/views/workspace-share/component.tsx
calbach: Thanks - my concern was that it would cause some kind of rendering issue or bug in the application. If not - then I'm not really worried about it.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+    return (value + '-user-role').replace(/[@\.]/g, '');
+  }
+
+  get hasPermission(): boolean {
+    return this.props.accessLevel === WorkspaceAccessLevel.OWNER;
```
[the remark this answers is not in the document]
UrsaStutsman: this enum does not have that value. WorkspaceAccessLevel enum has these values: Owner, Writer, Reader, Noaccess

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+  {value: WorkspaceAccessLevel.WRITER, label: 'Writer'},
+  {value: WorkspaceAccessLevel.OWNER, label: 'Owner'}
+];
+
+export interface WorkspaceShareState {
+  autocompleteLoading: boolean;
+  autocompleteUsers: User[];
+  userNotFound: string;
+  workspaceShareError: boolean;
+  usersLoading: boolean;
```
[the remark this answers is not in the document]
UrsaStutsman: nope this is specifically for the user search loading.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+      workspaceUpdateConflictError: false,
+      workspace: this.props.workspace,
+      searchTerm: '',
+      dropDown: false,
+    };
+    this.searchTermChangedEvent = fp.debounce(300, this.userSearch);
+  }
+
+  componentWillMount(): void {
+    document.addEventListener('mousedown', this.handleClickOutsideSearch, false);
```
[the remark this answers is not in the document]
UrsaStutsman: Think I may need a bit of direction on this change

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+        </ModalBody>
+        <ModalFooter>
+            <Button type='secondary' style={{marginRight: '.8rem', border: 'none'}}
+                    onClick={() => this.props.closeFunction()}>Cancel</Button>
+            <Button data-test-id='save' disabled={!this.hasPermission}
+                    onClick={() => this.save()}>Save</Button>
+        </ModalFooter>
+      </Modal>}
+      {!this.state.workspaceFound && <div>
+        <h3>The workspace {this.state.workspace.id} could not be found</h3></div>}
```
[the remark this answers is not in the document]
UrsaStutsman: You know what. I don't know. I can make a follow up ticket for addressing this.

ui/src/app/views/workspace-share/component.tsx
```diff
@@ -0,0 +1,508 @@
...
+      });
+  }
+
+  removeCollaborator(user: UserRole): void {
+    this.setState(({userRolesList, workspace}) => ({
+      userRolesList: fp.remove(({email}) => user.email === email, userRolesList),
+      workspace: {...workspace,
+        userRoles: fp.remove(({email}) => {
+          return user.email === email;
+        }, workspace.userRoles)} as Workspace}
```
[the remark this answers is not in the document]
UrsaStutsman: Resolved: removed userRolesList

ui/src/app/views/workspace-share/component.tsx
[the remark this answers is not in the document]
UrsaStutsman: Added ticket RW-2192 to address this and above comment around error handling.

PR #1716, opened 2019-01-29
[RW-1906][risk=no] Resource Card react

jacmrob:


jacmrob:
Still working on finishing tests but this is a big PR so thought I'd get a head start on reviewing.

ui/src/app/utils/resourceActionsReact.ts
```diff
@@ -0,0 +1,52 @@
+import {
```
rushtong: nit on naming. I think this could be `resource-actions` and possibly be a `.tsx` file too? That should avoid the naming conflict with `resourceActions.ts`

ui/src/app/views/cohort-list/component.spec.ts
```diff
@@ -103,6 +105,7 @@ describe('CohortListComponent', () => {
       ] }).compileComponents().then(() => {
         cohortListPage = new CohortListPage(TestBed);
       });
+    registerApiClient(CohortsApi, new CohortsApiStub());
```
rushtong: What I have found is that you will need to stub out a mock response for any api calls your component makes. If that isn't in place, then the stub will throw an error. See what I did for the profile api here: https://github.com/all-of-us/workbench/pull/1710/files#diff-8a1055dcdba0f29945195b2d731972f8

ui/src/app/views/resource-card/component.tsx
```diff
@@ -113,175 +113,313 @@ export class ResourceCardMenuComponent extends ReactWrapperBase {
...
+        };
+      } else if (props.resourceCard.conceptSet) {
+        // TODO [1/29/19]: Need to thread generated/fetch model types through for API calls
+        //  once parent components (notebook-list, cohort-list, concept-set-list) are converted
+        //  this should go away and resourceCard can be used as what comes through the props
+        const myTempConceptSet = {...this.props.resourceCard.conceptSet,
+          domain: props.resourceCard.conceptSet.domain as Domain};
+        this.state = {
+          resourceType: ResourceType.CONCEPT_SET,
+          resourceCard: {...props.resourceCard, conceptSet: myTempConceptSet},
```
panentheos: Both `resourceType` and `resourceCard` can be computed directly from props, and therefore don't need to be in state. (Even the temporary hack, which could just be factored out into e.g. a function if needed)

ui/src/app/views/cohort-list/component.spec.ts
jacmrob: I figured it out, thanks!!

ui/src/app/views/resource-card/component.tsx
jacmrob: even with this hack to cast the types in the case of a `ConceptSet`?  could you explain more what you mean?

ui/src/app/utils/resourceActionsReact.ts
jacmrob: it actually can't be a `tsx` file because of the syntax so I think we should just leave it until conversion is finished and then delete `resourceActions.ts` and rename it

ui/src/app/views/resource-card/component.tsx
jacmrob: nevermind actually figured it out!  will push up soon

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -1,13 +1,59 @@
...
+    notebookName: String): Promise<any> {
+    return new Promise<any>(resolve => {
+      setTimeout(() => {
+        const cloneName = notebookName.replace('.ipynb', '') + ' Clone.ipynb';
+        this.notebookList.push({
+          'name': cloneName,
+          'path': 'gs://bucket/notebooks/' + cloneName,
+          'lastModifiedTime': 100
+        });
+        resolve({});
```
panentheos: Do we really intend to implement a whole fake backend for the sake of tests? What's actually being verified here?

ui/src/app/views/resource-card/component.tsx
```diff
@@ -113,175 +113,310 @@ export class ResourceCardMenuComponent extends ReactWrapperBase {
...
+  // TODO [1/31/19] This method is only necessary until the parent components
+  //    (notebook-list, cohort-list, conceptSet-list) have been converted and use the
+  //    fetch API models.
+  static castConceptSet(resourceCard: RecentResource): RecentResource {
+    if (resourceCard.conceptSet) {
+      const myTempConceptSet = {...resourceCard.conceptSet,
+        domain: resourceCard.conceptSet.domain as Domain};
+      return {...resourceCard, conceptSet: myTempConceptSet};
+    }
+    return resourceCard;
```
panentheos: Good use of a pure function to make this transformation.

ui/src/testing/stubs/cohorts-api-stub.ts
```diff
@@ -0,0 +1,142 @@
...
+  Workspace,
+  WorkspaceAccessLevel
+} from 'generated/fetch';
+import {CohortListResponse} from 'generated/fetch/api';
+import {WorkspaceStubVariables} from './workspace-service-stub';
+
+export let DEFAULT_COHORT_ID = 1;
+export let DEFAULT_COHORT_ID_2 = 2;
+
+class CohortStub implements Cohort {
```
calbach: Why is this class needed? Can you just directly instantiate an object as needed which matches the interface type?

ui/src/testing/stubs/cohorts-api-stub.ts
```diff
@@ -0,0 +1,142 @@
...
+    };
+    this.cohorts = [exampleCohort, exampleCohortTwo];
+    this.workspaces = [stubWorkspace];
+    this.resourceList = convertToResources(this.cohorts, stubWorkspace.namespace,
+      stubWorkspace.id, WorkspaceAccessLevel.OWNER, ResourceType.COHORT);
+  }
+
+  updateCohort(ns: string, wsid: string, cid: number, newCohort: Cohort): Promise<Cohort> {
+    return new Promise<Cohort>((resolve, reject) => {
+      setTimeout(() => {
```
calbach: You can remove all of these `setTimeout()`s in the stub now. This was a workaround to Observable's inconsistent behavior around callback resolution. Namely if a value is present on an observable, a callbacks are executed synchronously. So without the setTimeout, the following would have worked in tests, but failed in a real browser:

```
let foo;
cohortsApi.updateCohort.subscribe(resp => {
  foo = resp;
});
console.log(foo.items);
```

Promises always execute callbacks asynchronously, so it's not a problem here - e.g. try this in Chrome console:

```
(() => {
    Promise.resolve(1).then(console.log);;
    console.log(2)
})();
2
1
```

ui/src/testing/stubs/concept-sets-api-stub.ts
```diff
@@ -0,0 +1,133 @@
...
+      setTimeout(() => {
+        const target = this.mustFindConceptSet(conceptSetId);
+        target.name = req.name;
+        target.description = req.description;
+        resolve(target);
+      }, 0);
+    });
+  }
+
+  // public createConceptSet(
```
calbach: nit: I prefer to avoid commented out code if possible. It should be easy enough to look at the service stub or git history once we need to port this.

ui/src/app/views/concept-set-list/component.ts
```diff
@@ -2,20 +2,27 @@ import {Component, OnInit, ViewChild} from '@angular/core';
...
+  ConceptSetsService,
+  Domain,
+  RecentResource,
+  Workspace,
+  WorkspaceAccessLevel,
 } from 'generated';
 
+import {generateDomain} from 'app/utils/index';
+
+// import {Domain} from 'generated/fetch';
```
calbach: rm?

ui/src/testing/stubs/workspaces-api-stub.ts
calbach: For behavioral tests (which are generally what I would favor in terms of value), having a single fake service implementation tends to be more maintainable and cleaner than setting up complicated multistage and implementation-dependent mock expectations.

e.g. test which does a notebook clone on the list page:

```
// mocks setup

mock(clone(...));
mock(listNotebooks(...)).return(fakePostCloneNotebooks)
// click clone
// verify UI shows cloned card
```

vs

```
// stub setup

// click clone
// check stub service or UI for cloned notebook
```

Note in the first case, that the actual details of what clone "does" have been simulated by the mocks (in each test case that needs it) - vs just implementing the logic in a fake. Using fakes also tends to scale a bit better in terms of readability in my experience especially when there are multiple services involved.

Anyways - this is my experience, but the value we see here will entirely depend on the types of tests we are writing.

ui/src/app/views/resource-card/component.tsx
```diff
@@ -113,175 +113,310 @@ export class ResourceCardMenuComponent extends ReactWrapperBase {
...
+          invalidResourceError: false,
+          ...defaultState
+        };
       } else {
-        this.resourceType = ResourceType.INVALID;
-        this.invalidResourceError = true;
+        this.state = {
+          invalidResourceError: true,
+          ...defaultState
+        };
```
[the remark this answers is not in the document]
jacmrob: I personally don't really see how an "invalid resource" could ever come about, since we control the creation of resource cards which stem from creating cohorts, concept sets, and notebooks only.  maybe @blrubenstein or @calbach have thoughts on the initial implementation?

ui/src/testing/stubs/cohorts-api-stub.ts
```diff
@@ -0,0 +1,142 @@
...
+    this.name = cohort.name;
+    this.type = cohort.type;
+    this.workspaceId = wsid;
+  }
+}
+
+export class CohortsApiStub extends CohortsApi {
+  public workspaces: Workspace[];
+  public cohorts: CohortStub[];
+  public resourceList: RecentResource[];
```
calbach: I don't this stub should have `workspaces` or a `resourceList`, they also aren't used anywhere AFAICT.

ui/src/app/views/resource-card/component.tsx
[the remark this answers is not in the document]
calbach: Yes, it indicates a programming error and as such we probably shouldn't be driving UI with it. My suggestion would be that this component treats "oneof notebook/cohort/conceptSet" as a precondition, and that the parent component does any necessary validation on the list (logs errors and filters out unsupported types).

I suspect this originated when we added "concept sets" to the recent resources and subsequently broke the resource cards in an unpleasant way. So I think validating this *somewhere* would be good.

ui/src/testing/stubs/cohorts-api-stub.ts
jacmrob: 👍

ui/src/testing/stubs/cohorts-api-stub.ts
calbach: Not addressed

ui/src/testing/stubs/cohorts-api-stub.ts
calbach: Not addressed

ui/src/app/views/resource-card/component.tsx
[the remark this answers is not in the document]
calbach: Not addressed (or maybe it was?). Please resolve these kinds of comment threads before merging ideally by replying

ui/src/app/views/resource-card/component.tsx
```diff
@@ -113,175 +113,310 @@ export class ResourceCardMenuComponent extends ReactWrapperBase {
...
+        + '/' + this.props.resourceCard.workspaceFirecloudName + '/cohorts/'
+        + this.props.resourceCard.cohort.id + '/review');
+  }
+
+  closeEditModal(): void {
+    this.setState({editing: false});
+  }
+
+  renameNotebook(): void {
+    this.setState({renaming: true});
```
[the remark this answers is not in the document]
calbach: Not addressed - @jacmrob maybe there was some offline discussion in this case and you decided not to do this, but please ack these kinds of comments if for no other reason than for the sake of remote observers

PR #1717, opened 2019-01-29
[RW-2015][RISK=LOW]User account creation checks wrong username

NehaBroad:


ui/src/app/views/account-creation/component.tsx
```diff
@@ -120,8 +120,8 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
   }
 
   usernameChanged(value: string): void {
-    const {username} = this.state.profile;
     this.updateProfile('username', value);
+    const {username} = this.state.profile;
```
calbach: Per offline discussion, this pattern is apparently not OK. See my comment here for further details/discussion. https://github.com/all-of-us/workbench/pull/1715#discussion_r252070271

Note that this worked in the test env when I tried it out, so it's probably not urgent - lets see the outcome of that other thread.

PR #1718, opened 2019-01-30
[RW-2016][risk=low] Add domain verification file for stable

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2016

### DIFF
--- a/ui/app.yaml
+++ b/ui/app.yaml
@@ -4,6 +4,11 @@ threadsafe: true
 default_expiration: "2m"
 
 handlers:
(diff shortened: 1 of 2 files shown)


calbach:
Yeah, since we don't run GAE locally it won't work - I deployed a test instance for this purpose: https://calbach-dot-all-of-us-workbench-test.appspot.com/.well-known/pki-validation/gsdv.txt

rushtong:
Thank you! Looks great 👍

PR #1723, opened 2019-01-30
[RW-1891][risk=no] Convert ws nav bar to react

zarsky-broad:
before:
before

after (matches Invision mocks & Terra):
after


ui/test-shim.js
```diff
@@ -1,10 +1 @@
 require('mutationobserver-shim');
-/**
```
zarsky-broad: Included automatically when the jest jsdom defaults set `pretendToBeVisual`.

ui/src/app/views/workspace-nav-bar/component.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+describe('WorkspaceNavBarComponent', () => {
+
+  let props: {};
+  const workspace = {
+    ...WorkspacesServiceStub.stubWorkspace(),
+    accessLevel: WorkspaceAccessLevel.OWNER,
+  };
+
+  const component = () => {
+    return mount(<WorkspaceNavBarReact {...props}/>, {attachTo: document.getElementById('root')});
```
zarsky-broad: This `attachTo` makes the PopupTrigger work.

ui/test-setup-script.js
```diff
@@ -1,9 +1,14 @@
 global.beforeEach(() => {
+    const appRoot = document.createElement('div');
+    appRoot.setAttribute('id', 'root');
+    document.body.appendChild(appRoot);
```
zarsky-broad: For the `attachTo` to make the PopupTrigger work.

ui/src/app/views/workspace-nav-bar/component.css
```diff
@@ -1,138 +1,6 @@
...
-  padding-left: 5rem;
-  box-shadow: 0 2px 5px 0 rgba(0, 0, 0, 0.26), 0 2px 10px 0 rgba(0, 0, 0, 0.16);
-}
-
-.btn-group .dropdown-menu clr-icon {
-  display: inline-grid;
-  margin-left: -0.4rem;
-}
-
 @media print {
```
zarsky-broad: Do we really expect anyone to print the page? Can this file be pulled entirely?

calbach: On further research, it appears this was intentional to allow for printing of one of the nav-bar'd pages (the cohort report). https://github.com/all-of-us/workbench/pull/1599

zarsky-broad: We should probably have a global definition for that class then, yes? I've seen it in a few other places

calbach: Yep, that sounds reasonable. Probably not worth breaking per-component CSS encapsulation in Angular, but given our current React setup it makes sense.

ui/src/app/icons/card-menu-icon/component.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import {mount} from 'enzyme';
+import * as React from 'react';
+
+import {
+  CardMenuIconComponentReact
+} from './component';
+
+// Note that the description is slightly different from the component for easier test filtering.
+describe('CardMenuIconComponent', () => {
```
calbach: This could be a point for discussion, but I'm not sure this test is worth it - assuming this icon gets rendered somewhere in the application which has a test.

ui/src/app/utils/index.tsx
```diff
@@ -190,7 +190,8 @@ export function reactStyles<T extends {[key: string]: React.CSSProperties }>(t:
 export class ReactWrapperBase implements OnChanges, OnInit, OnDestroy {
   @ViewChild('root') rootElement: ElementRef;
 
-  constructor(private WrappedComponent: React.ComponentType, private propNames: string[]) {}
+  constructor(private WrappedComponent: React.ComponentType | React.FunctionComponent,
```
calbach: Hm.. `ComponentType` is already a union of `ComponentClass<P> | FunctionComponent<P>`. Did this not compile?

ui/src/app/views/confirm-delete-modal/component.tsx
```diff
@@ -50,9 +49,7 @@ export class ConfirmDeleteModal
   }
 
   render() {
-    return <React.Fragment>
-      {this.props.deleting &&
-      <Modal className='confirmDeleteModal'>
+    return <Modal className='confirmDeleteModal'>
```
calbach: nit: while you're cleaning this, do we need the class?

ui/src/app/views/workspace-wrapper/component.ts
```diff
@@ -14,29 +15,29 @@ import {
...
   tabPath: string;
-  display = true;
+  displayNavBar = true;
   confirmDeleting = false;
 
   @ViewChild(BugReportComponent)
   bugReportComponent: BugReportComponent;
 
+  @ViewChild(WorkspaceNavBarComponent)
+  workspaceNavBarComponent: WorkspaceNavBarComponent;
```
calbach: This shouldn't be necessary since you don't utilize the reference anywhere. `@ViewChild` is only needed if you have to call a method or something on a child component.

ui/src/app/icons/card-menu-icon/component.spec.tsx
calbach: (even if it isn't... meh). Honestly these icons could probably all be thrown into one (or a small handful of) file(s) once we do away with the Angular wrapper; the boilerplate involved here always bothered me.

ui/src/app/views/confirm-delete-modal/component.tsx
jacmrob: it may be used in the test spec

ui/src/app/views/workspace-nav-bar/component.spec.tsx
jacmrob: awesome! good find

ui/src/app/views/workspace-nav-bar/component.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const navSpy = jest.fn();
+    NavStore.navigate = navSpy;
+    const wrapper = component();
+
+    wrapper.find({'data-test-id': 'Cohorts'}).first().simulate('click');
+    expect(navSpy).toHaveBeenCalledWith(
+      ['/workspaces', workspace.namespace, workspace.id, 'cohorts']);
+  });
+
+  it('should call delete method when clicked', () => {
```
jacmrob: testing if the delete method gets called seems to be a test of the delete modal or the dropdown menu.... should this component just test that the right options show up in the dropdown menu?

ui/src/app/views/workspace-nav-bar/component.tsx
```diff
@@ -0,0 +1,150 @@
...
+            icon='copy'
+            onClick={() => NavStore.navigate(['/workspaces', namespace, id, 'clone'])}>
+            Clone
+          </MenuItem>
+          <MenuItem
+            icon='pencil'
+            tooltip={isNotOwner && 'Requires owner permission'}
+            disabled={isNotOwner}
+            onClick={() => NavStore.navigate(['/workspaces', namespace, id, 'edit'])}
+          >
```
jacmrob: nit - put this `>` on above line

ui/src/app/views/workspace-wrapper/component.html
```diff
@@ -0,0 +1,21 @@
+<app-workspace-nav-bar *ngIf="displayNavBar"
```
jacmrob: quick ELI5 - what is workspace-wrapper as it relates to nav bar?  why aren't these all one thing?

ui/test-setup-script.js
jacmrob: 🎉🍾

ui/src/app/icons/card-menu-icon/component.spec.tsx
zarsky-broad: Once we have some control over webpack, I'd love to do what Terra does, and be able to import svg's directly into code.

.circleci/config.yml
```diff
@@ -226,7 +226,7 @@ jobs:
           name: Run React tests (Jest)
           working_directory: ~/workbench/ui
           command: |
-            npx jest
+            yarn test-react --runInBand
```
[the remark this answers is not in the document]
zarsky-broad: `--runInBand` is [the recommended fix](https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server) for that

ui/src/app/utils/index.tsx
zarsky-broad: Must have been a leftover from an experiment

ui/src/app/views/confirm-delete-modal/component.tsx
zarsky-broad: Can't find anywhere that uses it.

ui/src/app/views/workspace-nav-bar/component.spec.tsx
zarsky-broad: I want to make sure that both methods that are passed in as props are called correctly; I'm not worried that the right options won't be in the menu anymore than I am that the right tabs won't be in the bar.

ui/src/app/views/workspace-nav-bar/component.tsx
zarsky-broad: I've been doing this everywhere so that there's a line of visibly different indentation between the props of a component and its children, so that you can tell where one ends and the other begins. If we really hate it I can change it, but I find it very helpful for navigating code visually

ui/src/app/views/workspace-wrapper/component.html
zarsky-broad: The wrapper contains the nav bar, the workspace-level modals, and crucially, the router outlet. We may decide to do things differently once the react conversion is further along, but for the time being I needed an overarching angular component to control the modals and pass the functions to display them in.

PR #1732, opened 2019-02-01
un-revert workspace nav bar, fix circular reference AOT issue

panentheos:
This restores the workspace nav bar conversion with a fix for the AOT compilation issue.

The underlying issue appeared to be a circular reference, although it's not clear exactly where it was coming from. My assumption is that some of the angular core imports might do some sort of premature importing of e.g. the app module, which could lead to an circle.

Since the Component wrapper was unused in this case, we can just remove it to conveniently avoid the issue in this case. But we may need to move some files around if this comes up again.


calbach:
I'm wondering whether that may have actually been the issue (Angular component wasn't used, so compiler tried to remove something).

zarsky-broad:
@calbach Ooh, that makes sense. A little aggressive optimization...

PR #1733, opened 2019-02-02
[risk=no] Cache the yarn cache

calbach:
https://yarnpkg.com/lang/en/docs/cli/cache/
https://circleci.com/docs/2.0/caching/#yarn-node

Also freeze yarn.lock generation on Circle + deploys for improved determinism.

On a sample run, this reduced `yarn install` runtime from ~30 -> ~18s. Not great, but at least it's following the standard docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -212,7 +212,6 @@ jobs:
       - save_cache:
           paths:
             - ~/.cache/yarn
-            - ~/workbench/ui/node_modules
           key: ui-cache-{{ checksum "~/workbench/ui/yarn.lock" }}
       - run:
           name: Lint Angular app
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -212,6 +212,7 @@ jobs:
       - save_cache:
           paths:
             - ~/.cache/yarn
+            - ~/workbench/ui/node_modules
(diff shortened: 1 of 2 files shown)


zarsky-broad:
Why not cache the modules folder as well?

zarsky-broad:
Also: why does the cache folder have to be keyed? Wouldn't we want that for all builds? Even if it doesn't have all the exact deps it can only help, right?

calbach:
@blrubenstein thanks, added risk. "no ticket" is not a convention for this project.

@zarsky-broad this approach matches [Circle's docs](https://circleci.com/docs/2.0/caching/#yarn-node), I'd guess it may be redundant: it seems like the folders in node_modules are not quite symlinked from yarn cache (at least on my machine), but it contains the same contents and yarn likely just copies them from there. From a few data points testing this, it seems like it's slower (35-40s vs 20s). I can't quite explain this: perhaps yarn does some more expensive checksumming or something if there's already contents in node_modules. Reverting that for now.

The key defines when the cache gets stored again, so we definitely want it to change across runs so it's not getting overwritten on every PR push. We restore from the partial `ui-cache-` as a fallback, which should match any recent entry. I also added a cache restore in our deploy stage to account for the yarn chace.

PR #1734, opened 2019-02-04
[RW-1909][risk=low] Refactor Data Access Levels

rushtong:
## Addresses
https://precisionmedicineinitiative.atlassian.net/browse/RW-1909

See also: https://docs.google.com/document/d/1aE3ndSWjE0UguxET3oK9lTCu6dIqu7pPUwNiuSLYcvs/edit#heading=h.cbc54gkn65ua

## Changes
* `REVOKED` -> `UNREGISTERED`
* Refactored all data access level updates to go through `UserService`
* `Registered` logic updated to match tech doc


api/libproject/devstart.rb
```diff
@@ -984,23 +984,16 @@ def update_user_registered_status(cmd_name, args)
   common.run_inline %W{gcloud config set account #{op.opts.account}}
   header = "Authorization: Bearer #{token}"
   content_type = "Content-type: application/json"
-  payload = "{\"email\": \"#{op.opts.user}\"}"
+  payload = "{\"email\": \"#{op.opts.user}\", \"disabled\": \"#{op.opts.disabled}\"}"
   domain_name = get_auth_domain(op.opts.project)
-  if op.opts.action == "add"
-    common.run_inline %W{curl -H #{header} -H #{content_type}
+  common.run_inline %W{curl -X POST -H #{header} -H #{content_type}
       -d #{payload} https://api-dot-#{op.opts.project}.appspot.com/v1/auth-domain/#{domain_name}/users}
```
calbach: Existing issue - but these assumptions don't hold up anymore. We should add an explicit API endpoint field to our configs in this file and use that.

api/src/main/resources/workbench.yaml
```diff
@@ -429,39 +429,9 @@ paths:
           name: request
           description: request carrying user email to add
```
calbach: nit: update doc

api/libproject/devstart.rb
calbach: (the only outlier for now is that prod is: `api.workbench.researchallofus.org`)

rushtong: Yes, very true. Will do.

api/src/main/resources/workbench.yaml
rushtong: Good catch, thanks.

api/libproject/devstart.rb
rushtong: Done for this API, but have not gone through to refactor all of the other api calls. Filing a ticket to make sure we cover this.

api/src/main/resources/workbench.yaml
rushtong: Done.

api/src/test/java/org/pmiops/workbench/api/AuthDomainControllerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    ResponseEntity<EmptyResponse> response = this.authDomainController.createAuthDomain("");
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+  }
+
+  @Test
+  public void testDisableUser() {
+    AuthDomainDisableUserRequest request = new AuthDomainDisableUserRequest().
+        email(PRIMARY_EMAIL).
+        disabled(true);
+    ResponseEntity<Void> response = this.authDomainController.disableUser("", request);
```
calbach: Optional: This test would be a lot higher value if `UserService` were not mocked out (probably with a real UserDao as well). Right now these tests pretty much just assert that we had a method which called some other methods.

Even better would be if it called ProfileController afterwards and verified that the user is indeed disabled and/or changed in registration status (but I'd settle for just checking UserDao directly or verifying the FC call).

rushtong: Yes, I can see that - I'll spend some more time on that aspect.

rushtong: Test updated to check disabled status from `UserDao` for both enable and disable cases.

calbach: Nice - thanks!

PR #1735, opened 2019-02-04
[RW-1890][risk=no] convert recent work component to react

panentheos:
See ticket for details.


ui/src/app/views/recent-work/component.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {CohortsApi, ConceptSetsApi, UserMetricsApi, WorkspacesApi} from 'generated/fetch';
+import {CohortsApiStub} from 'testing/stubs/cohorts-api-stub';
+import {ConceptSetsApiStub} from 'testing/stubs/concept-sets-api-stub';
+import {UserMetricsApiStub} from 'testing/stubs/user-metrics-api-stub';
+import {WorkspacesApiStub} from 'testing/stubs/workspaces-api-stub';
+import {WorkspacesServiceStub} from 'testing/stubs/workspace-service-stub';
+import {WorkspaceAccessLevel} from 'generated';
+
+
+describe('RecentWorkComponent', () => {
```
jacmrob: Do we want to convert the rest of the existing recent work tests?  some of them seem useful

panentheos: Since this component is now relatively simple, it doesn't seem like having an extensive test suite would add much value relative to the time investment. Is there some particular logic that we're concerned about that's likely to have bugs?

gjuggler: (Comment from the peanut gallery)

Personally, I would hope to be checking that expected behavior is being roughly reproduced (and meanwhile documenting what that expected behavior is). Something like:
 - Verify we fetch & display components corresponding to data returned by the backend.
 - Verify that truncation is working as expected, e.g. 4 cards are shown if we have >4.
 - Verify that clicking on the shade produces the desired rotation of displayed cards.

It's probably feasible to shorten up the test code somewhat too (reducing / simplifying assertions and boilerplate). But IMO, for a component like this, a single render smoke test isn't enough for comfort (whereas for an icon component it certainly would be).

I'd still like to work together on a short test principles doc so we can have a shared understanding of what standards & guidelines to work against. @jacmrob would you be willing to get that started up?

panentheos: Agreed, it would be good to have a comprehensive team discussion about the costs and benefits of different kinds of tests.

PR #1736, opened 2019-02-04
[RW-1697][risk=no] ERA Commons acct linking UI

jacmrob:
*Notes*:
- The logic in @rushtong's `NihCallbackComponent` has been moved into the `AccountLinking` component.
- This also includes UI for Moodle training links (because I accidentally wrote it too).  Since this is behind a feature flag and the Moodle training functionality is going in in this sprint ([RW-1184](https://precisionmedicineinitiative.atlassian.net/browse/RW-1184)) I am okay keeping it in this PR as a stub.  
- The logic of if the AccountLinking component should render (via the `accountsLinked` var in the homepage) will mostly be refactored to go behind a guard component for [RW-1969](https://precisionmedicineinitiative.atlassian.net/browse/RW-1969) (lock out).


ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+    }
+  }
+
+  render() {
+    return <React.Fragment>
+      <div style={{display: 'flex', flexDirection: 'row'}}>
+        <div style={{display: 'flex', flexDirection: 'column', width: '50%'}}>
+          <div style={styles.mainHeader}>Researcher Workbench</div>
+          <div style={{marginLeft: '2rem', flexDirection: 'column'}}>
+            {/*<ClrIcon shape='exclamation-triangle' class='is-solid'/>*/}
```
calbach: Could you turn this into a TODO insetad? Or remove if no longer relevant?

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+              <div style={styles.infoBoxBody}>Clicking the training link will bring you to
+                the All of Us Compliance Training Portal, which will show you any
+                outstanding training material to be completed.</div>
+            </div>
+            <AccountLinkingButton failed={false}
+                                  completed={this.props.trainingCompleted}
+                                  defaultText='Complete Training'
+                                  completedText='Completed'
+                                  failedText=''
+                                  onClick={() => {}}/>
```
calbach: nit: Leave TODOs here for Moodle ticket?

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+@Component({
+  selector: 'app-account-linking',
+  template: '<div #root></div>',
+})
+export class AccountLinkingComponent extends ReactWrapperBase {
+  @Input('eraCommonsLinked') eraCommonsLinked: AccountLinkingProps['eraCommonsLinked'];
+  @Input('trainingCompleted') trainingCompleted: AccountLinkingProps['trainingCompleted'];
+  constructor() {
+    super(AccountLinking,
+        ['eraCommonsLinked', 'trainingCompleted']);
```
calbach: nit: unnecessary newline

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  }
+}
+
+
+@Component({
+  styleUrls: ['./component.css'],
+  templateUrl: './component.html',
+})
+
```
calbach: nit: rm \n

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  ngOnInit(): void {
+    this.profileService.getMe().subscribe(profile => {
+      if (profile.pageVisits) {
+        this.firstVisit = !profile.pageVisits.some(v =>
+        v.page === HomepageComponent.pageId);
+      }
+      if (environment.enableComplianceLockout) {
+        this.accountsLinked = !!profile.linkedNihUsername;
```
calbach: If `!enableComplianceLockout` then `this.accountsLinked` will never be set, so won't it be spinning forever? Seems like you should set it to `true` in that case.

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+    private profileStorageService: ProfileStorageService,
+    private route: ActivatedRoute,
+    private router: Router,
+  ) {
+    // create bound methods to use as callbacks
+    this.closeQuickTour = this.closeQuickTour.bind(this);
+  }
+
+  ngOnInit(): void {
+    this.profileService.getMe().subscribe(profile => {
```
calbach: Existing issue, please just add a TODO for now - but there are two profile requests happening in rapid succession on this page (this line, then the profileStorage reload below), which is really inefficient and might cause problems.

ui/src/environments/environment-type.ts
```diff
@@ -37,6 +37,10 @@ export interface Environment {
   // The base URL for the Zendesk help center / user forum.
   // Example value: https://aousupporthelp.zendesk.com/hc/
   zendeskHelpCenterUrl: string;
+  // The Shibboleth URL for linking ERA Commons accounts
+  shibbolethUrl: string;
+
+  rootUrl: string;
```
calbach: Were you unable to just get this from the window.location instead? I'd rather avoid encoding it here if possible since it's inherent to the application.

ui/src/environments/environment-type.ts
```diff
@@ -53,4 +57,6 @@ export interface Environment {
   // See RW-1966 for details.
   // Exit criteria: remove flag for Athens release.
   enableJupyterLab: boolean;
+
+  enableComplianceLockout: boolean;
```
calbach: Please add a comment similar to the others

calbach: I'm not sure whether `compliance` captures it. Maybe more broadly: `enableAthensRegistration`

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+      if (profile.pageVisits) {
+        this.firstVisit = !profile.pageVisits.some(v =>
+        v.page === HomepageComponent.pageId);
+      }
+      if (environment.enableComplianceLockout) {
```
calbach: This should be a combination of `enforceRegistered` and the new flag, most likely.

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+export class AccountLinking extends
+    React.Component<AccountLinkingProps, {eraCommonsLinked: boolean, eraCommonsError: string}> {
+
+  constructor(props: AccountLinkingProps) {
+    super(props);
+
+    this.state = {
+      eraCommonsError: '',
+      eraCommonsLinked: props.eraCommonsLinked
```
calbach: Why put this into state? It never changes.

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  styleUrls: ['./component.css'],
+  templateUrl: './component.html',
+})
+
+export class HomepageComponent implements OnInit, OnDestroy {
+  private static pageId = 'homepage';
+  @ViewChild('myVideo') myVideo: any;
+  profile: Profile;
+  view: any[] = [180, 180];
+  numberOfTotalTasks = 4;
```
calbach: Existing issue: There's definitely a bunch of dead code here - do you mind putting a TODO and filing a ticket for this cleanup (including my other comment around the double profile calls)

ui/src/app/views/nih-callback/component.spec.tsx
```diff
@@ -1,49 +0,0 @@
-import {mount, shallow} from 'enzyme';
```
calbach: Carry over test somehow?

ui/src/app/views/homepage/component.tsx
jacmrob: good catch!

ui/src/environments/environment-type.ts
jacmrob: we want to use the homepage specifically as the redirect callback even if we are on another page.  this seemed cleaner than putting some parsing logic on window.location. what do you think?

ui/src/app/views/homepage/component.tsx
jacmrob: that cleanup can probably be done when converting the homepage component.  I'll make a note in that ticket.

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  render() {
+    return <React.Fragment>
+      <div style={{display: 'flex', flexDirection: 'row'}}>
+        <div style={{display: 'flex', flexDirection: 'column', width: '50%'}}>
+          <div style={styles.mainHeader}>Researcher Workbench</div>
+          <div style={{marginLeft: '2rem', flexDirection: 'column'}}>
+            {/*<ClrIcon shape='exclamation-triangle' class='is-solid'/>*/}
+            <div style={styles.minorHeader}>In order to get access to data and tools
```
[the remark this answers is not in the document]
jacmrob: This is rendered behind the flag-- see the `environment.enableComplianceLockout` conditional in ngInit and lines 25-31 in the html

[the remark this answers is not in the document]
rushtong: Ah, thank you, I missed that!

ui/src/environments/environment-type.ts
```diff
@@ -37,6 +37,10 @@ export interface Environment {
   // The base URL for the Zendesk help center / user forum.
   // Example value: https://aousupporthelp.zendesk.com/hc/
   zendeskHelpCenterUrl: string;
+  // The Shibboleth URL for linking ERA Commons accounts
+  shibbolethUrl: string;
```
gjuggler: Super nitpick: at least in this file, try to keep keys alphabetized so things are easy to find. (I've been though this before... alpha is the only way to sort things like this without it becoming a complete mess, as configs like this would otherwise do.)

ui/src/app/views/homepage/component.tsx
```diff
@@ -0,0 +1,394 @@
...
+    </Clickable>;
+  }
+};
+
+export interface AccountLinkingProps {
+  eraCommonsLinked: boolean;
+  trainingCompleted: boolean;
+}
+
+export class AccountLinking extends
```
gjuggler: Naming nitpick: "Account Linking" doesn't quite capture that this component will also show compliance training completion steps required for access to the workbench.

And will we also use this component to show users some information when their account has been disabled – or do they just see a straight 403? (Same goes w/ ID Verification, which will probably become a blocking attribute that replaces ERA commons & will need to be included in this screen.)

Possible alternatives: "UserLockout", "WorkbenchAccessDenied", "AccessDeniedView", "StepsRequiredForAccess"?

@calbach for thoughts.

ui/src/app/views/homepage/component.tsx
calbach: When we return from the redirect, that's a hard page reload so this client state is no longer relevant. I don't see any `setState` affecting `eraCommonsLinked` in this file.

ui/src/app/views/homepage/component.tsx
calbach: Agreed. Naming is hard, I don't have a great recommendation here, another option: UnregisteredSection

ui/src/app/views/homepage/component.tsx
calbach: sounds good - thanks

ui/src/app/views/homepage/component.tsx
jacmrob: this has been resolved

PR #1738, opened 2019-02-04
[RW-2078][Risk=no] Remove signed in parent from routes

blrubenstein:
This looks very large, but should mostly be whitespace changes.

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -1,5 +1,5 @@
 import {NgModule} from '@angular/core';
-import {NavigationEnd, Router, RouterModule, Routes} from '@angular/router';
+import {NavigationEnd, Route, Router, RouterModule, Routes} from '@angular/router';
 
 import {RegistrationGuard} from './guards/registration-guard.service';
 import {SignInGuard} from './guards/sign-in-guard.service';
@@ -16,7 +16,6 @@ import {NotebookListComponent} from './views/notebook-list/component';
 import {NotebookRedirectComponent} from './views/notebook-redirect/component';
 import {ProfilePageComponent} from './views/profile-page/component';
 import {SettingsComponent} from './views/settings/component';
-import {SignedInComponent} from './views/signed-in/component';
 import {StigmatizationPageComponent} from './views/stigmatization-page/component';
 import {UnregisteredComponent} from './views/unregistered/component';
 import {WorkspaceEditComponent, WorkspaceEditMode} from './views/workspace-edit/component';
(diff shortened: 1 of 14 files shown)


ui/src/app/app-routing.module.ts
```diff
@@ -32,233 +31,242 @@ import {environment} from 'environments/environment';
...
+    component: SettingsComponent,
+    data: {title: 'Settings'}
+  }, {
+    path: 'workspaces/build',
+    component: WorkspaceEditComponent,
+    data: {title: 'Create Workspace', mode: WorkspaceEditMode.Create}
   }
 ];
 
+fp.forEach(route => {
```
calbach: nit: is it really preferable to use `fp` for this? Why not just use built-ins `[].forEach()`

ui/src/app/app-routing.module.ts
```diff
@@ -32,233 +31,242 @@ import {environment} from 'environments/environment';
...
+  if (!route.isPublic) {
+    route.canActivate = [SignInGuard];
+    route.canActivateChild = [SignInGuard, RegistrationGuard];
+    route.runGuardsAndResolvers = 'always';
```
calbach: I've messed around with TS annotations a bit and I'm surprised that something like this would work (but it might, since it compiled). I would recommend running locally with `--aot` here and manually testing to double check: https://github.com/all-of-us/workbench/blob/3862f270ed7f3360d7ce97a41dd741ffac7fecf2/ui/docker-compose.yaml#L12

ui/src/app/app-routing.module.ts
```diff
@@ -32,233 +31,242 @@ import {environment} from 'environments/environment';
...
+import * as fp from 'lodash/fp';
+
 declare let gtag: Function;
 
-const routes: Routes = [
+interface RouteExtension extends Route {
+  isPublic?: boolean;
+}
+
+declare type RoutesExtension = RouteExtension[];
```
calbach: nit: this type does not seem necessary, I'd just use `RouteExtension[]` directly below

ui/src/app/views/unregistered/component.spec.ts
```diff
@@ -73,11 +81,13 @@ describe('UnregisteredComponent', () => {
     profileStorageStub = new ProfileStorageServiceStub();
     TestBed.configureTestingModule({
       declarations: [
+        ...signedInDependencies.declarations,
         FakeAppComponent,
         FakeRootComponent,
         UnregisteredComponent,
       ],
       imports: [
+        ...fp.without([RouterTestingModule], signedInDependencies.imports),
```
calbach: Did this cause problems? Doesn't the last one win anyways?

ui/src/testing/test-helpers.ts
```diff
@@ -64,3 +83,29 @@ export function findElementsReact<C>(
 ) {
   return [].slice.call(fixture.debugElement.nativeElement.querySelectorAll(selector));
 }
+
+export const signedInDependencies = {
```
calbach: Thanks for factoring this out - this is nice.

nit: Is there a reason not to just define normal Angular module and import it into your tests?

ui/src/app/app-routing.module.ts
blrubenstein: No difference. Changing.

ui/src/app/app-routing.module.ts
blrubenstein: This totally doesn't work with `--aot` :( 
May I pair with you at some point? I can't figure out a good way to do this that doesn't add ridiculous amounts of boilerplate.

blrubenstein: Otherwise... I don't know, maybe we actually do punt on this PR and make the router conversion PR a bit bigger. I think this is much more straightforward in react/with the terra router.

PR #1727, opened 2019-02-01
[RW-1988][risk=no] Updating PPI tree to make Survey and Questions selectable

freemabd1:
Updating PPI tree to make Survey and Questions selectable

PR #1739, opened 2019-02-04
[RW-2076][risk=low] Fix BigQuery integration tests

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2076

- Reverts https://github.com/all-of-us/workbench/pull/1728 temporarily, tried fixing the tests here but I don't understand the original issue fully and I'm not sure about expected behavior so should wait on @freemabd1 . Per discussion with @dolbeew , sounds like this is an existing issue which is probably not critical to the next release.
- Fixes error message testing to match updated behavior from #1727 .


calbach:
Verified `./project.rb bigquerytest` passes locally.

PR #1745, opened 2019-02-05
[risk=no] Add some docs around test filtering via CLI

calbach:


calbach:
There's a passing mentions of this in the command strings in devstart.rb. These aren't very discoverable as AFAIK this only shows up when you list all the commands out:

https://github.com/all-of-us/workbench/blob/a567527417c1433813f0db23aadaa2590d7b2366/api/libproject/devstart.rb#L373

PR #1750, opened 2019-02-06
[RW-2034][risk=no] Auto-populate Institution and contact email address for new users

gjuggler:
This PR makes a few changes to populate the custom GSuite schema fields required for Moodle SAML integration.

While I was in this code, I also changed the way we store a user's contact email address in GSuite. This will now show up in two places:

1) The "Contact_email_address" custom schema field.
2) A secondary email address with type "home".

Plus some small additions to the integration test to verify these things work.

RELEASE COORDINATION: the custom schema has already been added to our dev GSuite, but not prod. I will make those changes before this PR gets merged, after which point this PR will be safe to run in all environments.

### DIFF
--- a/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
@@ -4,10 +4,12 @@ import static com.google.common.truth.Truth.assertThat;
 
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.http.apache.ApacheHttpTransport;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.time.Clock;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.pmiops.workbench.config.WorkbenchConfig;
@@ -30,8 +32,8 @@ public class DirectoryServiceImplIntegrationTest {
   public void setUp() throws MessagingException {
     MailService mailService = Mockito.mock(MailServiceImpl.class);
     service = new DirectoryServiceImpl(
-        Providers.of(googleCredential), Providers.of(workbenchConfig), httpTransport,
-        new GoogleRetryHandler(new NoBackOffPolicy()));
+      Providers.of(googleCredential), Providers.of(workbenchConfig), httpTransport,
+      new GoogleRetryHandler(new NoBackOffPolicy()));
   }
 
   @Test
@@ -52,9 +54,9 @@ public class DirectoryServiceImplIntegrationTest {
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(service.getContactEmailAddress(userName).equals("notasecret@gmail.com"));
     assertThat(
-            service.getUserByUsername(userName).getCustomSchemas()
-                    .get("All_of_Us_Workbench")
-                    .get("Institution").equals("All of Us Research Workbench"));
+      service.getUserByUsername(userName).getCustomSchemas()
+        .get("All_of_Us_Workbench")
+        .get("Institution").equals("All of Us Research Workbench"));
     service.deleteUser(userName);
     assertThat(service.isUsernameTaken(userName)).isFalse();
   }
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -102,39 +128,64 @@ public User getUser(String email) {
...
+        .setCustomSchemas(Collections.singletonMap(GSUITE_AOU_SCHEMA_NAME, aouCustomFields))
         .setChangePasswordAtNextLogin(true);
     retryHandler.run((context) -> getGoogleDirectoryService().users().insert(user).execute());
     return user;
   }
 
   @Override
-  public User resetUserPassword(String userKey) {
-    User user = getUser(userKey);
+  public User resetUserPassword(String email) {
```
gjuggler: Side note: I tried to clean up the terminology (at least within this file), so "email" is always the full email address "gjordan@researchallofus.net" and "username" is just the first part, "gjordan". In other files (and our SQL database) the distinction is not always so clear, and even Google isn't super specific about what is required for some of these API calls. For example, from the [API docs](https://developers.google.com/admin-sdk/directory/v1/guides/manage-users#update_user):

> The userKey can be the user's primary email address, the unique user id, or one of the user's alias email addresses. 

In my tests using the APIs explorer, the userKey parameter must be the full email address of the user.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -47,8 +47,14 @@ public void testDirectoryServiceUsernameIsTaken() {
   @Test
   public void testCreateAndDeleteTestUser() {
     String userName = String.format("integration.test.%d", Clock.systemUTC().millis());
-    service.createUser("Integration", "Test", userName, "notasecret");
+    service.createUser("Integration", "Test", userName, "notasecret@gmail.com");
```
gjuggler: Interestingly, now that we're storing a user's contact email address in a custom schema field with a typed value of "email address", this test was failing (API response 400 from Google) with the original "notasecret" value. Yay for extra field validation!

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -47,8 +47,14 @@ public void testDirectoryServiceUsernameIsTaken() {
...
     assertThat(service.isUsernameTaken(userName)).isTrue();
+    // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
+    assertThat(service.getContactEmailAddress(userName).equals("notasecret@gmail.com"));
+    assertThat(
+            service.getUserByUsername(userName).getCustomSchemas()
+                    .get("All_of_Us_Workbench")
+                    .get("Institution").equals("All of Us Research Workbench"));
```
rushtong: nit: spacing. I believe we're using 2-space tabs in these cases.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -1,3 +1,7 @@
+/**
+ * Note: this file is tested with integration tests rather than unit tests. See
+ * src/integration/.../DirectoryServiceImplIntegrationTest.java for test cases.
+ */
```
rushtong: 👍 for more integration testing!

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
gjuggler: Gotcha. Is there some way I can set up IntelliJ (& document / provide a standard config for the team), or run some formatting script which allows me to not have to think about this?

calbach: Would love to get auto-formatting setup in this project - right now our rules are extremely loosely defined, and only automatically checked in Codacy (by UI config only) , which is a huge pain and not config-driven.

Possibly:
- https://github.com/google/google-java-format
- https://clang.llvm.org/docs/ClangFormat.html

gjuggler: Fixed the spacing here. I'll look into auto-format sometime next week, thanks for the pointers!

gjuggler:
CH confirmed that the schema changes are now live in the prod GSuite, which means this can be included in the next release. Merging now.

PR #1751, opened 2019-02-06
[RW-1969][risk=no] Lock out users until ERA commons is authed

jacmrob:


ui/src/app/guards/era-common-guard.service.ts
```diff
@@ -0,0 +1,37 @@
+import {Injectable} from '@angular/core';
+import {
+  ActivatedRouteSnapshot, CanActivate, CanActivateChild, Router,
+  RouterStateSnapshot
+} from '@angular/router';
+import {ProfileStorageService} from 'app/services/profile-storage.service';
+import {Observable} from 'rxjs/Observable';
+
+@Injectable()
+export class EraCommonGuard implements CanActivate, CanActivateChild {
```
rushtong: nit: naming. `EraCommonsGuard` would be more accurate (note the `s` in Commons)

ui/src/app/guards/era-common-guard.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot):
+  Observable<boolean> {
+    if (route.routeConfig.path === '' ||
+        route.routeConfig.path.startsWith('nih-callback/')) {
+      // Leave /admin unguarded in order to allow bootstrapping of verified users.
+      return Observable.from([true]);
+    }
+
+    return this.profileStorageService.profile$.flatMap((profile) => {
+      if (profile.linkedNihUsername === null) {
```
rushtong: Should we also guard against the empty string?

ui/src/app/app-routing.module.ts
```diff
@@ -42,7 +43,7 @@ const routes: Routes = [
     path: '',
     component: SignedInComponent,
     canActivate: [SignInGuard],
-    canActivateChild: [SignInGuard, RegistrationGuard],
+    canActivateChild: [SignInGuard, RegistrationGuard, EraCommonGuard],
```
calbach: We need a plan for how this interacts with `RegistrationGuard` and the unregistered page. As it stands in this PR, you'd need to be manually ID verified before you could get to the dashboard to click the ERA commons link, which defeats the purpose of using ERA commons for ID verification. At the same time, we probably still want a manual IDV step so we can continue whitelisting initial users - I don't know whether this has been discussed yet.

Previous treatment for an unregistered user:
- Sign in
- RegistrationGuard redirects to /unregistered
- unregistered page autosubmits tos/training/manual IDV request/"demographics survey"

Possible new treatment?
- Sign in
- RegistrationGuard redirects to /unregistered, if tos/demo survey not submitted or manual IDV not done
- unregistered page autosubmits tos/"demographics survey", requests manual IDV

Alternatively:
- Repurpose the RegistrationGuard to redirect to the homepage instead (if the flag is enabled)
- Autosubmit tos/demo survey from that page if not submitted.
- Somehow also block the user here for manual IDV

ui/src/app/views/homepage/component.tsx
```diff
@@ -281,13 +282,23 @@ export class HomepageComponent implements OnInit, OnDestroy {
...
+      // Set Access Tasks flags
+      // TODO RW-1184 set trainingCompleted flag
+      this.eraCommonsLinked = !!profile.linkedNihUsername;
+
+      if (this.route.snapshot.queryParams.workbenchAccessTasks) {
+        // To reach the access tasks component from dev use /?workbenchAccessTasks=true
+        this.accessTasksRemaining = true;
+      } else {
+        this.serverConfigService.getConfig().subscribe((config) => {
+          if (environment.enableComplianceLockout || config.enforceRegistered) {
```
calbach: &&

ui/src/app/guards/access-tasks-guard.service.ts
```diff
@@ -4,23 +4,32 @@ import {
...
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot):
   Observable<boolean> {
     if (route.routeConfig.path === '' ||
-        route.routeConfig.path.startsWith('nih-callback/')) {
-      // Leave /admin unguarded in order to allow bootstrapping of verified users.
+        route.routeConfig.path.startsWith('nih-callback')) {
       return Observable.from([true]);
     }
+    let enforceRegistered: boolean;
+    this.serverConfigService.getConfig().subscribe((config) => {
```
calbach: These callbacks (here, and the profile call) are likely, but not guaranteed to execute in the order you are expecting here. Should wrap one of these callbacks in the other (or chain with flatmap)

ui/src/app/views/homepage/component.tsx
jacmrob: good catch. changed that during the demo

ui/src/app/guards/access-tasks-guard.service.ts
```diff
@@ -22,21 +22,20 @@ export class AccessTasksGuard implements CanActivate, CanActivateChild {
         route.routeConfig.path.startsWith('nih-callback')) {
       return Observable.from([true]);
     }
-    let enforceRegistered: boolean;
     this.serverConfigService.getConfig().subscribe((config) => {
-      enforceRegistered = config.enforceRegistered;
-    });
+      return this.profileStorageService.profile$.flatMap((profile) => {
+        if ((profile.linkedNihUsername === null || profile.linkedNihUsername === '')
+            && config.enforceRegistered) {
```
calbach: Seems like you should be checking `enableComplianceLockout` as well here, right?

jacmrob: good call :)

PR #1753, opened 2019-02-07
RW-1956 initial go at ElasticService.

freemabd1:
This PR is **purely a prototype** to get feedback on the approach with elasticsearch.
I'm using Java high-level REST client per your comments.
I've listed some questions in comments in the code. 
Please let me know what is good/bad about this.


api/src/main/java/org/pmiops/workbench/config/ElasticSearchConfig.java
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.inject.Provider;
+
+@Configuration
+public class ElasticSearchConfig {
+
+  @Autowired
+  private Provider<WorkbenchConfig> configProvider;
+
+  //Is Singleton scope appropriate here? or should we use another scope for elasticsearch?
```
calbach: Docs indicate Singleton is correct here. Could you add this link to the doc here as well? 

https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_changing_the_client_8217_s_initialization_code.html

api/src/main/java/org/pmiops/workbench/config/ElasticSearchConfig.java
```diff
@@ -0,0 +1,29 @@
...
+  //If a scope beside Singleton is used we need to make sure we close this client when finished.
+  @Bean
+  public RestHighLevelClient client() {
+    HttpHost httpHost = new HttpHost(
```
calbach: Could you also add a feature flag to the config in this PR? e.g. enableElasticsearchBackend - then just throw an exception or return null here if that's not on. Want to make sure this isn't crashing in higher environments where we don't have ES yet.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * This method will be used to generate counts from a
+   * org.pmiops.workbench.model.SearchRequest.
+   *
+   * @param userRequest
+   * @return
+   */
+  public Long elasticCount(org.pmiops.workbench.model.SearchRequest userRequest) throws IOException {
+    CountRequest countRequest = new CountRequest()
+      //What indices should we use here?
```
calbach: These are in the design doc, "type" is pretty much deprecated and will be "_doc" for all types. The index will change depending on the CDR version.

Open question: do we want to leverage the CdrVersionContext here as well?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,55 @@
...
+      .indices(INDEX, TYPE)
+      .source(new SearchSourceBuilder().query(createQueryBuilder(userRequest)));
+    return client.count(countRequest, RequestOptions.DEFAULT).getCount();
+  }
+
+  //Is it reasonable to user QueryBuilders for this? Or do we need something more flexible to
```
calbach: Good question, we should think about what interfaces we want here. My high level idea was that there would somehow be a common interface for use between BigQuery and Elastic to get counts. I'm not sure whether this holds up though.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * This method will be used to generate counts from a
+   * org.pmiops.workbench.model.SearchRequest.
+   *
+   * @param userRequest
+   * @return
+   */
+  public Long elasticCount(org.pmiops.workbench.model.SearchRequest userRequest) throws IOException {
```
calbach: FYI: For this PR - totally fine to just have a placeholder proof-of-concept method here, e.g. returning cluster status.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.stereotype.Service;
+
+import java.io.IOException;
+
+@Service
+public class ElasticSearchService {
+
+  private static final String INDEX = "index";
+  private static final String TYPE = "type";
+  private RestHighLevelClient client;
```
calbach: nit: final?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  //Is it reasonable to user QueryBuilders for this? Or do we need something more flexible to
+  //to accomplish all of our query needs?
+  private AbstractQueryBuilder createQueryBuilder(org.pmiops.workbench.model.SearchRequest userRequest) {
+    //Convert cohort definition into a QueryBuilder?
+    //May need to generate different QueryBuilders depending on the cohort definition complexity???
+    //Need to gather more knowledge about what needs to happen to get counts from elasticsearch
+
+    //This is an example of what might get returned from this conversion??
```
calbach: Nice - yes, I think we would use a request object like this as an intermediate. We can think about whether we actually need to expose a method like this outside the ElasticSearchService, or if that's just an implementation detail.

api/src/main/java/org/pmiops/workbench/config/ElasticSearchConfig.java
freemabd1: Added link.

api/src/integration/java/org/pmiops/workbench/TestWebMvcConfig.java
```diff
@@ -15,4 +19,23 @@
     public WorkbenchEnvironment workbenchEnvironment() {
         return new WorkbenchEnvironment(true, "appId");
     }
+
+    /**
+     * The loading of the spring application context for ApplicationTest
+     * was throwing a BeanCreationException: Scope 'request' is not active
+     * for the current thread. Since WorkbenchConfig is request scoped using
```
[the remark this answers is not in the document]
calbach: FirecloudConfig does this - I guess the difference is that it injects it into the provider rather than autowiring into the configuration class itself.

api/src/main/java/org/pmiops/workbench/config/ElasticSearchConfig.java
freemabd1: Done.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: Will do. I'm running into some issues that need to be solved before i can do this: https://docs.google.com/document/d/1hwXHlOW7wBpAQrj0UygMnQnO1pfBiSVSsKsf6C6DirM/edit#

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: I think we do.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: Agreed.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: Ok

api/config/config_prod.json
```diff
@@ -41,5 +41,9 @@
   "mandrill": {
     "fromEmail": "donotreply@researchallofus.org",
     "sendRetries": 3
+  },
+  "elasticsearch": {
+    "host": "elastic:9200",
```
calbach: nit: I'd just leave an `""` for these non-local hosts for now

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,116 @@
...
+    //Will probably use elasticsearch querybuilders
+    //https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-query-builders.html
+    return 0L;
+  }
+
+  /**
+   * Get the count with BigQuery.
+   * @param userRequest
+   */
+  private Long getBigQueryCount(org.pmiops.workbench.model.SearchRequest userRequest) {
```
calbach: I don't think anything BQ related in this service, unless we're renaming this something more general, e.g. `ParticipantCountService`. A level above this should probably deal with which service to call (e.g. just in the API controller is probably fine for now.)

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param userRequest
+   */
+  public Long elasticCount(org.pmiops.workbench.model.SearchRequest userRequest) {
+    //For now if enableElasticsearchBackend is true
+    //we get the cluster name, status and log it
+    //then fallback to bigquery count.
+    //TODO: Implement this method to call either getElasticCount()
+    //or getBigQueryCount() based on certain condition(s) like -> if it's
+    //temporal or not
+    if (configProvider.get().elasticsearch.enableElasticsearchBackend) {
```
calbach: I think I'd just move this up into the API controller and let this service handle Elastic-only.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * This method will generate counts per the
+   * org.pmiops.workbench.model.SearchRequest
+   * using elasticsearch with BigQuery being
+   * the fallback.
+   *
+   * @param userRequest
+   */
+  public Long elasticCount(org.pmiops.workbench.model.SearchRequest userRequest) {
```
calbach: nit: why fully qualify SearchRequest?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -0,0 +1,116 @@
...
+  public ElasticSearchService(Provider<WorkbenchConfig> configProvider,
+                              BigQueryService bigQueryService,
+                              ParticipantCounter participantCounter) {
+    this.configProvider = configProvider;
+    this.bigQueryService = bigQueryService;
+    this.participantCounter = participantCounter;
+  }
+
+  /**
+   * This method will generate counts per the
```
calbach: nit: this is some aggressive wrapping - are you optimizing for mobile IDE users?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: Done.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -30,29 +30,26 @@ public ElasticSearchService(Provider<WorkbenchConfig> configProvider) {
...
-  public Long elasticCount(SearchRequest userRequest) {
+  public Long elasticCount() {
     //For now if enableElasticsearchBackend is true
     //we get the cluster name, status and log it
     try {
       ClusterHealthRequest healthRequest = new ClusterHealthRequest();
       ClusterHealthResponse response = client().cluster().health(healthRequest, RequestOptions.DEFAULT);
 
       log.info(String.format("Cluster Name: %s - Cluster Status: %s", response.getClusterName(), response.getStatus().name()));
+      log.info(String.format("Cluster Name: %s - Cluster Status: %s", response.getClusterName(), response.getStatus().name()));
```
calbach: nit: duplicated logger line?

freemabd1: Doh.. not sure what happened there. Done.

PR #1763, opened 2019-02-08
[risk=no] extract current cohort from route

panentheos:
This creates `currentCohortStore` as a mechanism to expose the current cohort without depending on the router. Also converts `page-layout` and `overview-page` as example usages.


UrsaStutsman:
LGTM!

PR #1764, opened 2019-02-08
Srubenst/cohort list page

blrubenstein:


ui/src/app/views/cohort-list/component.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {CohortsApi, WorkspacesApi} from 'generated/fetch';
+import {CohortsApiStub, exampleCohortStubs} from 'testing/stubs/cohorts-api-stub';
+import {WorkspacesApiStub} from 'testing/stubs/workspaces-api-stub';
+
+import {
+  WorkspaceAccessLevel,
+} from 'generated';
+import {waitOneTickAndUpdate} from 'testing/react-test-helpers';
+
+describe('CohortList', () => {
```
jacmrob: I think this spec should also have the two tests that were in the angular spec - checking that if we delete a cohort the view updates, and that if we edit a cohort the view changes

ui/src/app/views/cohort-list/component.tsx
```diff
@@ -0,0 +1,147 @@
...
+  },
+
+  infoIcon: {
+    color: '#2691D0',
+    cursor: 'pointer',
+    marginLeft: '0.2rem',
+    height: '16px',
+    width: '16px'
+  },
+  addCard: {
```
jacmrob: could any of these styles be moved into `app/components/buttons` to be used commonly for all the add cards on the list pages?

ui/src/app/views/cohort-list/component.tsx
```diff
@@ -0,0 +1,147 @@
...
+    async reloadCohorts() {
+      try {
+        const {namespace, id, accessLevel} = this.props.workspace;
+        this.setState({
+          cohortsLoading: true
+        });
+        const resp = await cohortsApi().getCohortsInWorkspace(namespace, id);
+        this.setState({
+          cohortList: convertToResources(resp.items, namespace,
+            id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.COHORT),
```
jacmrob: nice! does this get around the casting for the fetch resource types?

blrubenstein: Sort of. It handles it at a single level, but nested values I think still cause a problem. But that's the syntax around it.

ui/src/app/views/cohort-list/component.tsx
blrubenstein: I don't think the icon could, because the abstraction used doesn't actually specify a specific action, and the margins and width/height also seem page specific, but I think the font size, weight, and height could be.

ui/src/app/views/cohort-list/component.spec.tsx
blrubenstein: I actually talked through with Brett on this one, and removing those was intentional. I don't think there's much value in testing it, because testing that it actually calls the `onDone` function should be done by the modal, and the logic for it here is a really simple callback that just calls `loadCohorts`, which should be tested by the basic test.

PR #1769, opened 2019-02-11
[DB-234] [risk=no] include physical measurements in search results

SrushtiGangireddy:
1. include physical measurements in the search results.

### DIFF
--- a/public-ui/src/app/app-routing.module.ts
+++ b/public-ui/src/app/app-routing.module.ts
@@ -51,7 +51,7 @@ const routes: Routes = [
         data: {title: 'View Full Results'}
       },
       {
-        path: 'physical-measurements/:searchString',
+        path: 'physical-measurements',
         component: PhysicalMeasurementsComponent,
         data: {title: 'Physical Measurements from Enrollment'}
       }
--- a/public-ui/src/app/app-routing.module.ts
+++ b/public-ui/src/app/app-routing.module.ts
@@ -51,7 +51,7 @@ const routes: Routes = [
         data: {title: 'View Full Results'}
       },
       {
-        path: 'physical-measurements',
+        path: 'physical-measurements/:searchString',
         component: PhysicalMeasurementsComponent,
         data: {title: 'Physical Measurements from Enrollment'}
       }
--- a/public-ui/src/app/views/pm/pm.component.ts
+++ b/public-ui/src/app/views/pm/pm.component.ts
@@ -40,10 +40,6 @@ export class PhysicalMeasurementsComponent implements OnInit, OnDestroy {
 
   constructor(private api: DataBrowserService, public dbc: DbConfigService,
               private route: ActivatedRoute) {
-    this.route.params.subscribe(params => {
-      this.searchString = params.searchString;
-    });
-
   }
 
   loading() {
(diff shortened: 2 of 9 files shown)


public-ui/src/app/app-routing.module.ts
```diff
@@ -51,7 +51,7 @@ const routes: Routes = [
         data: {title: 'View Full Results'}
       },
       {
-        path: 'physical-measurements',
+        path: 'physical-measurements/:searchString',
```
calbach: It's a bit more typical to use a query parameter for this kind of thing - you can subscribe to query parameters in the same way you're doing for normal parameters.

e.g. `/physical-measurements?q=asdf`

public-ui/src/app/views/pm/pm.component.ts
```diff
@@ -48,6 +54,10 @@ export class PhysicalMeasurementsComponent implements OnInit, OnDestroy {
     this.loadingStack.push(true);
     this.dbc.getPmGroups().subscribe(results => {
       this.conceptGroups = results;
+      if (this.searchString) {
+        this.conceptGroups = this.conceptGroups.filter(conceptgroup =>
+          conceptgroup.groupName.toLowerCase().includes(this.searchString));
```
calbach: SHould you also be lowercasing the search string?

public-ui/src/app/views/quick-search/quick-search.component.ts
```diff
@@ -189,5 +196,18 @@ export class QuickSearchComponent implements OnInit, OnDestroy {
     localStorage.setItem('searchText', this.prevSearchText);
     this.router.navigateByUrl('/ehr/' + r.domain.toLowerCase());
   }
+  public viewPhysicalMeasurements(searchString: string) {
```
calbach: nit:newlines?

public-ui/src/app/views/quick-search/quick-search.component.ts
```diff
@@ -189,5 +196,18 @@ export class QuickSearchComponent implements OnInit, OnDestroy {
...
+    this.router.navigateByUrl('/physical-measurements/' + searchString.toLowerCase());
+  }
+  public matchPhysicalMeasurements(searchString: string) {
+      if (searchString) {
```
calbach: I think this is missing some cases. Probably cleaner would be:

```
if (!this.pmConceptGroups) {
  return 0;
}
if (!searchString) {
 return this.pmConceptGropus.length
}
return filter...
```

SrushtiGangireddy: Addressed. Thanks.

public-ui/src/app/views/quick-search/quick-search.component.ts
SrushtiGangireddy: Done.

public-ui/src/app/views/pm/pm.component.ts
SrushtiGangireddy: Done.

public-ui/src/app/app-routing.module.ts
SrushtiGangireddy: Changed the param to query param.

public-ui/src/app/views/pm/pm.component.ts
```diff
@@ -35,19 +36,27 @@ export class PhysicalMeasurementsComponent implements OnInit, OnDestroy {
...
 
   ngOnInit() {
+    this.route.queryParams.subscribe(params => {
+      this.searchString = params.searchString || null;
+    });
     this.loadingStack.push(true);
     this.dbc.getPmGroups().subscribe(results => {
       this.conceptGroups = results;
+      if (this.searchString) {
+        this.conceptGroups = this.conceptGroups.filter(conceptgroup =>
```
calbach: camelCase

PR #1771, opened 2019-02-11
[risk=no] refactor away router dependencies

panentheos:
This transforms several components with the goal of removing direct router dependencies. The general approach is to read directly from our new 'stores', instead of the router data. Adds `queryParamsStore` to hold url query params. It also cleans up a couple pieces of data that were unused.

### DIFF
--- a/ui/src/app/cohort-search/demographics/demographics.component.spec.ts
+++ b/ui/src/app/cohort-search/demographics/demographics.component.spec.ts
@@ -2,7 +2,6 @@ import {dispatch, NgRedux} from '@angular-redux/store';
 import {MockNgRedux} from '@angular-redux/store/testing';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import {ReactiveFormsModule} from '@angular/forms';
-import {ActivatedRoute} from '@angular/router';
 import {ClarityModule} from '@clr/angular';
 import {MultiSelectComponent} from 'app/cohort-search/multi-select/multi-select.component';
 import {fromJS} from 'immutable';
@@ -49,12 +48,6 @@ describe('DemographicsComponent', () => {
         {provide: NgRedux, useValue: mockReduxInst},
         {provide: CohortBuilderService, useValue: {}},
         {provide: CohortSearchActions, useValue: new MockActions()},
-        {
-          provide: ActivatedRoute,
-          useValue: {
-            snapshot: {data: {workspace: {cdrVersionId: '1'}}}
-          }
-        },
       ],
     })
       .compileComponents();
--- a/ui/src/app/views/concept-homepage/component.spec.ts
+++ b/ui/src/app/views/concept-homepage/component.spec.ts
@@ -2,11 +2,11 @@ import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing'
 import {FormsModule, ReactiveFormsModule} from '@angular/forms';
 import {By} from '@angular/platform-browser';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
-import {ActivatedRoute} from '@angular/router';
 import {RouterTestingModule} from '@angular/router/testing';
 import {ClarityModule} from '@clr/angular';
 
 import {HighlightSearchComponent} from 'app/highlight-search/highlight-search.component';
+import {urlParamsStore} from 'app/utils/navigation';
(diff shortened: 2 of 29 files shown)


ui/src/app/cohort-search/demographics/demographics.component.spec.ts
```diff
@@ -49,12 +48,6 @@ describe('DemographicsComponent', () => {
...
         {provide: CohortBuilderService, useValue: {}},
         {provide: CohortSearchActions, useValue: new MockActions()},
-        {
-          provide: ActivatedRoute,
-          useValue: {
-            snapshot: {data: {workspace: {cdrVersionId: '1'}}}
-          }
-        },
       ],
     })
```
blrubenstein: Do you need to push to the `urlParamsStore` here?

ui/src/app/views/concept-set-details/component.spec.ts
```diff
@@ -3,12 +3,13 @@ import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing'
 import {FormsModule, ReactiveFormsModule} from '@angular/forms';
 import {By} from '@angular/platform-browser';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
-import {ActivatedRoute, Router} from '@angular/router';
+import {ActivatedRoute} from '@angular/router';
```
blrubenstein: Can we also pull activated route? Or is that needed for the factory? Are you planning to handle that in a follow up?

ui/src/app/views/concept-homepage/component.spec.ts
```diff
@@ -93,6 +70,10 @@ describe('ConceptHomepageComponent', () => {
         // This finishes the page reloading.
         updateAndTick(fixture);
       });
+    urlParamsStore.next({
+      ns: WorkspaceStubVariables.DEFAULT_WORKSPACE_NS,
+      wsid: WorkspaceStubVariables.DEFAULT_WORKSPACE_ID
+    });
```
blrubenstein: Do you need to do this here:
```
currentWorkspaceStore.next({
      ...WorkspacesServiceStub.stubWorkspace(),
      accessLevel: WorkspaceAccessLevel.OWNER,
    });
```

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -125,16 +122,15 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit(): void {
-    this.wsNamespace = this.route.snapshot.params['ns'];
-    this.wsId = this.route.snapshot.params['wsid'];
-    this.creating = this.route.snapshot.queryParams['creating'] || false;
-    this.playground = (this.route.snapshot.queryParams['playgroundMode'] === 'true');
-    this.jupyterLabMode = (this.route.snapshot.queryParams['jupyterLabMode'] === 'true');
+    const {ns, wsid} = urlParamsStore.getValue();
+    const {creating, playgroundMode, jupyterLabMode, kernelType} = queryParamsStore.getValue();
```
blrubenstein: Do you need to do the Kernels conversion as we saw below? `Kernels[kernelType]`

ui/src/app/cohort-search/demographics/demographics.component.spec.ts
UrsaStutsman: I'm curious about this too...

panentheos: It's not necessary in all cases, since these tests don't always hit the code paths that actually read this data. Since this particular test code will be going away soon, I was just aiming to get things passing for now.

ui/src/app/views/concept-set-details/component.spec.ts
panentheos: There's still one piece of data being pulled from the route in this component, so this has to stay for now.

ui/src/app/views/notebook-redirect/component.ts
panentheos: `this.kernelType` isn't actually used, so I got rid of it.

ui/src/app/views/concept-homepage/component.spec.ts
panentheos: This was extraneous data in the stub, the component doesn't actually use it.

PR #1773, opened 2019-02-11
[risk=no][RW-1888] convert profile page

zarsky-broad:


ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+});
+
+const labelSubstitutions = {
+  givenName: 'firstName',
+  familyName: 'lastName',
+  currentPosition: 'currentPosition',
+  organization: 'organization',
+  areaOfResearch: 'currentResearch'
+};
+const required = {presence: {allowEmpty: false}};
```
calbach: Can these validators be typed with the validator.js type definitions? I thought this was something totally custom on my first pass and a type would have made this obvious (and provide some basic compile-time checks)

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+const validators = {
+  firstName: {...required, ...notTooLong(80)},
+  lastName: {...required, ...notTooLong(80)},
+  currentPosition: {...required, ...notTooLong(255)},
+  organization: {...required, ...notTooLong(255)},
+  currentResearch: required,
+};
+
+export const ProfilePageReact = withUserProfile()
+(class ProfilePageReactComponent extends React.Component<
```
calbach: nit: alternate newline placement? I'm not a big fan of separating a method expression and its `(` for an invocation by newline. e.g. bring `withUserProfile()` down a line or move the paren up a line?

I realize it's a tricky balance to avoid indenting the whole class.

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+  { profileState: {profile: Profile, reload: Function} },
+  { profileEdits: Profile, updating: boolean }
+> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      profileEdits: props.profileState.profile || {},
```
calbach: Should this somehow be shallow-copied across? Or do we just hold the assumption that state will never be directly mutated?

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+      currentPosition,
+      organization,
+      currentResearch: areaOfResearch
+    }, validators);
+
+    const makeProfileInput = ({title, valueKey, isLong = false, ...props}) => {
+      const errorText = profile && errors && errors[labelSubstitutions[valueKey]];
+
+      const inputProps = {
+        value: fp.get(valueKey, profileEdits) || '',
```
calbach: fp.get vs `profileEdits[valueKey]`: is this purely style preference or is there something else happening here?

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+              />
+            </Clickable>
+            <div style={{border: `1px solid ${colors.gray[4]}`, flex: 1}}/>
+          </div>
+          <div style={{marginTop: 100}}>
+            <Button
+              type='purpleSecondary'
+              onClick={() => this.setState({profileEdits: profile})}
+            >
+              Discard Changes
```
calbach: The mocks says "Cancel"

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,314 @@
...
+            </Button>
+            <TooltipTrigger
+              side='top'
+              content={!!errors && 'You must correct errors before saving.'}
+            >
+              <Button
+                type='purplePrimary'
+                style={{marginLeft: 40}}
+                onClick={() => this.saveProfile()}
+                disabled={!!errors}
```
calbach: Should this have a different treatment when you haven't made any changes? e.g. disabled

ui/src/app/views/profile-page/component.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  const component = () => {
+    return mount(<ProfilePageReact/>);
+  };
+
+  beforeEach(() => {
+    userProfileStore.next({profile});
+  });
+
+  it('should render the profile', () => {
```
calbach: Would be good if there were test coverage of (1) validation behavior and (2) saving the profile.

ui/src/app/components/buttons.tsx
```diff
@@ -14,6 +14,17 @@ export const styles = {
     overflow: 'hidden', whiteSpace: 'nowrap', textOverflow: 'ellipsis',
     userSelect: 'none',
     margin: 0, padding: '0rem 0.77rem',
+  },
+  baseNeue: {
```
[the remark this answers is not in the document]
zarsky-broad: Heh, it does mean new, as used in things like Helvetica Neue (the font). I didn't just want to call it 'new' when not everybody may wish to switch to it yet, but there's an argument that this is needlessly confusing.

ui/src/app/views/profile-page/component.tsx
zarsky-broad: All updates are using `fp`, since they're deeper, so it's never mutated.

ui/src/app/views/profile-page/component.tsx
zarsky-broad: This allows deeper access using dot notation, see the institutional affiliations for examples.

ui/src/app/views/profile-page/component.tsx
zarsky-broad: Lou and I talked, we decided there's no real reason for a separate edit mode, so things are always editable. We decided to change this label as such.

ui/src/app/views/profile-page/component.spec.tsx
zarsky-broad: Talked with @blrubenstein about this.
For validation behavior, we thought that would be functionally just testing the validation library, which is unnecessary.
For saving, that would really just be testing the profile service, which is also tested independently, right?

Really, there's very little logic in this component.

ui/src/app/views/profile-page/component.tsx
zarsky-broad: I added a comment before them instead, as the types for this library are unclear to me.

ui/src/app/views/profile-page/component.spec.tsx
gjuggler: Personally I think we're making a mistake by removing all but a smoke test. There is definitely logic in components like this, and in the long term our velocity & stability will suffer if we don't encode core user expectations somewhere.

(That said, I don't think 100% of the existing tests were worth reproducing. More important / fragile are flows that make async calls and/or involve user interactions.)

I'm writing up something more concrete on the topic to share with the team for discussion.

zarsky-broad: I'm good with writing more tests, I just don't know what tests to write. Any suggestions?

ui/src/app/views/profile-page/component.tsx
```diff
@@ -0,0 +1,317 @@
...
+          </div>
+        </div>
+
+        <div style={{flex: '0 0 420px'}}>
+          <div style={styles.box}>
+            <div style={{...styles.h1, marginBottom: 12}}>
+              All of Us Training
+              <TooltipTrigger
+                side='left'
+                content=''
```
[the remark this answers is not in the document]
zarsky-broad: We don't have text for it yet.

ui/src/app/components/icons.tsx
```diff
@@ -21,8 +21,8 @@ export const styles = {
   }
 };
 
-export const ClrIcon = props => {
-  return React.createElement('clr-icon', props);
+export const ClrIcon = ({className = '', ...props}) => {
+  return React.createElement('clr-icon', {class: className, ...props});
```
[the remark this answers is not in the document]
zarsky-broad: I don't think any were using it yet, but they should be unaffected due to the spread merge.

ui/src/app/views/profile-page/component.spec.tsx
gjuggler: For a crud-type page component I'd usually go with a happy-path round-trip test:
 1. Render
 2. Find an input component, edit a value
 3. Trigger click on "submit", ensure the expected interface calls are made (looks like 1 call to updateProfile, and 1 call to props.reload)

I'd also scrutinize L83 a bit – there's no comment indicating what impact this has on the component, but if it's important it might be worth encoding in a test (more for documentation than actual logic)

The adding / removing of institutional affiliations would be lower-priority but useful. (And given the way we're using fp.set, we're not getting benefit from Typescript checking to protect against a typo in the field names.)

Finally, much lower priority but the user interaction to refresh the training completion state is also a bit of round-trip logic that could be checked... but since that's not the main purpose of this component that seems less worrisome if it breaks.

calbach: +1 to what Greg said. RE validation: I agree we don't need exhaustive testing here (i.e. re-encoding the validator configs you set up), but at least one test with an invalid input to verify that validation is wired up and working properly in the component seems like a good idea to me.

ui/src/app/components/buttons.tsx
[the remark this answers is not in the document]
calbach: Maybe others are more familiar, but I think "new" might be easier (and/or a comment advising when we should use this, or saying that all buttons should migrate to this newer style).

Also, just to clarify my understanding, is there a reason someone would not want to switch to this? This matches the target styling per Lou's mocks - right?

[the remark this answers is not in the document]
zarsky-broad: Yup. I guess new is reasonable, with the assumption that eventually we can switch it and then find-and-replace.

ui/src/app/views/profile-page/component.spec.tsx
zarsky-broad: Better now?

gjuggler: I'm happy – that's a nice example of a very concise test with user interaction & async service calls.

zarsky-broad: Thank @blrubenstein, their influence here is significant

PR #1777, opened 2019-02-12
[risk=no] bind react callbacks once

panentheos:
This addresses a short-term performance issue with rendering on the Angular/React boundary. Binding callbacks inline (while correct) creates a new function every time, which counts as a changed value. Because Angular updates on every mouse movement, this results in a lot of extra re-renders of react components. In addition to being a potential perf issue, it was interfering with the timing of `onChange` on selects.

Binding in the constructor creates a stable value which prevents these extra re-renders. This should be the preferred approach while we're in the transition. (Note that within react, inline binding via arrow functions is fine since there aren't nearly as many re-renders happening)

Also removes an unused html file.


panentheos:
The discovery came from debugging why selects were't working properly. The extra renders were overwriting the new value before the `onChange` handler could pick it up.

PR #1779, opened 2019-02-12
[RW-2012][Risk=moderate] Refactor user admin table and allow disabling

blrubenstein:


api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
```diff
@@ -40,7 +40,7 @@
   }
 
   @Override
-  @AuthorityRequired({Authority.MANAGE_GROUP})
+  @AuthorityRequired({Authority.REVIEW_ID_VERIFICATION})
```
gjuggler: I'm not quite sure why this change was made. It seems like disabling user should maybe be a distinct authority altogether? Maybe we should leave as-is and create a separate follow-up ticket?

Also: as I went looking for some clearer description of what each of these authorities allows, I could't find one. If you're willing to clean things up a bit while you're here, having a one-line sentence describing each authority level in o.p.w.m.Authority.java seems like a reasonably canonical place to document these enums.

(Also also... I'm noticing an issue with the disableUser method itself – the groupName string isn't actually needed anymore, and we should be logging the admin user action with the true / false disabled value rather than the access level. I think this wasn't fully cleaned up when we refactored the access levels. I'm happy to create a small follow-up PR for that if you'd rather not get involved :P)

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -72,6 +72,7 @@ public Profile getProfile(User user) {
     profile.setAreaOfResearch(user.getAreaOfResearch());
     profile.setRequestedIdVerification(user.getRequestedIdVerification());
     profile.setTwoFactorEnabled(user.getTwoFactorEnabled());
+    profile.setDisabled(user.getDisabled());
```
gjuggler: (SIde note) Ahh, the fact that we have to do stuff like this for the Profile data model still bugs me. It's a code smell that maybe we're not structuring things right. Perhaps the Profile object should contain a full User object inside of it, rather than a shallow copy of all of these fields...

Cleanup for a rainy day sometime.

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -22,6 +22,9 @@
   @Query("SELECT user FROM User user WHERE user.idVerificationIsValid IS NULL OR user.idVerificationIsValid = false")
   List<User> findUserNotValidated();
 
+  @Query("SELECT user FROM User user")
```
gjuggler: Nice – I was about to add this for my GSuite backfill PR as well.

ui/src/app/views/admin-user/component.ts
```diff
@@ -21,19 +22,31 @@ const verificationSortMap = {
...
+    this.loadProfiles();
+  }
+
+  loadProfiles(): void {
+    this.contentLoaded = false;
     this.profileService.getIdVerificationsForReview()
       .subscribe(
         profilesResp => {
           this.profiles = this.sortProfileList(profilesResp.profileList);
+          this.profiles.forEach((profile) => {
```
gjuggler: remove debug code?

ui/src/app/views/admin-user/component.ts
```diff
@@ -50,6 +63,13 @@ export class AdminReviewIdVerificationComponent implements OnInit {
     }
   }
 
+  disableUser(disable: boolean, profile: Profile): void {
+    this.authDomainService.disableUser('registered',
```
gjuggler: (Side note) Oof.. it's not great naming for our API method "disableUser" to actually be for disabling *or* enabling the user. This led me astray for a minute, and I thought we needed a separate method for enabling users... until I realized this is just an updateUserDisabledState method in disguise :)

I suppose it would be a pain to rename it at this point, but small mental hurdles like this can add up quickly & make it gummy to work in a codebase. Maybe I'll take a tech debt item at some point and work through some of these.

ui/src/app/views/signed-in/component.ts
```diff
@@ -163,8 +163,8 @@ export class SignedInComponent implements OnInit, OnDestroy {
     return this.locationService.path() === '/admin/review-workspace';
   }
 
-  get reviewIdActive(): boolean {
-    return this.locationService.path() === '/admin/review-id-verification';
+  get userAdminActive(): boolean {
+    return this.locationService.path() === '/admin/user';
```
gjuggler: nit: maybe admin/users with the plural is slightly better? I won't argue if you keep as-is.

ui/src/app/views/admin-user/component.html
```diff
@@ -9,13 +10,21 @@ <h2>Review ID Verifications</h2>
...
       <clr-dg-cell>{{profile.idVerificationRequestTime | date: 'medium'}}</clr-dg-cell>
       <clr-dg-cell>
-        <button class="btn" (click)="setIdVerificationStatus(profile,'VERIFIED')">Approve</button>
-        <button class="btn" (click)="setIdVerificationStatus(profile,'REJECTED')">Reject</button>
+        <div *ngIf="profile.idVerificationStatus !== IdVerificationStatus.VERIFIED">
+          <button class="btn"
+                  (click)="setIdVerificationStatus(profile,'VERIFIED')">Approve</button>
+          <button class="btn" (click)="setIdVerificationStatus(profile,'REJECTED')">Reject</button>
+        </div>
+        <div *ngIf="profile.idVerificationStatus === IdVerificationStatus.VERIFIED">
```
gjuggler: Thinking aloud on how we could handle this if / when the user-disable is a separate authority... do we currently have client-side access to the authority list for the current user? If not, that might be a nice next step to increase the flexibility of this page.

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
blrubenstein: I made this because I think we are moving more towards a place where we have user admins, and making this a distinct authority adds more overhead where I don't think it is necessary after our dataAccessLevel State Machine refactor.

The Authority.java file is autoGenerated, so we can't put a comment there. I can put a comment in the swagger though.

Agreed that the groupName isn't needed for disableUser. I think it would be valuable to pull it out, but didn't want to tackle that within the scope of this PR.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
blrubenstein: Agreed. I really hate that we do all this copying. And it led to a problem where we did have a disabled field on profile, but it wasn't actually being copied so I was just confused. 👍 to fixing it someday.

ui/src/app/views/admin-user/component.ts
blrubenstein: I... think it does make sense to rename it and refactor away group name. Let me do a quick look into it to see if this is actually used anywhere.

ui/src/app/views/admin-user/component.html
blrubenstein: No, and I don't really want to expose authorities clientside if we can avoid it? It might make sense, but something about it worries me. I will think through that while doing the refactor.

gjuggler: If we move towards a single "administer users" authority then I agree it won't be necessary... but if things develop in a way where "devs" should be able to perform more advanced administration than "support team" for some reason, we may end up wanting to pass this down to the client for a better user experience.

I wonder what the concerns would be? If all backend API calls are already protected separately, the only risk would be that someone could discover what actions are available to super-admins if they (a) guess an authority name and (b) somehow monkey-patch it into their client.

Maybe worthy of a tech talk topic?

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
gjuggler: Gotcha. So the idea would be that REVIEW_ID_VERIFICATION --> USER_ADMIN in an ideal world? I wonder how easily we can rename these enums... are they stored in the database as strings or integer values?

Maybe the safest alternative would be to create a brand-new authority w/ the right name, and migrate folks over as needed. More work though :\

Sorry I missed that the file is auto-gen. Swagger makes sense.

blrubenstein: I'll check with CH, if I recall correctly I think he did some work around enforcing/making sure that the enums don't change out from under us/are deterministic based on ordering.

ui/src/app/views/signed-in/component.ts
blrubenstein: I think it is the admin user page. I like the authority/page being called `User Admin`, so want the path to reflect that. If we find out later, happy changing it.

ui/src/app/views/admin-user/component.html
blrubenstein: Sure! Lets throw it in a tech talk.

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
blrubenstein: Follow-up tickets here:
https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2130&epics=visible
https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2131&epics=visible

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
blrubenstein: Follow up ticket here: 
https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2132&epics=visible

ui/src/app/views/admin-user/component.ts
blrubenstein: Follow up in this ticket:
https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2131&epics=visible

blrubenstein:
@gjuggler I am going to wait until after the tech talk to merge this.

blrubenstein:
screen shot 2019-02-13 at 4 08 05 pm
screen shot 2019-02-13 at 4 03 38 pm

PR #1781, opened 2019-02-13
[RW-2125][Risk=no] Add case insensitivity

blrubenstein:
@calbach this should fix future accounts. I am going to work out the backfill script as soon as I can.


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -406,6 +406,7 @@ private User initializeUserIfNeeded() {
   @Override
   public ResponseEntity<Profile> createAccount(CreateAccountRequest request) {
     verifyInvitationKey(request.getInvitationKey());
+    request.getProfile().setUsername(request.getProfile().getUsername().toLowerCase());
```
calbach: This is unrelated to your PR but something I just noticed - could you move `validateProfileFields()` directly after this line? We shouldn't be creating a gsuite user before validation

PR #1782, opened 2019-02-13
[RW-1887][risk=no] Homepage react conversion

jacmrob:


ui/src/app/views/homepage/component.tsx
```diff
@@ -341,72 +287,218 @@ export class HomepageComponent implements OnInit, OnDestroy {
...
+          <div style={homepageStyles.singleCard}>
+            {accessTasksLoaded ?
+              (accessTasksRemaining ?
+                (<WorkbenchAccessTasks eraCommonsLinked={eraCommonsLinked}
+                                       eraCommonsError={eraCommonsError}
+                                       trainingCompleted={trainingCompleted}/>
+                ) : (
+                  <div>
+                    <div style={{flexDirection: 'row', height: '17.47%', marginBottom: '0.5rem',
+                      justifyContent: 'flex-start', flexWrap: 'nowrap', display: 'flex'}}>
```
NehaBroad: nit: can this be moved to style tag. It would be better to limit inline styles to 3.

ui/src/app/views/homepage/component.tsx
```diff
@@ -186,153 +160,125 @@ export class WorkbenchAccessTasks extends
...
+  },
+  logo: {
+    top: '1rem', left: '6.25rem', height: '3.5rem', width: '7rem',
+    position: 'relative', lineHeight: '85px'
+  },
+  bottomLinks: {
+    color: '#9B9B9B', fontSize: '0.7rem', height: '1rem', left: '5.5rem',
+    top: '2rem', marginLeft: '2.5rem', position: 'relative', fontWeight: 400
+  },
+  addCard: {
```
NehaBroad: Question: do you think it should be moved to components/card.tsx as i can see this being used at other places as well or is it specific only to homepage?

ui/src/testing/stubs/profile-api-stub.ts
```diff
@@ -7,24 +7,32 @@ import {
   ProfileApi
 } from 'generated/fetch';
 
+export class ProfileStubVariables {
+  static PROFILE_STUB = <Profile>{
+    username: 'testers',
+    contactEmail: 'tester@mactesterson.edu🀓⚚><script>alert("hello");</script>',
```
NehaBroad: nit: junk characters

NehaBroad:
Also i just notice that the last line of footer link is being cropped of: 
screen shot 2019-02-15 at 11 57 10 am

ui/src/app/views/homepage/component.tsx
jacmrob: will do

jacmrob:
@NehaBroad those links actually scroll and resize with the page.  if you mouse over them you should be able to scroll down

ui/src/app/views/homepage/component.tsx
jacmrob: so there already is a `CardButton` element that has some default styles (in buttons.tsx).  i think some of the styles are custom to this specific card (where it sits on the homepage) but some could be moved out commonly.  In that case though I think we'd want it to be an "AddCardButton" built off of `CardButton` since the color/text size/plus button icon are all specific to the add function (we see them on the list pages).  I will make that change now.

jacmrob: scratch that, i think the common styles should be in `CardButton` itself

ui/src/app/views/homepage/component.spec.tsx
```diff
@@ -38,3 +47,93 @@ describe('WorkbenchAccessTasks', () => {
...
+
+  it('should not show the era commons linking page if user has an nih username', async () => {
+    const newProfile = {
+        ...profile,
+      linkedNihUsername: 'test'
+    };
+    userProfileStore.next({profile: newProfile as unknown as Profile, reload});
+    const wrapper = component();
+    await waitOneTickAndUpdate(wrapper);
+    expect(wrapper.find('[data-test-id="Login"]').exists()).toBeFalsy();
```
blrubenstein: Nit: I would use lowercase values for data test ids

ui/src/app/views/homepage/component.spec.tsx
```diff
@@ -38,3 +47,93 @@ describe('WorkbenchAccessTasks', () => {
...
+  });
+
```
blrubenstein: Nit: Extra space

ui/src/app/components/buttons.tsx
```diff
@@ -137,6 +137,8 @@ export const MenuItem = ({icon, tooltip = '', disabled = false, children, ...pro
 const cardButtonStyle = {
   style: {
     ...cardStyles.card,
+    alignItems: 'flex-start', alignContent: 'left',
+    marginTop: '1.9rem', marginRight: '106px', cursor: 'pointer',
```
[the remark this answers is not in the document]
jacmrob: i think it is - let me check and remove if so

PR #1787, opened 2019-02-14
[RW-1840][RISK=LOW] Offline sync of compliance training status

NehaBroad:
Create a cron job to run every day that which will update moodle information (moodle Id, training completion time) for all the user stored in database.


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -64,7 +64,6 @@
 import org.pmiops.workbench.model.UsernameTakenResponse;
 import org.pmiops.workbench.moodle.ApiException;
 import org.pmiops.workbench.compliance.ComplianceService;
```
gjuggler: ComplianceService can be ripped out of this class now, right? Or am I missing somewhere it's used?

NehaBroad: Yes, it should not be ref here now. Have updated the file

PR #1788, opened 2019-02-14
[RW-2004][RISK=LOW] Require higher minimum length for workbench usernames

NehaBroad:
1) Update UI to show error message if username is less than 3 characters
2) Update profile api
3) Update tooltip
screen shot 2019-02-14 at 2 14 29 pm

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -415,7 +415,7 @@ public class ProfileController implements ProfileApiDelegate {
     verifyInvitationKey(request.getInvitationKey());
     String userName = request.getProfile().getUsername();
     if(userName == null || userName.length() < 3 || userName.length() > 64 )
-      throw new BadRequestException("Username should be of at least 3 characters and not more than 64 characters");
+      throw new BadRequestException("Username should be at least 3 characters and not more than 64 characters");
     request.getProfile().setUsername(request.getProfile().getUsername().toLowerCase());
     validateProfileFields(request.getProfile());
     com.google.api.services.admin.directory.model.User googleUser =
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -413,6 +413,9 @@ private User initializeUserIfNeeded() {
   @Override
   public ResponseEntity<Profile> createAccount(CreateAccountRequest request) {
     verifyInvitationKey(request.getInvitationKey());
+    String userName = request.getProfile().getUsername();
+    if(userName.isEmpty() || userName.length() < 3 )
```
blrubenstein: Do we need the isEmpty anymore? Or will length just return 0?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -413,6 +413,9 @@ private User initializeUserIfNeeded() {
...
+      throw new BadRequestException("Username should be of atleast 3 characters");
```
blrubenstein: Nit: Space between `at` and `least`

ui/src/app/views/account-creation/component.spec.tsx
```diff
@@ -86,6 +86,17 @@ it('should handle username validity contains special chars', () => {
...
+  expect(wrapper.exists('#username')).toBeTruthy();
+  expect(wrapper.exists('#usernameError')).toBeFalsy();
+  // if username is long (not too long) but has a mismatch at end
+  let testInput = fp.repeat(50, 'abc');
+  testInput = testInput + ' abc';
+  wrapper.find('input#username').simulate('change', {target: {value: testInput}});
+  expect(wrapper.exists('#usernameError')).toBeTruthy();
+});
+
+it('should handle username validity length less than 3 characters', () => {
```
blrubenstein: This test looks like it actually puts a lot of characters in there? (The fp.repeat call). Can you verify this does what we want it to?

PR #1789, opened 2019-02-14
WIP Create Concept Set Modal

blrubenstein:
If anyone wants to pick this up, this is from our react boot camp. I am in the middle of other stuff and want to wrap that up first.


jacmrob:
Grabbing this to finish it off

ui/src/app/views/conceptset-create-modal/component.tsx
```diff
@@ -0,0 +1,122 @@
...
+      const {name, description, domain} = this.state;
+      const {workspace} = this.props;
+      const concept: ConceptSet = {
+        name: name,
+        description: description,
+        domain: domain
+      };
+      const request: CreateConceptSetRequest = {
+        conceptSet: concept
+      };
```
panentheos: FYI, typescript can perform type inference in straightforward cases like this, so it will be able to verify the shape of the request object without being explicit. This could be shortened to e.g.
```
const request = {conceptSet: {name, description, domain}};
```

ui/src/testing/stubs/concepts-api-stub.ts
```diff
@@ -0,0 +1,101 @@
...
+    super(undefined, undefined, (..._: any[]) => { throw Error('cannot fetch in tests'); });
+
+    this.concepts = ConceptStubVariables.STUB_CONCEPTS;
+  }
+
+  public getDomainInfo(
+    workspaceNamespace: string, workspaceId: string): Promise<DomainInfoResponse> {
+    return new Promise<DomainInfoResponse>(resolve => {
+      resolve({items: DomainStubVariables.STUB_DOMAINS});
+    });
```
panentheos: There's no need to use the explicit Promise constructor if you're just resolving immediately. This is equivalent:
```
return Promise.resolve({items: DomainStubVariables.STUB_DOMAINS});
```

ui/src/app/views/conceptset-create-modal/component.tsx
```diff
@@ -0,0 +1,122 @@
...
+      this.setState({saving: true});
+      await conceptSetsApi().createConceptSet(workspace.namespace, workspace.id, request);
+      this.props.onCreate();
+      this.props.onClose();
```
panentheos: nit: Instead of calling two callbacks together, it's usually a bit nicer to just call one (onCreate) and have the parent reset its open state as part of that. Most of the time it doesn't matter, but it's helpful if the parent wants to have a slightly non-standard flow.

ui/src/app/views/conceptset-create-modal/component.tsx
```diff
@@ -0,0 +1,122 @@
...
+    } catch (e) {
+      console.error(e);
+      this.setState({savingError: e});
```
panentheos: I think this will end up saving the response object itself, which probably isn't what we want. Since this is being used as a boolean, perhaps it should just be one.

ui/src/app/views/concept-set-list/component.html
```diff
@@ -43,5 +43,10 @@
     </div>
   </div>
 </app-top-box>
+<app-create-concept-modal *ngIf="isCreateModalOpen"
+                          [onCreate]="loadConceptSets.bind(this)"
```
panentheos: We should bind these once in the constructor, rather than inline here (perf reasons).

ui/src/app/views/conceptset-create-modal/component.tsx
jacmrob: I didn't want to put it in the parent method bc that method gets called in other places where the modal is not being used

PR #1790, opened 2019-02-14
[RW-2131][Risk=low] Srubenst/refactor disabled user

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
@@ -41,13 +41,13 @@ public class AuthDomainController implements AuthDomainApiDelegate {
 
   @Override
   @AuthorityRequired({Authority.REVIEW_ID_VERIFICATION})
-  public ResponseEntity<Void> updateUserDisabledStatus(String groupName, UpdateUserDisabledRequest request) {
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
```diff
@@ -41,13 +41,13 @@
...
   @AuthorityRequired({Authority.REVIEW_ID_VERIFICATION})
-  public ResponseEntity<Void> disableUser(String groupName, AuthDomainDisableUserRequest request) {
+  public ResponseEntity<Void> updateUserDisabledStatus(UpdateUserDisabledRequest request) {
     User user = userDao.findUserByEmail(request.getEmail());
     DataAccessLevel previousAccess = user.getDataAccessLevelEnum();
     User updatedUser = userService.setDisabledStatus(user.getUserId(), request.getDisabled());
     userService.logAdminUserAction(
         user.getUserId(),
-        "user access to  " + groupName + " domain",
+        "updated user access to registered domain",
```
gjuggler: I think the right thing to log here is the disabled / enabled status of the request, with text like "updated user disabled state."

blrubenstein: fixing that and the tests now!

PR #1791, opened 2019-02-15
[RW-1948][risk=low] Implement initial BigQuery OMOP -> Elastic indexer

calbach:
Ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-1948
Design: https://docs.google.com/document/d/1N_TDTOi-moTH6wrXn1Ix4dwUlw4j8GT9OsL9yXYXYmY/edit

# Overview
Command-line tool for indexing an OMOP BigQuery dataset into Elasticsearch. Executes the following pipeline:

- Create (and optionally replace) an existing Elasticsearch index for this CDR.
- Query BigQuery to produce JSON documents: the output of the query is the exact format we define for Elasticsearch (using a repeated STRUCT for repeated nested documents).
- [Optional] Export these results to a temporary table, and export that to GCS. Slower, but necessary for larger datasets (e.g. the entire Synthetic CDR).
- Bulk ingest the JSON documents into Elasticsearch (either inserts or updates).

The indexer can optionally be run on a subsample of participants in the CDR for faster test iterations and local development. The full index will be 10-100GB, so it is not advisable to run this locally against the entire synthetic CDR.

# Limitations
- Currently only indexes the person and condition_occurrence table.
- `./project.rb create-local-es-index` interface is limited and hardcoded to only work with the synthetic CDR -> local

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1176,6 +1176,7 @@ Common.register_command({
 
 def create_local_es_index(cmd_name, *args)
   op = WbOptionsParser.new(cmd_name, args)
+  # TODO(RW-2213): Generalize this subsampling approach for all local development work.
   op.opts.inverse_prob = 1000
   op.add_option(
       "--participant-inclusion-inverse-prob [DENOMINATOR]",
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1188,24 +1188,22 @@ def create_local_es_index(cmd_name, *args)
     exec(*(%W{docker-compose run --rm es-scripts ./project.rb #{cmd_name}} + args))
   end
 
-  Dir.chdir("tools") do
-    common = Common.new
-    # TODO(calbach): Parameterize most of these flags. For now this is hardcoded
-    # to work against the synthetic CDR into a local ES (using test Workbench).
-    create_flags = ([
-      ['--query-project-id', 'all-of-us-ehr-dev'],
-      ['--es-base-url', 'http://elastic:9200'],
-      ['--cdr-version', 'synth_r_2019_q1'],
-      ['--cdr-big-query-dataset', 'all-of-us-ehr-dev.synthetic_cdr20180606'],
-      ['--scratch-big-query-dataset', 'all-of-us-ehr-dev.workbench_elastic'],
-      ['--scratch-gcs-bucket', 'all-of-us-workbench-test-elastic-exports'],
-      ['--participant-inclusion-inverse-prob', op.opts.inverse_prob]
-    ].map { |kv| "#{kv[0]}=#{kv[1]}" } + [
-      '--delete-indices'
-      # Gradle args need to be single-quote wrapped.
-    ]).map { |f| "'#{f}'" }
-    common.run_inline %W{gradle elasticSearchIndexer -PappArgs=['create',#{create_flags.join(',')}]}
-  end
+  common = Common.new
+  # TODO(calbach): Parameterize most of these flags. For now this is hardcoded
+  # to work against the synthetic CDR into a local ES (using test Workbench).
+  create_flags = ([
+    ['--query-project-id', 'all-of-us-ehr-dev'],
+    ['--es-base-url', 'http://elastic:9200'],
+    ['--cdr-version', 'synth_r_2019_q1'],
+    ['--cdr-big-query-dataset', 'all-of-us-ehr-dev.synthetic_cdr20180606'],
+    ['--scratch-big-query-dataset', 'all-of-us-ehr-dev.workbench_elastic'],
+    ['--scratch-gcs-bucket', 'all-of-us-workbench-test-elastic-exports'],
+    ['--participant-inclusion-inverse-prob', op.opts.inverse_prob]
+  ].map { |kv| "#{kv[0]}=#{kv[1]}" } + [
+    '--delete-indices'
+    # Gradle args need to be single-quote wrapped.
+  ]).map { |f| "'#{f}'" }
+  common.run_inline %W{gradle elasticSearchIndexer -PappArgs=['create',#{create_flags.join(',')}]}
 end
 
 Common.register_command({
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1174,6 +1174,46 @@ Common.register_command({
   :fn => ->(*args) { list_clusters("list-clusters", *args) }
 })
 
+def create_local_es_index(cmd_name, *args)
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.inverse_prob = 10000
+  op.add_option(
+      "--participant-inclusion-inverse-prob [DENOMINATOR]",
+      ->(opts, v) { opts.inverse_prob = v},
+      "The inverse probabilty to index a participant, used to index a " +
+      "sample of participants. For example, 1000 would index ~1/1000 of participants in the " +
+      "target dataset. Defaults to 1 (all participants).")
+  op.parse.validate
+  unless Workbench.in_docker?
+    exec(*(%W{docker-compose run --rm es-scripts ./project.rb #{cmd_name}} + args))
+  end
+
+  Dir.chdir("tools") do
+    common = Common.new
+    # TODO(calbach): Parameterize most of these flags. For now this is hardcoded
+    # to work against the synthetic CDR into a local ES (using test Workbench).
+    create_flags = ([
+      ['--query-project-id', 'all-of-us-ehr-dev'],
+      ['--es-base-url', 'http://elastic:9200'],
+      ['--cdr-version', 'synth_r_2019_q1'],
+      ['--cdr-big-query-dataset', 'all-of-us-ehr-dev.synthetic_cdr20180606'],
+      ['--scratch-big-query-dataset', 'all-of-us-ehr-dev.workbench_elastic'],
+      ['--scratch-gcs-bucket', 'all-of-us-workbench-test-elastic-exports'],
+      ['--participant-inclusion-inverse-prob', op.opts.inverse_prob]
+    ].map { |kv| "#{kv[0]}=#{kv[1]}" } + [
+      '--delete-indices'
+      # Gradle args need to be single-quote wrapped.
+    ]).map { |f| "'#{f}'" }
+    common.run_inline %W{gradle elasticSearchIndexer -PappArgs=['create',#{create_flags.join(',')}]}
+  end
+end
+
+Common.register_command({
+  :invocation => "create-local-es-index",
+  :description => "Create Elasticsearch index",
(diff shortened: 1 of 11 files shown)


api/libproject/devstart.rb
```diff
@@ -1136,6 +1136,46 @@ def list_clusters(cmd_name, *args)
   :fn => ->(*args) { list_clusters("list-clusters", *args) }
 })
 
+def create_local_es_index(cmd_name, *args)
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.inverse_prob = 10000
```
gjuggler: nitpick: actually default to 1 before this goes in.

api/tools/build.gradle
```diff
@@ -52,6 +52,15 @@ task manageClusters(type: JavaExec) {
   }
 }
 
+task elasticSearchIndexer(type: JavaExec) {
+  classpath sourceSets.main.runtimeClasspath
+  main = "org.pmiops.workbench.tools.elastic.BigQueryOmopIndexer"
```
gjuggler: nitpick: would it be possible to aim for consistent naming of the task / command-line / Java class? The other tools were mostly good about this.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
```diff
@@ -0,0 +1,325 @@
...
+ *   necessary for larger datasets (e.g. the entire 1m participant Synthetic CDR).
+ * - Bulk ingest the JSON documents into Elasticsearch (either inserts or updates).
+ *
+ * This indexer can optionally be run on a subsample of participants in the CDR for faster test
+ * iterations and local development. The full index will be 10-100GB, so it is not advisable to run
+ * this locally against the entire synthetic CDR.
+ */
+public final class BigQueryOmopIndexer {
+
+  private static Option queryProjectIdOpt = Option.builder()
```
gjuggler: Would it be helpful / possible to add types to these options? It seems like the option API has a setType, but I have no idea if it works well and/or parses things as you'd expect. Booleans seem the most relevant.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/ElasticDocument.java
```diff
@@ -0,0 +1,175 @@
...
+  public int hashCode() {
+    return Objects.hash(id, source);
+  }
+
+  /**
+   * Converts a line of BigQuery results JSON to an Elasticsearch document.
+   */
+  public static ElasticDocument fromBigQueryJson(String line) throws IOException {
+    JsonObject doc = new JsonParser().parse(line).getAsJsonObject();
+    String id = doc.remove("_id").getAsString();
```
gjuggler: This code doesn't have tests, so liberally commenting important-but-obscure lines is more important. You'd commented on this issue (removing _id_ from the doc) in the other file, but it's warranted here too.

... but then again, it may also be worth adding basic unit tests on this class at some point. Seems reasonably straightforward to do.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
```diff
@@ -0,0 +1,325 @@
...
+
+    // Generate JSON documents from the CDR for Elasticsearch.
+    // Note: Here, BigQuery uses default credentials for the given environment.
+    BigQuery bq = BigQueryOptions.newBuilder()
+        .setProjectId(opts.getOptionValue(queryProjectIdOpt.getLongOpt()))
+        .build()
+        .getService();
+    String cdrDataset = opts.getOptionValue(cdrBigQueryDatasetOpt.getLongOpt());
+    int totalSampleSize = getPersonCount(bq, cdrDataset, inverseProb);
+    String personSQL = getPersonBigQuerySQL(cdrDataset, inverseProb);
```
freemabd1: can the count not be gotten from the result of personSql?

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
```diff
@@ -0,0 +1,325 @@
...
+        .replaceAll("\\{PERSON_ID_MOD\\}", Integer.toString(inverseProb));
+  }
+
+  private static RestHighLevelClient newClient(String esBaseUrl) throws MalformedURLException {
+    URL url = new URL(esBaseUrl);
+    return new RestHighLevelClient(RestClient.builder(
+        new HttpHost(url.getHost(), url.getPort(), url.getProtocol())));
+  }
+
+  @Bean
```
freemabd1: Do we need to expose this as a Spring Bean?

api/tools/src/main/resources/bigquery/es_person.sql
```diff
@@ -0,0 +1,56 @@
+SELECT
```
freemabd1: Is this sql going to be a template for all domains? If so should you parameterized anything that references a domain? or this can be done later as this code matures.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
calbach: Tried this out. It doesn't seem to do anything unless you use the `getParsedOptionValue()` method: https://commons.apache.org/proper/commons-cli/javadocs/api-1.3.1/org/apache/commons/cli/CommandLine.html

This API is kind of lame as it throws a parsing Exception and only supports returning an Object which I have to cast anyways, so I think it's probably clearer if I just parse it myself. Overall not really impressed with this package so far..

Also booleans are straightforward, they have no argument, so we just check for presence (there is no negation flag built in)

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
```diff
@@ -0,0 +1,325 @@
...
+ * - Bulk ingest the JSON documents into Elasticsearch (either inserts or updates).
+ *
+ * This indexer can optionally be run on a subsample of participants in the CDR for faster test
+ * iterations and local development. The full index will be 10-100GB, so it is not advisable to run
+ * this locally against the entire synthetic CDR.
+ */
+public final class BigQueryOmopIndexer {
+
+  private static Option queryProjectIdOpt = Option.builder()
+      .longOpt("query-project-id")
```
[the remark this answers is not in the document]
calbach: I'm effectively using this Option as the constant elsewhere in the file. I think this is probably overall lower boilerplate.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
calbach: Tested - yes, this is what triggers the SpringApplicationBuilder to actually execute something.

api/tools/src/main/resources/bigquery/es_person.sql
calbach: I'm planning to throw the whole mess into here in a subsequent PR as one giant query. I think each domain will have slightly different values being mapped into these structs.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
calbach: I'd assumed I couldn't get the total row count when I did an extraction to a table; I was wrong. Done - thanks.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/BigQueryOmopIndexer.java
```diff
@@ -0,0 +1,325 @@
...
+      .desc("GCP Cloud Project ID to run the extraction query in. Does not necessarily need " +
+          "to be the same as the CDR dataset's BigQuery project.")
+      .required()
+      .hasArg()
+      .build();
+  private static Option esBaseUrlOpt = Option.builder()
```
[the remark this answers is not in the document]
calbach: I think I prefer the bulider form since some of these need slightly different modifiers. I did a few iterations of this and I landed on this as an "OK" solution here. The thing I'd lose if I used that approach (if I'm understanding correctly), is the ability to reference the options throughout the code. I initially was retyping the flag strings, which was error-prone. Moving those strings to constants feels like higher boilerplate than this approach to me, though that might be wrong.

api/tools/src/main/java/org/pmiops/workbench/tools/elastic/ElasticDocument.java
calbach: Yes, I'm thinking about this. Unfortunately to write a meaningful test here, I'll need BigQuery and/or Elastic. It seems there is still no suitable fake for BigQuery locally aside from mocking out its exact behavior (a lot of work, and critical behavior we'd want to test). I'm thinking I might compromise by stubbing the GCS piece with golden data (saved BigQuery output), then running this tool via our "api-local-test" (runs on each master commit / PR), which starts up the API server, then tests. Another alternative would be to write an integration test, but I'd like some more immediate coverage here than "nightly".

Any thoughts?

freemabd1: This could be helpful: https://github.com/dadoonet/testcontainers-java-module-elasticsearch

calbach:
I started looking at adding tests here. I think I'll wait for @gjuggler 's CLI PR to go in and switch to his approach so I can avoid duplicating effort in setting up the test gradle deps here.

calbach:
Updated to use @gjuggler 's new gradle setup. I also refactored a few things to make it a bit more conducive to future test writing.

After mulling on this a bit, I think this tool would be best covered by an "integration" test - specifically a test harness which has an elastic docker image wired in via our docker-compose setup. I tried out the testcontainers stuff @freemabd1 linked - but it doesn't seem to be easily compatible with our current docker setup, e.g. for CircleCI they recommend doing something pretty custom: https://www.testcontainers.org/supported_docker_environment/continuous_integration/circle_ci/. I'm not yet sure whether that integration test would hit live BigQuery - but it might. As this will be a larger change, my preference would be to handle this as part of this ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-1957

Does this seem reasonable?

There's also a piece of work involved here which is figuring out how to incorporate tools tests into the gradle setup. I tried a couple of approaches without much immediate success.

api/libproject/devstart.rb
gjuggler: Makes sense – in that case, update the option description to match.

calbach: Oops, no wonder you asked for the change. Done.

PR #1795, opened 2019-02-15
[RW-1184][risk=no] Connect to moodle UI

jacmrob:


ui/src/app/views/homepage/component.tsx
```diff
@@ -105,6 +105,10 @@ export class WorkbenchAccessTasks extends
     window.location.assign(url);
   }
 
+  static redirectToTraining(): void {
+    window.location.assign('https://aoudev.nnlm.gov/static/data-researcher.html?saml=on');
```
NehaBroad: this is the dev url it should be https://aou.nnlm.gov/static/data-researcher.html?saml=on

gjuggler: We'll need an env variable for this. In prod the base URL will be https://aou.nnlm.gov/

PR #1797, opened 2019-02-15
[RW-1491][risk=no]Temporal Relationship UI with Undo

smazumd:
[RW-1491][risk=no]Temporal Relationship UI with Undo


ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -590,6 +661,10 @@ export class CohortSearchActions {
       entities.searchRequests[SR_ID][role] = data[role].map(group => {
         group.items = group.items.map(item => {
           item.searchParameters = item.searchParameters.map(param => {
+            group.mention = group.mention ? group.mention : '';
+            group.time = group.time ? group.time : '';
+            group.timeValue = group.timeValue ? group.timeValue : 0;
+            group.timeFrame = group.timeFrame ? group.timeFrame : '';
```
dolbeew: These temporal items should get set in the outer map so they only get set once. Here they are being set for each group item and each item parameter

dolbeew: We also need to set `temporalGroup` to 0 for each item if the group is not temporal

ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -119,6 +123,35 @@ export class CohortSearchActions {
     this.addId(newId);
     return newId;
   }
+  updateTemporal(flag: boolean, groupId: string, role: keyof SearchRequest) {
+    this._updatedTemporal(flag, groupId);
+    const group = getGroup(groupId)(this.state);
+    if (this.hasActiveItems(group)) {
```
dolbeew: This needs to check both temporalGroups for active items

ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -119,6 +123,35 @@ export class CohortSearchActions {
...
+    this._updateTemporalTime(time, groupId);
+    this.requestGroupCount(role, groupId);
+    this.requestTotalCount(groupId);
+  }
+
+  updateTemporalTimeValue(timeValue: any, groupId: string, role: keyof SearchRequest) {
+    this._updateTemporalTimeValue(timeValue, groupId);
+    this.requestGroupCount(role, groupId);
+    this.requestTotalCount(groupId);
+  }
```
dolbeew: We need to check both temporalGroups for active items in each of the above functions before making the calls for count. Probably need to make a separate function since we'll use it in several actions

ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -119,6 +123,35 @@ export class CohortSearchActions {
...
   }
+  updateTemporal(flag: boolean, groupId: string, role: keyof SearchRequest) {
+    this._updatedTemporal(flag, groupId);
+    const group = getGroup(groupId)(this.state);
+    if (this.hasActiveItems(group)) {
+      this.requestGroupCount(role, groupId);
+      this.requestTotalCount(groupId);
+    } else {
+      this.clearGroupCount(groupId);
+      this.clearTotalCount(groupId);
```
dolbeew: `groupId` not needed for this action

ui/src/app/cohort-search/redux/actions/service.ts
```diff
@@ -227,35 +260,51 @@ export class CohortSearchActions {
...
+        if (onlyChild) {
           if (groupItems.size === 1 && status === 'pending') {
             this.clearGroupCount(groupId);
           }
           if (this.otherGroupsWithActiveItems(groupId)) {
             this.requestTotalCount(groupId);
           } else {
             this.cancelTotalIfRequesting();
-            this.clearTotalCount();
+            this.clearTotalCount(groupId);
```
dolbeew: `groupId` not needed

ui/src/app/cohort-search/redux/actions/service.ts
smazumd: Done

ui/src/app/cohort-search/redux/actions/service.ts
smazumd: done

ui/src/app/cohort-search/redux/actions/service.ts
smazumd: done

PR #1801, opened 2019-02-19
[RW-2022][risk=no] Filter user responses on the current search

rushtong:
Addresses https://precisionmedicineinitiative.atlassian.net/browse/RW-2022

The approach this PR takes is to restrict the updates to the displayed list of users to happen only when the response from the server matches the query that UI actually made. Previously, if responses came out of order, the list of filtered users would not reflect the actual search query. This ensures that we're only updating the list for the correct query.


api/src/main/resources/workbench.yaml
```diff
@@ -2305,6 +2305,9 @@ definitions:
   UserResponse:
     type: object
     properties:
+      query:
```
calbach: Hmm.. I'm not sure I like this in principle - does this logically belong in the response? What's the rationale for adding this, but not adding the entirety of the request (present and future) into the response (or doing this on all our endpoints, for that matter)? e.g. imagine we added a second parameter here. I have some other suggestions below.

ui/src/app/views/workspace-share/component.ts
```diff
@@ -176,14 +176,18 @@ export class WorkspaceShareComponent implements OnInit {
...
     this.userService.user(this.searchTerm).subscribe((response) => {
-      this.autocompleteLoading = false;
       this.userResponse = response;
-      response.users = response.users.filter(user => {
-        return this.checkUnique(user.email, user.familyName, user.givenName);
-      });
-      this.autocompleteUsers = response.users.splice(0, 4);
-      if (this.autocompleteUsers.length === 0) {
-        this.autocompleteNoResults = true;
+      if (this.searchTerm === this.userResponse.query) {
```
calbach: This could simply be capture via closure above.

```
const searchTerm = this.searchTerm;
this.userService.user(searchTerm).sub(()=>{
  if (this.searchTerm !== searchTerm) return;
...
});
```

Alternatively, a more general solution:

```
const obs = this.userService.user(... => {
  if (this.pendingAutocomplete !== obs) return;
...
  this.pendingAutocomplete = null;
});
this.pendingAutocomplete = obs;
```

(also could replace "autocompleteLoading" with a helper to check null'ness on the observable).

Better would be to cancel the old obesrvable as well, but since we're moving to promises with the React changes I wouldn't touch that part.

api/src/main/resources/workbench.yaml
rushtong: The point of this is for the handler of the response to know what it was that was requested. For example, if there are two responses that arrive out of order, this allows the handler to know the full scope of the response, i.e. `This was searched on and this is what was found`. As it stands now, the response doesn't give the caller the full picture of what was requested. At the end of any series of requests that a caller might make (and whatever order they arrived in), the caller would be able to be able to distinguish between them.

I'm not sure I understand what you're referring to with `the entirety of the request (present and future)`. I don't think we're limited to any number of parameters here.

ui/src/app/views/workspace-share/component.ts
rushtong: Sorry, I don't follow your suggestions here. The point of comparing the last term the user searched with what is in the response allows the UI to perform an update IFF the response matches the search.

rushtong:
@UrsaStutsman - if you're following, I was looking at the ticket and thought the logic fix here would be clearer if it were separate from the react conversion. Hopefully that makes sense.

api/src/main/resources/workbench.yaml
calbach: The problem of responses coming back out of order is not unique to this endpoint. Clients have the ability to remember which request corresponds to which response, so there's no need to add API functionality to help them with this.

RE adding parameters, my concern was that the generalized version of this solution involved re-encoding the entire request into the response, which seems like not a great separation of concerns.

ui/src/app/views/workspace-share/component.ts
calbach: Yes, that is what my proposed code is doing. A client is able to store which request corresponds to which response either as an instance property or via a closure. Which part is not making sense?

rushtong: I don't see how this would work, but I'll spend some time with the approaches to see if I can grok them.

rushtong:
@calbach @jacmrob - I admittedly do not understand how this approach works, but PTAL

ui/src/app/views/workspace-share/component.ts
```diff
@@ -175,9 +175,11 @@ export class WorkspaceShareComponent implements OnInit {
       this.autocompleteNoResults = true;
       return;
     }
-    this.userService.user(this.searchTerm).subscribe((response) => {
-      this.autocompleteLoading = false;
+    const searchTerm = this.searchTerm;
```
calbach: This variable gets stored via a closure in the below function, copied-by-value so it retains whatever the query was at request time.

ui/src/app/views/workspace-share/component.ts
```diff
@@ -175,9 +175,11 @@ export class WorkspaceShareComponent implements OnInit {
...
+    this.userService.user(searchTerm).subscribe((response) => {
+      if (this.searchTerm !== searchTerm) { return; }
```
calbach: This code gets executed later when the API response comes back. At this point, `searchTerm` is unchanged as it refers to the value in the closure (and the same value we sent for the request), however `this.searchTerm` may be different if the user has input some different value.

calbach: opt_nit: I'd suggest either dropping the brackets or making this multiline. Single line with brackets is a bit cluttered in my opinion.

rushtong: Lint currently complains without the brackets.

calbach: OK - multiline then?

rushtong: 👍

rushtong: Thank you - this is helpful

PR #1803, opened 2019-02-19
[RW-2035][risk=no] Add a script to backfill GSuite custom fields & contact email

gjuggler:
The main goal of this PR was to create a script that does the following:

- Walk through each AoU user from the database,
- Find the equivalent user in GSuite and backfill the "Institution" and "Contact email" custom field

Unfortunately, I got pretty side-tracked by Gradle & Spring complications, since I needed to use a new dependency (DirectoryServiceImpl) that hadn't yet been used by any command-line tools.

After a fair bit of digging, I came up with a reasonably nice solution that will let us eventually get rid of the separate tools/build.gradle altogether, and have our command-line tools get built from the same Gradle config as our other code.

I also created a new Spring config file where we can set up tool-specific beans. (Context on this is that for some dependencies, our App Engine / WebMVC default config uses request-scoped beans and/or App Engine implementations, which need to be swapped out for command-line tools.)

Unfortunately it all ends up meaning that this PR is rather large and sprawling. Let me know if you'd strongly prefer I split something out into a separate piece.


gjuggler:
Note: I'm sending out for review, but there may be some CircleCI cleanup required. It gets a bit complicated, since some of these changes may affect other command-line tools. (And some of those tools aren't easily run locally, at least as far as I can tell.)

gjuggler:
Example output from a dry run against the test project:

$ ./project.rb backfill-gsuite-user-data --dry_run=true --project=all-of-us-workbench-test
...
INFO: DRY RUN:  Backfill complete. Updated 104, skipped 1, failed 23.

The 23 failed are from AoU users that don't have a corresponding GSuite account (e.g. the GSuite was deleted for some reason, without being removed from the AoU database).

The 1 skipped is gjordan@ who already has the correct custom field data.

api/libproject/devstart.rb
```diff
@@ -1006,6 +1006,52 @@ def update_user_registered_status(cmd_name, args)
...
+  op.opts.dry_run = true
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--dry_run=[dry_run]",
+      TrueClass,
+      ->(opts, v) { opts.dry_run = v},
+      "When true, print debug lines instead of performing writes. Defaults to true.")
+
+  print "Project is #{op.opts.project}\n"
```
calbach: debugging line?

api/libproject/devstart.rb
```diff
@@ -1006,6 +1006,52 @@ def update_user_registered_status(cmd_name, args)
...
+
+  # Create a cloud context and apply the DB connection variables to the environment.
+  # These will be read by Gradle and passed as Spring Boot properties to the command-line.
+  gcc = GcloudContextV2.new(op)
+  op.parse.validate
+  gcc.validate()
+
+  if op.opts.dry_run
+    print "DRY RUN -- CHANGES WILL NOT BE PERSISTED\n"
```
calbach: We've typically used `common.status` for these kinds of notifications. It has a slightly different color I guess to distinguish from stderr from other tools. May want to keep using that here for consistency at least (though I could imagine an argument for switching everything to `print`)

api/libproject/devstart.rb
```diff
@@ -1006,6 +1006,52 @@ def update_user_registered_status(cmd_name, args)
...
+  end
+
+  #ENV.update(read_db_vars(gcc))
```
calbach: nit: rm?

api/libproject/devstart.rb
```diff
@@ -1006,6 +1006,52 @@ def update_user_registered_status(cmd_name, args)
...
+  #ServiceAccountContext.new(op.opts.project).run do
+  with_cloud_proxy_and_db(gcc) do
+    #get_gsuite_admin_key(op.opts.project)
+
+    common = Common.new
+    common.run_inline %W{
+        gradle --info backfillGSuiteUserData
+       -PappArgs=[#{op.opts.dry_run}]}
+  end
+  #end
```
calbach: nit: rm commented out code lines?

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -156,22 +159,38 @@ public User createUser(String givenName, String familyName, String username, Str
     // case we want to pass it to other systems (e.g. Zendesk or Moodle) via SAML mapped fields.
     aouCustomFields.put(GSUITE_FIELD_CONTACT_EMAIL, contactEmail);
 
+    // In addition to the custom schema value, we store each user's contact email as a secondary
+    // email address with type "home". This makes it show up nicely in GSuite admin as the
+    // user's "Secondary email".
+      user.setEmails(Lists.newArrayList(
```
calbach: nit: -2sp (we should figure out the linting)

api/tools/src/main/java/org/pmiops/workbench/tools/BackfillGSuiteUserData.java
```diff
@@ -0,0 +1,131 @@
...
+ * Once the backfill is completed for all environments, this script may be deleted.
+ *
+ * TODO(RW-2032): Before this script goes away, but after recoveryEmail becomes available
+ * to the directory API, we should augment this script & run a backfill for recovery emails.
+ */
+@SpringBootApplication
+// Most of the "config" module contains Spring beans that we don't want. But the RetryConfig
+// is required for the DirectoryService class, so we specifically include only that one.
+//
+// TODO(gjuggler): find a better way to manage component scans like this... either cenralize
```
calbach: Probably we need to split out packages / folder structure and define clear relationships so we can easily pull in non-request-scoped stuff. I'm not sure whether that's something we can easily enforce with our current build tooling or if it would have to just be convention.

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
```diff
@@ -0,0 +1,76 @@
+package org.pmiops.workbench.tools;
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.http.apache.ApacheHttpTransport;
+import com.google.gson.Gson;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.db.dao.ConfigDao;
+import org.pmiops.workbench.db.model.Config;
+import org.springframework.context.annotation.*;
```
calbach: Should avoid `*` imports. I think you can switch this in Intellij (another thing to cover with the linter choice).

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
```diff
@@ -0,0 +1,76 @@
...
+   * Any command-line tool which loads this bean needs to be called from a project.rb command
+   * which is preceded with "get_gsuite_admin_key" to ensure the local key file si populated.
+   *
+   * @return
+   */
+  @Lazy
+  @Bean
+  GoogleCredential googleCredential() {
+    try {
+      String saKeyPath = "src/main/webapp/WEB-INF/gsuite-admin-sa.json";
```
calbach: hmm... I think this would really only work for the test environment. This file only exists after you've run dev-up (or possibly if you ran a build on a given environment, which would be unusual). We'd either need to `gsutil cp` into here before running the tool - or probably better somehow load the relevant Config for the appropriate environment, and grab the `gs://` path from there. Pulling from gs:// in here would be preferred, as we'd rather not pull the prod gsuite admin key file (for example) onto a dev workstation.

Alternatively, you could instead try to pass this gs:// path as a flag to your CLI, then somehow get it into here, but this might just be creating more work for the next config value someone adds (or CLI someone writes).

api/build.gradle
```diff
@@ -448,3 +463,32 @@ task incrementalHotSwap(type: IncrementalHotSwapTask) {
 incrementalHotSwap.dependsOn compileJava
 
 task listProjectAPIs(type: GenerateAPIListingTask, dependsOn: generateMergedSwagger)
+
+
+// Define database-related Spring Boot properties required for running
+// command-line tools. These are set up elsewhere for the main App Engine
+// runtime.
```
calbach: Maybe make a note here to keep in sync with the app engine xml template file

api/build.gradle
```diff
@@ -448,3 +463,32 @@ task incrementalHotSwap(type: IncrementalHotSwapTask) {
...
+
+// See api/project.rb backfill-gsuite-user-data
+task backfillGSuiteUserData(type: JavaExec) {
+  classpath = sourceSets.tools.runtimeClasspath
+  main = "org.pmiops.workbench.tools.BackfillGSuiteUserData"
+  systemProperties = commandLineSpringProperties
+  if (project.hasProperty("appArgs")) {
+    args Eval.me(appArgs)
+  }
+}
```
rushtong: nit: newline

api/libproject/devstart.rb
```diff
@@ -1038,12 +1084,12 @@ def set_authority(cmd_name, *args)
...
-      common.run_inline %W{
-        gradle --info setAuthority
-       -PappArgs=['#{op.opts.email}','#{op.opts.authority}',#{op.opts.remove},#{op.opts.dry_run}]}
-    end
+      Dir.chdir("tools") do
+        common = Common.new
+        common.run_inline %W{
+          gradle --info setAuthority
+         -PappArgs=['#{op.opts.email}','#{op.opts.authority}',#{op.opts.remove},#{op.opts.dry_run}]}
+      end
```
rushtong: nit: revert?

api/tools/src/main/java/org/pmiops/workbench/tools/BackfillGSuiteUserData.java
```diff
@@ -0,0 +1,131 @@
...
+      log.info(String.format("%s Backfill complete. Updated %d, skipped %d, failed %d.",
+        dryRun ? "DRY RUN: " : "", updateCount, skipCount, errorCount));
+    };
+  }
+
+  public static void main(String[] args) throws Exception {
+    new SpringApplicationBuilder(BackfillGSuiteUserData.class).web(false)
+      .run(args);
+  }
+}
```
rushtong: nit: newline

api/libproject/devstart.rb
gjuggler: Ah I didn't realize – switched, thanks.

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
gjuggler: AFAICT this would technically have worked correctly as written – the backfill_gsuite_user_data method in devstart.rb contained a "get_gsuite_admin_key" command, which fetches and stores the gsuite key for the current project. Yes, it would have stored the prod key locally when run against the prod project... but it doesn't seem to be any different than how we store the prod key locally when deploying the App Engine project, https://github.com/all-of-us/workbench/blob/c2837ea680eac13cf03b32feadacbf5f8b8fae31/api/libproject/devstart.rb#L1382 but maybe I'm misunderstanding what 'deploy_app' is used for.

Anyway – I was able to change the command-line config to instead fetch the GSuite admin credentials from GCS, which feels like a much better solution in the long run.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -162,10 +162,10 @@ public static void addCustomSchemaValues(User user, String primaryEmail, String
     // In addition to the custom schema value, we store each user's contact email as a secondary
     // email address with type "home". This makes it show up nicely in GSuite admin as the
     // user's "Secondary email".
-      user.setEmails(Lists.newArrayList(
-      new UserEmail().setType("work").setAddress(primaryEmail).setPrimary(true),
-      new UserEmail().setType("home").setAddress(contactEmail)))
-      .setCustomSchemas(Collections.singletonMap(GSUITE_AOU_SCHEMA_NAME, aouCustomFields));
+    user.setEmails(Lists.newArrayList(
+    new UserEmail().setType("work").setAddress(primaryEmail).setPrimary(true),
```
calbach: nit: indent

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
calbach: Ah I missed that line (looks like it was commented out when I was reviewing). Yeah this would have worked.

Note: Minor distinction - when deploy_app runs locally, it typically gets run inside of a "hermetic" docker container (via this deploy stuff: https://github.com/all-of-us/workbench/blob/c2837ea680eac13cf03b32feadacbf5f8b8fae31/deploy/README.md), so the key files aren't getting pulled into your development git client, at least.

PR #1804, opened 2019-02-19
[risk=no] refactor cohort review detail state

panentheos:
This changes the state management on the review detail page to be more react-friendly. It will make gradual conversion of this page much more straightforward. Highlights:
* Several resolvers were removed, and the page component now does its own data loading.
* Most state now formally lives on the page component. Child components receive the data as `Input` bindings, as well as callback functions to request updates, as appropriate.


ui/src/app/cohort-review/annotation-item/annotation-item.component.ts
```diff
@@ -74,7 +71,7 @@ export class AnnotationItemComponent implements OnInit, OnChanges, AfterContentC
     private cdref: ChangeDetectorRef
   ) {}
 
-  ngOnChanges() {
+  ngOnChanges(changes) {
```
blrubenstein: Does this need to be added here?

ui/src/app/cohort-review/annotation-list/annotation-list.component.html
```diff
@@ -6,16 +6,15 @@ <h4 class="annotations-color">Annotations</h4>
     (click)="openManager()">
     <clr-icon shape="plus-circle" size="21" title="Manage Annotations"></clr-icon>
   </button>
-  <button *ngIf="(annotations$ | async).length> 0"
+  <button *ngIf="annotationList.length> 0"
```
blrubenstein: Nit: Add a space before `>`

ui/src/app/cohort-review/detail-all-events/detail-all-events.component.html
```diff
@@ -18,8 +18,7 @@
     </clr-dg-filter>
   </clr-dg-column>
 
-  <clr-dg-row *clrDgItems="let datum of data"
-    (click)="onClick(datum)">
```
blrubenstein: Why was this removed? Just out of curiosity/not understanding.

ui/src/app/cohort-review/detail-page/detail-page.ts
```diff
@@ -12,19 +15,55 @@ export class DetailPage implements OnInit, OnDestroy {
...
   sidebarOpen = true;
   participant: Participant;
-  subscription: Subscription;
-  participantId: number;
-  constructor(private route: ActivatedRoute) {}
+  annotations: ParticipantCohortAnnotation[];
+  annotationDefinitions: CohortAnnotationDefinition[];
+  subscriptions: Subscription[] = [];
+  constructor(
+    private reviewAPI: CohortReviewService,
```
blrubenstein: I would name these `reviewService` and `defsService` because we have a distinction between service and API. (Service is generated for angular, and API is generated for fetch)

ui/src/app/cohort-review/annotation-item/annotation-item.component.ts
panentheos: no, that was left over from development

ui/src/app/cohort-review/detail-all-events/detail-all-events.component.html
panentheos: The downstream event was not actually used.

ui/src/app/cohort-review/detail-page/detail-page.ts
panentheos: I agree with the reasoning, but there is an established pattern of calling these 'API' in the angular code, so I'm inclined to keep that consistency to make the refactoring smoother.

ui/src/app/cohort-review/detail-tabs/detail-tabs.component.html
```diff
@@ -1,6 +1,6 @@
 <clr-tabs>
   <clr-tab>
-    <button clrTabLink id="detail-summary-tab"(click) = "getDomainsParticipantsData()">Summary</button>
+    <button clrTabLink id="detail-summary-tab">Summary</button>
```
smazumd: when we are switching between other tabs summary tab clicking is not smooth it takes time to open the tab. Not sure what is causing the issue.

ui/src/app/cohort-review/detail-page/detail-page.html
```diff
@@ -1,33 +1,41 @@
-<div class="detail-page" [class.sidebar-open]="sidebarOpen">
+<ng-container *ngIf="participant && annotations && annotationDefinitions">
```
smazumd: I observed if I click on participant from participants page, it is routing to the clicked participant and no spinner comes up, and it waits to get all data and then charts spinners start spinning. there is a time gap between getting chart data and all other data(like participants id and all). So I think we need to add a spinner here before ng-container , where we need to check ' <div class="spinner spinner-inverse" *ngIf="!participant && !annotations && !annotationDefinitions">Loading...</div>' -- something like this. but may be there will be more better way  to handle it. it was an observation.

panentheos: Good call, I added a spinner.

ui/src/app/cohort-review/detail-tabs/detail-tabs.component.html
panentheos: I noticed this as well on certain participants. I think there are some issues with the data manipulation in the charts, and it ends up doing substantial extra work when there are many data points, and this leads to a noticeable pause on render. I think this is already happening without this PR, but it's masked because we reload the data every time, so the initial render just shows the spinner. We should investigate that separately.

ui/src/app/cohort-review/detail-page/detail-page.html
smazumd: nit : I think we need some css so that spinner will be in the center of the page. you can add this below css-
.review-spinner {
  position: absolute;
  top: 45%;
  left: 50%;
}
and add this class to the spinner div.

PR #1808, opened 2019-02-19
[DB-164] [risk=no] fixing cloudsql import

SrushtiGangireddy:
1. This pr fixes the cloudsql import which fails when the replication / backup job runs between the csv import jobs.


SrushtiGangireddy:
Tested prod data import with this script and it seems to work. The backup/ failover job did not take over and the import did not fail in between.

PR #1810, opened 2019-02-20
[RW-1876][Risk=no] Change "Clone" to "Duplicate"

UrsaStutsman:
Grep'd views/ for clone. Think I got them all...

### DIFF
--- a/ui/src/app/views/resource-card/component.tsx
+++ b/ui/src/app/views/resource-card/component.tsx
@@ -36,7 +36,7 @@ const ResourceCardMenu: React.FunctionComponent<{
         ['notebook', () => {
           return <React.Fragment>
             <MenuItem icon='pencil' onClick={onRenameNotebook}>Rename</MenuItem>
-            <MenuItem icon='copy' onClick={onCloneResource}>Clone</MenuItem>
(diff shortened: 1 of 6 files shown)


ui/src/app/views/resource-card/component.tsx
```diff
@@ -36,7 +36,7 @@ const ResourceCardMenu: React.FunctionComponent<{
         ['notebook', () => {
           return <React.Fragment>
             <MenuItem icon='pencil' onClick={onRenameNotebook}>Rename</MenuItem>
-            <MenuItem icon='copy' onClick={onCloneResource}>Clone</MenuItem>
+            <MenuItem icon='copy' onClick={onCloneResource}>Duplicate</MenuItem>
```
jacmrob: should we change the names of the methods too? i.e. onCloneResource -> onDuplicateResource.  Just to have no lingering confusion

UrsaStutsman: So I think we discussed this, we're leaving the api as clone (as this mirrors what FC has), and I think we decided the "line" here was what actually appears in the UI, everything else should stay the same

jacmrob: okay in that case 👍

PR #1812, opened 2019-02-20
[RW-2181][risk=low] 500 error calling Profile API syncTrainingStatus()

NehaBroad:


api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -56,6 +56,13 @@ public String readMandrillApiKey() {
     return mandrillKeys.getString("api-key");
   }
 
+  @Override
+  public String getMoodleApiKey() {
+    JSONObject mandrillKeys = new JSONObject(readToString(getCredentialsBucketName(), "moodle-keys.json"));
```
gjuggler: Nit: s/mandrill/moodle/

Also, for better consistency w/ the Mandrill method name, maybe readMoodleApiKey for a method name?

gjuggler: Is it important that this file is a JSON file? Or would just reading a plain-text string be simpler?

(I'm all for consistency, so if you say it's better for us to be consistent w/ how we store Mandrill keys, I'm fine with that! Just update the API keys section in our Moodle integration docs to describe the JSON structure, so the next person who has to regenerate keys knows what format they should be in.)

NehaBroad: I copy pasted the code <embarrassed>

NehaBroad: I was debating this myself to make it like invitation-key or mandrill. I think i will revert it to be like invitation-key which is a text as we have currently just one element.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -15,14 +17,23 @@
...
   private static final String RESPONSE_FORMAT = "json";
-  private static final String TOKEN = "d106acbb7eb050a10e9b200eb82b8be8";
   private static final String GET_MOODLE_ID_SEARCH_FIELD = "email";
   private static final String MOODLE_EXCEPTION = "moodle_exception";
   private static final String MOODLE_USER_NOT_ALLOWED_ERROR_CODE = "guestsarenotallowed";
+  private CloudStorageService cloudStorageService;
 
+  @Autowired
+  public ComplianceServiceImpl(
+      CloudStorageService cloudStorageService) {
```
rushtong: nit: extraneous line break here.
```suggestion
      public ComplianceServiceImpl(CloudStorageService cloudStorageService) {
```

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -15,14 +17,23 @@
...
+    this.cloudStorageService = cloudStorageService;
+  }
+
+  private String getToken() {
+    return this.cloudStorageService.getMoodleApiKey();
+  }
```
rushtong: nit: add line between methods

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
rushtong: We did start down the path of json-ifying configs, but I don't think it's necessary.

PR #1816, opened 2019-02-21
[RISK=LOW]Correcting path for failing bulk sync training status cron job

NehaBroad:
Cron job xml had the following entry
  <cron>
    <url>/v1/cron/bulkSyncTrainingStatus</url>
    <target>api</target>
    <description>Periodic Sync users training status</description>
    <schedule>every 24 hours</schedule>
  </cron>

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -484,7 +484,7 @@ paths:
           type: string
           required: false
 
-  /v1/account/bulkSyncTrainingStatus:
+  /v1/cron/bulkSyncTrainingStatus:
     post:
       tags:
         - user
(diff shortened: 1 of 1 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -484,7 +484,7 @@ paths:
           type: string
           required: false
 
-  /v1/account/bulkSyncTrainingStatus:
+  /v1/cron/bulkSyncTrainingStatus:
```
calbach: Could you also move this down to 812? There's a relevant comment header there https://github.com/all-of-us/workbench/blob/7cc0a4b0275ae21b0da63b38f67d761b5202cdf0/api/src/main/resources/workbench.yaml#L812

PR #1822, opened 2019-02-22
[RW-2035] [risk=no] Handle null contactEmail in GSuite backfill script

gjuggler:
While running the script against staging, I found that we have some users where the contactEmail is null. Just in case this might also happen in the request-handling path somehow, this PR updates the logic to not include an empty contactEmail.


calbach:
FYI: this may happen when we clear the database (which we've done a few times in test/staging in the past) or possibly for some very old user accounts. Users from prior version of the app appear on the site and we create an account for them, without knowing their contact email (https://github.com/all-of-us/workbench/blob/dd11c400fade01078bfd32571a175ceae08d9ffd/api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java#L150).

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -69,8 +70,8 @@
 
   @Autowired
   public DirectoryServiceImpl(Provider<GoogleCredential> googleCredentialProvider,
-      Provider<WorkbenchConfig> configProvider,
-      HttpTransport httpTransport, GoogleRetryHandler retryHandler) {
+                              Provider<WorkbenchConfig> configProvider,
+                              HttpTransport httpTransport, GoogleRetryHandler retryHandler) {
```
rushtong: nit: revert?

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -155,17 +157,23 @@ public static void addCustomSchemaValues(User user, String primaryEmail, String
...
+    }
 
     // In addition to the custom schema value, we store each user's contact email as a secondary
     // email address with type "home". This makes it show up nicely in GSuite admin as the
     // user's "Secondary email".
-    user.setEmails(Lists.newArrayList(
-    new UserEmail().setType("work").setAddress(primaryEmail).setPrimary(true),
-    new UserEmail().setType("home").setAddress(contactEmail)))
-    .setCustomSchemas(Collections.singletonMap(GSUITE_AOU_SCHEMA_NAME, aouCustomFields));
+    ArrayList<UserEmail> emails = Lists.newArrayList(
```
rushtong: nit: prefer the `List` interface.
```suggestion
    List<UserEmail> emails = Lists.newArrayList(
```

PR #1823, opened 2019-02-22
[RW-2194] [risk=no] Try reducing the Java memory limit for CircleCI.

gjuggler:
See the Jira ticket and comment in config.yml for more info.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
             - ~/workbench/api/build/exploded-api/WEB-INF/lib/
           key: api-cache-{{ checksum "~/workbench/api/build.gradle" }}
 
-  api-local-test:
+  build:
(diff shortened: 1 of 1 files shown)


gjuggler:
I managed to kick off a CircleCI run with this config change, https://circleci.com/gh/all-of-us/workbench/38610, which was green. I'll re-run it a few times to see if it's still flaky, but at least the 2g limit doesn't totally break our build.

calbach:
That's a new one - dockerhub flake maybe? Looks good assuming rerun passes

gjuggler:
Yeah, that failure looked better on a re-run.

I'm still not 100% confident on this change, since flaky OOMs are really hard to test one way or another. But there's a chance this will fix it, and it's easy to revert if it actually makes things worse.

PR #1825, opened 2019-02-22
[RW-2196] Remove completeEthicsTraining() endpoint

jacmrob:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -475,12 +475,6 @@ public class ProfileController implements ProfileApiDelegate {
   }
 
   @Override
-  public ResponseEntity<Profile> completeEthicsTraining() {
-    User user = userService.submitEthicsTraining();
-    return getProfileResponse(saveUserWithConflictHandling(user));
-  }
-
-  @Override
   public ResponseEntity<Profile> submitTermsOfService() {
     User user = userService.submitTermsOfService();
     return getProfileResponse(saveUserWithConflictHandling(user));
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -355,19 +355,6 @@ paths:
           schema:
             $ref: "#/definitions/Profile"
 
-  # TODO: add other state pertaining to ethics training?
-  /v1/account/complete-ethics-training:
-    post:
-      tags:
-        - profile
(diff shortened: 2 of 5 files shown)


calbach:
We can't really remove this until Moodle integration is fully launched in all environments, unless we're changing the registration criteria or repurposing syncTrainingStatus() in the meantime. Am I missing something?

gjuggler:
> We can't really remove this until Moodle integration is fully launched in all environments, unless we're changing the registration criteria or repurposing syncTrainingStatus() in the meantime. Am I missing something?

CH, is this related to the call to completeEthicsTraining from unregistered/component.ts? I'm confused how this API call was / is being used today to allow access to the workbench (maybe partly because I'm usually testing in test, where we're not blocking anything on being in the registered group).

I'm hoping there's a way we can simplify / coordinate how the workbench decides whether a user should have access to the workbench, and how we control this across environments & users. Basically, if Moodle isn't launched in prod, we shouldn't require Moodle completion to access the workbench. And similar for eRA commons and all other access requirements.

I'm chatting with Karthik now to better understand what the requirements here will be, especially w.r.t. era commons where it looks like we may also need user-specific overrides.

rushtong:
@gjuggler - I think you're describing a good chunk of https://precisionmedicineinitiative.atlassian.net/browse/RW-2124

jacmrob:
I'm going to close this PR for now and open one that is just changing the button in the profile, not touching any API endpoints

calbach:
@gjuggler the "compliance training completion" bit is required for registered access. The way that's getting set currently is via this call on the unregistered page. If we remove this staging/stable/prod users will be unable to become registered. Three possible solutions:

1. Leave things as they are, remove this endpoint once we've launched Moodle
2. Modify syncTrainingStatus() to simply set compliance training "complete" if Moodle is disabled.
3. Remove training from the registration criteria until Moodle is launched (i.e. change the logic based on the flag). We'd have to think through what will happen vs what we want to happen when Moodle is enabled (do we want to require all prod users to suddenly become unregistered when we launch Moodle integration and push them to complete training - or is that optional?).

gjuggler:
@calbach Thanks for the details. I think we'll want to aim for something like #3 in the long term, but to have a consistent working model across all of our "access requirements". Karthik and I worked out a rough model for this... lemme put out a rough doc describing this & I'll run it by you early next week.

calbach:
@gjuggler sounds good, after typing it out fully I also think (3) is probably the cleanest path.

PR #1826, opened 2019-02-22
[RISK=LOW]Moodle flag and host change

NehaBroad:
Call moodle api IF moodleApi is enabled.

Depending on the environment updates moodle host

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
@@ -1,15 +1,12 @@
 package org.pmiops.workbench.compliance;
 
 import org.pmiops.workbench.google.CloudStorageService;
-import org.pmiops.workbench.moodle.ApiClient;
 import org.pmiops.workbench.moodle.ApiException;
-import org.pmiops.workbench.moodle.MoodleConfig;
 import org.pmiops.workbench.moodle.api.MoodleApi;
 import org.pmiops.workbench.moodle.model.BadgeDetails;
 import org.pmiops.workbench.moodle.model.MoodleUserResponse;
 import org.pmiops.workbench.moodle.model.UserBadgeResponse;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 import org.pmiops.workbench.config.WorkbenchConfig;
@@ -27,7 +24,7 @@ public class ComplianceServiceImpl implements ComplianceService {
   private CloudStorageService cloudStorageService;
   private Provider<WorkbenchConfig> configProvider;
 
-  private Provider<MoodleApi> moodleApi;
+  private Provider<MoodleApi> moodleApiProvider;
 
 
   @Autowired
@@ -36,7 +33,7 @@ public class ComplianceServiceImpl implements ComplianceService {
       Provider<MoodleApi> moodleApiProvider) {
     this.cloudStorageService = cloudStorageService;
     this.configProvider = configProvider;
-    moodleApi = moodleApiProvider;
+    this.moodleApiProvider = moodleApiProvider;
 
   }
 
@@ -53,7 +50,7 @@ public class ComplianceServiceImpl implements ComplianceService {
     if (!enableMoodleCalls()) {
       return null;
     }
-    List<MoodleUserResponse> response = moodleApi.get().getMoodleId(getToken(), GET_MOODLE_ID_SEARCH_FIELD,  email);
+    List<MoodleUserResponse> response = moodleApiProvider.get().getMoodleId(getToken(), GET_MOODLE_ID_SEARCH_FIELD,  email);
     if (response.size() == 0) {
       return null;
     }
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -21,18 +23,35 @@
...
+  }
+
+  private void setApiHost() {
+    ApiClient apiClient = api.getApiClient();
+    apiClient.setBasePath("https://" + configProvider.get().moodle.host +"/webservice/rest");
+  }
+
   @Override
   public Integer getMoodleId(String email) throws ApiException {
+    if (!enableMoodleCalls())
```
gjuggler: style nit: add open / close braces on conditionals (here and below)

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -21,18 +23,35 @@
...
+      return null;
+    setApiHost();
+    api.getApiClient().setDebugging(true);
```
gjuggler: I'm not sure what the impact of turning on API client debugging is – should this be on in production or was it just for local testing?

api/src/main/resources/moodle.yaml
```diff
@@ -2,6 +2,7 @@
 # MOodle API information can be found at :
 # https://docs.moodle.org/dev/Web_service_API_functions
 
+#The host will get updated as per the environment it will be aoudev.nnlm.gov for dev and test
```
gjuggler: I'd move this down to L13 where the host value is. And it may be worth referencing "ComplianceServiceImpl.java" as the place where this configuration happens.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -42,6 +61,11 @@ public Integer getMoodleId(String email) throws ApiException {
 
   @Override
   public List<BadgeDetails>  getUserBadge(int userMoodleId) throws ApiException {
+    if (!enableMoodleCalls())
+      return null;
+    setApiHost();
+    api.getApiClient().setDebugging(true);
```
calbach: Should this perhaps be a config value as we have for Firecloud (`debugEndpoints`)?

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -21,18 +23,35 @@
...
+  private void setApiHost() {
+    ApiClient apiClient = api.getApiClient();
+    apiClient.setBasePath("https://" + configProvider.get().moodle.host +"/webservice/rest");
+  }
+
   @Override
   public Integer getMoodleId(String email) throws ApiException {
+    if (!enableMoodleCalls())
+      return null;
+    setApiHost();
```
calbach: Manually setting this before every call seems like a lot of duplication. Perhaps it would make more sense to copy the approach we use for the other API clients: add a `MoodleConfig.java` file in this directory similar to https://github.com/all-of-us/workbench/blob/dd11c400fade01078bfd32571a175ceae08d9ffd/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java

and provide a properly configured `MoodleApi` from there. Then this service takes a `Provider<MoodleApi>` and uses that for calls.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
NehaBroad: i meant to remove this but forgot, it was for my testing

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
NehaBroad: I forgot to remove this, it was just for my testing purpose

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -25,12 +27,17 @@
...
   private Provider<WorkbenchConfig> configProvider;
 
+  private MoodleConfig moodleConfig;
+
 
   @Autowired
   public ComplianceServiceImpl(CloudStorageService cloudStorageService,
-      Provider<WorkbenchConfig> configProvider) {
+      Provider<WorkbenchConfig> configProvider,
+      Provider<MoodleConfig> moodleConfigProvider) {
```
calbach: In my suggested approach, this would be `Provider<MoodleApi>` instead. The `moodleApi` method you defined will be called automatically by spring and injected.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -21,19 +25,35 @@
   private static final String MOODLE_EXCEPTION = "moodle_exception";
   private static final String MOODLE_USER_NOT_ALLOWED_ERROR_CODE = "guestsarenotallowed";
   private CloudStorageService cloudStorageService;
+  private Provider<WorkbenchConfig> configProvider;
+
+  private Provider<MoodleApi> moodleApi;
```
calbach: nit: `moodleApiProvider` for consistency

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -58,16 +60,14 @@ public UserService(Provider<User> userProvider,
       Clock clock,
       Random random,
       FireCloudService fireCloudService,
-      Provider<WorkbenchConfig> configProvider,
-      Provider<ComplianceService> complianceServiceProvider) {
+      Provider<WorkbenchConfig> configProvider) {
```
calbach: I was actually suggesting:

```
public UserService(
    ...
    ComplianceService complianceService) {
  ...
  this.complianceService = complianceService;
```

YOu don't need the provider in this case because ComplianceService is a singleton.

PR #1830, opened 2019-02-25
[risk=no] convert annotation definition modals to react

panentheos:
This converts the 'add' and 'edit' modals for annotation definitions. The basic UX is the same, but there are several improvements, mostly around validation and error handling. It also standardizes the visual styles.


ui/src/app/cohort-review/annotation-definition-modals/annotation-definition-modals.component.tsx
```diff
@@ -0,0 +1,284 @@
...
+      saving: false
+    };
+  }
+
+  async create() {
+    try {
+      const {onCreate, urlParams: {ns, wsid, cid}} = this.props;
+      const {name, annotationType, enumValues} = this.state;
+      this.setState({saving: true});
+      const newAd = await cohortAnnotationDefinitionApi().createCohortAnnotationDefinition(
```
blrubenstein: Nit: I'd call this `newAnnotationDefinition`, especially since it isn't used very often? I think its more understandable.

Also, can this be inlined below?

panentheos: It could be inlined, yes. I usually do it this way for readability, to highlight the `await`, otherwise it can get lost inside the expression. The other confounding issue here is that a longer name would go past the line limit and require additional restructuring. (FWIW I don't like line limits for that reason, but that's another discussion)

PR #1831, opened 2019-02-25
[RW-2195][risk=no] Convert Bug Report

jacmrob:


ui/src/app/views/admin-review-workspace/component.html
```diff
@@ -34,4 +34,6 @@ <h3 class="modal-title">Error:</h3>
     <span class="loading-text">Loading workspaces for review...</span>
   </div>
 </ng-template>
-<app-bug-report></app-bug-report>
\ No newline at end of file
+<app-bug-report *ngIf="bugReportOpen"
```
blrubenstein: I don't know if we actually have this automatic bug report thing anywhere else. This may be able to just be removed.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+      submitting: false,
+      sendBugReportError: false,
+      bugReport: {...this.emptyBugReport,
+        shortDescription: props.bugReportDescription || '',
+        contactEmail: props.profileState.profile.username
+      }
+    };
+  }
+
+  async send() {
```
panentheos: nit: Using `async` here suggests that the function will use `await` on promises, but it doesn't.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+import {AlertDanger} from 'app/components/alert';
+import {Button} from 'app/components/buttons';
+import {ClrIcon} from 'app/components/icons';
+import {Modal, ModalBody, ModalFooter, ModalInput, ModalTitle} from 'app/components/modals';
+import {
+  BugReport,
+  BugReportType,
+  Profile
+} from 'generated/fetch';
+import * as React from 'react';
```
blrubenstein: Nit: I'd bump library imports to the top, and separate out the generated into its own import block.

ui/src/app/views/signed-in/component.html
```diff
@@ -93,4 +93,6 @@ <h6 style="padding: 0.5rem;">
     <app-routing-spinner></app-routing-spinner>
   </div>
 </div>
-<app-bug-report></app-bug-report>
+<app-bug-report *ngIf="bugReportOpen"
```
panentheos: TOL: If we want to be able to launch this arbitrarily on any page, we could think about making it a more global piece of infrastructure at some point, e.g. rendering it at the top level of the app and driving it off some global state.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  render() {
+    const {bugReport, submitting, sendBugReportError} = this.state;
+    const {onClose} = this.props;
+    return <Modal width={700} loading={submitting}>
+      <ModalTitle style={{fontWeight: 400}}>Submit a Bug</ModalTitle>
+      <ModalBody>
+        <label className='required' style={styles.fieldHeader}>Description: </label>
+        <ModalInput value={bugReport.shortDescription}
```
panentheos: This should use `TextInput`. `ModalInput` should be deprecated, actually. (Note that the `onChange` parameter on `TextInput` is different)

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+                onChange={(e) => {this.setState(
+                  {bugReport: {
+                    ...bugReport,
+                    bugReportType: e.target.value as unknown as BugReportType}
+                  }); }}>
+          <option value={BugReportType.APPLICATION}>Workbench Application</option>
+          <option value={BugReportType.DATA}>Data Quality</option>
+        </select>
+        <label style={styles.fieldHeader}>How to Reproduce: </label>
+        <textarea onChange={(v) => {bugReport.reproSteps = v.target.value; } }/>
```
panentheos: This should use `TextArea`. (And note different `onChange` signature)

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+        <div style={{display: 'flex', flexDirection: 'row', marginTop: '1rem'}}>
+          <input type='checkbox'
+                 defaultChecked={bugReport.includeNotebookLogs}
```
panentheos: nit: You should be able to use `checked` rather than `defaultChecked` here.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+                   {bugReport: {
+                     ...bugReport,
+                     includeNotebookLogs: !bugReport.includeNotebookLogs
+                   }})}
+          />
+          <label style={{...styles.fieldHeader, marginTop: '0rem', marginLeft: '0.2rem'}}>
+            Attach notebook logs? </label>
+        </div>
+        <label style={styles.fieldHeader}>Contact Email: </label>
+        <ModalInput value={bugReport.contactEmail}
```
panentheos: `TextInput`

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+  render() {
+    const {bugReport, submitting, sendBugReportError} = this.state;
+    const {onClose} = this.props;
+    return <Modal width={700} loading={submitting}>
+      <ModalTitle style={{fontWeight: 400}}>Submit a Bug</ModalTitle>
+      <ModalBody>
+        <label className='required' style={styles.fieldHeader}>Description: </label>
+        <ModalInput value={bugReport.shortDescription}
+                    onChange={(v) => this.setState({bugReport:
+                        {...bugReport, shortDescription: v.target.value}})}/>
```
panentheos: FYI, lodash has some nice utils for performing deep updates on data structures. See `_.set` and `_.update` for useful ones.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+  async send() {
+    this.setState({submitting: true, sendBugReportError: false});
+    bugReportApi().sendBugReport(this.state.bugReport).then(() => {
+      this.setState({submitting: false});
+      this.props.onClose();
+    }
+    ).catch(() => {
+      this.setState({sendBugReportError: true, submitting: false});
+    });
+
```
blrubenstein: Nit: Remove extra space.

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const {bugReport, submitting, sendBugReportError} = this.state;
+    const {onClose} = this.props;
+    return <Modal width={700} loading={submitting}>
+      <ModalTitle style={{fontWeight: 400}}>Submit a Bug</ModalTitle>
+      <ModalBody>
+        <label className='required' style={styles.fieldHeader}>Description: </label>
+        <ModalInput value={bugReport.shortDescription}
+                    onChange={(v) => this.setState({bugReport:
+                        {...bugReport, shortDescription: v.target.value}})}/>
+        <label className='required' style={styles.fieldHeader}>Bug Type: </label>
```
blrubenstein: Could you add validation with the validate.js library here? Just to make sure we handle validation cleanly and consistently?

ui/src/app/views/bug-report/component.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <option value={BugReportType.APPLICATION}>Workbench Application</option>
+          <option value={BugReportType.DATA}>Data Quality</option>
+        </select>
+        <label style={styles.fieldHeader}>How to Reproduce: </label>
+        <textarea onChange={(v) => {bugReport.reproSteps = v.target.value; } }/>
+        <div style={{display: 'flex', flexDirection: 'row', marginTop: '1rem'}}>
+          <input type='checkbox'
+                 defaultChecked={bugReport.includeNotebookLogs}
+                 onClick={() => this.setState(
+                   {bugReport: {
```
calbach: nit: may want to use the `updater` form here with `fp.set` throughout here. It is technically possible state may have changed before this update gets triggered.

ui/src/app/views/signed-in/component.html
jacmrob: I dunno if it is global state though because we want to open it conditionally with certain parameters when we trigger certain errors

ui/src/app/views/bug-report/component.tsx
jacmrob: it was giving me a complier error

ui/src/app/views/signed-in/component.html
calbach: Driveby: What is TOL? (I'm assuming it's not Throne of Lies).

Also I think Brett means the global state would be "show or not show BugReport". The determination as to whether you toggle visibility could still be made at the component level (i.e. here). It feels like there's a good argument for making this component more of a singleton - the modal is already rendered in global scope, and you'd never want to open multiple separate bug report modals simultaneously.

panentheos: Heh, sorry. TOL means 'thinking out loud'. Terra started using it as a way to suggest longer-term ideas that are relevant but shouldn't necessarily be part of the current PR.

And yes, that was what I was thinking regarding the bug report modal. With that in place, any code path could just call e.g. `openBugReportModal('subject')` without having to manage its own instance of it.

PR #1834, opened 2019-02-25
[RW-2107] adding data set page

UrsaStutsman:
For feedback from Scout

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -219,7 +219,7 @@ const routes: Routes = [
                 path: 'data/datasets',
                 component: DataSetComponent,
                 data: {
-                  title: 'DataSet Page',
+                  title: 'Dataset Page',
                   breadcrumb: {
                     value: 'Dataset'
                   }
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -221,7 +221,7 @@ const routes: Routes = [
                 data: {
                   title: 'DataSet Page',
                   breadcrumb: {
-                    value: 'DataSet'
+                    value: 'Dataset'
                   }
                 }
               },
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -29,6 +29,7 @@ import {ConceptSetResolver} from './resolvers/concept-set';
 import {WorkspaceResolver} from './resolvers/workspace';
 
 import {DataPageComponent} from 'app/views/data-page/component';
+import {DataSetComponent} from 'app/views/dataset/component';
 import {environment} from 'environments/environment';
 import {NavStore} from './utils/navigation';
 import {SignInComponent} from './views/sign-in/component';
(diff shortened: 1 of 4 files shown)


ui/src/app/app-routing.module.ts
```diff
@@ -214,6 +215,16 @@ const routes: Routes = [
                   }
                 }
               },
+              {
+                path: 'data/datasets',
+                component: DataSetComponent,
+                data: {
+                  title: 'DataSet Page',
```
blrubenstein: I’d probably do `Dataset` or `Data Set` rather than `DataSet` for anything user facing? Keri would have a solid answer there?

ui/src/app/views/dataset/component.tsx
```diff
@@ -0,0 +1,99 @@
...
+          variables and values for one or more of your cohorts. Then export the completed dataset
+          to Notebooks where you can perform your analysis</div>
+      </FadeBox>
+      <div style={{display: 'flex'}}>
+        <div style={{marginLeft: '1.5rem', marginRight: '1.5rem', width: '33%'}}>
+          <h2>Select Cohorts</h2>
+          <div style={{backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
+            <div style={styles.selectBoxHeader}>
+              Cohorts
+              <ClrIcon shape='plus-circle' class='is-solid' style={styles.addIcon}/>
```
blrubenstein: I think the + should probably route to the cohort builder?

ui/src/app/views/dataset/component.tsx
```diff
@@ -0,0 +1,99 @@
...
+            <div style={{height: '8rem'}}/>
+          </div>
+        </div>
+        <div style={{width: '58%'}}>
+          <h2>Select Concept Sets</h2>
+          <div style={{display: 'flex', backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
+            <div style={{flexGrow: 1}}>
+              <div style={{...styles.selectBoxHeader, borderRight: '1px solid #E5E5E5'}}>
+                Concept Sets
+                <ClrIcon shape='plus-circle' class='is-solid' style={styles.addIcon}/>
```
blrubenstein: Likewise should probably open the new concept set modal

blrubenstein:
I think I saw that createConceptSet modal error when I was doing my work, I believe it has to do with null vs undefined check? I don't recall exactly though.

UrsaStutsman:
@blrubenstein I think I addressed everything. I also disabled the save button. And the problem with the concept sets was somehow I missed adding the componentDidMount which loads the concept domain info. I am getting those same Snyk errors again, so not sure if I can merge, or if we're still not blocking on those.

UrsaStutsman:
Screenshot: (Blank spaces are TODO's for cards: RW-2095 and RW-2098)
workspaces-screenshot

ui/src/app/views/dataset/component.tsx
```diff
@@ -0,0 +1,147 @@
...
+          </div>
+        </div>
+      </FadeBox>
+      <FadeBox style={{marginTop: '1rem'}}>
+        <div style={{backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
+          <div style={{...styles.selectBoxHeader, display: 'flex', position: 'relative'}}>
+            <div>Preview Dataset</div>
+            <div style={{marginLeft: '1rem', color: '#000000', fontSize: '14px'}}>A visualization
+              of your data table based on the variable and value you selected above</div>
+            <Button style={{position: 'absolute', right: '1rem', top: '.25rem'}} disabled={true}>
```
jacmrob: why setting disabled to true here?  if it's for code that hasn't been written yet add a comment?

blrubenstein:
@UrsaStutsman sounds good! This looks great.

PR #1835, opened 2019-02-26
[no ticket][risk=no] Add an alert when waiting on badge completion API from moodle

jacmrob:


jacmrob:
Context: the moodle api takes about 1-2 minutes after you complete training to update your badges; show this warning so that the user is not confused if they've completed training but still see it as an access task.

TODO: add a page-refresh

PR #1836, opened 2019-02-26
[RW-2194] [risk=no] Add a script to run CircleCI with local config changes.

gjuggler:
While working on RW-2194, I dug around the forums a bit and found an example script which lets you test local Circle changes via their API. This helped a ton in letting me actually see whether the memory change had an effect, without committing the change to master before verifying.


calbach:
To clarify the behavior, Circle should be picking up changes to the config when you push a PR. The exception here is that certain tasks only run on the `master` branch post-merge. In the rare case where you need to mess with those without merging, you can technically do this by temporarily changing the matching logic here to pickup your branch: 
https://github.com/all-of-us/workbench/blob/324c5850a36a80461df25082cc1b4e1e36f96c80/.circleci/config.yml#L349

calbach:
Perhaps things like machine size aren't getting picked up though?

gjuggler:
> Perhaps things like machine size aren't getting picked up though?

The behavior I was seeing was that when I pushed my branch gj/circle-mem (with config.yml changes to the Java memory env flag), to GH, it would trigger CircleCI tests to run, but they wouldn't incorporate the config.yml changes from that branch.

Maybe I was missing something though, or Circle's merge logic doesn't include changes to the java_defaults section (which would seem somewhat odd).

I mostly put this up because I found this script useful... happy to drop the PR until we have a chance to verify the behavior.

calbach:
Thanks for sending the PR - I'm just trying to reconcile my understanding first, this script may make sense to add. How were you verifying that your machine size change wasn't picked up? For example, the configuration page on your PR run seems to show the right values, and the logs show `Picked up JAVA_TOOL_OPTIONS: -Xmx2g`:

https://circleci.com/gh/all-of-us/workbench/38642#config/containers/0

gjuggler:
> Thanks for sending the PR - I'm just trying to reconcile my understanding first, this script may make sense to add. How were you verifying that your machine size change wasn't picked up? For example, the configuration page on your PR run seems to show the right values, and the logs show `Picked up JAVA_TOOL_OPTIONS: -Xmx2g`:
> 
> https://circleci.com/gh/all-of-us/workbench/38642#config/containers/0

Yeah – where I was seeing issues were in runs when I was testing the original PR, for example https://circleci.com/gh/all-of-us/workbench/38593#config/containers/0 and https://circleci.com/gh/all-of-us/workbench/38593#tests/containers/0.

The config tab correctly shows the "JAVA_TOOL_OPTIONS: -Xmx2g" entry, but the test run was incorrectly picking up the stale value: "Picked up JAVA_TOOL_OPTIONS: -Xmx3g"

My API-triggered runs seemed to work: https://circleci.com/gh/all-of-us/workbench/38610#tests/containers/0

Maybe this issue is somehow restricted to environment-style config changes like the memory variable? Scout showed me their PR-triggered CircleCI test runs, so I do see it picking up most normal branch-local changes.

calbach:
In the circle run you linked to, the config still had xmx3g on the api-local-test (likely an older version). Do you have a different example which shows the issue?

gjuggler:
> In the circle run you linked to, the config still had xmx3g on the api-local-test config (likely an older version of your PR). Do you have a different example which shows the issue?

Ahhh, yeah that must have been what I was fighting against the whole time... I think I'd convinced myself so strongly that the config changes weren't getting picked up (while not noticing the second 3g > 2g change required) that I didn't recognize the correctness of CircleCI's behavior after the fact.

I'll revert this PR, as it must be WAI then.

calbach:
Phew, ok. Unfortunately I've found Circle's docs to be very lacking and determining the actual/expected behaviors in these cases typically reduces to a guess and check. I've also spent a fair amount of time debugging/configuring Circle so I may be able to help if more questions come up here.

PR #1839, opened 2019-02-26
[RW-2001][risk=no] creating 1 denormalized search table.

freemabd1:
creating 1 denormalized search table.


api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/PMQueryBuilder.java
```diff
@@ -65,22 +68,25 @@
   private static final String VALUE_AS_NUMBER = "and value_as_number ${operator} ${value}\n";
 
   private static final String BP_INNER_SQL_TEMPLATE =
-    "select person_id from `${projectId}.${dataSetId}.search_pm`\n" +
-    "   where source_concept_id = ${conceptId}\n";
+    "select person_id from `${projectId}.${dataSetId}." + TABLE_ID + "`\n" +
```
calbach: Do we still need these separate query builders?

freemabd1: I thought about removing them but I decided to wait until the redesign of CB search. We are moving to a new design were calling universal search that combines concept search/tree searches into one experience. This will allow users to search by many other standard vocabularies like HCPCS, LOINC.. etc, as well as source vocabularies. We are still in discovery of what this means from an engineering standpoint, but i envision that these will probably go away with this redesign.

PR #1841, opened 2019-02-26
[RW-1905][risk=no] Workspace List conversion

jacmrob:


ui/src/app/views/workspace-list/component.tsx
```diff
@@ -0,0 +1,350 @@
...
+  },
+  permissionBox: {
+    color: '#FFFFFF', height: '1rem', width: '3rem', fontSize: 10, textAlign: 'center',
+    borderRadius: '0.2rem', padding: 0
+  }
+});
+
+
+const WorkspaceCardMenu: React.FunctionComponent<{
+  disabled: boolean, wp: WorkspacePermissions, onShare: Function, onDelete: Function
```
UrsaStutsman: There's something funky going on with the formatting of these couple lines, made it difficult to parse.

ui/src/app/views/workspace-list/component.tsx
```diff
@@ -0,0 +1,350 @@
...
+      this.setState({confirmDeleting: false});
+      this.props.reload();
+    }).catch(() => {
+      this.setState({workspaceDeletionError: true});
+    });
+  }
+
+  shareWorkspace(): void {
+    this.setState({sharing: false});
+    this.props.reload();
```
UrsaStutsman: So I know that this was something I did when I converted the workspace share card, but that was because I could not think of how to communicate back to the angular parent the changes from the modal child. (And I was sick of that card, so used the way that worked easily). But I think there maybe a better way here, like instead of this being a "reload" function per se, you could have this function update the state in the parent (thereby re-rendering the page, without a full reload). Questions, don't hesitate. (And I could be wrong!)

ui/src/app/views/workspace-list/component.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {
+  WorkspacesApiStub,
+  workspaceStubs,
+  WorkspaceStubVariables
+} from 'testing/stubs/workspaces-api-stub';
+import {waitOneTickAndUpdate} from 'testing/react-test-helpers';
+
+// Mock the navigate function but not userProfileStore
+jest.mock('app/utils/navigation', () => ({
+    ...(jest.requireActual('app/utils/navigation')),
```
UrsaStutsman: running jest locally throws an error on this line for me.

ui/src/app/views/workspace-list/component.tsx
jacmrob: i think this actually is the right way to do things; for all our modals we have a piece of state in the parent that says whether the modal is open or not, and the callback function when the modal closes changes the state and reloads (bc elements of the parent page may have changed with the actions of that modal, like deleting workspaces for example).  That's how we're doing it for workspace delete so i'm okay keeping it this way

ui/src/app/views/workspace-list/component.spec.tsx
jacmrob: fixed this!

ui/src/app/views/workspace-list/component.tsx
UrsaStutsman: Sounds good. I'm putting on my back burner, because I think on this page for share, it's not that necessary (view won't change at all), and it would be good if we could update the page without a full reload on delete. But not stressing it now.

PR #1842, opened 2019-02-27
[RW-1948][risk=no] Switch to events, fix visits, minor schema code cleanup

calbach:
Splitting this PR out for discussion primarily. This is a design change (also made a comment in the design doc: https://docs.google.com/document/d/1N_TDTOi-moTH6wrXn1Ix4dwUlw4j8GT9OsL9yXYXYmY/edit?hl=en#heading=h.jduorqros7hm)

This changes from a per-domain event approach to a single schema and nested field for all events. This maintains the top level per-domain ID fields to allow for "TOP N conditions" type aggregations.

Pros:
- Single template for all search group queries, don't need to worry about mixed domain types in a criteria tree (as much)

Cons:
- May be less useful for aggregation later in unforeseen use cases.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
@@ -64,7 +64,7 @@ public class ElasticDocument {
       .build();
 
   private static Map<String, Object> esType(ElasticType t) {
-    return ImmutableMap.of("type", ElasticType.KEYWORD.lower());
+    return ImmutableMap.of("type", t.lower());
   }
 
   public final String id;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
```diff
@@ -38,31 +38,35 @@ String lower() {
...
-      .put("condition_source_concept_ids", ImmutableMap.of("type", ElasticType.KEYWORD.lower()))
-      .put("conditions", NESTED_FOREIGN_SCHEMA)
+      .put("gender_concept_id", esType(ElasticType.KEYWORD))
+      .put("condition_concept_ids", esType(ElasticType.KEYWORD))
+      .put("condition_source_concept_ids", esType(ElasticType.KEYWORD))
+      .put("events", NESTED_FOREIGN_SCHEMA)
       .build();
 
+  private static Map<String, Object> esType(ElasticType t) {
+    return ImmutableMap.of("type", ElasticType.KEYWORD.lower());
```
gjuggler: Unless I'm missing something, should this be t.lower()?

calbach: :man_facepalming: 

Odd that the indexer still ran successfully. If only someone had thought to add tests for this.

calbach:
@gjuggler @freemabd1 PTAL

PR #1844, opened 2019-02-27
[RW-2010][Risk=low] Adding in gtag user agent custom dimension to allow us to filter in GA

UrsaStutsman:
Ready for review. Any questions don't hesitate!

### DIFF
--- a/ui/src/app/views/app/component.ts
+++ b/ui/src/app/views/app/component.ts
@@ -137,10 +137,6 @@ export class AppComponent implements OnInit {
    */
   private setGTagManager() {
     const s = this.doc.createElement('script');
-    // This is set in Google Analytics, all the other envs this was the first
-    // custom dimension, but in test it was the second, no way to change it that
-    // I can find -US 2/27/18
-    const d = environment.gaId === 'UA-112406425-1' ? 'dimension2' : 'dimension1';
     s.type = 'text/javascript';
     s.innerHTML =
       '(function(w,d,s,l,i){' +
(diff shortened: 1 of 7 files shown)


ui/src/app/views/app/component.ts
```diff
@@ -137,6 +137,10 @@ export class AppComponent implements OnInit {
    */
   private setGTagManager() {
     const s = this.doc.createElement('script');
+    // This is set in Google Analytics, all the other envs this was the first
+    // custom dimension, but in test it was the second, no way to change it that
+    // I can find -US 2/27/18
+    const d = environment.gaId === 'UA-112406425-1' ? 'dimension2' : 'dimension1';
```
rushtong: An alternative to calculating it here is to set it as an environment variable, i.e. 
```
const d = environment.gaDimension;
```
And of course, set that value in the ts files.

ui/src/environments/environment-type.ts
```diff
@@ -18,6 +18,9 @@ export interface Environment {
   // The Google Analytics account ID for logging actions and page views.
   // Example value: 'UA-112406425-3'
   gaId: string;
+  // The Google Analytics custom dimension ID for sending User Agent
+  // info, allowing us to filter out Pingdom, Appscan, etc
+  gaDimension: string;
```
gjuggler: Naming nitpick: since we may have *other* dimension in the future, I'd go with something like

"gaUserAgentDimension"

I'd also update the comment to include an explicit example, e.g. "dimension1". Just so it's extremely clear to everyone what the format should be :)

PR #1850, opened 2019-02-28
[RW-1955][RW-2114][risk=low] Initial ElasticFilters implementation

calbach:
Sending for initial feedback.

Future work:
- Handling of "attributes"
- Integration testing

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -77,7 +77,9 @@ public final class ElasticFilters {
   }
 
   /**
-   * A hand-maintained mapping of which criteria trees map to standard vs source concept IDs.
+   * A hand-maintained mapping of which criteria trees map to standard vs source concept IDs. This
+   * indicates which concept ID field we should be querying against for a SearchRequest on a given
+   * tree.
    * TODO(RW-2249): Document or encode this more centrally.
    */
   private static Set<TreeType> STANDARD_TREES = ImmutableSet.of(
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -8,17 +8,12 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Multimap;
-import com.google.common.collect.Multimaps;
-import com.google.common.collect.SetMultimap;
-import com.google.rpc.BadRequest;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import jdk.nashorn.internal.ir.annotations.Immutable;
 import joptsimple.internal.Strings;
 import org.apache.lucene.search.join.ScoreMode;
 import org.elasticsearch.index.query.BoolQueryBuilder;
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -0,0 +1,254 @@
...
+          // TODO: This two-step lookup is gross - dig into why we don't seem to have enough
+          // information available in the SearchRequest to more directly translate a criteria
+          // selection into a set of concept IDs. Also there is no uniqueness guarantee on codes
+          // nor do we seem to have documentation of what a "criteria code" is and where it comes
+          // from. Lastly, the code gets stuff into param.getValue() - this seems to be undocumented
+          // and unclear whether that field is polymorphic.
+          Criteria parent = criteriaDao.findCriteriaByTypeAndCodeAndGroup(param.getType(), param.getValue(), true);
+
+          // TODO: Try to batch this request for all search params.
+          Set<Long> children = criteriaDao.findCriteriaChildrenByTypeAndParentId(
```
freemabd1: I think we can lookup children of parent in 3 different ways depending on search parameter type:
1) if ICD9/ICD10 and is_group = 1
`select concept_id
  from criteria
  where type = :type
  and subtype = :subtype
  and code REGEXP '^(:value)' 
  and is_selectable = 1
  and concept_id != ''`
2) if PPI and is_group = 1 and concept_id is null
`select concept_id
      from criteria
      where path like (
      select concat('%', id, '%') as path
      from criteria
      where subtype = :subtype
      and parent_id = 0)`
3) if Drug and is_group = 1
`select a.concept_id from
    criteria a
    join (select CONCAT( '%.', id, '%') as path
    from criteria
    where concept_id in (:conceptId)) b
    on a.path like b.path
    and is_group = 0
    and is_selectable = 1
    and type = 'DRUG'
    and subtype = 'ATC'`

calbach: OK - I tried to generalize these cases out rather than tie them to specific tree types, see `buildCriteriaGroupLookup`. Let me know if this makes sense.

I decided to skip straight to minimal SQL calls rather than optimizing later, so the client code here is a bit more verbose. I think this is probably the effective minimal number of SQL calls we can make given our current SearchRequest requirements and SQL table schemas, without writing some crazy complex union query.

calbach:
This PR escalated a bit after the revelations from @freemabd1 - now has tests though. PTAL

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -0,0 +1,395 @@
...
+    Map<FullTreeType, Map<String, Set<Long>>> childrenByParentCode = Maps.newHashMap();
+    Map<FullTreeType, Map<Long, Set<Long>>> childrenByParentConcept = Maps.newHashMap();
+
+    // Within each category/tree type combination, we can batch MySQL requests to translate these
+    // groups into child concept IDs. First we build up maps to denote which groups to query and
+    // which will eventually hold the results (for now, we mark them with an empty set).
+    for (SearchGroup sg : Iterables.concat(req.getIncludes(), req.getExcludes())) {
+      for (SearchGroupItem sgi : sg.getItems()) {
+        for (SearchParameter param : sgi.getSearchParameters()) {
+          if (!param.getGroup()) {
```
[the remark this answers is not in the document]
freemabd1: If a CB search request contains #2 we only want to count participants that have answered anything for that specific question(criteria already has that question concept_id). If criteriaDao#findCriteriaLeavesByType is called in this case it will return all answers for specific survey(type and subtype). This is not intended.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -0,0 +1,395 @@
...
+              leaves.add(c);
+            }
+          });
+
+      for (Criteria c : leaves) {
+        // Technically this could scale poorly with many criteria groups. We don't expect this
+        // number to be very high as it requires a user action to add a group, but a better data
+        // structure could be used here if this becomes too slow.
+        for (Criteria parent : parents) {
+          if (c.getCode().startsWith(parent.getCode())) {
```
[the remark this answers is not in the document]
freemabd1: So the codes for ICD9/ICD10 and Drugs.. i can't really speak on who created them(Doctors, Hospitals or Accountants) but I can say that whomever created ICD9/ICD10 probably didn't create the drug codes. So I wouldn't hold my breath thinking that different vocabularies will have any similarities. 

Yes, value is only used for ICD9/ICD10 except for demographics -> pass deceased in value

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -0,0 +1,395 @@
...
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(type, subType);
+    }
+  }
+
+  private Map<SearchParameter, Set<Long>> buildCriteriaGroupLookup(SearchRequest req) {
+    // Three categories of criteria groups are currently supported in a SearchRequest:
+    // 1. Groups with only tree types specified, e.g. PPI.
```
calbach: Per Brian discussion make the order match here - this matches PPI survey.

calbach: doc TODOs:
- SearchParameter.vaue: ICD9/10 codes, deceased, PPI answer, (something else)
- Answers need to be handled in a follow-up
- Add indexing to README and examples of codes

calbach:
Added docs and found a bug in the PPI stuff. Confirmed the question queries are now working (will need to do more work to get answers working).

PR #1851, opened 2019-02-28
[RW-1948][risk=no] Add remaining domains to events index

calbach:
TODO:
- May need more indexing for physical measurements.

### DIFF
--- a/api/tools/src/main/resources/bigquery/es_person.sql
+++ b/api/tools/src/main/resources/bigquery/es_person.sql
@@ -9,22 +9,55 @@ SELECT
   ec.concept_name ethnicity_concept_name,
   condition_concept_ids,
   condition_source_concept_ids,
-  /* TODO(calbach): Expand this to UNION events across all domains. */
-  conditions AS events
+  observation_concept_ids,
+  observation_source_concept_ids,
+  drug_concept_ids,
+  drug_source_concept_ids,
+  procedure_concept_ids,
+  procedure_source_concept_ids,
+  measurement_concept_ids,
+  measurement_source_concept_ids,
+  visit_concept_ids,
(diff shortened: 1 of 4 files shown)


api/tools/src/main/resources/bigquery/es_person.sql
```diff
@@ -41,6 +77,102 @@ LEFT JOIN (
...
+  SELECT
+    m.person_id person_id,
+    ARRAY_AGG(DISTINCT measurement_concept_id) measurement_concept_ids,
+    ARRAY_AGG(DISTINCT measurement_source_concept_id) measurement_source_concept_ids,
+    ARRAY_AGG(STRUCT( measurement_concept_id AS concept_id,
+        measurement_source_concept_id AS source_concept_id,
+        measurement_date AS start_date,
+        DATE_DIFF(measurement_date, DATE(p.YEAR_OF_BIRTH, p.MONTH_OF_BIRTH, p.DAY_OF_BIRTH), YEAR) AS age_at_start_date,
+        v.visit_concept_id,
+        value_as_number,
```
calbach: @freemabd1 please double check this - added the value_as_number/concept

freemabd1: Looks good to me. Also, to cleanup data that is invalid/incorrectly coded should each join to domain table include all source or standard concept ids that match concept ids in the concept table?

calbach: Yikes - is that a real thing? Let me file a ticket to fix this in a follow-up; if I'm understanding correctly, it shouldn't cause any bugs (yet) to have invalid data here but it will bloat the index size.

Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-2242

PR #1852, opened 2019-02-28
[risk=no] convert breadcrumb to React

panentheos:
This reworks the breadcrumb logic, and converts it to react. Notes:
* The breadcrumb hierarchy is now explicitly implemented in the `getTrail` function, rather than being embedded in the hierarchy of the routes.
* Each leaf route now specifies a single token corresponding to the breadcrumb it should display.
* For the breadcrumbs that show names of objects that are loaded from the server, there is a fallback string in case the object is not there yet. This doesn't currently happen due to the way resolvers work, but as we refactor those, it will become relevant.
* There is a temporary duplication of information between the url strings in the router vs `getTrail`. In the future, once we have a system that allows us to refer to routes by name, we can use those to generate the breadcrumb URLs instead of hard coding them.


ui/src/app/utils/index.tsx
```diff
@@ -277,6 +277,16 @@ export const withCurrentWorkspace = () => {
   return connectBehaviorSubject(currentWorkspaceStore, 'workspace');
 };
 
+// HOC that provides a 'workspace' prop with current WorkspaceData
```
calbach: Stale doc, here and on the other changes in this file

ui/src/app/views/breadcrumb/component.tsx
```diff
@@ -0,0 +1,133 @@
...
+  },
+  lastLink: {
+    color: '#262262',
+    fontWeight: 600,
+    fontSize: '1rem',
+    textDecoration: 'none'
+  }
+};
+
+export const getTrail = (type, data) => {
```
calbach: Doc?

Return type? `type` as an enum instead?

ui/src/app/views/breadcrumb/component.tsx
```diff
@@ -0,0 +1,133 @@
...
+      ];
+    case 'workspaceClone':
+      return [
+        ...getTrail('workspace', data),
+        {label: 'Duplicate Workspace', url: `${prefix}/clone`}
+      ];
+    case 'notebook':
+      return [
+        ...getTrail('workspace', data),
+        {label: 'Notebooks', url: `${prefix}/notebooks`},
```
calbach: There was some intentional logic based on Lou's mocks implemented by the `intermediate: true` approach we had previously. AFAICT that's lost here, though I'm not 100% sure from tracing the code.

ui/src/app/cohort-review/routing/routing.module.ts
```diff
@@ -27,28 +28,19 @@ const routes: Routes = [{
...
     }
   }, {
     path: 'report',
     component: QueryReportComponent,
     data: {
-      breadcrumb: {
-        value: 'Query Report',
-        intermediate: false
-      }
+      breadcrumb: 'report'
```
calbach: This doesn't appear in `getTrail()` (I'm assuming it should). An enum rather than a string might avoid these desynchronizing and make the linkage clearer.

ui/src/app/utils/index.tsx
```diff
@@ -277,6 +277,16 @@ export const withCurrentWorkspace = () => {
...
+export const withCurrentCohort = () => {
+  return connectBehaviorSubject(currentCohortStore, 'cohort');
+};
+
+// HOC that provides a 'workspace' prop with current WorkspaceData
+export const withCurrentConceptSet = () => {
```
calbach: Tangent: General question on these HOC stores. (In cases like the breadcrumb this question is probably not applicable since it stays on the page, but more a question for the other content below the nav bar.)

Does this always work cleanly when there's a transition between contexts? For example, would this ever be rendered before the currentWorkspace() updated on a navigate - or similarly would currentWorkspace() ever trigger a state change during navigation while this component was still on the screen?

ui/src/app/utils/index.tsx
panentheos: Good catch, fixed

ui/src/app/views/breadcrumb/component.tsx
panentheos: Done.

ui/src/app/views/breadcrumb/component.tsx
panentheos: I was careful to capture this logic in the new version. For example, the `notebooks` page has the `Workspace` breadcrumb, while the `notebooks/foo` page has the `Notebook` breadcrumb, which contains both the former 'intermediate' crumb and the new leaf.

ui/src/app/cohort-review/routing/routing.module.ts
panentheos: Good catch, although it turns out this page isn't actually reachable any more, as its content is now shown as the second tab on the participants page. I removed the route entry instead.

ui/src/app/utils/index.tsx
panentheos: It depends on the exact case, but generally yes. React (and I think Angular too) unmounts components from leaves upward, so as long as the state in question is being managed by a component higher up in the tree, the child will be removed before it sees any changes.

Obviously if we're making a lateral move, e.g. going from one participant to the next, then the component will stay mounted and see the transition, but the component expects that.

ui/src/app/views/breadcrumb/component.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const prefix = `/workspaces/${ns}/${wsid}`;
+  switch (type) {
+    case 'workspaces':
+      return [
+        {label: 'Workspaces', url: '/workspaces'}
+      ];
+    case 'workspace':
+      return [
+        ...getTrail('workspaces', data),
+        {label: workspace ? workspace.name : '...', url: prefix}
```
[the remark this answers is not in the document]
panentheos: It's a fallback string, see the third bullet in the description.

ui/src/app/views/breadcrumb/component.tsx
```diff
@@ -0,0 +1,141 @@
...
+  withCurrentConceptSet(),
+  withUrlParams(),
+  withRouteConfigData()
+)(({workspace, cohort, conceptSet, urlParams, routeConfigData}) => {
+  const trail = getTrail(
+    routeConfigData.breadcrumb,
+    {urlParams, workspace, conceptSet, cohort}
+  );
+  const first = fp.dropRight(1, trail);
+  const last = fp.last(trail) as any;
```
calbach: Does this still need an `any` now that you typed it? Or does putting it through lodash drop the type info?

panentheos: Ah, looks like it's no longer needed, thanks.

PR #1857, opened 2019-02-28
[RW-1452][Risk=moderate] Working rainforest test script

blrubenstein:
This will attempt to schedule a run of all of the rainforest tests from a script. This currently will not work, because we do not have enough tabular rows for emails and passwords to execute them all in one command.

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -456,6 +456,19 @@ Common.register_command({
   :fn => ->(*args) { run_bigquery_tests("bigquerytest", *args) }
 })
 
+def run_rainforest_tests(cmd_name, *args)
+  ensure_docker cmd_name, args
+  common = Common.new
+  token = `gsutil cat gs://all-of-us-rw-staging-credentials/rainforest-key.txt`
+  common.run_inline %W{rainforest run all --token #{token}}
+end
+
+Common.register_command({
+  :invocation => "rainforesttest",
+  :description => "Runs rainforest tests.",
+  :fn => ->(*args) { run_rainforest_tests("rainforesttest", *args) }
+})
+
 def run_gradle(cmd_name, args)
   ensure_docker cmd_name, args
   begin
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -459,6 +459,9 @@ Common.register_command({
 def run_rainforest_tests(cmd_name, *args)
   ensure_docker cmd_name, args
   common = Common.new
+  # The bucket is hardcoded to staging, because that is currently the only
+  # environment we can run tests in. There is, however, an identical key in
+  # each of the other environments.
   token = `gsutil cat gs://all-of-us-rw-staging-credentials/rainforest-key.txt`
   common.run_inline %W{rainforest run all --token #{token}}
 end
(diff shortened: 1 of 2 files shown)


api/src/dev/server/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN apk --no-cache add \
         libc6-compat \
         openssh-client \
         git \
+        make \
```
calbach: Were these other added deps all required by the rainforest CLI?

api/libproject/devstart.rb
```diff
@@ -456,6 +456,19 @@ def run_bigquery_tests(cmd_name, *args)
   :fn => ->(*args) { run_bigquery_tests("bigquerytest", *args) }
 })
 
+def run_rainforest_tests(cmd_name, *args)
+  ensure_docker cmd_name, args
+  common = Common.new
+  token = `gsutil cat gs://all-of-us-rw-staging-credentials/rainforest-key.txt`
+  common.run_inline %W{rainforest run all --token #{token}}
```
calbach: Doesn't this need to be parameterized to run a particular test suite or in a particular environment?

api/libproject/devstart.rb
```diff
@@ -456,6 +456,19 @@ def run_bigquery_tests(cmd_name, *args)
   :fn => ->(*args) { run_bigquery_tests("bigquerytest", *args) }
 })
 
+def run_rainforest_tests(cmd_name, *args)
+  ensure_docker cmd_name, args
+  common = Common.new
+  token = `gsutil cat gs://all-of-us-rw-staging-credentials/rainforest-key.txt`
```
calbach: Is hardcoding to staging intentional? If so why? (please add a comment)

api/src/dev/server/Dockerfile
```diff
@@ -64,8 +65,12 @@ RUN unzip gradle.zip && rm gradle.zip \
 ENV PATH="$PATH:/gradle/bin"
 WORKDIR /
 
+RUN apk add gcc libc-dev g++
+
 RUN gem install --no-document googleauth
 
+RUN gem install rainforest-cli
```
calbach: FYI: you're likely to need this here as well: https://github.com/all-of-us/workbench/blob/f35b531e374fb40b59d41e7b1f69a93a49ca9083/ci/Dockerfile.circle_build

Doesn't need to be this PR though - should probably wait until you're hooking this up to circle so you can test it.

api/libproject/devstart.rb
blrubenstein: I did want to hardcode to staging, because thats the environment that Rainforest runs against. We do have an identical key in each environment though.

api/libproject/devstart.rb
blrubenstein: It may need to, I want to talk to Alex Minor about our strategy around this. Right now we can just run in staging, so environment can't be parameterized.

api/src/dev/server/Dockerfile
blrubenstein: Yes

api/src/dev/server/Dockerfile
blrubenstein: Yeah, I think that should go into a PR for the circle integration.

PR #1859, opened 2019-02-28
[RW-2098][Risk=no] View and Select Concept Sets in DataSet Builder

UrsaStutsman:
All comments addressed. Going to work on tests. And going to change this a regular PR so that if you have time and want to approve you can go ahead and do that.

### DIFF
--- a/ui/src/app/views/dataset/component.tsx
+++ b/ui/src/app/views/dataset/component.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import {Button} from 'app/components/buttons';
 import {FadeBox} from 'app/components/containers';
 import {ClrIcon} from 'app/components/icons';
+import {ResourceListItem} from 'app/components/resources';
 import {SpinnerOverlay} from 'app/components/spinners';
 import {WorkspaceData} from 'app/resolvers/workspace';
 import {cohortsApi, conceptsApi, conceptSetsApi} from 'app/services/swagger-fetch-clients';
@@ -14,7 +15,6 @@ import {convertToResource, ResourceType} from 'app/utils/resourceActionsReact';
 import {CreateConceptSetModal} from 'app/views/conceptset-create-modal/component';
 import {ConfirmDeleteModal} from 'app/views/confirm-delete-modal/component';
 import {EditModal} from 'app/views/edit-modal/component';
-import {ResourceCardMenu} from 'app/views/resource-card/component';
 import {
   Cohort,
   ConceptSet,
@@ -40,29 +40,12 @@ export const styles = {
   }
 };
 
-const ResourceListItem: React.FunctionComponent <
-  {conceptSet: ConceptSet, openConfirmDelete: Function, edit: Function}
-  > = ({conceptSet, openConfirmDelete, edit}) => {
-    return<div style={{border: '0.5px solid #C3C3C3', margin: '.4rem',
-      height: '1.5rem', display: 'flex'}}>
-      <div style={{width: '.75rem', paddingTop: 5, paddingLeft: 10}}>
-        <ResourceCardMenu disabled={false}
-                          resourceType={ResourceType.CONCEPT_SET}
-                          onDeleteResource={openConfirmDelete}
-                          onEditConceptSet={edit}/>
-      </div>
-      <input type='checkbox' value={conceptSet.name} style={{height: 17, width: 17,
-        marginLeft: 10, marginTop: 10, marginRight: 10, backgroundColor: '#7CC79B'}}/>
-      <div style={{lineHeight: '1.5rem'}}>{conceptSet.name}</div>
-    </div>;
-  };
-
 export const DataSet = withCurrentWorkspace()(class extends React.Component<
   {workspace: WorkspaceData},
   {creatingConceptSet: boolean, conceptDomainList: DomainInfo[],
     conceptSetList: ConceptSet[], loadingConceptSets: boolean,
     confirmDeleting: boolean, editing: boolean, resource: RecentResource,
-    rType: ResourceType
+    rType: ResourceType, selectedConceptSets: ConceptSet[]
   }> {
 
   constructor(props) {
@@ -75,7 +58,8 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
       confirmDeleting: false,
       editing: false,
       resource: undefined,
-      rType: undefined
+      rType: undefined,
+      selectedConceptSets: []
     };
   }
 
@@ -101,7 +85,6 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
   convertResource(r: ConceptSet | Cohort, rType: ResourceType): RecentResource {
     const {workspace} = this.props;
     this.setState({rType: rType});
-    console.log(rType);
     return convertToResource(r, workspace.namespace, workspace.id,
       workspace.accessLevel as unknown as WorkspaceAccessLevel, rType);
   }
@@ -113,18 +96,10 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
 
   closeConfirmDelete(): void {
     this.setState({confirmDeleting: false});
-    if (this.state.rType === ResourceType.CONCEPT_SET) {
-      const deleted = this.state.resource.conceptSet;
-      const updatedList = fp.filter(conceptSet => conceptSet.id !== deleted.id,
-        this.state.conceptSetList);
-      this.setState({conceptSetList: updatedList});
-    }
-    this.setState({resource: undefined, rType: undefined});
   }
 
   edit(r: Cohort | ConceptSet, rType: ResourceType): void {
     const rc = this.convertResource(r, rType);
-    console.log(rType);
     this.setState({editing: true, resource: rc});
   }
 
@@ -142,7 +117,13 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
     } else {
       call = cohortsApi().deleteCohort(...argList);
     }
-
+    if (this.state.rType === ResourceType.CONCEPT_SET) {
+      const deleted = this.state.resource.conceptSet;
+      const updatedList = fp.filter(conceptSet => conceptSet.id !== deleted.id,
+        this.state.conceptSetList);
+      this.setState({conceptSetList: updatedList});
+    }
+    this.setState({resource: undefined, rType: undefined});
     call.then(() => this.closeConfirmDelete());
   }
 
@@ -169,6 +150,23 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
     });
     this.setState({editing: false, conceptSetList: updatedList,
       resource: undefined, rType: undefined});
+  }
+
+  select(resource: ConceptSet | Cohort, rtype: ResourceType): void {
+    if (rtype === ResourceType.CONCEPT_SET) {
+      const origSelected = this.state.selectedConceptSets;
+      const selected = fp.filter((c) => c.id === resource.id, origSelected);
+      if (selected.length > 0) {
+        this.setState({selectedConceptSets: fp.remove((c) => c.id === resource.id, origSelected)});
+      } else {
+        this.setState({selectedConceptSets: origSelected.concat(origSelected)});
+      }
+    } else {
+      // else = cohort
+      //  determine if already in selected
+      //  remove if it is
+      //  add if it isn't
+    }
   }
 
   getCurrentResource(): Cohort | ConceptSet {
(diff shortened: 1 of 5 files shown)


ui/src/app/views/dataset/component.tsx
```diff
@@ -31,17 +40,42 @@ export const styles = {
   }
 };
 
+const ResourceListItem: React.FunctionComponent <
```
blrubenstein: Nice helper component here! Do you think we will need these list items elsewhere on the site? It may be worth making it a common component instead of inside the dataset builder file.

ui/src/app/views/dataset/component.tsx
```diff
@@ -31,17 +40,42 @@ export const styles = {
...
+  {conceptSet: ConceptSet, openConfirmDelete: Function, edit: Function}
+  > = ({conceptSet, openConfirmDelete, edit}) => {
+    return<div style={{border: '0.5px solid #C3C3C3', margin: '.4rem',
+      height: '1.5rem', display: 'flex'}}>
+      <div style={{width: '.75rem', paddingTop: 5, paddingLeft: 10}}>
+        <ResourceCardMenu disabled={false}
```
blrubenstein: On the mocks, concept sets appear not to have the dropdown menu. That said, I think they probably should? We should check with Lou to confirm here.

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      this.setState({conceptSetList: conceptSets.items, loadingConceptSets: false});
     } catch (error) {
       console.log(error);
     }
   }
 
+  convertResource(r: ConceptSet | Cohort, rType: ResourceType): RecentResource {
+    const {workspace} = this.props;
+    this.setState({rType: rType});
+    console.log(rType);
```
blrubenstein: Nit: rm console log

ui/src/app/views/dataset/component.tsx
```diff
@@ -31,17 +40,42 @@ export const styles = {
...
+  > = ({conceptSet, openConfirmDelete, edit}) => {
+    return<div style={{border: '0.5px solid #C3C3C3', margin: '.4rem',
+      height: '1.5rem', display: 'flex'}}>
+      <div style={{width: '.75rem', paddingTop: 5, paddingLeft: 10}}>
+        <ResourceCardMenu disabled={false}
+                          resourceType={ResourceType.CONCEPT_SET}
+                          onDeleteResource={openConfirmDelete}
+                          onEditConceptSet={edit}/>
+      </div>
+      <input type='checkbox' value={conceptSet.name} style={{height: 17, width: 17,
```
blrubenstein: I think it would probably valuable for this ticket to also cover managing selected concept sets. I don't think we can do anything with them yet, but it would probably be good to work out that behavior sooner rather than later.

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+  openConfirmDelete(r: ConceptSet, rType: ResourceType): void {
+    const rc = this.convertResource(r, rType);
+    this.setState({confirmDeleting: true, resource: rc});
+  }
+
+  closeConfirmDelete(): void {
+    this.setState({confirmDeleting: false});
+    if (this.state.rType === ResourceType.CONCEPT_SET) {
+      const deleted = this.state.resource.conceptSet;
+      const updatedList = fp.filter(conceptSet => conceptSet.id !== deleted.id,
```
blrubenstein: Interesting question on this one (I don't know the answer): On deleting, should we just filter out of the list, or reload the conceptSetList? I don't know what we do elsewhere, but we should be consistent. (Tradeoff being that this is faster, but means a user is more likely to be left with a stale list.)

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+        this.state.conceptSetList);
+      this.setState({conceptSetList: updatedList});
+    }
+    this.setState({resource: undefined, rType: undefined});
+  }
+
+  edit(r: Cohort | ConceptSet, rType: ResourceType): void {
+    const rc = this.convertResource(r, rType);
+    console.log(rType);
```
blrubenstein: Nit: rm console log

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+    return convertToResource(r, workspace.namespace, workspace.id,
+      workspace.accessLevel as unknown as WorkspaceAccessLevel, rType);
+  }
+
+  openConfirmDelete(r: ConceptSet, rType: ResourceType): void {
+    const rc = this.convertResource(r, rType);
+    this.setState({confirmDeleting: true, resource: rc});
+  }
+
+  closeConfirmDelete(): void {
```
blrubenstein: Technically it should only do this if you actually click confirm, correct? Semantically this seems like it is called no matter how the confirm delete modal is closed. (Maybe: `onConfirmDelete`?)

Or at least, the filtering out of the list should only be on delete.

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+  receiveDelete() {
+    const {resource, rType} = this.state;
+    let call;
+    const id = rType === ResourceType.COHORT ? resource.cohort.id : resource.conceptSet.id;
+    const argList = [resource.workspaceNamespace,
+      resource.workspaceFirecloudName, id];
+    if (rType === ResourceType.CONCEPT_SET) {
+      // these also throw errors in the logs
+      // functions correctly, and I think is so pretty, but
+      // I don't like all the errors in the logs, so torn.
```
blrubenstein: I think errors in the logs is fine for now, if they are real errors. Eventually we will probably want to handle more cleanly. 

Do they throw errors even on success?

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      call = conceptSetsApi().updateConceptSet(...argList);
+    }
+    call.then(() => this.closeEditModal());
+  }
+
+  closeEditModal(): void {
+    const edited = this.state.resource.conceptSet;
+    const updatedList = this.state.conceptSetList.map((conceptSet) => {
+      return edited.id === conceptSet.id ? edited : conceptSet;
+    });
```
blrubenstein: Likewise to above, we should figure out whether we want to issue a reload or do the in-browser update. The only reason I bring it up is that I think we've done the reload in the past.

ui/src/app/views/dataset/component.tsx
```diff
@@ -58,15 +92,102 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+    this.setState({editing: false, conceptSetList: updatedList,
+      resource: undefined, rType: undefined});
+  }
+
+  getCurrentResource(): Cohort | ConceptSet {
+    if (this.state.resource) {
+      return fp.compact([this.state.resource.cohort, this.state.resource.conceptSet])[0];
```
blrubenstein: Neat! This is an awesome helper.

ui/src/app/views/dataset/component.tsx
```diff
@@ -133,6 +266,18 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
       onClose={() => {this.setState({ creatingConceptSet: false}); }}
       conceptDomainList={conceptDomainList}
       existingConceptSets={conceptSetList}/>}
+      {console.log(resource)}
```
blrubenstein: Nit: rm console log line.

ui/src/app/views/dataset/component.tsx
```diff
@@ -133,6 +266,18 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      {this.state.confirmDeleting &&
+      // This throws a type error in the logs, figure out how to fix that
+      //  Functions as it's supposed to
```
blrubenstein: you might be able to use the cast `____ as unknown as _____` pattern? Is it from fetch vs generated?

ui/src/app/views/resource-card/component.tsx
```diff
@@ -19,13 +19,14 @@ import {Domain, RecentResource} from 'generated/fetch';
 import {cohortsApi, conceptSetsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import {environment} from 'environments/environment';
 
-const ResourceCardMenu: React.FunctionComponent<{
+export const ResourceCardMenu: React.FunctionComponent<{
```
blrubenstein: Now that this is used multiple places in different ways, we may want to move it to a common component, and maybe rename as `ResourceMenu`

ui/src/app/views/dataset/component.tsx
UrsaStutsman: I'd been thinking the same thing! But not sure if it will be used other places, maybe through it in a resource related file? I'll go through our dirs and see if I can think of a good place for it.

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Yeah I wasn't sure, it looked like _some_ of them had it, and it does say on the card to include it...

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Ok, will definitely add that in then, wasn't sure.

ui/src/app/views/dataset/component.tsx
UrsaStutsman: This is the way I'd _like_ to have it implemented. It's much faster, and since we have etags to prevent updates to stale workspaces I feel like it's safe.

ui/src/app/views/dataset/component.tsx
UrsaStutsman: This was one of the extra thoughts I had after putting this up last night. I will sort this logic appropriately. (just close if cancel hit, etc.)

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Not sure what you mean by "on success", but from what I can tell about the UI tests in circle is that this fails strict compilation, so will probably have to pull it out. Going to consult with Brett first though.

ui/src/app/views/dataset/component.tsx
UrsaStutsman: And I think that's a bad reason to stick with something, I think moving forward we should use this, and have a tech debt ticket to change the other places.... Maybe should be a tech talk subject to get consensus from the team.

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Typescript compilation errors.

ui/src/app/views/resource-card/component.tsx
UrsaStutsman: Yup and maybe to where ever I put that resourceList thing too...

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Fixed

ui/src/app/views/dataset/component.tsx
UrsaStutsman: Fixed for both modals

ui/src/app/views/dataset/component.tsx
UrsaStutsman: fixed

ui/src/app/views/dataset/component.tsx
blrubenstein: Totally reasonable, I do think that this is probably the better way to do this, but it depends. If there’s a lot of people likely making edits on the same workspace/concept sets, then reloading is probably the way to do. A quick sync around it makes sense, but I’m comfortable going this way until we have a decision, and then adding a tech debt ticket to do cleanup one way or the other.

ui/src/app/views/dataset/component.tsx
blrubenstein: I’d also look and see if the mocks reuse this component at all. If they do, maybe `common/resources.tsx`?

UrsaStutsman: Put it in components/resources.tsx? Let me know if you want me to move it common/

blrubenstein: I think components/resources.tsx is a totally reasonable place for it!

PR #1861, opened 2019-03-01
[risk=no] refactor away resolvers

panentheos:
This removes the router resolvers in favor of fetching the data inside the various wrapper components, and having them gate the rendering of children as appropriate. This brings things more in line with the typical React approach, and removes more router dependencies.

Notes:
* Instead of the dark routing spinner that blocks navigation, page transitions happen sooner, and there is an inline spinner while e.g. the workspace object is loaded. This should be a largely equivalent UX. (and in some cases better, because we can show e.g. the workspace tab bar earlier, so the page feels a bit more complete while it's loading).
* Because the transition happens earlier, the breadcrumb will sometimes show the new path before the underlying object is loaded, so it will briefly show a fallback string in place of a name. This is currently '...', but we could change that if desired.


ui/src/app/cohort-review/page-layout/page-layout.ts
```diff
@@ -1,31 +1,50 @@
...
+  ) {}
 
   ngOnInit() {
-    const review = cohortReviewStore.getValue();
+    const {ns, wsid, cid} = urlParamsStore.getValue();
+    const cdrid = +(currentWorkspaceStore.getValue().cdrVersionId);
+    this.reviewAPI.getParticipantCohortStatuses(ns, wsid, cid, cdrid, {
+      page: 0,
+      pageSize: 25,
+      sortColumn: ParticipantCohortStatusColumns.PARTICIPANTID,
```
calbach: Existing issue, but: `sortColumn` seem to be a real request field. Should be able to remove and then subsequently this should allow you to remove the below `PageFilterRequest` cast

ui/src/app/cohort-review/page-layout/page-layout.ts
```diff
@@ -1,31 +1,50 @@
...
+      cohortReviewStore.next(review);
+      if (review.reviewStatus === ReviewStatus.NONE) {
+        this.reviewPresent = false;
+      } else {
+        this.reviewPresent = true;
+        navigate(['workspaces', ns, wsid, 'cohorts', cid, 'review', 'participants']);
+      }
+    });
+    this.cohortsAPI.getCohort(ns, wsid, cid).subscribe(cohort => {
+      currentCohortStore.next(cohort);
```
calbach: I'm a bit concerned with this callsite pushing global state. SHould we consider moving this to the same location as the `currentWorkspaceStore` updates?

There are other components that depend on the `currentCohortStore` value, so now it feels like there some implicit dependencies on this component being rendered at the same time as all of those (it's also not clear in which component I can or cannot use `currentCohortStore`). Alternatively if this component is indeed meant to wrap anything with a cohort context, perhaps it should pass the cohort value down to its children as a property instead.

panentheos: It's good to call this out, since it's a bit outside the ordinary approach. Essentially, this component is now acting as the resolver, and is the sole manager of this state. And since it's a wrapper around all cohort-related pages, any component within that part of the hierarchy can expect to be able to connect to this store and get see up-to-date info. In that sense, it's actually relatively straightforward. A bigger issue, in my opinion, is the `cohortReviewStore`, which has multiple writers that push data at different times, and no clear owner. That stems from an underlying UX design issue, though.

Back to this store. Yes, we could just pass the cohort directly down to children, with one exception: the breadcrumb. Since that's rendered higher up in the tree, the only way for child components to affect it is via some sort of push mechanism. So we'd still need something like this, even if the rest of the data flow was made more standard.

To summarize: I think this is a reasonable refactoring step, since it addresses the underlying needs and has relatively low complexity. We can certainly continue to evolve this, and could also rethink the breadcrumb rendering in the future if desired.

calbach: I see, this component is pretty much playing the same role as `with-workspace`. I was a bit thrown by the naming here since "cohort-review" doesn't really encompass every context where you might need to load/update a cohort (e.g. cohort clone page?) - though maybe it covers everything that exists today.

Could you capture this in a comment here describing how we would go about extending this approach and any current limitations? I can imagine a future dev being a bit lost figuring out how to add a new component which needs a cohort (and introducing bugs).

ui/src/app/cohort-review/page-layout/page-layout.html
```diff
@@ -2,8 +2,9 @@
   CohortReview main content outlet
   Pages: Overview, Table (participant master), Detail
 -->
-<div class="cohort-review-wrapper">
+<div class="cohort-review-wrapper" *ngIf="cohortLoaded && reviewPresent === true">
```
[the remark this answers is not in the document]
panentheos: Since these three items are mutually exclusive with one another, I think it's clearer to have them on the same level.

PR #1871, opened 2019-03-04
[risk=no] Handling cdr version name in init-new-cdr script

SrushtiGangireddy:
1. Adding the version-flag in script to add necessary permission to public user based on the value in flag.

### DIFF
--- a/api/db-cdr/generate-cdr/generate-cloudsql-db.sh
+++ b/api/db-cdr/generate-cdr/generate-cloudsql-db.sh
@@ -46,13 +46,20 @@ then
   exit 1
 fi
 
+if [[ ${BUCKET} =~ .*public.* ]]
+then
+  VERSION_FLAG="public"
+else
+  VERSION_FLAG="cdr"
+fi
+
 startDate=$(date)
 echo " Starting generate-cloudsql-db $DATABASE from bucket $BUCKET $startDate"
 
 
 # Init the local database
 echo "Initializing new  $DATABASE"
-if ./generate-cdr/init-new-cdr-db.sh --drop-if-exists --cdr-db-name ${DATABASE}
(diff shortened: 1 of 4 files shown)


api/db-cdr/generate-cdr/generate-cloudsql-db.sh
```diff
@@ -46,13 +46,20 @@ then
...
+fi
+
 startDate=$(date)
 echo " Starting generate-cloudsql-db $DATABASE from bucket $BUCKET $startDate"
 
 
 # Init the local database
 echo "Initializing new  $DATABASE"
-if ./generate-cdr/init-new-cdr-db.sh --drop-if-exists --cdr-db-name ${DATABASE}
+if ./generate-cdr/init-new-cdr-db.sh --drop-if-exists --cdr-db-name ${DATABASE} --version-flag ${VERSION_FLAG}
```
calbach: A few more callsites here to update: https://github.com/all-of-us/workbench/blob/72b5eafff9700cc4587094597bdd5705d19fbffb/api/libproject/devstart.rb

api/libproject/devstart.rb
```diff
@@ -194,8 +194,8 @@ def run_local_migrations()
     common.run_inline %W{./run-migrations.sh main}
   end
   Dir.chdir('db-cdr/generate-cdr') do
-    common.run_inline %W{./init-new-cdr-db.sh --cdr-db-name cdr}
-    common.run_inline %W{./init-new-cdr-db.sh --cdr-db-name public}
+    common.run_inline %W{./init-new-cdr-db.sh --cdr-db-name cdr --version-flag cdr}
```
calbach: Also please update all the calls that hit this function: https://github.com/all-of-us/workbench/blob/72b5eafff9700cc4587094597bdd5705d19fbffb/api/libproject/devstart.rb#L88

SrushtiGangireddy: Updated. Tested with local data migrations. Merging in.

PR #1872, opened 2019-03-04
[RW-2235][Risk=low] Add era commons fields to db

blrubenstein:
@jacmrob to make sure the UI eRA commons stuff all looks good.
@NehaBroad to make sure all the API eRA commons stuff all looks good.
@gjuggler to review the overall approach as a foundation for future data access frameworks.

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -51,8 +51,6 @@
     "host": "aoudev.nnlm.gov",
     "enableMoodleBackend": true
   },
-  // The access object specifies whether each of the following access requirements block access
-  // to the workbench.
   "access": {
     "enableComplianceTraining": false,
     "enableEraCommons": false
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -50,5 +50,11 @@
   "moodle": {
     "host": "aoudev.nnlm.gov",
     "enableMoodleBackend": true
+  },
+  // The access object specifies whether each of the following access requirements block access
+  // to the workbench.
+  "access": {
+    "enableComplianceTraining": false,
+    "enableEraCommons": false
   }
 }
--- a/api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
+++ b/api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
@@ -6,8 +6,9 @@
                     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
   <changeSet author="blrubenst" id="db.changelog-57-add-era-commons-completion-time">
     <addColumn tableName="user">
-      <column name="era_linked_nih_username" type="varchar(255)"/>
-      <column name="era_link_expire_time" type="datetime"/>
+      <column name="era_commons_linked_nih_username" type="varchar(255)"/>
+      <column name="era_commons_link_expire_time" type="datetime"/>
+      <column name="era_commons_completion_time" type="datetime"/>
     </addColumn>
   </changeSet>
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
+++ b/api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="blrubenst" id="db.changelog-57-add-era-commons-completion-time">
+    <addColumn tableName="user">
+      <column name="era_linked_nih_username" type="varchar(255)"/>
+      <column name="era_link_expire_time" type="datetime"/>
+    </addColumn>
+  </changeSet>
+</databaseChangeLog>
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -503,15 +503,21 @@ public class ProfileController implements ProfileApiDelegate {
     User user = userProvider.get();
     NihStatus nihStatus = fireCloudService.getNihStatus();
     if (nihStatus != null) {
+      Timestamp eraCommonsCompletionTime = user.getEraCommonsCompletionTime();
       if ((nihStatus.getLinkedNihUsername() != null &&
-          nihStatus.getLinkedNihUsername() != user.getEraCommonsLinkedNihUsername()) ||
+          !nihStatus.getLinkedNihUsername().equals(user.getEraCommonsLinkedNihUsername())) ||
           nihStatus.getLinkExpireTime() != user.getEraCommonsLinkExpireTime().getTime()) {
-        user.setEraCommonsCompletionTime(new Timestamp(clock.instant().toEpochMilli()));
+        eraCommonsCompletionTime = new Timestamp(clock.instant().toEpochMilli());
+      } else if (nihStatus.getLinkedNihUsername() == null) {
+        eraCommonsCompletionTime = null;
       }
-      user.setEraCommonsLinkedNihUsername(nihStatus.getLinkedNihUsername());
-      user.setEraCommonsLinkExpireTime(new Timestamp(nihStatus.getLinkExpireTime()));
+      user = userService.setEraCommonsStatus(
+          nihStatus.getLinkedNihUsername(),
+          new Timestamp(nihStatus.getLinkExpireTime()),
+          eraCommonsCompletionTime
+      );
     } else {
-      user.setEraCommonsCompletionTime(null);
+      user = userService.setEraCommonsStatus(null, null, null);
     }
     userDao.save(user);
     return getProfileResponse(user);
@@ -725,6 +731,8 @@ public class ProfileController implements ProfileApiDelegate {
             !nihStatus.getLinkedNihUsername().equals(user.getEraCommonsLinkedNihUsername())) ||
             nihStatus.getLinkExpireTime() != user.getEraCommonsLinkExpireTime().getTime()) {
           eraCommonsCompletionTime = new Timestamp(clock.instant().toEpochMilli());
+        } else if (nihStatus.getLinkedNihUsername() == null) {
+          eraCommonsCompletionTime = null;
         }
         user = userService.setEraCommonsStatus(
             nihStatus.getLinkedNihUsername(),
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -502,23 +502,7 @@ public class ProfileController implements ProfileApiDelegate {
   public ResponseEntity<Profile> syncEraCommonsStatus() {
     User user = userProvider.get();
     NihStatus nihStatus = fireCloudService.getNihStatus();
-    if (nihStatus != null) {
-      Timestamp eraCommonsCompletionTime = user.getEraCommonsCompletionTime();
-      if ((nihStatus.getLinkedNihUsername() != null &&
-          !nihStatus.getLinkedNihUsername().equals(user.getEraCommonsLinkedNihUsername())) ||
-          nihStatus.getLinkExpireTime() != user.getEraCommonsLinkExpireTime().getTime()) {
-        eraCommonsCompletionTime = new Timestamp(clock.instant().toEpochMilli());
-      } else if (nihStatus.getLinkedNihUsername() == null) {
-        eraCommonsCompletionTime = null;
-      }
-      user = userService.setEraCommonsStatus(
-          nihStatus.getLinkedNihUsername(),
-          new Timestamp(nihStatus.getLinkExpireTime()),
-          eraCommonsCompletionTime
-      );
-    } else {
-      user = userService.setEraCommonsStatus(null, null, null);
-    }
+    userService.setEraCommonsStatus(nihStatus);
     userDao.save(user);
     return getProfileResponse(user);
   }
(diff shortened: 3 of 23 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -493,15 +488,26 @@ private User initializeUserIfNeeded() {
...
       throw new ServerErrorException(e);
     }
-    return ResponseEntity.ok(profile);
+    return getProfileResponse(user);
+  }
+
+  @Override
+  public ResponseEntity<Profile> syncEraCommonsStatus() {
+    User user = userProvider.get();
+    NihStatus nihStatus = fireCloudService.getNihStatus();
```
NehaBroad: Question: Can user linked nih username change? I guess i am trying to think if it will be usefull to check if the username already exist in DB and set it accordingly. Thoughts?

NehaBroad: But since we are not making a special call to get the username i think its fine

blrubenstein: I don't believe it can change, but I am unsure...? This is meant to update with what is in FC though, so I'm inclined to do a full update.

api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="blrubenst" id="db.changelog-57-add-era-commons-completion-time">
+    <addColumn tableName="user">
+      <column name="era_linked_nih_username" type="varchar(255)"/>
```
gjuggler: Let's try to be consistent with a single module name prefix for all columns.

In the doc I'd proposed "eraCommons" for the prefix, which would lead to column names like:

era_commons_linked_nih_username
era_commons_link_expire_time

If we wanted to shorten the module name to "era_link" that's fine too – but needs to be consistent:

era_link_nih_username
era_link_expire_time

(My $.02 is that "era commons" is generally thought of as an atomic phrase, so it would feel a bit awkward to shorten to just "era link". I don't have super strong feelings on the name – but we do need to be internally consistent.)

api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
```diff
@@ -0,0 +1,13 @@
...
+      <column name="era_link_expire_time" type="datetime"/>
```
gjuggler: I think we're missing a third column here: the core "moduleCompletionTime" column, which tracks the last timestamp at which a given access module was known to be completed by a user.

This column would be named like

era_commons_completion_time

or

era_link_completion_time

depending on which prefix we choose (see above comment).

This column would capture the timestamp when the user goes from an unlinked to linked state. And would be cleared when going from a linked to unlinked state. This and the bypass field would be guaranteed to be present for all access modules, and this is what we'd be looking at from L105 of UserService.java.

The motivation here was that by ensuring all access modules are modeled with this bare minimum of similarly-typed fields, it becomes much clearer to treat these different modules in a consistent way. (Versus the alternative, where one module has to be checked against an empty string, another module is a boolean, another is an enum, etc... which quickly becomes unnecessary mental burden to grok)

api/config/config_local.json
```diff
@@ -50,5 +50,11 @@
   "moodle": {
     "host": "aoudev.nnlm.gov",
     "enableMoodleBackend": true
+  },
+  // The access object specifies whether each of the following access requirements block access
```
gjuggler: Thoughts on comment placement: I wonder if it might be easier to maintain this comment as a single code-comment in WorkbenchConfig.java, rather than replicated across each of the config files here.

api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
blrubenstein: link vs linked is mirroring what FC/Shibboleth returns. I was just doing the prefix/module as `era`.

ui/src/app/views/homepage/component.tsx
```diff
@@ -334,8 +334,13 @@ export const Homepage = withUserProfile()(class extends React.Component<
           this.setState({firstVisitTraining: false});
         }
       }
-
-      this.setState({eraCommonsLinked: !!profile.linkedNihUsername});
+      try {
+        const syncEraCommonsStatus = await profileApi().syncEraCommonsStatus();
```
jacmrob: does this call firecloud?  if not why not just use the `profileState` prop?

blrubenstein: It does, but I realized that we actually don't need to sync here if we update when the callback from NIH happens.

api/db/changelog/db.changelog-57-add-era-commons-completion-time.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="blrubenst" id="db.changelog-57-add-era-commons-completion-time">
```
NehaBroad: nit: the file should be rename to changelog-57-add-era-commons-entries or something not just restricting to completion time

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -493,15 +488,33 @@ private User initializeUserIfNeeded() {
...
+    if (nihStatus != null) {
+      if ((nihStatus.getLinkedNihUsername() != null &&
+          nihStatus.getLinkedNihUsername() != user.getEraCommonsLinkedNihUsername()) ||
+          nihStatus.getLinkExpireTime() != user.getEraCommonsLinkExpireTime().getTime()) {
+        user.setEraCommonsCompletionTime(new Timestamp(clock.instant().toEpochMilli()));
+      }
+      user.setEraCommonsLinkedNihUsername(nihStatus.getLinkedNihUsername());
```
NehaBroad: What should be the completion time if linkedNihUsername is null?

blrubenstein: If linkedNihUsername is null, completion time should be null. I was acting under the assumption that shibboleth will return an error if there is no username? But don't know if that's correct in truth.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -701,11 +714,29 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
+            nihStatus.getLinkExpireTime() != user.getEraCommonsLinkExpireTime().getTime()) {
+          eraCommonsCompletionTime = new Timestamp(clock.instant().toEpochMilli());
+        }
+        user = userService.setEraCommonsStatus(
+            nihStatus.getLinkedNihUsername(),
+            new Timestamp(nihStatus.getLinkExpireTime()),
+            eraCommonsCompletionTime
+        );
+      } else {
+        user = userService.setEraCommonsStatus(null, null, null);
```
NehaBroad: Do we need this or will these values be explicitly null we initialize the user here ?

blrubenstein: Initialize user if needed won't always create a new user. If there is already a user it will do nothing, so we still need this set.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -101,8 +101,9 @@ private User updateWithRetries(Function<User, User> userModifier, User user) {
 
   private void updateDataAccessLevel(User user) {
     boolean shouldBeRegistered = Optional.ofNullable(user.getIdVerificationIsValid()).orElse(false)
+        && (!(configProvider.get().access.enableEraCommons) || user.getEraCommonsCompletionTime() != null)
```
gjuggler: This is starting to look really clean – thanks for adding the enableComplianceTraining env flag while you're here, it really helps to set the pattern. 👍

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -210,6 +211,18 @@ public User apply(User user) {
     });
   }
 
+  public User setEraCommonsStatus(String linkedNihUsername, Timestamp linkExpireTime, Timestamp completionTime) {
```
gjuggler: I'm a little unclear on the purpose of this method. In the syncEraCommonsStatus method on ProfileController, we simply call userDao.save(user). Is there a reason to have this separate method set up? (If so... it's worth a method-level comment to help guide the user.)

Just looking around at other methods in this file... one thing that ended up happening for the compliance training module was that we wanted to have ProfileController and UserController call a single service-level method to execute the sync for a given user. So the full logic for syncing a module's state lives within UserService. That's not an unreasonable way to approach this – but would involve moving all of the logic from ProfileController to here.

api/src/main/resources/workbench.yaml
```diff
@@ -2072,10 +2091,10 @@ definitions:
         description: Timestamp when the user requests identity verification.
       twoFactorEnabled:
         type: boolean
-      linkedNihUsername:
+      eraCommonsLinkedNihUsername:
         type: string
         description: The user's NIH username
-      linkExpireTime:
+      eraCommonsLinkExpireTime:
```
gjuggler: Maybe we should include the eraCommonsCompletionTime here? We do include it for compliance training (see L2066), so it would be nice to include for consistency. (Plus, I could imagine eventually showing users the timestamp when they completed each requirement in a tooltip or similar.)

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -753,7 +753,7 @@ public void reviewIdVerification_sendsEmail() throws Exception {
 
   @Test
   public void testUpdateNihToken() {
-    doNothing().when(fireCloudService).postNihCallback(any());
```
gjuggler: It's a little disconcerting that we don't have any tests explicitly checking the logic of the era commons sync.

What I'd hope to see would be something akin to the testSyncTraining method below – which explicitly walks through a sync call, with a mocked-out badge response, and verifies that we update the user data model accordingly.

Feel free to push back if there's some reason it's harder to test the NIH service in the same way. But for consistency, it would be nice to have something in the same pattern.

ui/src/app/guards/access-tasks-guard.service.ts
```diff
@@ -25,7 +25,8 @@ export class AccessTasksGuard implements CanActivate, CanActivateChild {
     }
     return this.serverConfigService.getConfig().flatMap((config) => {
       return this.profileStorageService.profile$.flatMap((profile) => {
-        if ((profile.linkedNihUsername === null || profile.linkedNihUsername === '')
+        if ((profile.eraCommonsLinkedNihUsername === null
```
gjuggler: Optional nitpick: It would be nice to check the same field as we check in the backend – eraCommonsCompletionTime. That will set up another nice pattern here, where every access module (regardless of what supplementary data it tracks) is checked against the same Timestamp-type field for completion.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: FYI, linked username **CAN** change. It is not a thing that is within our control, and we should not be trying to enforce restrictions there.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
rushtong: Shibboleth doesn't return errors, per se. Remember the flow is:
* AoU -> Shibboleth -> NIH
* User authenticates, or fails to authenticate
  - If the user fails to auth, NIH doesn't redirect back to Shibboleth
* NIH -> Shibboleth -> AoU with a (theoretically) valid token
* AoU -> Post token to FC -> AoU with NIH Status information.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -701,11 +714,29 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
     try {
-      fireCloudService.postNihCallback(wrapper);
+      NihStatus nihStatus = fireCloudService.postNihCallback(wrapper);
+      if (nihStatus.getLinkedNihUsername() == null) {
+        log.log(Level.WARNING, "No Username Found when updating nih token");
+      }
       User user = initializeUserIfNeeded();
+      if (nihStatus != null) {
+        Timestamp eraCommonsCompletionTime = user.getEraCommonsCompletionTime();
+        if ((nihStatus.getLinkedNihUsername() != null &&
```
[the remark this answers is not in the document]
blrubenstein: There is no compelling reason to have it separate. Will update.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
blrubenstein: Will add that.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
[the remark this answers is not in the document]
blrubenstein: I have a third proposal, which is adding it into the user service and simplifying the controller logic. Will push up shortly with other changes.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
blrubenstein: Agreed. Doing that.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -493,15 +488,23 @@ private User initializeUserIfNeeded() {
...
-    return ResponseEntity.ok(profile);
+    return getProfileResponse(user);
+  }
+
+  @Override
+  public ResponseEntity<Profile> syncEraCommonsStatus() {
+    User user = userProvider.get();
+    NihStatus nihStatus = fireCloudService.getNihStatus();
+    userService.setEraCommonsStatus(nihStatus);
+    userDao.save(user);
```
gjuggler: This line should be unnecessary now that userService is saving w/ retries.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -493,15 +488,23 @@ private User initializeUserIfNeeded() {
...
     }
-    return ResponseEntity.ok(profile);
+    return getProfileResponse(user);
+  }
+
+  @Override
+  public ResponseEntity<Profile> syncEraCommonsStatus() {
+    User user = userProvider.get();
+    NihStatus nihStatus = fireCloudService.getNihStatus();
+    userService.setEraCommonsStatus(nihStatus);
```
gjuggler: nit: reassign the user w/ the return value from userService.setEraCommonsStatus?

user = userService.setEraCommonsStatus(nihStatus);

PR #1874, opened 2019-03-04
[P3][RW-2258][risk=no] fix styling on cohort create card

jacmrob:

### DIFF
--- a/ui/src/app/components/card.tsx
+++ b/ui/src/app/components/card.tsx
@@ -1,6 +1,6 @@
 import {reactStyles, withStyle} from 'app/utils';
 
-export const styles = reactStyles({
+const baseStyles = reactStyles({
   card: {
     padding: '1rem',
     borderRadius: '0.2rem',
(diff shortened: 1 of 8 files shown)


jacmrob:
makes it the size of all the other resource "create" cards

blrubenstein:
(Found a regression, so I think it does need the changes we discussed offline)

screen shot 2019-03-06 at 10 42 26 am

jacmrob:
So I did a bit more refactor to get some consistency with card sizing.  There are now two types of "card" - `WorkspaceCard` (the large ones for workspaces) and `ResourceCard` (small ones for cohorts, concept sets, etc).  They both build off a common style base and then have unique styles for sizing.  For CardButtons we have "small" (add cohort, add notebook, etc) and "large" (add workspace), which also build off common styling.  

so now any addCard or resourceCard styling inline in components should solely be fore unique ways we position the cards on the page.  Actual styling of the cards themselves should not need to be set in every parent component.

ui/src/app/components/card.tsx
```diff
@@ -9,10 +9,22 @@ export const styles = reactStyles({
     border: '1px solid #d7d7d7',
     display: 'flex',
     flexDirection: 'column',
-    width: '12rem',
```
jacmrob: these were always overwritten it turns out and we never had any cards with these sizing dimensions -____-

ui/src/app/components/card.tsx
```diff
@@ -9,10 +9,22 @@ export const styles = reactStyles({
...
-    height: '9rem',
     margin: '0 1rem 1rem 0'
   }
 });
 
-export const Card = withStyle(styles.card)('div');
+export const styles = reactStyles({
+  workspaceCard: {
+    ...baseStyles.card,
+    minWidth: '300px', maxWidth: '300px',
```
jacmrob: adding both max and min for width/height should fix any issues with card size changing on page resize

ui/src/app/views/workspace-list/component.tsx
```diff
@@ -44,8 +44,7 @@ const styles = reactStyles({
     display: 'flex', justifyContent: 'flex-start', flexWrap: 'wrap'
   },
   addCard: {
-    display: 'flex', fontSize: '20px', lineHeight: '28px', marginTop: '0',
-    fontWeight: 600, color: '#216FB4'
+    margin: '0 1rem 1rem 0', fontWeight: 600, color: 'rgb(33, 111, 180)'
```
jacmrob: custom styling for the add card on the workspace list page

PR #1878, opened 2019-03-05
[RW-1302][RW-319][risk=no] Return a 400 on bad requests

calbach:
Fixes a long-standing bug where we return a 500 errors on malformed requests. Looking at a security ticket complaining about these 500s finally forced the issue.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/exceptions/ExceptionAdvice.java
+++ b/api/src/main/java/org/pmiops/workbench/exceptions/ExceptionAdvice.java
@@ -6,6 +6,7 @@ import java.util.logging.Logger;
 import org.pmiops.workbench.model.ErrorResponse;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/exceptions/ExceptionAdvice.java
```diff
@@ -6,13 +6,23 @@
...
 
 @ControllerAdvice
 public class ExceptionAdvice {
+  private static final Logger log = Logger.getLogger(ExceptionAdvice.class.getName());
+
+  @ExceptionHandler({HttpMessageNotReadableException.class})
+  public ResponseEntity<?> messageNotReadableError(Exception e) {
+    log.log(Level.INFO, "failed to parse HTTP request message, returning 400", e);
+    return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(
+        WorkbenchException.errorResponse("failed to parse valid JSON request message")
```
blrubenstein: I think you can use our BadRequestException here? https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/exceptions/BadRequestException.java

calbach: It's possible that would work, but the code that translates our exceptions into responses (including BadRequestException) is 2 lines below this method. These handlers are logically being applied at the same level so even if it worked today I'm not sure I want to depend on that working in future versions of Spring.

PR #1880, opened 2019-03-06
[RW-2095][Risk=no] Adding cohorts to Data Set page

UrsaStutsman:
Ready for review. Tests up.


ui/src/app/views/dataset/component.tsx
```diff
@@ -128,39 +137,57 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
         id, updatedResource as Cohort);
+      updatedCohortList = this.state.cohortList.map((cohort) => {
+        return updatedResource.id === cohort.id ? (updatedResource as Cohort) : cohort;
+      });
     } else if (resource.conceptSet) {
       call = conceptSetsApi().updateConceptSet(workspaceNamespace, workspaceFirecloudName,
         id, updatedResource);
+      updatedConceptSetList = this.state.conceptSetList.map((conceptSet) => {
+        return updatedResource.id === conceptSet.id ? (updatedResource as ConceptSet) : conceptSet;
+      });
```
panentheos: It seems like there are a couple places with very similar code paths that only differ by a few pieces of information. It might be worth trying to refactor those at some point.

ui/src/app/views/dataset/component.tsx
```diff
@@ -104,20 +110,23 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      call = cohortsApi().deleteCohort(namespace, id, resourceId);
     }
     if (this.state.rType === ResourceType.CONCEPT_SET) {
-      const deleted = this.state.resource.conceptSet;
-      const updatedList = fp.filter(conceptSet => conceptSet.id !== deleted.id,
+      const updatedList = fp.filter(conceptSet => conceptSet.id !== resourceId,
         this.state.conceptSetList);
       this.setState({conceptSetList: updatedList});
+    } else {
+      const updatedList = fp.filter(cohort => cohort.id !== resourceId,
```
panentheos: FYI: formulating this as a `remove` could allow you to take advantage of the shorthand:
```
fp.remove({id: resourceId}, this.state.cohortList)
```

ui/src/app/views/dataset/component.tsx
```diff
@@ -128,39 +137,57 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+        this.setState({selectedConceptSets: (origSelected).concat(selected)});
       }
     } else {
-      // else = cohort
-      //  determine if already in selected
-      //  remove if it is
-      //  add if it isn't
+      const origSelected = this.state.selectedCohorts;
+      const selected = fp.filter((c) => c.id === resource.id, origSelected);
+      if (selected.length > 0) {
```
panentheos: FYI: since we only care about the presence, not the actual items, we could use
```
fp.some({id: resource.id}, origSelected)
```

ui/src/app/views/dataset/component.tsx
```diff
@@ -128,39 +137,57 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
   }
 
   closeEditModal(): void {
     this.setState({editing: false});
   }
 
+  clone(cohort: Cohort): void {
+    const {namespace, id} = this.props.workspace;
+    navigateByUrl(`/workspaces/${namespace}/${id}/cohorts/build?cohortId=${cohort.id}`);
+  }
```
panentheos: TOL: This seems like a strange thing to have to put here. It might be worth trying to disentangle this polymorphic 'resource' notion a bit at some point.

blrubenstein:
Since Ursa is out on vacation this week, I am going to take this over, fyi.

ui/src/app/views/dataset/component.tsx
```diff
@@ -196,43 +219,73 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
                 <ClrIcon shape='plus-circle' class='is-solid' style={styles.addIcon}
                   onClick={() => navigate(['workspaces', namespace, id,  'cohorts', 'build'])}/>
               </div>
-              {/*TODO: load cohorts and display here*/}
-              <div style={{height: '8rem'}}/>
+              <div style={{height: '10rem', overflowY: 'auto'}}>
+                {this.state.cohortList.length > 0 && !loadingResources &&
+                  this.state.cohortList.map(cohort =>
```
panentheos: nit: Don't need to check length explicitly, if the array is empty then nothing will get rendered.

PR #1883, opened 2019-03-06
[RW-2259][risk=no][P4] "temporal" toggle on cohort builder runs into text

smazumd:
[P4] "temporal" toggle on cohort builder runs into text


ui/src/app/cohort-search/search-group/search-group.component.css
```diff
@@ -206,3 +206,10 @@ ul {
   padding-top: 0rem;
 }
 
+.card-footer .temporal-container {
+  display: flex;
+}
+
+.temporal-container .temporal-text {
+  padding-left: 2rem;
```
jacmrob: could just move this style in-line.  will make it easier when we convert to react.

PR #1885, opened 2019-03-07
[RW-2265][risk=no] Add more criteria docs

calbach:
I took a stab at this, could use some help filling in the blanks and on my TODOs.

### DIFF
--- a/api/src/main/resources/cb_search_api.yaml
+++ b/api/src/main/resources/cb_search_api.yaml
@@ -371,8 +371,10 @@ definitions:
           id will be used in the case that a parent is selectable.
         type: string
       hasAttributes:
-        # TODO: More detail - what are attributes and what does "are needed" mean?
-        description: specifies if extra attributes are needed
+        description: >
+          Whether this criteria has associated attributes which can be filtered
+          by value during a search, for example a blood pressure measurement
+          critieria might have an associated measurement value attribute.
         type: boolean
       path:
         description: >
--- a/api/src/main/resources/cb_search_api.yaml
+++ b/api/src/main/resources/cb_search_api.yaml
@@ -289,7 +289,6 @@ definitions:
       - id
       - parentId
       - type
-      - code
       - name
       - group
       - selectable
@@ -309,15 +308,13 @@ definitions:
         format: int64
       type:
         description: The tree type of this criteria, see TreeType.
-        # TODO: Use TreeType enum?
+        # TODO(RW-2278): Use TreeType enum.
         type: string
       subtype:
         description: The tree subtype of this criteria, see TreeSubType.
-        # TODO: Use TreeSubType enum?
+        # TODO(RW-2278): Use TreeSubType enum.
         type: string
       code:
-        # TODO: Some of the criteria table has null codes, but it's required
-        # here. What does that mean and how does that manifest in the API?
         description: >
           Code that identifies this criteria. In some vocabularies such as ICD9
           and ICD10, this code captures the tree hierarchy, e.g. '001.002.003'.
@@ -364,11 +361,10 @@ definitions:
         format: int64
       domainId:
         description: >
-          # TODO: More detail - what does this represent exactly? What is the
-          # relationship to the tree type?
-          clue to determine which OMOP tables to search, but these only exist
-          for leaves in the tree. Parents don't have domain ids and concept
-          id will be used in the case that a parent is selectable.
+          # TODO: Remove this field.
+          Deprecated. Clue to determine which OMOP tables to search, but these
+          only exist for leaves in the tree. Parents don't have domain ids and
+          concept id will be used in the case that a parent is selectable.
         type: string
       hasAttributes:
         description: >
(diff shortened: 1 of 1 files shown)


api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
         description: >
-          clue to determine which tables to search, but these only exist
+          # TODO: More detail - what does this represent exactly? What is the
+          # relationship to the tree type?
+          clue to determine which OMOP tables to search, but these only exist
           for leaves in the tree. Parents don't have domain ids and concept
           id will be used in the case that a parent is selectable.
         type: string
       hasAttributes:
+        # TODO: More detail - what are attributes and what does "are needed" mean?
```
jscherdin: If this flag is set, it means that there may be additional details a user might want to add (ie, selecting a lab and entering a value)

api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
-          code that identifies this criteria, ex: icd9, icd10 or cpt all
-          have codes that map to the concept_code column in the concept
-          table.
+          Code that identifies this criteria. In some vocabularies such as ICD9
+          and ICD10, this code captures the tree hierarchy, e.g. '001.002.003'.
+          Multiple criteria may exist for the same code within a CDR version if
+          a given concept has multiple entries at different locations in the
+          criteria tree (this is common in SNOMED).
+
+          Criteria codes should generally be stable across CDR versions.
```
calbach: @jscherdin do you mind double-checking my statements here about IDs and how we expect them to be affected across CDR version changes?

In summary what I've documented here:
id - not stable across CDR verisons
code - "mostly"? stable across CDR versions
conceptIds - also "mostly" stable?

api/src/main/resources/cb_search_api.yaml
calbach: Thanks - attempted to reword the docs based on this.

api/src/main/resources/cb_search_api.yaml
jscherdin: ID is not stable. Code and concept ID - yes mostly stable is a good way to put it

api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
-        description: parent id of the criteria
+        description: >
+          The parent id of the criteria. 0 if this is the root node of a
+          criteria tree.
         type: integer
         format: int64
       type:
-        description: type of criteria
+        description: The tree type of this criteria, see TreeType.
+        # TODO: Use TreeType enum?
```
freemabd1: Yes, we should be using these enums. We can address this in a separate PR if you like. I think it will entail UI and test case fixes.

api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
         type: string
       subtype:
-        description: subtype of this criteria
+        description: The tree subtype of this criteria, see TreeSubType.
+        # TODO: Use TreeSubType enum?
```
freemabd1: Same as above.

api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
         type: string
       code:
+        # TODO: Some of the criteria table has null codes, but it's required
+        # here. What does that mean and how does that manifest in the API?
```
freemabd1: Code can be removed from required. The api doesn't depend on the code being set unless it's ICD9/ICD10 type. We have validation here: https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/CodesQueryBuilder.java#L121

api/src/main/resources/cb_search_api.yaml
```diff
@@ -296,56 +296,88 @@ definitions:
...
+
+          Concept IDs should generally be stable across CDR versions.
         type: integer
         format: int64
       domainId:
         description: >
-          clue to determine which tables to search, but these only exist
+          # TODO: More detail - what does this represent exactly? What is the
+          # relationship to the tree type?
+          clue to determine which OMOP tables to search, but these only exist
```
freemabd1: domainId is deprecated. I think this property is still around because we wanted to make sure cloning still worked on old cohort definitions. I will research and see if we can remove this completely.

api/src/main/resources/cb_search_api.yaml
freemabd1:
```suggestion
        This flag indicates that a criteria has auxiliary information in the criteria_attribute table, which helps the UI provide information about data ranges on user inputs.
```

api/src/main/resources/cb_search_api.yaml
calbach: Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-2278 and added TODOs

api/src/main/resources/cb_search_api.yaml
calbach: Done

api/src/main/resources/cb_search_api.yaml
calbach: Added a comment here and noted it is deprecated

PR #1887, opened 2019-03-07
[RW-2239][Risk=low] Adding per user bypass for Access Modules

UrsaStutsman:


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -100,11 +100,30 @@ private User updateWithRetries(Function<User, User> userModifier, User user) {
...
-    boolean shouldBeRegistered = Optional.ofNullable(user.getIdVerificationIsValid()).orElse(false)
-        && user.getDemographicSurveyCompletionTime() != null
-        && user.getTrainingCompletionTime() != null
-        && user.getTermsOfServiceCompletionTime() != null
+    // TODO: Add in when we add this module
+    // boolean dataUseAgreementCompliant = user.getDataUseAgreementCompletionTime() != null ||
+    // user.getDataUseAgreementBypassTime() != null || !configProvider.get().access.enableDataUseAgreement;
+    Timestamp current = new Timestamp(clock.instant().toEpochMilli());
+    boolean complianceTrainingNotExpired = user.getComplianceTrainingExpirationTime() == null ||
+      current.before(user.getComplianceTrainingExpirationTime()) ||
```
gjuggler: Further to my comment above about punting the compliance_training_expiration_time field to a separate PR, I'd simply remove this logic.

Even when we do implement the field rename, I don't think we want to involve the compliance-training expiration time in the logic that determines whether a user should be in the registered group or not. Frankly, we don't quite understand what the "expiration time" we get from Moodle will actually represent. At best, we may use it to drive some user-facing notifications, but even that's uncertain. :)

Plus – for consistency & predictability, all modules should simply check the same core set of fields: completion time, bypass time, and environment-enabled.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -100,11 +100,30 @@ private User updateWithRetries(Function<User, User> userModifier, User user) {
...
+      !(configProvider.get().access.enableComplianceTraining);
+    boolean complianceTrainingCompliant = complianceTrainingNotExpired &&
+      (user.getComplianceTrainingCompletionTime() != null ||
+      user.getComplianceTrainingBypassTime() != null ||
+      !(configProvider.get().access.enableComplianceTraining));
+    // TODO: add in the check on completion time once RW-2235 (Scout's) has been merged in
+    boolean eraCommonsCompliant = user.getEraCommonsBypassTime() != null || !configProvider.get().access.enableEraCommons;
+    boolean idVerificationCompliant = user.getIdVerificationCompletionTime() != null ||
+      user.getIdVerificationBypassTime() != null || !configProvider.get().access.enableIdVerification ||
+      Optional.ofNullable(user.getIdVerificationIsValid()).orElse(false);
```
gjuggler: It's worth making a note here that we need to eventually migrate from the "id_verification_is_valid" field to the "id_verification_completion_time" field... maybe a TODO above this line of logic?

api/db/changelog/db.changelog-58-update-access-modules.xml
```diff
@@ -0,0 +1,25 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="ustutsman" id="db.changelog-58-update-access-modules">
+    <addColumn tableName="user">
+      <column name="data_use_agreement_completion_time" type="datetime"/>
+      <column name="data_use_agreement_bypass_time" type="datetime"/>
+      <column name="compliance_training_expiration_time" type="datetime"/>
```
gjuggler: I know the ticket wasn't very precise about how to handle column renames like this... sorry about that!

After looking through the PR I think for this particular field (compliance_training_expiration_time), we should exclude it from this PR and handle the column rename (training_expiration_time --> compliance_training_expiration_time) as a separate follow-up. It'll help keep this PR simpler, and won't affect the introduction of the core access module fields.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -666,6 +667,45 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
+      case "eraCommons":
+        user.setEraCommonsBypassTime(valueToSet);
+        break;
+      case "idVerification":
+        user.setIdVerificationBypassTime(valueToSet);
+        break;
+      case "twoFactorAuth":
+        user.setTwoFactorAuthBypassTime(valueToSet);
+        break;
+      default:
```
gjuggler: Nitpick: return an error response if we get an unrecognized module name.

PR #1890, opened 2019-03-08
[risk=no] Upgrade Mockito and add the mock-maker-inline extension

gjuggler:
While working on some unit tests for RW-2236 I found myself stuck in a corner where I couldn't make progress without being able to mock a method on GoogleCredential. The problem was that all of GoogleCredential's core methods were marked as final.

Based on my reading of https://www.baeldung.com/mockito-final and https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2, Mockito 2 added some tricks to support this in a reasonably easy-to-use way.

I'm not by any means an expert here, so I don't know if this would have unintended consequences. But it worked like a charm for my use case, and doesn't seem to break any other tests.

I'm not attached to this at all. If there are strong objections, I'm happy to revert! The only impact is, I won't be able to test the impersonated-credential code as well... but testing ability there is already somewhat limited, so it wouldn't be a huge loss.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -341,7 +341,7 @@ dependencies {
   toolsCompile 'commons-cli:commons-cli:1.4'
 
   testCompile 'junit:junit:4.12'
-  testCompile 'org.mockito:mockito-core:1.10.19'
+  testCompile 'org.mockito:mockito-core:2.18.3'
   testCompile "com.google.appengine:appengine-api-stubs:${GAE_VERSION}"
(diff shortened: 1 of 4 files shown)


gjuggler:
A bit more reading on the mockito-inline functionality suggests there is a risk of test performance degradation. I'm not sure if it applies to *all* mocked functions, or only pays a performance penalty for when the inline-mocking is used.

I didn't notice an obvious difference in the CircelCI unit test runtime, but I added a README to the extension directory with some pointers & a note that we might disable this / revert if we notice a slowdown.

PR #1893, opened 2019-03-08
POC code for new router

panentheos:
This provides a proof-of-concept implementation for the new router. While it should be mostly valid, as of the time of writing it has not been wired up or tested. The intention is that once all the relevant components have been converted, this PR can serve as the foundation for the final router replacement.

This code depends on three libraries: `path-to-regexp`, `qs`, and `history`, which will need to be installed.

See comments for detailed notes.

### DIFF
--- a/ui/src/app/components/routing.tsx
+++ b/ui/src/app/components/routing.tsx
@@ -1,0 +1,35 @@
+import * as React from 'react';
+
+import {withRoute} from 'app/routes';
+import {history} from 'app/services/history';
+
+export const Router = withRoute()(({route}) => {
+  if (!route) {
+    return <div>Not found</div>;
+  }
+  return route.render();
+});
+
+export class Redirect extends React.Component<{to: any}> {
+  componentDidMount() {
+    const {to} = this.props;
+    history.replace(to);
+  }
+
+  render() {
+    return null;
+  }
+}
+
+// Example of how to enforce login on particular pages
+/*
+export const AuthWrapper = fp.flow(
+  withRoute(),
+  withAuthState(),
+)(({route, authState, children}) => {
+  if (!route.public && !authState.loggedIn) {
+    return <Redirect to={getPath('login')} />;
+  }
+  return children;
+});
+*/
--- a/ui/src/app/routes.tsx
+++ b/ui/src/app/routes.tsx
@@ -1,0 +1,265 @@
+import pathToRegexp from 'path-to-regexp';
+import * as qs from 'qs';
+
+import {history} from 'app/services/history';
+import {BreadcrumbType} from 'app/utils/navigation';
+import {AdminReviewWorkspace} from 'app/views/admin-review-workspace/component';
+import {AdminUser} from 'app/views/admin-user/component';
+import {CohortList} from 'app/views/cohort-list/component';
+import {ConceptHomepage} from 'app/views/concept-homepage/component';
+import {ConceptSetDetails} from 'app/views/concept-set-details/component';
+import {ConceptSetList} from 'app/views/concept-set-list/component';
+import {DataPage} from 'app/views/data-page/component';
+import {DataSet} from 'app/views/dataset/component';
+import {Homepage} from 'app/views/homepage/component';
+import {NotebookList} from 'app/views/notebook-list/component';
+import {NotebookRedirect} from 'app/views/notebook-redirect/component';
+import {ProfilePage} from 'app/views/profile-page/component';
+import {Settings} from 'app/views/settings/component';
+import {SignIn} from 'app/views/sign-in/component';
+import {StigmatizationPage} from 'app/views/stigmatization-page/component';
+import {Unregistered} from 'app/views/unregistered/component';
+import {WorkspaceEdit, WorkspaceEditMode} from 'app/views/workspace-edit/component';
+import {WorkspaceList} from 'app/views/workspace-list/component';
+import {WorkspaceWrapper} from 'app/views/workspace-wrapper/component';
+import {Workspace} from 'app/views/workspace/component';
+
+const routes = [
+  {
+    name: 'login',
+    path: '/login',
+    render: () => <SignIn />,
+    title: 'Sign In',
+    public: true
+  },
+  {
+    name: 'homepage',
+    path: '/',
+    render: () => <Homepage />,
+    title: 'Homepage',
+  },
+  {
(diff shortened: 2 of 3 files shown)


ui/src/app/components/routing.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import {withRoute} from 'app/routes';
+import {history} from 'app/services/history';
+
+export const Router = withRoute()(({route}) => {
+  if (!route) {
+    return <div>Not found</div>;
```
panentheos: This is where the 'not found' page is rendered. Alternatively, we could add a catch-all route at the end of the routes list, and handle it that way.

ui/src/app/components/routing.tsx
```diff
@@ -0,0 +1,35 @@
...
+});
+
+export class Redirect extends React.Component<{to: any}> {
+  componentDidMount() {
+    const {to} = this.props;
+    history.replace(to);
+  }
+
+  render() {
+    return null;
```
panentheos: Renders nothing, but navigates as a side effect. This allows us to perform a redirect by rendering this component, which is convenient (see below).

ui/src/app/components/routing.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+// Example of how to enforce login on particular pages
```
panentheos: This is partly pseudo-code since the 'authState' concept hasn't been written yet, but it shows the basic concept of how to write a 'guard' by conditionally rendering children.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+import {WorkspaceWrapper} from 'app/views/workspace-wrapper/component';
+import {Workspace} from 'app/views/workspace/component';
+
+const routes = [
+  {
+    name: 'login',
+    path: '/login',
+    render: () => <SignIn />,
+    title: 'Sign In',
+    public: true
```
panentheos: Routes can contain arbitrary data with a variety of uses.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+  {
+    name: 'workspace',
+    path: '/workspaces/:ns/:wsid',
+    render: () => <WorkspaceWrapper><Workspace /></WorkspaceWrapper>,
+    title: 'View Workspace Details',
+    breadcrumb: BreadcrumbType.Workspace
+  },
+  {
+    name: 'workspaceEdit',
+    path: '/workspaces/:ns/:wsid/edit',
```
panentheos: These paths are fed to the `path-to-regexp` library. See its documentation for available options.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+  {
+    name: 'notebooks',
+    path: '/workspaces/:ns/:wsid/notebooks',
+    render: () => <WorkspaceWrapper><NotebookList /></WorkspaceWrapper>,
+    title: 'View Notebooks',
+    breadcrumb: BreadcrumbType.Workspace
+  },
+  {
+    name: 'notebook',
+    path: '/workspaces/:ns/:wsid/notebooks/:nbName',
```
panentheos: If the duplication of e.g. `/workspaces/:ns/:wsid` is a concern, it would be reasonable to extract that out into a variable, e.g. `workspacePrefix`.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+      <WorkspaceEdit mode={WorkspaceEditMode.Edit}/>
+    </WorkspaceWrapper>,
+    title: 'Edit Workspace',
+    breadcrumb: BreadcrumbType.WorkspaceEdit
+  },
+  {
+    name: 'workspaceClone',
+    path: '/workspaces/:ns/:wsid/clone',
+    render: () => <WorkspaceWrapper>
+      <WorkspaceEdit mode={WorkspaceEditMode.Clone}/>
```
panentheos: Some data that was previously route data might be simpler as props, e.g. `mode` here.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+    query: qs.parse(search, {ignoreQueryPrefix: true, plainObjects: true}),
+  };
+};
+
+/**
+ * HOC that injects a `route` prop, whose value is the matched route handler, extended
+ * with `params` (path params) and `query` (query params). If no route matches, the value
+ * will be undefined.
+ */
+export const withRoute = () => WrappedComponent => {
```
panentheos: This is the main tool for injecting and using information about the current route.

An important and useful detail here is that the route _matching_ is separate from the route _rendering_. Any component can use this HOC to make decisions based on the route, even ones higher in the tree than the `Router` component itself.

For example, we will likely have a top-level 'signed in' wrapper that will examine the route to determine whether to redirect to the login page. There could also be a 'title renderer' component whose sole task is to watch the current route and update the page title. The `Router` component is just one of many consumers of this data, which can all be freely composed with one another.

ui/src/app/routes.tsx
```diff
@@ -0,0 +1,265 @@
...
+    render() {
+      const {location} = this.state;
+      return <WrappedComponent {...this.props} route={parseRoute(location)} />;
+    }
+  }
+  return Wrapper;
+};
+
+/**
+ * Return a URL path string, given a route name and its params.
```
panentheos: This is the main function for generating urls. E.g. `getPath('notebooks', {ws, nsid})`

Navigation is done via `history.push`. We'll probably want some sort of `Link` component that abstracts this behavior nicely for links.

ui/src/app/components/routing.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import {withRoute} from 'app/routes';
+import {history} from 'app/services/history';
+
```
calbach: Possibly another component we'd want eventually here would be something like a `<RouterLink to="/" />` which renders an `<a href="/">...</a>` with a click handler that handles this via the html lib.

Not something I'm necessarily asking you to add to the PR but noting we may need it and raising in case you had any thoughts in this area.

panentheos: Yes, that's definitely an abstraction we'll want.

PR #1897, opened 2019-03-08
[RW-2289][Risk=low] Move to binary of Rainforest CLI

blrubenstein:

### DIFF
--- a/api/src/dev/server/Dockerfile
+++ b/api/src/dev/server/Dockerfile
@@ -22,7 +22,6 @@ RUN apk --no-cache add \
         libc6-compat \
         openssh-client \
         git \
-        make \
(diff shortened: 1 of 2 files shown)


api/src/dev/server/Dockerfile
```diff
@@ -22,7 +22,6 @@ RUN apk --no-cache add \
         libc6-compat \
         openssh-client \
         git \
-        make \
```
gjuggler: Can you explain this and the other removed / changed lines? They seem unrelated to the purpose of this PR. If they aren't – I'm wondering what purpose they served, or what about the Rainforest change has made them unnecessary?

Or were they just cruft to begin with?

blrubenstein: Sure! Those were added to support the ruby version of the rainforest cli, which is deprecated. Wanted to remove them as they are now unneeded dependencies.

gjuggler: Awesome – thanks for clarifying.

PR #1902, opened 2019-03-12
[RW-2236] [risk=low] Add support for generating impersonated OAuth credentials.

gjuggler:
This PR adds support, docs and integration tests for generating impersonated OAuth credentials for communicating with FireCloud APIs.

As demonstrated in the integration test & command-line tool, this will allow us to pull FireCloud data on behalf of users, outside the context of the web application.

*Deployment note*: Part of this system involves loading and using credentials for a new service account "firecloud-admin". Before impersonation can be used by the webapp in production (e.g. as part of a cron job), we need to create these new service accounts and get them whitelisted by FireCloud. Docs for this have been added to http://broad.io/aou-new-environment.

However, this PR on its own doesn't require these changes, since the cron job isn't implemented yet. That will be handled as a follow-up PR, after which we need to make sure new SAs have been created for all environments.


api/doc/domain-wide-delegation.md
```diff
@@ -0,0 +1,48 @@
+# Domain-wide delegation in AoU RW
```
gjuggler: I'm trying a new thing here, adding some relatively-lightweight .md documentation into the codebase. Compared to a separate tech doc, this makes it more easily linkable and searchable within the repo.

I'm hoping this might be a good pattern for documenting "the way things work" for certain parts of the system that may be non-obvious. This was a pretty successful pattern at Google (including team-focused documentation inside the codebase); we'll see if it catches on here :)

api/libproject/devstart.rb
```diff
@@ -1060,6 +1060,38 @@ def backfill_gsuite_user_data(cmd_name, *args)
   :fn => ->(*args) {backfill_gsuite_user_data("backfill-gsuite-user-data", *args)}
 })
 
+def fetch_firecloud_user_profile(cmd_name, *args)
```
gjuggler: You might argue that the "Fetch user profile" command-line tool is useless, which it is – but it did force me to ensure the creds impersonation actually works from the command-line. And it provides a simple pattern for anyone who needs to write a tool which actually does something useful w/ the FireCloud APIs.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -29,46 +47,131 @@
...
+    return createService(loadConfig("config_prod.json"));
+  }
+
+  @Test
+  public void testStatusProd() throws Exception {
+    assertThat(getProdService().getFirecloudStatus()).isTrue();
+  }
+
+  @Test
+  public void testStatusDev() throws Exception {
```
gjuggler: AFAICT, we only used to integration test against prod. This adds calls to both dev & prod. Unless there's a particular reason we don't want to test pinging both endpoints? If we find FireCloud dev is flaky at all, we can add a retry and/or remove the test.

api/src/integration/java/org/pmiops/workbench/IntegrationTestConfig.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.retry.backoff.ExponentialRandomBackOffPolicy;
+import org.springframework.retry.backoff.Sleeper;
+import org.springframework.retry.backoff.ThreadWaitSleeper;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+
+import static org.springframework.context.annotation.FilterType.ASSIGNABLE_TYPE;
+
+@Configuration
```
gjuggler: This config ends up repeating a lot of what's in the CommandLineTool config. It may behoove us to merge them at some point. I'll leave that for follow-up, probably combined w/ the work of pulling out our separate tools-specific Gradle config.

api/src/main/java/org/pmiops/workbench/config/AppEngineConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import com.google.api.client.extensions.appengine.http.UrlFetchTransport;
+import com.google.api.client.http.HttpTransport;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class AppEngineConfig {
+
+    @Bean
+    HttpTransport httpTransport() {
```
gjuggler: The HttpTransport bean is one of the key App Engine specific dependencies we inject. I'd originally tried putting this bean within WebMvcConfig, but that was causing a cyclic bean dependency (since Spring tries to load all beans for a given class at the same time). Pulling it out into its own file here resolved the issue.

api/src/main/java/org/pmiops/workbench/config/CommonConfig.java
```diff
@@ -18,14 +19,13 @@ JsonFactory jsonFactory() {
     return new JacksonFactory();
   }
 
-  @Bean
```
gjuggler: Removing HttpTransport from CommonConfig allows us to make CommonConfig actually be shared across our three main configs: WebMvc, CommandLineTool, and IntegrationTest.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
```diff
@@ -77,9 +88,11 @@ public WorkbenchEnvironment workbenchEnvironment() {
    *
    * We may in future rotate key files in production, but will be sure to keep the ones currently
    * in use in cloud environments working when that happens.
+   *
+   * TODO(gjuggler): should we start pulling this file from GCS instead?
    */
   @Lazy
-  @Bean
+  @Bean(name= Constants.GSUITE_ADMIN_CREDS)
```
gjuggler: I don't love that we need to include these bean names, but the alternative of creating dumb wrapper classes (GSuiteAdminCreds, which has a single method getGoogleCredential) seemed like more effort than it would be worth.

api/tools/build.gradle
```diff
@@ -86,8 +86,15 @@ sourceSets {
                   // In general, command-line tools won't need access to
                   // org.pmiops.workbench.model (Swagger-generated classes),
                   // so we selectively whitelist as needed.
+                  'org/pmiops/workbench/auth/Constants.java',
```
gjuggler: Rant mode: dealing with this per-file dependency management was a nightmare. I can't wait until we get rid of this, it will make it 5x easier to write small command-line tools.

api/libproject/devstart.rb
```diff
@@ -1060,6 +1060,38 @@ def backfill_gsuite_user_data(cmd_name, *args)
...
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--username=[username]",
+      String,
+      ->(opts, v) { opts.username = v},
+      "The AoU username to fetch FireCloud data for (e.g. 'gjordan'")
```
blrubenstein: What happens if you pass in `gjordan@fake-research-aou`?

api/libproject/devstart.rb
blrubenstein: If the goal is to provide that pattern for people who want to write a tool which does something useful with the FireCloud APIs, can we remove the `register_command` block? Just to avoid unused commands showing up in the blank arg `./project.rb`?

api/libproject/devstart.rb
```diff
@@ -1798,12 +1830,19 @@ def setup_project_data(gcc, cdr_db_name, public_db_name)
...
   with_cloud_proxy_and_db(gcc) do
-    common.status "Copying service account key to GCS..."
+    common.status "Copying GSuite service account key to GCS..."
     gsuite_admin_creds_file = Tempfile.new("gsuite-admin-sa.json").path
     common.run_inline %W{gcloud iam service-accounts keys create #{gsuite_admin_creds_file}
         --iam-account=gsuite-admin@#{gcc.project}.iam.gserviceaccount.com --project=#{gcc.project}}
     common.run_inline %W{gsutil cp #{gsuite_admin_creds_file} gs://#{gcc.project}-credentials/gsuite-admin-sa.json}
 
+    common.status "Copying FireCloud service account key to GCS..."
+    firecloud_admin_creds_file = Tempfile.new("firecloud-admin-sa.json").path
```
blrubenstein: Is this secret? Can you make sure it is blocked from being committed by git secrets if so?

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
blrubenstein: Let's keep an eye on this one for sure, to make sure that we aren't blocking releases to staging if FC dev is down.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -81,24 +88,24 @@ public WorkspacesApi workspacesApi(@Qualifier(END_USER_API_CLIENT) ApiClient api
...
-  public GroupsApi groupsApi(@Qualifier(ALL_OF_US_API_CLIENT) ApiClient apiClient) {
+  public GroupsApi groupsApi(@Qualifier(SERVICE_ACCOUNT_API_CLIENT) ApiClient apiClient) {
     // Group/Auth Domain creation and addition are made by the AllOfUs service account
     GroupsApi api = new GroupsApi();
     api.setApiClient(apiClient);
     return api;
   }
 
-  @Bean(name = END_USER_GROUPS_API)
+  @Bean(name=END_USER_GROUPS_API)
```
blrubenstein: Nit: Can we do consistent spacing here? We have no spaces here, and above have a single space after the `=` sign. I don't have any strong feelings what we should do, just as long as we are consistent.

api/tools/build.gradle
blrubenstein: Can we bump that Tech Debt ticket up? It sounds like it could be really valuable to deal with this before we start getting into more command line overrides and breaking up our command line tools.

api/libproject/devstart.rb
gjuggler: Is it possible to run the command at all if I leave it out? Part of the imagined utility here would be that we can let someone see that it actually works.

Maybe adding a prefix to the command like "example_" or "tool_example" could help make it obvious in the command list?

Either way... project.rb is a huge mess and I'm hoping https://precisionmedicineinitiative.atlassian.net/browse/RW-605 will give us a better model for maintaining many commands in an organized state.

api/libproject/devstart.rb
gjuggler: It probably errors out. I didn't think edge cases were too important for a demo command-line tool.

api/libproject/devstart.rb
gjuggler: Yes – it's a secret in the same way all of our service account keys are secrets. IAFAICT the git-secrets config that we apply as part of the prepare-commit-msg hook (https://github.com/all-of-us/utils/blob/master/hooks/apply-git-secrets.sh#L17) should cover all Google-style JSON service account keys.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
gjuggler: Good point. Maybe a more ideal solution would be for integration tests to support being parameterized by the deployment environment they're blocking, e.g. a staging deploy would run integration tests against FC prod.

But before that, we should develop a clearer strategy and definition on what our integration tests should cover. For instance, right now there's a warning at the top of these tests that says "Integration tests with real projects/clusters against production {workbench|notebooks} is not recommended at this time". I'm not quite sure that's enough to guide expansion of our integration-level coverage.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
gjuggler: Good catch, done. I'm 100% for consistency, I just haven't set up IntelliJ perfectly yet. I'm hoping we can get some pre-commit hooks that relieve us all from the mental burden of remembering code-formatting rules like this.

api/tools/build.gradle
gjuggler: Agreed. This is https://precisionmedicineinitiative.atlassian.net/browse/RW-2172

api/libproject/devstart.rb
blrubenstein: Fair. I just know we often use the full email address as username in the DB, so wanted to make sure we handled that? (If I query the db, copy the username field, and run the tool, it will fail. Similarly if we want to run this impersonation powered by the user object in the DB, I am a little worried about failure there.)

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
blrubenstein: If you do figure that out, it would be great to add that documentation to the handbook.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -29,46 +47,131 @@
...
+   * WorkbenchConfig. Note that this will always use the test environment's default service account
+   * credentials when making API calls. It shouldn't be possible to make authenticated calls to the
+   * FireCloud prod environment.
+   * <p>
+   * This method mostly exists to allow us to run a status-check against both FC dev & prod within
+   * the same integration test run.
+   **/
+  private FireCloudService createService(WorkbenchConfig config) {
+    ApiClient apiClient = new ApiClient();
+    apiClient.setBasePath(config.firecloud.baseUrl);
```
calbach: opt_nit: these setters are chainable

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -29,46 +47,131 @@
...
+        "integration-test-user@fake-research-aou.org");
+
+    // Run the most basic API call against the /me/ endpoint.
+    ProfileApi profileApi = new ProfileApi(apiClient);
+    Me me = profileApi.me();
+    assertThat(me.getUserInfo().getUserEmail()).isEqualTo("integration-test-user@fake-research-aou.org");
+    assertThat(me.getUserInfo().getUserSubjectId()).isEqualTo("101727030557929965916");
+
+    // Run a test against a different FireCloud endpoint. This is important, because the /me/
+    // endpoint is accessible even by service accounts whose subject IDs haven't been whitelisted
```
calbach: That's... slightly surprising.. I'm sure there's some history here. I don't suppose you know it off-hand? (probably not worth adding to the comments - just curious)

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -29,46 +47,131 @@
...
+    // by FireCloud devops.
+    //
+    // If we haven't had our "firecloud-admin" service account whitelisted,
+    // then the following API call would result in a 401 error instead of a 404.
+    NihApi nihApi = new NihApi(apiClient);
+    int responseCode = 0;
+    try {
+      NihStatus nihStatus = nihApi.nihStatus();
```
calbach: nit: unused value

api/src/integration/java/org/pmiops/workbench/IntegrationTestConfig.java
```diff
@@ -0,0 +1,130 @@
...
+public class IntegrationTestConfig {
+
+  /**
+   * Loads the GSuite admin service account key from GCS.
+   *
+   * This needs to be annotated with @Lazy so only classes that use it (e.g. BackfillGSuiteUserData
+   * which requires a WorkbenchConfig instance) will trigger the file load attempt.
+   *
+   * Any command-line tool which loads this bean needs to be called from a project.rb command
+   * which is preceded with "get_gsuite_admin_key" to ensure the local key file si populated.
```
calbach: nit: si -> is

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -27,4 +50,38 @@ public static String workbenchAccessToken(
...
+   * @param userEmail Email address of the user to impersonate.
+   * @param scopes The list of Google / OAuth API scopes to be authorized for.
+   * @return
+   * @throws IOException
+   */
+  public GoogleCredential getImpersonatedCredential(
+      GoogleCredential serviceAccountCredential,
+      String userEmail,
+      List<String> scopes) throws IOException {
+    // Build derived csredentials for impersonating the target user.
```
calbach: nit: spelling

api/src/main/java/org/pmiops/workbench/config/CommonConfig.java
```diff
@@ -18,14 +19,13 @@ JsonFactory jsonFactory() {
...
-  }
-
   @Bean
   Clock clock() { return Clock.systemUTC(); }
 
   @Bean
   Random random() { return new SecureRandom(); }
+  
+  @Bean
+  GoogleCredential.Builder googleCredentialBuilder() { return new GoogleCredential.Builder(); }
```
calbach: Beans are singleton by default (surprised? I was too), so I don't think this will work as intended. Adding `@Scope(value = ConfigurableBeanFactory.SCOPE_PROTOTYPE)` should fix.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -74,6 +95,38 @@ public FireCloudServiceImpl(Provider<WorkbenchConfig> configProvider,
     this.workspacesApiProvider = workspacesApiProvider;
     this.statusApiProvider = statusApiProvider;
     this.retryHandler = retryHandler;
+    this.serviceAccounts = serviceAccounts;
+    this.firecloudCredentialProvider = firecloudCredentialProvider;
```
calbach: nit: I'd probably include admin somewhere in this variable name; could potentially be misinterpreted as is

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -74,6 +95,38 @@ public FireCloudServiceImpl(Provider<WorkbenchConfig> configProvider,
...
+  public ApiClient getApiClientWithImpersonation(String userEmail) throws IOException {
+    // Load credentials for the firecloud-admin Service Account. This account has been granted
+    // domain-wide delegation for the OAuth scopes required by FireCloud.
+    GoogleCredential googleCredential = firecloudCredentialProvider.get();
+
+    GoogleCredential impersonatedUserCredential = serviceAccounts.getImpersonatedCredential(
+        googleCredential, userEmail, FIRECLOUD_API_OAUTH_SCOPES
+    );
+
+    ApiClient apiClient = new ApiClient();
```
calbach: opt_nit: can chain all of these, also bypassing the need for a local var.

api/tools/build.gradle
calbach: +1, though I'm a bit confused - why did you define the tool here rather than using your new dependency approach? (defining the tool at the api/build.gradle level?)  I imagined we might move towards eliminating this gradle file entirely at some point.

calbach:
(modulo Codacy issues - both look legitimate)

PR #1905, opened 2019-03-12
[RISK=LOW][RW-1916] React - workspace edit

NehaBroad:


blrubenstein:
Couple of things from running it:
1) The CDR dropdown should have a default (I believe to the latest CDR)? Currently starts at 'Select'
2) I think a lot of the formatting got thrown off. This is what I am seeing:
Screen Shot 2019-03-13 at 11 28 52 AM
Screen Shot 2019-03-13 at 11 29 49 AM

I believe that most of the research purpose information stuff should have a bounding box that is smaller than the full width of the page (I think lining up with the underserved population bounding box) and that there should be a bit more spacing between request a review and the create/cancel button.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+            {this.state.show &&
+            <ClrIcon shape='minus-circle' style={styles.icon}/>
+            }
+          </Clickable>
+        </div>
+      <div>
+        <div style={styles.header}>
+          Focus on an underserved population
+          <TooltipTrigger content='A primary mission of the All of Us Research Program is to include
+          populations that are medically underserved and/or historically underrepresented in
```
blrubenstein: Nit: This should be indented more.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+          populations that will be included in your research.'>
+            <InfoIcon style={styles.infoIcon}/>
+          </TooltipTrigger>
+        </div>
+        <label style={styles.focusLabel}>
+          This research will focus on, or include findings on, distinguishing characteristics
+          related to one or more underserved populations
+        </label>
+      </div>
+
```
blrubenstein: Nit: rm extra line

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+    </div>
+      {this.state.show && <div style={{paddingTop: '1rem'}}>
+        <h5>Additional information for underserved population</h5>
+        <label>Select all that apply. Only options for underserved populations are provided.
+        </label>
+        <div>
+          { FocusCategories.map (({title, categories} ) => {
```
blrubenstein: Nit: Extra space before `)`

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,321 @@
+import {Clickable} from 'app/components/buttons';
```
blrubenstein: This should probably go into `ui/src/app/components`, rather than `ui/src/app/views`.

ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -1,52 +1,28 @@
-import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
```
blrubenstein: Should this move to being a tsx file? I believe you mentioned still working on tests?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+import {ReactWrapperBase, withCurrentWorkspace, withRouteConfigData} from 'app/utils';
+import {navigate, userProfileStore} from 'app/utils/navigation';
+import {WorkspaceUnderservedPopulation} from 'app/views/workspace-edit-underserved-population/component';
+import {CdrVersion, DataAccessLevel, Workspace} from 'generated/fetch';
+import * as fp from 'lodash/fp';
+import * as React from 'react';
+
+export const ResearchPurposeItems = {
+  diseaseFocusedResearch: {
+    name: 'diseaseFocusedResearch',
```
blrubenstein: Does this need name? If not, probably worth removing. (I can't find anywhere it is used?)

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+          await workspacesApi().cloneWorkspace(
+            this.state.workspace.namespace, this.state.workspace.id,
+            {
+              includeUserRoles: this.state.cloneUserRole,
+              workspace: this.state.workspace
+            });
+        } else {
+          await workspacesApi()
+              .updateWorkspace(this.state.workspace.namespace, this.state.workspace.id,
+                  {workspace: this.state.workspace});
```
blrubenstein: When I changed one of the checkboxes on this, and updated workspace, I got a null pointer exception on the backend. It looks like it may not be passing up research purpose?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+    height: '10em',
+    resize: 'none',
+    padding: '0.1rem 0.2rem',
+    background: '#fff',
+    border: '1px solid #ccc',
+    color: '#000',
+    borderRadius: '.125rem'
+  },
+  shortDescription: {
+    color: '#262262',
```
blrubenstein: Optional Nit for all colors: You may be able to use `app/styles/colors.ts`? (We don't do that very much in the codebase right now, but I'd love to move in that direction)

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: I found this on a clone as well, but only the second clone I did?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+  routeConfigData: any;
+  workspace: Workspace;
+}
+
+export interface WorkspaceEditState {
+  cdrVersionItems: Array<CdrVersion>;
+  disableButton: Boolean;
+  workspace: Workspace;
+  workspaceCreationConflictError: Boolean;
+  workspaceCreationError: Boolean;
```
blrubenstein: I notice these are typed as `Boolean` whereas below are `boolean`? Is there a reason for that? If so, mind throwing a comment on here? If not, mind making all `boolean`?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+      };
+    }
+
+    componentWillMount() {
+      this.updateWorkspace();
+      this.setCdrVersions();
+    }
+
+
+    updateWorkspace() {
```
blrubenstein: Nit: This method name makes me think we are issuing an update request. Maybe `initializeWorkspace?`? It looks like it is always called on `componentWillMount`? It may also be worth just inlining there.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+      });
+      this.setState({cdrVersionItems: versions});
+      if (this.isMode(WorkspaceEditMode.Create)) {
+        this.setState(fp.set(['workspace', 'cdrVersionId'], versions[0].value));
+      }
+    }
+
+
+    openStigmatization() {
+      window.open('/definitions/stigmatization', '_blank');
```
blrubenstein: This feels like it should be an href, especially since we aren't doing any other logic here, and not a function used for onClick?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+    disableButton() {
+      return this.isEmpty('name') || this.isEmpty('description');
+    }
+
+    updateWorkspaceCategory(category, value) {
+      this.setState(fp.set(['workspace', 'researchPurpose', category], value));
+    }
+
+
+    updateUnderserverPopulation(populationDetails) {
```
blrubenstein: Nit: spelling. Should be `updateUnderservedPopulation`

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+
+    render() {
+      return <React.Fragment>
+        <div style={{width: '60%'}}>
+        <WorkspaceEditSection header={this.renderHeader()} tooltip={toolTipText.header}
+                              section={{marginTop: '24px'}} largeHeader required>
+          <div style={{display: 'flex', flexDirection: 'row'}}>
+            <TextInput type='text' style={styles.textInput} autoFocus placeholder='Workspace Name'
+                       value = {this.state.workspace.name}
+                       onChange={v =>    this.setState(fp.set(['workspace', 'name'], v))}/>
```
blrubenstein: Nit: There's some inconsistent spacing here.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,590 @@
...
+            </label>
+          </div>
+        </WorkspaceEditSection>
+        <div>
+          <div style={{display: 'flex', flexDirection: 'row', marginTop: '1rem',
+            marginBottom: '1rem'}}>
+            <Button type='secondary' style={{marginRight: '1rem'}}>
+              Cancel
+            </Button>
+            <TooltipTrigger content={[<ul>Missing Required Fields:
```
blrubenstein: Optional: For some reason, this tooltip appears super far from the button. This appears to exist on test as well, so no big deal, but if you wanted to fix it, wanted to bring it up.

NehaBroad:
@blrubenstein : The checkbox for underserved population in the current version is working in this version for you?

blrubenstein:
Yes, have verified that it is working. I will take another look at this after lunch!

ui/src/app/views/workspace-edit/component.tsx
NehaBroad: So since workspace in initialized in constructor and this is updating the initialized workspace as per the mode, i named it 'updateWorkspace'.

Also i find it neater to have it as a separate method than inline

ui/src/app/views/workspace-edit-underserved-population/component.tsx
NehaBroad: Question: So does component contains classes/modals that are used by multiple pages like button /alerts etc?
If no i can move this to component. However since it is a part of Workspace-Edit and will be used only by that , i prefer to see it in views.

ui/src/app/views/workspace-edit/component.tsx
NehaBroad: I am able to clone and update workspace. But there is some inconsistency regarding etag, i am checking that

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,320 @@
...
+      show: !this.state.show
+    });
+  }
+
+  focusCategoryChange(subCategory, value) {
+    if (!value) {
+      const index = this.state.value.findIndex(item => item === subCategory.id);
+      this.state.value.splice(index , 1);
+    } else if (value) {
+      this.state.value.push(subCategory);
```
panentheos: Mutating state in-place like this is not a good idea, since React has no way of knowing that the value has been modified. We'll be getting rid of this state anyway, but I wanted to mention it since it's an important principle to follow.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  flipShow() {
+    this.setState({
+      show: !this.state.show
+    });
+  }
+
+  focusCategoryChange(subCategory, value) {
+    if (!value) {
+      const index = this.state.value.findIndex(item => item === subCategory.id);
```
panentheos: FYI, take a look at the `toggleIncludes` utility function, which may be useful here.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,320 @@
...
+            return <div key={title}>
+              <h4 style={styles.title}>
+                {title}
+              </h4>
+              <div style={{display: 'flex', flexWrap: 'wrap'}}>
+                {
+                  categories.map(({id, label}) => {
+                    const isIncluded = fp.includes(id, this.state.value);
+                    return <div style={styles.row}>
+                      <input type='checkbox' style={styles.checkBox} checked={isIncluded}
```
panentheos: FYI, there's a `CheckBox` component that abstracts over the basic checkbox and unpacks the `{target: {value}}` structure in `onChange`

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+    longDescription: 'This research will focus on, or include findings on, distinguishing \
+    characteristics related to one or more underserved populations'
+  },
+  requestReview: {
+    shortDescription: 'Request a review of your research purpose'
+  }
+};
+
+export const toolTipText = {
+  header: 'A Workspace is your place to store and analyze data for a specific project.Each ' +
```
panentheos: FYI, backtick quotes allow multiline strings, which can make long blocks of text much nicer to edit.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+  cdrVersionItems: Array<CdrVersion>;
+  disableButton: boolean;
+  workspace: Workspace;
+  workspaceCreationConflictError: boolean;
+  workspaceCreationError: boolean;
+  cloneUserRole: boolean;
+  loading: boolean;
+}
+
+export const WorkspaceEdit = withRouteConfigData()(withCurrentWorkspace()(
```
panentheos: nit: If you have multiple HOCs, it's sometimes nicer to compose them with `fp.flow`, which limits the depth of nesting.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+  class extends React.Component<WorkspaceEditProps, WorkspaceEditState> {
+
+    constructor(props: WorkspaceEditProps) {
+      super(props);
+      this.isEmpty.bind(this);
```
panentheos: This is creating a bound method but not storing it anywhere, so it's not doing anything. I think this can be deleted.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+      this.state = {
+        cdrVersionItems: [],
+        disableButton: true,
```
panentheos: This piece of state looks unused.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+          }
+        },
+        workspaceCreationConflictError: false,
+        workspaceCreationError: false,
+        cloneUserRole: false,
+        loading: false
+      };
+    }
+
+    componentWillMount() {
```
panentheos: `componentWillMount` is deprecated. If we want to make ajax calls on mount, `componentDidMount` is preferred.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+          fp.set(['workspace', 'name'], ('Duplicate of ' + this.props.workspace.name)));
+      }
+    }
+
+    async setCdrVersions() {
+      const cdrVersions = await cdrVersionsApi().getCdrVersions();
+      // Convert cdrVersion to select input format of label and value
+      const versions = [];
+      cdrVersions.items.map((version , i) => {
+        versions.push({label: version.name, value: version.cdrVersionId});
```
panentheos: Rather than building this here, we could just store the raw data and then compute the options list during render.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+      }
+      this.setState({workspace : this.props.workspace});
+      if (this.isMode(WorkspaceEditMode.Clone)) {
+        this.setState(
+          fp.set(['workspace', 'name'], ('Duplicate of ' + this.props.workspace.name)));
+      }
+    }
+
+    async setCdrVersions() {
+      const cdrVersions = await cdrVersionsApi().getCdrVersions();
```
panentheos: This function needs a try/catch handler in case this call fails.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+            text='Share cloned workspace with same collaborators'/>
+        </div>
+        }
+        <WorkspaceEditSection header='Billing Account' subHeader='National Institutes of Health'
+            tooltip={toolTipText.billingAccount}
+            text='To fulfill program requirements, All of Us requests the following information
+            for each workspace.'/>
+        <WorkspaceEditSection header='Describe your research purpose'
+            tooltip={toolTipText.researchPurpose}
+            text={['Please include the ', <strong key='question'>research question</strong>,
```
panentheos: FYI: since the `text` prop just gets rendered, it can be anything renderable, including a `div` or `Fragment` which would probably be more convenient here.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+
+    updateWorkspace() {
+      if (this.isMode(WorkspaceEditMode.Create)) {
+        const profile =  userProfileStore.getValue().profile;
+        this.setState(fp.set(['workspace', 'namespace'], profile.freeTierBillingProjectName));
+        return;
+      }
+      this.setState({workspace : this.props.workspace});
```
panentheos: Rather than setting this after the component mounts, we could merge this logic into the constructor so we just initialize the state to the right values (remember that we have access to the props in the constructor as well).

ui/src/app/views/workspace-edit-underserved-population/component.tsx
blrubenstein: I am sort of leaning towards having views just be top level pages, and then sub-components either live in the page file, but this seems fine for now until we can have that conversation for sure.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,316 @@
...
+    width: '33.3%',
+    display: 'flex',
+    flexDirection: 'row',
+    height: '3rem'
+  },
+  checkbox: {
+    color: '#333333',
+    fontSize: 14,
+    fontWeight: 400,
+    lineHeight: '30px', paddingRight: '0.3rem'},
```
blrubenstein: Nit, curly brace on new line

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+export class WorkspaceUnderservedPopulation extends
+    React.Component<{value: Array<UnderservedPopulationEnum>, onChange: Function},
+    {show: boolean}> {
+  constructor(props: any) {
+    super(props);
+    this.state = {show: false};
+  }
+
+  flipShow() {
```
blrubenstein: Nit: This could probably just be inlined? I'm not sure it needs to be its own function.

ui/src/app/views/workspace-edit-underserved-population/component.tsx
```diff
@@ -0,0 +1,316 @@
...
+      {
+        id: UnderservedPopulationEnum.EDUCATIONINCOMELESSTHANTWENTYFIVETHOUSANDFORFOURPEOPLE,
+        label: 'Less than $25,000 for 4 people'
+      }
+    ]
+  }
+];
+
+export class WorkspaceUnderservedPopulation extends
+    React.Component<{value: Array<UnderservedPopulationEnum>, onChange: Function},
```
blrubenstein: I might call this `selectedValues` rather than `value`? I think it might be more clear.

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: That is completely reasonable, and I also find true, although in general I tend to think of functions as reuseable, but the idea with this is that it is called only once? I think that's why my preference is for inline.

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: Thank you for checking!

PR #1910, opened 2019-03-13
[RW-591][Risk=low] Rework get workspaces

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -30,6 +30,7 @@ import org.pmiops.workbench.db.dao.ConceptSetDao;
 import org.pmiops.workbench.db.dao.UserDao;
 import org.pmiops.workbench.db.dao.UserRecentResourceService;
 import org.pmiops.workbench.db.dao.UserService;
+import org.pmiops.workbench.db.dao.WorkspaceDao;
 import org.pmiops.workbench.db.dao.WorkspaceService;
 import org.pmiops.workbench.db.model.CdrVersion;
 import org.pmiops.workbench.db.model.Cohort;
@@ -66,6 +67,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
       Pattern.compile(NOTEBOOKS_WORKSPACE_DIRECTORY + "/[^/]+(\\.(?i)(ipynb))$");
 
   private final WorkspaceService workspaceService;
+  private final WorkspaceDao workspaceDao;
   private final CdrVersionDao cdrVersionDao;
   private final CohortDao cohortDao;
   private final ConceptSetDao conceptSetDao;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -91,6 +94,7 @@
       UserService userService,
       UserRecentResourceService userRecentResourceService) {
     this.workspaceService = workspaceService;
+    this.workspaceDao = workspaceDao;
```
calbach: tl;dr: Should just call `workspaceService.getDao()` where needed for consistency with the rest of the file.

Looking again at the rest of this controller - this is probably a broader issue.

Arguably it might be more consistent for access to go through a new method in WorkspaceService instead of straight to workspaceDao. I don't see clear rationale for why we'd use workspaceService for some things, and workspaceDao directly for others.

The existing code has this same issue by just calling `getDao()` on the service, which equally broke the abstraction.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -514,46 +518,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-          }
-          responseList.add(currentWorkspace);
-        }
+
+    for (org.pmiops.workbench.db.model.Workspace dbWorkspace : dbWorkspaces) {
+      org.pmiops.workbench.firecloud.model.WorkspaceResponse fcWorkspace = fcWorkspaces
+          .stream()
+          .filter(w -> w.getWorkspace().getWorkspaceId().equals(dbWorkspace.getFirecloudUuid()))
+          .collect(Collectors.toList())
+          .get(0);
```
calbach: Would prefer some error checks here rather than hitting the index out of bound exception. Previous code more-or-less had this. It's definitely possible for FC workspaces and Workbench workspaces to get out of sync.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -514,46 +518,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-          } else {
-            currentWorkspace.setAccessLevel(WorkspaceAccessLevel.fromValue(fcWorkspace.getAccessLevel()));
-          }
-          responseList.add(currentWorkspace);
-        }
+
+    for (org.pmiops.workbench.db.model.Workspace dbWorkspace : dbWorkspaces) {
+      org.pmiops.workbench.firecloud.model.WorkspaceResponse fcWorkspace = fcWorkspaces
+          .stream()
+          .filter(w -> w.getWorkspace().getWorkspaceId().equals(dbWorkspace.getFirecloudUuid()))
```
calbach: I realize this was carried over from before, but this is n^2. Should just create a map<workspace id, fcWorkspace> ahead of this loop.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: Yeah, okay. I totally missed that `getDao()` existed :( 
I think we call directly to the Dao when it is simple, and the Service when there is more complicated logic around what to do.

blrubenstein:
@calbach should be good for another look. PTAL

PR #1912, opened 2019-03-13
[RW-2240][risk=low] Put bypass variables in Profile

jacmrob:
follow-up to @UrsaStutsman's work storing the bypass variables in the db, first step to making them used in the UI is making them accessible from the profile.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
+++ b/api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
@@ -94,6 +94,9 @@ public class ProfileService {
     if (user.getEraCommonsCompletionTime() != null) {
       profile.setEraCommonsCompletionTime(user.getEraCommonsCompletionTime().getTime());
     }
+    if (user.getEraCommonsBypassTime() != null) {
+      profile.setEraCommonsBypassTime(user.getEraCommonsBypassTime().getTime());
+    }
     if (user.getDemographicSurveyCompletionTime() != null) {
       profile.setDemographicSurveyCompletionTime(user.getDemographicSurveyCompletionTime()
           .getTime());
--- a/api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
+++ b/api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
@@ -88,6 +88,12 @@ public class ProfileService {
     if (user.getTrainingCompletionTime() != null) {
       profile.setTrainingCompletionTime(user.getTrainingCompletionTime().getTime());
     }
+    if (user.getComplianceTrainingCompletionTime() != null) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -88,12 +88,21 @@ public Profile getProfile(User user) {
     if (user.getTrainingCompletionTime() != null) {
```
jacmrob: question-- will `trainingCompletion` eventually get replaced by `complianceTrainingCompletion`?

api/src/main/resources/workbench.yaml
```diff
@@ -2191,6 +2211,46 @@ definitions:
         type: integer
         format: int64
         description: Timestamp when the user completed era commons linking.
+      eraCommonsBypassTime:
+        type: integer
+        format: int64
+        description: Timestamp when the user was bypassed for completing era commons linking.
+      complianceTrainingCompletionTime:
```
calbach: opt: Alternatively we could implement this with a standard structure to avoid some duplication, e.g.:

```
eraCommonsAccess:
  ref: "$/WorkbenchAccess"

WorkbenchAccess:
  properties:
    completionTime:
    bypassTime:
```

I don't feel very strongly either way. It does add a few more potential null-ness states to worry about.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
calbach: Yes, not sure what the plan is on sequencing though. I think @gjuggler probably has the latest thinking on that.

gjuggler: Sorry for the delayed response here.

Yes, we'll eventually migrate the field name.

We'll want to do the migration (DB migration script, and update Moodle API calls to refer to the new field) before we turn on the access.enableComplianceTraining flag in production. I'll propose we take care of this (and other migrations) in the next sprint.

blrubenstein: @gjuggler we should talk to Karthik about that ASAP if so, I think we can do it, but we will need to make sure that gets prioritized in.

gjuggler: I'll make a note in https://precisionmedicineinitiative.atlassian.net/browse/RW-2331 (Athens production readiness). As it stands we're not quite ready for Moodle to be turned on, so this cleanup is a required part of that ticket. I don't think this reflects any shift in priorities – just that there's still a few steps left before we can just "switch it on"

blrubenstein: Got it. I can also sync up with Karthik in person on this on Monday, if there's anything we feel is unresolved

PR #1913, opened 2019-03-13
[risk=no] Fix docs and value_as_number index type

calbach:
Number may be floats and should be indexed as such.


calbach:
Ignoring the flaky UI test, since this PR does not touch UI code in any way.

PR #1918, opened 2019-03-14
[RW-2123][risk=no]Show a policy reminder before sending users to GCS bucket

NehaBroad:
Please note : Modal design is pending Lou approv
Screen Shot 2019-03-14 at 6 55 39 PM
al

### DIFF
--- a/ui/src/app/views/workspace/component.html
+++ b/ui/src/app/views/workspace/component.html
@@ -166,8 +166,7 @@
         </div>
       </div>
     </div>
-    <a href="https://console.cloud.google.com/storage/browser/{{workspace.googleBucketName}}?authuser={{username}}"
-        target="_blank" rel="noopener noreferrer">
+    <a rel="noopener noreferrer" (click)="googleBucketModal = true">
       Google Bucket
     </a>
   </div>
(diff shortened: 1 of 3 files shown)


ui/src/app/views/workspace/component.html
```diff
@@ -166,8 +166,7 @@ <h6>
         </div>
       </div>
     </div>
-    <a href="https://console.cloud.google.com/storage/browser/{{workspace.googleBucketName}}?authuser={{username}}"
-        target="_blank" rel="noopener noreferrer">
+    <a rel="noopener noreferrer" (click)="googleBucketModal = true">
       Google Bucket
```
blrubenstein: Do you still need the `rel` attribute here? Also, semantically, this probably shouldn't be an anchor tag anymore. Maybe a button link?

ui/src/app/views/workspace/component.html
```diff
@@ -177,6 +176,19 @@ <h6>
...
+  <h3 class="modal-title">Note</h3>
+  <div class="modal-body">It is All of Us data use policy that researchers should not make copies of or download
+    individual-level data (including taking screenshots or other means of viewing individual-level data) outside of
+    the All of Us research environment without approval from All of Us Resource Access Board (RAB).<br>
+    Notebooks should rarely be downloaded directly from Google Cloud Console, as output cells in Notebooks may
+    contain sensitive individual level data.
+  </div>
+  <div class="modal-footer">`
+    <button type="button" class="btn btn-secondary" (click)="googleBucketModal = false">Cancel</button>
+    <button type="button" class="btn btn-primary" (click)="openGoogleBucket()">Continue</button>
```
blrubenstein: This should probably be a styled anchor tag so that we get the nice benefits from cmd-click and accessibility benefits.

PR #1920, opened 2019-03-15
[RW-2175][RISK=NO]No error message on submit of account creation page if the email is i…

NehaBroad:


NehaBroad:
Screen Shot 2019-03-15 at 12 57 54 AM

ui/src/app/views/account-creation/component.tsx
```diff
@@ -191,6 +201,10 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
                      value={contactEmail}
                        style={{width: '16rem'}}
                        onChange={v => this.updateProfile('contactEmail', v)}/>
+          {this.state.invalidEmail &&
+          <Error id='invalidEmailError'>
+            Email Id is invalid
```
blrubenstein: Nit: I would probably say `Contact Email is invalid`

NehaBroad: Error message has been updated and yes the regex is picked up online as standard for email structure check

PR #1926, opened 2019-03-18
[RW-2236] [risk=low] eRA Commons sync cron job

gjuggler:
The main goal of this PR is to implement a cron job to periodically sync the eRA Commons status of all users.

When starting this work I noticed that a few things might be organized better, especially as we move towards (1) fleshing out the access module pattern, and (2) more cron-type synchronization jobs.

So if you'll pardon my refactoring here, I moved some of the code from Moodle / complianceTraining along with the eRA Commons code with the following pattern:
 - OfflineUserController: handlers for CRON endpoints related to user management.
 - UserService: service-level methods for syncing the access module status for a user (or an impersonated user via FireCloud).

The end result feels like a much cleaner abstraction: the OfflineUserController can fully mock out the User Service for test, and the (new) UserServiceTest holds unit tests for the actual communication with external services.


gjuggler:
Putting this out for slightly-early review: I haven't tested the cron job locally yet, but looking for comments / critiques from Neha and Scout, since this refactor touches access module code that's recently gone in.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -29,46 +47,131 @@
...
+    //
+    // If we haven't had our "firecloud-admin" service account whitelisted,
+    // then the following API call would result in a 401 error instead of a 404.
+    NihApi nihApi = new NihApi(apiClient);
+    int responseCode = 0;
+    try {
+      NihStatus nihStatus = nihApi.nihStatus();
+    } catch (ApiException e) {
+      responseCode = e.getCode();
+    }
```
NehaBroad: Expected Exceptions rule can be used here instead of try catch.

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
```diff
@@ -41,15 +68,25 @@
...
     try {
       return cloudStorageService.getGSuiteAdminCredentials();
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
 
+  @Lazy
+  @Bean(name=Constants.FIRECLOUD_ADMIN_CREDS)
+  GoogleCredential fireCloudCredentials(CloudStorageService cloudStorageService) {
```
NehaBroad: Like googleCredentials can this be also named fireCloudAdminCredentials?

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
```diff
@@ -74,4 +111,19 @@ WorkbenchConfig workbenchConfig(ConfigDao configDao) {
...
   }
+
+  @Bean
+  public Sleeper sleeper() {
+    return new ThreadWaitSleeper();
+  }
+
+  @Bean
+  public BackOffPolicy backOffPolicy(Sleeper sleeper) {
+    // Defaults to 100ms initial interval, doubling each time, with some random multiplier.
```
NehaBroad: Like the fact you mentioned the initial interval in comments 👍

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+import static org.mockito.Mockito.when;
+
+@RunWith(SpringRunner.class)
+@DataJpaTest
+@Import({LiquibaseAutoConfiguration.class})
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@AutoConfigureTestDatabase(replace= AutoConfigureTestDatabase.Replace.NONE)
+public class UserServiceTest {
+
+  private final String EMAIL_ADDRESS = "a@fake-research-aou.org";
```
blrubenstein: Nit: I believe we actually block users from making emails under 3 characters. I worry about potential for weird behavior if we don't follow that for tests as well. (Maybe `abc@fake-research-aou.org`?)

api/src/test/java/org/pmiops/workbench/firecloud/FireCloudServiceImplTest.java
```diff
@@ -55,16 +57,31 @@
...
+  private GoogleCredential impersonatedCredential;
+
+  private WorkbenchConfig workbenchConfig;
 
   @Rule
   public MockitoRule mockitoRule = MockitoJUnit.rule();
 
   @Before
   public void setUp() {
+    workbenchConfig = new WorkbenchConfig();
```
blrubenstein: Could you use the new `createEmptyConfig` here and then override what you want to override?

gjuggler:
> This all looks reasonable. For the future, we should think about if we can break up PRs like this. (I know I had a hard time going through all this, but it does seem like one thread. Mainly just around can we break PRs into smaller chunks)

Hey – I'm super sorry for the messy review! Unfortunately, I'd forgotten that this branch had some out of date changes from a prior PR that hadn't been rebased. So most, if not all of the FireCloudService stuff was outdated. I've resolved conversations which were related to those out-of-date changes.

My apologies again – I'm still figuring out how best to manage multiple PRs that depend on each other.

For the other comments, I'm working on those now and will re-request a review when this is ready & I've tested the cron job locally.

blrubenstein:
No worries at all, large PRs sometimes just happen! Just want to make sure we are continuing to think about this moving forwards.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
```diff
@@ -82,12 +82,16 @@ WorkspaceACLUpdateResponseList updateWorkspaceACL(String projectName, String wor
 
   void removeUserFromGroup(String email, String groupName);
 
-  boolean isUserMemberOfGroup(String groupName);
+  boolean isUserMemberOfGroup(String email, String groupName);
```
gjuggler: Note: the change here (*and in FireCloudServiceImpl and fireCloud.yaml) is a workaround / bugfix to allow the UserService.updateDataAccessLevel method to work correctly within the context of a Cron job, which is updating user data for more than just the currently-authenticated user.

More gory details:

The updateDataAccessLevel method needs to determine if a given user is a member of the registered-tier access group. To do this, we previously made a call like:

fireCloudService.isUserMemberOfGroup(registeredTierGroup);

and the fireCloudService would make a call to FireCloud using the *currently-authenticated user* credentials.

This works fine when handling end-user web requests. But when we're running a batch job, userService.updateDataAccessLevel may be called with a user that *isn't* currently-authenticated.

To fix this, I added the FireCloud API call that retrieves all members of a given group. A comment left by Dan Rodney in FireCloudServiceImpl had said that "There is no endpoint in FireCloud for checking whether a user is a member of a particular group". I can't tell whether Dan didn't recognize the GET method, or was concerned that fetching all users from the registered-tier group would be inefficient – which it will be, but I don't think we need to be too concerned about that for now. Even a list of ~10k emails won't be a huge issue, and we can lean on the FireCloud team to create a more focused API endpoint if we ever need it.

The call can now correctly propagate the target user email from UserService --> FireCloudService:

fireCloudService.isUserMemberOfGroup(userEmail, registeredTierGroup);

(Note: by removing this use of the "endUserGroups" API, we can also follow-up on this by cleaning up the awkward use of two API clients within FireCloudService.)

api/src/test/java/org/pmiops/workbench/firecloud/FireCloudServiceImplTest.java
gjuggler: This is a good idea, but this change shouldn't have been in the PR to begin with... I'll consider following-up with a cleanup on all unit tests which had been manually constructing WorkbenchConfigs.

gjuggler:
Adding @calbach in case the FireCloudService change could have known negative consequences (see my long comment above).

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -743,10 +746,9 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
     }
     JWTWrapper wrapper = new JWTWrapper().jwt(token.getJwt());
     try {
-      NihStatus nihStatus = fireCloudService.postNihCallback(wrapper);
-      User user = initializeUserIfNeeded();
```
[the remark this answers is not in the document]
calbach: Yep, I don't see any reason to be calling`initializeUser` here.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -412,4 +422,68 @@ public void syncUserTraining(User user) throws ApiException, NotFoundException {
...
+      NihStatus nihStatus = api.nihStatus();
+      setEraCommonsStatus(user, nihStatus);
+    } catch (org.pmiops.workbench.firecloud.ApiException e) {
+      if (e.getCode() == HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
+        // We'll catch the NOT_FOUND ApiException here, since we expect many users to have an empty
+        // eRA Commons linkage.
+        log.info(String.format("NIH Status not found for user %s", user.getEmail()));
+        return;
+      } else {
+        throw new org.pmiops.workbench.firecloud.ApiException(e);
```
calbach: Wrapping a firecloud exception in a firecloud exception - was that intentional?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -318,17 +318,10 @@ public void removeUserFromGroup(String email, String groupName) {
   }
 
   @Override
-  public boolean isUserMemberOfGroup(String groupName) {
+  public boolean isUserMemberOfGroup(String email, String groupName) {
     return retryHandler.run((context) -> {
-      // There is no endpoint in FireCloud for checking whether a user is a member of a particular
-      // group; so instead, fetch all the group memberships. (There won't be that many for our
-      // users anyway.)
-      for (ManagedGroupAccessResponse group : endUserGroupsApiProvider.get().getGroups()) {
```
calbach: This appears to have been the only usage of `endUserGroupsApiProvider`, should be able to remove

api/src/test/java/org/pmiops/workbench/api/OfflineUserControllerTest.java
```diff
@@ -0,0 +1,103 @@
...
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+public class OfflineUserControllerTest {
+
+  @TestConfiguration
+  @Import({
+      OfflineUserController.class
+  })
+  static class Configuration {
+    @MockBean
```
calbach: nit: This is interesting - I didn't even know you could do a mock this way. I think the pattern we have most of the other locations is to put this in an annotation on the `Configuration` itself, e.g.

https://github.com/all-of-us/workbench/blob/f3341eaf4b0988806120310bdac26e5d6b1b54b7/api/src/test/java/org/pmiops/workbench/api/AuditControllerTest.java#L56

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -70,6 +79,17 @@ public UserService(Provider<User> userProvider,
     this.complianceService = complianceService;
   }
 
+  /**
+   * Provides unfettered access to the DAO instance that sits beneath this service class.
+   *
+   * This accessor is meant primarily for tests, or scenarios where a controller needs to bypass
```
calbach: Is this entirely necessary? Is there a reason you can't just `@Autowire UserDao` straight into your unit test?

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+    Provider<WorkbenchConfig> configProvider = Providers.of(WorkbenchConfig.createEmptyConfig());
+    testUser = insertUser(EMAIL_ADDRESS);
+
+    userService = new UserService(Providers.of(testUser), userDao, adminActionHistoryDao, CLOCK,
+        new Random(), fireCloudService, configProvider, complianceService);
+  }
+
+  @After
+  public void tearDown() {
+    userDao.deleteAll();
```
calbach: This shouldn't be necessary based on my understanding of how spring data jpa testing works. If it is, please let me know what issue this is working around

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -0,0 +1,170 @@
...
+  private UserService userService;
+
+  private User testUser;
+
+  @Before
+  public void setUp() {
+    Provider<WorkbenchConfig> configProvider = Providers.of(WorkbenchConfig.createEmptyConfig());
+    testUser = insertUser(EMAIL_ADDRESS);
+
+    userService = new UserService(Providers.of(testUser), userDao, adminActionHistoryDao, CLOCK,
```
calbach: I think you've walked the right line here. This is a case where the provider value cannot be statically declared in the `@TestConfiguration`, since it needs to be initialized and potentially changed per test. I've not yet determined whether there's a way to dynamically change a provider value like this in tests, e.g. based on request scope or something

There's an alternate way of addressing this issue which is used in a few tests:
- In the test config, provide a `null` Bean for User (satisfying the Provider injection)
- Expose an `@VisibleForTesting setUserProvider()` method on the service
- Autowire the service, then override the userProvider in `setUp()` using this method

Example:
https://github.com/all-of-us/workbench/blob/f3341eaf4b0988806120310bdac26e5d6b1b54b7/api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java#L211

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
gjuggler: Yes – noted in my comment above. I think this will involve enough refactoring that I'd prefer to handle in a follow-up PR. Filed RW-2353 to track.

api/src/test/java/org/pmiops/workbench/api/OfflineUserControllerTest.java
gjuggler: Gotcha, thanks for the pointer. Done.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
gjuggler: Thanks – I guess the tl;dr is that this is going to be painful one way or another. I see how for WorkspacesControllerTest, the awkwardness of adding .setUserProvider to all controllers is offset by how much more boilerplate would have been required to use concrete constructors for the many controllers involved in the test.

But I also found it quite painful to debug and diagnose Spring-related errors, especially since they're only encountered when you actually run the tests. (Which is further exacerbated by how slowly this all executes on a macbook within Docker.)

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
gjuggler: Yep – just leftover from prior futzing around with the tests. Removed.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
calbach: Yes, I noticed this during a recent review. The existence of that method felt like it might be breaking some abstractions, but I didn't do any deeper investigation into the merits of that approach.

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -133,7 +133,7 @@ public void testStatusDev() throws Exception {
   }
 
   /**
-   * Ensure swe can successfully use delegation of authority to make FireCloud API calls on behalf
```
gjuggler: Apologies in advance for the unrelated typo fix. I find myself seeing these things, and not wanting to lose state on current work (e.g. stashing changes, creating a new branch, creating a separate PR... altogether takes me 5-10 mins) and end up making the documentation fix within the same branch. I'll try to catch myself earlier next time... if folks have suggestions on how they handle this, I'm all ears.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /**
+   * Updates moodle information for all users in the database.
+   *
+   * This API method is called by a cron job and is not part of our normal user-facing surface.
+   */
+  @Override
+  public ResponseEntity<Void> bulkSyncComplianceTrainingStatus() {
+    List<User> allUsers = userService.getAllUsers();
```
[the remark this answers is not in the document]
gjuggler: I ended up adding a short doc/cron-jobs.md file in this PR with some docs on cron jobs. I figure it's better to get something in (even if basic) sooner than later.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -0,0 +1,86 @@
...
+        errorCount++;
+        log.severe(String.format("Error syncing compliance training status for user %s: %s",
+            user.getEmail(), e.getMessage()));
+      }
+    }
+
+    if (errorCount > 0) {
+      throw new ServerErrorException(
+          String.format("%d errors encountered during complince training sync", errorCount));
+    } else {
```
calbach: opt_nit: I would probably drop the `else` here and below to keep the "normal" code flow easier to scan.

gjuggler: Ack. I heard similar feedback from Scout in a separate file context; I'll adopt the in-house style on this one :)

PR #1928, opened 2019-03-18
[RW-2331][Risk=moderate] Backfill Script for Bypass

blrubenstein:
To be merged before the release where we plan to turn on eRA commons, Moodle, and Beta Access. @gjuggler PTAL to verify backfill script.

I have verified locally that `id_verification_is_valid` is true if approved, false if rejected, and null if no action.

### DIFF
--- a/api/db/changelog/db.changelog-60-backfill-bypass.xml
+++ b/api/db/changelog/db.changelog-60-backfill-bypass.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+  <changeSet author="blrubenstein" id="changelog-60-backfill-bypass">
+    <sql>UPDATE user
+      SET era_commons_bypass_time = (SELECT NOW()),
+      compliance_training_bypass_time = (SELECT NOW()),
+      beta_access_completion_time = CASE WHEN id_verification_is_valid = true THEN (select NOW()) ELSE NULL END</sql>
+  </changeSet>
+</databaseChangeLog>
--- a/api/db/changelog/db.changelog-60-backfill-bypass.xml
+++ b/api/db/changelog/db.changelog-60-backfill-bypass.xml
@@ -8,7 +8,7 @@
   <changeSet author="blrubenstein" id="changelog-60-backfill-bypass">
     <sql>UPDATE user
       SET era_commons_bypass_time = CASE WHEN id_verification_is_valid = true THEN (select NOW()) ELSE NULL END,
-      compliance_training_bypass_time = (SELECT NOW()),
+      compliance_training_bypass_time = CASE WHEN id_verification_is_valid = true THEN (select NOW()) ELSE NULL END,
       beta_access_completion_time = CASE WHEN id_verification_is_valid = true THEN (select NOW()) ELSE NULL END</sql>
   </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


api/db/changelog/db.changelog-60-backfill-bypass.xml
```diff
@@ -0,0 +1,14 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+  <changeSet author="blrubenstein" id="changelog-60-backfill-bypass">
+    <sql>UPDATE user
+      SET era_commons_bypass_time = (SELECT NOW()),
+      compliance_training_bypass_time = (SELECT NOW()),
+      beta_access_completion_time = CASE WHEN id_verification_is_valid = true THEN (select NOW()) ELSE NULL END</sql>
```
calbach: Should use the same logic for `era_commons_bypass` as well

api/db/changelog/db.changelog-60-backfill-bypass.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+  <changeSet author="blrubenstein" id="changelog-60-backfill-bypass">
+    <sql>UPDATE user
```
calbach: There's an argument to backfill `data_use_agreement_bypass_time` here as well; I'll defer to @gjuggler on that.

api/db/changelog/db.changelog-60-backfill-bypass.xml
blrubenstein: Hmm... going to push back a bit on that, because our mapping is moving the current IDV process over to beta access, and then eRA commons is a new thing. (Because of the nature of it being a manual process, its functionally the same action for beta access, but something new for eRA

api/db/changelog/db.changelog-60-backfill-bypass.xml
blrubenstein: Fair. We will wait to hear from Greg

api/db/changelog/db.changelog-60-backfill-bypass.xml
calbach: I'd say we either don't backfill ERA commons here at all, or we match IDV. Currently this migration force-sets ERA commons on all users in Workbench regardless of whether they are trusted or were ever granted access to the system. They might be locked out for Bedford, but then would suddenly gain access with Clifton and skip ERA commons. That is clearly not what we want.

An even more conservative approach would revert this change, then get a list of users who need ERA commons exceptions for Algood (don't have an account) and manually bypass all of those.

blrubenstein: They shouldn't gain access though, because they don't have beta_access set. You will need all three to have registered access.

calbach: The beta_access requirement would be removed for Clifton (or later if Clifton is not a public release). Regardless I don't think it should be controversial to say we should avoid setting ERA commons bypasses where it is not necessary or warranted as that might cause access problems later on.

blrubenstein: Ah, okay. Got it. That makes sense, and is a strong argument to me to do it the cleaner way here. Will update.

api/db/changelog/db.changelog-60-backfill-bypass.xml
gjuggler: If we don't actually have the data use agreement click-through set up (which I believe is the case?) then we shouldn't backfill here. Presumably asking users to actually click through the agreement once we do implement & enable that module won't be an excessive burden.

blrubenstein:
@calbach fyi, per offline conversation with Greg and Jackie, we discussed merging this in before this release. Risk is relatively low, and it gives us time to get things set/figured out before we release for Athens.

blrubenstein:
Not anymore. I am going to rebase, add the changelog to the master, and then merge.

PR #1929, opened 2019-03-18
[No Ticket][Risk=no] Add back id verification

blrubenstein:

### DIFF
--- a/api/config/config_prod.json
+++ b/api/config/config_prod.json
@@ -54,6 +54,6 @@
     "enableComplianceTraining": false,
     "enableEraCommons": false,
     "enableDataUseAgreement": false,
-    "enableIdVerification": false
+    "enableIdVerification": true
   }
 }
--- a/api/config/config_stable.json
+++ b/api/config/config_stable.json
@@ -54,6 +54,6 @@
     "enableComplianceTraining": false,
     "enableEraCommons": false,
     "enableDataUseAgreement": false,
-    "enableIdVerification": false
+    "enableIdVerification": true
   }
 }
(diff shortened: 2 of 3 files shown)


api/config/config_prod.json
```diff
@@ -54,6 +54,6 @@
     "enableComplianceTraining": false,
     "enableEraCommons": false,
     "enableDataUseAgreement": false,
-    "enableIdVerification": false
+    "enableIdVerification": true
```
calbach: Why is `enableDataUseAgreement` also false? Is that correct?

blrubenstein: I don't believe we have anything in dataUseAgreement yet. It should be getting set though, so I don't believe we need it false?

UrsaStutsman: As far as I know we don't have that module enabled yet?

blrubenstein: Codewise I am seeing it not enabled. I think we used to autopopulate as part of the unregistered page? (@calbach correct me if I am wrong) but I don't think we lose anything by turning it off. From what I can see in the DB, the two things autopopulated are:
`terms_of_service_completion_time`
`demographic_survey_completion_time`

blrubenstein: I think we lost the `terms_of_service_completion_time` at some point though, I don't see it getting read anywhere in UserService

calbach: Yeah, this is fine, I wasn't clear on whether this was being implemented with the same field or a new one. Looks like it's a totally new field. Also the code is commented out so this wouldn't make much of a different. `false` should be the correct value for now then.

PR #1931, opened 2019-03-18
[RW-2263][risk=no] Implement counts for all types except(Deceased and BP)

freemabd1:
Implement counts for all types except(Deceased and BP). We now produce counts for all trees:
1) Surveys - All Surveys, Questions and Answers
2) PM - All types except **Blood Pressure**
3) Condition(ICD9/10, Snomed) - All types
4) Procedure(ICD9/10, CPT, Snomed) - All types
5) Demographics - All age, gender, race and ethnicity except **Deceased**
6) Drugs - All types
7) Measurements - All types
8) Visits - All types

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -23,7 +23,6 @@ import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.index.query.RangeQueryBuilder;
-import org.elasticsearch.index.query.TermsQueryBuilder;
 import org.pmiops.workbench.cdr.dao.CriteriaDao;
 import org.pmiops.workbench.cdr.model.Criteria;
 import org.pmiops.workbench.exceptions.BadRequestException;
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -245,11 +245,14 @@ public final class ElasticFilters {
       right = now.minusYears(Long.parseLong(attr.getOperands().get(0))).toLocalDate();
       if (attr.getOperands().size() > 1) {
         //use high end of the age range to calculate the low end(left) of the date range
-        left = now.minusYears(Long.parseLong(attr.getOperands().get(1))).minusYears(1).plusDays(1).toLocalDate();
+        //Ex: 2019-03-19(current date) - 55year(age) - 1 year = 1963-03-19
+        //Need to use GT to make sure not to include 1963-03-19 which evaluates to 56 years old
+        //which is out the range of 55. 1963-03-20 evaluates to 55 years 11 months 30 days.
+        left = now.minusYears(Long.parseLong(attr.getOperands().get(1))).minusYears(1).toLocalDate();
       }
       switch (attr.getOperator()) {
         case BETWEEN:
-          rq.gte(left).lte(right).format("yyyy-MM-dd");
+          rq.gt(left).lte(right).format("yyyy-MM-dd");
           break;
         default:
           throw new BadRequestException("Bad operator for attribute: " + attr.getOperator());
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -212,50 +212,49 @@ public final class ElasticFilters {
     if (AttrName.CAT.equals(attr.getName())) {
       //Currently the UI only uses the In operator for CAT which fits the terms query
       return QueryBuilders.termsQuery("events.value_as_concept_id", attr.getOperands());
-    } else {
-      Object left = null, right = null;
-      RangeQueryBuilder rq;
-      if (AttrName.NUM.equals(attr.getName())) {
-        rq = QueryBuilders.rangeQuery("events.value_as_number");
-        left = Float.parseFloat(attr.getOperands().get(0));
-        if (attr.getOperands().size() > 1) {
-          right = Float.parseFloat(attr.getOperands().get(1));
-        }
-        switch (attr.getOperator()) {
-          case LESS_THAN_OR_EQUAL_TO:
-            rq.lte(left);
-            break;
-          case GREATER_THAN_OR_EQUAL_TO:
-            rq.gte(left);
-            break;
-          case BETWEEN:
-            rq.gte(left).lte(right);
-            break;
-          case EQUAL:
-            rq.gte(left).lte(left);
-            break;
-          default:
-            throw new BadRequestException("Bad operator for attribute: " + attr.getOperator());
-        }
-        return rq;
-      } else if (AttrName.AGE.equals(attr.getName())) {
-        rq = QueryBuilders.rangeQuery("birth_datetime");
-        //use the low end of the age range to calculate the high end(right) of the date range
-        OffsetDateTime now = OffsetDateTime.now(ZoneOffset.UTC);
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
+              QueryBuilders.nestedQuery(
+                // We sum a constant score for each matching document, yielding the total number
+                // of matching nested documents (events).
+                "events", QueryBuilders.constantScoreQuery(b), ScoreMode.Total))
+              .setMinScore(occurredAtLeast));
+        }
       }
-      filter.filter(sgiFilter);
+      //SGI should "OR"
+      filter.should(sgiFilter);
```
calbach: Oops, thanks

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
+          case BETWEEN:
+            rq.gte(left).lte(right).format("yyyy-MM-dd");
+            break;
+          default:
+            throw new BadRequestException("Bad operator for attribute: " + attr.getOperator());
+        }
+        return rq;
+      }
+    }
+    throw new RuntimeException("attribute name is not an attr name type: " + attr.getName());
```
calbach: nit: would go for BadRequestException here, technically it is a malformed request.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
 
     return filter;
   }
 
+  private static QueryBuilder attributeToQuery(Attribute attr) {
+    //Attributes with a name of CAT map to the value_as_concept_id column
+    if (AttrName.CAT.equals(attr.getName())) {
+      //Currently the UI only uses the In operator for CAT which fits the terms query
+      return QueryBuilders.termsQuery("events.value_as_concept_id", attr.getOperands());
+    } else {
```
calbach: nit: this `else` isn't needed since you return above - can just unindent the rest

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
+        }
+        return rq;
+      } else if (AttrName.AGE.equals(attr.getName())) {
+        rq = QueryBuilders.rangeQuery("birth_datetime");
+        //use the low end of the age range to calculate the high end(right) of the date range
+        OffsetDateTime now = OffsetDateTime.now(ZoneOffset.UTC);
+        right = now.minusYears(Long.parseLong(attr.getOperands().get(0))).toLocalDate();
+        if (attr.getOperands().size() > 1) {
+          //use high end of the age range to calculate the low end(left) of the date range
+          left = now.minusYears(Long.parseLong(attr.getOperands().get(1))).minusYears(1).plusDays(1).toLocalDate();
```
calbach: Why +1 day? Comment?

Shouldn't GT vs GTE handle this?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
+            break;
+          default:
+            throw new BadRequestException("Bad operator for attribute: " + attr.getOperator());
+        }
+        return rq;
+      } else if (AttrName.AGE.equals(attr.getName())) {
+        rq = QueryBuilders.rangeQuery("birth_datetime");
+        //use the low end of the age range to calculate the high end(right) of the date range
+        OffsetDateTime now = OffsetDateTime.now(ZoneOffset.UTC);
+        right = now.minusYears(Long.parseLong(attr.getOperands().get(0))).toLocalDate();
```
calbach: It seems like there's an assumption that these are actually integers. Is that true? Otherwise if you actually support floats the below -1y logic doesn't really make sense.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -157,31 +165,99 @@ private QueryBuilder searchGroupToFilter(SearchGroup sg) {
...
       for (SearchParameter param : sgi.getSearchParameters()) {
         String conceptField = "events." + (isStandardConcept(param) ? "concept_id" : "source_concept_id");
-        BoolQueryBuilder b = QueryBuilders.boolQuery()
-            .filter(QueryBuilders.termsQuery(conceptField, toleafConceptIds(ImmutableList.of(param))));
+        if (isNonNestedSchema(param)) {
+          conceptField = nonNestedFields.get(param.getSubtype());
+        }
+        Set<String> leafConceptIds = toleafConceptIds(ImmutableList.of(param));
+        BoolQueryBuilder b = QueryBuilders.boolQuery();
+        if (!leafConceptIds.isEmpty()) {
```
calbach: Why would this be empty now? For the demo tree is the concept information not provided?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -241,7 +322,9 @@ private static boolean isStandardConcept(SearchParameter param) {
                 .map(id -> Long.toString(id))
                 .collect(Collectors.toSet()));
       } else {
-        out.add(Long.toString(param.getConceptId()));
+        if (param.getConceptId() != null) {
+          out.add(Long.toString(param.getConceptId()));
```
calbach: Same as above - is it legal for a search parameter to correspond to 0 matching concept IDs?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -241,7 +326,9 @@ private static boolean isStandardConcept(SearchParameter param) {
                 .map(id -> Long.toString(id))
                 .collect(Collectors.toSet()));
       } else {
-        out.add(Long.toString(param.getConceptId()));
+        if (param.getConceptId() != null) {
```
calbach: nit: else if ()

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -241,7 +326,9 @@ private static boolean isStandardConcept(SearchParameter param) {
...
+          out.add(Long.toString(param.getConceptId()));
```
calbach: nit: Could you add a comment here explaining that some search parameters are attribute/modifiers only.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
calbach: I'm thinking with this change that `sgiFilter` is no longer necessary. It's basically:

```
filter = sgiFilter1 || sgiFilter2 || ...
sgiFilter = nestedFilter1 || nestedFilter2 || ... || nonNestedFilter1 || nonNestedFilter2 || ...
```

calbach: No need to include that in this PR unless you feel inspired. I think the change would just be:
- Remove `sgiFilter`
- Replace existing mentions of `sgiFilter` with `filter`

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
freemabd1: Done.

PR #1935, opened 2019-03-19
[RW-2240][risk=no] Admin bypass UI

jacmrob:


jacmrob:
Unopened: 
Screen Shot 2019-03-19 at 4 00 28 PM

Opened: 
Screen Shot 2019-03-19 at 4 00 20 PM

Modified:
Screen Shot 2019-03-19 at 4 00 35 PM

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -233,6 +241,76 @@ public User apply(User user) {
     });
   }
 
+  public User setDataUseAgreementBypassTime(Timestamp bypassTime) {
```
gjuggler: I think, instead of adding a whole bunch of new methods here, we could make relatively concise calls to updateUseWithRetries directly from ProfileController using lambdas, e.g. something like:

case "dataUseAgreement":
        user = userService.updateWithRetries(user -> {
          user.setDataUseAgreementBypassTime(valueToSet);
          return user;
        });
        break;

ui/src/app/guards/access-tasks-guard.service.ts
```diff
@@ -25,8 +25,11 @@ export class AccessTasksGuard implements CanActivate, CanActivateChild {
     }
     return this.serverConfigService.getConfig().flatMap((config) => {
       return this.profileStorageService.profile$.flatMap((profile) => {
-        if (profile.eraCommonsCompletionTime === null
-            && config.enforceRegistered && environment.enableComplianceLockout) {
```
gjuggler: The logic here is starting to scare me a little... for example, do we have a clear understanding of the difference between config.enforceRegistered and environment.enableComplianceLockout? AFAICT, they're only ever checked in tandem on the frontend. And it's strange that "complianceLockout" boolean is controlling eraCommons logic, when eraCommons has nothing to do with compliance training. I worry we're at risk of confusing ourselves into oblivion here if we don't proactively clean this up.

Just brainstorming here, I can think of a couple options:

1) Use dataAccessLevel enum to drive all frontend logic.

Since we're already implementing all of this logic in the backend (in UserService.updateDataAccessLevel), we could just use the output of that method, which sets the dataAccessLevel enum, to drive our guard logic here. For example:

if (profile.dataAccessLevel == DataAccessLevel.UNREGISTERED) {
  // Re-route to homepage, where access tasks would be expected to be shown.
}

2. Use backend config flags to drive all access module logic.

Since we're already propagating the backend config into the frontend via an API method, we could drive all of this lockout from purely the config flags, where we're already set up with per-module flags and a global enforceRegistered flag. (But note that part of our plan is to eventually remove enforceRegistered, since all access control should be mediated by access modules. That's for later, though.) For example:

if (profile.eraCommonsCompletionTime === null && profile.eraCommonsBypassTime === null &&
    config.access.enableEraCommons && config.firecloud.enforceRegistered) {
  // Navigate to homepage where access tasks should be shown.
}


Separate from the above consideration, we also have an issue where much of the access tasks logic is duplicated both here and in views/homepage/component.tsx L361. It would be awesome if we could centralize the logic somehow, maybe by pulling it out into a utility method that takes in the profile & config and decides if the user should be blocked, and which modules should be shown to the user. So something like:

// access-tasks-guard.service.ts
if (!accessUtils.shouldAccessWorkbench(profile, config)) {
  // Navigate to homepage where access tasks should be shown.
}

// homepage/component.tsx
if (!accessUtils.shouldAccessWorkbench(profile, config)) {
  this.setState({
    accessTasksRemaining: true
    accessTasksLoaded: true
  });
}


Thoughts? I'm just thinking aloud here; maybe there are good reasons not to move in this direction, now or ever.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jacmrob: @blrubenstein what do you think of this?  i know pulling out those db call wrappers into the UserService was a convention thing. not sure how i feel about having it done differently in just one place

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -116,18 +116,28 @@ private void updateDataAccessLevel(User user) {
...
+
+    // TODO: Add in when we add idVerification module
+    // boolean idVerificationCompliant = user.getIdVerificationCompletionTime() != null ||
+    //   user.getIdVerificationBypassTime() != null || !configProvider.get().access.enableIdVerification ||
+    // TODO: Add in when we add demographics survey
+    // boolean demographicSurveyComplete = user.getDemographicSurveyCompletionTime() != null;
+    // // TODO: can be removed once we totally move off old validation
+    //   Optional.ofNullable(user.getIdVerificationIsValid()).orElse(false);
+
+    log.severe(String.valueOf(betaAccessGranted));
```
gjuggler: nit: remove before merging

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -116,18 +116,28 @@ private void updateDataAccessLevel(User user) {
...
     boolean complianceTrainingCompliant = user.getComplianceTrainingCompletionTime() != null ||
       user.getComplianceTrainingBypassTime() != null || !configProvider.get().access.enableComplianceTraining;
-    boolean idVerificationCompliant = user.getIdVerificationCompletionTime() != null ||
-      user.getIdVerificationBypassTime() != null || !configProvider.get().access.enableIdVerification ||
-    // TODO: can be removed once we totally move off old validation
-      Optional.ofNullable(user.getIdVerificationIsValid()).orElse(false);
+    boolean betaAccessGranted = user.getBetaAccessBypassTime() != null ||
+            !configProvider.get().access.enableBetaAccess;
+
+    // TODO: Add in when we add idVerification module
```
gjuggler: IMO we should probably just clear out these TODOs. I've heard reasonable people argue that commented-out code should categorically never be left in place. I try not to be too much of a stickler there, but at this point I think the patterns are well-set enough that we don't need to keep this around.

PR #1937, opened 2019-03-19
[No ticket][Risk=no] Fix routing infinite loop

blrubenstein:


ui/src/app/guards/access-tasks-guard.service.ts
```diff
@@ -20,6 +20,7 @@ export class AccessTasksGuard implements CanActivate, CanActivateChild {
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot):
   Observable<boolean> {
     if (route.routeConfig.path === '' ||
+        route.routeConfig.path.startsWith('unregistered') ||
```
gjuggler: I'd add a short comment describing what's happening here, and why this line fixes the issue – otherwise this is slightly cryptic, and may be very difficult for a naive reader to understand what's going on. (Especially when the PR isn't attached to a Jira ticket w/ more context.)

PR #1943, opened 2019-03-22
[RW-2115][risk=no] elasticsearch API for demographic charts

freemabd1:
Elasticsearch API for demographic charts
Not much unit/integration testing around this since we can't currently test es responses very easily.
Had to make a change to the indexing to account for 'Unknown'  for races/ethnicities


api/src/main/java/org/pmiops/workbench/elasticsearch/AggregationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public static final String RANGE_19_44 = "19-44";
+  public static final String RANGE_45_64 = "45-64";
+  public static final String RANGE_GT_65 = "65";
+  public static final String DATE = "d_";
+  public static final String GENDER = "g_";
+  public static final String RACE = "r_";
+
+  /**
+   * Comment
```
calbach: nit: replace or remove

api/src/main/java/org/pmiops/workbench/elasticsearch/AggregationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+      .addRange((start == null) ? null : start.toString(), end.toString())
+      .subAggregation(AggregationBuilders
+        .terms(GENDER + ageRange)
+        .field("gender_concept_name")
+        .order(BucketOrder.key(true))
+        .subAggregation(AggregationBuilders
+          .terms(RACE + ageRange)
+          .field("race_concept_name")
+          .order(BucketOrder.key(true))
+          .minDocCount(1)));
```
calbach: nit: Why is this param needed? Comment?

api/src/main/java/org/pmiops/workbench/elasticsearch/AggregationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    LocalDate start = null;
+    //use the low end of the age range to calculate the end of the date range
+    LocalDate end = ElasticUtils.todayMinusYears(Integer.parseInt(ages[0]));
+    if (ages.length > 1) {
+      //use high end of the age range to calculate the start of the date range
+      //need to add 1 year to adjust the start date to beginning of age range
+      start = ElasticUtils.todayMinusYears(Integer.parseInt(ages[1]) + 1);
+    }
+
+    return AggregationBuilders
```
calbach: nit: Could you add a comment about the bucket ordering? Is this because you'll be pulling them back out in the same ordre?

api/src/main/java/org/pmiops/workbench/elasticsearch/AggregationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    List<String> demoInformation = new ArrayList<>();
+    for (String ageRange : ageRanges) {
+      ParsedDateRange parsedDate = searchResponse.getAggregations().get(DATE + ageRange);
+      for (Range.Bucket dateBucket : parsedDate.getBuckets()) {
+        Terms gender = dateBucket.getAggregations().get(GENDER + ageRange);
+        for (Terms.Bucket genderBucket : gender.getBuckets()) {
+          Terms race = genderBucket.getAggregations().get(RACE + ageRange);
+          for (Terms.Bucket raceBucket : race.getBuckets()) {
+            demoInformation.add(
+              new JSONObject()
```
calbach: nit: I'd probably just return a DomainChartInfo directly here since I'm not sure this can really be used for anything else, though I appreciate the effort to return a generic response here.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
```diff
@@ -59,11 +59,11 @@ String lower() {
   public static final Map<String, Object> PERSON_SCHEMA = ImmutableMap.<String, Object>builder()
       .put("birth_datetime", esType(ElasticType.DATE))
       .put("gender_concept_id", esType(ElasticType.KEYWORD))
-      .put("gender_concept_name", esType(ElasticType.TEXT))
+      .put("gender_concept_name", esType(ElasticType.KEYWORD))
```
calbach: Note: I was wrong about the keyword subfield - that's just the default ES behavior. We could add a field type like this, but I don't think it's worth indexing this as text anyways (it's a small handful of values).

```
                "type" : "text",
                "fields" : {
                  "keyword" : {
                    "type" : "keyword",
                    "ignore_above" : 256
                  }
                }
```

api/tools/src/main/resources/bigquery/es_person.sql
```diff
@@ -4,9 +4,9 @@ SELECT
   p.gender_concept_id,
   gc.concept_name gender_concept_name,
```
calbach: Is it needed here too?

api/src/main/java/org/pmiops/workbench/elasticsearch/AggregationUtils.java
freemabd1: Done.

api/tools/src/main/resources/bigquery/es_person.sql
freemabd1: We currently don't have an unmapped genders in synthetic or prod dbs. But i added it anyway to be safe. Done.

PR #1947, opened 2019-03-26
[RW-2240] [risk=moderate] Simplify frontend access guard and add backfill for betaAccessBypass

gjuggler:
Digging into what was causing QA failures for the upcoming weekly release, I found two largely-separate issues:

1) The frontend access guard involved two pieces of complex logic based on a combination of config.enforceRegistered and environment.enableComplianceLockout. There was a subtle error in homepage/component.tsx that was causing access tasks not to be shown to new and non-beta-approved users.

I realized that the logic would be simpler to reason about & test if we removed the front-end "enableComplianceLockout" flag altogether. With this change, all front-end access decisions are made purely based on whether the user's data access level is registered or not (and whether config.enforceRegistered is true).

2) There was an issue where our bypass backfill script was updating the betaAccessCompletionTime, but the bypass UI and UserService were only looking at betaAccessBypassTime. Updating the backfill script to also change betaAccessBypassTime will cause existing users to be correctly granted access by default.

### DIFF
--- a/ui/src/app/guards/registration-guard.service.ts
+++ b/ui/src/app/guards/registration-guard.service.ts
@@ -10,7 +10,6 @@ import {Observable} from 'rxjs/Observable';
 import {ProfileStorageService} from 'app/services/profile-storage.service';
 import {ServerConfigService} from 'app/services/server-config.service';
 import {hasRegisteredAccess} from 'app/utils';
-import {environment} from 'environments/environment';
 
 
 @Injectable()
--- a/ui/src/app/guards/registration-guard.service.ts
+++ b/ui/src/app/guards/registration-guard.service.ts
@@ -33,18 +33,12 @@ export class RegistrationGuard implements CanActivate, CanActivateChild {
           return Observable.from([true]);
         }
         return this.profileStorageService.profile$.flatMap((profile) => {
-          // TODO: Remove this logic when enableComplianceLockout feature goes in
-          //   Since dataAccessLevel is computed using the the training and eraCommons modules,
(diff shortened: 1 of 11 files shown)


ui/src/app/guards/registration-guard.service.ts
```diff
@@ -33,18 +32,12 @@ export class RegistrationGuard implements CanActivate, CanActivateChild {
...
-          //   Since dataAccessLevel is computed using the the training and eraCommons modules,
-          //   for now recalculate it using only betaAccess and emailVerification.
-          const hasAccess = environment.enableComplianceLockout ?
-              hasRegisteredAccess(profile.dataAccessLevel) :
-              (!!profile.betaAccessBypassTime &&
-                profile.emailVerificationStatus === 'subscribed');
-          if (hasAccess) {
+          if (hasRegisteredAccess(profile.dataAccessLevel)) {
             return Observable.from([true]);
+          } else {
```
blrubenstein: Nit: you don’t need the else clause because the return above will short circuit.

gjuggler: Ack. My personal aesthetic has always been for evenly-balanced if/else statements in situations like this (where the first return isn't obviously a "return early" kind of scenario). But if there's a strong codebase style in the other direction, I'll be happy to follow-up with a change.

blrubenstein: Not strong enough that I think it’s something super important to do. I remember getting the suggestion at some point a while ago, but explicit else seems fine.

PR #1951, opened 2019-03-27
[RW-2054][Risk=no] Show a warning on unsupported browsers

ericsong:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2054

Shows a warning when loading the app on unsupported browsers. The only supported browser is chrome >= 67


ui/src/main.ts
```diff
@@ -17,3 +18,36 @@ setupCustomValidators();
 ReactModal.defaultStyles.overlay = {};
 ReactModal.defaultStyles.content = {};
 platformBrowserDynamic().bootstrapModule(AppModule);
+checkBrowserSupport();
+
+function checkBrowserSupport() {
```
calbach: I think app component would probably be slightly more appropriate for this kind of initialization, unless there's a reason that won't work:

https://github.com/all-of-us/workbench/blob/2898b4b7cd2336fd529736dadf27301f581c31c7/ui/src/app/views/app/component.ts

ui/src/typings.d.ts
```diff
@@ -11,3 +11,5 @@ other source files but are globally available.
 */
 
 declare var ResizeObserver: any;
+
+declare module 'outdated-browser-rework';
```
blrubenstein: Nit: Add newline at end of file.

ui/src/main.ts
blrubenstein: 👍 to this. Maybe following the same pattern we used for `cookiesEnabled`

PR #1952, opened 2019-03-27
[RW-2359][risk=no] Fallback to BigQuery on isApproximate

freemabd1:
Fallback to BigQuery on isApproximate


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
```diff
@@ -53,7 +53,7 @@ public ElasticSearchService(CriteriaDao criteriaDao, Provider<WorkbenchConfig> c
     log.info("Elastic filter: "  + filter.value().toString());
     long count = client().count(new CountRequest(personIndex)
         .source(SearchSourceBuilder.searchSource().query(filter.value())), RequestOptions.DEFAULT).getCount();
-    return new ElasticFilterResponse<>(count, filter.isApproximate());
+    return new ElasticFilterResponse<>(count);
```
calbach: opt: Probably could remove this response if you want - it might touch a lot more code though, so optional.

api/src/main/java/org/pmiops/workbench/api/CohortBuilderController.java
```diff
@@ -311,6 +309,18 @@
...
 
+  /**
+   * This method helps determine what request can only be approximated by elasticsearch
+   * and must fallback to the BQ implementation.
+   *
+   * @param request
+   * @return
+   */
+  private boolean isApproximate(SearchRequest request) {
+    return request.getIncludes().stream().anyMatch(sg -> sg.getTemporal())
```
calbach: Could you capture blood pressure here?

freemabd1: Added BP and Deceased to this check.

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticSearchService.java
freemabd1: Done.

PR #1955, opened 2019-03-28
[No Ticket][Risk=low] Added foreign key

blrubenstein:
I think we missed this...


api/config/cdm/cdm_5_2.json
```diff
@@ -545,6 +545,7 @@
           "name": "visit_occurrence_id",
           "mode": "nullable",
           "description": "A foreign key to the visit in the VISIT_OCCURRENCE table during which the device was used."
+          "foreignKey": "visit_occurrence"
```
freemabd1: looks like your missing a comman

PR #1956, opened 2019-03-28
[RW-2219][RW-2167][risk=no] Add cohort action icons, navigation from review

dolbeew:
Add action icons to the edit cohort screen for reviewing, deleting and exporting a cohort to a notebook. Also added navigation to edit a cohort from the participants table in review.


ui/src/app/cohort-search/overview/overview.component.html
```diff
@@ -55,41 +31,87 @@ <h2 class="count-header" id="total-count">
...
+      </span>
+      <span *ngIf="(error$ | async) && !temporal.flag">
+        <clr-icon
+          [popper]="errorPopper"
+          [popperTrigger]="'hover'"
+          shape="error-standard"
+          size="18"></clr-icon>
+      </span>
+      <span *ngIf="!(isRequesting$ | async) && ((temporal?.flag && temporal.tempLength && (total$ | async) === null) || (total$ | async) === -1)">
+        - -
```
blrubenstein: Just to confirm, this is intentional? (Fine if it is, just strange visually for me)

ui/src/app/cohort-search/overview/overview.component.html
```diff
@@ -55,41 +31,87 @@ <h2 class="count-header" id="total-count">
...
+    </div>
+  </div>
+  <div class="card bg-faded">
+    <div class="card-header header-text">
+      Results By Gender, Age Range, and Race
+      <clr-icon shape="sort-by"
+        [class.is-info]="stackChart"
+        (click)="toggleChartMode()">
+      </clr-icon>
+      <!--<clr-icon shape="angle"-->
```
blrubenstein: Do you still need this commented out code? (Just making sure its still giving value.)

ui/src/app/cohort-search/overview/overview.component.html
```diff
@@ -162,3 +181,28 @@ <h2>Save New Cohort</h2>
     </button>
   </div>
 </clr-modal>
+<clr-modal #deleteModal
```
blrubenstein: I think you can use the confirmDeleteModal here: https://github.com/all-of-us/workbench/tree/master/ui/src/app/views/confirm-delete-modal 

Just want to make sure we aren't redoing code paths/getting out of sync stylistically. 

Example usage here: https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/resource-card/component.tsx#L414

ui/src/app/cohort-search/overview/overview.component.html
dolbeew: The double `-` is intentional. I think Shanta may have added the space in between, which looks strange to me also, so I removed it.

ui/src/app/cohort-search/overview/overview.component.html
dolbeew: AFAIK we don't need it, removed

ui/src/app/cohort-search/overview/overview.component.html
dolbeew: I did look at that component first and it would work but it didn't look like there was a way to set the custom message specified in the ticket (at least not without making changes to the component).

blrubenstein: I think we will want to support this verbiage change for all cohort deletion, so it probably makes sense to change that component. (To either support custom language, or change the delete language for cohorts)

dolbeew: Good call. I didn't think about deleting cohorts from the card having a different message.

I moved confirmDeleteModal to the cohort-common.module so I could use it in cohort-search.

PR #1957, opened 2019-03-28
[RW-2146][Risk=moderate] Srubenst/values

blrubenstein:
@freemabd1 to verify the bigquery schema request for values.
@NehaBroad to validate UI logic and look/feel
@panentheos ... there's definitely a cleaner React/LoDash way to do some of this... mind taking a look?

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
+++ b/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
@@ -130,13 +130,13 @@ public class BigQueryService {
     public FieldList getTableFieldsFromDomain(Domain d) {
         CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
         String tableName;
-        if (d.equals(Domain.CONDITION)) {
+        if (Domain.CONDITION.equals(d)) {
             tableName = "ds_condition_occurrence";
-        } else if (d.equals(Domain.PROCEDURE)) {
+        } else if (Domain.PROCEDURE.equals(d)) {
             tableName = "ds_procedure_occurrence";
-        } else if (d.equals(Domain.DRUG)) {
+        } else if (Domain.DRUG.equals(d)) {
             tableName = "ds_drug_exposure";
-        } else if (d.equals(Domain.MEASUREMENT)) {
+        } else if (Domain.MEASUREMENT.equals(d)) {
             tableName = "ds_measurement";
         } else {
             throw new BadRequestException("Invalid domain, unable to fetch fields from table");
--- a/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
+++ b/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
@@ -3,9 +3,11 @@ package org.pmiops.workbench.api;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.FieldValue;
 import com.google.cloud.bigquery.JobInfo;
 import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableResult;
 
 import java.util.Collections;
@@ -22,9 +24,11 @@ import org.joda.time.format.DateTimeFormatter;
 import org.pmiops.workbench.cdr.CdrVersionContext;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.model.CdrVersion;
+import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.exceptions.ForbiddenException;
 import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.exceptions.ServerUnavailableException;
+import org.pmiops.workbench.model.Domain;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -122,4 +126,23 @@ public String getDate(List<FieldValue> row, int index) {
         }
         return row.get(index).getStringValue();
     }
+
+    public FieldList getTableFieldsFromDomain(Domain d) {
+        CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
+        String tableName;
+        if (d.equals(Domain.CONDITION)) {
```
freemabd1: nit: you can reverse these to avoid NullPointerExceptions since the equals method is null safe. 
```suggestion
        if (Domain.CONDITION.equals(d)) {
```

api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
```diff
@@ -241,7 +242,7 @@ public void setUp() {
     // SpringBootTest, which causes problems with CdrDbConfig. Just construct the service and
     // controller directly.
     ConceptService conceptService = new ConceptService(entityManager, conceptDao);
-    conceptsController = new ConceptsController(conceptService, workspaceService,
+    conceptsController = new ConceptsController(bigQueryService, conceptService, workspaceService,
```
freemabd1: Do we need a test case in here that test this new endpoint? Or is that coming in a later PR?

api/src/main/resources/workbench.yaml
```diff
@@ -2003,6 +2023,23 @@ definitions:
         items:
           $ref: "#/definitions/DomainInfo"
 
+  DomainValuesResponse:
```
freemabd1: Maybe add a description property explaining what a DomainValue is?

api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
blrubenstein: I think I am going to a followup PR with tests.

ui/src/app/views/dataset/component.tsx
```diff
@@ -40,12 +43,34 @@ export const styles = {
...
+interface DomainValuePair {
+  domain: Domain;
+  value: string;
+}
+
+export const ValueListItem: React.FunctionComponent <
+  {domainValue: DomainValue, onSelect: Function}> =
+  ({domainValue, onSelect}) => {
+    return <div style={{display: 'flex', color: 'black', height: '1.2rem'}}>
+      <input type='checkbox' value={domainValue.value} onClick={() => onSelect()}
```
panentheos: Without a `checked` prop, this input is uncontrolled and won't necessarily reflect the underlying state.

ui/src/app/views/dataset/component.tsx
```diff
@@ -187,6 +254,16 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
     }
   }
 
+  selectDomainValue(domain: Domain, domainValue: DomainValue): void {
+    const origSelected = this.state.selectedValues;
+    const selectObj = {domain: domain, value: domainValue.value};
+    if (fp.includes(selectObj, origSelected)) {
```
panentheos: `includes` uses `===` for comparisons, which only uses reference identity on objects, so this probably isn't working as intended. Is there some kind of unique id that we can use instead of a compound object? If not, then we'll need to use e.g. `some` instead.

ui/src/app/views/dataset/component.tsx
```diff
@@ -169,13 +196,53 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      const newDomains = fp.without(origDomains, currentDomains) as unknown as Domain[];
+      const removedDomains = fp.without(currentDomains, origDomains);
+      const updatedValueSets =
+        valueSets.filter(valueSet => !(fp.contains(valueSet.domain, removedDomains)));
+      const updatedSelectedValues =
+        selectedValues.filter(selectedValue =>
+          !fp.contains(selectedValue.domain, removedDomains));
+      if (newDomains.length > 0) {
+        this.getValuesList(newDomains)
+          .then(newValueSets => this.setState((state) => ({
```
panentheos: If we're not using this `state`, then using the callback form here is unnecessary.

ui/src/app/views/dataset/component.tsx
```diff
@@ -169,13 +196,53 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+            selectedConceptSetIds: newSelectedConceptSets,
+            valueSets: updatedValueSets.concat(newValueSets),
+            selectedValues: updatedSelectedValues
+          })));
+      } else {
+        this.setState({selectedConceptSetIds: newSelectedConceptSets,
+          valueSets: updatedValueSets,
+          selectedValues: updatedSelectedValues});
```
panentheos: TOL: How large is the full set of all values? Would it be simpler to just load everything once on page load, and then have all the UI interactions be synchronous?

ui/src/app/views/dataset/component.tsx
blrubenstein: Got it. Will add.

ui/src/app/views/dataset/component.tsx
```diff
@@ -169,13 +196,53 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
...
+      const {valueSets, selectedValues} = this.state;
       const origSelected = this.state.selectedConceptSetIds;
+      let newSelectedConceptSets: number[];
       if (fp.includes(resource.id, origSelected)) {
-        this.setState({selectedConceptSetIds: fp.remove((c) => c === resource.id, origSelected)});
+        newSelectedConceptSets = fp.remove((c) => c === resource.id, origSelected) as number[];
       } else {
-        this.setState({selectedConceptSetIds: (origSelected).concat(resource.id)});
+        newSelectedConceptSets = (origSelected).concat(resource.id);
+      }
```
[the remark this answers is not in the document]
blrubenstein: 👍 Will do

ui/src/app/views/dataset/component.tsx
blrubenstein: That could very well be better. Will do that.

ui/src/app/views/dataset/component.tsx
blrubenstein: I was noticing that. There isn't a unique id, unfortunately, but I can use `some`.

ui/src/app/views/dataset/component.tsx
blrubenstein: Talked offline. Think that the asynchronous will be better as we add more domains.

ui/src/app/views/dataset/component.tsx
```diff
@@ -40,12 +43,35 @@ export const styles = {
...
+  domain: Domain;
+  value: string;
+}
+
+export const ValueListItem: React.FunctionComponent <
+  {domainValue: DomainValue, onSelect: Function, checked: boolean}> =
+  ({domainValue, onSelect, checked}) => {
+    return <div style={{display: 'flex', color: 'black', height: '1.2rem'}}>
+      <input type='checkbox' value={domainValue.value} onChange={() => onSelect()}
+             style={{height: 17, width: 17, marginLeft: 10, marginTop: 10,
```
NehaBroad: this can be pulled out in style variable

api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
blrubenstein: I ended up having the time, so I just added a contract test.

PR #1961, opened 2019-03-30
[RW-2360][risk=moderate] Initial Elastic Cloud configuration for querying/indexing

calbach:
- Include the scheme in the ES base url
- Enable basic auth for ES cloud usage
- More generalization of `load-es-index` command
- Minor perf optimization for index loading
- Enable ES in test/staging/stable


api/libproject/devstart.rb
```diff
@@ -1265,13 +1293,15 @@ def create_local_es_index(cmd_name, *args)
     '--delete-indices'
     # Gradle args need to be single-quote wrapped.
   ]).map { |f| "'#{f}'" }
-  common.run_inline %W{gradle elasticSearchIndexer -PappArgs=['create',#{create_flags.join(',')}]}
+  ServiceAccountContext.new(TEST_PROJECT).run do
```
gjuggler: I'm confused about the hard-coded TEST_PROJECT. Shouldn't this refer to opts.env project name as well?

api/libproject/devstart.rb
```diff
@@ -26,6 +26,10 @@
 DRY_RUN_CMD = %W{echo [DRY_RUN]}
 
 ENVIRONMENTS = {
+  "local" => {
+    :config_json => "config_local.json",
+    :cdr_versions_json => "cdr_versions_local.json"
+  },
   "all-of-us-workbench-test" => {
```
gjuggler: Side comment: it might be cleaner to eventually refer to all environments by keyword, and include "project_id" as a field here. That might help reduce some conditional logic when dealing with 'local', and make it easier if we (heaven forbid) have to change a project name in the future.

api/libproject/devstart.rb
```diff
@@ -1609,6 +1639,11 @@ def get_auth_domain(project)
   return JSON.parse(File.read("config/#{config_json}"))["firecloud"]["registeredDomainName"]
 end
 
+def get_es_base_url(env)
```
gjuggler: Side note: I didn't fully realize that our config_*.json files were also being used to control deploy-side scripts. I guess it's attractive to centralize the source of truth. But it feels a bit awkward (and possibly risky / difficult to reason about) that we're using the config to control both the runtime environment and some characteristics of how that environment gets deployed.

For instance: I would definitely feel less safe about making an emergency config push directly to the SQL database, given that the deploy script is assuming some consistency between the deploy logic and the pushed config.

I can't quite decide if this is this an actual risk, or just a theoretical / hypothetical. Food for thought...

api/libproject/devstart.rb
```diff
@@ -26,6 +26,10 @@
 DRY_RUN_CMD = %W{echo [DRY_RUN]}
 
 ENVIRONMENTS = {
+  "local" => {
+    :config_json => "config_local.json",
```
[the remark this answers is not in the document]
calbach: Added them. Also added a TODO to clean this up more generally. I think there's a need for something like this but I agree it's not quite right in the current iteration.

api/libproject/devstart.rb
calbach: Yes, I think there needs to be a refactoring  across commands here - added a TODO.

api/libproject/devstart.rb
```diff
@@ -1234,29 +1238,53 @@ def list_clusters(cmd_name, *args)
...
+    ->(opts, v) { opts.env = v},
+    "Environment to load into; 'local' or a GCP project name, e.g. " +
+    "'all-of-us-workbench-test'")
+  op.add_validator ->(opts) { raise ArgumentError unless ENVIRONMENTS.has_key? opts.env }
+
+  op.opts.cdr_version = 'cdr'
+  op.add_option(
+    "--cdr-version [VERSION]",
+    ->(opts, v) { opts.cdr_version = v},
+    "CDR version to specify, should match cdr_versions_*.json configurations. Defaults to 'cdr'")
```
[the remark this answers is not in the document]
calbach: Rewrote flag comment to hopefully make this clearer.

api/libproject/devstart.rb
calbach: Yes, thanks. Did that with a fallback to TEST_PROJECT for local runs.

api/libproject/devstart.rb
calbach: There's only a handful of commands here that do this. One of them is the command to `load_config` (which is actually the command that pushes the config). The rest of these are unrelated to deployment.

Without some higher order config language I'm not sure we can avoid this without duplicating flag values in various places.

api/db/changelog/db.changelog-master.xml
```diff
@@ -73,4 +73,5 @@
   <include file="changelog/db.changelog-63.xml"/>
   <include file="changelog/db.changelog-64.xml"/>
   <include file="changelog/db.changelog-65.xml"/>
+  <include file="changelog/db.changelog-66.xml"/>
```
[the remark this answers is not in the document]
calbach: I don't have a strong feeling either way. These files are pretty boilerplatey and it does require putting the string in multiple locations (file name, here, and within the file itself), which is annoying if the scope changes during the course of your PR or you hit a merge conflict.

Looks like @dolbeew was the maverick to break with convention. Any thoughts on this?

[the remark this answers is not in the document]
dolbeew: I don't have a strong preference either, that's just the way I was shown how to do it.  @freemabd1 may be able to give more info.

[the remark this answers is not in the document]
freemabd1: I'm fine either way, but Liquibase has best practices: https://www.liquibase.org/bestpractices.html

api/config/cdr_versions_staging.json
```diff
@@ -22,6 +23,7 @@
     "releaseNumber": 1,
     "numParticipants": 946237,
     "cdrDbName": "synth_r_2019q1_2",
+    "elasticIndexBaseName": "synth_r_2019q1_2",
```
freemabd1: So did you name this 'synth_r_2019q1_2' to be consistent with cdrDbName? Or was is the second version of the index?

api/config/cdr_versions_test.json
```diff
@@ -9,6 +9,7 @@
     "releaseNumber": 1,
     "numParticipants": 946237,
     "cdrDbName": "cdr20181213",
+    "elasticIndexBaseName": "synth_r_2019q1_2",
     "publicDbName": "public20181213"
```
freemabd1: Do we need this property anymore since databrowser is separate? Prob a change for a different PR.

api/config/cdr_versions_staging.json
calbach: It was for consistency with cdrDbName. There's an argument to just use _1 here and make it correspond to the elastic index version instead. Don't feel too strongly on this (will probably be a follow-up if we want to change the convention).

api/config/cdr_versions_test.json
calbach: Right, will no longer be needed. There is a ticket near the top of the backlog for tearing down databrowser code

calbach:
Reindexed into AWS-backed Elastic cloud. Per discussion with security, enabling this in stable (and not prod).

PR #1962, opened 2019-04-01
[RW-2392] [risk=low] Fix eRA Commons logic, and finish module rename.

gjuggler:
This PR fixes an issue with the eRA Commons logic where we were too aggressive in returning error statuses from the UserService method.

Specifically: when we have a Moodle ID for a user, but that user hasn't completed any trainings, we seem to get an empty response from the Moodle API. (See examples in the RW-2392 description.) Instead of returning an error in that case, we should clear out the compliance training completion bit.

I also took the chance to finish up the rename of training --> complianceTraining.

### DIFF
--- a/api/db/changelog/db.changelog-67-compliance-training-expiration.xml
+++ b/api/db/changelog/db.changelog-67-compliance-training-expiration.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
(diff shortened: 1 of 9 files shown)


api/db/changelog/db.changelog-66-compliance-training-expiration.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="gjuggler" id="db.changelog-66-compliance-training-expiration">
+        <addColumn tableName="user">
+            <column name="compliance_training_expiration_time" type="datetime"/>
```
NehaBroad: Question: Will we be removing the existing training_expiration column?

gjuggler: Good point – I added in code to remove the existing column.

PR #1967, opened 2019-04-01
[RW-2111][risk=no] Visually distinguish stable and Prod environments

ericsong:
Add a label to the banner to indicate what environment is in use.

Screen Shot 2019-04-01 at 1 18 49 PM

### DIFF
--- a/ui/src/app/views/signed-in/component.html
+++ b/ui/src/app/views/signed-in/component.html
@@ -4,7 +4,14 @@
       <clr-icon shape="bars" (click)="sidenavToggle = !sidenavToggle"
           style="width: 1.5rem; height: 1.5rem; fill: #216FB4"></clr-icon>
     </div>
-    <img [src]="headerImg" class="header-image" routerLink="/">
+    <div>
+      <img [src]="headerImg" class="header-image" routerLink="/">
+      <div *ngIf="shouldShowDisplayTag" style="margin-left: 1rem;height: 12px;width: 126px;border-radius: 2px;background-color: #262262; box-shadow: 0 2px 5px 0 rgba(0,0,0,0.26), 0 2px 10px 0 rgba(0,0,0,0.16);">
+        <div style="color: #FFFFFF;	font-family: Montserrat;	font-size: 8px;	line-height:12px; text-align: center;">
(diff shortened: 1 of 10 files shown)


ui/src/app/views/signed-in/component.html
```diff
@@ -4,7 +4,14 @@
       <clr-icon shape="bars" (click)="sidenavToggle = !sidenavToggle"
           style="width: 1.5rem; height: 1.5rem; fill: #216FB4"></clr-icon>
     </div>
-    <img [src]="headerImg" class="header-image" routerLink="/">
+    <div>
+      <img [src]="headerImg" class="header-image" routerLink="/">
+      <div *ngIf="shouldShowDisplayTag" class="display-tag">
+        <div>
```
NehaBroad: the two div can be combined

ui/src/environments/environment-type.ts
```diff
@@ -15,6 +15,9 @@ export interface Environment {
   // Example value: 'Test' would cause the following full title:
   // "Homepage | [Test] All of Us Research Workbench"
   displayTag: string;
+  // Indicates if the displayTag should be shown in the web app. If it is true,
+  // a small label will be added under the "All of Us" logo in the header.
+  shouldShowDisplayTag: boolean;
```
gjuggler: I'm not sure this extra flag is necessary.

We already have a convention where in production the displayTag is an empty string, and app/component.ts (L89) then checks for a non-empty value in order to decide whether to add a prefix to the tab title.

It might be more consistent to just use the same convention for the in-app display tag as well.

Counterpoint: I can see an argument that having explicit "shouldShowDisplayTag" is clearer to understand. If we believe that's the case, then we should be consistent in how we're applying this logic, and use shouldShowDisplayTag in L89 of app/component.ts to control whether the prefix is included in the tab title.

Your call – after reading this through, my slight preference would be for adding explicit control to app/component.ts, but I can see it either way.

ericsong: I don't have a strong preference either but I think I prefer to keep it separate since I almost view the `displayTag` as an environment name. I think this opens us up to using it as such since its no longer tied to being an empty string.

PR #1968, opened 2019-04-01
[No Ticket][Risk=no] Dataset yaml object

blrubenstein:

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -2031,6 +2031,34 @@ definitions:
           type: integer
           format: int64
 
+  DataSet:
+    type: object
+    required:
+      - name
+    properties:
+      name:
+        type: string
+      conceptSetIds:
+        type: array
+        description: >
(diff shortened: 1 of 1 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -2031,6 +2031,34 @@ definitions:
           type: integer
           format: int64
 
+  DataSet:
```
NehaBroad: So i was thinking about the Data set response, and since we display just the name of the cohort/concept. Does it make sense to create an object : {id, name}. use that like
cohorts:
        type: array
        description: >
          The ids and names of all cohorts in the data set
        items:
          ref...
 This was we can have the same Request and Response object

blrubenstein: For displaying name, I believe we need to display all cohorts and concept sets, not just the ones selected for the DataSet? I've been using getCohortsFromWorkspace for fetching that information. (And then we can have the selected ones be based on Ids in the DataSet)

PR #1972, opened 2019-04-01
[RW-2381][Risk=low] linking table sql

blrubenstein:

### DIFF
--- a/api/db-cdr/generate-cdr/make-bq-dataset-linking.sh
+++ b/api/db-cdr/generate-cdr/make-bq-dataset-linking.sh
@@ -9,7 +9,7 @@ IFS=$'\n\t'
 # get options
 
 # --cdr=cdr_version ... *optional
-USAGE="./generate-clousql-cdr/make-bq-denormalized-dataset.sh --bq-project <PROJECT> --bq-dataset <DATASET>"
+USAGE="./generate-clousql-cdr/make-bq-dataset-linking.sh --bq-project <PROJECT> --bq-dataset <DATASET>"
 
 while [ $# -gt 0 ]; do
   echo "1 is $1"
@@ -60,7 +60,7 @@ bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
     DENORMALIZED_NAME               STRING,
     OMOP_SQL                        STRING,
     JOIN_VALUE                      STRING,
-    DOMAIN                          STRING,
+    DOMAIN                          STRING
 )"
 
 ################################################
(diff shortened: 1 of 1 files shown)


api/db-cdr/generate-cdr/make-bq-dataset-linking.sh
```diff
@@ -0,0 +1,127 @@
...
+################################################
+# INSERT DATA
+################################################
+echo "ds_linking - inserting conditions data"
+
+bq --quiet --project=$BQ_PROJECT query --nouse_legacy_sql \
+"INSERT INTO \`$BQ_PROJECT.$BQ_DATASET.ds_linking\`
+VALUES
+    ('PERSON_ID', 'a.PERSON_ID', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` a', 'Condition'),
+    ('CONDITION_CONCEPT_ID', 'a.CONDITION_CONCEPT_ID', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` a', 'Condition'),
```
gjuggler: IIUC, this table will be responsible for generating the user-facing SQL for a dataset – correct?

I don't think we should expect users to understand the reasoning behind these table names "a", "c1", "c2", etc.

If possible, we should switch to using meaningful (but concise) nicknames for these intermediate tables?

For example, a first pass on this table:

a -- > c_occurrence
c1 --> c_concept
c2 --> c_type
c3 --> visit
c4 --> c_source
c5 --> c_status

This would immediately make the resulting SQL much more interpetable to someone reading this for the first time. (And also easier for a developer to understand what's going on here.)

blrubenstein: Will change. That works for me.

blrubenstein:
@jscherdin @gjuggler this should be good for a look now. I think this is everything.

api/db-cdr/generate-cdr/make-bq-dataset-linking.sh
```diff
@@ -0,0 +1,271 @@
...
+# This generates the big query de-normalized tables for dataset builder.
+
+set -xeuo pipefail
+IFS=$'\n\t'
+
+
+# get options
+
+# --cdr=cdr_version ... *optional
+USAGE="./generate-clousql-cdr/make-bq-denormalized-dataset.sh --bq-project <PROJECT> --bq-dataset <DATASET>"
```
gjuggler: nit: fix usage script name

api/db-cdr/generate-cdr/make-bq-dataset-linking.sh
```diff
@@ -0,0 +1,271 @@
...
+    ('STANDARD_VOCABULARY', 'c_standard_concept.vocabulary_id as STANDARD_VOCABULARY', 'left join \`$BQ_PROJECT.$BQ_DATASET.concept\` c_standard_concept on c_occurrence.CONDITION_CONCEPT_ID = c_standard_concept.CONCEPT_ID', 'Condition'),
+    ('CONDITION_START_DATETIME', 'CONDITION_START_DATETIME', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` c_occurrence', 'Condition'),
+    ('CONDITION_END_DATETIME', 'CONDITION_END_DATETIME', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` c_occurrence', 'Condition'),
+    ('CONDITION_TYPE_CONCEPT_ID', 'CONDITION_TYPE_CONCEPT_ID', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` c_occurrence', 'Condition'),
+    ('CONDITION_TYPE_CONCEPT_NAME', 'c_type.concept_name as CONDITION_TYPE_CONCEPT_NAME', 'left join \`$BQ_PROJECT.$BQ_DATASET.concept\` c_type on c_occurrence.CONDITION_TYPE_CONCEPT_ID = c_type.CONCEPT_ID', 'Condition'),
+    ('STOP_REASON', 'STOP_REASON', 'from \`$BQ_PROJECT.$BQ_DATASET.condition_occurrence\` c_occurrence', 'Condition'),
```
gjuggler: nit: is this missing a c_occurrence. prefix, e.g. 'c.occurrence.STOP_REASON'? It may be valid SQL if the field name is unique among all joined tables, but it will seem clearer / more consistent to include the table prefix everywhere (here and for all other domains too).

blrubenstein: I can add this. 
It did actually run, so this is valid SQL, but being clearer can't hurt

PR #1975, opened 2019-04-02
[risk=no] Fix cohort builder API test

calbach:


calbach:
OK, done waiting for Codacy. This change is small enough that I'm not too worried (I'll fix any issues in a follow-up on the off-chance it finds something).

PR #1976, opened 2019-04-02
[risk=no] Escape JSON slashes...

calbach:


calbach:
Not waiting for tests, since tests clearly don't cover this. Will file an issue for that.

PR #1988, opened 2019-04-03
[RW-2104][Risk:LOW]Create Save Data Set Functionality

NehaBroad:


api/db/changelog/db.changelog-67-dataset.xml
```diff
@@ -0,0 +1,92 @@
...
+      <column name="name" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="values" type="clob">
+        <constraints nullable="false"/>
+      </column>
+      <column name="description" type="clob">
+        <constraints nullable="true"/>
+      </column>
+      <column name="creator_id" type="bigint">
```
blrubenstein: Do we need to worry about creator? I suspect this may not be something we actually need to store.

api/db/changelog/db.changelog-67-dataset.xml
```diff
@@ -0,0 +1,92 @@
...
+
+     <addForeignKeyConstraint baseColumnNames="data_set_id"
+        baseTableName="data_set_cohort_id"
+        constraintName="fk_data_set_data_set_id"
+        onDelete="CASCADE"
+        referencedColumnNames="data_set_id"
+        referencedTableName="data_set"/>  
+
+  </changeSet>
+</databaseChangeLog>
```
blrubenstein: Nit: Newline at EOF

api/db/changelog/db.changelog-master.xml
```diff
@@ -74,4 +74,5 @@
   <include file="changelog/db.changelog-64.xml"/>
   <include file="changelog/db.changelog-65.xml"/>
   <include file="changelog/db.changelog-66.xml"/>
+  <include file="changelog/db.changelog-67-dataset"/>
```
blrubenstein: I think you are missing the `.xml` at the end.

api/db/changelog/db.changelog-67-dataset.xml
```diff
@@ -0,0 +1,92 @@
...
+      <column name="data_set_id" type="bigint" autoIncrement="true">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="workspace_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
+      <column name="name" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="values" type="clob">
```
blrubenstein: Given what we said, should values just be a clob? Since we need to maintain domain/valueSet groupings, does it make more sense for this to be another sub table?

api/db/changelog/db.changelog-67-dataset.xml
NehaBroad: Should we  consistent with cohort and concept?

api/db/changelog/db.changelog-67-dataset.xml
NehaBroad: good point

api/db/changelog/db.changelog-67-dataset.xml
blrubenstein: If cohort and concept set store creator, then probably. I didn't realize they did :|

api/src/main/java/org/pmiops/workbench/db/model/DataSetValues.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public DataSetValues(String domainId, String value) {
+    this.domainId = domainId;
+    this.value = value;
+  }
+
+
+  @Column(name = "domain_id")
+  public String getDomainId() {
```
blrubenstein: If possible, it would be good to have this return the Domain itself, or make a helper method titled `getDomainEnum` and `setDomainEnum` to get and set via enum. (See the concept set model for an example)

ui/src/app/views/dataset/component.tsx
```diff
@@ -98,6 +106,10 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
       selectedCohortIds: [],
       valueSets: [],
       selectedValues: [],
+      save: false,
```
blrubenstein: Nit: I'd probably call this saving, rather than save, because it is only true while there is a save in progress.

blrubenstein: (On further look lower down, `creating` might be another good name.

ui/src/app/views/dataset/component.tsx
```diff
@@ -415,6 +455,30 @@ export const DataSet = withCurrentWorkspace()(class extends React.Component<
       <EditModal resource={resource}
                  onEdit={e => this.receiveEdit(e)}
                  onCancel={() => this.closeEditModal()}/>}
+      {this.state.save && <Modal>
+        <ModalTitle>Save Dataset</ModalTitle>
+        <ModalBody>
```
blrubenstein: Can you use `validate.js` here? That feels like semantically what you want.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  public void setConceptSetId(List<Long> conceptSetId) {
+    this.conceptSetId = conceptSetId;
+  }
+
+  @ElementCollection
+  @CollectionTable(name = "data_set_cohort", joinColumns = @JoinColumn(name = "data_set_id"))
+  @Column(name = "cohort_id")
+  public List<Long> getCohortSetId() {
```
blrubenstein: It might be nice to have something which returned the actual concept sets and cohorts here, rather than just the Ids (similar to values)

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+  @Override
+  public DataSet saveDataSet(String name, String description, long workspaceId,
+      List<Long> cohortIdList, List<Long> conceptIdList, List<DataSetValues> values, long creatorId,
+      Timestamp creationTime) {
+    DataSet dataSetDb = new DataSet();
+    dataSetDb.setName(name);
+    dataSetDb.setDescription(description);
+    dataSetDb.setWorkspaceId(workspaceId);
+    dataSetDb.setInvalid(true);
```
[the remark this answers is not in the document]
NehaBroad: So invalid is there to mark DataSet dirty if the concept or cohort of a dataset is deleted.

api/db/changelog/db.changelog-70-data-set.xml
```diff
@@ -0,0 +1,107 @@
...
+    <addForeignKeyConstraint baseColumnNames="creator_id"
+        baseTableName="data_set"
+        constraintName="fk_data_set_creator_id"
+        onDelete="SET NULL"
+        referencedColumnNames="user_id"
+        referencedTableName="user"/>
+    <addForeignKeyConstraint baseColumnNames="workspace_id"
+        baseTableName="data_set"
+        constraintName="fk_data_set_workspace_id"
+        onDelete="CASCADE"
```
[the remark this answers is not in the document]
NehaBroad: Confirmed This will safely delete data_set if the corresponding workspace is deleted

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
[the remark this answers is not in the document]
blrubenstein: Got it, okay. I would think that invalid would be falsy after creation then. (Because we haven't deleted any of those yet.)

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -0,0 +1,95 @@
...
+      throw new BadRequestException("Invalid Request: missing concept set ids");
+    } else if (dataSetRequest.getCohortIds() == null || dataSetRequest.getCohortIds().size() == 0) {
+      throw new BadRequestException("Invalid Request: missing cohort ids");
+    } else if (dataSetRequest.getValues() == null || dataSetRequest.getValues().size() == 0) {
+      throw new BadRequestException("Invalid Request: missing values");
+    }
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    long wId = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
+    List<DataSetValues> dataSetValuesList = dataSetRequest.getValues().stream().map(
+        (domainValueSet) -> {
```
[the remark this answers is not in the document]
NehaBroad: i tried that however it was throwing an error saying return is required

NehaBroad:
@blrubenstein I had change the Value data type to have 1-1 mapping of domain and value in the database rather than storing 1-many. (we can talk more about it if you think otherwise) because of that i made some changes to generateQuery

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -1,5 +1,32 @@
 package org.pmiops.workbench.api;
 
+import com.google.common.base.Strings;
+import com.google.common.collect.Streams;
+import org.pmiops.workbench.cdr.dao.ConceptDao;
```
blrubenstein: Nit: Small thing, but can you integrate these into the alphabetized imports below, so we don't have two separate lists?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -60,57 +107,140 @@
...
-  }
-}
 
 @RestController
 public class DataSetController implements DataSetApiDelegate {
-  private BigQueryService bigQueryService;
 
-  private CdrBigQuerySchemaConfigService cdrBigQuerySchemaConfigService;
+  private Provider<User> userProvider;
+  private final WorkspaceService workspaceService;
```
blrubenstein: Nit: If you could alphabetize these, that would be awesome.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -60,57 +107,140 @@
...
+  private CdrBigQuerySchemaConfigService cdrBigQuerySchemaConfigService;
+
   private ParticipantCounter participantCounter;
 
-  private WorkspaceService workspaceService;
 
 
   @Autowired
   DataSetController(
+      DataSetService dataSetService, Provider<User> userProvider,
```
blrubenstein: Nit: If you could alphabetize these, that would be awesome.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -60,57 +107,140 @@
...
+  }
+
+  static final Function<DataSetValues, DomainValuePair> TO_CLIENT_DOMAIN_VALUE =
+      new Function<DataSetValues, DomainValuePair>() {
+        @Override
+        public DomainValuePair apply(DataSetValues dataSetValue) {
+          DomainValuePair domainValuePair = new DomainValuePair();
+          domainValuePair.setValue(dataSetValue.getValue());
+          domainValuePair.setDomain(dataSetValue.getDomainEnum());
+          return  domainValuePair;
```
blrubenstein: Nit: It looks like there's an extra space after return.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -60,57 +107,140 @@
...
+      };
+
+  private ConceptSet toClientConceptSet(org.pmiops.workbench.db.model.ConceptSet conceptSet) {
+    ConceptSet result = ConceptSetsController.TO_CLIENT_CONCEPT_SET.apply(conceptSet);
+    if (!conceptSet.getConceptIds().isEmpty()) {
+      Iterable<org.pmiops.workbench.cdr.model.Concept> concepts =
+          conceptDao.findAll(conceptSet.getConceptIds());
+      result.setConcepts(Streams.stream(concepts)
+         .map(ConceptsController.TO_CLIENT_CONCEPT)
+          .collect(Collectors.toList()));
```
blrubenstein: Nit: There's an extra space before `.collect`

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -150,9 +281,8 @@ public String getStandardConceptIdColumn() {
     for (Domain d: domainList) {
       String query = "SELECT ";
       // VALUES HERE:
-      Optional<ValueSet> valueSetOpt = dataSet.getValues().stream()
-          .filter(valueSet -> valueSet.getDomain() == d)
-          .findFirst();
+      Optional<List<DomainValuePair>> valueSetOpt =  Optional.of(dataSet.getValues().stream()
```
blrubenstein: I don't think this needs to be an optional, and the check below can be `valueSetList.isEmpty()` (I think?)

api/src/main/resources/workbench.yaml
```diff
@@ -2040,7 +2074,50 @@ definitions:
...
         items:
-          $ref: "#/definitions/ValueSet"
+          $ref: "#/definitions/DomainValuePair"
+
+  DomainValuePair:
+    type: object
+    properties:
+      domain:
+        description: >
+          Domain corresponding to an OMOP table that can contain rows for the concepts in this
```
blrubenstein: It was pointed out to me that most of this description we don't need, since its just copy paste. (Maybe `Domain corresponding to an OMOP table`?

api/src/main/resources/workbench.yaml
```diff
@@ -2040,7 +2074,50 @@ definitions:
...
+          concept set. Note that the Domain values RACE, GENDER, and ETHNICITY are not allowed
+          here; it makes sense to specify concepts in these domains in cohort criteria, but
+          there isn't much value in having concept sets defined for them.
+        $ref: "#/definitions/Domain"
+      value:
+        type: string
+
+  DataSetResponse:
```
blrubenstein: I think this could just be `DataSet`? Because it could be part of other response objects.

PR #1994, opened 2019-04-05
[RW-2407][risk=no] adding cb_criteria and corresponding tables to local/cloud scripts

freemabd1:
adding cb_criteria and corresponding tables to local/cloud scripts


api/db-cdr/generate-cdr/bq-schemas/cb_criteria.json
```diff
@@ -0,0 +1,92 @@
...
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "name",
+    "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "value",
+    "type": "INTEGER"
```
jscherdin: I thought we made value a STRING?

freemabd1: I made it a string in liquibase/mysql. We can change it to a string here as well, but if we do that we have to recreate the origin table(all-of-us-ehr-dev.synthetic_cdr20180606.criteria)

jscherdin: That's fine. I just wanted to make sure this wasn't a problem

PR #1995, opened 2019-04-05
[RW-2291][risk=no]: Change "Duplicate" Functionality in workspace and dataset builder

ericsong:
I approached this by refactoring some of the Cohort creation logic in `CohortsController` and then adding an endpoint for duplicating cohorts.

I chose this approach over reusing the existing create cohort endpoint because I believe it is better to have a stricter and more verbose API that matches actions in the domain. With this approach, the client can duplicate a cohort by only providing an identifier instead of having to be aware of business logic such as appending `_2` to the name or dropping the cohort review models.


api/src/main/java/org/pmiops/workbench/api/CohortsController.java
```diff
@@ -136,28 +127,61 @@ public void setUserProvider(Provider<User> userProvider) {
...
-    // TODO: Make this a service method to avoid duplication with WorkspacesController
-    dbCohort.setCreator(userProvider.get());
-    dbCohort.setWorkspaceId(workspace.getWorkspaceId());
-    dbCohort.setCreationTime(now);
-    dbCohort.setLastModifiedTime(now);
-    dbCohort.setVersion(1);
+
+    org.pmiops.workbench.db.model.Cohort newCohort = cohortFactory.createCohort(cohort, userProvider.get(), workspace.getWorkspaceId());
+
+    if (cohortDao.findCohortByNameAndWorkspaceId(newCohort.getName(), workspace.getWorkspaceId()) != null) {
```
blrubenstein: Can you add this check above the creation via the factory? I think the information should be available there.

api/src/main/java/org/pmiops/workbench/api/CohortsController.java
```diff
@@ -136,28 +127,61 @@ public void setUserProvider(Provider<User> userProvider) {
...
+    }
+
+    try {
+      newCohort = cohortDao.save(newCohort);
+      userRecentResourceService.updateCohortEntry(workspace.getWorkspaceId(),
+              userProvider.get().getUserId(),
+              newCohort.getCohortId(),
+              new Timestamp(clock.instant().toEpochMilli()));
+    } catch (Exception e) {
+      throw new ServerErrorException(
```
blrubenstein: Opt-nit: Do we know what sorts of errors we could be seeing here? It would be good to throw something more descriptive than a 500 if we can.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -389,29 +393,20 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-    }
+
+    org.pmiops.workbench.db.model.Workspace finalDbWorkspace = dbWorkspace;
+    cloudStorageService.readAllDemoCohorts().stream()
+            .map(apiCohort -> cohortFactory.createCohort(apiCohort, userProvider.get(), finalDbWorkspace.getWorkspaceId())
+            ).forEach(dbCohort -> {
+              try {
+                dbCohort = cohortDao.save(dbCohort);
+              } catch (DataIntegrityViolationException e) {
+                throw new ServerErrorException(String.format(
```
blrubenstein: Nit: This should be a conflict exception I believe. 

Edit: Although, this is about loading in demos. This may actually not be an exception that we want to block creating a workspace on, but rather log and move on. I would probably approach it that way here, rather than throwing a `ServerErrorException`

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import org.pmiops.workbench.db.model.Cohort;
+import org.pmiops.workbench.db.model.User;
+
+public interface CohortFactory {
+
+    Cohort createCohort(org.pmiops.workbench.model.Cohort apiCohort, User creator, long workspaceId);
+
+    // duplicateCohort provides the functionality for the "Duplicate" Cohort function in the AoU workspace.
+    // The name is set to the original cohort's name appended by "_2"
+    // It is NOT a deep cloning method so some fields will not be copied over.
```
blrubenstein: What fields do we expect won't be copied over? Review Sets?

api/src/main/resources/workbench.yaml
```diff
@@ -1100,6 +1100,22 @@ paths:
...
+    post:
+      tags:
+        - cohorts
+      description: Creates a duplicate of the cohort specified by cohortId
+      operationId: "duplicateCohort"
+      responses:
+        200:
+          description: The cohort definition that was created.
+          schema:
+            $ref: "#/definitions/Cohort"
```
blrubenstein: Nit: This can also return a BadRequestException, right? If so we should document.

api/src/test/java/org/pmiops/workbench/cohorts/CohortFactoryTest.java
```diff
@@ -0,0 +1,68 @@
...
+        apiCohort.setType("type");
+        apiCohort.setCriteria("criteria");
+
+        User user = mock(User.class);
+
+        long workspaceId = 1l;
+
+        Cohort dbCohort = cohortFactory.createCohort(apiCohort, user, workspaceId);
+
+        assertThat(dbCohort.getDescription()).isEqualTo("desc");
```
blrubenstein: Nit: I might compare these against apiCohort rather than against the hardcoded string. Just to reduce a bit of brittleness.

ui/src/app/views/resource-card/component.tsx
```diff
@@ -380,6 +398,14 @@ export class ResourceCard extends React.Component<ResourceCardProps, ResourceCar
           <Button onClick={() => this.setState({invalidResourceError: false})}>OK</Button>
         </ModalFooter>
       </Modal>}
+      {this.state.showErrorModal &&
```
blrubenstein: This feels like a useful reusable component to me. Do you think it would be valuable to put this in `components/modals.tsx`?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: (We may also want to do that with the other demos, but that seems out of scope for this work)

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactory.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.pmiops.workbench.db.model.Cohort;
+import org.pmiops.workbench.db.model.User;
+
+public interface CohortFactory {
+
+    Cohort createCohort(org.pmiops.workbench.model.Cohort apiCohort, User creator, long workspaceId);
+
+    // duplicateCohort provides the functionality for the "Duplicate" Cohort function in the AoU workspace.
+    // The name is set to the original cohort's name appended by "_2"
```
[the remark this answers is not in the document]
blrubenstein: I wonder if the desired product behavior here is for it to go `_2`, then `_3`...

api/src/main/java/org/pmiops/workbench/api/CohortsController.java
ericsong: Not really sure unfortunately.

That said, I believe it is an improvement over what was happening previously where all `DataIntegrityViolationException` would get caught as the naming conflict exception message and any other errors get bubbled up to a 500 which is what is happening here, just more explicitly.

I could get rid of the try/catch and it should return 500s in the same conditions but we would lose the exception message.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
ericsong: a `BadRequestException` seems more appropriate than `ServerErrorException`. I'm not sure about removing the exception message altogether though. The changes in this file are just for refactoring so I don't want to make too many functionally visible changes.

ericsong: we could file a ticket if we find repro steps for a bug where this manifests.

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactory.java
ericsong: I believe so. Waiting on confirmation from PM

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactoryImpl.java
```diff
@@ -0,0 +1,62 @@
...
+                apiCohort.getName(),
+                apiCohort.getType(),
+                apiCohort.getCriteria(),
+                creator,
+                workspaceId
+        );
+    }
+
+    @Override
+    public Cohort duplicateCohort(Cohort original, User creator) {
```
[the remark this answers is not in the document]
ericsong: so I read through this a bit and I think it might be tough to reuse since the use cases here and with the "Duplicate" feature are a bit different.

It seems like `cloneCohortAndReviews` is more like deep cloning the entire Cohort object into a new workspace whereas `duplicateCohort` is copying a Cohort (without review sets) within its current current workspace.

The functional differences are...
- the name is changed with `duplicateCohort` and not with `cloneCohort`
- `creator` is copied from the target workspace in `cloneCohort`, passed in as an argument for `duplicateCohort`
-  `lastModifiedTime` is copied from the target workspace in `cloneCohort`, `now()` in `duplicateCohort`
-  `setCreationTime` is copied from the target workspace in `cloneCohort`, `now()` in `duplicateCohort`
- `CohortReview` is not copied over in `duplicateCohort`, no annotations either


That said, if we can settle on a single source for `creator/lastModifiedTime/setCreationTime` when being copied over, I think I can refactor the clone functionality to look something like the following

`cloneCohort(String newName, Cohort original, Workspace targetWorkspace)`
`cloneCohortWithReviewSets(...)`

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactory.java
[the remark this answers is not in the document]
ericsong: hmm, I can change it to be a client specified value. I could see a future where the user is asked to provide the desired name.

api/src/main/java/org/pmiops/workbench/cohorts/CohortFactoryImpl.java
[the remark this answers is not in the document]
calbach: In cloneWorkspace, creator/lastModifiedTime/creationTime actually match the behavior in this PR, it's just coded a bit unintuitively (targetWorkspace is always a freshly created workspace by the cloner), so those changes are not a problem.

Another issue we'd need to deal with is the ID copying in CloneWorkspace - i.e. the current method assumes its being cloned into a fresh workspace (which is not a great separation of concerns). I think these are different enough that we may not need to hit the same code paths right now. That said, if there's any code which would make sense to move into `CohortService` (e.g. `cloneCohort` with a `save()`?) and comments which clearly explain the differences, that would be very useful going forward. That may be what you are suggesting.

One other thing I'll note is that I'm not convinced on the permanence of this decision to not copy cohort reviews. I could easily imagine this become a feature request (e.g. optional checkbox when cloning).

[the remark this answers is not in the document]
ericsong: What IDs are being copied that you are worried about? the annotation copy methods?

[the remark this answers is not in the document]
calbach: Actually ignore that, I was thinking ID was copied but that wouldn't work since it's the primary key. It's just the name that gets copied, which you noted.

api/src/main/resources/workbench.yaml
```diff
@@ -1100,6 +1100,22 @@ paths:
           schema:
             $ref: "#/definitions/Cohort"
 
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/cohorts/{cohortId}/duplicate:
```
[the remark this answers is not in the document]
ericsong: I think I prefer keeping it as Duplicate so it follows the same naming convention as in the UI. It might not be consistent with the other `clone` methods but IMO its easier for a developer to find the right code to change if the code matches the UI.

If we change the language back to clone, its easy enough to change. I can actually change the existing `cloneX` methods to be `duplicateX` if you're OK with it.

ericsong:
@blrubenstein @calbach I think I've addressed everything

ui/src/app/views/text-modal/component.tsx
```diff
@@ -0,0 +1,39 @@
+import {
```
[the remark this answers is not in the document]
blrubenstein: Yeah, that is what we landed on. (Super generic in app/components, and workbench-specific in app/views)

api/src/main/resources/workbench.yaml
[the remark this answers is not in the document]
calbach: Sure feel free to file a ticket + handle in a follow-up PR. My view is that we shouldn't be chasing product/UI changes with our API (it's a moving target), but if you feel strongly about the naming in this case I'm not going to object - as long as we reach consistency.

api/src/main/resources/workbench.yaml
```diff
@@ -1099,6 +1099,38 @@ paths:
...
+      - $ref: '#/parameters/workspaceNamespace'
+      - $ref: '#/parameters/workspaceId'
+    post:
+      tags:
+        - cohorts
+      description: Creates a duplicate of the cohort specified by cohortId
+      operationId: "duplicateCohort"
+      parameters:
+        - in: body
+          name: DuplicateCohortParams
```
calbach: This is an interesting pattern. What does the JSON look like here? The convention we have elsewhere is to define a request object, but I think we should probably switch if this is equivalent (if we want to do this pattern, please also file a ticket on this).

api/src/test/java/org/pmiops/workbench/cohorts/CohortFactoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.sql.Timestamp;
+import java.time.Clock;
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+public class CohortFactoryTest {
```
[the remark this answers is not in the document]
ericsong: I do agree that this test+class is pretty trivial but the reason I wanted it out of `CohortsControllerTest` is because it seemed a bit too heavy and integration like for the kind of tests I needed. Do we want to build out a larger suite of unit tests?

api/src/main/resources/workbench.yaml
ericsong: I changed it to look more like what we have elsewhere. Which I believe is... `[Method]Request` defined in definitions and making a reference to it.

The definition of the object has remained the same.

api/src/test/java/org/pmiops/workbench/cohorts/CohortFactoryTest.java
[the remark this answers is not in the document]
calbach: Our tests in general may need improvement - but testing at the level of the API controller I don't think is too heavy at all, this is the only place in this PR where you can probably fully test the method behavior. I'm not totally sure what you mean by larger test suite - but I don't think we need more unit tests at this lower level.

ericsong:
@calbach ready for another review!

ericsong:
Oh, didn't realize you meant on top of that. I can add tomorrow

ericsong:
@calbach  test added

PR #2010, opened 2019-04-16
[RW-1792][risk=no] Faster API server boot up

ericsong:


README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
...
+      ./project.rb get-test-service-creds
+      ```
+
+##### Usage
+* Start services required for API server
+    * ```Shell
+      ./project-rb start-api-reqs
+      # the counterpart command is ./project.rb stop-api-reqs
+      ```
+* Start API server through gradle
```
calbach: Did you compare this approach to our existing `./project.rb run-api`? What was the time difference?

README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
 ./project.rb
 ```
 
+#### API: Faster API startup for MacOS
+The above steps for starting the API server can take upwards of 8-10 minutes on MacOS, most likely due to performance issues with Docker for Mac. Follow these steps to set up your developer environment to start the API server outside of docker. A full restart should take ~30 seconds with this method.
+
+All commands should be run from `workbench/api`
+
+##### Setup
```
calbach: This implies to me a one-time setup, which is not really accurate. Here are the normal steps that happen in dev-up, and which typically may need to run during development or just when switching branches:

- Bring up services (i.e. equivalent to your `start-api-reqs`)
- Run schema migrations (pickup changes in liquibase XML files)
- Apply configuration updates (pickup changes in config/*_local.json)

README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
...
+
+##### Usage
+* Start services required for API server
+    * ```Shell
+      ./project-rb start-api-reqs
+      # the counterpart command is ./project.rb stop-api-reqs
+      ```
+* Start API server through gradle
+    * ```Shell
+      ./gradlew appengineRun
```
calbach: An alternative approach would be to keep everything else the same, and just switch `run-api` to do this instead.

README.md
ericsong: `./project-rb run-api` consistently takes 3 - 3.5m on my machine

README.md
ericsong: I'm not sure about updating `./project-rb run-api` since I believe it can be confusing if `./project-rb` starts to mix dockerized and non dockerized setups.

So I think it should stay as is or we can just remove it entirely if it is not being used. I'm planning to add some usage metrics around `./project-rb` to track how often developers are using the various commands so we can do some cleanup of older/unused code.

README.md
calbach: Is this resolved? I don't see any mention of how this should be handled when you're using this approach.

README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
...
+All commands should be run from `workbench/api`
+
+##### Setup
+* Go through the [normal docker developer startup process](#api-dev-appengine-appserver) at least once successfully
+* Install Java 8 and set JAVA_HOME. Note: Your Java8 library directory may be different
+    * ```Shell
+      export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_201.jdk/Contents/Home
+      ```
+    * Optional - Add the export command to your `.bash_profile` or `.bashrc` and `source` it 
+* Install envsubst
```
calbach: Is this still needed? Look like you removed its usage below

README.md
calbach: The original intent with ./project.rb was to be the location for developer commands in lieu of the more typical ad-hoc shell scripting. It's not directly related to docker usage, though it may have developed that association over time (there are already commands in there that do not use docker).

Putting all of these things into one command seems a bit easier to me, but if that would cause a significant slowdown then this is fine as-is I guess, e.g.:

```
get-test-service-creds
start-api-reqs
begin:
  run ./gradlew ....
ensure:
  stop-api-reqs
```

api/libproject/generate_appengine_web_xml.sh
```diff
@@ -1,6 +1,29 @@
...
+  echo "Exporting value $var=$evaluatedString"
+  export $var=$evaluatedString
+done
+
 set -euo pipefail
 IFS=$'\n\t'
 
-envsubst < src/main/webapp/WEB-INF/appengine-web.xml.template \
+cat src/main/webapp/WEB-INF/appengine-web.xml.template \
+  | sed "s|\${DB_DRIVER}|${DB_DRIVER}|" \
```
calbach: This is a pretty verbose standin for `envsubst`. This is also not the only place we use envsubst in the codebase. Were you not able to include this dependency in the setup?

api/src/main/resources/workbench.yaml
```diff
@@ -883,6 +883,7 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
+
```
calbach: nit: would revert - unrelated

api/libproject/generate_appengine_web_xml.sh
ericsong: this was added by @gjuggler as `envsubst` is not easy to install on Google macbooks?

gjuggler: envsubst is a huge pita to install on OSX, especially in Google corp environment.

Agree this is verbose, but removes a dependency in the code-reload cycle. Also, the set of env variables substituted here hasn't changed for ~2y, so a bit of verbosity seems OK.

Other usages are outside of the main iteration loop, and are reasonable to be run from within the docker env where it's always available.

calbach: OK, sounds fine to remove the dep. Please at least move the substitution up into the above loop instead so we don't need to hardcode all of these variable names though. I'd just copy the template and run `sed -i` on each turn of the loop. Or is `vars.env` not a superset of these?

gjuggler: Ah, I hadn't thought of that, that would be nice. Though I don't think it's a huge deal to hard-code a few vars that rarely change. (It also makes this file show up more prominently when searching for these env variables, instead of the somewhat obscure dependency via vars.env.)

But probably best to do the cleaner solution w/ less maintenance burden overall. If Eric's willing to code it up :)

README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
...
+    * ```Shell
+      gcloud components install app-engine-java
+      ```
+* Generate Google Cloud access token
+    * ```Shell
+      ./project.rb get-test-service-creds
+      ```
+
+##### Usage
+* If you have schema migrations or pending configuration updates, run through the [normal docker developer startup process](#api-dev-appengine-appserver) at least once successfully. This is typically only necessary when switching between branches.
```
jacmrob: is normal docker developer startup process "./project.rb dev-up" in /api? is there an easier command to run migrations without also starting the api, so i can then start it through this new process?

README.md
```diff
@@ -97,6 +97,43 @@ Other available operations may be discovered by running:
...
+* Start services required for API server
+    * ```Shell
+      ./project-rb start-api-reqs
```
jacmrob: nit - it is `./project.rb` not `./project-rb`

README.md
ericsong: there is definitely something quicker we can run but I haven't worked out what those commands are yet

api/libproject/generate_appengine_web_xml.sh
```diff
@@ -1,6 +1,22 @@
 #!/bin/bash
-set -euo pipefail
```
jacmrob: do we want to keep this command now that you aren't debugging anymore?  probably want this script to fail on error

ericsong: added

PR #2016, opened 2019-04-18
API Server start speed up - take 2

ericsong:
There was a bug with the first revision where the generate xml step would always use the vars from db/vars.env.

We now use the vars in the environment and added steps on how to load the db/vars.env to your local setup to the README

### DIFF
--- a/api/db/load_vars.sh
+++ b/api/db/load_vars.sh
@@ -1,0 +1,11 @@
+for line in $(awk '!/^ *#/ && NF' $WORKBENCH_DIR/api/db/vars.env); do
+  IFS='=' read -r var val <<< "$line"
+
+  if [ "$var" == "DB_HOST" ]; then
+    export DB_HOST=localhost
+    continue
+  fi
+
+  evaluatedString=$(echo $(eval "echo $val"))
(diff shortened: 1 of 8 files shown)


api/libproject/devstart.rb
```diff
@@ -220,6 +250,7 @@ def start_local_api()
   setup_local_environment
   common = Common.new
   common.status "Starting API server..."
+  ENV["OVERWRITE_WORKBENCH_DB_HOST"] = "true"
```
gjuggler: What does this line do? I can't find any references to this string elsewhere.

api/db/load_vars.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
gjuggler: I'd add a line or two of comments here, providing context on what this script is and when / from where it should expect to be run. Just for posterity.

api/libproject/devstart.rb
ericsong: ah, should have been removed. good catch

ericsong:
addressed the comments here: https://github.com/all-of-us/workbench/pulls

PR #2020, opened 2019-04-19
Srubenst/synonyms update

blrubenstein:

### DIFF
--- a/ui/src/app/views/concept-table/component.tsx
+++ b/ui/src/app/views/concept-table/component.tsx
@@ -2,6 +2,7 @@ import {Component, Input} from '@angular/core';
 import {Clickable} from 'app/components/buttons';
 import {CheckBox} from 'app/components/inputs';
 import {PopupTrigger} from 'app/components/popups';
+import colors from 'app/styles/colors';
 import {ReactWrapperBase, toggleIncludes} from 'app/utils';
 import {reactStyles} from 'app/utils';
 import {Concept} from 'generated/fetch/api';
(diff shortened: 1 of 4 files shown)


ui/src/app/views/concept-table/component.tsx
```diff
@@ -17,10 +19,60 @@ const styles = reactStyles({
...
+    const hasOverflowingChildren = element.offsetHeight < element.scrollHeight ||
+      element.offsetWidth < element.scrollWidth;
+    this.setState({willOverflow: hasOverflowingChildren});
+  }
 
-export class ConceptTable extends React.Component<{concepts: Concept[];
-  loading: boolean; placeholderValue: string, getSelectedConcepts: Function,
-  setSelectedConcepts: Function; },
+  render() {
+    return <div style={{display: 'flex'}}>
```
NehaBroad: not: we can extrapolate props and state here.

ui/src/app/views/concept-table/component.tsx
```diff
@@ -17,10 +19,60 @@ const styles = reactStyles({
...
+      <div style={{minWidth: '170px', maxWidth: '170px',
+        fontStyle: 'italic',
```
NehaBroad: Styles with more than 2 attributes can be moved under styles const

ui/src/app/views/concept-table/component.tsx
```diff
@@ -17,10 +19,60 @@ const styles = reactStyles({
   }
 });
 
+export class SynonymsObject extends React.Component<{},
```
NehaBroad: since there is no property we can have const SynonymsObject instead of class

blrubenstein: Resolved offline.

PR #2022, opened 2019-04-22
[RW-2284][risk-no] UI Search Standard Measurement Concepts in a List

dolbeew:
This PR contains a LOT of code additions, but most of that is duplication of existing components needed for the new criteria. So I would recommend focusing on testing the flow of the UI rather than reviewing the actual code. Also, none of the calculate or final counts will work currently but the rest of the UI should be functional.


dolbeew:
Feature flag for the new UI is disabled. Set the `enableCBListSearch` environment variable to true (for local or test environments depending on which api is being used) to test the UI.

PR #2028, opened 2019-04-23
[RW-2457][risk=no] API Workspaces Refactoring

ericsong:
This originally contained a change to the API where a requiredAccessLevel was added to getWorkspaces() but I realized it was unnecessary and we could just do the filtering on the front end. That said, I think there were useful refactoring changes I made that I'd like to keep so here they are.

It includes...
- Creating a Workspaces package
- Moving API mapping code out of WorkspacesController and into a Mapping service
- Setting up WorkspaceService to be unit tested
- Refactor some helper methods in WorkspaceControllerTest
- Add an integration test for getWorkspaces()


api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -0,0 +1,100 @@
...
+    doReturn(workspaceId).when(workspace).getWorkspaceId();
+    WorkspaceResponse workspaceResponse = mock(WorkspaceResponse.class);
+    doReturn(workspace).when(workspaceResponse).getWorkspace();
+    doReturn(accessLevel.toString()).when(workspaceResponse).getAccessLevel();
+    return workspaceResponse;
+  }
+
+  private org.pmiops.workbench.db.model.Workspace mockDbWorkspace(String name, String firecloudUuid) {
+    org.pmiops.workbench.db.model.Workspace workspace = mock(org.pmiops.workbench.db.model.Workspace.class);
+    doReturn(new Timestamp(1000000l)).when(workspace).getLastModifiedTime();
```
blrubenstein: Could this timestamp be pulled out into a variable? (I know other tests use 'NOW'? Is it important to use this constant? If so, why?)

ui/src/app/components/resources.tsx
```diff
@@ -11,11 +11,11 @@ import {Cohort, ConceptSet} from 'generated/fetch';
...
   disabled: boolean, resourceType: ResourceType, onRenameNotebook?: Function,
   onRenameCohort?: Function, onOpenJupyterLabNotebook?: any, onCloneResource?: Function,
-  onDeleteResource?: Function, onEditCohort?: Function, onReviewCohort?: Function,
-  onEditConceptSet?: Function
+  onCopyResource?: Function, onDeleteResource?: Function, onEditCohort?: Function,
+  onReviewCohort?: Function, onEditConceptSet?: Function
 }> = ({
         disabled, resourceType, onRenameNotebook = () => {}, onRenameCohort = () => {},
-        onOpenJupyterLabNotebook = () => {}, onCloneResource = () => {},
+        onOpenJupyterLabNotebook = () => {}, onCloneResource = () => {}, onCopyResource = () => {},
```
blrubenstein: I think you need to add `onCopyResource` to the notebook case below.

ericsong: I left it out here so it wouldn't have any user facing changes. The follow up PR has the rest of the front end code.

PR #2031, opened 2019-04-23
[RW-2456][risk=no] registration homepage refactor

jacmrob:

### DIFF
--- a/ui/src/app/views/homepage/component.tsx
+++ b/ui/src/app/views/homepage/component.tsx
@@ -17,7 +17,7 @@ import {configApi, profileApi} from 'app/services/swagger-fetch-clients';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {QuickTourReact} from 'app/views/quick-tour-modal/component';
 import {RecentWork} from 'app/views/recent-work/component';
-import {WorkbenchAccessTasks} from 'app/views/registration-page/component';
+import {RegistrationPage} from 'app/views/registration-page/component';
 import {
   BillingProjectStatus,
   Profile,
@@ -285,11 +285,11 @@ export const Homepage = withUserProfile()(class extends React.Component<
           <div style={homepageStyles.singleCard}>
             {accessTasksLoaded ?
               (accessTasksRemaining ?
-                (<WorkbenchAccessTasks eraCommonsLinked={eraCommonsLinked}
-                                       eraCommonsError={eraCommonsError}
-                                       trainingCompleted={trainingCompleted}
-                                       firstVisitTraining={firstVisitTraining}
-                                       betaAccessGranted={betaAccessGranted}/>
+                (<RegistrationPage eraCommonsLinked={eraCommonsLinked}
+                                   eraCommonsError={eraCommonsError}
+                                   trainingCompleted={trainingCompleted}
+                                   firstVisitTraining={firstVisitTraining}
+                                   betaAccessGranted={betaAccessGranted}/>
                 ) : (
                   <div style={{display: 'flex', flexDirection: 'row', paddingTop: '2rem'}}>
                     <div style={homepageStyles.contentWrapperLeft}>
--- a/ui/src/app/views/homepage/component.tsx
+++ b/ui/src/app/views/homepage/component.tsx
@@ -17,13 +17,13 @@ import {configApi, profileApi} from 'app/services/swagger-fetch-clients';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {QuickTourReact} from 'app/views/quick-tour-modal/component';
 import {RecentWork} from 'app/views/recent-work/component';
-import {RegistrationPage} from 'app/views/registration-page/component';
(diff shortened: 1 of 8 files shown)


ui/src/app/views/homepage/component.tsx
```diff
@@ -453,15 +285,15 @@ export const Homepage = withUserProfile()(class extends React.Component<
           <div style={homepageStyles.singleCard}>
             {accessTasksLoaded ?
               (accessTasksRemaining ?
-                (<WorkbenchAccessTasks eraCommonsLinked={eraCommonsLinked}
-                                       eraCommonsError={eraCommonsError}
-                                       trainingCompleted={trainingCompleted}
-                                       firstVisitTraining={firstVisitTraining}
-                                       betaAccessGranted={betaAccessGranted}/>
+                (<RegistrationPage eraCommonsLinked={eraCommonsLinked}
```
blrubenstein: Nit: I wouldn't call this `RegistrationPage`, because it isn't actually a top level page? Maybe just `Registration`?

ui/src/app/views/homepage/component.tsx
```diff
@@ -1,201 +1,33 @@
...
 import {RecentWork} from 'app/views/recent-work/component';
+import {RegistrationPage} from 'app/views/registration-page/component';
 import {
   BillingProjectStatus,
   Profile,
 } from 'generated/fetch';
 
-
-const styles = reactStyles({
+export const homepageStyles = reactStyles({
```
blrubenstein: Is there a reason to change this from `styles` to `homepageStyles`? It breaks our convention of page independent styles being called `styles`.

ui/src/app/views/registration-page/component.tsx
```diff
@@ -0,0 +1,183 @@
+import * as React from 'react';
+
+import {AlertClose, AlertDanger, AlertWarning} from 'app/components/alert';
+import {Button} from 'app/components/buttons';
+import {ResourceCardBase} from 'app/components/card';
+import {baseStyles} from 'app/components/card';
```
blrubenstein: Nit: I would leave the variable as `styles` in card, and then do this import as the following:
`import {styles as cardStyles} from 'app/components/card';`

ui/src/app/views/registration-page/component.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <div style={{...styles.mainHeader, fontSize: '18px', marginBottom: '1rem'}}>
+        <ClrIcon shape='warning-standard' class='is-solid'
+                 style={{color: '#fff', marginRight: '0.3rem'}}/>
+        In order to get access to data and tools please complete the following steps:
+      </div>
+      {!betaAccessGranted && <div data-test-id='beta-access-warning'
+                                  style={{...baseStyles.card, ...styles.warningModal}}>
+        <ClrIcon shape='warning-standard' class='is-solid'
+                 style={styles.warningIcon}/>
+        You have not been granted beta access. Please contact help@research-aou.org.
```
blrubenstein: Is this the right email? I thought it was `support@researchallofus.org`?

ui/src/app/views/registration-page/component.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </div>}
+
+      <div style={{display: 'flex', flexDirection: 'row', margin: '3%'}}>
```
blrubenstein: Nit: I think this looked better without the margin, we should check with Lou on that.

ui/src/app/views/registration-page/component.tsx
```diff
@@ -0,0 +1,183 @@
...
+              opacity: '0.6', maxHeight: this.allTasksCompleted() ? '160px' : '305px',
+              minHeight: this.allTasksCompleted() ? '160px' : '305px'}}>
+            <div style={{display: 'flex', flexDirection: 'column', justifyContent: 'flex-start'}}>
+              <div style={styles.cardHeader}>STEP {i + 1}</div>
+              <div style={styles.cardHeader}>{card.title}</div>
+            </div>
+            {!this.allTasksCompleted() &&
+            <div style={styles.cardDescription}>{card.description}</div>}
+            {taskCompletionMap.get(i) ?
+              <Button disabled={true} style={{backgroundColor: '#8BC990', width: 'max-content'}}
```
blrubenstein: If it is disabled, the cursor should be default rather than a pointer.

ui/src/app/views/registration-page/component.tsx
jacmrob: mocks have it without margin so i will make this change

ui/src/app/views/homepage/component.tsx
jacmrob: I'm going to do `RegistrationDashboard`

ui/src/app/views/homepage/component.tsx
jacmrob: 👍

ui/src/app/views/registration-page/component.tsx
jacmrob: pushing back on that since this technically is the base style for a card that then for all our other cards in the workbench have styles on top of them (see card.tsx).  in this case i just wanted to adopt a few common things like border radius and shadows, etc

PR #2034, opened 2019-04-24
[RW-2457][risk=low] Copy notebooks between workspaces

ericsong:
There is a sizable refactoring here with moving Notebooks related code into a NotebooksService as well as refactoring the implementation to be a bit cleaner and reuse more code.

Sorry about the size of the PR, got kind of carried away with moving stuff around. That said, a good chunk of the lines indicated are just file moves. Anything with `Leonardo` in the filename can be assumed to be a file move and ignored.

The code to really focus on is... 
- NotebooksService @calbach @blrubenstein 
  - Even though I pulled it out of WorkspacesController, I made enough changes along the way that it warrants a good look.
  - Also look at how it is being used within the context of WorkspacesController
- Front end code @jacmrob @NehaBroad


ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+export class CopyNotebookModal extends React.Component<CopyNotebookModalProps,
+CopyNotebookModalState> {
+  constructor(props: CopyNotebookModalProps) {
+    super(props);
+
+    this.state = {
+      writeableWorkspaces: [],
+      newName: props.fromNotebook.name,
+      destination: null,
+      state: State.FORM,
```
blrubenstein: Nit: I would name this something about what state its talking about. Is this `copyNotebookResponseState`? In general I prefer verboseness when it gives more clarity.

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+            <ModalBody style={{ textAlign: 'center' }}><Spinner /></ModalBody> :
+            <ModalBody>
+                {this.state.state === State.FORM && this.renderFormBody()}
+                {this.state.state === State.ERROR && this.renderErrorBody()}
+                {this.state.state === State.SUCCESS && this.renderSuccessBody()}
+            </ModalBody>
+        }
+        <ModalFooter>
+          <Button type='secondary' onClick={this.props.onClose}>
+              {this.getCloseButtonText()}
```
blrubenstein: See above on the state variable. I think we can eliminate the `Success` state and then turn this into a ternary operator.

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+          </Button>
+          {this.renderActionButton()}
```
blrubenstein: Likewise, if we eliminate the success state, we can make this a ternary operator.

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+            autoFocus
+            value={this.state.newName}
+            onChange={v => this.setState({ newName: v })}
+        />
+      </div>
+    );
+  }
+
+  renderErrorBody() {
+    return <div style={headerStyles.formLabel}>{this.state.errorMsg}</div>;
```
blrubenstein: I would have this error show up in the form, rather than as a separate modal body implementation. (Spin the button on loading, and if it fails, just have the error message show up on the body.)

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+          Go to Copied Notebook
+        </Button>
+      );
+    }
+  }
+
+  renderFormBody() {
+    return (
+      <div>
+        <div style={headerStyles.formLabel}>Destination *</div>
```
jacmrob: is the `*` to indicate a required field?  i believe that we have a way to use that using `validate.js` that will prevent the user from submitting if required fields aren't filled out.  see `rename-modal` for example

ui/src/app/views/copy-notebook-modal/component.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+describe('CopyNotebookModal', () => {
+  let props: CopyNotebookModalProps;
+
+  const component = () => {
+    return mount<CopyNotebookModal, CopyNotebookModalProps, CopyNotebookModalState>
+    (<CopyNotebookModal {...props}/>);
+  };
+
+  const workspaces = [
```
jacmrob: nice climbing references :)

ui/src/app/views/copy-notebook-modal/component.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+    let select = wrapper.find(Select);
+    select.instance().setState({menuIsOpen: true});
+    wrapper.update();
+
+    const options = wrapper.find(Select).find({role: 'option'}).map(e => e.text());
+
+    expect(options).toEqual([workspaces[0].name, workspaces[2].name]);
+  });
+
+  it('should set destination state upon clicking an option', async() => {
```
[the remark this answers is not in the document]
ericsong: Sure. I  did it this way because I was thinking have more tests enable you to find the root cause of a test failure more quickly but combining them is more future proofed since its testing against its usage rather than the implementation.

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { FileDetail, Workspace } from 'generated/fetch';
+
+import { Spinner } from 'app/components/spinners';
+import { workspacesApi } from 'app/services/swagger-fetch-clients';
+import { WorkspacePermissions } from 'app/utils/workspace-permissions';
+
+enum State { FORM, ERROR, SUCCESS }
+
+export interface CopyNotebookModalProps {
+  fromWorkspaceNamespace: string;
```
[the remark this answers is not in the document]
ericsong: spent some time wrangling with this but couldn't get too far. The problem I ran into was that the output of `getCurrentWorkspace()` is anonymous so I can't import it from other files. I looked at other classes to see how its used but it looks like it is always wrapped in a component that extends ReactWrapperBase. Which seems like its used in the angular transition? I see imports from the angular core library mixed in.

Anyone run into this before?

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+      this.props.fromWorkspaceNamespace,
+      this.props.fromWorkspaceName,
+      this.props.fromNotebook.name,
+      {
+        toWorkspaceName: this.state.destination.id,
+        toWorkspaceNamespace: this.state.destination.namespace,
+        newName: this.state.newName
+      }
+      ).then((response) => {
+        this.setState({ state: State.SUCCESS, loading: false });
```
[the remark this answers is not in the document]
ericsong: I followed the Terra UI for this. I agree that it would simplify the code but I think its a nice UI touch to be able to decide where to go after. Ex. what if you want to copy a bunch of notebooks?

ui/src/app/views/copy-notebook-modal/component.tsx
ericsong: Within the form body? To allow the user to retry?

ui/src/app/views/copy-notebook-modal/component.tsx
```diff
@@ -0,0 +1,185 @@
...
+        this.setState({
+          errorMsg: errorMsg,
+          state: State.ERROR,
+          loading: false
+        });
+      });
+  }
+
+  goToDestinationWorkspace() {
+    window.location.href = this.getWorkspaceUrl(this.state.destination);
```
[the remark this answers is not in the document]
ericsong: I tried that at first but it didn't work so I just did this. I spent some time debugging it but the angular/react routing stuff was getting a bit hard to follow. Happy to give it another shot if someone can guide me!

ui/src/app/views/copy-notebook-modal/component.tsx
ericsong: I took a look at `validate.js` and it seems like a bit of overhead for what I need which is just to check if the user selected anything. I'm not doing any smart validation on what the string needs to look like and I believe thats the value add of using `validate.js`

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -61,9 +62,8 @@
   // If we later decide to tune this value, consider moving to the WorkbenchConfig.
   private static final int MAX_NOTEBOOK_SIZE_MB = 100;
   // "directory" for notebooks, within the workspace cloud storage bucket.
-  private static final String NOTEBOOKS_WORKSPACE_DIRECTORY = "notebooks";
-  private static final Pattern NOTEBOOK_PATTERN =
-      Pattern.compile(NOTEBOOKS_WORKSPACE_DIRECTORY + "/[^/]+(\\.(?i)(ipynb))$");
+  private static final String NOTEBOOKS_WORKSPACE_DIRECTORY = NotebooksServiceImpl.NOTEBOOKS_WORKSPACE_DIRECTORY;
```
calbach: Can you define these constants on the interface instead? Odd to hav ea direct impl reference here

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -617,112 +597,58 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-
-  @Override
-  public ResponseEntity<FileDetail> cloneNotebook(String workspace, String workspaceName,
-      String notebookName) {
-    String newName = notebookName.replaceAll("\\.ipynb", " ") + "Clone.ipynb";
-    FileDetail fileDetail = notebookCloneOperation(workspace, workspaceName, notebookName, newName);
-    return ResponseEntity.ok(fileDetail);
+  public ResponseEntity<List<FileDetail>> getNoteBookList(String workspaceNamespace,
+      String workspaceId) {
+    List<FileDetail> fileList = notebooksService.getNotebooks(workspaceNamespace, workspaceId);
```
calbach: Is there a broader design reason to move all of the logic out of the controller? A lot of the code that was here is pretty tightly married to the details of the API specification. AFAICT  the newly added NotebooksService effectively duplicates the API controller interface, which is already autogenerated for us via Swagger.

I can maybe buy it if this method is being reused from multiple locations or if there is high complexity, but I'm not seeing much of an advantage in these cases. In general I would prioritize test coverage of the actual API method behavior over testing of a service.

api/src/main/java/org/pmiops/workbench/auth/UserProvider.java
```diff
@@ -0,0 +1,18 @@
+package org.pmiops.workbench.auth;
+
+import javax.inject.Provider;
+import org.pmiops.workbench.db.model.User;
+import org.springframework.stereotype.Service;
+
+@Service
```
calbach: Per offline discussion, I would definitely favor our previous approach over this. I think this convention changes the production code path too significantly to justify saving some test boilerplate. I don't see a particularly high cost in this PR to justify it.

I didn't get a chance to try the static stubbing approach per slack - alternatively if you can find a way to make that work, that would fix all the problems here.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -1,126 +1,151 @@
...
-    retryHandler.run((context) -> {
-      clusterApi.deleteCluster(googleProject, clusterName);
-      return null;
-    });
+  public FileDetail copyNotebook(String fromWorkspaceNamespace, String fromWorkspaceName, String fromNotebookName,
+      String toWorkspaceNamespace, String toWorkspaceName, String newNotebookName) {
+    GoogleCloudLocators fromNotebookLocators = getNotebookLocators(fromWorkspaceNamespace, fromWorkspaceName, fromNotebookName);
+    GoogleCloudLocators newNotebookLocators = getNotebookLocators(toWorkspaceNamespace, toWorkspaceName, newNotebookName);
+
+    if (!cloudStorageService.blobsExist(Collections.singletonList(newNotebookLocators.blobId)).isEmpty()) {
```
calbach: This should happen after the access check to avoid unnecessary exposure of information

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -1,126 +1,151 @@
...
+      throw new BlobAlreadyExistsException();
+    }
+    workspaceService.enforceWorkspaceAccessLevel(toWorkspaceNamespace, toWorkspaceName, WorkspaceAccessLevel.WRITER);
+
+    cloudStorageService.copyBlob(fromNotebookLocators.blobId, newNotebookLocators.blobId);
```
calbach: This uses our service account credentials. This needs an access check of READ on the fromWorkspace as well. Currently this would allow me to read the contents of any notebook for which I knew the name and copy it to my own workspace

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -617,112 +597,58 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-    fileDetail.setName(newName);
-    fileDetail.setPath(opDto.fullPath);
-    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
-    fileDetail.setLastModifiedTime(now.getTime());
-    userRecentResourceService
-        .updateNotebookEntry(opDto.workspaceId, opDto.userId, opDto.fullPath, now);
-    return fileDetail;
-  }
+  @Override
+  public ResponseEntity<FileDetail> copyNotebook(String fromWorkspaceNamespace, String fromWorkspaceId,
```
calbach: Shouldn't this be called `duplicateNotebook`? I don't think this should be a different method from clone. If the plan is to remove the other method in a follow-up, that's fine (please add a TODO if so).

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
ericsong: Yes. this is something I feel pretty strongly about. I can prepare a larger presentation on this topic but as a quick rundown... Even if we do not test at the Service level, I think there are a few reasons to break logic out of the controllers.

1. Our controllers are too big which make them hard to navigate and also understand the responsibilities for each.
Ex. whats the difference in responsibilities between WorkspaceController and WorkspaceService? ConceptSetController and ConceptSetService?

2. Encourage layered architecture
For example, we can set the controllers as the layer between our business logic and HTTP and have it handle everything related to HTTP such as request/response parsing, as well as extracting state out of the request so the services can be stateless. (Grabbing User out of Provider<User>)

3. Enable code reuse and exportability
Related to #2: if we create a business logic layer that is purely Java and not tied to Spring, it enables us to extract libraries and jars that can be used in other repositories such as Terra. 
Of course, it also enables our own controllers and services to reuse code with less work.

4. Enable unit testing
I think there is another discussion to be had on whether or not we fully embrace unit testing but I think we should at least we designing our code in a way that it is possible.

My reasoning for pulling NotebookService out of WorkspaceController was that I identified a chunk of logic that was in WorkspaceController but had very little interaction with the rest of the controller. In my mind, WorkspaceController was effectively playing two roles of servicing Workspace related API calls and Notebooks related API alls. I think we should actually take things a step further and create a NotebooksController which handles notebooks related API calls. This work is a step in that direction.

calbach: Thanks - glad to see some strong opinions here. I think it would a useful topic for next week's tech talk if you want to present this and (if possible) outline the before/after of the dependency graph / codebase structure. Seems fine to proceed with the approach in your current PR in the meantime.

Two points that I will add here to consider:
- We control which methods go in which controller. This is defined by the tag in our yaml file. We could decide to instead break up large controllers, e.g. place all the methods you've refactored here into their own controller (the interface would look the same, or perhaps identical to, the NotebooksService). This would also change the client interfaces, so we'd need to think about whether that would actually be desirable overall. My guess is the current distribution of methods could be improved significantly.
- Provider<User> is a request-context-driven injection of data. If your service is making remote calls, e.g. to Leo, it is also going to be reading this context information out, i.e. to authenticate the backend call. If the layer above and below a service are both pulling the User out of the context in this way, having a stateless service between them is less compelling - though I can buy there are still some benefits.

ericsong: Good points! I'd be happy to prepare a tech talk on this.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
ericsong: good catch. added a test case for this

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
ericsong: I can sync up with product on this but in my mind, Copy means an entity can be copied into some other container while Clone/Duplicate means creating a new version within the same container.

api/src/integration/java/org/pmiops/workbench/NotebooksIntegrationTest.java
```diff
@@ -6,8 +6,8 @@
 import org.mockito.Mock;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.notebooks.NotebooksRetryHandler;
-import org.pmiops.workbench.notebooks.NotebooksService;
-import org.pmiops.workbench.notebooks.NotebooksServiceImpl;
+import org.pmiops.workbench.notebooks.LeonoardoNotebooksClient;
```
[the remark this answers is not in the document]
ericsong: haha. good catch

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -85,7 +85,7 @@
         return allOfUsCluster;
       };
 
-  private final NotebooksService notebooksService;
+  private final LeonoardoNotebooksClient leonoardoNotebooksClient;
```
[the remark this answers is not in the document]
ericsong: so the reason I actually made the change is because I wanted to use the name of `NotebooksService` to contain our business logic around notebooks. It seemed like the previous version of `NotebooksService` was mainly acting as a layer between us and leonardo so I renamed it as such.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -280,6 +275,8 @@ public void setUp() {
 
     doNothing().when(cloudStorageService).copyAllDemoNotebooks(any());
 
+    // doAnswer(returnsFirstArg()).when(workspaceDao).save(any(org.pmiops.workbench.db.model.Workspace.class));
```
calbach: rm

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -255,11 +254,7 @@ User user() {
   @Before
   public void setUp() {
     User user = createUser(LOGGED_IN_USER_EMAIL);
-    when(userProvider.get()).thenReturn(user);
-    workspacesController.setUserProvider(userProvider);
-    cohortsController.setUserProvider(userProvider);
-    cohortReviewController.setUserProvider(userProvider);
-    conceptSetsController.setUserProvider(userProvider);
+    currentUser = user;
```
calbach: nit: would just assign directly, `user` var not needed

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1666,9 +1663,137 @@ public void testRenameNotebookWoExtension() throws Exception {
...
+    fromWorkspace = workspacesController.createWorkspace(fromWorkspace).getBody();
+    String fromNotebookName = "origin";
+
+    Workspace toWorkspace = createStubbedWorkspace("toWorkspaceNs", "toworkspace", WorkspaceAccessLevel.OWNER);
+    toWorkspace = workspacesController.createWorkspace(toWorkspace).getBody();
+    String newNotebookName = "new";
+    String expectedNotebookName = newNotebookName + ".ipynb";
+
+    CopyNotebookRequest copyNotebookRequest = new CopyNotebookRequest();
+    copyNotebookRequest.setToWorkspaceName(toWorkspace.getName());
```
calbach: nit: You can chain all swagger models:

```
new CopyNotebookRequest()
  .toWorkspaceName(...)
  .toWorkspaceNamespace(...)
```

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -1666,9 +1663,137 @@ public void testRenameNotebookWoExtension() throws Exception {
...
+    workspacesController.copyNotebook(
+        fromWorkspace.getNamespace(),
+        fromWorkspace.getName(),
+        fromNotebookName,
+        copyNotebookRequest);
+  }
+
+  @Test(expected = ForbiddenException.class)
+  public void copyNotebook_hasReadPermissions() {
+    Workspace fromWorkspace = createStubbedWorkspace("fromWorkspaceNs", "fromworkspae", WorkspaceAccessLevel.NO_ACCESS);
```
calbach: opt_nit: typo in "fromworkspae", not that it matters

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
calbach: I think this is entirely an engineering call, product will probably not be looking at our REST APIs. I think that functionally, duplicating within a workspace is a proper subset of duplicating across workspaces, with no other semantic distinction. Product can call these buttons two different things if they want, we don't need our API to mirror that.

ericsong: True. Are you saying we should just remove `cloneNotebook` and use `copyNotebook` in its place?

calbach: Yep - could also be in a follow-up PR if you prefer.

ericsong: I'll do it in a follow up. It's going to be hard to pick out the changes in this PR since its already so big. Also there's a good chunk of changes that need to be made across the stack.

https://precisionmedicineinitiative.atlassian.net/browse/RW-2551

ui/src/app/views/copy-notebook-modal/component.tsx
blrubenstein: Yes, within the form body.

ui/src/app/views/copy-notebook-modal/component.spec.tsx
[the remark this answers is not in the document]
blrubenstein: Since test failures give line number, I am not too worried about combining them. I don't think its complex enough that we as a team are going to have too much trouble finding root cause.

ui/src/app/views/copy-notebook-modal/component.tsx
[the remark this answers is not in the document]
blrubenstein: Sure! basically navigate takes an array of paths. Let me take a quick look and see if I can give you the method.

ui/src/app/views/copy-notebook-modal/component.tsx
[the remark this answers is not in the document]
blrubenstein: I think that is correct in this case. In that case, this seems reasonable and we can pass it down this way from the parent.

ui/src/app/views/copy-notebook-modal/component.tsx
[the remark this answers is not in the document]
blrubenstein: It should just be `navigate(['workspaces', workspace.namespace, workspace.id, 'notebooks'])` to do this. If that doesn't work, do you want to try hopping on a screen share to figure it out?

PR #2035, opened 2019-04-24
[RW-2460][risk=no] set up 2FA module backend & cron job to sync from google

jacmrob:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -270,11 +270,8 @@ public class ProfileController implements ProfileApiDelegate {
       return user;
     }
 
-    if (user.getTwoFactorAuthCompletionTime() == null &&
-            directoryService.getUser(user.getEmail()).getIsEnrolledIn2Sv()) {
-      user.setTwoFactorAuthCompletionTime(now);
-      user = saveUserWithConflictHandling(user);
-    }
+    // Check if 2FA is set up on account
+    userService.syncTwoFactorAuthStatus(user);
 
     // On first sign-in, create a FC user, billing project, and set the first sign in time.
     if (user.getFirstSignInTime() == null) {
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -270,11 +270,10 @@ public class ProfileController implements ProfileApiDelegate {
       return user;
     }
 
-    if (user.getTwoFactorAuthCompletionTime() == null) {
-      if (directoryService.getUser(user.getEmail()).getIsEnrolledIn2Sv()) {
-        user.setTwoFactorAuthCompletionTime(now);
-        user = saveUserWithConflictHandling(user);
-      }
+    if (user.getTwoFactorAuthCompletionTime() == null &&
+            directoryService.getUser(user.getEmail()).getIsEnrolledIn2Sv()) {
+      user.setTwoFactorAuthCompletionTime(now);
+      user = saveUserWithConflictHandling(user);
     }
 
     // On first sign-in, create a FC user, billing project, and set the first sign in time.
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -502,6 +502,7 @@ public class ProfileController implements ProfileApiDelegate {
     return getProfileResponse(userProvider.get());
   }
 
+  @Override
   public ResponseEntity<Profile> syncTwoFactorAuthStatus() {
     userService.syncTwoFactorAuthStatus();
     return getProfileResponse(userProvider.get());
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -264,7 +264,6 @@ public class ProfileController implements ProfileApiDelegate {
   private User initializeUserIfNeeded() {
     UserAuthentication userAuthentication = userAuthenticationProvider.get();
     User user = userAuthentication.getUser();
-    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     if (userAuthentication.getUserType() == UserType.SERVICE_ACCOUNT) {
       // Service accounts don't need further initialization.
       return user;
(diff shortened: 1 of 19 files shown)


api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -167,4 +178,13 @@ public void testClearsEraCommonsStatus() throws Exception {
     assertThat(user.getEraCommonsCompletionTime()).isNull();
   }
 
+  @Test
+  public void testSyncTwoFactorAuthStatus() throws Exception {
+    when(directoryService.getUser(EMAIL_ADDRESS)).thenReturn(googleUser);
```
ericsong: it looks like `googleUser` is only used in this test. Should we move narrow down the scope to this function since it is the only one that needs it?

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -167,4 +178,13 @@ public void testClearsEraCommonsStatus() throws Exception {
     assertThat(user.getEraCommonsCompletionTime()).isNull();
   }
 
+  @Test
+  public void testSyncTwoFactorAuthStatus() throws Exception {
```
ericsong: i think the other case warrants a test case too. should be a quick add.

(setting 2fa completion time to false)

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -264,15 +264,17 @@ private String createFirecloudBillingProject(User user) {
...
     if (userAuthentication.getUserType() == UserType.SERVICE_ACCOUNT) {
       // Service accounts don't need further initialization.
       return user;
     }
 
-    Boolean twoFactorEnabled = Optional.ofNullable(user.getTwoFactorEnabled()).orElse(false);
-    if (!twoFactorEnabled) {
-      user.setTwoFactorEnabled(directoryService.getUser(user.getEmail()).getIsEnrolledIn2Sv());
-      user = saveUserWithConflictHandling(user);
+    if (user.getTwoFactorAuthCompletionTime() == null) {
```
ericsong: nit: we can make this slightly more readable by adding a `completedTwoFactorAuth()` function to the User class which does this check

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -485,4 +489,24 @@ public void syncEraCommonsStatusUsingImpersonation(User user) throws IOException
       }
     }
   }
+
+  /**
+   * Syncs the 2FA status for the current user.
+   */
+  public void syncTwoFactorAuthStatus() {
+    this.syncTwoFactorAuthStatus(userProvider.get());
```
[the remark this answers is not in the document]
ericsong: optional: I would say just exposing the parameterized function is sufficient for the UserService's API. The reason why I prefer that is because I think Services (or any code really) should be as stateless as possible so it should try not to import and use stateful dependencies like Provider<User>. Instead, any state related to the request should be captured in the Controllers (which are actually aware of how to interact with HTTP) and passed to the services.

That said, I don't think this is a pattern we have adopted so it's up to you if you want to make the change. Not a big deal either way in this case.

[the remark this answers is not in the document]
jacmrob: i'm gonna leave it as-is because we do use this convention in other places and there is a case for when this function is called by the current logged-in user

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jacmrob: i'm going to argue that that makes this less readable because i have to go click around more, and also adds more boilerplate code

ericsong: That's fine. But in regards to clicking around, the goal for me is that I shouldn't have to. I don't really care how its implemented, I just want to know if its enabled or not which I should be able to glean from the function name. I had to read through about half the PR before I understood the purpose of the 2fa completion time, and then was able to come back here to reread it and understand whats going on.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -485,4 +489,23 @@ public void syncEraCommonsStatusUsingImpersonation(User user) throws IOException
...
+
+  public void syncTwoFactorAuthStatus() {
+    syncTwoFactorAuthStatus(userProvider.get());
+  }
+
+  public void syncTwoFactorAuthStatus(User targetUser) {
+    updateUserWithRetries(user -> {
+      Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+      if (directoryService.getUser(user.getEmail()).getIsEnrolledIn2Sv()) {
+        user.setTwoFactorAuthCompletionTime(now);
```
ericsong: I believe this will overwrite the twoFactorAuthCompletionTime if it called on a user that already has it set. Is that what we want? This is different from what is happening in ProfileController. ProfileController only sets it if the value is originally null.

Side note: this is another reason for sharing the code here and in ProfileController. Less likelihood of differing implementation and bugs.

jacmrob: good catch. will do.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -485,4 +489,23 @@ public void syncEraCommonsStatusUsingImpersonation(User user) throws IOException
...
     }
   }
+
+  public void syncTwoFactorAuthStatus() {
+    syncTwoFactorAuthStatus(userProvider.get());
+  }
+
+  public void syncTwoFactorAuthStatus(User targetUser) {
+    updateUserWithRetries(user -> {
+      if (user.getTwoFactorAuthCompletionTime() == null &&
```
ericsong: i think this actually unsets the user's 2fa completion time if it set to anything. `user.getTwoFactorAuthCompletionTime() = something` -> `user.getTwoFactorAuthCompletionTime() == null` is false -> flows into else statement.

might be good to write a test case to cover this as well as the previous condition. its not a lot of code its covering but boolean statements like this can be tricky so I think its good to have the edge cases covered.

jacmrob: thanks for checking.... so here's what i think the logic should be -- since we need to be able to unset the var on our side if the user disables 2FA on the google side (and this method gets called by our cron endpoint that checks that daily to keep them in sync), it should check if the user is enrolled in google, then if the value is null in workbench, set it to a new timestamp.  otherwise (if user is unenrolled in google), set to null in workbench.

PR #2036, opened 2019-04-25
[RW-2342]	 API: Create API for Preview Data Table

NehaBroad:


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -187,11 +132,11 @@ public DataSet apply(org.pmiops.workbench.db.model.DataSet dataSet) {
           result.setName(dataSet.getName());
           Iterable<org.pmiops.workbench.db.model.ConceptSet> conceptSets =
               conceptSetDao.findAll(dataSet.getConceptSetId());
-          result.setConceptSets(Streams.stream(conceptSets)
+          result.setConceptSets(StreamSupport.stream(conceptSets.spliterator(), false)
```
blrubenstein: I couldn't get the streams to work here, it kept giving me import errors, but maybe I was missing something. @NehaBroad when you get in tomorrow, mind taking a look?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -229,125 +174,49 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+    Map<String, QueryJobConfiguration> bigQueryJobConfig = dataSetService.generateQuery(dataSet);
+    bigQueryJobConfig.forEach((domain, queryJobConfiguration) -> {
+      String query = queryJobConfiguration.getQuery().concat(" LIMIT "+ NO_OF_PREIVEW_ROWS);
+      queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
+
+      TableResult queryResponse = bigQueryService.executeQuery(bigQueryService
+          .filterBigQueryConfig(queryJobConfiguration));
+
+      // Construct a list of all values with empty results.
+      List<DataSetPreviewValueList> valuePreviewList =
```
blrubenstein: I had to change the logic around here because it wasn't working with multiple domains. @NehaBroad please review to make sure it looks good to you!

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -320,7 +329,32 @@ private void stubGetWorkspace(String ns, String name, String creator,
...
+      schemaFields.add(Field.of("JOIN_VALUE", LegacySQLTypeName.STRING));
+      FieldList schema = FieldList.of(schemaFields);
+      ArrayList<FieldValue> rows = new ArrayList<>();
+      rows.add(FieldValue.of(FieldValue.Attribute.PRIMITIVE, "PERSON_ID"));
+      rows.add(FieldValue.of(FieldValue.Attribute.PRIMITIVE, domainBaseTable));
+      FieldValueList fieldValueList = FieldValueList.of(rows, schema);
+      values.add(fieldValueList);
+    });
+    doReturn(values).when(tableResultMock).getValues();
+    doReturn(tableResultMock).when(bigQueryService).executeQuery(any());
```
blrubenstein: @calbach FYI, I figured out how to mock the execute query call. Of note, this does _not_ handle constructing the actual query job param, so this will hit everything :( but I think that is doable. (Just not something I wanted to tackle here until we need to.)

PR #2047, opened 2019-04-29
[RW-2489][risk=no] Use original Workspace Name on Clone/Duplicate Workspace page

jmthibault79:
Use the `props` Workspace Name instead of the `state` Workspace Name in the header text, making it equal to the _fixed_ original name instead of the _editable_ cloned name.

Also fix a typo


jmthibault79:
Old behavior: Screen Shot 2019-04-29 at 4 45 00 PM

New behavior: 
Screen Shot 2019-04-29 at 4 43 51 PM

PR #2048, opened 2019-04-30
[RW-2553][risk=no] Do not log potentially sensitive GAE XML configs

calbach:


calbach:
Force submitting to-be-reviewed since there's some urgency on this and no one around to review. Please review after.

ericsong:
Looks good

PR #2049, opened 2019-04-30
[RW-2553][risk=no] Drop gradle --info logging

calbach:
Potentially logs secrets from system vars


calbach:
Force merging, please review post-merge.

PR #2050, opened 2019-04-30
[risk=no][RW-2453] Add registration tasks to user profile page

ericsong:
Screen Shot 2019-04-30 at 1 27 26 PM

### DIFF
--- a/ui/src/app/views/profile-page/component.tsx
+++ b/ui/src/app/views/profile-page/component.tsx
@@ -13,7 +13,7 @@ import {profileApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
 import {reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import { ProfileRegistrationStepStatus } from 'app/views/profile-registration-step-status/component';
-import { RegistrationTasks } from 'app/views/registration-dashboard/component';
+import { RegistrationTasksMap } from 'app/views/registration-dashboard/component';
 import {environment} from 'environments/environment';
 import {Profile} from 'generated/fetch';
 
@@ -61,23 +61,6 @@ const validators = {
   organization: {...required, ...notTooLong(255)},
   areaOfResearch: required,
 };
-
-function isTwoFactorAuthEnabled(profile: Profile): boolean {
-  return !!profile.twoFactorAuthCompletionTime || !!profile.twoFactorAuthBypassTime;
-}
-
-function isTrainingComplete(profile: Profile): boolean {
-  return !!profile.complianceTrainingCompletionTime || !!profile.complianceTrainingBypassTime;
-}
-
-function isEraCommonsLinked(profile: Profile): boolean {
-  return !!profile.eraCommonsCompletionTime || !!profile.eraCommonsBypassTime;
-}
-
-function isDataUseAgreementSigned(profile: Profile): boolean {
-  return !!profile.dataUseAgreementCompletionTime || !!profile.dataUseAgreementBypassTime;
-}
-
 
 export const ProfilePage = withUserProfile()(class extends React.Component<
   { profileState: { profile: Profile, reload: Function } },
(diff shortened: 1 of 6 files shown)


ui/src/app/views/profile-registration-step-status/component.tsx
```diff
@@ -0,0 +1,98 @@
...
+interface Props {
+  title: string;
+  wasBypassed: boolean;
+  isComplete: boolean;
+  incompleteButtonText: string;
+  completedButtonText: string;
+  completeStep: Function;
+}
+
+class ProfileRegistrationStepStatus extends React.Component<Props> {
```
jacmrob: if this component doesn't have state, you could make it a React.FunctionComponent (see the `WorkspaceCardMenu` in workspace-list for example).  Then I would move this code into the profile component file, since it's not really a separate "view" and is unique to the profile

ui/src/app/views/profile-registration-step-status/component.tsx
```diff
@@ -0,0 +1,98 @@
+import * as React from 'react';
+
+import { Button } from 'app/components/buttons';
+import { ClrIcon } from 'app/components/icons';
+import colors from 'app/styles/colors';
+import { reactStyles } from 'app/utils';
+
+const styles = reactStyles({
+  container: {
```
jacmrob: following the existing convention, can you put some of these styles on the same line to reduce the number of lines here?

ui/src/app/views/profile-registration-step-status/component.tsx
```diff
@@ -0,0 +1,98 @@
...
+  button: {
+    height: 38
+  },
+  successButton: {
+    backgroundColor: '#8BC990',
+    cursor: 'default'
+  }
+});
+
+interface Props {
```
jacmrob: nit: would give this a more specific name

ui/src/app/views/profile-registration-step-status/component.tsx
```diff
@@ -0,0 +1,98 @@
...
+  completedButtonText: string;
+  completeStep: Function;
+}
+
+class ProfileRegistrationStepStatus extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+  }
+
+  render() {
```
jacmrob: also following convention, you can extrapolate props at the top of this function (`const {isComplete, ...} = this.props`) and then not have to keep calling "this.props.*" below

ui/src/app/views/profile-registration-step-status/component.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <div style={styles.detailsContainer}>
+            { this.props.children }
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+export {
```
jacmrob: you can put the export at the start of the class definition, although as per my above comment i think you should move this into the profile component file

ui/src/app/views/registration-dashboard/component.tsx
```diff
@@ -48,20 +48,68 @@ const styles = reactStyles({
...
+  twoFactorAuthCompleted: boolean;
+  dataUseAgreementCompleted: boolean;
 }
 
-export class RegistrationDashboard extends
-    React.Component<RegistrationDashboardProps,
-    {trainingWarningOpen: boolean, showRefreshButton: boolean,
-      taskCompletionMap: Map<number, boolean>}> {
+interface State {
+  showRefreshButton: boolean;
```
jacmrob: don't think this needs to be defined above.  adds more boilerplate imo

ui/src/app/views/profile-page/component.tsx
```diff
@@ -58,6 +62,23 @@ const validators = {
   areaOfResearch: required,
 };
 
+function isTwoFactorAuthEnabled(profile: Profile): boolean {
```
jacmrob: Can we reuse these functions in the homepage when we check these things for the registration dashboard?   see here and below: https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/homepage/component.tsx#L191

ui/src/app/views/profile-registration-step-status/component.tsx
ericsong: I made the `body` css inline but I kept the rest for the following reasons

container, title - I think its easier to read multiple lines of CSS if they are line by line versus inline.

buttonsContainer, detailsContainer, successButton - i think the class name adds some context on what its being used for. Makes it a quicker read IMO, instead of having to load the UI or read the code to figure out what this div is being used for.

button - its being used twice. Although, it would be nice if I could just apply a style to a tag or children of a tag... Is there a way to do that using `reactStyles`?

ui/src/app/views/profile-registration-step-status/component.tsx
ericsong: so this is a stylistic question I was going to ask the team. When we define State/Props interfaces, should we be prefixing the whole name of the class?

I think for State, it's unnecessary to do so since it's internal to the class. For example, we do not define the css variable above as `profileRegistarationStepStatusStyles`.

For Props, there is an argument that it should be named since it is exposed outside of the class but I think we can define that in the export statement like I did below. That said, the only gain here is that we just save on a little bit of typing but I did it this way since I don't see much of a downside.

ui/src/app/views/profile-registration-step-status/component.tsx
ericsong: Sounds good. I made the change to FunctionComponent. However, I think I prefer it in the separate file since it's easier to read shorter files.

aourwmob: Discussed and decided that separate file is fine.

ui/src/app/views/profile-registration-step-status/component.tsx
aourwmob: Discussed and decided separate lines is clear here.

ui/src/app/views/profile-registration-step-status/component.tsx
aourwmob: Discussed. We like the consistency between this generic `Props` and `styles` above. Decided to make this the default practice and may refactor other code to be consistent with this.

ui/src/app/views/registration-dashboard/component.tsx
aourwmob: Discussion involving `Props` applies.

ui/src/app/views/profile-page/component.tsx
aourwmob: 👍

PR #2051, opened 2019-05-01
[risk=no][RW-2559][P1] enforce data use agreement in setting data access level

jacmrob:


jacmrob:
This will make it so that completing the DUA is actually a requirement for getting access to the workbench.  right now it's just a smokeshow in the registration dash

PR #2052, opened 2019-05-01
[RW-2558][risk=no][P1] Allow users to request review on workspaces

jacmrob:


jacmrob:
also fixes some spacing on modal text/button text (because jsx is weird about putting spaces between interpolations)

jacmrob:
`v.value` ended up unsetting the whole state variable.  like it full on disappeared.

PR #2053, opened 2019-05-01
[risk=no] Adds program to generate API map.

bopwerks:


tools/api.py
```diff
@@ -0,0 +1,414 @@
+"""Produce a list of API endpoints that are dialed by AoU endpoints.
```
blrubenstein: I would rather go for a more descriptive file name than `api`? Maybe `generate-api-endpoint-mappings.py` or something?

tools/api.py
```diff
@@ -0,0 +1,414 @@
...
+To generate the call graph, do the following:
+
+```
+git clone https://github.com/gousiosg/java-callgraph.git
+cd path/to/java-callgraph
+mvn install # install 'maven' before running this command
+java -jar target/javacg-0.1-SNAPSHOT-static.jar path-to-aou-repo/{aou-utils/swagger-codegen-cli.jar,common-api/build/libs/common-api-0.1.0.jar} >path/to/callgraph.txt
+```
+
+Then, to produce the list of endpoints, run `python3 api.py path/to/aou-repo path/to/callgraph.txt`. The endpoints are output in CSV.
```
blrubenstein: Nice documentation on how to use this. Love it.

tools/api.py
```diff
@@ -0,0 +1,414 @@
...
+        return EXIT_FAILURE
+    project_root = argv[1]
+    callgraph_path = argv[2]
+    apis = [
+        API("Workbench", os.path.join(project_root, "api/src/main/resources/workbench.yaml"), "org.pmiops.workbench.api"),
+        API("Terra", os.path.join(project_root, "api/src/main/resources/fireCloud.yaml"), "org.pmiops.workbench.firecloud.api"),
+        API("Notebooks", os.path.join(project_root, "api/src/main/resources/notebooks.yaml"), "org.pmiops.workbench.notebooks.api"),
+        API("JIRA", os.path.join(project_root, "api/src/main/resources/jira.yaml"), "org.pmiops.workbench.jira.api"),
+        API("Mandrill", os.path.join(project_root, "api/src/main/resources/mandrill_api.yaml"), "org.pmiops.workbench.mandrill.api"),
+        API("Moodle", os.path.join(project_root, "api/src/main/resources/moodle.yaml"), "org.pmiops.workbench.moodle.api"),
```
blrubenstein: Do we need to manually add in APIs when we add them here? If we do, what happens if an API isn't included here? Is it just ignored?

This is something I could see us needing to edit, probably more than anything else in the file. Can we make it easily discoverable? (Either by moving to a variable at the top of the file, or throwing some sort of obvious comment?)

tools/api.py
bopwerks: Good idea, will rename.

tools/api.py
bopwerks: Thanks!

tools/api.py
bopwerks: The program will crash if a yaml file is unreadable. Your idea is good -- I'll move this list as near the top of the file as I can, and print a warning if a file is unreadable. If we find that this tool is useful for other repositories then I think it would be wise to soft-code the yaml file paths.

PR #2054, opened 2019-05-01
[RW-2441][risk=moderate] Only init Zendesk Widget when user is signed in

jmthibault79:
Potential security risk: imports external executable code (no change from current)

### DIFF
--- a/ui/src/app/utils/zendesk.ts
+++ b/ui/src/app/utils/zendesk.ts
@@ -1,5 +1,9 @@
 import {ElementRef} from '@angular/core';
 
+// ideally this initialization would be done by including <script> tags in
+// the component template html file.  However, Angular does not allow this.
+// TODO: investigate whether this will be possible after we migrate to React
+
 export function initializeZendeskWidget(
   elementRef: ElementRef, accessKey: string): void {
   // This external script loads the Zendesk web widget, connected to our
--- a/ui/src/app/utils/zendesk.ts
+++ b/ui/src/app/utils/zendesk.ts
@@ -1,3 +1,49 @@
+import {ElementRef} from '@angular/core';
+
+export function initializeZendeskWidget(
+  elementRef: ElementRef, accessKey: string): void {
+  // This external script loads the Zendesk web widget, connected to our
+  // production Zendesk account. If we ever get a test Zendesk account, the key
+  // should be templated in using an environment variable.
+  const s = document.createElement('script');
+  s.type = 'text/javascript';
+  s.id = 'ze-snippet';
+  s.src = 'https://static.zdassets.com/ekr/snippet.js?key=' + accessKey;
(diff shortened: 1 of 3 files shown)


ui/src/app/utils/zendesk.ts
```diff
@@ -1,3 +1,49 @@
...
+  const s = document.createElement('script');
+  s.type = 'text/javascript';
+  s.id = 'ze-snippet';
+  s.src = 'https://static.zdassets.com/ekr/snippet.js?key=' + accessKey;
+  elementRef.nativeElement.appendChild(s);
+
+  // This data configures the Zendesk web widget with settings to show only
+  // the "contact us" form. See https://developer.zendesk.com/embeddables/docs/widget/
+  // for API docs.
+  window['zESettings'] = {
```
jmthibault79: This is purely cut/paste from index.html

jmthibault79:
The JIRA ticket suggests moving the `<script>` calls to `signed-in/component.html`.  This was not possible however, because Angular ignores script tags in templates.  I used the solution given in https://stackoverflow.com/questions/38088996/adding-script-tags-in-angular-component-template instead.

ui/src/app/utils/zendesk.ts
```diff
@@ -1,3 +1,49 @@
...
+export function initializeZendeskWidget(
+  elementRef: ElementRef, accessKey: string): void {
+  // This external script loads the Zendesk web widget, connected to our
+  // production Zendesk account. If we ever get a test Zendesk account, the key
+  // should be templated in using an environment variable.
+  const s = document.createElement('script');
+  s.type = 'text/javascript';
+  s.id = 'ze-snippet';
+  s.src = 'https://static.zdassets.com/ekr/snippet.js?key=' + accessKey;
+  elementRef.nativeElement.appendChild(s);
```
jmthibault79: This is intended to be equivalent to
```
<script id="ze-snippet" src="https://static.zdassets.com/ekr/snippet.js?key=5a7d70b9-37f9-443b-8d0e-c3bd3c2a55e3"> </script>
```

blrubenstein:
Taking a look now. (Also, FYI @ericsong I use they/them pronouns)

ui/src/app/utils/zendesk.ts
```diff
@@ -1,3 +1,49 @@
...
+
+  // This data configures the Zendesk web widget with settings to show only
```
blrubenstein: Nit: I'd throw a comment on here about why it needs to be done this way.

ui/src/app/utils/zendesk.ts
```diff
@@ -1,3 +1,53 @@
+import {ElementRef} from '@angular/core';
+
+// ideally this initialization would be done by including <script> tags in
```
jmthibault79: @blrubenstein how is this?

PR #2055, opened 2019-05-01
[risk=no][RW-2261][RW-2492] change all frontend 'clone' language to 'duplicate'

jacmrob:

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -102,7 +102,7 @@ const routes: Routes = [
                 component: WorkspaceEditComponent,
                 data: {
                   title: 'Clone Workspace',
-                  mode: WorkspaceEditMode.Clone,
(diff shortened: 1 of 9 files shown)


ui/src/app/app-routing.module.ts
```diff
@@ -102,7 +102,7 @@ const routes: Routes = [
                 component: WorkspaceEditComponent,
                 data: {
                   title: 'Clone Workspace',
```
jmthibault79: Update title

jacmrob: Good catch

jmthibault79: Q: should we also update the `clone` path?

jacmrob:
This also includes changing API endpoints and routes, however the backend methods the correspond to are currently still using "clone"

PR #2057, opened 2019-05-02
[RW-2545] [risk=no] Adds performance directory for perf tests.

bopwerks:


performance/src/gatling/simulations/org/pmiops/databrowser/BasicSimulation.scala
```diff
@@ -0,0 +1,29 @@
+package org.pmiops.databrowser
```
gjuggler: I'm a little conflicted on whether we should check in the actual databrowser code files. We'll almost certainly delete this in a future PR, so my initial inclination would be to keep them out of this PR and just bring in the adapted code we'll actually use, once we have at least an MVP that runs against the workbench.

Your call though – if you think it's important to have this tracked in our own version control, I won't block it.

bopwerks: Sounds good. I've removed the scala files for now.

PR #2059, opened 2019-05-02
[RW-2145][RISK: LOW] Create Preview Data Table

NehaBroad:
Screen Shot 2019-05-02 at 1 29 20 PM


ui/src/app/views/dataset-page/component.tsx
```diff
@@ -38,8 +38,9 @@ import {
   ValueSet,
   WorkspaceAccessLevel,
 } from 'generated/fetch';
-
-
+import {DataSetPreviewList} from 'generated/model/dataSetPreviewList';
```
blrubenstein: I'd make this line import from fetch:
`import {DataSetPreviewList} from 'generated/fetch'`

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -89,15 +90,33 @@ export const ValueListItem: React.FunctionComponent <
...
-    confirmDeleting: boolean, editing: boolean, resource: RecentResource,
-    rType: ResourceType, selectedConceptSetIds: number[], selectedCohortIds: number[],
-    valueSets: ValueSet[], selectedValues: DomainValuePair[], openSaveModal: boolean,
-    queries: Array<DataSetQuery>, includesAllParticipants: boolean
-  }> {
+interface Props {
+  workspace: WorkspaceData;
+}
+
+interface State {
```
blrubenstein: Becoming more and more sold on the idea of interfaces. This is way more readable.

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -311,6 +351,28 @@ export const DataSetPage = withCurrentWorkspace()(class extends React.Component<
     };
     const sqlQueries = await dataSetApi().generateQuery(namespace, id, dataSet);
     this.setState({queries: sqlQueries.queryList});
+    this.setState({dataSets: []});
```
blrubenstein: I would fold this in to the setState call before.

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -89,15 +90,33 @@ export const ValueListItem: React.FunctionComponent <
...
+  workspace: WorkspaceData;
+}
+
+interface State {
+  confirmDeleting: boolean;
+  cohortList: Cohort[];
+  conceptDomainList: DomainInfo[];
+  conceptSetList: ConceptSet[];
+  creatingConceptSet: boolean;
+  dataSets: Array<DataSetPreviewList>;
```
blrubenstein: Nit: Can we call this `dataSetTablePreviews` or something? I just picture this as multiple data sets as currently written.

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -455,24 +534,54 @@ export const DataSetPage = withCurrentWorkspace()(class extends React.Component<
...
                     onClick ={() => this.setState({openSaveModal: true})}
                     disabled={this.disableSave()}>
-              SAVE DATASET
+              SAVE DATA SET
             </Button>
           </div>
-          {/*TODO: Display dataset preview*/}
+          {previewDataLoading && <div style={{display: 'flex', flexDirection: 'column'}}>
+            <Spinner style={{position: 'relative', top: '2rem',
+              left: '20rem'}}></Spinner>
```
blrubenstein: Rather than using a left property, can you center it and get rid of the position relative?

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -117,6 +136,9 @@ export const DataSetPage = withCurrentWorkspace()(class extends React.Component<
       selectedValues: [],
       openSaveModal: false,
       queries: [],
+      dataSets: [],
+      selectedPreviewDomain: '',
```
blrubenstein: I would probably try to default this to null, and type it as a `Domain`? I think it makes more sense as a typing than a free-form string.

NehaBroad: this is a string the data coming back as domain as a string value

NehaBroad:
Test will be part of RW-2397: Test for select values

PR #2063, opened 2019-05-03
[RW-2467][risk=no] Billing Project Buffer - Create entries

ericsong:
First part of the Billing Project Buffer project. This PR adds the liquibase migrations for the table that will represent our buffer as well as start running a cron which will spin up billing projects until the buffer is full.

High Level Summary - Exposing a cron job endpoint called bufferBillingProject() which will check if the buffer if the full and create a buffered billing project if it is not. Every call can only create one and it is currently configured to be called every 5 seconds. The buffer max capacity is set in the WorkbenchConfig while the current buffer size is determined by the number of buffer entries that have the qualifying statuses.

Some questions
- How can I actually if a cron is working locally? I'm not too sure if my cron schedule string is correct
- How does WorkbenchConfig get updated? Changing it via SQL didn't get the Provider to automatically reload the values. I had to do a server restart for new config values to get picked up.


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -0,0 +1,62 @@
...
+    BillingProjectBufferEntry entry = new BillingProjectBufferEntry();
+    entry.setProjectName(projectName);
+    entry.setCreationTime(new Timestamp(clock.instant().toEpochMilli()));
+    entry.setStatus(CREATING);
+
+    billingProjectBufferEntryDao.save(entry);
+  }
+
+  private String createBillingProjectName() {
+    return workbenchConfigProvider.get().firecloud.billingProjectPrefix + "-" + UUID.randomUUID().toString();
```
calbach: Did we want to make the UUID shorter? i.e. substring this?

api/db/changelog/db.changelog-73-add-BillingProjectBufferEntry-table.xml
```diff
@@ -0,0 +1,28 @@
...
+
+      <column name="project_name" type="varchar(255)">
+        <constraints unique="true" nullable="false" />
+      </column>
+
+      <column name="creation_time" type="datetime">
+        <constraints nullable="false" />
+      </column>
+
+      <column name="status" type="varchar(32)">
```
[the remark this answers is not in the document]
ericsong: Made the change. But I am wondering... how much flexibility does this really give us in being able to change our enum values? If we expose the enum string values as part of our API, we're still unable to rename values without a major API version upgrade.

On the downside, this makes analyzing our data through SQL a little bit harder. And SQL that we do write needs to use short values like the one in my Dao, `SELECT COUNT(*) FROM BillingProjectBufferEntry WHERE status=0 OR status=2 OR status=3`.

Do you have a suggestion for being able to do the above without having to use the short values?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -0,0 +1,62 @@
...
+      Clock clock,
+      FireCloudService fireCloudService,
+      Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.billingProjectBufferEntryDao = billingProjectBufferEntryDao;
+    this.clock = clock;
+    this.workbenchConfigProvider = workbenchConfigProvider;
+    this.fireCloudService = fireCloudService;
+  }
+
+  public void bufferBillingProject() {
```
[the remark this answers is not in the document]
ericsong: yeah... this was a tough one to name. I went through a bunch of iterations that used add/create and eventually found `buffer` which I hoped would imply that the action will not exceed the buffer's capacity.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: i believe we decided it was fine to keep the full UUID since its not shown to the user at any point.

gjuggler: I might have misunderstood our tech talk conclusion, but I thought we'd move from

https://all-of-us-workbench-test.appspot.com/workspaces/aou-rw-test-324/my-workspace-name

to something like

https://all-of-us-workbench-test.appspot.com/workspaces/aou-rw-test-12345/

e.g. get rid of the workspace name in the URL path, since it's no longer required once project==workspace.

This maintains the nice property that the URL is globally unique, since no two GCP projects can have the same project ID.

ericsong: oh, right.

I was thinking we could still decouple the billing project id === workspace name and allow the user to create a customizable workspace id. Given that the billing project id has to be unique given the current architecture constraints with the buffer.

All that said, i think we decided it was expose the full guid in the url as long as its the last part of the URL. Which it will be in the workspace === billing project model.

gjuggler: For the workspace URL itself it will be the last token, but for e.g. notebooks URLs it will still be stuck in the middle. A URL from the test RW today is:

https://all-of-us-workbench-test.appspot.com/workspaces/aou-rw-test-324/asdf/notebooks/Using%2520Cohorts%2520and%2520Concepts.ipynb?playgroundMode=false

The project ID also shows up in the cloud console, and may eventually be used to derive a workspace bucket (which is currently "fc-{GUID}", but could be changed once we have 1-to-1 mapping from project to workspace).

So overall there's still probably some aesthetic pressure to keep the project ID ~short.

The nice thing is, the chance of collision is small and Google will handle rejecting an attempted duplicate project creation. So there's probably no architecture change necessary beyond your already-planned failure tolerance.

(Quick historical note: Firecloud used to solve this for test projects by sticking a few random words at the end, e.g. "fiab-sshah-known-sawfly". It's cute, but random UUID chars are probably safer.)

ericsong: OK. I'm cool with sizing it down in case we end up using it in the future. Thoughts on how big?

api/db/changelog/db.changelog-73-add-BillingProjectBufferEntry-table.xml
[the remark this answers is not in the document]
calbach: The guiding principle for me on this subject is to avoid situations where your storage implementation restricts your ability to change an enum name in code for clarity. APIs I think are a separate question - API definitions should generally be decoupled from storage representations as a design principle; you need to maintain flexibility to change these independently.

Agreed this makes raw SQL interactions less readable; I'd guess there are tricks for dealing with this but I'm not aware of any easy fix. In that particular case, one alternative is to paramterize the set of statuses rather than encoding them directly in the Dao, and pass that in as a `Set<Short>`, `WHERE status IN @statuses`

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
calbach: 32 bits of entropy seems sufficient to me indefinitely. Could probably get away with ~24 bits but I don't know if it's worth the optimization (and we might need to increase it later). Supporting up to 100K workspaces comfortably is probably a reasonable enough upper bound to consider for now.

api/src/main/webapp/WEB-INF/cron.xml
```diff
@@ -30,4 +30,10 @@
     <description>Daily sync of 2FA set up status (via Google Directory Service)</description>
     <schedule>every 24 hours</schedule>
   </cron>
+  <cron>
+    <url>/v1/cron/bufferBillingProject</url>
+    <target>api</target>
+    <description>If the AoU Billing Project buffer is not full, create a firecloud billing project and add one</description>
+    <schedule>every 5 seconds</schedule>
```
calbach: This may be too rapid; I think 1 per minute might suffice (we can tune this later if needed)

api/src/main/java/org/pmiops/workbench/db/model/BillingProjectBufferEntry.java
```diff
@@ -0,0 +1,63 @@
...
+
+@Entity
+public class BillingProjectBufferEntry {
+
+  private long id;
+  private String projectName;
+  private Timestamp creationTime;
+  private Status status;
+
+  public enum Status {
```
calbach: Should probably add documentation for these enum values

api/db/changelog/db.changelog-73-add-BillingProjectBufferEntry-table.xml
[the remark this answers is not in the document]
calbach: Also, more context on that decision here: https://precisionmedicineinitiative.atlassian.net/browse/RW-872

I think it's reasonable to disagree on preferred approach here, both have merit. Since we're already using this approach, I would definitely prefer to keep that current pattern unless I am overwhelmingly overruled by the rest of the team.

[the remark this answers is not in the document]
ericsong: I am cool with following the current paradigm.

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -8,7 +8,7 @@
 @Repository
 public interface BillingProjectBufferEntryDao  extends CrudRepository<BillingProjectBufferEntry, Long> {
 
-  BillingProjectBufferEntry findByProjectName(String projectName);
+  BillingProjectBufferEntry findByFireCloudProjectName(String fireCloudProjectName);
 
   @Query("SELECT COUNT(*) FROM BillingProjectBufferEntry WHERE status=0 OR status=2 OR status=3")
```
calbach: nit: `status IN (0, 2, 3)` might be slightly more readable/extensible

api/db/changelog/db.changelog-73-add-BillingProjectBufferEntry-table.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
calbach: nit: filename and ID name should match updated table name

api/db/changelog/db.changelog-73-add-BillingProjectBufferEntry-table.xml
```diff
@@ -0,0 +1,28 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="ericsong" id="db.changelog-73-add-BillingProjectBufferEntry-table">
+    <createTable tableName="billing_project_buffer_entry">
+
+      <column name="billing_projet_buffer_entry_id" type="bigint" autoIncrement="true">
```
calbach: project

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface BillingProjectBufferEntryDao  extends CrudRepository<BillingProjectBufferEntry, Long> {
+
+  BillingProjectBufferEntry findByFireCloudProjectName(String fireCloudProjectName);
+
+  @Query("SELECT COUNT(*) FROM BillingProjectBufferEntry WHERE status=0 OR status=2 OR status=3")
```
calbach: I'd not include `3` here. I may have said differently in my doc - I don't remember.

If it's actively assigning, it's been claimed (either for assignment, or if that fails for deletion), so I think it's correct to exclude it from the buffer size.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+  @Autowired
+  private Provider<WorkbenchConfig> workbenchConfigProvider;
+
+  private BillingProjectBufferService billingProjectBufferService;
+
+  private final long BUFFER_CAPACITY = 5;
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
```
calbach: Is this required? We don't have this in our other tests AFAIK.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    billingProjectBufferService = new BillingProjectBufferService(
```
calbach: It seems like you should be able to just export a `@Bean Clock` from your test config, then just `@Autowire BillingProjectBufferService`, and avoid the manual construction.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    String billingProjectName = captor.getValue();
+
+    assertThat(billingProjectName).startsWith(workbenchConfig.firecloud.billingProjectPrefix);
+    assertThat(billingProjectBufferEntryDao.findByFireCloudProjectName(billingProjectName).getStatusEnum())
+        .isEqualTo(CREATING);
+  }
+
+  @Test
+  public void fillBuffer_capacity() {
```
calbach: nit: one case worth covering would be reduction of buffer capacity below the number of buffered projects. I'd probably make this a separate test case

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
ericsong: i think you can use it if you have `@Mock` annotations in your test class. AFAIK, it just finds the mock annotations and assigns them to a mock of the given type so it just saves you a few lines.

I don't have mocks in this class though so I removed it

PR #2066, opened 2019-05-03
[RW:2343][RISK:LOW]Preview issues: 500 AND column mismatch

NehaBroad:


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -218,7 +220,14 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
 
       queryResponse.getValues().forEach(fieldValueList -> {
         IntStream.range(0, fieldValueList.size()).forEach(columnNumber -> {
-          valuePreviewList.get(columnNumber).addQueryValueItem(fieldValueList.get(columnNumber).getValue().toString());
+          try {
+            valuePreviewList.get(columnNumber).addQueryValueItem(
+                fieldValueList.get(columnNumber).getValue().toString());
+          } catch (NullPointerException ex) {
+            log.severe(
```
blrubenstein: Do you know what was causing this? If it's expected, this should maybe be log.warning or log.info rather than severe?

NehaBroad: I still have not been able to track down the exact column and its value. But keeping an eye on it.

NehaBroad:
@blrubenstein : PR should be better now as the preview data is in.

PR #2067, opened 2019-05-05
[RW-2578] [risk=no] Temporarily disable the two-factor sync cron job.

gjuggler:
See linked ticket for discussion. The goal is to quickly cherry-pick the job to be disabled, with an aim to re-enable once we understand and mitigate the impact on Terra.


gjuggler:
> LGTM. If it’s a temporary disable, should we just comment out for now?

I always prefer to remove – it's easy enough to find removed code in Git to revive.

PR #2073, opened 2019-05-06
[RW-2467][risk=no] Fix JPA issues with BillingProjectBufferEntry

ericsong:
Fixes the following exception on create - 
- Caused by: java.sql.SQLSyntaxErrorException: Unknown column 'statusEnum' in 'field list'

Fixes the following exception on getCurrentBufferSize() - 
- org.hibernate.hql.internal.ast.QuerySyntaxException: billing_project_buffer_entry is not mapped

One important question that we should look into... why did our tests not catch this? The tests go through both of these operations using a real Dao but they pass without these changes. These failures only come up when running it manually through the application.


calbach:
Optional: while you're fixing this, one thing I noticed after you merged the last one is that there are some unused variables at L65 of BillingProjectBufferServiceTest.java

PR #2074, opened 2019-05-06
[RW-2578] [risk=no] Additional logging for data access level changes and sync jobs.

gjuggler:
While digging into the past weekend's Terra incident (see RW-2578 for details) I found it was very hard to tell exactly how many times the workbench was making changes to the all-of-us-registered-prod group.

This PR adds some add'l logging which will help us more confidently assess how many changes are happening.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
@@ -162,12 +162,14 @@ public class UserService {
       if (!isInGroup) {
         this.fireCloudService.addUserToGroup(user.getEmail(),
             configProvider.get().firecloud.registeredDomainName);
+        log.info(String.format("Added user %s to registered-tier group.", user.getEmail()));
       }
       user.setDataAccessLevelEnum(DataAccessLevel.REGISTERED);
     } else {
       if (isInGroup) {
         this.fireCloudService.removeUserFromGroup(user.getEmail(),
             configProvider.get().firecloud.registeredDomainName);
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -385,8 +387,8 @@ public boolean getContactEmailTaken(String contactEmail) {
   /**
    *  Syncs the current user's training status from Moodle.
    */
-  public void syncComplianceTrainingStatus() throws org.pmiops.workbench.moodle.ApiException, NotFoundException {
-    syncComplianceTrainingStatus(userProvider.get());
+  public User syncComplianceTrainingStatus() throws org.pmiops.workbench.moodle.ApiException, NotFoundException {
+    return syncComplianceTrainingStatus(userProvider.get());
```
jacmrob: why change this to return a user?

gjuggler: This makes it easier to compare the user state before & after doing the status sync.

While we tend to make changes directly on the user object passed into these methods, it's theoretically possible to get a new User reference if the saveUserWithRetries process kicks in and fetches a new user from the database. So I thought making these methods explicitly return a User would be safest.

(Possible follow-up: change these methods to not modify the inbound User object and always return a new copy. I didn't bite this off in this PR since there isn't an easily-accessible User.clone() method... it may be this pattern would be too much effort to implement given the limited gains.)

PR #2078, opened 2019-05-06
[RW-2561][risk=no] Add cron to sync BillingProjectBufferEntries

ericsong:
Adds a cron that will fetch a BillingProjectBufferEntry and sync its status with Fireclouds


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -51,11 +54,36 @@ public void bufferBillingProject() {
     billingProjectBufferEntryDao.save(entry);
   }
 
+  public void syncBillingProjectStatus() {
+    BillingProjectBufferEntry entry = billingProjectBufferEntryDao
```
calbach: I have some concerns about degenerate situations here, e.g. the oldest project stuck in this state forever. Maybe grab the first N instead and iterate? (maybe 5)

There's another ticket for handling that situation, but we'll probably want to allow a generous timeout and we don't want all projects to be stuck in the meantime.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -51,11 +54,36 @@ public void bufferBillingProject() {
...
+    if (entry == null) {
+      return;
+    }
+
+    switch (fireCloudService.getBillingProjectStatus(entry.getFireCloudProjectName()).getCreationStatus()) {
+      case READY:
+        entry.setStatusEnum(AVAILABLE);
+        break;
+      case ERROR:
+        entry.setStatusEnum(ERROR);
```
calbach: It is probably worth logging an error here, and/or an error message if we have one from FC.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -51,11 +54,36 @@ public void bufferBillingProject() {
...
+    BillingProjectBufferEntry entry = billingProjectBufferEntryDao
+        .findFirstByStatusOrderByCreationTimeAsc(StorageEnums.billingProjectBufferStatusToStorage(CREATING));
+
+    if (entry == null) {
+      return;
+    }
+
+    switch (fireCloudService.getBillingProjectStatus(entry.getFireCloudProjectName()).getCreationStatus()) {
+      case READY:
+        entry.setStatusEnum(AVAILABLE);
```
calbach: opt: could be nice to log the time it took to create the project. Terra should track this as well, but it's been a long time coming

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -164,4 +184,87 @@ public void fillBuffer_decreaseCapacity() {
...
+  @Test
+  public void syncBillingProjectStatus_error() {
+    billingProjectBufferService.bufferBillingProject();
+
+    ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
+    verify(fireCloudService).createAllOfUsBillingProject(captor.capture());
+    String billingProjectName = captor.getValue();
+
+    BillingProjectStatus billingProjectStatus = new BillingProjectStatus();
+    billingProjectStatus.setCreationStatus(CreationStatusEnum.ERROR);
```
calbach: opt: can chain this

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -164,4 +184,87 @@ public void fillBuffer_decreaseCapacity() {
...
+  @Test
+  public void syncBillingProjectStatus_multiple() {
+    billingProjectBufferService.bufferBillingProject();
+    billingProjectBufferService.bufferBillingProject();
+    billingProjectBufferService.bufferBillingProject();
+
+    ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
+    verify(fireCloudService, times(3)).createAllOfUsBillingProject(captor.capture());
+    List<String> capturedProjectNames = captor.getAllValues();
+    doReturn(createBillingProjectStatus(CreationStatusEnum.READY)).when(fireCloudService)
```
calbach: nit: loop?

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -164,4 +184,87 @@ public void fillBuffer_decreaseCapacity() {
...
+    doReturn(createBillingProjectStatus(CreationStatusEnum.ERROR)).when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(2));
+
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.syncBillingProjectStatus();
+
+    assertThat(billingProjectBufferEntryDao.countByStatus(StorageEnums.billingProjectBufferStatusToStorage(CREATING)))
+    .isEqualTo(0);
```
calbach: nit: indent

api/src/main/webapp/WEB-INF/cron.xml
```diff
@@ -30,4 +30,10 @@
     <description>If the AoU Billing Project buffer is not full, create a firecloud billing project and add one</description>
     <schedule>every 1 minutes</schedule>
   </cron>
+  <cron>
+    <url>/v1/cron/syncBillingProjectStatus</url>
+    <target>api</target>
+    <description>Fetch a BillingProjectBufferEntry that is in the CREATING state and check its status on Firecloud</description>
+    <schedule>every 5 seconds</schedule>
```
calbach: I think 1 minute should be fine again, especially if you iterate multiple projects per run (per other comment)

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: Hmm, yeah that's a good point. I think it's a very valid concern given the instability of FC. How about adding another field to the model to track the last synced time? We can change the orderBy to use that field and the sync will give every entry a chance to be synced without being blocked by a previous failing entry.

This also tracks the project creation time duration for free. `SELECT (last_synced_time - creation_time) FROM e WHERE e.status=AVAILABLE, ASSIGNED, ASSIGNING`

This also lets us keep the pattern where each invocation of the sync does a small piece of work (n=1) and we can control its throughput by adjusting only the cron frequency. I prefer that because it's easier to adjust and understand throughput with just 1 variable and it distributes load more evenly.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: I don't believe there's an error message in this code path since this is technically a 200 from FC. If the actual get request fails, we should see the stack trace getting logged.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
calbach: Sure - that sounds reasonable

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
ericsong: I decided to just write them out since the third mock return is different. It would end up being a n=2 loop + another statement which is just as verbose as writing out all three

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: I did add a log for when this code path gets hit though

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -56,19 +60,28 @@ public void bufferBillingProject() {
...
+        case READY:
+          entry.setStatusEnum(AVAILABLE);
+          break;
+        case ERROR:
+          log.warning(String.format("SyncBillingProjectStatus: BillingProject %s creation failed",
+              entry.getFireCloudProjectName()));
+          entry.setStatusEnum(ERROR);
+          break;
+      }
+    } catch (Exception e) {
```
calbach: I'd make this a more specific type to avoid suppressing programming errors - are you trying to catch a FirecloudApiException here?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -56,19 +60,28 @@ public void bufferBillingProject() {
...
+      e.printStackTrace();
```
calbach: nit: Would probably be more consistent to write this exception via the logger, e.g. `log.log(Level.WARNING, "Firecloud project check failed", e);`

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: yeah. It doesn't look like Firecloud. ApiException gets thrown though.

It seems like FireCloudService always returns a WorkbenchException so I catch that instead.

PR #2080, opened 2019-05-07
[RW-2392] [risk=no] Improve stability of access module completion timestamps

gjuggler:
While testing the cron jobs locally, I noticed that a couple of the sync processes were continually updating the "completionTime" for the given access module every time the sync ran. This stabilizes the completion times so they don't get updated for all users on every run.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
@@ -422,17 +422,21 @@ public class UserService {
       List<BadgeDetails> badgeResponse = complianceService.getUserBadge(moodleId);
       // The assumption here is that the User will always get 1 badge which will be AoU
       if (badgeResponse != null && badgeResponse.size() > 0) {
+        BadgeDetails badge = badgeResponse.get(0);
+        Timestamp badgeExpiration = badge.getDateexpire() == null ? null :
+            new Timestamp(Long.parseLong(badge.getDateexpire()));
+
         if (user.getComplianceTrainingCompletionTime() == null) {
+          // This is the user's first time with a Moodle badge response, so we reset the completion
+          // time.
+          user.setComplianceTrainingCompletionTime(now);
+        } else if (badgeExpiration != null && !badgeExpiration.equals(user.getComplianceTrainingExpirationTime())) {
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -150,8 +156,13 @@ public void testSyncEraCommonsStatus() throws Exception {
...
         new Timestamp(TIMESTAMP_MSECS));
     assertThat(user.getEraCommonsLinkExpireTime()).isEqualTo(
-        new Timestamp(TIMESTAMP_MSECS + 1));
+        new Timestamp(TIMESTAMP_MSECS / 1000));
     assertThat(user.getEraCommonsLinkedNihUsername()).isEqualTo("nih-user");
+
+    // Completion timestamp should not change when the method is called again.
+    Timestamp completionTime = user.getEraCommonsCompletionTime();
+    userService.syncEraCommonsStatus();
+    assertThat(user.getEraCommonsCompletionTime()).isEqualTo(completionTime);
```
jacmrob: what will happen if this assertion fails?

gjuggler: This was meant to be a regression test for the logic change to avoid updating the completion time , following the pattern we use for 2fa at L196.

Looking at this now, I realize that the current timestamp isn't changing between calls, so it wouldn't actually catch a regression. Fixed.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -421,11 +422,13 @@ public User syncComplianceTrainingStatus(User user) throws org.pmiops.workbench.
       List<BadgeDetails> badgeResponse = complianceService.getUserBadge(moodleId);
       // The assumption here is that the User will always get 1 badge which will be AoU
       if (badgeResponse != null && badgeResponse.size() > 0) {
-        user.setComplianceTrainingCompletionTime(now);
+        if (user.getComplianceTrainingCompletionTime() == null) {
+          user.setComplianceTrainingCompletionTime(now);
```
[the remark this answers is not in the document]
jacmrob: looks good 👍

PR #2084, opened 2019-05-07
[RW-2483][risk=low] workspace deletion policy

jacmrob:


ui/src/app/views/workspace-list/component.tsx
```diff
@@ -246,6 +246,12 @@ export const WorkspaceList = withUserProfile()
     this.reloadWorkspaces();
   }
 
+  componentDidUpdate() {
```
jacmrob: this fixes a little bug I noticed when testing deletion that the create workspace button on the list workspaces page never updated to be active when billing proj loaded

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -294,6 +295,7 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
   @Override
   public ResponseEntity<EmptyResponse> deleteWorkspace(String workspaceNamespace,
       String workspaceId) {
+    // TODO: do we want to delete any workspace resources?
```
jacmrob: currently we are not; we need to save workbench metadata and collaborator lists (the `user_workspace` table), but could get rid of other pieces.  however if we keep that stuff we can easily implement workspace restore

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -120,7 +121,7 @@ public Workspace getByName(String ns, String name) {
   @Override
   public Workspace getRequired(String ns, String firecloudName) {
     Workspace workspace = get(ns, firecloudName);
-    if (workspace == null) {
+    if (workspace == null || (workspace.getWorkspaceActiveStatusEnum() == WorkspaceActiveStatus.DELETED)) {
```
jacmrob: it makes sense to me that the get endpoint should not fetch deleted workspaces; would like to hear some other opinions.

calbach: I would probably flip this to != ACTIVE, it is unlikely we'd want other statuses showing up here, if we added any others in the future

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -461,6 +466,7 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     dbWorkspace.setCreationTime(now);
     dbWorkspace.setLastModifiedTime(now);
     dbWorkspace.setVersion(1);
+    dbWorkspace.setWorkspaceActiveStatusEnum(WorkspaceActiveStatus.ACTIVE);
```
ericsong: Can we set this as a default value in the model class? I believe it should work

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -106,6 +106,7 @@
 import org.pmiops.workbench.model.UserRole;
 import org.pmiops.workbench.model.Workspace;
 import org.pmiops.workbench.model.WorkspaceAccessLevel;
+import org.pmiops.workbench.model.WorkspaceActiveStatus;
```
ericsong: unused import?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -96,31 +97,37 @@ public Workspace get(String ns, String firecloudName) {
...
                 fcWorkspace -> fcWorkspace.getWorkspace().getWorkspaceId(),
                 fcWorkspace -> fcWorkspace));
   }
 
   @Override
   public Workspace getByName(String ns, String name) {
-    return workspaceDao.findByWorkspaceNamespaceAndName(ns, name);
+    Workspace workspace = workspaceDao.findByWorkspaceNamespaceAndName(ns, name);
+    if (workspace == null || workspace.getWorkspaceActiveStatusEnum() != WorkspaceActiveStatus.ACTIVE) {
+      return null;
```
jacmrob: @calbach I added status checks for some of our other getter methods, however for this one (which is called to check if a ws name already exists when creating/cloning a ws) i think we do not want to error out with 404 if a workspace is found but not active, since we probably do want to allow people to create/clone workspaces with the same names as deleted workspaces, yes?

calbach: Well.. I'm pretty confused. Our controllers seem to be acting under the assumption that workspace name is unique, yet there is no uniqueness constraint applied at the database level. In several of these cases it seems like we're expecting to return null rather than 404, which is not well documented. I think in the getByName case it should be returning null if there are no active matching workspaces.

Would need to think more about whether we want an index uniqueness constraint on name (if so, this approach of reusing the table presents some issues). If not, we should think carefully about create/clone behaviors. But I think returning null should do the trick here for now.

api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
```diff
@@ -94,6 +97,10 @@ public boolean equals(Object obj) {
 
   private Set<WorkspaceUserRole> usersWithAccess = new HashSet<WorkspaceUserRole>();
 
+  public Workspace() {
+    setWorkspaceActiveStatusEnum(WorkspaceActiveStatus.ACTIVE);
```
calbach: This looks a little scary to me; are you certain this works and won't just overwrite whatever status was in the database when we load a model from the database? (and then overwrite the database entry when we save it back?)

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -96,31 +97,42 @@ public Workspace get(String ns, String firecloudName) {
...
+    }
+    return workspace;
   }
 
   @Override
   public Workspace getRequired(String ns, String firecloudName) {
     Workspace workspace = get(ns, firecloudName);
-    if (workspace == null) {
+    if (workspace == null || (workspace.getWorkspaceActiveStatusEnum() != WorkspaceActiveStatus.ACTIVE)) {
```
calbach: Therea re a few references to `enforceWorkspaceAccessLevel` which I think should be covered, e.g. copyNotebook().

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
gjuggler: Quick comment from the peanut gallery: we should definitely avoid trying to maintain any uniqueness of names unless we're otherwise forced to (e.g. notebook filenames which need to be unique within a directory).

I was going to say we should aim to use workspace IDs wherever possible in our URLs and API calls...

but it seems like Terra's workspace APIs only ever take the workspace name, not workspace ID, as parameters. In fact, it's not possible to change a workspace's name via Terra, so they must be treating it as some sort of actual ID (even though there's a separate workspaceUuid). Dang.

Still, we should probably strive to do things as "right" as possible in our own environment, and consider urging Terra to improve their APIs over time.

calbach: This is another point of terminology confusion, I'm 99% sure firecloudName is locally unique within a project (the docs don't really say). workspaceId is a UUID which is globally unique. The name referenced here is the display name for the AoU workspace.

I don't have any context on why there's this secondary UUID "workspaceId", but I would consider the RESTFUL identifier for a Firecloud workspace to be `workspaces/{namespace}/{name}`. Our referencing that primary API identifier seems correct (assuming there aren't some strange caveats about them changing under you).

api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
jacmrob: it seems to work!  when I reload my workspaces after deleting, the deleted ones are still deleted

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jacmrob: what do you mean? like a user should not be able to copy a notebook from a deleted workspace?

calbach: Yes - and it should be 404ing if I try that. Perhaps it already works since we're checking FC for the ACL, but worth verifying.

There are several cohort methods that also use this method as their only access check.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
gjuggler: Agreed, what we're doing re: referring to the FireCloud workspaceName as a unique & immutable identifier is correct. 

But there's still plenty of room for confusion in our codebase and on the user's end. For example, today if I create a workspace and then rename it in AoU, our workspace URL shows the old workspace name (but lowercased and with spaces removed):

Screen Shot 2019-05-10 at 4 28 29 PM

The billing model change further complicates things. We'll soon have ~5 fields all referring to some variant of the same thing:

1. GCP project ID
2. GCP project name (Terra doesn't read/write this right now, but it does show up in a user's cloud console)
3. AoU workspace name (actually a name)
4. Terra workspace name (actually an ID)
5. Terra workspace ID (a UUID)

But the change is also an opportunity – I'm optimistic that when the dust settles we can ensure / enforce that 1, 2, 4 are all identical and our URLs end up looking something like http://workbench.researchallofus.org/workspaces/aou-d2fbe9c/. Which is probably the model that Terra should work towards as well.

calbach: :+1: I agree this would be the ideal behavior. There are definitely some issues to confront in getting there, e.g. global namespacing (a workspace name cannot match any existing Google project ID) but they are probably surmountable.

PR #2088, opened 2019-05-07
[RW-2561][risk=no] switch cron schedule to every minute and increase syncs per invocation

ericsong:
Turns out the highest frequency cron schedule is once a minute

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -23,6 +23,7 @@ import org.springframework.stereotype.Service;
 @Service
 public class BillingProjectBufferService {
 
+  private static final int SYNCS_PER_INVOCATION = 5;
   private static final int PROJECT_BILLING_ID_SIZE = 8;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -59,35 +60,38 @@ public void bufferBillingProject() {
   }
 
   public void syncBillingProjectStatus() {
-    BillingProjectBufferEntry entry = billingProjectBufferEntryDao
-        .findFirstByStatusOrderByLastSyncRequestTimeAsc(StorageEnums.billingProjectBufferStatusToStorage(CREATING));
+    for (int i = 0; i < SYNCS_PER_INVOCATION; i++) {
```
calbach: opt: Probably a bit more efficient to just do a limit 5 on the dao, then iterate on the list in memory.

ericsong: Yeah... I did look into it but it requires adding pagination code to the repository query since Hibernate/JPA does not support LIMIT queries. Seemed like enough additional effort that wasn't worth the return.  I am also hoping to switch back to an n=1 function if the Google Tasks thing works out.

PR #2090, opened 2019-05-07
[RW-2105][Risk=low] Working functionality for export to existing notebook

blrubenstein:


blrubenstein:
Good call on tests. Will add those now.

PR #2093, opened 2019-05-07
[No Ticket][risk=no]Start UI Coding Style Guide

ericsong:
Action item of 4/26 retrospective with patterns discussed in mob programming session on 4/30

### DIFF
--- a/ui/CodeStyleGuide.md
+++ b/ui/CodeStyleGuide.md
@@ -1,0 +1,58 @@
+# Coding Style Guide
+
+## React Component Structure
+
+  <a name="react-component-structure--props-and-state"></a><a name="1.1"></a>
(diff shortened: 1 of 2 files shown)


ui/CodeStyleGuide.md
```diff
@@ -0,0 +1,58 @@
...
+
+    class MyReactComponent extends React.Component<Props, State> { }
+
+    export {
+      MyReactComponent,
+      Props as MyReactComponentProps
+    };
+    ```
+    - One place to look at to determine what a file is exposing versus having to scan the file for export statements
+    - Aliases can be used to namespace only when necessary. Ex. Props can just be Props in this file but can be prefixed
```
jmthibault79: I would word this more prescriptively, something like "Only provide an alias when it's necessary to reference this externally"

PR #2094, opened 2019-05-08
[RW-2143][RW-2323][risk=no] Add list and tree views for drug domain

dolbeew:
Add Drugs to criteria dropdown for the new CB list search

### DIFF
--- a/ui/src/app/cohort-search/constant.ts
+++ b/ui/src/app/cohort-search/constant.ts
@@ -37,7 +37,8 @@ export const LIST_PROGRAM_TYPES = [];
 
 export const LIST_DOMAIN_TYPES = [
   {name: 'Measurements', domain: DomainType.MEASUREMENT, type: null},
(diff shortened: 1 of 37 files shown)


ui/src/app/cohort-search/constant.ts
```diff
@@ -37,7 +37,8 @@ export const LIST_PROGRAM_TYPES = [];
 
 export const LIST_DOMAIN_TYPES = [
   {name: 'Measurements', domain: DomainType.MEASUREMENT, type: null},
-  {name: 'Conditions', domain: DomainType.CONDITION, type: null}
+  {name: 'Conditions', domain: DomainType.CONDITION, type: null},
+  {name: 'Drugs', domain: DomainType.DRUG, type: null}
```
freemabd: if type is null for each do we need init that prop?

dolbeew: We may need it eventually, but I removed it for now. Will be easy to put back if needed.

PR #2098, opened 2019-05-08
[RW-1975][risk=no] Remove public db/user refs and corresponding scripts

calbach:
Follow-up TODO:
- remove the public_db_name column from the cdr_version table.


api/config/cdr_versions_local.json
```diff
@@ -10,7 +10,6 @@
     "releaseNumber": 1,
     "numParticipants": 946237,
     "cdrDbName": "cdr",
-    "elasticIndexBaseName": "cdr",
-    "publicDbName": "public"
```
freemabd: Does removing this property break the Cdr updater? https://github.com/all-of-us/workbench/blob/master/api/tools/src/main/java/org/pmiops/workbench/tools/UpdateCdrVersions.java#L34

calbach: Tested - still works. publicDbName is nullable so it's not a problem.

PR #2099, opened 2019-05-08
[RW-2260][risk=no] Fix "Save As" URI encoding on shared workspaces

ericsong:
Shared Workspaces cannot use "Save As" since the modal encodes the URI which turns our current delimiter `:` into characters that Jupyter doesn't decode correctly when it goes back to their API.

The solution is to change it into characters that do not get encoded. We chose `__`.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -21,6 +21,7 @@ import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.UserDao;
 import org.pmiops.workbench.db.dao.UserRecentResourceService;
 import org.pmiops.workbench.db.dao.UserService;
+import org.pmiops.workbench.firecloud.FireCloudServiceImpl;
 import org.pmiops.workbench.workspaces.WorkspaceService;
 import org.pmiops.workbench.db.model.CdrVersion;
 import org.pmiops.workbench.db.model.User;
@@ -51,7 +52,6 @@ public class ClusterController implements ClusterApiDelegate {
   // delocalization of saved files back to a given GCS location. See
   // https://github.com/DataBiosphere/leonardo/blob/develop/jupyter-docker/jupyter_delocalize.py#L12
   private static final String DELOCALIZE_CONFIG_FILENAME = ".delocalize.json";
-  public static final String WORKSPACE_DELIMITER = "__";
 
   // This file is used by the All of Us libraries to access workspace/CDR metadata.
   private static final String AOU_CONFIG_FILENAME = ".all_of_us_config.json";
@@ -205,11 +205,11 @@ public class ClusterController implements ClusterApiDelegate {
     );
     String workspacePath = body.getWorkspaceId();
     if (!projectName.equals(body.getWorkspaceNamespace())) {
-      workspacePath = body.getWorkspaceNamespace() + WORKSPACE_DELIMITER + body.getWorkspaceId();
+      workspacePath = body.getWorkspaceNamespace() + FireCloudServiceImpl.WORKSPACE_DELIMITER + body.getWorkspaceId();
     }
     String apiDir = "workspaces/" + workspacePath;
     if (body.getPlaygroundMode()) {
-      apiDir = "workspaces" + WORKSPACE_DELIMITER + "playground/" + workspacePath;
+      apiDir = "workspaces_playground/" + workspacePath;
     }
     String localDir = "~/" + apiDir;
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -1,5 +1,7 @@
 package org.pmiops.workbench.firecloud;
 
+import static org.pmiops.workbench.api.ClusterController.WORKSPACE_DELIMITER;
```
calbach: Could you put this constant somewhere central? This creates a bit of an awkward dependency graph (low level service referencing a controller).

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -204,11 +205,11 @@ public void setUserService(UserService userService) {
     );
     String workspacePath = body.getWorkspaceId();
     if (!projectName.equals(body.getWorkspaceNamespace())) {
-      workspacePath = body.getWorkspaceNamespace() + ":" + body.getWorkspaceId();
+      workspacePath = body.getWorkspaceNamespace() + WORKSPACE_DELIMITER + body.getWorkspaceId();
     }
     String apiDir = "workspaces/" + workspacePath;
     if (body.getPlaygroundMode()) {
-      apiDir = "workspaces:playground/" + workspacePath;
+      apiDir = "workspaces" + WORKSPACE_DELIMITER + "playground/" + workspacePath;
```
calbach: I'd name the constant more generically, or else just hardcode _ here. `workspaces:playground` doesn't have any of the same uniqueness constraints as the other directory. Could also just be a single underscore in this case.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -192,6 +194,10 @@ public void registerUser(String contactEmail, String firstName, String lastName)
 
   @Override
   public void createAllOfUsBillingProject(String projectName) {
+    if (projectName.contains(WORKSPACE_DELIMITER)) {
+      throw new RuntimeException("Attempting to create billing project with name that contains workspace delimiter : " + projectName);
```
calbach: Maybe just log severe instead. This wouldn't be the end of the world, and I'm not sure we want to be 500ing requests at runtime for this reason.

ericsong: hmm, I think throwing an exception is correct here since it should really never happen. There is no code path that should allow it at the moment and if a developer makes a change so that it does happen, it is likely because they were unaware of how it interacts with the delimiter. In that case, I think it should fail quickly and apparently so we can detect the change before it gets merged.

ericsong:
@calbach  responded to 1 comment. addressed the rest

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -21,6 +21,7 @@
 import org.pmiops.workbench.db.dao.UserDao;
 import org.pmiops.workbench.db.dao.UserRecentResourceService;
 import org.pmiops.workbench.db.dao.UserService;
+import org.pmiops.workbench.firecloud.FireCloudServiceImpl;
```
calbach: nit: Sorry for another nit on this, but should probably be on the service interface instead. The controllers shouldn't have a direct reference to the implementation classes.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
calbach: I'm not convinced this would be caught before merging since the prefix is config-driven and grabbed at runtime, but I guess this at least covers the suffix (alternatively, you could check for this in a test instead).

IllegalArgumentException is probably better if you leave the assertion here.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -353,9 +354,9 @@ public void testLocalize_differentNamespace() throws Exception {
 
     Map<String, String> localizeMap = mapCaptor.getValue();
     assertThat(localizeMap).containsEntry(
-        "~/workspaces/proj:wsid/foo.ipynb", "gs://workspace-bucket/notebooks/foo.ipynb");
-    assertThat(resp.getClusterLocalDirectory()).isEqualTo("workspaces/proj:wsid");
-    JSONObject aouJson = dataUriToJson(localizeMap.get("~/workspaces/proj:wsid/.all_of_us_config.json"));
+        "~/workspaces/proj" + WORKSPACE_DELIMITER + "wsid/foo.ipynb", "gs://workspace-bucket/notebooks/foo.ipynb");
```
calbach: opt_nit: I would recommend just hardcoding the exact string here for testing. It is a bit more maintenance to keep it in sync, but in this case I'd favor being explicit in showing the expected formatted value, rather than duplicating the path construction logic in the test.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
ericsong: that's my bad. You gave me the same feedback before

PR #2100, opened 2019-05-09
[No Ticket][Risk=low] Srubenst/parameterize app header

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
@@ -23,7 +23,6 @@ import org.springframework.web.context.annotation.RequestScope;
 public class FireCloudConfig {
 
   public static final String X_APP_ID_HEADER = "X-App-ID";
-  public static final String X_APP_ID_HEADER_VALUE = "AoU-RW";
 
   // Bean names used to differentiate between an API client authenticated as the end user (via
   // UserAuthentication) and an API client authenticated as the service account user (via
@@ -49,7 +48,7 @@ public class FireCloudConfig {
     apiClient.setBasePath(workbenchConfig.firecloud.baseUrl);
     apiClient.setAccessToken(userAuthentication.getCredentials());
     apiClient.setDebugging(workbenchConfig.firecloud.debugEndpoints);
-    addFireCloudDefaultHeader(apiClient);
+    addFireCloudDefaultHeader(apiClient, workbenchConfig.firecloud.xAppIdValue);
     return apiClient;
   }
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -23,7 +23,6 @@
 public class FireCloudConfig {
 
   public static final String X_APP_ID_HEADER = "X-App-ID";
```
gjuggler: Can you add a note describing our use of this header, and what expectations exist on the FireCloud side?

An alternative place to put this documentation / comment would be as a comment on the WorkbenchConfig class definition. I know we have zero comments there right now, but that's a pretty poor state of affairs & it's never too late to start leaving breadcrumbs of knowledge :)

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -128,13 +127,13 @@ public StatusApi statusApi(WorkbenchConfig workbenchConfig) {
...
     apiClient.setDebugging(workbenchConfig.firecloud.debugEndpoints);
-    addFireCloudDefaultHeader(apiClient);
+    addFireCloudDefaultHeader(apiClient, workbenchConfig.firecloud.xAppIdValue);
     statusApi.setApiClient(apiClient);
     return statusApi;
   }
 
-  private void addFireCloudDefaultHeader(ApiClient apiClient) {
-    apiClient.addDefaultHeader(X_APP_ID_HEADER, X_APP_ID_HEADER_VALUE);
+  private void addFireCloudDefaultHeader(ApiClient apiClient, String value) {
```
gjuggler: This method feels a bit awkward, given it has only a single line and takes a parameter derived from the workbenchConfig, which this class already has access to.

Idea: maybe refactor this into a buildApiClient() method which handles as much common ApiClient logic as possible:

ApiClient apiClient = new ApiClient();
apiClient.setBasePath(workbenchConfig.firecloud.baseUrl);
apiClient.setDebugging(workbenchConfig.firecloud.debugEndpoints);
apiClient.addDefaultHeader(X_APP_ID_HEADER, workbenchConfig.firecloud.xAppIdValue);
return apiClient;

If you make that method public, we could even call it from FireCloudServiceImpl to remove a few lines of code from there too.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksConfig.java
```diff
@@ -89,8 +89,8 @@ public ClusterApi serviceClusterApi(@Qualifier(NOTEBOOKS_SERVICE_CLIENT) ApiClie
     return api;
   }
 
-  private void addNotebooksDefaultHeader(ApiClient apiClient) {
+  private void addNotebooksDefaultHeader(ApiClient apiClient, String xAppHeaderValue) {
```
gjuggler: Same comment here as above – maybe refactoring as an ApiClient builder method would make more sense?

If you don't think that path is worthwhile, I'd be inclined to remove this method and simply make the apiClient.addDeafultHeader call independently in each of the usage locations. This method feels a bit like needless redirection.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
blrubenstein: 👍 to this, adding it into WorkbenchConfig, and adding a tech debt ticket to do an audit of our config file and figure out what isn't clear so we can comment it.

blrubenstein: Filed: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2616&epics=visible&selectedEpic=RW-1801

blrubenstein:
Codacy isn't analyzing for some reason. Going to merge since everything else is passing.

PR #2101, opened 2019-05-09
[No Ticket][Risk=no] Suppress unchecked warning to reduce logging noise

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -318,6 +318,10 @@ public class DataSetController implements DataSetApiDelegate {
     return sqlSection + "\n\n" + namedParamsSection + "\n\n" + dataFrameSection;
   }
 
+  // BigQuery api returns parameter values either as a list or an object.
+  // To avoid warnings on our cast to list, we suppress those warnings here,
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -318,22 +318,27 @@ private String convertQueryToString(DataSetQuery query, String prefix) {
     return sqlSection + "\n\n" + namedParamsSection + "\n\n" + dataFrameSection;
   }
 
+  // BigQuery api returns parameter values either as a list or an object.
+  // To avoid warnings on our cast to list, we suppress those warnings here,
+  // as they are expected.
+  @SuppressWarnings("unchecked")
```
ericsong: If you prefer, you can cast within the map to avoid the suppress warning.

```
            arrayValues.stream()
                .map(o -> (NamedParameterValue) o)
                .map(namedParameterValue -> "{\'value\': " + namedParameterValue.getParameterValue() + "}") + "]" :
```

ericsong: arrayValues would be a `List<Object>` and you would cast to a `List`. This shouldn't fire a warning since its a checked cast.

PR #2104, opened 2019-05-09
[RW-2451][Risk=no] Change workspace edit and caching to do the right …

blrubenstein:
…thing

### DIFF
--- a/ui/src/app/services/workspace-storage.service.ts
+++ b/ui/src/app/services/workspace-storage.service.ts
@@ -14,8 +14,7 @@ export interface WorkspaceData extends Workspace {
 
 @Injectable()
 export class WorkspaceStorageService {
-  // Cache of loading or completed Promises for workspace data. Key is of the
-  // form "ns/id".
+  // Cache of loaded workspace data. Key is of the form "ns/id".
   private cache = new Map<string, WorkspaceData>();
 
   constructor(private workspacesService: WorkspacesService) {}
--- a/ui/src/app/services/workspace-storage.service.ts
+++ b/ui/src/app/services/workspace-storage.service.ts
@@ -12,10 +12,6 @@ export interface WorkspaceData extends Workspace {
   accessLevel: WorkspaceAccessLevel;
 }
 
-const nextCurrentWorkspace = (workspace: WorkspaceData) => {
-  currentWorkspaceStore.next(workspace);
-};
-
 @Injectable()
 export class WorkspaceStorageService {
   // Cache of loading or completed Promises for workspace data. Key is of the
@@ -33,11 +29,10 @@ export class WorkspaceStorageService {
     const workspace = await this.workspacesService.getWorkspace(wsNs, wsId)
       .toPromise()
       .then((resp) => {
-        const workspaceResp = {
+        return {
           ...resp.workspace,
           accessLevel: resp.accessLevel
         };
-        return workspaceResp;
       })
       .catch((e) => {
         // Purge the cache on error to allow for retries.
@@ -47,7 +42,7 @@ export class WorkspaceStorageService {
         }
         throw e;
       });
-    nextCurrentWorkspace(workspace);
+    currentWorkspaceStore.next(workspace);
     this.cache.set(key, workspace);
     return workspace;
   }
@@ -57,7 +52,7 @@ export class WorkspaceStorageService {
     if (!this.cache.has(key)) {
       return await this.reloadWorkspace(wsNs, wsId);
     }
-    nextCurrentWorkspace(this.cache.get(key));
+    currentWorkspaceStore.next(this.cache.get(key));
     return Promise.resolve(this.cache.get(key));
   }
 }
--- a/ui/src/app/services/workspace-storage.service.ts
+++ b/ui/src/app/services/workspace-storage.service.ts
@@ -2,7 +2,6 @@ import 'rxjs/Rx';
 
 import {Injectable} from '@angular/core';
 
-
 import {Workspace} from 'generated';
 import {WorkspacesService} from 'generated';
 import {WorkspaceAccessLevel} from 'generated';
@@ -13,8 +12,9 @@ export interface WorkspaceData extends Workspace {
 
 @Injectable()
 export class WorkspaceStorageService {
-  // Cache of loaded workspace data. Key is of the form "ns/id".
-  private cache = new Map<string, WorkspaceData>();
+  // Cache of loading or completed Promises for workspace data. Key is of the
+  // form "ns/id".
+  private cache = new Map<string, Promise<WorkspaceData>>();
 
   constructor(private workspacesService: WorkspacesService) {}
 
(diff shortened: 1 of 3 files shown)


ui/src/app/services/workspace-storage.service.ts
```diff
@@ -10,27 +12,32 @@ export interface WorkspaceData extends Workspace {
   accessLevel: WorkspaceAccessLevel;
 }
 
+const nextCurrentWorkspace = (workspace: WorkspaceData) => {
+  currentWorkspaceStore.next(workspace);
+};
+
 @Injectable()
 export class WorkspaceStorageService {
   // Cache of loading or completed Promises for workspace data. Key is of the
```
jmthibault79: update comment

ui/src/app/services/workspace-storage.service.ts
```diff
@@ -10,27 +12,32 @@ export interface WorkspaceData extends Workspace {
   accessLevel: WorkspaceAccessLevel;
 }
 
+const nextCurrentWorkspace = (workspace: WorkspaceData) => {
```
calbach: nit: wrapping this in a helper method saves a few characters at callsites, but adds a layer of indirection for the reader. This is not a good tradeoff in my view - would prefer just inlining.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -596,8 +596,8 @@ export class WorkspaceEditComponent extends ReactWrapperBase {
     this._location.back();
   }
 
-  reloadWorkspace(namespace, id): void {
-    this.workspaceStorage.reloadWorkspace(namespace, id);
+  reloadWorkspace(namespace, id): Promise<WorkspaceData> {
+    return this.workspaceStorage.reloadWorkspace(namespace, id);
```
calbach: I don't see anyone consuming this, which leads me to believe this is either not necessary or it has something to do with getting Angular to detect the promise. If it's the latter, please add a comment explaining.

ui/src/app/services/workspace-storage.service.ts
```diff
@@ -40,15 +47,17 @@ export class WorkspaceStorageService {
         }
         throw e;
       });
-    this.cache.set(key, load);
-    return load;
+    nextCurrentWorkspace(workspace);
```
calbach: I don't think I agree with the idea that interactions with `WorkspaceStorageService` should affect the currentWorkspaceStore. Those interactions are currently handled by the workspace-wrapper, which is not perfect, but it is at least centralized to a clear owner and matches what is rendered on the page (there is nothing to indicate that if I call `WorkspaceStorageService.getWorkspace()`, I'd like to immediately update my context and change things currently shown on the page).

I would instead suggest perhaps plumbing a `reloadWorkspace()` callback down from the workspace-wrapper to its children, or having workspace-wrapper listen for updates on the workspace store and perhaps even having the workspace edit page trigger the update with the reloaded value, e.g.:

```
reloadWorkspace().then((ws) => currentWorkspaceStore.next(ws))
```

ui/src/app/services/workspace-storage.service.ts
```diff
@@ -10,27 +12,32 @@ export interface WorkspaceData extends Workspace {
...
 
-  reloadWorkspace(wsNs: string, wsId: string): Promise<WorkspaceData> {
+  async reloadWorkspace(wsNs: string, wsId: string): Promise<WorkspaceData> {
     const key = this.wsKey(wsNs, wsId);
-    const load = this.workspacesService.getWorkspace(wsNs, wsId)
+    const workspace = await this.workspacesService.getWorkspace(wsNs, wsId)
       .toPromise()
       .then((resp) => {
-        return {
+        const workspaceResp = {
```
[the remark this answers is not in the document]
blrubenstein: I will revert this. I think this may have been done during my debugging process.

ui/src/app/services/workspace-storage.service.ts
blrubenstein: I can just inline it, since it is relatively the same. (As compared to the ProfileStorageService which needs to pass some boilerplate)

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: It has to do with allowing the workspace edit component to properly await the completion of `reloadWorkspace` so it doesn't preload the stale one. 

Specifically, this bug was caused by this line here: 
https://github.com/all-of-us/workbench/pull/2104/files#diff-6bc93013088145c1dbc94c5714d63f39R363

We wanted to block navigation on the update finishing, but since we weren't actually handing down the promise, it wasn't functioning as intended.

ui/src/app/views/workspace-wrapper/component.ts
```diff
@@ -69,7 +69,7 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
         // do not render the child component with a stale workspace.
         this.workspace = undefined;
         this.accessLevel = undefined;
-        return this.workspaceStorageService.getWorkspace(ns, wsid);
+        return Promise.resolve(this.workspaceStorageService.getWorkspace(ns, wsid));
```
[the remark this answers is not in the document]
blrubenstein: Might be stale from previous changes, will investigate.

ui/src/app/services/workspace-storage.service.ts
blrubenstein: That seems reasonable to me. I think if workspaceWrapper is subscribed to the store, we may just get it for free. I will try it

blrubenstein: It broke when I removed these from the storage service. If you like, we can hop on a call to look into it? I'm trying to avoid passing a reload down to every child component.

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: Updating to be promise<void> though, because the actual object returned was not consumed.

ui/src/app/services/workspace-storage.service.ts
calbach: How about my second suggestion?

> or having workspace-wrapper listen for updates on the workspace store and perhaps even having the workspace edit page trigger the update with the reloaded value

calbach: i.e. workspace-edit page calling `currentWorkspaceStore.next`

blrubenstein: Hmmm... let me try that. `currentWorkspaceStore` uses WorkspaceData, so there is a bit of manipulation/merging of data structures, but it could work. However, from a fundamental perspective, what is the purpose of `WorkspaceStorageService` if it isn't affecting `currentWorkspaceStore`. My understanding was that `currentWorkspaceStore` just exposes the current workspace in the cache.

blrubenstein: I'm worried we are just generating two workspace caches, and should just delete `WorkspaceStorageService` if so.

ui/src/app/views/workspace-edit/component.tsx
blrubenstein: Resolving now that we are using that object to publish to the store.

ui/src/app/services/workspace-storage.service.ts
```diff
@@ -2,6 +2,7 @@ import 'rxjs/Rx';
 
 import {Injectable} from '@angular/core';
 
+
```
calbach: Can the changes to this file be reverted? Agree per slack it probably makes sense to look at removing these storage services as a follow-up now that we have the workspace/user stores.

ui/src/app/views/workspace-edit/component.tsx
calbach: opt: I don't really need to code comment here anymore - I think it's clear given the new code that it wants to wait on the response. IN the initial PR the return value wasn't actually used by anything.

PR #2107, opened 2019-05-09
[RW-2476][risk=no] Add billing project name param to listClusters

jmthibault79:
Note: this does not make any changes to cluster creation or status checking - only the API signature.  ~So: the parameter must still be equal to the free tier billing project!~

Also some small opportunistic refactoring.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -138,9 +138,9 @@ public class ClusterController implements ClusterApiDelegate {
 
     User user = this.userProvider.get();
     if (billingProjectId.equals(user.getFreeTierBillingProjectName()) &&
-            user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
+        user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
       throw new FailedPreconditionException(
-              "User billing project is not yet initialized, cannot list/create clusters");
+          "User billing project is not yet initialized, cannot list/create clusters");
     }
 
     org.pmiops.workbench.notebooks.model.Cluster fcCluster;
@@ -166,7 +166,7 @@ public class ClusterController implements ClusterApiDelegate {
       }
     } else if (
         org.pmiops.workbench.notebooks.model.ClusterStatus.RUNNING.equals(fcCluster.getStatus()) &&
-        retries != 0) {
+            retries != 0) {
       this.userService.setClusterRetryCount(0);
     }
     ClusterListResponse resp = new ClusterListResponse();
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -127,19 +127,26 @@ public class ClusterController implements ClusterApiDelegate {
   }
 
   @Override
-  public ResponseEntity<ClusterListResponse> listClusters() {
+  public ResponseEntity<ClusterListResponse> listClusters(String projectName) {
     User user = this.userProvider.get();
+    // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
+    // until then: continue to check the user's free tier project status
+    // because that is the only project we'll pass in here
+    if (!user.getFreeTierBillingProjectName().equals(projectName)) {
+      throw new FailedPreconditionException(
+          "NOT YET IMPLEMENTED: Cannot list clusters using arbitrary project name");
+    }
     if (user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
       throw new FailedPreconditionException(
           "User billing project is not yet initialized, cannot list/create clusters");
     }
-    String project = user.getFreeTierBillingProjectName();
     org.pmiops.workbench.notebooks.model.Cluster fcCluster;
     try {
-      fcCluster = this.leonardoNotebooksClient.getCluster(project, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+      fcCluster = this.leonardoNotebooksClient.getCluster(projectName,
+              LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
     } catch (NotFoundException e) {
-      fcCluster = this.leonardoNotebooksClient.createCluster(
-          project, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+      fcCluster = this.leonardoNotebooksClient.createCluster(projectName,
+              LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
     }
 
     int retries = Optional.ofNullable(user.getClusterCreateRetries()).orElse(0);
@@ -154,7 +161,8 @@ public class ClusterController implements ClusterApiDelegate {
         }
         log.warning("Retrying cluster creation.");
 
-        this.leonardoNotebooksClient.deleteCluster(project, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+        this.leonardoNotebooksClient.deleteCluster(projectName,
+                LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
       }
     } else if (
         org.pmiops.workbench.notebooks.model.ClusterStatus.RUNNING.equals(fcCluster.getStatus()) &&
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -128,18 +128,21 @@ public class ClusterController implements ClusterApiDelegate {
 
   @Override
   public ResponseEntity<ClusterListResponse> listClusters(String billingProjectId) {
+    if (billingProjectId == null) {
+      throw new FailedPreconditionException("Must specify billing project");
+    }
+
+    // future-proofing the transition from using free-tier projects to using the billing buffer:
+    // verify that the project has been properly initialized if it's the user's free tier project.
+    // billing buffer projects are guaranteed to be initialized at this point.
+
     User user = this.userProvider.get();
-    // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
-    // until then: continue to check the user's free tier project status
-    // because that is the only project we'll pass in here
-    if (!user.getFreeTierBillingProjectName().equals(billingProjectId)) {
+    if (billingProjectId.equals(user.getFreeTierBillingProjectName()) &&
+            user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
       throw new FailedPreconditionException(
-          "NOT YET IMPLEMENTED: Cannot list clusters using arbitrary project name");
+              "User billing project is not yet initialized, cannot list/create clusters");
     }
-    if (user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
-      throw new FailedPreconditionException(
-          "User billing project is not yet initialized, cannot list/create clusters");
-    }
+
     org.pmiops.workbench.notebooks.model.Cluster fcCluster;
     try {
       fcCluster = this.leonardoNotebooksClient.getCluster(billingProjectId, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -127,12 +127,12 @@ public class ClusterController implements ClusterApiDelegate {
   }
 
   @Override
-  public ResponseEntity<ClusterListResponse> listClusters(String projectName) {
+  public ResponseEntity<ClusterListResponse> listClusters(String billingProjectId) {
     User user = this.userProvider.get();
     // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
     // until then: continue to check the user's free tier project status
     // because that is the only project we'll pass in here
-    if (!user.getFreeTierBillingProjectName().equals(projectName)) {
+    if (!user.getFreeTierBillingProjectName().equals(billingProjectId)) {
       throw new FailedPreconditionException(
           "NOT YET IMPLEMENTED: Cannot list clusters using arbitrary project name");
     }
@@ -142,11 +142,9 @@ public class ClusterController implements ClusterApiDelegate {
     }
     org.pmiops.workbench.notebooks.model.Cluster fcCluster;
     try {
-      fcCluster = this.leonardoNotebooksClient.getCluster(projectName,
-              LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+      fcCluster = this.leonardoNotebooksClient.getCluster(billingProjectId, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
     } catch (NotFoundException e) {
-      fcCluster = this.leonardoNotebooksClient.createCluster(projectName,
-              LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+      fcCluster = this.leonardoNotebooksClient.createCluster(billingProjectId, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
     }
 
     int retries = Optional.ofNullable(user.getClusterCreateRetries()).orElse(0);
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -127,19 +127,26 @@ public void setUserService(UserService userService) {
   }
 
   @Override
-  public ResponseEntity<ClusterListResponse> listClusters() {
+  public ResponseEntity<ClusterListResponse> listClusters(String projectName) {
     User user = this.userProvider.get();
+    // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
```
jmthibault79: My interpretation of RW-2476 is that implementing arbitrary billing project creation, configuration, and monitoring is not in scope, implying that no functional change is desired here.

Please let me know if you disagree!

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -330,12 +330,12 @@ private User initializeUserIfNeeded() {
             log.log(Level.INFO, "Failed to remove user from errored billing project");
           }
           String billingProjectName = createFirecloudBillingProject(user);
-          return this.userService.setBillingProjectNameAndStatus(billingProjectName, BillingProjectStatus.PENDING);
+          return this.userService.setFreeTierBillingProjectNameAndStatus(billingProjectName, BillingProjectStatus.PENDING);
```
jmthibault79: Making explicit that this is not applicable to arbitrary billing projects

api/src/main/resources/workbench.yaml
```diff
@@ -501,18 +501,24 @@ paths:
 
   # Notebook clusters ####################################################################
 
-  /v1/clusters:
+  /v1/clusters/{clusterNamespace}:
```
jmthibault79: Following the existing naming convention of referring to billing projects in the swagger.  Should we rename this?

api/src/test/java/org/pmiops/workbench/api/AuditControllerTest.java
```diff
@@ -1,7 +1,7 @@
 package org.pmiops.workbench.api;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.mockito.Matchers.any;
+import static org.mockito.ArgumentMatchers.any;
```
jmthibault79: `Matchers` looks like this:
```
@Deprecated
public class Matchers extends ArgumentMatchers { }
```

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -130,7 +130,7 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     this.jupyterLabMode = jupyterLabMode === 'true';
     this.setNotebookNames();
 
-    this.loadingSub = this.clusterService.listClusters()
+    this.loadingSub = this.clusterService.listClusters(this.wsNamespace)
```
ericsong: we have access to a `cluster` model in this class which contains `clusterNamespace`. Based on the API naming, I would guess that `cluster.clusterNamespace` is the right value to use here. Do they just happen to the same in our business logic?

jmthibault79: That is the better choice.  Thanks.

jmthibault79:
Note: this is currently untested.  I won't merge until I have done so.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -127,19 +127,26 @@ public void setUserService(UserService userService) {
   }
 
   @Override
-  public ResponseEntity<ClusterListResponse> listClusters() {
+  public ResponseEntity<ClusterListResponse> listClusters(String projectName) {
```
gjuggler: Naming nit: we should be as precise & consistent with naming as possible here, since the "billing project" concept is now spread across AoU, Terra, and GCP. It may not be reasonable to do too much in this PR, but I think it's an important long-term goal for us to be consistent.

What we're passing through here is an identifier for the GCP project, which Terra generally calls the billing project. GCP calls this the "project ID":

Screen Shot 2019-05-10 at 10 55 19 AM

I'm not sure whether Terra refers to the project ID as the "billing project name" or the "billing project ID". But "projectName" is probably not right. This is a globally unique ID that's managed by Google, so we should give it a name that signals that. (And if we someday want to show the separate human-editable GCP project name, then we won't run into mass confusion.)

A few suggested options:
 - cloudProjectId
 - projectId
 - billingProjectId
 - something else?

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -127,19 +127,26 @@ public void setUserService(UserService userService) {
...
+    }
     if (user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
       throw new FailedPreconditionException(
           "User billing project is not yet initialized, cannot list/create clusters");
     }
-    String project = user.getFreeTierBillingProjectName();
     org.pmiops.workbench.notebooks.model.Cluster fcCluster;
     try {
-      fcCluster = this.leonardoNotebooksClient.getCluster(project, LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME);
+      fcCluster = this.leonardoNotebooksClient.getCluster(projectName,
```
gjuggler: These lines look like auto-format churn. I don't recall for sure, but maybe our team uses 100 char width for Java files and your editor is set to 80?

TBH, I don't usually worry too much about formatting issues like this. We should eventually set up a clang-format commit-hook which frees all of us from ever having to think about formatting again.

api/src/main/resources/workbench.yaml
gjuggler: See my note above re: naming. I think we should aim to align on a single way to refer to GCP project IDs and use that consistently throughout the project. ClusterNamespace just confuses things IMO, since it's another word referring to the same thing.

api/src/main/resources/workbench.yaml
```diff
@@ -501,18 +501,24 @@ paths:
...
+        and this cluster should always either exist or be in the process of being
+        initialized. In a future where researchers have more control over cluster
+        creation, this endpoint would be extended to return all clusters.
       operationId: listClusters
       tags:
         - cluster
+      parameters:
+        - in: path
+          name: clusterNamespace
+          description: clusterNamespace
```
gjuggler: Whatever we end up landing on for naming of this parameter, please include a slightly longer description than a re-statement of the param name :)

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -225,7 +226,7 @@ private JSONObject dataUriToJson(String dataUri) {
   public void testListClusters() throws Exception {
     when(notebookService.getCluster(WORKSPACE_NS, "all-of-us")).thenReturn(testFcCluster);
 
-    assertThat(clusterController.listClusters().getBody().getDefaultCluster())
+    assertThat(clusterController.listClusters(WORKSPACE_NS).getBody().getDefaultCluster())
```
gjuggler: Same as above: consider renaming this to CLOUD_PROJECT_ID or whatever we decide on as the canonical term for billing project / GCP project IDs.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
calbach: +1 to reaching consistency. projectName is at least partially consistent though - here are the various names for this thing:

- Firecloud: projectName - https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/resources/swagger/api-docs.yaml#L2055
- Leo: googleProject - https://github.com/all-of-us/workbench/blob/48cab8e66c723be603c9931527c82aa39e0dba16/api/src/main/resources/notebooks.yaml#L122
- Firecloud again: arguably a different concept which happens to be the same value "workspaceNamespace" - https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/resources/swagger/api-docs.yaml#L2473
- Our code: mix of projectId, workspaceNamespace, projectName, maybe others

One thing that is critical here would be to understand whether the "projectName" is documented by Firecloud as being the same thing as a Google project ID - or if that's just an implementation detail.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
calbach: Agreed the things listed here are out of scope, but to slightly future proof this and avoid having to revisit this in the next sprint, I would suggest instead:

```
if (user.getFreeTierBillingProjectName() != null &&
    user.getFreeTierBillingProjectName().equals(projectName) &&
    user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
  throw new FailedPreconditionException(
          "User billing project is not yet initialized, cannot list/create clusters");
}
```

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
jmthibault79: I like `billingProjectId` here

ericsong: 👍 this has been a source of confusion for me. I think it would be good to take some time and decide on how we want to name things

calbach: Plugging the relevant tech debt ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-1806

jmthibault79: Heh.  This sounds awfully familiar to me -> https://broadinstitute.atlassian.net/browse/GAWB-41

For those without access, that's a FireCloud tech debt ticket I filed in 2015 which we never acted on:

"I learned that a Billing Project is the same as a Workspace Namespace.  I knew that both represented a Google Project, but it was not clear to me that they were identical.  Additionally, we have the concept of a Project ID in GoogleServicesDAO, which is the same value yet again.

We should standardize on a single term to avoid confusion, and do a mass renaming of all the others in Rawls."

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -132,11 +132,9 @@ public void setUserService(UserService userService) {
     // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
     // until then: continue to check the user's free tier project status
     // because that is the only project we'll pass in here
-    if (!user.getFreeTierBillingProjectName().equals(billingProjectId)) {
-      throw new FailedPreconditionException(
-          "NOT YET IMPLEMENTED: Cannot list clusters using arbitrary project name");
-    }
-    if (user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
+    if (user.getFreeTierBillingProjectName() == null ||
```
calbach: This was not quite my suggestion, logically here is what I'm proposing:

- billing project null --> continue (this supports a future where new users lack a free tier billing project)
- billing project != freeTier --> continue (this supports a future where an older user creates a new workspace with its own billing project)
- billing project == freeTier && notReady --> error (existing logic)

jmthibault79: Sorry - this was not clear to me.  I'm having trouble thinking about these future states and will follow up with you in a Googledoc.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -128,16 +128,25 @@ public void setUserService(UserService userService) {
...
     User user = this.userProvider.get();
-    // TODO as part of Billing Project epic RW-1205: enable status checking for arbitrary project
-    // until then: continue to check the user's free tier project status
-    // because that is the only project we'll pass in here
-    if (user.getFreeTierBillingProjectName() == null ||
-            ! user.getFreeTierBillingProjectName().equals(billingProjectId) ||
-            user.getFreeTierBillingProjectStatusEnum() != BillingProjectStatus.READY) {
-      throw new FailedPreconditionException(
-          "User billing project is not yet initialized, cannot list/create clusters");
+    if (user.getFreeTierBillingProjectName() != null) {
```
calbach: I would not override the provided billingProjectId in any situation, I would only run the below status check if the incoming billingProjectId == freeTierBillingProjectName.

Here are the migration events going forwards:
t0. All users have a free tier billing project, all clusters are created within each user's project
t1. Enable 1 project == 1 workspace - all current users still have a free tier billing project, any newly created users don't going forward. All clusters should be created in the billing project associated with the relevant workspace, which may or may not be the free tier billing project.
t2. All workspaces migrated to 1 project == 1 workspace. Older users still have freeTierBillingProjectName set, but it isn't used for anything. Clusters now exclusively get created in new per-workspace projects.
t3. Teardown fields / code for freeTierBillingProjectName/Status, now no users have this.

In all cases, we should just respect the incoming billingProject. IF it is a freeTierBillingProject, we can also check the status since we know it.

ui/src/app/views/notebook-redirect/component.ts
```diff
@@ -130,7 +130,10 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     this.jupyterLabMode = jupyterLabMode === 'true';
     this.setNotebookNames();
 
-    this.loadingSub = this.clusterService.listClusters()
+    this.loadingSub = userProfileStore.asObservable()
```
jmthibault79: Need to Observe here because `freeTierBillingProjectName` is not yet populated on initial load

ui/src/app/views/settings/component.tsx
```diff
@@ -46,7 +47,12 @@ export class SettingsReact extends React.Component<{}, SettingsState> {
   }
 
   componentDidMount() {
-    this.pollCluster();
+    userProfileStore.asObservable().subscribe((profileStore) => {
```
jmthibault79: Need to Observe here because `freeTierBillingProjectName` is not yet populated on initial load

ui/src/app/views/settings/component.tsx
```diff
@@ -25,6 +26,7 @@ export interface SettingsState {
   resetClusterPending: boolean;
   resetClusterModal: boolean;
   clusterDeletionFailure: boolean;
+  waitingForUserProfile: boolean;
```
jmthibault79: Only used to select more appropriate tooltip text

ui/src/app/views/settings/component.tsx
```diff
@@ -61,8 +70,8 @@ export class SettingsReact extends React.Component<{}, SettingsState> {
       <h3>Notebooks</h3>
       <div style={styles.notebookSettings}>
         <TooltipTrigger content={
-                          (!this.state.cluster) ?
-                            'Your notebook server is still being created' : undefined}
+          this.state.waitingForUserProfile ? 'Waiting for notebook server status' :
```
jmthibault79: We haven't asked Leo about the cluster yet, so it might be ready - IMO it's inappropriate to claim that it's still being created

jmthibault79:
Since this PR has substantially changed from your original approvals, most notably in the fact that it:
* implements parameter checking correctly
* passes all tests
* no longer breaks multiple screens of the UI
I would appreciate a re-review from one or more of you @calbach @ericsong @gjuggler 

Thanks! :)

ui/src/app/views/settings/component.tsx
calbach: +1 thanks for improving this

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -223,35 +226,57 @@ private JSONObject dataUriToJson(String dataUri) {
...
         testFcCluster.status(null));
 
-    assertThat(clusterController.listClusters().getBody().getDefaultCluster().getStatus())
+    assertThat(clusterController.listClusters(BILLING_PROJECT_ID).getBody().getDefaultCluster().getStatus())
         .isEqualTo(ClusterStatus.UNKNOWN);
   }
 
+  @Test(expected = FailedPreconditionException.class)
+  public void testListClustersNullBillingProject() throws Exception {
+    when(notebookService.getCluster(BILLING_PROJECT_ID, "all-of-us")).thenReturn(testFcCluster);
```
ericsong: unnecessary mock?

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -223,35 +226,57 @@ private JSONObject dataUriToJson(String dataUri) {
...
+  public void testListClustersFreeTierNotReady() throws Exception {
+    when(notebookService.getCluster(BILLING_PROJECT_ID, "all-of-us")).thenReturn(testFcCluster);
+
+    User notReadyUser = new User();
+    notReadyUser.setEmail(LOGGED_IN_USER_EMAIL);
+    notReadyUser.setUserId(123L);
+    notReadyUser.setFreeTierBillingProjectName(BILLING_PROJECT_ID);
+    notReadyUser.setFreeTierBillingProjectStatusEnum(BillingProjectStatus.PENDING);
+
+    when(userProvider.get()).thenReturn(notReadyUser);
```
ericsong: i try to use the `doReturn` syntax everywhere. covers a few more cases than when - https://stackoverflow.com/a/20360269

jmthibault79: Interesting, thanks!

Looks like we use both patterns frequently.  This class uses `when`/`thenReturn` exclusively so in this case I'll keep it for internal consistency.

PR #2108, opened 2019-05-09
[RISK:NO][RW:2397]  Test for select values

NehaBroad:


ui/src/app/views/dataset-page/component.spec.tsx
```diff
@@ -51,5 +52,56 @@ describe('DataSet', () => {
       .toBe(exampleCohortStubs.length);
   });
 
+  it('should display views based on Domain of Concept selected in workspace', async() => {
```
blrubenstein: Nit: 'values' not views?

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -83,7 +83,8 @@ export const ValueListItem: React.FunctionComponent <
   {domainValue: DomainValue, onSelect: Function, checked: boolean}> =
   ({domainValue, onSelect, checked}) => {
     return <div style={{display: 'flex', color: 'black', height: '1.2rem'}}>
-      <input type='checkbox' value={domainValue.value} onChange={() => onSelect()}
+      <input type='checkbox' value={domainValue.value} onClick={() => onSelect()}
```
blrubenstein: You shouldn't need to add a click handler for this. Just, above, in the tests, pass 'change' in as the arg to simulate?

NehaBroad: I was actually hoping to find solution for input type checkbox to remove onChange, technically change would work but since its a click event i was trying to let the test be. I am going to try some other thing if that will not work i will revert this and change test to have "change" rather than click

NehaBroad: Ok so far what i have read since onChange and onClick behaves different only for IE i'll revert this and use 'change' in test

PR #2111, opened 2019-05-10
[RW-2618][RISK=NO]Skin Preview Data Set

NehaBroad:
preview data table will be loaded  Once cohort , concept and view are selected.

After that the user will have to click the refresh button to refresh the table,
Screen Shot 2019-05-10 at 4 30 04 PM
Screen Shot 2019-05-10 at 4 30 15 PM


ui/src/app/views/dataset-page/component.tsx
```diff
@@ -545,16 +573,24 @@ const DataSetPage = withCurrentWorkspace()(class extends React.Component<Props,
         <div style={{backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
           <div style={{...styles.selectBoxHeader, display: 'flex', flexDirection: 'row',
             position: 'relative'}}>
+            <div style={styles.previewDataHeader}>
+              Preview Data Set
+            </div>
+            {!defaultPreviewView && <Clickable data-test-id='preview-icon'
+                                               onClick={() => this.getPreviewList()}
+                                               style={styles.refreshIcon}>
+              <ClrIcon style={{fill: colors.gray[7], marginLeft: '0.1rem', marginTop: '-1rem'}}
```
blrubenstein: Can you use align-self rather than a negative marginTop? Or maybe the clickable can do it? Just want to make sure we are using relative rather than absolute styling when we can.

ui/src/app/styles/colors.ts
```diff
@@ -15,7 +15,8 @@ export default {
     '#85bde4',
     '#afd3ed',
     '#d7eaf6',
-    '#eaf4fb'
+    '#eaf4fb',
+    '#2F2E7E'
```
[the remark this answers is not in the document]
NehaBroad: I really try to avoid changing the order of colors that are being referred from multiple files and has high chances to create issues which are difficult to catch. We can talk more about this offline

ui/src/app/views/dataset-page/component.tsx
NehaBroad: I was not able to align the actual icon to the middle of the button. I tried align-items and vertical align. We can take a look at it together?

calbach:
Driveby - **[risk=no]**

PR #2112, opened 2019-05-10
[RW-2619][Risk=low] Remove workspace storage service.

blrubenstein:
@dolbeew to make sure I didn't murder cohort builder with this change...

@calbach for approval on the removal of workspacestorageservice.


ui/src/app/views/workspace-edit/component.spec.ts
```diff
@@ -1,58 +1,59 @@
-import {fakeAsync, TestBed} from '@angular/core/testing';
```
blrubenstein: Keeping this file per Neha's request, so we know what tests we need to write for the React version.

ui/src/app/views/workspace-wrapper/component.ts
```diff
@@ -69,7 +66,18 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
         // do not render the child component with a stale workspace.
         this.workspace = undefined;
         this.accessLevel = undefined;
-        return this.workspaceStorageService.getWorkspace(ns, wsid);
+        // This needs to happen for testing. Otherwise it tries to make an api
```
calbach: I hit this as well. Seeding the urlParamSTore with `{}` in the tests is the root of the issue - we may need a better story there eventually. I wouldn't say we *need* to deal with the issue in production code (but certainly this is more convenient).

ui/src/app/views/workspace-wrapper/component.ts
```diff
@@ -69,7 +66,18 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
...
+        // call with undefined, because the component initializes before we have
+        // access to the route.
+        if (ns === undefined || wsid === undefined) {
+          return Promise.resolve({name: '', accessLevel: WorkspaceAccessLevel.NOACCESS});
```
calbach: Returning a stub workspace seems problematic here. I might recommend just returning `Observable.empty();` which would not trigger the next callback.

ui/src/app/views/workspace-wrapper/component.ts
blrubenstein: I will clarify the comment.

ui/src/app/views/workspace-wrapper/component.ts
blrubenstein: Agreed that it could be problematic. However `Observable.empty()` did actually trigger the next callback. I am going to try adding an empty object check in the subscribe, and returning an empty object.

calbach: OK, that didn't work then. I'd just use null instead. I think passing `Observable.empty()` as a sentinel value would be pretty confusing.

PR #2115, opened 2019-05-13
[RW-2622][risk=no] Add consume project to Billing Project Buffer API

ericsong:
I ended up going with a MySQL lock after trying out transactions and table locks. I ruled out transactions since I believe they lock on a row by row basis and I needed something that locked on the table level. I chose a generic lock over table locks since I just needed the consume code to run in sequence. The consume code doesn't have to block other updates to the table.

Had to do some funky stuff to be able to test concurrency. I'm not super happy with what came out of it since there is manual clean up that needs to be done to prevent side effects and the underlying locking implementation had to be stubbed out but it works.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -42,8 +42,8 @@ public class BillingProjectBufferService {
       Provider<WorkbenchConfig> workbenchConfigProvider) {
     this.billingProjectBufferEntryDao = billingProjectBufferEntryDao;
     this.clock = clock;
-    this.workbenchConfigProvider = workbenchConfigProvider;
     this.fireCloudService = fireCloudService;
+    this.workbenchConfigProvider = workbenchConfigProvider;
   }
 
   public void bufferBillingProject() {
@@ -111,12 +111,17 @@ public class BillingProjectBufferService {
   private BillingProjectBufferEntry consumeBufferEntryForAssignment() {
     while (billingProjectBufferEntryDao.acquireAssigningLock() != 1) {}
 
-    BillingProjectBufferEntry entry = billingProjectBufferEntryDao.findFirstByStatusOrderByCreationTimeAsc(
-        StorageEnums.billingProjectBufferStatusToStorage(AVAILABLE));
-    entry.setStatusEnum(ASSIGNING);
-    billingProjectBufferEntryDao.save(entry);
+    BillingProjectBufferEntry entry;
+    try {
+      entry = billingProjectBufferEntryDao
+          .findFirstByStatusOrderByCreationTimeAsc(
+              StorageEnums.billingProjectBufferStatusToStorage(AVAILABLE));
+      entry.setStatusEnum(ASSIGNING);
+      billingProjectBufferEntryDao.save(entry);
+    } finally {
+      billingProjectBufferEntryDao.releaseAssigningLock();
+    }
 
-    billingProjectBufferEntryDao.releaseAssigningLock();
     return entry;
   }
 
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -30,10 +30,10 @@ public class BillingProjectBufferService {
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   private static final Logger log = Logger.getLogger(BillingProjectBufferService.class.getName());
 
-  private final BillingProjectBufferEntryDao billingProjectBufferEntryDao;
-  private final Clock clock;
-  private final FireCloudService fireCloudService;
-  private final Provider<WorkbenchConfig> workbenchConfigProvider;
+  private BillingProjectBufferEntryDao billingProjectBufferEntryDao;
+  private Clock clock;
+  private FireCloudService fireCloudService;
+  private Provider<WorkbenchConfig> workbenchConfigProvider;
 
   @Autowired
   public BillingProjectBufferService(BillingProjectBufferEntryDao billingProjectBufferEntryDao,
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -27,10 +30,10 @@
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   private static final Logger log = Logger.getLogger(BillingProjectBufferService.class.getName());
 
-  private final BillingProjectBufferEntryDao billingProjectBufferEntryDao;
-  private final Clock clock;
-  private final FireCloudService fireCloudService;
-  private final Provider<WorkbenchConfig> workbenchConfigProvider;
+  private BillingProjectBufferEntryDao billingProjectBufferEntryDao;
```
calbach: Why drop final here?

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -8,12 +8,22 @@
 @Repository
 public interface BillingProjectBufferEntryDao  extends CrudRepository<BillingProjectBufferEntry, Long> {
 
+  String ASSIGNING_LOCK = "ASSIGNING_LOCK";
```
calbach: nit: private?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -94,6 +97,34 @@ public void syncBillingProjectStatus() {
...
+    fireCloudService.addUserToBillingProject(user.getEmail(), entry.getFireCloudProjectName());
+    entry.setStatusEnum(ASSIGNED);
+    entry.setAssignedUser(user);
+    billingProjectBufferEntryDao.save(entry);
+
+    return entry;
+  }
+
+  private BillingProjectBufferEntry consumeBufferEntryForAssignment() {
+    while (billingProjectBufferEntryDao.acquireAssigningLock() != 1) {}
```
calbach: A tight retry loop here is a bit scary. What do you expect the behavior to be if the lock is currently held?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -94,6 +97,34 @@ public void syncBillingProjectStatus() {
...
+
+    BillingProjectBufferEntry entry;
+    try {
+      entry = billingProjectBufferEntryDao
+          .findFirstByStatusOrderByCreationTimeAsc(
+              StorageEnums.billingProjectBufferStatusToStorage(AVAILABLE));
+      entry.setStatusEnum(ASSIGNING);
+      billingProjectBufferEntryDao.save(entry);
+    } finally {
+      billingProjectBufferEntryDao.releaseAssigningLock();
```
calbach: What are the properties of the lock, e.g. if this call fails, or the process somehow crashes before this line can execute?

api/src/test/java/org/pmiops/workbench/TestLock.java
```diff
@@ -0,0 +1,24 @@
+package org.pmiops.workbench;
+
+public class TestLock {
```
calbach: Should there be some kind of synchronization on this test lock?

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -284,4 +323,75 @@ public void syncBillingProjectStatus_iteratePastStallingRequests() {
...
+    callableTasks.add(t1);
+    callableTasks.add(t2);
+
+    ExecutorService executor = Executors.newFixedThreadPool(2);
+    List<Future<BillingProjectBufferEntry>> futures = executor.invokeAll(callableTasks);
+
+    assertThat(futures.get(0).get().getFireCloudProjectName())
+        .isNotEqualTo(futures.get(1).get().getFireCloudProjectName());
+
+    // Need to clean up entities since they actually get committed to the database because we're not running within a transaction
```
calbach: This is the test annotation that handles cleanup:

```
@DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD)
```

Also I don't see Propagation.NOT_SUPPORTED on this test - is it really not being cleaned up or was this comment from an earlier version?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: I'll revert it back but for context, there is a bug with our version of Spring and App Engine where adding `@Transactional` to a method required a default constructor which is why I needed to drop `final`. It has something to do with its usage of Objenesis to create the proxy methods to support declarative transactions.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: the underlying `GET_LOCK` call will timeout after 1 second, after which it will try again until it succeeds. So long as the release gets locked at some point, this loop should be able to acquire it when its free

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: If the process is alive, it should be able to reach this line even if an exception is thrown since its within a `finally` block. If the process crashes, MySQL should let go of the lock since the session is dead.

I did not think of the case where the release call just fails. I can add a retry loop like the one above that keeps getting called until it succeeds.

maybe with a max retry counter

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
ericsong: it looks like `private` modifier is not supported in interfaces for Java8

api/src/test/java/org/pmiops/workbench/TestLock.java
ericsong: good catch

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
ericsong: I added it as an annotation.

The dirties context annotation also works! Thanks for the tip

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
calbach: Could you add a comment here which mentions the 1s delay

ericsong:
@calbach  that should cover everything

PR #2118, opened 2019-05-13
[RW-2271][RW-2297][risk=no] concept set details react

jacmrob:


ui/src/app/views/concept-set-details/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+        selectedConcepts: []
+      };
+    }
+
+    componentDidMount() {
+      this.getConceptSet();
+    }
+
+    async getConceptSet() {
+      const {urlParams: {ns, wsid, csid}} = this.props;
```
gjuggler: I think I made this comment to Scout on some non-frontend PR once, but is there a strong reason to use such short variable name abbreviations here? It seems like in this case they're coming directly from the urlParams props, so destructuring into a local variable is super convenient.

But on the other hand, "ns" isn't exactly clear to someone looking at this code for the first time. I'd almost always prefer a real word, even if slightly shortened.

I suspect this is a pretty strong pattern throughout the codebase, so it's not worth breaking consistency to change it here – I just wanted to note that this feels slightly out of place with respect to how we name variables elsewhere.

ui/src/app/views/concept-set-details/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+      const {conceptSet, editName, editDescription} = this.state;
+      try {
+        this.setState({editSaving: true});
+        await conceptSetsApi().updateConceptSet(ns, wsid, csid,
+          {...conceptSet, name: editName, description: editDescription});
+        await this.getConceptSet();
+        this.setState({editing: false, editSaving: false});
+      } catch (error) {
+        console.log(error);
+        this.setState({editing: false, editSaving: false});
```
gjuggler: This feels a little odd -- if an error occurs, we do the same thing as if it succeeds? I'd almost rather not change state, and leave the same TODO here as in L118, e.g. we should come back to this & clean up our error-handling strategy en masse.

Alternatively: could you adopt the same approach as L152 over here?

ui/src/app/views/concept-set-details/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+    async onDeleteConceptSet() {
+      const {urlParams: {ns, wsid, csid}} = this.props;
+      try {
+        await conceptSetsApi().deleteConceptSet(ns, wsid, csid);
+        navigate(['workspaces', ns, wsid, 'concepts', 'sets']);
+      } catch (error) {
+        console.log(error);
+        this.setState({error: true,
+          errorMessage: 'Could not concept set \'' + this.state.conceptSet.name + '\''});
```
gjuggler: nit: "could not delete"

ui/src/app/views/concept-set-details/component.tsx
```diff
@@ -0,0 +1,321 @@
...
+        console.log(error);
+        this.setState({editing: false, editSaving: false});
+      }
+    }
+
+    onSelectConcepts(concepts) {
+      this.setState({selectedConcepts: concepts});
+    }
+
+    async onDeleteConcepts() {
```
gjuggler: Wording nitpick: maybe we should call this "onRemoveConcepts" since we're not actually deleting them in any real sense? The user-facing copy on L152 might also make more sense with "remove"

ui/src/app/views/concept-set-details/component.tsx
jacmrob: yes i think the state-setting should be in a `finally` block

ui/src/app/views/concept-set-details/component.tsx
jacmrob: good thought

PR #2122, opened 2019-05-14
[RW-2425][risk=no] Prevent read-only users from creating reviews

dolbeew:
If a user has READER access for a workspace and clicks on a cohort that does not yet have a review, show a modal with a message and redirect back to the cohort list


ui/src/app/cohort-review/page-layout/page-layout.html
```diff
@@ -2,9 +2,23 @@
...
 -->
-<div class="cohort-review-wrapper" *ngIf="cohortLoaded && reviewPresent === true">
+<div class="cohort-review-wrapper" *ngIf="cohortLoaded && reviewPresent">
   <router-outlet></router-outlet>
 </div>
 
-<app-create-review-page *ngIf="cohortLoaded && reviewPresent === false" (created)="reviewCreated($event)"></app-create-review-page>
-<div *ngIf="!cohortLoaded || reviewPresent === undefined" style="position: absolute; top: 50%; left: 50%;" class="spinner"></div>
+<app-create-review-page *ngIf="cohortLoaded && !reviewPresent && !readonly" (created)="reviewCreated($event)"></app-create-review-page>
+<clr-modal [(clrModalOpen)]="cohortLoaded && !reviewPresent && readonly" [clrModalSize]="'md'" [clrModalClosable]="false">
```
freemabd: would it be easier to understand if we wrap this logic in a function? Maybe call it something like permissionToReview()? or allowedToReview()

dolbeew: Good call, done.

PR #2125, opened 2019-05-14
[RW-2574][RISK=LOW] Create "Select All" for Values for Data Set Builder

NehaBroad:
This pr also has solution for 500 on selecting include all participants


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -228,12 +233,12 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
       queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
 
       TableResult queryResponse = bigQueryService.executeQuery(bigQueryService
-          .filterBigQueryConfig(queryJobConfiguration));
+          .filterBigQueryConfig(queryJobConfiguration), 42000l);
```
blrubenstein: Curious why 42000, mind adding a comment or something?

calbach:
Please use `=` rather than `:` in the PR title. The release ticket script will not accept `:`s

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
NehaBroad: Have updated it to a minute, Select All can sometimes take up to a minute to get the result. I have added a comment however i would want to change the logic to add retries i.e if it time out call the query again with less number of rows.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -224,30 +229,66 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+              previewValue.setQueryValue(queryValues);
+            }
+          });
+          retry = 2;
+        } catch (ServerErrorException ex) {
+          if(ex.getCause().getMessage().contains("Read timed out")) {
+            rowsRequested = (rowsRequested / 2);
+            retry++;
+          } else {
+            retry = 2;
```
blrubenstein: I might throw an exception here rather than setting retry = 2? Maybe the same exception.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -224,30 +229,66 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+                  Double fieldValue = Double.parseDouble(value);
+                  queryValues.add(dateFormat.format(new Date(fieldValue.longValue())));
+                } catch (NumberFormatException ex) {
+                  queryValues.add("");
+                }
+              });
+              previewValue.setQueryValue(queryValues);
+            }
+          });
+          retry = 2;
```
blrubenstein: You may be able to use `break`, rather than setting retry = 2. I think that would be cleaner if we decided to do more or less retries, not needing to change as often.

PR #2131, opened 2019-05-15
[RW-2605][risk=low] Create new workspaces in buffered projects

ericsong:
Pretty straightforward. Most of the changes were actually wrangling the over excessing stubbing we do in WorkspaceControllerTest. I created a more generic stub that mocks the behavior of FireCloudService so we don't have to explicitly stub every call made to it.

I created a follow up story to handle the empty buffer scenario. https://precisionmedicineinitiative.atlassian.net/browse/RW-2651

### DIFF
--- a/api/config/config_test.json
+++ b/api/config/config_test.json
@@ -62,5 +62,8 @@
   },
   "cohortbuilder": {
     "enableListSearch": false
+  },
+  "featureFlags": {
+    "useBillingProjectBuffer": true
   }
 }
--- a/api/config/config_test.json
+++ b/api/config/config_test.json
@@ -64,6 +64,6 @@
     "enableListSearch": false
   },
   "featureFlags": {
-    "useBillingProjectBuffer": true
+    "useBillingProjectBuffer": false
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -81,7 +81,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
   private final Clock clock;
   private final NotebooksService notebooksService;
   private final UserService userService;
-  private final Provider<WorkbenchConfig> workbenchConfig;
+  private final Provider<WorkbenchConfig> workbenchConfigProvider;
 
   @Autowired
   WorkspacesController(
@@ -99,7 +99,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
       Clock clock,
       NotebooksService notebooksService,
       UserService userService,
-      Provider<WorkbenchConfig> workbenchConfig) {
+      Provider<WorkbenchConfig> workbenchConfigProvider) {
     this.billingProjectBufferService = billingProjectBufferService;
     this.workspaceService = workspaceService;
     this.workspaceMapper = workspaceMapper;
@@ -114,7 +114,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
     this.clock = clock;
     this.notebooksService = notebooksService;
     this.userService = userService;
-    this.workbenchConfig = workbenchConfig;
+    this.workbenchConfigProvider = workbenchConfigProvider;
   }
 
   @VisibleForTesting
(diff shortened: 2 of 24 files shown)


api/config/config_test.json
```diff
@@ -62,5 +62,8 @@
   },
   "cohortbuilder": {
     "enableListSearch": false
+  },
+  "featureFlags": {
+    "useBillingProjectBuffer": true
```
calbach: I don't think we're ready to enable this yet. There's a bunch of free tier billing project + cluster + UI tickets which would need to be resolved first. I will try to create a ticket which maps out all the dependencies.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -77,9 +81,11 @@
   private final Clock clock;
   private final NotebooksService notebooksService;
   private final UserService userService;
+  private final Provider<WorkbenchConfig> workbenchConfig;
```
calbach: nit: name should probably inclue `Provider`

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -171,31 +178,52 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     } else if (workspace.getName().length() > 80) {
       throw new BadRequestException("Workspace name must be 80 characters or less");
     }
+
+    final boolean useBillingProjectBuffer = workbenchConfig.get().featureFlags.useBillingProjectBuffer;
+    if (useBillingProjectBuffer) {
```
calbach: Shouldn't this be a `!`?

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -171,31 +178,52 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-      } catch (ConflictException e) {
-        if (attemptValue >= 5) {
-          throw e;
-        } else {
-          fcWorkspaceId =
-              new FirecloudWorkspaceId(workspaceId.getWorkspaceNamespace(),
-                  workspaceId.getWorkspaceName() + Integer.toString(attemptValue));
+    org.pmiops.workbench.firecloud.model.Workspace fcWorkspace;
+
+    if (useBillingProjectBuffer) {
```
calbach: Why fork here? I think the existing retry logic is sketchy, but we probably want it equally in either case.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -171,31 +178,52 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+      fcWorkspace = attemptFirecloudWorkspaceCreation(fcWorkspaceId);
```
calbach: I missed this in the previous review. There's a question of whether we should consider a billing project to be "assigned" before or after we add the user to the project.

If we fail on this line, we've assigned the project but it's effectively an orphan which we'd ideally clean up later.

But I think since that update could fail as well - I'm convincing myself that the way you have it now is probably ideal. We could also have another infrequent cron which cleans up any billing projects in this state (assigned, but don't have any associated workspace).

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
ericsong: I felt like this code doesn't apply anymore once the bp == workspace rule is in place. It seems to be guarding against the possibility of duplicate workspace names within a billing project but that should no longer be a problem since every billing project at this point should be unique once its taken off of the buffer.

If the retry was on `Exception` I might have kept it around but its only handling the conflict case

calbach: Good point - I overlooked that this was just for conflict handling. Makes sense to fork.

Thinking more - there's actually an argument here that we use a constant for all firecloud workspace IDs. I think I prefer the way you have it now though.

PR #2132, opened 2019-05-15
[RW-2537][Risk=low] Update data set

blrubenstein:


api/db/changelog/db.changelog-78-data-set-etag.xml
```diff
@@ -0,0 +1,14 @@
...
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="srubenst" id="db.changelog-76-data-set-last-modified">
+    <addColumn tableName="data_set">
+      <column name="version" type="smallint" defaultValue="1">
+        <constraints nullable="false"/>
+      </column>
+    </addColumn>
+  </changeSet>
+</databaseChangeLog>
```
NehaBroad: nit: eof line

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -352,6 +352,44 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+        getDbDataSet(workspaceNamespace, workspaceId, dataSetId, WorkspaceAccessLevel.WRITER);
+    if(Strings.isNullOrEmpty(request.getEtag())) {
+      throw new BadRequestException("missing required update field 'etag'");
+    }
+    int version = Etags.toVersion(request.getEtag());
+    if (dbDataSet.getVersion() != version) {
+      throw new ConflictException("Attempted to modify outdated data set version");
+    }
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    dbDataSet.setLastModifiedTime(now);
```
NehaBroad: I likes the way how attributes are set in data set in **createDataSet** ( eg . DataSet result = new DataSet()
              .name(dataSet.getName())
              .includesAllParticipants(dataSet.getIncludesAllParticipants())....)
Can this be done in the same way

blrubenstein: That only works on chaining off of a constructor. It would theoretically be possible here, but we'd need to add a wrapper method around set that does things the way we want, which would require some DB modeling refactoring.

ui/src/app/views/dataset-page/component.tsx
```diff
@@ -463,7 +494,15 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
                                       onChange={
                                         () => this.select(cohort, ResourceType.COHORT)
                                       }
-                                      edit={
+                                      onEdit={
+                                        () => {
+                                          const url =
+                                            '/workspaces/' + namespace + '/' +
+                                            id + '/cohorts/build?cohortId=';
```
NehaBroad: we can append  cohort.id here since url is not being used anywhere else

PR #2133, opened 2019-05-15
[RW-2651] Show error message to User when buffer is full

ericsong:
More gracefully handle the case where the billing project buffer is empty. 

Returns a 429 instead of a 500 and the client displays a relevant error message. BufferService will also log when it's consumed while empty and throws a more specific error message.

Scoped out refilling the buffer on empty.


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -30,6 +31,8 @@
 import org.pmiops.workbench.db.dao.ConceptSetDao;
 import org.pmiops.workbench.db.dao.UserDao;
 import org.pmiops.workbench.db.dao.UserService;
+import org.pmiops.workbench.exceptions.*;
```
blrubenstein: Nit: I prefer (and I think established convention has been) to import specific exceptions rather than `*`.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -138,11 +144,15 @@ private String createBillingProjectName() {
     return prefix + randomString;
   }
 
+  private int getRemainingBufferSlotsSize() {
+    return getBufferMaxCapacity() - (int) getCurrentBufferSize();
```
blrubenstein: I'd expect `getCurrentBufferSize` to return the number of projects in the buffer right now, not the number of unfilled slots. I understand its an existing thing, but maybe `getCurrentBufferSpace` (or `getCurrentUnfilledBufferSpace`?)

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -3,13 +3,7 @@
...
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
```
blrubenstein: Nit: Prefer specific imports to * imports.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
ericsong: that is what `getCurrentBufferSize()` returns. I really like `getCurrentUnfilledBufferSpace` over `getRemainingBufferSlotsSize`. will make the switch!

PR #2140, opened 2019-05-16
[No Ticket][Risk=no] Remove menu from data set list and refactor to remove code

blrubenstein:
Removes menus on the data set page per Lou and Greg's recommendations, and then refactor and tear things out.

### DIFF
--- a/ui/src/app/components/resources.tsx
+++ b/ui/src/app/components/resources.tsx
@@ -3,22 +3,9 @@ import * as React from 'react';
 import {Clickable, MenuItem} from 'app/components/buttons';
 import {ClrIcon} from 'app/components/icons';
 import {PopupTrigger} from 'app/components/popups';
-import colors from 'app/styles/colors';
 import {switchCase} from 'app/utils';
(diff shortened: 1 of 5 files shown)


ui/src/app/components/resources.tsx
```diff
@@ -3,22 +3,9 @@ import * as React from 'react';
...
-    border: '0.5px solid #C3C3C3', margin: '.4rem',
-    height: '1.5rem', display: 'flex'
-  },
-  listItemCheckbox: {
-    height: 17, width: 17, marginLeft: 10, marginTop: 10,
-    marginRight: 10, backgroundColor: colors.green[1]
-  }
-};
 
 export const ResourceCardMenu: React.FunctionComponent<{
```
blrubenstein: I wasn't sure if this should fold under resource card now, or continue being here. It is currently only used in one place, but I think will probably fold out again in the next few weeks.

calbach:
Could this PR title be made more specific?

blrubenstein:
@calbach done.

PR #2141, opened 2019-05-16
[RW-2308][risk=low] Add snippets menu support and define an update process

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2308

Enables the [snippets menu extension](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/snippets_menu/readme.html) and loads AoU specific code snippets defined [here](https://github.com/all-of-us/workbench-snippets). These snippets will be periodically (manually) synced into the Workbench repo, and then packaged up and deployed via GCS for consumption by the Leo createCluster API (same as our playground extension works).

The Snippets menu customization is done via a method similar to [this](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/snippets_menu/readme.html#advanced-menu-customization). Deviations follow:
- We use a UI extension instead of `custom.js`. `custom.js` is a singleton and is already written by Leo for another purpose. An extension appears to function identically.
- We insert different menus depending on the active Kernel (R vs Python)

See the README for process details on updating these snippets going forward.

CC @jmthibault79 

Screenshot below - clicking on a menu entry inserts code into the current cell:
image


calbach:
Also added a screenshot to the PR description ^

api/cluster-resources/README.md
```diff
@@ -51,7 +51,45 @@ Jupyter UI extension for playground mode. Passed via GCS at cluster creation tim
...
+https://github.com/all-of-us/workbench-snippets. AoU configures Leo clusters as
+follows to enable this:
+
+1. Enable the snippets_menu/main extension in ./setup_notebook_cluster.sh
+1. Deploy a Jupyter UI extension to configure the menu with AoU-specific snippets
+
+## Updating Snippets
+
+As code snippets are updated in the source repository, we will want to
+periocially update the menus in the Workbench. For now, this process is fairly
```
gjuggler: How often do we expect these snippets to change / be updated? We may benefit from fully automating this into our build process sooner than later if it's going to be reasonably often, so the only thing we'd have to do is update a submodule reference.

Maybe file a tracking ticket with some loose notes on a feasible approach?

api/cluster-resources/build.rb
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
```
gjuggler: Nice example how an isolated bit of build logic can be separate from devstart but still well integrated with the build!

api/cluster-resources/README.md
calbach: Fair point - I don't think we know yet how frequent this will be. A submodule is a possible solution, though it is a bit more opaque than having the explicit menu JSON checked into source. I put some thoughts on https://precisionmedicineinitiative.atlassian.net/browse/RW-2665 and noted here.

api/cluster-resources/snippets-menu.js.template
```diff
@@ -0,0 +1,30 @@
...
+  'nbextensions/snippets_menu/main'
+], (Jupyter, snippets_menu) => {
+  // Jupyter UI extension to customize the "Snippets Menu" extension with AoU
+  // specific code snippets. See
+  // https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/snippets_menu/readme.html
+  const load = () => {
+    const pyMenu = {{PY_MENU_JSON}};
+    const rMenu = {{R_MENU_JSON}};
+
+    const {session: {kernel_model: {name: kernelName}}} = Jupyter.notebook;
```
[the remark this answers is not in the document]
calbach: Agreed - done

calbach:
FYI I further changed the snippets JS name, as that apparently impacts how it gets loaded and displayed, this was making me very uncomfortable (_ = standard snippets menu extension, - = our extension which depends on the standard one):

image

PR #2142, opened 2019-05-16
[RW-2609][RW-2608][risk=low] Remove usages of free tier billing project

ericsong:
Delivering RW-2608 "Remove client-side usage of freeTierBillingProject" and RW-2609 "Stop creating free tier billing projects on user creation" together so the app is somewhat usable when the feature flag is turned on.

We have to expose the feature flag on the front end so we can override the UI code that disables workspace creation if the free tier billing project is not ready yet.


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -275,10 +275,12 @@ private User initializeUserIfNeeded() {
     if (user.getFirstSignInTime() == null) {
       // TODO(calbach): After the next DB wipe, switch this null check to
       // instead use the freeTierBillingProjectStatus.
-      if (user.getFreeTierBillingProjectName() == null) {
-        String billingProjectName = createFirecloudUserAndBillingProject(user);
-        user.setFreeTierBillingProjectName(billingProjectName);
-        user.setFreeTierBillingProjectStatusEnum(BillingProjectStatus.PENDING);
+      if (!workbenchConfigProvider.get().featureFlags.useBillingProjectBuffer && user.getFreeTierBillingProjectName() == null) {
+        if (user.getFreeTierBillingProjectName() == null) {
```
calbach: redundant with above if

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -275,10 +275,12 @@ private User initializeUserIfNeeded() {
...
+          String billingProjectName = createFirecloudUserAndBillingProject(user);
```
calbach: We still need the first part of this method call in the new codepath - which registers the user with Firecloud.

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -63,11 +67,17 @@ public UserController(UserService userService) {
...
     Sort sort = new Sort(new Sort.Order(direction, DEFAULT_SORT_FIELD));
+
+    List<User> users = userService.findUsersBySearchString(term, sort);
+
     // We want to filter out users not initialized in firecloud yet to avoid sharing with a not yet existent user.
-    List<User> users = userService.findUsersBySearchString(term, sort)
-        .stream()
-        .filter(user -> user.getFreeTierBillingProjectName() != null)
-        .collect(Collectors.toList());
+    // This is no longer relevant if we create new workspaces with new billing projects
```
calbach: Ideally we'd still filter out users who had never signed in - as they will not be registered in Firecloud (else the FC share call will fail with some opaque error). It's a bit of an indirect check, but I'd suggest just changing this code (and don't fork) to check for firstSignInTime != null

ui/src/app/views/homepage/component.tsx
```diff
@@ -252,16 +255,23 @@ export const Homepage = withUserProfile()(class extends React.Component<
     }
   }
 
+  async checkUseBillingProjectBufferFeatureFlag() {
+    const config = await configApi().getConfig();
```
calbach: This is pretty rough, there's an existing instance of this on the homepage too.

1. The server config is already loaded on the client side and it blocks the application load.
2. This config can be considered to be static for the lifetime of the app's client session.

I think we need a `configStore`, or something even more streamlined for synchronous access; this PR is kind of forcing the issue for me. @blrubenstein I don't know whether this has come up on other React PRs already (I'm surprised if not).

ericsong: Do we have a pattern for centrally storing application data? I'm thinking something like redux

calbach: Here is our current pattern - it's not super elegant and it's led to a few bugs and some issues during testing. In this case extending the current pattern is probably sufficient. We should probably revisit once we're fully on React:

https://github.com/all-of-us/workbench/blob/00a906d0e936cead8ca2ea7cf5b63cdf324e86cc/ui/src/app/utils/navigation.tsx#L14

Note: we do use redux in the cohort builder in a few places.

ui/src/app/views/app/component.ts
```diff
@@ -190,4 +192,9 @@ export class AppComponent implements OnInit {
     });
   }
 
+  private async loadConfig() {
+    const config = await configApi().getConfig();
```
calbach: Please get this via `ServerConfigService.getConfig()` to avoid the extra request:

https://github.com/all-of-us/workbench/blob/00a906d0e936cead8ca2ea7cf5b63cdf324e86cc/ui/src/app/services/server-config.service.ts#L14

You might need to change the type or coerce the type for now (while we're still mixed React/Angular)

ui/src/app/utils/navigation.tsx
```diff
@@ -17,6 +17,7 @@ export const currentConceptSetStore = new BehaviorSubject<ConceptSet>(undefined)
 export const urlParamsStore = new BehaviorSubject<any>({});
 export const queryParamsStore = new BehaviorSubject<any>({});
 export const routeConfigDataStore = new BehaviorSubject<any>({});
+export const configDataStore = new BehaviorSubject<any>({});
```
calbach: Should use a type - the above 3 are `any` because they originate from Angular and don't have a type (or it's complex).

calbach: nit: naming convention would be `configStore` or `serverConfigStore`. The above variable is a "routeConfigData" "store", not a "routeConfig" "dataStore".

PR #2146, opened 2019-05-17
[RW-2655][risk=no] Attempt fix for flaky UI test failure

dolbeew:
Also tried to clean up console errors that occur when running ui tests as those may contribute to circle-ci failures


ui/src/app/cohort-search/cohort-search/cohort-search.component.spec.ts
```diff
@@ -150,7 +153,6 @@ describe('CohortSearchComponent', () => {
   }));
 
   beforeEach(() => {
-    registerApiClient(CohortsApi, new CohortsApiStub());
```
calbach: This line seemed correct to me since the component uses the cohortsApi. I'd suggest leaving this even if the current test coverage doesn't actually hit the lines that require this.

ui/src/app/views/workspace/component.spec.ts
```diff
@@ -32,7 +32,7 @@ import {
   UserService,
 } from 'generated';
 import {
-  CohortsApi,
+  CohortsApi, ConceptSetsApi,
```
calbach: nit: move to next line

ui/src/app/views/workspace/component.spec.ts
```diff
@@ -121,6 +122,7 @@ describe('WorkspaceComponent', () => {
       });
     currentWorkspaceStore.next(workspaceDataStub);
     registerApiClient(CohortsApi, new CohortsApiStub());
+    registerApiClient(ConceptSetsApi, new ConceptSetsApiStub());
```
calbach: Hmm, I don't see any relationship between this and the component being tested - were you seeing an error originate from here? If so what was it?

dolbeew: It wasn't causing a test failure but it did throw a console error that looked very similar to the circle-ci flake so I thought it might help with flakes:

`./src/app/views/recent-work/component.tsx?:62 Error: API client is not registered: ensure you are not retrieving an API client before app initialization. In unit tests, be sure to call registerApiClient() for all API clients in use, else call bindApiClients(): function ConceptSetsApi() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    at Object.eval [as conceptSetsApi] (./src/app/services/swagger-fetch-clients.ts?:48)
    at class_1.eval (./src/app/views/recent-work/component.tsx?:45)
    at step (./node_modules/tslib/tslib.es6.js?:117)
    at Object.eval [as next] (./node_modules/tslib/tslib.es6.js?:98)
    at eval (./node_modules/tslib/tslib.es6.js?:91)
    at new ZoneAwarePromise (./node_modules/zone.js/dist/zone.js?:890)
    at Module.__awaiter (./node_modules/tslib/tslib.es6.js?:87)
    at eval (./src/app/views/recent-work/component.tsx?:35)
    at class_1.eval (./src/app/views/recent-work/component.tsx?:55)
    at step (./node_modules/tslib/tslib.es6.js?:117)`

ui/src/app/cohort-search/cohort-search/cohort-search.component.spec.ts
calbach: Oops, sorry - I thought these were two different tests - carry on

ui/src/app/views/workspace/component.spec.ts
calbach: Ah good catch - I see the dep now coming in from app-recent-work.

PR #2148, opened 2019-05-17
[risk=no] Pull updated menu snippets

calbach:
Follow up to https://github.com/all-of-us/workbench-snippets/pull/16


calbach:
Discussed in slack - autogenerated and copied from workbench-snippets; review would happen there. Would be nice to format these better, but unfortunately I don't think we can break up the newlines and still have valid JSON.

PR #2149, opened 2019-05-17
[RW-2475][risk=low] Rework cluster reset button for cluster-per-workspace

ericsong:
Removes the global `/settings` page which allowed the user to reset their notebook cluster. The page was refactored into a button component and then used in the workspace details page so users can reset clusters on a workspace to workspace basis.

Decided not to hide it behind a feature flag since the new workflow still works for the purposes of resetting the free tier billing project. Hitting it from any workspace will do the job. 

Screen Shot 2019-05-17 at 3 27 15 PM


ui/src/app/views/reset-cluster-button/component.tsx
```diff
@@ -84,12 +82,14 @@ export class SettingsReact extends React.Component<{}, SettingsState> {
...
-          <br/><br/>
-          Resetting should not be necessary under normal conditions. Please help us to
-          improve this experience by using "Report a Bug" from the profile drop-down
-          to describe the reason for this reset.
+            <div>
+              <strong>Warning:</strong> Any unsaved changes to your notebooks may be lost
+              and your cluster will be offline for 5-10 minutes.
+              <br/><br/>
+              Resetting should not be necessary under normal conditions. Please help us to
+              improve this experience by using "Report a Bug" from the profile drop-down
```
jacmrob: this text should change to refer users to "Contact Support" from the side hamburger menu.  "Report a Bug" on the profile drop down no longer exists since we released zendesk.

ui/src/app/views/reset-cluster-button/component.tsx
```diff
@@ -147,10 +149,19 @@ export class SettingsReact extends React.Component<{}, SettingsState> {
 }
 
 @Component({
+  selector: 'app-react-cluster-button',
   template: '<div #root></div>'
 })
-export class SettingsComponent extends ReactWrapperBase {
+class ResetClusterButtonComponent extends ReactWrapperBase {
+  @Input() billingProjectId;
```
jacmrob: probably want to type this so the complier is happy

ui/src/app/views/workspace/component.ts
```diff
@@ -204,6 +195,19 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     this.bugReportOpen = true;
   }
 
+  workspaceClustersBillingProjectId(): string {
+    if (this.workspace.namespace === this.freeTierBillingProject) {
+      return this.freeTierBillingProject;
+    }
+
+    if (this.accessLevel === WorkspaceAccessLevel.WRITER ||
```
calbach: opt_nit: `[WorkspaceAccessLevel.WRITER, WorkspaceAccessLevel.OWNER].includes(this.accessLevel)`

ui/src/app/views/workspace/component.ts
```diff
@@ -204,6 +195,19 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     this.bugReportOpen = true;
   }
 
+  workspaceClustersBillingProjectId(): string {
```
calbach: nit: I would not recommend naming methods/variables with a possessive, I think it's a bit ambiguous with the plural. I would just name this workspaceClusterBillingProjectId

PR #2151, opened 2019-05-20
[RW-2646][Risk=no] Srubenst/export to notebook menu item

blrubenstein:


NehaBroad:
So i was able to create a new notebook without providing notebook name.

PR #2155, opened 2019-05-21
[RW-2552][risk=no] log admin user bypass action

jacmrob:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -714,7 +714,7 @@ public class ProfileController implements ProfileApiDelegate {
     }
     userService.logAdminUserAction(
         userId,
-        "updated user bypass on " + moduleName,
+        "set bypass status for module " + moduleName + " to " + bypassed,
         previousValue,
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -672,37 +672,52 @@ private boolean userNeverLoggedIn(com.google.api.services.admin.directory.model.
...
       case "twoFactorAuth":
+        previousValue = user.getTwoFactorAuthBypassTime();
         userService.setTwoFactorAuthBypassTime(userId, valueToSet);
         break;
       default:
         throw new BadRequestException("There is no access module named: " + moduleName);
     }
+    userService.logAdminUserAction(
+        userId,
+        "updated user bypass on " + moduleName,
```
gjuggler: Wording nit: maybe "Set bypass status for module " + moduleName + " to " + bypassed

I think including the true/false value is useful since otherwise an admin (who I imagine might eventually be reading these log messages in a table in the UI somewhere) would have to grok the previousValue / valueToSet which may not be immediately clear how to interpret.

PR #2157, opened 2019-05-21
[RW-1665][risk=no] Speed up GET /me

ericsong:
Doing a 2FA sync on every call to GET /me added a significant amount of lag to the endpoint based on local testing. This PR removes the 2FA sync. It should be a safe change since we make an explicit call to sync 2FA whenever the user hits the homepage.

Local Testing Results of 50 GET /me calls
average w/ sync - .96312 s
average w/o sync - .163540 s


jacmrob:
Can you test this with a new user before you merge? 

Create a new user, *don't* set up 2FA for them, and make sure on the access modules page it shows 2fa as incomplete.

ericsong:
Tested

PR #2160, opened 2019-05-23
[RW-1304][RISK=NO] Add support for pagination in workbench concept search

NehaBroad:


ui/src/app/views/concept-table/component.tsx
```diff
@@ -93,9 +110,16 @@ export class ConceptTable extends React.Component<{concepts: Concept[];
...
+      selectedVocabularies: [],
+      loading: false,
+      first: 0,
+      rows: 10,
+      totalRecords: props.concepts.length,
+      pageNumber: 0,
+      pageConcepts: props.concepts.slice(0, 10)
     };
     this.filterImageSrc = 'filter';
+    this.onPage = this.onPage.bind(this);
```
blrubenstein: Do you need to use bind, instead of binding the function below to `() => this.onPage()`? (I think we prefer to use lambdas over binds when we can)

ui/src/app/views/concept-table/component.tsx
```diff
@@ -80,11 +80,28 @@ export class SynonymsObject extends React.Component<{},
...
+  nextPage: Function;
+  placeholderValue: string;
+  reactKey: string;
+  selectedConcepts: Concept[];
+}
+
+interface State {
+  first: number;
+  loading: boolean ;
+  rows: number;
```
blrubenstein: Does this ever change? If not, we should make it a constant.

ui/src/app/views/concept-table/component.tsx
```diff
@@ -80,11 +80,28 @@ export class SynonymsObject extends React.Component<{},
...
+  onSelectConcepts: Function;
+  nextPage: Function;
+  placeholderValue: string;
+  reactKey: string;
+  selectedConcepts: Concept[];
+}
+
+interface State {
+  first: number;
+  loading: boolean ;
```
blrubenstein: Can we name this something different to differentiate from the prop `loading`? What are they loading? We should specify

NehaBroad:
So the way it is working right now is it gets the first 100 rows and then gets the next set once the user is gets to the last page.  So rather than getting all the rows it gets 100 per request

PR #2161, opened 2019-05-23
[RW-2669][Risk=high] Srubenst/delete profile

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -741,6 +741,9 @@ public class ProfileController implements ProfileApiDelegate {
 
   @Override
   public ResponseEntity<Void> deleteProfile() {
+    if (!workbenchConfigProvider.get().featureFlags.unsafeAllowDeleteUser) {
+      throw new ForbiddenException("Self account deletion is disallowed in this environment.");
+    }
     User user = userProvider.get();
     log.log(Level.WARNING, "User email: " + user.getEmail());
     directoryService.deleteUser(user.getEmail().split("@")[0]);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -745,7 +745,7 @@ public class ProfileController implements ProfileApiDelegate {
       throw new ForbiddenException("Self account deletion is disallowed in this environment.");
     }
     User user = userProvider.get();
-    log.log(Level.WARNING, "User email: " + user.getEmail());
+    log.log(Level.WARNING, "Deleting profile: user email: " + user.getEmail());
     directoryService.deleteUser(user.getEmail().split("@")[0]);
     userDao.delete(user.getUserId());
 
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -739,4 +739,17 @@ private void updateBypass(long userId, AccessBypassRequest request) {
     );
   }
 
+  @Override
+  public ResponseEntity<Void> deleteProfile() {
+    if (!workbenchConfigProvider.get().featureFlags.unsafeAllowDeleteUser) {
+      throw new ForbiddenException("Self account deletion is disallowed in this environment.");
+    }
+    User user = userProvider.get();
+    log.log(Level.WARNING, "User email: " + user.getEmail());
```
calbach: Was this for debugging? I think it would probably be useful to keep this log in, but I would add a bit more context, e.g. deleting user "..."

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -128,5 +128,6 @@ public static WorkbenchConfig createEmptyConfig() {
 
   public static class FeatureFlagsConfig {
     public boolean useBillingProjectBuffer = false;
+    public boolean unsafeAllowDeleteUser;
```
calbach: Do we intend to eventually turn this on in production? If not, it probably doesn't qualify as a feature flag.

api/src/main/resources/workbench.yaml
```diff
@@ -114,6 +114,35 @@ paths:
...
+        204:
+          description: Request received.
+        400:
+          description: Bad request
+          schema:
+            $ref: "#/definitions/ErrorResponse"
+    delete:
+      tags:
+        - profile
+      description: Deletes the user's profile and all associated information
```
calbach: This potentially leaves around a bunch of orphaned Firecloud data. I would definitely not try to address that in this PR but probably worth noting here.

blrubenstein: Actually, that's a good point. @bopwerks FYI that we can't clean up all information, so it makes sense to make sure usernames are still UUID coded or something so we don't have FireCloud account creation clashes.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
blrubenstein: Probably not. I wasn't sure where this should go though, do you have thoughts on where to put this flag?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
blrubenstein: It was originally for debugging, but I'll add context and take credit for leaving it in! :D

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
calbach: Per slack discussion, on further thought it's probably fine here; this tuple doesn't need to strictly apply to transient features. More generally a feature could be a permanent fixture which controls whether some functionality is enabled or disabled.

PR #2164, opened 2019-05-23
[RW-2681][Risk=no] Add back highlight search term concept search

blrubenstein:


ui/src/app/views/concept-table/component.tsx
```diff
@@ -135,10 +143,35 @@ export class ConceptTable extends React.Component<{concepts: Concept[];
...
+          words.push(splits[i], matches[i]);
+        }
+        words.push(splits[splits.length - 1]);
+      }
+      return words.map(word => <span
+        style={matchString.test(word.toLowerCase()) ? styles.highlighted : {}}>
+          {word}
+        </span>);
+    } else {
+      return stringToHighlight;
```
NehaBroad: nit: we can also write the above as 
 if (!searchTerm) {
return stringToHighlight;
}
const words....

ui/src/app/views/concept-table/component.tsx
```diff
@@ -27,11 +27,19 @@ const styles = reactStyles({
     marginLeft: 10,
     verticalAlign: 'middle',
     color: colors.blue[3]
+  },
+  highlighted: {
+    backgroundColor: 'yellow',
```
NehaBroad: if this yellow is same as '#F8C954' we can use colors.ts

ui/src/app/views/concept-table/component.tsx
```diff
@@ -135,10 +143,35 @@ export class ConceptTable extends React.Component<{concepts: Concept[];
...
     </SynonymsObject>);
   }
 
+  highlightWithSearchTerm(stringToHighlight: string) {
+    const {searchTerm} = this.props;
+    if (searchTerm) {
+      const words: string[] = [];
+      let searchWords = searchTerm.split(new RegExp(',| '));
+      searchWords = searchWords.filter(w => w.length > 0 );
+      searchWords = searchWords.map(word => word.replace(/[&!^\/\\#,+()$~%.'":*?<>{}]/g, ''));
```
NehaBroad: Can we write this as searchTerm.split(new RegExp(',| '))
        .filter(w => w.length > 0 )
        .map(word => word.replace(/[&!^\/\\#,+()$~%.'":*?<>{}]/g, ''))?

ui/src/app/views/concept-table/component.tsx
```diff
@@ -27,12 +27,21 @@ const styles = reactStyles({
     marginLeft: 10,
     verticalAlign: 'middle',
     color: colors.blue[3]
+  },
+  highlighted: {
+    color: colors.green[2],
+    backgroundColor: colors.green[2].alpha(0.2),
+    backgroundOpacity: '0.2',
```
NehaBroad: question: do we need backgroundOpacity if its specified in backgroundColor?

blrubenstein: we don't, that's a meaningless css property that doesn't exist that I was trying out and need to remove!

PR #2170, opened 2019-05-28
[RW-2595][RW-2518][risk=no] Purpose of Use questions & ontology

jacmrob:


ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-                         placeholder='Name of Disease' onChange={v =>
-                  this.setState(fp.set(['workspace', 'researchPurpose', 'diseaseOfFocus'], v))}
-                         disabled={!this.state.workspace.researchPurpose.diseaseFocusedResearch}/>
-            </TooltipTrigger>
-          </WorkspaceCategory>
-          <div style={{display: 'inline-block'}}>
-            <div style={{display: 'flex'}}>
+            tooltip={toolTipText.billingAccount}/>
+        <WorkspaceEditSection header='Research Use Statement Questions'
+            text={['The AoU Research Program requires each user of AoU data to provide a \
```
blrubenstein: I know this is an existing problem, but can we make this prop more meaningful than `text`? I'd expect `text` to be a child. Maybe `sectionDescription`?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+             meaningful description of the intended purpose of data use for each Workspace they \
+             create. The responses provided below will be posted publicly in the AoU Research Hub \
+             website to inform the AoU Research Participants.  Therefore, please provide \
+             sufficiently detailed responses at a 5th grade reading level.  Your responses will \
+             not be used to make decisions about data access.', <br/>, <br/>,
+              <i>Note that you are required to create separate Workspaces for each project
+              for which you access AoU data, hence the responses below are expected to be specific
+              to the project for which you are creating this particular Workspace.</i>
+            ]}/>
+        <WorkspaceEditSection header={researchPurposeQuestions[0].header}
```
blrubenstein: There's some boilerplate here that feels like it may be doable via `researchPurposeQuestions.map`. I wonder if that's usable here.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-          <div style={{display: 'inline-block'}}>
-            <div style={{display: 'flex'}}>
-              <WorkspaceCategory item={ResearchPurposeItems.population}
-                  value={this.state.workspace.researchPurpose.population}
-                  onChange={v => this.updateWorkspaceCategory('population', v)}/>
-              <WorkspaceCategory item={ResearchPurposeItems.commercialPurpose}
-                  value={this.state.workspace.researchPurpose.commercialPurpose}
-                  onChange={v => this.updateWorkspaceCategory('commercialPurpose', v)}/>
+        </WorkspaceEditSection>
+        <WorkspaceEditSection
```
blrubenstein: This especially is where I feel like there is value in the map. All of these seem like repeated code.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+                    onChange={v => this.updateResearchPurpose('intendedStudy', v)}
+                    style={{marginTop: '0.5rem'}}/>
+        </WorkspaceEditSection>
+        <WorkspaceEditSection header={researchPurposeQuestions[3].header}
+                              text={researchPurposeQuestions[3].description} required>
+          <TextArea value={this.state.workspace.researchPurpose.anticipatedFindings}
+                    onChange={v => this.updateResearchPurpose('anticipatedFindings', v)}
+                    style={{marginTop: '0.5rem'}}/>
+        </WorkspaceEditSection>
+        <WorkspaceEditSection required header={researchPurposeQuestions[4].header}>
```
blrubenstein: This one doesn't reference text here. Should it?

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+            A primary mission of the AoU Research Program is to include research participants
+            who are medically underserved or are historically underrepresented in Biomedical
+            Research, or who, because of systematic social disadvantage, experience health
+            disparities.  As a way to assess the research being conducted with a focus on these
+            populations, AoU requires that you indicate the demographic categories you intend
+            to focus your analysis on.
+          </div>}
+          <div style={{marginTop: '0.5rem'}}>
+            <RadioButton name='population' style={{marginRight: '0.5rem'}}
+              onChange={v => this.updateResearchPurpose('population', true)}
```
blrubenstein: The spacing on this radio button vs the lower one is different. Should probably be consistent (surprised our linter didn't get upset)

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+            <strong>If "Yes": </strong> Please specify the demographic category or categories of the
+            population(s) that you are interested in exploring in your study.
+            Select as many as applicable.
+            <div style={{display: 'flex', flexDirection: 'row', flex: '1 1 0',
+              marginTop: '0.5rem'}}>
+              <div style={{display: 'flex', flexDirection: 'column'}}>
+                <LabeledCheckBox label='Race/Ethnicity'
+                  value={this.specificPopulationSelected(SpecificPopulationEnum.RACEETHNICITY)}
+                  onChange={v =>
+                     this.updateSpecificPopulation(SpecificPopulationEnum.RACEETHNICITY, v)}
```
blrubenstein: Opt nit: Because this is the same enum as value uses, there's a bit of code smell around duplication. Would it be possible to reduce that with variable or map? (Using an array and then mapping over it may give a bit more documentation of what is on which side also.)

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -449,78 +532,189 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
           </div>
-          <WorkspaceUnderservedPopulation
-            selectedValues={
-              !this.state.workspace.researchPurpose.underservedPopulationDetails ? [] :
-                  this.state.workspace.researchPurpose.underservedPopulationDetails}
-              onChange={v => this.updateUnderservedPopulation(v)}>
-          </WorkspaceUnderservedPopulation>
         </WorkspaceEditSection>
         <WorkspaceEditSection header='Request a review of your research purpose'
                               tooltip={toolTipText.reviewRequest}>
```
blrubenstein: This appears not to have `text`. Is that intentional?

ui/src/app/views/workspace-list/component.tsx
```diff
@@ -172,7 +172,7 @@ export class WorkspaceCard extends React.Component<
                   {wp.workspace.name}</div>
               </Clickable>
             </div>
-            <div style={styles.workspaceDescription}>{wp.workspace.description}</div>
+            {/*<div style={styles.workspaceDescription}>{wp.workspace.description}</div>*/}
```
blrubenstein: I'd just delete this if we don't need it. But it should probably be a different one of those text fields? Maybe `scientificQuestion`? This is a good Karthik question.

ui/src/app/views/workspace-edit/component.tsx
```diff
@@ -547,8 +741,14 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-              { this.isEmpty('name') && <li> Name </li> }
-              { this.isEmpty('description') && <li> Description </li> }
+              { this.isEmpty(this.state.workspace, 'name') && <li> Name </li> }
+              { this.isEmpty(this.state.workspace.researchPurpose, 'intendedStudy') &&
+              <li>Field of intended study</li>}
+              { this.isEmpty(this.state.workspace.researchPurpose, 'anticipatedFindings') &&
+              <li>Anticipated findings</li>}
+              { this.isEmpty(this.state.workspace.researchPurpose, 'softwareChoice') &&
+              <li>Reason for choosing AoU</li>}
+              { !this.categoryIsSelected() && <li>Research focus</li>}
```
blrubenstein: Should this also validate and error on "Yes I am focusing on a specific population, but no I have not selected which one"?

ui/src/app/views/workspace/component.spec.ts
```diff
@@ -125,8 +124,8 @@ describe('WorkspaceComponent', () => {
     registerApiClient(WorkspacesApi, new WorkspacesApiStub());
   }));
 
-  it('displays research purpose', fakeAsync(() => {
-    expect(fixture.debugElement.queryAll(
-      By.css('.research-purpose-item')).length).toEqual(2);
+  // TODO [RW-2594] add new tests
```
blrubenstein: If you aren't adding tests here, would it be worth throwing a quick comment with what tests you think should be added? Alternatively, is there a reason _not_ to just add the tests?

ui/src/app/views/workspace-list/component.tsx
jacmrob: I believe that RW-2594 answers this

ui/src/app/views/workspace/component.spec.ts
jacmrob: reason not to add the tests is that the ws about page is about to be refactored in RW-2594.  tests here should probably test that it displays the correct answers to rp questions as per the new UI

ui/src/app/views/workspace-edit/component.tsx
jacmrob: i'm not sure if it is worth pulling out because each of the sections are so different

ui/src/app/views/workspace-edit/component.tsx
jacmrob: good thought

ui/src/app/views/workspace-edit/component.tsx
jacmrob: it is, i believe

ui/src/app/views/workspace-edit/component.tsx
jacmrob: yeah i think this is a good place to use mapping.  i'm not sold on doing it above.  although even here there is some weird ordering with the two columns, and how "other" needs to go last

ui/src/app/views/workspace-edit/component.tsx
jacmrob: ^ oh nvm you meant for the checkboxes

ui/src/app/views/workspace/component.spec.ts
blrubenstein: Got it. Can you add that as a comment?

PR #2171, opened 2019-05-28
[RW-2624][Risk=no] Require concepts for valid concept set

blrubenstein:


ui/src/app/views/concept-set-list/component.tsx
```diff
@@ -48,11 +43,8 @@ export const ConceptSetsList = withCurrentWorkspace()(
     }
 
     componentDidMount() {
-      try {
-        this.getDomainInfo();
-      } finally {
-        this.loadConceptSets();
-      }
+      this.getDomainInfo();
```
NehaBroad: Do we need to log errors if domainInfo returns error?

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -253,6 +257,16 @@ private void addConceptsToSet(org.pmiops.workbench.db.model.ConceptSet dbConcept
       String workspaceId, Long conceptSetId, UpdateConceptSetRequest request) {
     org.pmiops.workbench.db.model.ConceptSet dbConceptSet = getDbConceptSet(
         workspaceNamespace, workspaceId, conceptSetId, WorkspaceAccessLevel.WRITER);
+
+    Set<Long> allConceptSetIds = dbConceptSet.getConceptIds();
+    if (request.getAddedIds() != null) {
+      allConceptSetIds.addAll(request.getAddedIds());
+    }
+    int sizeOfAllConceptSetIds = allConceptSetIds.stream().distinct().collect(Collectors.toSet()).size();
```
NehaBroad: Since allConceptSetIds is  set we do not need distinct()

NehaBroad:
Do we need CreateConceptSetModal anymore? i noticed it being referenced  in dataPage but not sure its being used anymore

ui/src/app/views/concept-set-list/component.tsx
blrubenstein: We are, as part of the `getDomainInfo` function.

PR #2172, opened 2019-05-28
[ RW-2662][RISK=NO] Rename Data Set

NehaBroad:


blrubenstein:
Small Nit: PR isn't about the tiles, right? It's about the rename modal? Do you mind changing that?

ui/src/app/views/data-page/component.tsx
```diff
@@ -123,6 +124,8 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
         conceptSetsApi().getConceptSetsInWorkspace(namespace, id),
         dataSetApi().getDataSetsInWorkspace(namespace, id)
       ]);
+      const existingDatasetName = dataSets.items.map(dataSet => dataSet.name);
```
blrubenstein: Nit: I'd avoid the intermediate variable, and just put the map inside the setState call.

ui/src/app/views/recent-work/component.tsx
```diff
@@ -72,6 +75,13 @@ export const RecentWork = (fp.flow as any)(
     }
   }
 
+  existingName(resource) {
```
blrubenstein: For consistency, can we name this `existingList` or the above `existingName`?

ui/src/app/views/recent-work/component.tsx
```diff
@@ -52,7 +53,9 @@ export const RecentWork = (fp.flow as any)(
             cohortsApi().getCohortsInWorkspace(namespace, id),
             conceptSetsApi().getConceptSetsInWorkspace(namespace, id)
           ]);
-           // TODO Remove this cast when we switch to fetch types
+          const notebookName = notebooks.map(notebook => notebook.name);
```
blrubenstein: Nit: I'd avoid the intermediate variable and just put this into the setState call.

ui/src/app/views/resource-card/component.tsx
```diff
@@ -503,6 +558,13 @@ export class ResourceCard extends React.Component<ResourceCardProps, ResourceCar
                           workspaceNamespace={this.props.resourceCard.workspaceNamespace}
                           workspaceFirecloudName={this.props.resourceCard.workspaceFirecloudName}
                           closeFunction={() => this.setState({exportingDataSet: false})}/>}
+      {this.state.renamingDataSet &&
```
blrubenstein: Following above, could we do `this.state.renaming && this.isDataSet`? (Just to minimize state?)

ui/src/app/views/rename-modal/component.tsx
```diff
@@ -87,26 +53,23 @@ export class RenameModal extends React.Component<RenameModalProps, {
         message: 'already exists'
       }
     }});
-    return <Modal loading={saving}>
-      <ModalTitle>Please enter the new name for {notebookName}</ModalTitle>
+    return <Modal>
+      <ModalTitle>Please enter new name for {oldName}</ModalTitle>
```
blrubenstein: Should this also have a description? Perhaps optionally? If so, can we also use this for the cohortRename modal? (I think its called `cohort-edit-modal`?)

ui/src/app/views/resource-card/component.tsx
```diff
@@ -481,18 +538,30 @@ export class ResourceCard extends React.Component<ResourceCardProps, ResourceCar
             {fp.startCase(fp.camelCase(this.resourceType.toString()))}</div>
         </div>
       </ResourceCardBase>
-      {this.state.editing && (this.isCohort  || this.isConceptSet) &&
-      <EditModal resource={ResourceCard.castConceptSet(this.props.resourceCard)}
-                 onEdit={v => this.receiveEdit(v)}
-                 onCancel={() => this.closeEditModal()}/>}
+      {this.state.editing && (this.isCohort) &&
```
blrubenstein: Nit: Can this use `renaming` in state instead? And maybe we can do away with editing?

PR #2174, opened 2019-05-29
[RW-2661][Risk=no] Srubenst/concept set actions

blrubenstein:

### DIFF
--- a/ui/src/app/views/concept-add-modal.tsx
+++ b/ui/src/app/views/concept-add-modal.tsx
@@ -107,10 +107,10 @@ export const ConceptAddModal = withCurrentWorkspace()
         addedIds: conceptIds
       };
       try {
-        await conceptSetsApi().updateConceptSetConcepts(
+        const conceptSet = await conceptSetsApi().updateConceptSetConcepts(
           namespace, id, selectedSet.id, updateConceptSetReq);
         this.setState({saving: false});
-        onSave();
+        onSave(conceptSet);
       } catch (error) {
         console.error(error);
       }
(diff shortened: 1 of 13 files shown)


ui/src/app/views/concept-add-modal.tsx
```diff
@@ -125,9 +125,10 @@ export const ConceptAddModal = withCurrentWorkspace()
         addedIds: conceptIds
       };
       try {
-        await conceptSetsApi().createConceptSet(namespace, id, request);
+        const createdConceptSet =
+          await conceptSetsApi().createConceptSet(namespace, id, request);
         this.setState({saving: false});
-        onSave();
+        onSave(createdConceptSet);
```
dolbeew: The concept set also needs to be passed to the `onSave()` on line [113](https://github.com/all-of-us/workbench/pull/2174/files#diff-d0aaa77339d78f2f09ba5efb57acb13aR113). Currently it throws an error if I try to add to an existing set.

ui/src/app/views/concept-add-modal.tsx
```diff
@@ -125,9 +125,10 @@ export const ConceptAddModal = withCurrentWorkspace()
         addedIds: conceptIds
       };
       try {
-        await conceptSetsApi().createConceptSet(namespace, id, request);
+        const createdConceptSet =
+          await conceptSetsApi().createConceptSet(namespace, id, request);
```
dolbeew: Just curious, is there a difference/benefit to using async/await here instead of putting the next two lines in `.then()`?

ui/src/app/views/concept-add-modal.tsx
blrubenstein: Good catch. That actually maybe means that saveNew and updateExisting should be different. Will check product behavior with Keri.

ui/src/app/views/concept-add-modal.tsx
blrubenstein: Not particularly, I do think its more readable though, if there's no need to do other work in a synchronous manner.

PR #2177, opened 2019-05-29
[RW-2647][RISK=NO]Column names on Data Set Builder don't fit

NehaBroad:
Tooltip will appear only when textOverflow property is applied 
Screen Shot 2019-05-29 at 4 33 01 PM

### DIFF
--- a/ui/src/app/views/dataset-page.tsx
+++ b/ui/src/app/views/dataset-page.tsx
@@ -8,8 +8,6 @@ import {ClrIcon} from 'app/components/icons';
 import {TooltipTrigger} from 'app/components/popups';
 import {Spinner} from 'app/components/spinners';
 
-import * as ReactDOM from 'react-dom';
-
 import {
   cohortsApi,
   conceptsApi,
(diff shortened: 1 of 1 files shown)


ui/src/app/views/dataset-page.tsx
```diff
@@ -378,6 +380,33 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
       }
     }
 
+    isEllipsisActive(text) {
+      // For Data table with column title as active text, enable the tooltip if column's scroll
+      // width is greater than offsetWidth as that represents the textOverflow style has been
+      // applied and it has ellipsis at the end
+
+      const column = document.getElementsByClassName('p-column-title');
```
blrubenstein: Nit: I'd call this `columns` rather than `column` because it covers all columns.

ui/src/app/views/dataset-page.tsx
```diff
@@ -378,6 +380,33 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+      for (let i = 0; i < column.length; i++) {
```
blrubenstein: Rather than doing a loop here, it may make sense to generate a map at render time. Right now this is n^2, but generating a map at render time takes it to n and is probably simpler?

PR #2178, opened 2019-05-30
[P2][RW-2610][risk=no] fix reader permissions on concept set

jacmrob:
per the react conversion, most of the permissions things were fixed, however I 
* changed the styling of ellipses menu to not disable the whole menu but rather show tooltips for disabled actions
* disabled adding concepts to sets if user is a reader, plus added a tooltip

### DIFF
--- a/ui/src/app/views/concept-homepage.tsx
+++ b/ui/src/app/views/concept-homepage.tsx
@@ -25,7 +25,7 @@ import {
   DomainCount,
   DomainInfo,
   StandardConceptFilter,
-  VocabularyCount
+  VocabularyCount, WorkspaceAccessLevel
 } from 'generated/fetch';
 
 const styles = reactStyles({
@@ -327,6 +327,20 @@ export const ConceptHomepage = withCurrentWorkspace()(
         'concepts', 'sets', conceptSet.id, 'actions']);
     }
 
+    get canEdit(): boolean {
+      return this.isOwner || this.isWriter;
+    }
+
+    get isWriter(): boolean {
+      return this.props.workspace.accessLevel as unknown as WorkspaceAccessLevel
+        === WorkspaceAccessLevel.WRITER;
+    }
+
+    get isOwner(): boolean {
+      return this.props.workspace.accessLevel as unknown as WorkspaceAccessLevel
+        === WorkspaceAccessLevel.OWNER;
+    }
+
     render() {
       const {loadingDomains, conceptDomainList, standardConceptsOnly, showSearchError,
         searching, concepts, searchLoading, conceptDomainCounts, selectedDomain,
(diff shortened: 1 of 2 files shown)


ui/src/app/views/concept-homepage.tsx
```diff
@@ -25,7 +25,7 @@ import {
   DomainCount,
   DomainInfo,
   StandardConceptFilter,
-  VocabularyCount
+  VocabularyCount, WorkspaceAccessLevel
```
ericsong: newline? so its consistent with the rest of this object

ui/src/app/views/concept-homepage.tsx
```diff
@@ -327,6 +327,20 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
+    get canEdit(): boolean {
+      return this.isOwner || this.isWriter;
+    }
+
+    get isWriter(): boolean {
+      return this.props.workspace.accessLevel as unknown as WorkspaceAccessLevel
+        === WorkspaceAccessLevel.WRITER;
+    }
+
+    get isOwner(): boolean {
```
ericsong: Is it possible to use `WorkspacePermissions` to replace these functions?

I'm not a fan of how `WorkspacePermissions` forces you to construct an object to use its functions but its probably better to reuse than reimplement.

ui/src/app/views/concept-homepage.tsx
```diff
@@ -407,8 +421,10 @@ export const ConceptHomepage = withCurrentWorkspace()(
                             nextPage={(page) => this.getNextConceptSet(page)}/>
               <SlidingFabReact submitFunction={() => this.setState({conceptAddModalOpen: true})}
                                iconShape='plus'
+                               tooltip={!this.canEdit}
+                               tooltipContent={<div>Requires Owner or Writer permission</div>}
```
ericsong: is this tooltip still relevant in the case where the user has edit permissions but `this.activeSeletedConceptCount` === 0`?

jacmrob: good catch.

ui/src/app/views/concept-homepage.tsx
jacmrob: will fix

ui/src/app/views/concept-homepage.tsx
jacmrob: wait this is actually correct, it will only show a tooltip if the user doesn't have edit permission.  (that's what the `tooltip` prop is for-- whether to show or not).  so it will not show up for a user with edit with 0 concepts selected

ui/src/app/views/concept-homepage.tsx
jacmrob: yeah i agree.  i may also fix the other place where it's reimplemented to do that too

PR #2183, opened 2019-05-31
[RW-2621][Risk=no] Srubenst/workspace share side panel update

blrubenstein:

### DIFF
--- a/ui/src/app/views/workspace-share.spec.tsx
+++ b/ui/src/app/views/workspace-share.spec.tsx
@@ -4,27 +4,27 @@ import * as Lolex from 'lolex';
 import * as fp from 'lodash/fp';
 import Select from 'react-select';
 
-import {WorkspaceShare, WorkspaceShareProps, WorkspaceShareState} from './workspace-share';
+import {Props, State, WorkspaceShare} from './workspace-share';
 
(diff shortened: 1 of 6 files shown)


ui/src/app/views/workspace-share.spec.tsx
```diff
@@ -4,27 +4,27 @@ import * as Lolex from 'lolex';
 import * as fp from 'lodash/fp';
 import Select from 'react-select';
 
-import {WorkspaceShare, WorkspaceShareProps, WorkspaceShareState} from './workspace-share';
+import {Props, State, WorkspaceShare} from './workspace-share';
```
jacmrob: is it a new convention to not give props/state objects descriptive names, and just use `Props` and `State`?

ui/src/app/views/workspace-share.tsx
```diff
@@ -250,13 +241,11 @@ export class WorkspaceShare extends React.Component<WorkspaceShareProps, Workspa
...
           this.setState({workspaceShareError: true});
         }
-        this.setState({usersLoading: false});
+        this.setState({saving: false});
       });
   }
 
   onCancel(): void {
-    // Reload the workspace to remove the temp added user/user roles in state variable workspace
-    this.reloadWorkspace();
```
jacmrob: why do we no longer need to do this?

PR #2184, opened 2019-06-01
[RW-2735][risk=low] Unique cluster names + test refactors

calbach:
- Include user ID in cluster names for uniqueness, if useBillingProjectBuffer is on
- Added test
- Refactored test to take advantage of improved provider stubbing approach

Alternative considered:
Generate a random ID for the cluster name. Rejected for now, as it would be more challenging to identify the right cluster via `listClusters` - we would need to filter down by owner (and there might be multiple matches). A 1:1 mapping is simplest for now.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -1,6 +1,5 @@
 package org.pmiops.workbench.api;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -88,14 +87,14 @@ public class ClusterController implements ClusterApiDelegate {
           };
 
   private final LeonardoNotebooksClient leonardoNotebooksClient;
-  private Provider<User> userProvider;
+  private final Provider<User> userProvider;
   private final WorkspaceService workspaceService;
   private final FireCloudService fireCloudService;
-  private Provider<WorkbenchConfig> workbenchConfigProvider;
-  private UserService userService;
-  private UserRecentResourceService userRecentResourceService;
+  private final Provider<WorkbenchConfig> workbenchConfigProvider;
+  private final UserService userService;
+  private final UserRecentResourceService userRecentResourceService;
   private final UserDao userDao;
-  private Clock clock;
+  private final Clock clock;
 
   @Autowired
   ClusterController(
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -143,11 +133,17 @@ public void setUserService(UserService userService) {
           "User billing project is not yet initialized, cannot list/create clusters");
     }
 
+    String clusterName = LeonardoNotebooksClient.DEFAULT_CLUSTER_NAME;
+    if (workbenchConfigProvider.get().featureFlags.useBillingProjectBuffer) {
+      // With 1 billing project == 1 workspace, a billing project may contain multiple clusters (up
```
ericsong: Not related to these changes but I think we should have a term to refer to the `1 billing project == 1 workspace` design/work just so its easy to refer to in our documentation.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -151,26 +151,24 @@ User user() {
...
-    when(userProvider.get()).thenReturn(user);
-    clusterController.setUserProvider(userProvider);
 
     createUser(OTHER_USER_EMAIL);
 
-    UserService userService = new UserService(
-        userProvider, userDao, adminActionHistoryDao, CLOCK, new FakeLongRandom(123),
-        fireCloudService, Providers.of(config), complianceService, directoryService);
-    clusterController.setUserService(userService);
-
```
ericsong: should the `testFcCluster` and `testCluster` have their names changed to match the new pattern?

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
gjuggler: Maybe project-per-workspace is the shortest (yet still clear) description, or 1PPW as an acronym... neither is elegant, but as long as we all know what it means.

I tried linking http://broad.io/1ppw to the core design doc (which itself can link to Jira epics, tickets, etc.). Maybe that helps?

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
calbach: Good catch. Since these tests have the flag off by default, I think we just leave it as-is for now, but added a TODO to update these later. The name here is actually not super pertinent to the tests so I don't think it's awful to have some temporary inconsistency here.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
calbach: Good point... this is what I've been calling it, but it's a bit of a mouthful. I think it's probably solidified enough that we shouldn't change it, though in hindsight something like "hermetic workspaces" might have been less choppy, albeit more opaque.

Updated this comment to match verbage.

calbach:
Ignoring UI build flake, since this doesn't touch the UI.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
ericsong: I like 1PPW, I wasn't aware of that acronym actually.

calbach: Hah - I just meant the "1 project per workspace" part, I think Greg just invented the acronym. Would shorten for typing in comments at least.

PR #2187, opened 2019-06-03
[RW-2613][risk=no] Tells GCP to launch >=10 active API instances by d…

bopwerks:
…efault to reduce ramp-up time during performance tests, which causes time-outs


calbach:
What is the plan here? I don't think we need 10 instances in all environments indefinitely. The PR I linked allows per-environment configuration of task counts, was there a reason not to just copy what was done there and increase the tasks for test only?

Also for purposes of testing I would recommend setting both a min and max, else the task count will scale up with load, resulting in new server tasks which timeout initially.

api/libproject/devstart.rb
```diff
@@ -25,38 +25,51 @@
...
+    :cdr_versions_json => "cdr_versions_local.json",
+    :gae_vars => TEST_GAE_VARS
   },
   "all-of-us-workbench-test" => {
     :api_endpoint_host => "api-dot-#{TEST_PROJECT}.appspot.com",
     :cdr_sql_instance => "#{TEST_PROJECT}:us-central1:workbenchmaindb",
     :config_json => "config_test.json",
-    :cdr_versions_json => "cdr_versions_test.json"
+    :cdr_versions_json => "cdr_versions_test.json",
+    :gae_vars => make_gae_vars(10, 64)
```
calbach: Again - I would suggest setting min === max to constrain variables for initial profiling. I expect your load test to trigger more task startups which may timeout your requests. Eventually we would want to include the autoscaling in the scope of the load testing but I'd guess it will be too noisy to start there.

I'm guessing you'll be tuning these numbers several times as you go anyways, so I leave it to you how you want to proceed.

PR #2189, opened 2019-06-03
[RW-2714][risk=no] Convert gender chart to React

dolbeew:
Convert `gender-chart` component from Angular to React (no design or functionality changes).

Screen Shot 2019-06-03 at 1 05 50 PM

Also adds basic render tests for `gender-chart` and `combo-chart` components and removes `ngx-charts` dependency since this was the last component where it was used.


ui/src/app/cohort-search/gender-chart/gender-chart.component.tsx
```diff
@@ -0,0 +1,128 @@
...
+  data: any;
+}
+
+interface State {
+  options: any;
+}
+
+export class GenderChart extends React.Component<Props, State> {
+  readonly codeMap = {
+    'M': 'Male',
```
blrubenstein: Just a quick question, is this a good opportunity with the component conversion to move to using a more inclusive gender set for the chart?

dolbeew: As far as the UI goes, we can do this anytime. I know it's been a re-identification issue in the past. @jscherdin any idea about this?

jscherdin: We are just waiting on final CAPs bucket information before we change this

PR #2195, opened 2019-06-04
[RW-2102][RISK=LOW] Create Demographics visual selector Data Set builder

NehaBroad:
This PR:
Creates a dummy concept set "Demographics" only for DataSet.
(which refers to the PERSON table)


NehaBroad:
@blrubenstein : Save demographics data set all fixed

blrubenstein:
@NehaBroad how does this behave for after the save? If you try to export an existing data set?

PR #2201, opened 2019-06-05
[RW-2764][Risk=no] Fix share workspace

blrubenstein:

### DIFF
--- a/ui/src/app/utils/index.tsx
+++ b/ui/src/app/utils/index.tsx
@@ -280,7 +280,9 @@ export const connectBehaviorSubject = <T extends {}>(subject: BehaviorSubject<T>
 
       render() {
         const {value} = this.state;
-        return <WrappedComponent {...this.props} {...{[name]: value}} />;
+        // We allow overriding of the currentValue, for reuse of the same
+        // logic outside of the scope of a current workspace.
+        return <WrappedComponent {...{[name]: value}} {...this.props}/>;
       }
     }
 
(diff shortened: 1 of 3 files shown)


blrubenstein:
Good question. The regression that came up here is that I added `withCurrentWorkspace` to workspace-share, forgetting that workspace share was also accessible from the workspace list page. I do think that is still the right way to do that, but it made me think that there should probably be a way to access those components outside the context of current workspace, especially for admin work down the road. That led to the prop overriding. The follow up is that workspace share also published to the current workspace store, which we don't want to do outside that context, so we need to clear it. 

I think it is possible to do a test on this, which would probably be opening the workspace share modal and sharing from the workspace list page (or even just opening it). If you would like, I can build out that unit test.

gjuggler:
> Good question. The regression that came up here is that I added `withCurrentWorkspace` to workspace-share, forgetting that workspace share was also accessible from the workspace list page. I do think that is still the right way to do that, but it made me think that there should probably be a way to access those components outside the context of current workspace, especially for admin work down the road. That led to the prop overriding. The follow up is that workspace share also published to the current workspace store, which we don't want to do outside that context, so we need to clear it.
> 
> I think it is possible to do a test on this, which would probably be opening the workspace share modal and sharing from the workspace list page (or even just opening it). If you would like, I can build out that unit test.

Ah, I see – so a user of this component was already trying to override props, but it wasn't working due to the ordering here.

A unit test would be great to have given the subtleness of the fix. Maybe time-box the effort to bail out if it's taking longer than 30mins?

PR #2205, opened 2019-06-05
[RW-2062] [risk=no] rubenst/attempt at cross org reporting

blrubenstein:
Please take an initial look. This is probably not ready to merge, but want some eyes on it. Next step is setting up the cron.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
@@ -2,7 +2,10 @@ package org.pmiops.workbench.api;
 
 import java.io.IOException;
 import java.sql.Timestamp;
+import java.util.List;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import org.pmiops.workbench.db.dao.UserService;
 import org.pmiops.workbench.db.model.User;
 import org.pmiops.workbench.exceptions.NotFoundException;
@@ -15,16 +18,6 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
 /** Handles offline / cron-based API requests related to user management. */
-import java.io.IOException;
-import java.sql.Timestamp;
-import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import java.util.stream.Collectors;
-
-/**
- * Handles offline / cron-based API requests related to user management.
- */
 @RestController
 public class OfflineUserController implements OfflineUserApiDelegate {
   private static final Logger log = Logger.getLogger(OfflineUserController.class.getName());
@@ -34,8 +27,8 @@ public class OfflineUserController implements OfflineUserApiDelegate {
   private final UserService userService;
 
   @Autowired
-  public OfflineUserController(CloudResourceManagerService cloudResourceManagerService,
-                               UserService userService) {
+  public OfflineUserController(
+      CloudResourceManagerService cloudResourceManagerService, UserService userService) {
     this.cloudResourceManagerService = cloudResourceManagerService;
     this.userService = userService;
   }
@@ -234,21 +227,24 @@ public class OfflineUserController implements OfflineUserApiDelegate {
   /**
    * Audits GCP access for all users in the database.
    *
-   * This API method is called by a cron job and is not part of our normal user-facing surface.
+   * <p>This API method is called by a cron job and is not part of our normal user-facing surface.
    */
   @Override
   public ResponseEntity<Void> bulkAuditProjectAccess() {
-    int errorCount = 0;
-    int userCount = 0;
-
     for (User user : userService.getAllUsers()) {
-      userCount++;
       // TODO(RW-2062): Move to using the gcloud api for list all resources when it is available.
-      List<String> unauthorizedLogs = cloudResourceManagerService.getAllProjectsForUser(user).stream()
-          .filter(project -> !(project.getParent().getId().equals(PMI_OPS_ORG_ID)))
-          .map(project -> project.getName() + " in organization " + project.getParent().getId()).collect(Collectors.toList());
+      List<String> unauthorizedLogs =
+          cloudResourceManagerService.getAllProjectsForUser(user).stream()
+              .filter(project -> !(project.getParent().getId().equals(PMI_OPS_ORG_ID)))
+              .map(project -> project.getName() + " in organization " + project.getParent().getId())
+              .collect(Collectors.toList());
       if (unauthorizedLogs.size() > 0) {
-        log.log(Level.WARNING, "User " + user.getEmail() + " has access to projects: " + String.join(", ", unauthorizedLogs));
+        log.log(
+            Level.WARNING,
+            "User "
+                + user.getEmail()
+                + " has access to projects: "
+                + String.join(", ", unauthorizedLogs));
       }
     }
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
@@ -7,6 +7,7 @@ import org.pmiops.workbench.db.dao.UserService;
 import org.pmiops.workbench.db.model.User;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.exceptions.ServerErrorException;
+import org.pmiops.workbench.google.CloudResourceManagerService;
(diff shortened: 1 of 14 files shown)


api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -201,4 +209,26 @@ private boolean timestampsEqual(Timestamp a, Timestamp b) {
...
+   */
+  @Override
+  public ResponseEntity<Void> bulkAuditProjectAccess() {
+    int errorCount = 0;
+    int userCount = 0;
+
+    for (User user : userService.getAllUsers()) {
+      userCount++;
+      List<String> unauthorizedLogs = cloudResourceManagerService.getAllProjectsForUser(user).stream()
+          .filter(project -> !(project.getParent().getId().equals(PMI_OPS_ORG_ID)))
```
gjuggler: Have you tried running the cron against our test database?

I'm pretty sure we should see AoU users with roles on projects in firecloud.org, e.g. billing projects created on their behalf. So we may well need to store an environment variable to track the firecloud org ID for each environment, or simpler yet, a constant list of all expected firecloud org IDs.

In fact, for my fake-research-aou user, I don't see any project-level roles on projects in PMI-Ops.org when I run the projects.list command.

api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+      ServiceAccounts serviceAccounts) {
+    this.serviceAccounts = serviceAccounts;
+    this.cloudResourceManagerAdminCredsProvider = cloudResourceManagerAdminCredsProvider;
+    this.httpTransport = httpTransport;
+    this.retryHandler = retryHandler;
+  }
+
+  private CloudResourceManager getCloudResourceManagerServiceWithImpersonation(User user) throws IOException {
+// Load credentials for the cloud-resource-manager Service Account. This account has been granted
+    // domain-wide delegation for the OAuth scopes required by FireCloud.
```
gjuggler: nit: update copy-pasta comment

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -201,4 +209,26 @@ private boolean timestampsEqual(Timestamp a, Timestamp b) {
...
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
+
+  /**
+   * Audits GCP access for all users in the database.
+   *
+   * This API method is called by a cron job and is not part of our normal user-facing surface.
+   */
+  @Override
+  public ResponseEntity<Void> bulkAuditProjectAccess() {
```
gjuggler: We're missing a test for this method – see examples to draw from in OfflineUserControllerTest.java. Don't need to go overboard, but having nonzero test coverage is important.

api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+      CloudResourceManagerScopes.CLOUD_PLATFORM_READ_ONLY
+  );
+
+  private final Provider<GoogleCredential> cloudResourceManagerAdminCredsProvider;
+  private final HttpTransport httpTransport;
+  private final GoogleRetryHandler retryHandler;
+  private final ServiceAccounts serviceAccounts;
+
+  @Autowired
+  public CloudResourceManagerServiceImpl(
```
gjuggler: For testing implementations of Google services, we've tended to skip unit tests and only have integration-level tests. It may make sense to do the same here. See the integration/j/o/p/workbench/ test files for examples.

A couple options for testing this scenario:

1. Assume there is a test user "cloud-resource-manager-test@fake-research-aou.org" with a single billing project, and make the ListAllProjects call on behalf of this user. Ensure the list contains 1 project w/ the expected project name. (The trouble here is that if we ever inadvertently clear out users from our test environment, this test would start to break.)
2. Same as the above, but do as much setup & teardown within the test as you can (instead of assuming the existence of a project / user). This may require way more effort, to the point of not being worthwhile.

@aweng98 I'd love to hear if you have any other suggestions for integration testing here...

aweng98: I agree. thumb up for adding integration tests. makes perfect sense to me.
The code relies on the contract of a external service (Google Cloud Resource API), integration test that exercising Google API service calls increases our confidence that Google services are working as expected and is used correctly.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -201,4 +209,26 @@ private boolean timestampsEqual(Timestamp a, Timestamp b) {
...
+    int errorCount = 0;
+    int userCount = 0;
+
+    for (User user : userService.getAllUsers()) {
+      userCount++;
+      List<String> unauthorizedLogs = cloudResourceManagerService.getAllProjectsForUser(user).stream()
```
calbach: Per discussion yesterday - this check is definitely much more limited in scope than is suggested in the ticket; for example if someone just adds me to a bucket, that project won't be returned by this API. Was there follow-up on this? Does this approach meet the acceptance criteria, or do we need to reconsider?

api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
aweng98: `GoogleCredential` provides a `getApplicationDefault`. maybe that instead providing our own test user like "cloud-resource-manager-test@fake-research-aou.org".

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
gjuggler: Yep – see latest comments on https://precisionmedicineinitiative.atlassian.net/browse/RW-2062. This isn't perfect, but it's a first step. The security requirements were somewhat vague anyway – the main goal is to make a reasonable effort to detect unauthorized cross-org permissions. Project-level roles will at least capture the most pernicious abuses of these GSuite accounts.

Plus, it seems like there's a GCP API in development that will more directly hit our use case; rather than try to cobble something together, we may be better off just waiting for that.

@blrubenstein, maybe add a TODO(RW-2062) comment somewhere in this method to mark our intention to improve this if/when a better API becomes available?

blrubenstein: Will add the todo.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
blrubenstein: https://stackoverflow.com/questions/4913891/how-to-run-gae-cron-jobs-as-specific-app-version

It looks like we can't run crons without merging it in against a versioned deployment. I think the only way to actually test out the cron is to merge it in and make sure it runs then.

blrubenstein: I have no clue what orgId we should be seeing/whitelisting, but I figure for now, we might log extra. The orgId I whitelisted here is what I got when I ran the query authed as my test user, and is what showed up for this sort of project: 
{
      "projectNumber": "731830181574",
      "projectId": "aou-rw-test-614",
      "lifecycleState": "ACTIVE",
      "name": "aou-rw-test-614",
      "labels": {
        "billingaccount": "fc-pmi-drc-workbench-testing"
      },
      "createTime": "2019-06-06T20:05:13.162Z",
      "parent": {
        "type": "organization",
        "id": "400176686919"
      }
    },

gjuggler: It's definitely possible to trigger a cron execution from a local machine -- see https://github.com/all-of-us/workbench/blob/master/api/doc/cron-jobs.md#local-testing, I found this extremely helpful when iterating on some of the other crons.

gjuggler: Ah – it may have been a simple naming issue, then. The aou-rw-test-614 project is part of the FireCloud org, since that's where user projects live. So s/PMI_OPS/FIRECLOUD/ might be enough to make me happy :)

blrubenstein: Oh, I totally tested the endpoint. Sorry, I thought this meant testing the cron execution itself, as triggered by appengine. 

Yes, I have run it locally, and curled the endpoint.

blrubenstein:
@gjuggler please re-review when you have a chance

PR #2207, opened 2019-06-06
[RW-2494][RISK=NO] Include Surveys in Concept Set Selector - Home Page

NehaBroad:
Screen Shot 2019-06-06 at 10 56 35 AM

### DIFF
--- a/ui/src/app/views/concept-homepage.spec.tsx
+++ b/ui/src/app/views/concept-homepage.spec.tsx
@@ -4,11 +4,25 @@ import * as React from 'react';
 import {conceptsApi, registerApiClient} from 'app/services/swagger-fetch-clients';
 import {currentWorkspaceStore} from 'app/utils/navigation';
 import {ConceptHomepage} from 'app/views/concept-homepage';
-import {ConceptsApi, ConceptSetsApi, DomainInfo, StandardConceptFilter, WorkspacesApi} from 'generated/fetch';
+import {
+  ConceptsApi,
+  ConceptSetsApi,
+  DomainInfo,
+  StandardConceptFilter,
+  WorkspacesApi} from 'generated/fetch';
 import {waitOneTickAndUpdate} from 'testing/react-test-helpers';
 import {ConceptSetsApiStub} from 'testing/stubs/concept-sets-api-stub';
-import {ConceptsApiStub, ConceptStubVariables, DomainStubVariables} from 'testing/stubs/concepts-api-stub';
-import {workspaceDataStub, WorkspacesApiStub, WorkspaceStubVariables} from 'testing/stubs/workspaces-api-stub';
+import {
+  ConceptsApiStub,
+  ConceptStubVariables,
+  DomainStubVariables,
+  SurveyStubVariables
+} from 'testing/stubs/concepts-api-stub';
+import {
+  workspaceDataStub,
+  WorkspacesApiStub,
+  WorkspaceStubVariables
+} from 'testing/stubs/workspaces-api-stub';
 
 
 function isSelectedDomain(
@@ -53,6 +67,13 @@ describe('ConceptHomepage', () => {
     await waitOneTickAndUpdate(wrapper);
     expect(wrapper.find('[data-test-id="domain-box-name"]').length)
       .toBe(DomainStubVariables.STUB_DOMAINS.length);
+  });
+
+  it('should have one card per survey.', async() => {
+    const wrapper = mount(<ConceptHomepage />);
+    await waitOneTickAndUpdate(wrapper);
+    expect(wrapper.find('[data-test-id="survey-box-name"]').length)
+      .toBe(SurveyStubVariables.STUB_SURVEYS.length);
   });
 
   it('should default to standard concepts only, and performs a full search', async() => {
--- a/ui/src/app/views/concept-homepage.spec.tsx
+++ b/ui/src/app/views/concept-homepage.spec.tsx
@@ -9,7 +9,8 @@ import {
   ConceptSetsApi,
   DomainInfo,
   StandardConceptFilter,
-  WorkspacesApi} from 'generated/fetch';
+  WorkspacesApi
+} from 'generated/fetch';
 import {waitOneTickAndUpdate} from 'testing/react-test-helpers';
 import {ConceptSetsApiStub} from 'testing/stubs/concept-sets-api-stub';
 import {
--- a/ui/src/app/views/concept-homepage.tsx
+++ b/ui/src/app/views/concept-homepage.tsx
@@ -26,6 +26,7 @@ import {
   DomainCount,
   DomainInfo,
   StandardConceptFilter,
+  SurveyModule,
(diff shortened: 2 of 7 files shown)


ui/src/app/views/concept-homepage.spec.tsx
```diff
@@ -4,11 +4,25 @@ import * as React from 'react';
 import {conceptsApi, registerApiClient} from 'app/services/swagger-fetch-clients';
 import {currentWorkspaceStore} from 'app/utils/navigation';
 import {ConceptHomepage} from 'app/views/concept-homepage';
-import {ConceptsApi, ConceptSetsApi, DomainInfo, StandardConceptFilter, WorkspacesApi} from 'generated/fetch';
+import {
+  ConceptsApi,
+  ConceptSetsApi,
+  DomainInfo,
+  StandardConceptFilter,
+  WorkspacesApi} from 'generated/fetch';
```
blrubenstein: Nit: Close curly brace on new line.

ui/src/app/views/concept-homepage.tsx
```diff
@@ -65,6 +66,22 @@ const styles = reactStyles({
   },
   clearSearchIcon: {
     fill: colors.blue[0], transform: 'translate(-1.5rem)', height: '1rem', width: '1rem'
+  },
+  sectionHeader: {
+    height: '24px',
```
blrubenstein: Nit: This could just be 24 (px for height is implicit)

ui/src/app/views/concept-homepage.tsx
```diff
@@ -84,12 +101,27 @@ const DomainCard: React.FunctionComponent<{conceptDomainInfo: DomainInfo,
...
-          <b>{conceptDomainInfo.participantCount}</b> participants in domain.</div>
+          <div><b>{conceptDomainInfo.participantCount}</b> participants in domain.</div>
+        </div>
         <Clickable style={styles.domainBoxLink}
                    onClick={browseInDomain}>Browse Domain</Clickable>
       </WorkspaceCardBase>;
     };
 
+const SurveyCard: React.FunctionComponent<{survey}> = ({survey}) => {
+  return <WorkspaceCardBase style={{minWidth: '11rem', maxHeight: 'auto', width: '11.5rem'}}>
```
blrubenstein: What's the intent behind minWidth of 11, but width of 11.5?

ui/src/app/views/concept-homepage.tsx
```diff
@@ -84,12 +101,27 @@ const DomainCard: React.FunctionComponent<{conceptDomainInfo: DomainInfo,
...
+    <div style={styles.domainBoxHeader} data-test-id='survey-box-name'>{survey.name}</div>
+    <div style={styles.conceptText}>
+      <span style={{fontSize: '30px'}}>{survey.questionCount}</span> survey questions with
```
blrubenstein: Nit: this can just be 30, rather than '30px', pixels are implicit.

ui/src/app/views/concept-homepage.tsx
```diff
@@ -159,33 +194,17 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
-        });
-        this.setState({
-          conceptsCache: conceptsCache,
-          conceptDomainCounts: conceptDomainCounts,
-          selectedDomain: conceptDomainCounts[0],
-        });
+        const [conceptDomainInfo, surveysInfo] = await Promise.all([
+          conceptsApi().getDomainInfo(namespace, id),
+          conceptsApi().getSurveyInfo(namespace, id)]);
+        this.setState({conceptSurveysList: surveysInfo.items});
```
blrubenstein: I would probably try to put this into the setState called by loadConceptDomainInfo, to go down to a single setState call, rather than two.

blrubenstein: It may be worth not abstracting the loadConceptDomainInfo into another function.

ui/src/app/views/concept-homepage.tsx
```diff
@@ -194,6 +213,29 @@ export const ConceptHomepage = withCurrentWorkspace()(
       }
     }
 
+    loadConceptDomainInfo(conceptDomainInfo) {
+      const conceptsCache: ConceptCacheItem[] = [];
+      const conceptDomainCounts: DomainCount[] = [];
+      conceptDomainInfo.items.forEach((domain) => {
```
blrubenstein: Rather than do this as a forEach, it may be worth doing these as two maps. (For simplicity reasons.)

That might make it feel more reasonable inlined.

ui/src/app/views/concept-homepage.tsx
```diff
@@ -416,14 +458,27 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
+                                         standardConceptsOnly={standardConceptsOnly}
+                                         browseInDomain={() => this.browseDomain(domain)}
+                                         key={i} data-test-id='domain-box'/>;
+                  })}
+                  </div>
+                  <div style={styles.sectionHeader}>
+                    Survey Questions
+                  </div>
+                  <div style={styles.cardList}>
+                    {conceptSurveysList.map((surveys, i) => {
```
blrubenstein: If we aren't using `i`, we don't need it in the map

ui/src/app/views/concept-homepage.tsx
NehaBroad: I left it as is because this story is just first step for surveys (was not sure if we would need more logic for survey stories)

ui/src/app/views/concept-homepage.tsx
NehaBroad: sure. This was pre existing i'll change this

ui/src/app/views/concept-homepage.tsx
NehaBroad: I am thinking about it i might put it back for this story , however in future if there is any manipulation to survey data we should separate out the logic in two different functions

ui/src/app/views/concept-homepage.tsx
```diff
@@ -159,17 +194,18 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
       try {
         const conceptsCache: ConceptCacheItem[] = [];
         const conceptDomainCounts: DomainCount[] = [];
-        const resp = await conceptsApi().getDomainInfo(namespace, id);
-        this.setState({conceptDomainList: resp.items});
-        resp.items.forEach((domain) => {
+        const [conceptDomainInfo, surveysInfo] = await Promise.all([
+          conceptsApi().getDomainInfo(namespace, id),
+          conceptsApi().getSurveyInfo(namespace, id)]);
+        conceptDomainInfo.items.map((domain) => {
```
blrubenstein: Sorry, I might have miscommunicated this :( 

I would have done above: `const conceptsCache: ConceptCacheItem[] = conceptDomainInfo.items.map...` and `const conceptDomainCounts: DomainCount[] = conceptDomainInfo.items.map...`

PR #2209, opened 2019-06-06
[risk=no] fix research purpose regression on ws about page

jacmrob:
Updates to the workspace create/edit page with new research purpose fields created a regression in the workspace About page causing nothing to show up underneath the "Research Purpose" header.  This PR fixes it so that it displays the same information that it used to (the "description" field was renamed "intendedStudy" and reallocated to that rp question).  The `ResearchPurposeItems` object also changed into a list with new fields. 

Although this page will change with the release of the new About page with dataset builder, this fixes things until Bedford.

### DIFF
--- a/ui/src/app/views/workspace/component.ts
+++ b/ui/src/app/views/workspace/component.ts
@@ -105,11 +105,14 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     this.workspace.userRoles = fp.sortBy('familyName', workspace.userRoles);
     this.accessLevel = workspace.accessLevel;
     this.researchPurposeArray = [];
-    Object.keys(ResearchPurposeItems).forEach((key) => {
-      if (this.workspace.researchPurpose[key]) {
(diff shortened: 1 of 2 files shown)


ui/src/app/views/workspace/component.ts
```diff
@@ -105,12 +105,15 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     this.workspace.userRoles = fp.sortBy('familyName', workspace.userRoles);
     this.accessLevel = workspace.accessLevel;
     this.researchPurposeArray = [];
-    Object.keys(ResearchPurposeItems).forEach((key) => {
-      if (this.workspace.researchPurpose[key]) {
-        let shortDescription = ResearchPurposeItems[key].shortDescription;
-        if (key === 'diseaseFocusedResearch') {
+    ResearchPurposeItems.forEach((key) => {
```
ericsong: `key` sounds like it might be a misnomer? Its type is the item right?

PR #2215, opened 2019-06-06
[RW-2594][risk=no] New research purpose on ws About page

jacmrob:

### DIFF
--- a/ui/src/app/components/tables.tsx
+++ b/ui/src/app/components/tables.tsx
@@ -1,3 +1,4 @@
+import colors from 'app/styles/colors';
 import {nextSort} from 'app/utils';
 import * as fp from 'lodash/fp';
 import * as React from 'react';
@@ -29,6 +30,10 @@ const styles = {
   paginationRow: {
     display: 'flex', alignItems: 'center', height: rowHeight,
     border: '1px solid #ccc', backgroundColor: '#fafafa', fontSize: 13
+  },
+  paddedTable: {
+    color: colors.purple[0], fontSize: 13, lineHeight: '24px',
+    background: colors.white, padding: '0.2rem 0.5rem', display: 'flex'
   }
 };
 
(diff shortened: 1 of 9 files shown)


ui/src/app/components/tables.tsx
```diff
@@ -114,3 +119,32 @@ export const TextCell = ({style = {}, ...props}) => {
...
 };
+
+export const PaddedTableCell = ({left, content, leftWidth = '50%', rightWidth = '50%'}) => {
+  return <div style={{padding: '.2rem', width: left ? leftWidth :
+      rightWidth, display: 'flex'}}>
+    <div style={{...styles.paddedTable, width: '100%'}}>{content}</div>
+  </div>;
+};
+
+export const TwoColPaddedTable = ({style = {}, header = false, headerLeft = '',
```
gjuggler: +1 on the effort to create a reusable component here – looks nice.

jacmrob:
Screen Shot 2019-06-07 at 9 34 29 AM

ui/src/app/views/workspace/component.ts
```diff
@@ -50,6 +51,7 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
   bugReportOpen: boolean;
   bugReportDescription = '';
   googleBucketModal = false;
+  showUpdatedResearchPurpose = environment.enableDatasetBuilder;
```
gjuggler: The tying together of these features is a bit surprising – your offline explanation makes total sense, but the code here might benefit from a quick comment (and maybe a link to the dataset builder Epic) saying why we're tying these together and when it should be removed.

ui/src/app/views/workspace-about.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <div style={styles.mainHeader}>Research Purpose
+          <Clickable disabled={!workspacePermissions.canWrite}
+                     data-test-id='edit-workspace'
+                     onClick={() => navigate(
+                       ['workspaces',  workspace.namespace, workspace.id, 'edit'])}>
+            <EditComponentReact disabled={!workspacePermissions.canWrite}
+                                style={{marginTop: '0.1rem'}}/>
+          </Clickable>
+        </div>
+        <div style={styles.sectionText}>{ResearchPurposeDescription}</div>
```
gjuggler: I know Lou's mocks included this, but it feels kind of awkward to have such a long piece of boilerplate text being shown to all users who ever visit any workspace. Maybe it's worth quickly checking w/ Karthik or Lou on whether this is really the intent?

If we do need to keep this in, I'd argue it should be towards the bottom, so more useful content (e.g. the workspace description) is above the fold.

ui/src/app/views/workspace/component.ts
jacmrob: will add a comment

ui/src/app/views/workspace-about.tsx
jacmrob: I totally agree :/ will check with karthik

PR #2217, opened 2019-06-07
[RW-2657][risk=moderate] Grant canCompute to workspace editors/owners

calbach:
- Upon sharing, this enables writers/owners to launch clusters in that project.
- This will not actually be utilized yet within the application.
- Recode the reset cluster button, as this would have created conflicting clusters in the wrong project before we launched `useProjectBillingBuffer`

Will backfill these roles as a follow-up.

FYI @jmthibault79

### DIFF
--- a/ui/src/app/views/workspace/component.ts
+++ b/ui/src/app/views/workspace/component.ts
@@ -208,6 +208,8 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
 
   workspaceClusterBillingProjectId(): string {
     if (this.useBillingProjectBuffer === undefined) {
+      // The server config hasn't loaded yet, we don't yet know which billing
+      // project should be used for clusters.
       return null;
     }
     if (!this.useBillingProjectBuffer) {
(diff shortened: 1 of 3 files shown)


ui/src/app/views/workspace/component.ts
```diff
@@ -200,14 +207,16 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
   }
 
   workspaceClusterBillingProjectId(): string {
-    if (this.workspace.namespace === this.freeTierBillingProject) {
+    if (this.useBillingProjectBuffer === undefined) {
```
ericsong: Took me a few minutes trying to figure out why we should return `null` in this case. Do you think it's worth adding a comment or adding a mini function to capture the context?

```
isServerConfigLoaded() { return this.useBillingProjectBuffer !== undefined } 
```

calbach: Added a comment, this is the downside of getting flags from the server, though hopefully our React version of this will be a bit easier to use synchronously. In hindsight, we should have just provided an instance of the serverConfig for injection rather than using this service.

PR #2219, opened 2019-06-07
[risk=no] Fix invalid value for GAE_MIN_INSTANCES

ericsong:


calbach:
0 is valid - I suspect this is another case of a local gcloud version being off: https://github.com/all-of-us/data-browser/pull/213

PR #2226, opened 2019-06-11
[RW-2817][P1] fix null pointer bugs on ws edit 

jacmrob:
fixed two null pointer errors on workspace edit page:
* running "includes" on null object when checking if the "other population" was selected
* checking if we have any specific populations selected when verifying all fields filled for the save button.  (this one happened when we had "no" selected in the initial ws then tried to edit it to say "yes"; see below comment)

note that these happened on duplicate and edit because we are fetching a ws object that has already been created, and when these fields are left empty on initial create they are returned as null.

### DIFF
--- a/ui/src/app/views/workspace-edit.tsx
+++ b/ui/src/app/views/workspace-edit.tsx
@@ -461,7 +461,8 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
 
     get noSpecificPopulationSelected() {
       return this.state.workspace.researchPurpose.population &&
-        this.state.workspace.researchPurpose.populationDetails.length === 0;
+        (!this.state.workspace.researchPurpose.populationDetails ||
+          this.state.workspace.researchPurpose.populationDetails.length === 0);
     }
 
     get noDiseaseOfFocusSpecified() {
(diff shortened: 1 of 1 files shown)


ui/src/app/views/workspace-edit.tsx
```diff
@@ -740,8 +741,8 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                    disabled={!this.state.workspace.researchPurpose.population}/>
                 <TextInput type='text' autoFocus placeholder='Please specify'
                            value={this.state.workspace.researchPurpose.otherPopulationDetails}
-                           disabled={!this.state.workspace.researchPurpose.populationDetails
-                             .includes(SpecificPopulationEnum.OTHER)}
+                           disabled={!fp.includes(SpecificPopulationEnum.OTHER,
```
jacmrob: fp includes handles null object

gjuggler:
It looks like this might solve https://precisionmedicineinitiative.atlassian.net/browse/RW-2817 which is blocking today's release. Jackie, does that seem right? If so, please attach this PR to RW-2817 and I'll cherrypick it in.

jacmrob:
@gjuggler attached.  am merging to master

PR #2229, opened 2019-06-12
[risk=no][RW-2474] Billing Project buffer error handling

ericsong:
"Error" Handling.

All its really doing is finding projects that have been stuck in the temporary states for too long and set them to ERROR.

The timeout for CREATING is 60m and ASSIGNING is 10m.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -6,9 +6,11 @@ import static org.pmiops.workbench.db.model.BillingProjectBufferEntry.BillingPro
 import static org.pmiops.workbench.db.model.BillingProjectBufferEntry.BillingProjectBufferStatus.CREATING;
 import static org.pmiops.workbench.db.model.BillingProjectBufferEntry.BillingProjectBufferStatus.ERROR;
 
+import com.google.common.collect.Iterables;
 import com.google.common.hash.Hashing;
 import java.sql.Timestamp;
 import java.time.Clock;
+import java.time.temporal.ChronoUnit;
 import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/db/model/BillingProjectBufferEntry.java
```diff
@@ -84,7 +95,9 @@ public BillingProjectBufferStatus getStatusEnum() {
     return StorageEnums.billingProjectBufferStatusFromStorage(status);
   }
 
-  public void setStatusEnum(BillingProjectBufferStatus status) {
+  public void setStatusEnum(
+      BillingProjectBufferStatus status, Supplier<Timestamp> currentTimestamp) {
```
calbach: This is definitely on the fence for me. We don't really have business logic in these model classes anywhere else in the code AFAIK, but it also makes all the call sites pretty clean and ensures we don't miss any. Of course, someone could go straight to the `setStatus()` method and bypass this.

api/src/main/resources/workbench.yaml
```diff
@@ -1078,6 +1078,22 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
+  /v1/cron/cleanBillingBuffer:
+    get:
+      security: []
+      tags:
+        - billing
+        - cron
+      description: ind BillingProjectBufferEntries that have failed during the creation or assignment step and set their statuses to ERROR
```
calbach: Find

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -101,11 +110,34 @@ public void syncBillingProjectStatus() {
     }
   }
 
+  public void cleanBillingBuffer() {
+    Iterables.concat(
```
calbach: Nice

api/src/main/java/org/pmiops/workbench/db/model/BillingProjectBufferEntry.java
ericsong: Yeah... I wasn't sure about this either but I decided the protection against misuse is worth it. `setStatus()` and `setLastStatusChangedTime()` are actually both private so the only way to update either is through this method.

PR #2243, opened 2019-06-18
[RW-1381][risk=low] Delete user_workspace table

jacmrob:


jacmrob:
This deletes the user_workspace table and uses firecloud as the one source of truth for workspace collaborators.  Because of the performance issues with calling firecloud for ACLs every time we load a workspace/list workspaces, I got rid of the `userRoles` attribute of workspaces and created an endpoint for fetching ACLs, so that we only fetch them when we actually need them (for showing collaborators in workspace about & for doing workspace share)

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -127,6 +126,22 @@ public Workspace get(String ns, String firecloudName) {
                 fcWorkspace -> fcWorkspace));
   }
 
+  @Override
+  public Map<User, WorkspaceAccessEntry> getFirecloudWorkspaceAcls(Workspace workspace) {
```
blrubenstein: Lets simplify this method to take just the namespace and firecloud name rather than the full workspace.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -763,4 +726,15 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
 
     return ResponseEntity.ok(fileDetail);
   }
+
+  @Override
+  public ResponseEntity<WorkspaceUserRolesResponse> getWorkspaceUserRoles(
+      String workspaceNamespace, String workspaceId) {
+    org.pmiops.workbench.db.model.Workspace dbWorkspace =
+        workspaceService.getRequired(workspaceNamespace, workspaceId);
+    List<UserRole> userRoles = workspaceService.getWorkspaceUserRoles(dbWorkspace);
```
blrubenstein: Can we rename this method to `getFirecloudWorkspaceUserRoles` just to be clear that its a call to FireCloud to load this information.

blrubenstein: I would pass in the map here, and have this just be a helper that converts a Map<User, Acl> to List<Acl>

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -127,6 +126,22 @@ public Workspace get(String ns, String firecloudName) {
...
+    WorkspaceACL firecloudWorkspaceAcls =
+        fireCloudService.getWorkspaceAcl(
```
blrubenstein: Opt: this is the only place we call getWorkspaceAcl, could the logic after this fold into `fireCloudService.getWorkspaceAcl` and then have the firecloudService method return the data structure we want?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -334,4 +352,17 @@ public Workspace findByWorkspaceId(long workspaceId) {
...
+  @Override
+  public List<UserRole> getWorkspaceUserRoles(Workspace workspace) {
+    Map<User, WorkspaceAccessEntry> rolesMap = getFirecloudWorkspaceAcls(workspace);
+    List<UserRole> userRoles = new ArrayList<>();
+    for (Map.Entry<User, WorkspaceAccessEntry> entry : rolesMap.entrySet()) {
+      userRoles.add(workspaceMapper.toApiUserRole(entry));
+    }
+    return userRoles.stream()
+        .sorted(
+            Comparator.comparing(UserRole::getRole).thenComparing(UserRole::getEmail).reversed())
```
blrubenstein: Nice!

ui/src/app/views/workspace/component.ts
```diff
@@ -140,6 +141,19 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     this.wsId = wsid;
   }
 
+  private loadUserRoles() {
+    workspacesApi().getWorkspaceUserRoles(this.wsNamespace, this.wsId)
+      .then(
+        resp => {
+          this.workspaceUserRoles = fp.sortBy('familyName', resp.items);
```
blrubenstein: We no longer need to sort here, right? Because we're sorting on the server-side?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -127,6 +126,22 @@ public Workspace get(String ns, String firecloudName) {
...
+  public Map<User, WorkspaceAccessEntry> getFirecloudWorkspaceAcls(Workspace workspace) {
+    WorkspaceACL firecloudWorkspaceAcls =
+        fireCloudService.getWorkspaceAcl(
+            workspace.getWorkspaceNamespace(), workspace.getFirecloudName());
+    Map<String, Object> aclsMap = (Map) firecloudWorkspaceAcls.getAcl();
+    Map<User, WorkspaceAccessEntry> userToAcl = new HashMap<>();
+    for (Map.Entry<String, Object> entry : aclsMap.entrySet()) {
+      WorkspaceAccessEntry acl =
+          new Gson().fromJson(entry.getValue().toString(), WorkspaceAccessEntry.class);
+      User currentUser = userDao.findUserByEmail(entry.getKey());
```
calbach: Is it necessary to fetch users like this, or could we just return strings? It's going to get slow for large ACL lists to be hitting the database in a for loop like this.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -199,57 +214,60 @@ public void setResearchPurposeApproved(String ns, String firecloudName, boolean
...
-    Iterator<WorkspaceUserRole> dbUserRoles = workspace.getWorkspaceUserRoles().iterator();
-    while (dbUserRoles.hasNext()) {
-      WorkspaceUserRole currentUserRole = dbUserRoles.next();
+  }
+
+  @Override
+  public Workspace updateUserRoles(
+      Workspace workspace, Map<User, WorkspaceAccessLevel> userRoleMap) {
+    // userRoleMap is a map of the new permissions for ALL users on the ws
+    Map<User, WorkspaceAccessEntry> aclsMap = getFirecloudWorkspaceAcls(workspace);
```
calbach: Using an etag here would be ideal; does FC support it? If not can we file the FR? Either that or we track the delta client-side and only send that (could be a follow-up ticket).

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -199,57 +214,60 @@ public void setResearchPurposeApproved(String ns, String firecloudName, boolean
     saveWithLastModified(workspace, now);
   }
 
-  @Override
-  public Workspace updateUserRoles(Workspace workspace, Set<WorkspaceUserRole> userRoleSet) {
-    Map<Long, WorkspaceUserRole> userRoleMap = new HashMap<Long, WorkspaceUserRole>();
-    for (WorkspaceUserRole userRole : userRoleSet) {
-      userRole.setWorkspace(workspace);
-      userRoleMap.put(userRole.getUser().getUserId(), userRole);
+  private void updateFirecloudAclsOnUser(
```
calbach: This method doing mutation feels a bit awkward, I would probably restructure this as:

```
static WorkspaceACLUpdate accessLevelToWorkspaceACLUpdate(accessLevel)
```

Optionally include user email there, or just add it after.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -79,7 +81,8 @@ public FireCloudServiceImpl(
       Provider<BillingApi> billingApiProvider,
       Provider<GroupsApi> groupsApiProvider,
       Provider<NihApi> nihApiProvider,
-      Provider<WorkspacesApi> workspacesApiProvider,
+      @Qualifier("workspacesApi") Provider<WorkspacesApi> workspacesApiProvider,
+      @Qualifier("workspaceAclsApi") Provider<WorkspacesApi> workspaceAclsApiProvider,
```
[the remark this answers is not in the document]
jacmrob: i will check and add

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jacmrob: done; this fixed my testing bug 👍

PR #2245, opened 2019-06-19
[RW-2851][risk=no] Add Surveys and PM to CB list search

dolbeew:
**These feature flags need to be enabled to test this PR:
-`enableListSearch` in [/api/config/config_local.json](https://github.com/all-of-us/workbench/blob/master/api/config/config_local.json#L64)
-`enableCBListSearch ` in [/ui/src/environments/environment.local.ts](https://github.com/all-of-us/workbench/blob/master/ui/src/environments/environment.local.ts#L21)

Adds Surveys and Physical Measurements options in the criteria dropdown for new list search UI. These do not use the new list search workflow and will open the wizard directly to the tree view. No design or functionality changes, these should work the same as the current wizards.

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -63,7 +63,7 @@
     "unsafeAllowSelfBypass": true
   },
   "cohortbuilder": {
-    "enableListSearch": true
+    "enableListSearch": false
   },
   "featureFlags": {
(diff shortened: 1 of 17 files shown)


dolbeew:
Closing and including these changes in the PR for RW-2852

PR #2248, opened 2019-06-20
[RW-2752][RISK=NO] Surveys in Concept Set Selector - See Answers expansion

NehaBroad:
Pagination of answer will be coming in the next PR which will also cover % answered and participation count


api/src/main/java/org/pmiops/workbench/cdr/ConceptBigQueryService.java
```diff
@@ -108,8 +130,30 @@ private QueryJobConfiguration buildSurveyQuery(String survey) {
...
+        bigQueryService.executeQuery(
+            bigQueryService.filterBigQueryConfig(buildSurveyAnswerQuery(questionConceptId)),
+            360000L);
+    result
+        .getValues()
+        .forEach(
+            surveyValue -> {
+              SurveyAnswerResponse answer = new SurveyAnswerResponse();
+              answer.setParticipationCount(
+                  Long.parseLong(surveyValue.get(2).getValue().toString()));
```
blrubenstein: What happens if any of these are null? Can they be null?

api/src/main/resources/workbench.yaml
```diff
@@ -1569,26 +1569,50 @@ paths:
...
+          type: string
+      responses:
+        200:
+          description: information about the surveys
+          schema:
+            type: "array"
+            items:
+              $ref: "#/definitions/SurveyQuestionsResponse"
+
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/getSurveyAnswer:
```
blrubenstein: Nit: Following above pattern, this should probably be `/surveyAnswers`

ui/src/app/views/survey-details.tsx
```diff
@@ -77,7 +84,17 @@ export const SurveyDetails = withCurrentWorkspace()(
         const {workspace, surveyName} = this.props;
         const surveys = await conceptsApi().getSurveyQuestions(
           workspace.namespace, workspace.id, surveyName);
-        this.setState({surveyList: surveys, loading: false});
+        const seeMyAnsList = [];
+        surveys.forEach(survey => seeMyAnsList.push(false));
+        const surveyDetails = [];
+        surveys.map((survey) => {
```
blrubenstein: `.map` in general should be used for mutating an array. I would probably do this as:
```
const surveyDetails = surveys.map((survey => {
  return {
    question: survey,
    answer: []
  })
```

(Or even cut out the intermediate variable and just put the map in the setState)

PR #2252, opened 2019-06-21
[RW-2736][risk=no]Convert cards and breadcrumbs to anchor links

jmthibault79:
With this change, cards (like notebooks, data sets, cohorts) and breadcrumbs are now anchor links.  This enables user behavior like right-click, middle-click, shift-click, and command-click.

### DIFF
--- a/ui/src/app/utils/navigation.tsx
+++ b/ui/src/app/utils/navigation.tsx
@@ -58,6 +58,15 @@ export const navigateByUrl = (...args) => {
   return NavStore.navigateByUrl(...args);
 };
 
+// if modifier keys are pressed (like shift or cmd) use the href
+// if no keys are pressed, prevent default behavior and route using navigateByUrl
+export const navigateAndPreventDefaultIfNoKeysPressed = (e: React.MouseEvent, url: string) => {
+  if (!(e.shiftKey || e.altKey || e.ctrlKey || e.metaKey)) {
+    e.preventDefault();
+    navigateByUrl(url);
+  }
+};
+
 export enum BreadcrumbType {
   Workspaces = 'Workspaces',
   Workspace = 'Workspace',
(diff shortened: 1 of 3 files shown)


ui/src/app/views/resource-card.tsx
```diff
@@ -234,12 +234,6 @@ export class ResourceCard extends React.Component<Props, State> {
     }
   }
 
-  reviewCohort(): void {
```
jmthibault79: Moved to `getResourceUrl()`

ui/src/app/views/resource-card.tsx
```diff
@@ -439,51 +433,47 @@ export class ResourceCard extends React.Component<Props, State> {
...
-        navigate(
-          ['workspaces', this.props.resourceCard.workspaceNamespace,
-            this.props.resourceCard.workspaceFirecloudName, 'notebooks',
-            encodeURIComponent(this.props.resourceCard.notebook.name)], {
-              queryParams,
-              relativeTo: null,
-            });
-        break;
+        return `${workspacePrefix}/notebooks/${encodeURIComponent(notebook.name)}?` +
+          fp.toPairs(queryParams).map(kvp => `${kvp[0]}=${kvp[1]}`).join('&');
```
gjuggler: Optional: there should be a simpler & safer way to render a set of query params to string... maybe use the URL API (https://developer.mozilla.org/en-US/docs/Web/API/URL) with

new URLSearchParams(queryParams).toString()

ui/src/app/utils/navigation.tsx
```diff
@@ -58,6 +58,15 @@ export const navigateByUrl = (...args) => {
   return NavStore.navigateByUrl(...args);
 };
 
+// if modifier keys are pressed (like shift or cmd) use the href
+// if no keys are pressed, prevent default behavior and route using navigateByUrl
+export const navigateAndPreventDefaultIfNoKeysPressed = (e: React.MouseEvent, url: string) => {
```
gjuggler: Optional: a slightly shorter / pithier name could be "navigateOrAllowClick"? I'm not sure if it still correctly reflects the intent of the method, but I found the name as-is to be a bit of a mouthful (to the point where it detracts from readability / comprehension)

PR #2256, opened 2019-06-25
[RW-2855][risk=no] add icon links for cohort and concept sets to dataset builder

andy7i:
Adding these links:

image-20190618-145124


ui/src/app/views/dataset-page.tsx
```diff
@@ -43,9 +45,17 @@ export const styles = reactStyles({
...
     color: colors.blue[7],
-    borderBottom: '1px solid #E5E5E5'
+    borderBottom: '1px solid #E5E5E5',
+    display: 'flex',
+    justifyContent: 'space-between',
+    flexDirection: 'row'
+  },
+
+  selectBoxHeaderIconLinks: {
+    fill: '#216FB4'
```
NehaBroad: Can you use colors.tsx here

ui/src/app/views/dataset-page.tsx
```diff
@@ -43,9 +45,17 @@ export const styles = reactStyles({
     fontSize: '16px',
     height: '2rem',
     lineHeight: '2rem',
-    paddingLeft: '13px',
+    paddingLeft: '0.55rem',
```
jmthibault79: what do these style changes do?

ui/src/app/views/dataset-page.tsx
```diff
@@ -460,6 +480,13 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
               <div style={{backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
                 <div style={styles.selectBoxHeader}>
                   Cohorts
+                    <a data-test-id='cohorts-link' href={cohortsPath}
```
jmthibault79: Could these two use a common component?

ui/src/app/views/dataset-page.tsx
```diff
@@ -504,12 +538,12 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
                   </div>
                 </div>
                 <div style={{width: '40%'}}>
-                  <div style={{...styles.selectBoxHeader, display: 'flex'}}>
+                  <div style={styles.selectBoxHeader}>
```
jmthibault79: similar questions on styles

ui/src/app/views/dataset-page.tsx
andy7i: >paddingLeft: '0.55rem',
- convert px to rem since rem is better to use than px when possible.
> paddingRight: '0.55rem',
- pad the right side the same amount as the left.

```
display: 'flex',
    justifyContent: 'space-between',
    flexDirection: 'row'
```
these attributes space elements so that they are as far apart as possible.

```
  selectBoxHeaderIconLinks: {
    fill: colors.blue[9]
  },	  },
```
defining the icon color in one location instead of 2 places

ui/src/app/views/dataset-page.tsx
```diff
@@ -539,7 +573,7 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
         </FadeBox>
         <FadeBox style={{marginTop: '1rem'}}>
           <div style={{backgroundColor: 'white', border: '1px solid #E5E5E5'}}>
-            <div style={{...styles.selectBoxHeader, ...styles.refreshPreviewHeader}}>
+            <div style={styles.previewDataHeaderBox}>
```
andy7i: use a single style instead of using 2 disparate styles

ui/src/app/views/dataset-page.tsx
andy7i: moved the display flex into selectBoxHeader since it's used in all selectBoxHeaders now

ui/src/app/views/dataset-page.tsx
andy7i: good idea!

PR #2258, opened 2019-06-25
[RW-2844][RISK=NO]Make export to Notebooks selected by default when saving Dataset

NehaBroad:
Default update dataset modal 
Screen Shot 2019-06-25 at 12 51 38 PM

Button text for default update 
Screen Shot 2019-06-25 at 12 51 26 PM

Button text if data set is updated
Screen Shot 2019-06-25 at 12 51 33 PM

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -154,7 +154,7 @@ public class DataSetController implements DataSetApiDelegate {
               now);
       return ResponseEntity.ok(TO_CLIENT_DATA_SET.apply(savedDataSet));
     } catch (DataIntegrityViolationException ex) {
-      throw new ConflictException("Data set with the same name already exist");
+      throw new ConflictException("Data set with the same name already exists");
     }
   }
 
@@ -486,7 +486,7 @@ public class DataSetController implements DataSetApiDelegate {
     } catch (OptimisticLockException e) {
       throw new ConflictException("Failed due to concurrent concept set modification");
     } catch (DataIntegrityViolationException ex) {
-      throw new ConflictException("Data set with the same name already exist");
+      throw new ConflictException("Data set with the same name already exists");
     }
 
     return ResponseEntity.ok(TO_CLIENT_DATA_SET.apply(dbDataSet));
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -485,6 +485,8 @@ public class DataSetController implements DataSetApiDelegate {
       // TODO: add recent resource entry for data sets
     } catch (OptimisticLockException e) {
       throw new ConflictException("Failed due to concurrent concept set modification");
+    } catch (DataIntegrityViolationException ex) {
+      throw new ConflictException("Data set with the same name already exist");
     }
 
     return ResponseEntity.ok(TO_CLIENT_DATA_SET.apply(dbDataSet));
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -1662,6 +1662,10 @@ paths:
           description: The updated data set.
           schema:
             $ref: "#/definitions/DataSet"
+        409:
+          description: Data set with same name exist
+          schema:
+            $ref: "#/definitions/ErrorResponse"
     delete:
       tags:
         - dataSet
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -1627,7 +1627,7 @@ paths:
             schema:
               $ref: "#/definitions/ErrorResponse"
           409:
-            description: Data set with same name exist
+            description: Data Set with the same name already exists
             schema:
               $ref: "#/definitions/ErrorResponse"
 
@@ -1663,7 +1663,7 @@ paths:
           schema:
             $ref: "#/definitions/DataSet"
         409:
-          description: Data set with same name exist
+          description: Data Set with the same name already exists
           schema:
             $ref: "#/definitions/ErrorResponse"
     delete:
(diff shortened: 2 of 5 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -1662,6 +1662,10 @@ paths:
           description: The updated data set.
           schema:
             $ref: "#/definitions/DataSet"
+        409:
+          description: Data set with same name exist
```
andy7i: I like the idea of using the same message as above
```suggestion
          description: Data set with the same name already exists
```

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -485,6 +485,8 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
       // TODO: add recent resource entry for data sets
     } catch (OptimisticLockException e) {
       throw new ConflictException("Failed due to concurrent concept set modification");
+    } catch (DataIntegrityViolationException ex) {
+      throw new ConflictException("Data set with the same name already exist");
```
andy7i:
```suggestion
      throw new ConflictException("Data set with the same name already exists");
```

ui/src/app/views/new-dataset-modal.tsx
```diff
@@ -227,7 +257,8 @@ class NewDataSetModal extends React.Component<Props, State> {
         <div style={{display: 'flex', alignItems: 'center', marginTop: '1rem'}}>
           <CheckBox style={{height: 17, width: 17}}
                     data-test-id='export-to-notebook'
-                    onChange={() => this.changeExportToNotebook()} />
+                    onChange={() => this.changeExportToNotebook()}
+                    checked={this.state.exportToNotebook} />
           <div style={{marginLeft: '.5rem',
             color: colors.black[0]}}>Export to notebook</div>
```
andy7i: probably not in the scope of this ticket, but it would be nice if this text `Export to notebook` could be clicked to check the checkbox. Currently, clicking the text does nothing.

ui/src/app/views/dataset-page.tsx
```diff
@@ -560,11 +544,11 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-              <Button data-test-id='save-button' style={{marginRight: '1rem'}}
-                      onClick={this.editing ? () => this.updateDataSet() :
-                () => this.setState({openSaveModal: true})}
-                disabled={this.disableSave() || (this.editing && !dataSetTouched)}>
-                {this.editing ? 'UPDATE DATA SET' : 'SAVE DATA SET'}
+              <Button data-test-id='save-button' style={{position: 'absolute', right: '1rem',
+                top: '.25rem'}} onClick ={() => this.setState({openSaveModal: true})}
+                disabled={this.disableSave()}>
+                {this.editing ? !dataSetTouched ? 'Analyze' :
+                    'Update And Analyze' : 'Save And Analyze'}
```
andy7i: Clicked around here to test and it looks good.  One minor thing I noticed that's probably outside of the scope of this ticket:
when the user goes into an existing data set, the button shows `Analyze`. Then if the user makes a change, the button changes to `Update And Analyze`. This is great. If the user undoes the change they made, then there is no change to the data set but the button still says `Update And Analyze`. As a user, I would expect it to go back to saying `Analyze`, but this might be much more involved than what we want to do right now.

NehaBroad: Good point. Thats happening because we are just changing on the basis of any action rather than what the action actual is (like select or unselect rather than if the data set options is same as original). 
we/team can discuss more on this, and if we agree it should be another story as the edit indicator is same as what we currently have on test (update button is enabled even after selecting and unselecting)

ui/src/app/views/new-dataset-modal.tsx
NehaBroad: We can talk to lou/team but if we do agree to do this it should be consistent throughout the site

ui/src/app/views/new-dataset-modal.spec.tsx
```diff
@@ -59,7 +60,7 @@ describe('NewDataSetModal', () => {
     expect(exportSpy).not.toHaveBeenCalled();
   });
 
-  it('should not export if export not clicked', async() => {
+  it('should not export if export  clicked', async() => {
```
andy7i:
```suggestion
  it('should not export if export not clicked', async() => {
```

ui/src/app/views/new-dataset-modal.spec.tsx
```diff
@@ -131,4 +130,14 @@ describe('NewDataSetModal', () => {
       kernelType: KernelTypeEnum.Python
     });
   });
+  it ('should have default dataSet name if data set is passed as props', () => {
```
andy7i: nice

ui/src/app/views/new-dataset-modal.spec.tsx
NehaBroad: This is because now export functionality is selected by default, if its un checked it would mean user does not want to export notebook

andy7i: Oh that makes sense. how about this then?

```suggestion
  it('should not export if export not checked', async() => {
```

ui/src/app/views/dataset-page.tsx
blrubenstein: I don't think it would be bad for us to change this here. It shouldn't be hard to store an immutable dataSet in state, and then do a deep compare to see if its been touched. (And you can avoid changing any other code by making `get dataSetTouched` return the deep compare)

NehaBroad:
My concern for doing that right now is that there is a good chance of this implementation introducing bugs because we are adding complexity to it. Can we create another story for this for after bedford?

NehaBroad:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2919 created for taking care of the dataSetTouche indicator

PR #2260, opened 2019-06-25
[RW-2747][risk=low] publish workspace api

jacmrob:


api/db/changelog/db.changelog-86-published-workspaces.xml
```diff
@@ -0,0 +1,12 @@
...
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="jroberti" id="db.changelog-86-published-workspaces">
+        <addColumn tableName="workspace">
+            <column name="published" type="boolean" defaultValueBoolean="false"/>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
```
blrubenstein: Nit: Newline at eof

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -751,4 +758,39 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+
+  @Override
+  public ResponseEntity<WorkspaceResponseListResponse> getPublishedWorkspaces() {
+    WorkspaceResponseListResponse response = new WorkspaceResponseListResponse();
+
+    return ResponseEntity.ok(response);
+  }
+
+  @Override
+  @AuthorityRequired({Authority.WORKSPACE_ADMIN})
```
blrubenstein: Could we call this authority `FEATURED_WORKSPACE_ADMIN`? I imagine there are other things that will be separate workspace actions. (Like reviewing research purposes)

This is more about publishing than it is administering workspaces.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -368,12 +368,37 @@ public Workspace findByWorkspaceId(long workspaceId) {
...
+          entry.getKey(), WorkspaceAccessLevel.fromValue(entry.getValue().getAccessLevel()));
+    }
+
+    if (publish) {
+      updatedAcls.put(publishedWorkspaceGroup, WorkspaceAccessLevel.READER);
+    } else {
+      updatedAcls.put(publishedWorkspaceGroup, WorkspaceAccessLevel.NO_ACCESS);
+    }
+
+    return updateWorkspaceAcls(workspace, updatedAcls);
```
blrubenstein: Where does this update the ACLs in FC? Presumably at some point we need to make a post call?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -56,4 +56,7 @@ Workspace getWorkspaceEnforceAccessLevelAndSetCdrVersion(
 
   WorkspaceACLUpdate updateFirecloudAclsOnUser(
       WorkspaceAccessLevel updatedAccess, WorkspaceACLUpdate currentUpdate);
+
+  Workspace doWorkspacePublish(
```
calbach: `publishWorkspace`?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -368,12 +368,37 @@ public Workspace findByWorkspaceId(long workspaceId) {
...
+        log.log(Level.WARNING, "No user found for " + entry.getKey());
+      }
     }
     return userRoles.stream()
         .sorted(
             Comparator.comparing(UserRole::getRole).thenComparing(UserRole::getEmail).reversed())
         .collect(Collectors.toList());
   }
+
+  public Workspace doWorkspacePublish(
```
calbach: I think this should either `setPublished()` or not return a workspace

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -368,12 +368,37 @@ public Workspace findByWorkspaceId(long workspaceId) {
       Map<String, WorkspaceAccessEntry> rolesMap) {
     List<UserRole> userRoles = new ArrayList<>();
     for (Map.Entry<String, WorkspaceAccessEntry> entry : rolesMap.entrySet()) {
-      User user = userDao.findUserByEmail(entry.getKey());
-      userRoles.add(workspaceMapper.toApiUserRole(user, entry.getValue()));
+      // Filter out groups
+      try {
+        User user = userDao.findUserByEmail(entry.getKey());
+        userRoles.add(workspaceMapper.toApiUserRole(user, entry.getValue()));
+      } catch (NullPointerException e) {
```
calbach: Please don't catch NPEs, just do an if check on the `null` user if that's what this is checking

api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
```diff
@@ -85,6 +85,7 @@ public boolean equals(Object obj) {
   private Short activeStatus;
   private Short billingMigrationStatus =
       StorageEnums.billingMigrationStatusToStorage(BillingMigrationStatus.OLD);
+  private boolean published;
```
calbach: We wanted to filter these out of getWorkspaces - if that's not meant to be in this PR, is that work tracked separately?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jacmrob: `updateWorkspaceAcls` makes the firecloud call to update on the FC end

api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
jacmrob: unclear; i will check with karthik

jacmrob: published workspaces should not show up unless user is explicitly a reader/writer/owner on the ws.  changed getWorkspaces() to reflect that.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -294,9 +294,14 @@ public void setUp() {
     WorkbenchConfig testConfig = new WorkbenchConfig();
     testConfig.cohortbuilder = new WorkbenchConfig.CohortBuilderConfig();
     testConfig.cohortbuilder.enableListSearch = false;
+    testConfig.firecloud = new WorkbenchConfig.FireCloudConfig();
```
calbach: You can add this at L250 instead of exposing the configProvider like this.

The cohortReviewController setter also shouldn't be necessary, but that's an existing issue which I wouldn't advise digging into in this PR.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -126,6 +126,23 @@ public Workspace get(String ns, String firecloudName) {
...
+            workspaceResponse ->
+                !(workspaceResponse.getAccessLevel() != WorkspaceAccessLevel.OWNER
+                    && workspaceResponse.getWorkspace().getPublished()))
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public List<WorkspaceResponse> getPublishedWorkspaces() {
+    return getWorkspacesAndPublicWorkspaces().stream()
+        .filter(workspaceResponse -> (workspaceResponse.getWorkspace().getPublished()))
```
calbach: nit: unnecessary parens around the lambda

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -263,10 +280,15 @@ public Workspace updateWorkspaceAcls(
       } else {
         // This is how to remove a user from the FireCloud ACL:
         // Pass along an update request with NO ACCESS as the given access level.
-        WorkspaceACLUpdate removedUser = new WorkspaceACLUpdate();
-        removedUser.setEmail(currentUserEmail);
-        removedUser = updateFirecloudAclsOnUser(WorkspaceAccessLevel.NO_ACCESS, removedUser);
-        updateACLRequestList.add(removedUser);
+        // Note: do not do for all users group.  Unpublish will pass the specific NO_ACCESS acl
+        // TODO [jacmrob] : have all users pass NO_ACCESS explicitly? Handle filtering on frontend?
+        User foundUser = userDao.findUserByEmail(currentUserEmail);
```
calbach: First: See below comment - this usage is making me nervous; I think it would be better to avoid this method entirely.

Else: How about a simpler filter of checking the domain? This would not require serial database lookups running in a loop.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -685,4 +697,37 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+
+  @Override
+  @AuthorityRequired({Authority.FEATURED_WORKSPACE_ADMIN})
+  public ResponseEntity<EmptyResponse> publishWorkspace(
+      String workspaceNamespace, String workspaceId) {
+    org.pmiops.workbench.db.model.Workspace dbWorkspace =
+        workspaceService.getRequired(workspaceNamespace, workspaceId);
+
+    dbWorkspace = workspaceService.setPublished(dbWorkspace, getRegisteredUserDomainEmail(), true);
+    dbWorkspace.setPublished(true);
```
calbach: My previous comment was not addressed: why is this outside of the `setPublished` method?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -366,12 +388,38 @@ public Workspace findByWorkspaceId(long workspaceId) {
...
+          entry.getKey(), WorkspaceAccessLevel.fromValue(entry.getValue().getAccessLevel()));
+    }
+
+    if (publish) {
+      updatedAcls.put(publishedWorkspaceGroup, WorkspaceAccessLevel.READER);
+    } else {
+      updatedAcls.put(publishedWorkspaceGroup, WorkspaceAccessLevel.NO_ACCESS);
+    }
+
+    return updateWorkspaceAcls(workspace, updatedAcls);
```
calbach: This feels different enough from `updateWorkspaceAcls` that I would just call `fireCloudService.updateWorkspaceACL` directly with a single input. Most of what that method doing is not necessary for what you're trying to achieve here.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jacmrob: i think both are necessary bc of how we are calling the config in the controller

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
jacmrob: fixed it

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jacmrob: 👍

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -126,6 +126,23 @@ public Workspace get(String ns, String firecloudName) {
         .collect(Collectors.toList());
   }
 
+  @Override
+  public List<WorkspaceResponse> getWorkspaces() {
+    return getWorkspacesAndPublicWorkspaces().stream()
+        .filter(
+            workspaceResponse ->
+                !(workspaceResponse.getAccessLevel() != WorkspaceAccessLevel.OWNER
```
gjuggler: My reading of this line is that we'll include published workspaces in the "default" list only if the user is an OWNER on that workspace.

Should we also show them if the user is an EDITOR? It seems like the content team may be operating w/ a model where multiple folks are EDITORS, but only one individual is an OWNER. It may be odd to them if only the OWNER sees the workspace.

[Optional] Also – do we have any unit tests covering this logic? If it's not too much effort to add one, this logic feels just complicated enough to warrant some test coverage.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -281,10 +280,9 @@ public Workspace updateWorkspaceAcls(
       } else {
         // This is how to remove a user from the FireCloud ACL:
         // Pass along an update request with NO ACCESS as the given access level.
-        // Note: do not do for all users group.  Unpublish will pass the specific NO_ACCESS acl
+        // Note: do not do groups.  Unpublish will pass the specific NO_ACCESS acl
         // TODO [jacmrob] : have all users pass NO_ACCESS explicitly? Handle filtering on frontend?
-        User foundUser = userDao.findUserByEmail(currentUserEmail);
-        if (foundUser != null) {
+        if (!currentUserEmail.split("_", 0)[0].equals("GROUP")) {
```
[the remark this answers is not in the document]
jacmrob: let's do this as a follow-up ticket? I think we probably want to filter out for any GROUP email

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jacmrob: yes i think we should add editors.  i don't love how this will not include people who are explicit READERS vs those in the all users group, but there isn't really a way to do that now that we are only fetching acls from firecloud.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
jacmrob: Actually I fixed in this PR

PR #2261, opened 2019-06-26
[RW-2843][RISK=LOW]Behavior when a cohort/concept set referenced in a Dataset is deleted

NehaBroad:
Delete cohort/concept used in a dataset : should see the following warning message

Screen Shot 2019-06-26 at 12 29 26 PM

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -534,7 +534,7 @@ public class DataSetController implements DataSetApiDelegate {
   public ResponseEntity<Boolean> dataSetByIdExist(
       String workspaceNamespace, String workspaceId, String resourceType, Long id) {
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -441,6 +445,34 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
     return ResponseEntity.ok(response);
   }
 
+  @Override
+  public ResponseEntity<Boolean> markDirty(
+      String workspaceNamespace, String workspaceId, MarkDataSetRequest markDataSetRequest) {
+    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
+        workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
+    List<org.pmiops.workbench.db.model.DataSet> dbDataSet = new ArrayList<>();
```
blrubenstein: `dbDataSet` makes me think there's just one. Maybe `dbDataSetList`?

PR #2262, opened 2019-06-26
[RW-2771][Risk=no] Disable showing empty info for eRA commons

blrubenstein:


ui/src/app/views/profile-page.tsx
```diff
@@ -109,6 +109,8 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
   render() {
     const {profileState: {profile}} = this.props;
     const {profileEdits, updating} = this.state;
+    console.log('profile.eraCommonsLinkedNihUsername');
```
jacmrob: are these debug lines?  get rid of em?

blrubenstein:
The config is where we have decided to do API based feature flags.

PR #2265, opened 2019-06-26
[RW-1605][risk=no] Fix cluster load messaging on notebooks redirect

calbach:
When cluster is already running, don't show the "may take 10 minutes" language. Some minor cleanup so that we only set `progressComplete` on one of Unknown, Initializing, Resuming.

image

### DIFF
--- a/ui/src/app/views/notebook-redirect/component.spec.ts
+++ b/ui/src/app/views/notebook-redirect/component.spec.ts
@@ -158,6 +158,20 @@ describe('NotebookRedirectComponent', () => {
     expect(fixture.debugElement.queryAll(By.css('.i-frame')).length).toBe(1);
   }));
 
+  it('should display "Connecting" after initially finding a RUNNING cluster', fakeAsync(() => {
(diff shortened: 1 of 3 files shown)


ui/src/app/views/notebook-redirect/component.spec.ts
```diff
@@ -158,6 +158,20 @@ describe('NotebookRedirectComponent', () => {
...
+
+    updateAndTick(fixture);
+    fixture.detectChanges();
+
+    const initBox = fixture.debugElement.queryAll(By.css('#initializing'))[0];
+    expect(initBox.children[1].nativeElement.innerText)
+      .toContain('Connecting');
+
+    tick(10000);
+    fixture.detectChanges();
```
ericsong: whats being asserted after the tick and detectChanges?

calbach: Nothing - this was to clear an outstanding timer. Turns out only the `tick()` was actually needed - also added a comment.

PR #2266, opened 2019-06-26
[RW-2899][risk=no] Fix resource card lastModified time handling

calbach:
RecentResource uses a `string`; this was being coerced as a string number on the workspace page, while recent-work was broken.


calbach:
It handled it as well as the previous version did. That's not really the issue I was tackilng here, and I'm not sure whether or not it can happen - but doesn't hurt to cover it. Done.

PR #2267, opened 2019-06-26
[RW-2894][risk=no] Require registration for user search

calbach:
Update the testing pattern for UserControllerTest as well.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
@@ -1,6 +1,8 @@
 package org.pmiops.workbench.api;
 
 import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/pmiops/workbench/api/UserControllerTest.java
```diff
@@ -48,66 +52,69 @@
...
-    this.userController = new UserController(userService);
+
     saveFamily();
   }
 
-  @After
-  public void tearDown() {
-    userDao.deleteAll();
+  @Test(expected = ForbiddenException.class)
+  public void testEnforceRegisteredUnregistered() {
```
NehaBroad: Does it make sense to test when enforce register is false as well

calbach: Yes - the existing tests cover this (by default, enforceRegistered=false in these tests).

PR #2268, opened 2019-06-26
[RW-2897][risk=no] Clean up profile endpoint bit-rot

calbach:
- TOS is subsumed by DUA, remove all references
- Demo survey is not implemented, return this HTTP status for now


jacmrob:
need to fix one test in `ProfileControllerTest`

calbach:
I checked with Karthik on that yesterday. He said DUA subsumes TOS (for now). It's possible we'll need it in the future, but at this point that's speculative and having the code or field in the database (with values populated in some cases) will only add confusion at that point.

PR #2221, opened 2019-06-10
[RW-2657][risk=moderate] Implement canCompute backfill script

calbach:
https://precisionmedicineinitiative.atlassian.net/browse/RW-2657

Acting as the GAE default SA for the given environment:
```
for each owned workspace:
  ensure writers and owners have canCompute
``` 

- Required pulling in the getWorkspaceACL Firecloud swagger
- Includes a billing project filter to avoid touching local/old workspaces
- There are surprisingly many workspaces for which the GAE SA has "NO ACCESS", this warrants some follow-up investigation.

PR #2269, opened 2019-06-26
[risk=no] Refactor WorkspaceACL handling to avoid Java warning

calbach:
Slightly more automated parsing + gets rid of the unchecked Java warning.

Copied from my now deleted backfill script: https://github.com/all-of-us/workbench/pull/2221/files#diff-3e77476e30c3fda73b42c53d56c9f977R96

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
@@ -135,20 +135,16 @@ public class WorkspaceServiceImpl implements WorkspaceService {
                 fcWorkspace -> fcWorkspace));
   }
 
-  /**
-   * Swagger Java codegen does not handle the WorkspaceACL model correctly; it returns a GSON map
-   * instead. Run this through a typed Gson conversion process to parse into the desired type.
-   */
-  private static Map<String, WorkspaceAccessEntry> extractAclResponse(WorkspaceACL aclResp) {
-    Type accessEntryType = new TypeToken<Map<String, WorkspaceAccessEntry>>() {}.getType();
-    Gson gson = new Gson();
-    return gson.fromJson(gson.toJson(aclResp.getAcl(), accessEntryType), accessEntryType);
-  }
-
   @Override
   public Map<String, WorkspaceAccessEntry> getFirecloudWorkspaceAcls(
       String workspaceNamespace, String firecloudName) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -134,19 +135,20 @@ public Workspace get(String ns, String firecloudName) {
...
-        fireCloudService.getWorkspaceAcl(workspaceNamespace, firecloudName);
-    Map<String, Object> aclsMap = (Map) firecloudWorkspaceAcls.getAcl();
-    Map<String, WorkspaceAccessEntry> userToAcl = new HashMap<>();
-    for (Map.Entry<String, Object> entry : aclsMap.entrySet()) {
-      WorkspaceAccessEntry acl =
-          new Gson().fromJson(entry.getValue().toString(), WorkspaceAccessEntry.class);
-      userToAcl.put(entry.getKey(), acl);
-    }
-    return userToAcl;
+    return extractAclResponse(fireCloudService.getWorkspaceAcl(workspaceNamespace, firecloudName));
```
jacmrob: looks good except why make this its own method?  seems like you could just put the above code into `getFirecloudWorkspaceAcls` to consolidate, and add a comment for clarity

jacmrob:
seems like something went wrong in the return of that method; see the failed test.  i can give more context, but basically during clone when looking up the acls of the cloneFrom ws it is not finding anything, which is why it is returning a NO_ACCESS for each of the users who were on the ws (it thinks their acls were removed bc they are not found).   this is bc of how the method iterates; basically [this call](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java#L252) is returning empty which is then causing the code to go to the else clause.  so something from the changes you made are causing `getFirecloudWorkspaceAcls` to return null or something.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Done.

calbach:
The tests were not quite matching what is returned by Swagger codegen. It returns a Map, but more specifically a Gson TreeMap, which Gson can further parse. This is what my method was taking advantage of. I updated the tests to better match what happens on the server.

calbach:
(I also ran the server locally to verify)

PR #2270, opened 2019-06-27
[risk=medium][RW-2706] Read Only Notebooks

ericsong:
Notebooks will now load a read only preview by default. Clicking edit will redirect to the old notebooks page when the cluster is ready.


ui/src/app/services/notebooks-swagger-fetch-clients.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * This package configures/provides swagger "typescript-fetch" client
```
calbach: Maybe just reference the other file instead to avoid duplicating this documentation.

ui/src/app/views/interactive-notebook.tsx
```diff
@@ -0,0 +1,151 @@
...
+    return (
+      <div>
+        <div style={styles.navBar}>
+          <div style={{...styles.navBarItem, ...styles.active, width: 227}}>
+            Preview (Read-Only)
+          </div>
+          {this.state.userRequestedEditMode ? (
+            <div style={{...styles.navBarItem, width: 550}}>
+              <ClrIcon shape="sync" style={{...styles.navBarIcon, ...styles.rotate}}></ClrIcon>
+              Preparing your Jupyter environment. This may take up to 2 minutes.
```
calbach: We can do this as a follow-up, but this text should be dependent on the initial state of the cluster.

- Already running: don't show any "may take up to" text
- Resuming: "may take up to 2 minutes"
- Starting: "may take up to 10 mintues"

For now, I'd just hardcode "may take up to 10 minutes" and leave a TODO for the rest.

ui/src/app/app-routing.module.ts
```diff
@@ -125,6 +126,14 @@ const routes: Routes = [
                       breadcrumb: BreadcrumbType.Notebook,
                       minimizeChrome: true
                     }
+                  }, {
+                    path: 'v2/:nbName',
```
calbach: nit: I might just call this `preview` rather than v2

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -169,6 +174,20 @@ public void saveNotebook(String bucketName, String notebookName, JSONObject note
         notebookContents.toString().getBytes(StandardCharsets.UTF_8));
   }
 
+  @Override
+  public String getReadOnlyHtml(
+      String workspaceNamespace, String workspaceName, String notebookName) {
+    String bucketName =
```
calbach: nit: probably worth commenting that this call here is what is enforcing the ACLs for the notebook bucket

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -737,6 +738,14 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
     return ResponseEntity.ok(fileDetail);
   }
 
+  @Override
+  public ResponseEntity<ReadOnlyNotebookResponse> readOnlyNotebook(
+      String workspace, String workspaceName, String notebookName) {
+    ReadOnlyNotebookResponse response = new ReadOnlyNotebookResponse();
```
calbach: opt_nit: can chain this with the next line

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -344,6 +348,15 @@ public boolean isUserMemberOfGroup(String email, String groupName) {
         });
   }
 
+  @Override
+  public String staticNotebooksConvert(String notebook) {
+    return retryHandler.run(
+        (context) -> {
+          String html = staticNotebooksApiProvider.get().convertNotebook(notebook);
```
calbach: Can just return

opt: drop the brackets: `context -> staticNotebooksApiProvider.get().convertNotebook(notebook)`

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -169,6 +174,20 @@ public void saveNotebook(String bucketName, String notebookName, JSONObject note
...
+        fireCloudService
+            .getWorkspace(workspaceNamespace, workspaceName)
+            .getWorkspace()
+            .getBucketName();
+    String notebook =
+        getNotebookContents(bucketName, notebookName).toString(); // this might be wrong
+
+    return fireCloudService.staticNotebooksConvert(notebook);
```
calbach: This could be a follow-up, but we should have a sanitization step here. At a glance, this looks pretty straightforward to use: https://github.com/OWASP/java-html-sanitizer

nbconvert does do sanitization, but we should practice defense in depth

ui/src/app/app-routing.module.ts
ericsong: does `preview` make sense as more features like edit and playground mode get added?

calbach: Could go either way, putting "edit" / "playground" in the URL may make sense. Alternatively if at some point we move the Jupyter editor to this page and deprecate the current notebooks redirect page, we could rename this back to just "notebooks" and delete the other path.

ericsong: yeah. I was thinking we'll eventually move back to just `notebooks`

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
calbach: Where is this check? Also I do think calling `getWorkspace()` is sufficient - we should avoid making two serial calls to FC if we can avoid it.

calbach: Nevermind, I must have been looking at a stale commit, I see it now. Second statement still stands though

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -343,86 +341,92 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
-    } else {
-      notebookFile.getJSONArray("cells").put(createNotebookCodeCellWithString(queriesAsStrings));
-    }
-
-    notebooksService.saveNotebook(
-        workspace.getWorkspace().getBucketName(),
-        dataSetExportRequest.getNotebookName(),
-        notebookFile);
-
+    //    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
```
jmthibault79: Reminder to revert this.  (Debugging detritus from our pairing session)

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -181,8 +183,9 @@ public GoogleCredential getDefaultServiceAccountCredentials() throws IOException
   }
 
   @Override
-  public JSONObject getFileAsJson(String bucketName, String fileName) throws IOException {
-    return new JSONObject(readToString(bucketName, fileName));
+  public JsonObject getFileAsJson(String bucketName, String fileName) throws IOException {
```
jmthibault79: This change is necessary because of the incompatibility between `org.json.JSONObject` and `com.google.gson.JsonObject`.  We should fix this: https://precisionmedicineinitiative.atlassian.net/browse/RW-2929

ui/src/app/views/interactive-notebook.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    render() {
+      return (
+        <div>
+          <div style={styles.navBar}>
+            <div style={{...styles.navBarItem, ...styles.active, width: 227}}>
+              Preview (Read-Only)
+            </div>
+            {this.state.userRequestedEditMode ? (
+              <div style={{...styles.navBarItem, width: 620}}>
```
calbach: 620 seems like a large value to hardcode, where'd this come from? In the mocks the buttons are `229`, the loading dialog is 509px. Might be better to let these flow more naturally though, e.g. `display: flex`

ui/src/app/views/interactive-notebook.tsx
```diff
@@ -0,0 +1,175 @@
...
+                Object.assign({}, styles.navBarItem,
+                  this.canWrite() ? styles.clickable : styles.disabled, {width: 135})}
+                   onClick={() => { this.onEditClick(); }}>
+                <EditComponentReact enableHoverEffect={false}
+                                    disabled={!this.canWrite()}
+                                    style={{...styles.navBarIcon}}/>
+                Edit
+              </div>
+            </div>)
+            }
```
calbach: Note: we don't have the X button. I don't think it's needed for MVP but we may want to track this.

ui/package.json
```diff
@@ -11,13 +11,14 @@
...
     "merge-swagger": "../api/build.rb merge-yaml --no-docker",
     "codegen-clean": "rm -rf src/generated src/notebooks-generated",
     "codegen-swagger": "java -jar ../aou-utils/swagger-codegen-cli.jar generate --lang typescript-angular --input-spec ../api/src/main/resources/merged.yaml --output src/generated --additional-properties ngVersion=2",
     "codegen-swagger-fetch": "java -jar ../aou-utils/swagger-codegen-cli.jar generate --lang typescript-fetch --input-spec ../api/src/main/resources/merged.yaml --output src/generated/fetch",
     "post-codegen": "sed -i.bak -e 's/OpaqueToken/InjectionToken/' src/generated/variables.ts",
     "notebooks-codegen-swagger": "java -jar ../aou-utils/swagger-codegen-cli.jar generate --lang typescript-angular --input-spec ../api/src/main/resources/notebooks.yaml --output src/notebooks-generated --additional-properties ngVersion=2",
+    "notebooks-codegen-swagger-fetch": "java -jar ../aou-utils/swagger-codegen-cli.jar generate --lang typescript-fetch --input-spec ../api/src/main/resources/notebooks.yaml --output src/notebooks-generated/fetch",
```
jmthibault79: What does this do?

ui/src/app/icons/edit.spec.tsx
```diff
@@ -18,6 +18,7 @@ describe('EditIconComponent', () => {
   beforeEach(() => {
     props = {
       disabled: false,
+      enableHoverEffect: true,
```
jmthibault79: What does this do?

ui/src/app/views/resource-card.tsx
```diff
@@ -484,7 +484,7 @@ export class ResourceCard extends React.Component<Props, State> {
           queryParams.set('jupyterLabMode', 'true');
         }
 
-        return `${workspacePrefix}/notebooks/${encodeURIComponent(notebook.name)}?` +
+        return `${workspacePrefix}/notebooks/preview/${encodeURIComponent(notebook.name)}?` +
```
jmthibault79: Is it still appropriate to include the queryParams here?  Preview/Readonly ignores them.

Actually - do we have any way to access Playground Mode or JupyterLab now?  Maybe we should remove all references to these.

ui/src/app/views/interactive-notebook.tsx
ericsong: tried using flex and the styling is SO MUCH better now. Thanks

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -154,13 +155,17 @@ public FileDetail renameNotebook(
...
       return cloudStorageService.getFileAsJson(
-          bucketName, "notebooks/".concat(notebookName.concat(".ipynb")));
+          bucketName, "notebooks/".concat(withNotebookExtension(notebookName)));
     } catch (IOException e) {
       throw new ServerErrorException(
           "Failed to get notebook " + notebookName + " from bucket " + bucketName);
     }
   }
 
+  private String withNotebookExtension(String notebookName) {
```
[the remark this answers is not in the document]
ericsong: good catch! I even wrote both haha - woops. 

I decided to keep my current version with `endsWith` since thats easier to read than regex.

ui/package.json
ericsong: We use swagger to create our API clients but we actually need to generate TWO versions of each since angular and react use different API client libraries.

We were already generating both versions for `codegen` in lines 17-19 but we were only generating the angular version for the notebooks client since none of the react components hit the notebooks API until this PR.

This line adds the step to generate the React version of the notebooks API.

ui/src/app/icons/edit.spec.tsx
ericsong: I added `enableHoverEffect` as a prop to the edit component so I can toggle if I want that behavior or not.

The test needed to be updated with the new prop parameters.

ui/src/app/views/resource-card.tsx
ericsong: good catch. The new component doesnt do anything with it so we should just drop it.

The old notebooks page still uses them though so references to that page should append the query parameters.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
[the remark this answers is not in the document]
jmthibault79: agreed

PR #2276, opened 2019-06-28
[RW-2928][Risk=no] Srubenst/delete demos

blrubenstein:


blrubenstein:
This also fixes a small bug I noticed around concept routing when adding a new concept set.

blrubenstein:
@calbach I will let Rohini and the QA folks know as well.

PR #2280, opened 2019-06-28
[RW-2848][Risk=no] Srubenst/select all wording

blrubenstein:

### DIFF
--- a/ui/src/app/views/dataset-page.tsx
+++ b/ui/src/app/views/dataset-page.tsx
@@ -4,10 +4,10 @@ import * as React from 'react';
 
 import {Button, Clickable} from 'app/components/buttons';
 import {FadeBox} from 'app/components/containers';
-import {TooltipTrigger} from 'app/components/popups';
-import {Spinner} from 'app/components/spinners';
 import {ClrIcon} from 'app/components/icons';
 import {CheckBox} from 'app/components/inputs';
+import {TooltipTrigger} from 'app/components/popups';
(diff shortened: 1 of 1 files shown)


ui/src/app/views/dataset-page.tsx
```diff
@@ -332,21 +330,23 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
       this.setState({selectedValues: valuesSelected, dataSetTouched: true});
     }
 
+    get selectAll() {
```
NehaBroad: if this is not being used anywhere else except selectAllValues we can just directly use it in the if statement

ui/src/app/views/dataset-page.tsx
```diff
@@ -522,11 +522,17 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
                     <div>
                       Values
                     </div>
-                    <Clickable data-test-id='select-all'
-                               style={{marginLeft: 'auto'}}
-                               onClick={() => this.selectAllValues()}>
-                      Select All
-                    </Clickable>
+                    <div style={{marginLeft: 'auto',
+                      width: '5rem', display: 'flex', alignItems: 'center'}}>
```
NehaBroad: can you extract that to styles constant

ui/src/app/views/dataset-page.tsx
NehaBroad: Ohh its being used on select so never mind

PR #2285, opened 2019-07-01
[No Ticket][Risk=no] update-intellij-setup

andy7i:
Updated the link to be generalized to UI and API and moved it to the setup section.


ericsong:
I believe the first google doc link "workbench IntelliJ setup" has the steps for the autoformatting so we don't need to link out to the "Workbench: google-java-format" document.

PR #2287, opened 2019-07-01
[No Ticket][Risk=no] Srubenst/style cleanups

blrubenstein:
These are the new colors as requested by Lou. Please make sure it looks reasonable and that nothing is bad. (We are hoping to just keep the core set of colors in the color file, and apply lightness as we need hues of the same base color.)

### DIFF
--- a/ui/src/app/styles/colors.ts
+++ b/ui/src/app/styles/colors.ts
@@ -1,47 +1,28 @@
+import * as Color from 'color';
+
 export default {
-  purple: [
-    '#262262',
-    '#4356a7',
-    '#AB87B3'
-  ],
-  backgroundGrey: '#f1f2f2',
-  red: '#d0021b',
-  green: [
-    '#8bc990',
-    '#7AC79B',
-    '#7CC79B',
-    '#659F3D'
-  ],
-  blue: [
-    '#2691d0',
-    '#5aa6da',
-    '#6CACE4',
-    '#85bde4',
-    '#afd3ed',
-    '#d7eaf6',
-    '#eaf4fb',
-    '#2F2E7E',
-    '#302C71',
-    '#216FB4' // link blue
-  ],
-  gray: [
-    '#4a4a4a',
-    '#6e6e6e',
-    '#929292',
-    '#b7b7b7',
-    '#c3c3c3',
-    '#dbdbdb',
-    '#ededed',
-    '#F1F1F1'
-  ],
+  // Used for buttons and text
+  primary: '#262262',
+  // Used for tab navigation
+  secondary: '#6CACE4',
+  success: '#8BC990',
+  // Used for tables
+  light: '#E9ECEF',
+  highlight: '#F8C954',
+  warning: '#F7981C',
+  danger: '#DB3214',
+  dark: '#4A4A4A',
+  // Text links and Icon links
+  accent: '#216FB4',
+  resourceCardHighlights: {
+    conceptSet: '#A27BD7',
+    cohort: '#F8C954',
+    notebook: '#8BC990',
+    dataSet: '#6CACE4'
+  },
   white: '#fff',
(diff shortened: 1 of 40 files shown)


ui/src/app/styles/colors.ts
```diff
@@ -1,47 +1,28 @@
...
-  ],
-  yellow: [
-    '#F8C954',
-    '#F7981C',
-    '#F68D76'
-  ]
+};
+
+export const colorWithLightness = (color: string, intensity: number) => {
+  return Color(color).mix(Color('white'), 1 - intensity).toString();
```
gjuggler: The mixing is an interesting approach, but would color.lighten (or color.whiten) be simpler and do something similar if not identical?

ui/src/app/styles/colors.ts
```diff
@@ -1,47 +1,28 @@
...
-    '#333333'
-  ],
-  yellow: [
-    '#F8C954',
-    '#F7981C',
-    '#F68D76'
-  ]
+};
+
+export const colorWithLightness = (color: string, intensity: number) => {
```
gjuggler: What's the expected range for "intensity"? I would have expected 0 to 1 based on the implementation, but I see this being called with a value of 1.6. That doesn't seem quite right.

Also... based on my interpretation of Color.mix tests https://github.com/Qix-/color/blob/master/test/index.js#L628, it seems like the intensity value is flipped from what it should be.  In other words,

Color('black').mix(Color('white'), 0) --> black
Color('black').mix(Color('white'), 1) --> white

which would make me think our 'intensity' should be passed directly to Color.mix, rather than subtracted from 1. Unless I'm missing something?

(But see below – it may be simpler & more reliable to use 'lighten' instead anyway)

ui/src/app/styles/colors.ts
```diff
@@ -1,47 +1,28 @@
...
-    '#929292',
-    '#b7b7b7',
-    '#c3c3c3',
-    '#dbdbdb',
-    '#ededed',
-    '#F1F1F1'
-  ],
+  // Used for buttons and text
+  primary: '#262262',
+  // Used for tab navigation
```
gjuggler: The comments are useful – maybe there should be one on each color, succinctly describing its intended use (and current overall color)?

ui/src/app/styles/colors.ts
blrubenstein: color.lighten actually affected the rgb values, because I think I would need to mess with hue/saturation as well to get it to work. This is the approach Terra uses, and also lets us use higher numbers to darken, so I wanted to go with their approach.

ui/src/app/styles/colors.ts
blrubenstein: Talking to Lou, we decided the other ones were relatively clear semantically.

ui/src/app/styles/colors.ts
```diff
@@ -1,47 +1,28 @@
...
-    '#4a4a4a',
-    '#6e6e6e',
-    '#929292',
-    '#b7b7b7',
-    '#c3c3c3',
-    '#dbdbdb',
-    '#ededed',
-    '#F1F1F1'
-  ],
+  // Used for buttons and text
```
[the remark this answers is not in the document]
blrubenstein: This is the ordering it is in our new style guide: https://projects.invisionapp.com/d/main#/console/17894189/371138106/inspect which is part of why I left it this way. I can move to alphabetical if we want to though.

ui/src/app/styles/colors.ts
blrubenstein: Talking to Isaac later. I am tempted to just make it intensity though, if what Brett said is correct.
(`.mix` gives us a range of -1 to 1, where -1 is black, 1 is white, and 0 is the base color)

PR #2301, opened 2019-07-03
[RW-2847][risk=no]Open Notebook from Data set in a "new tab" by default

NehaBroad:
Export data set via snowman : opens the notebook in a new tab and returns back to Data tab

Update/save data set: opens notebook in a new tab and then closes the save modal and return back to data tab

### DIFF
--- a/ui/src/app/views/new-dataset-modal.tsx
+++ b/ui/src/app/views/new-dataset-modal.tsx
@@ -149,9 +149,7 @@ class NewDataSetModal extends React.Component<Props, State> {
       } else {
         await dataSetApi().createDataSet(workspaceNamespace, workspaceId, request);
       }
-      if (!this.state.exportToNotebook) {
-        window.history.back();
-      } else {
+      if (this.state.exportToNotebook) {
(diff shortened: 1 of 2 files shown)


ui/src/app/views/new-dataset-modal.tsx
```diff
@@ -161,9 +160,11 @@ class NewDataSetModal extends React.Component<Props, State> {
...
             newNotebook: this.state.newNotebook
           });
-        navigate(['workspaces',
-          workspaceNamespace,
-          workspaceId, 'notebooks', this.state.notebookName + '.ipynb']);
+        // Open notebook in a new tab and return back to the Data tab
+        const notebookUrl = '\\workspaces\\' + workspaceNamespace + '\\' + workspaceId +
+          '\\notebooks\\' + this.state.notebookName + '.ipynb';
+        window.history.back();
+        window.open(notebookUrl);
```
blrubenstein: Nit: Ordering-wise I would probably open the notebook URL before going back in the history.

PR #2304, opened 2019-07-03
[RW-2977][risk=no] Update drugs UI for CB list search

dolbeew:
Call api for ingredients of brands returned by list search

### DIFF
--- a/ui/src/app/cohort-search/list-search/list-search.component.tsx
+++ b/ui/src/app/cohort-search/list-search/list-search.component.tsx
@@ -301,11 +301,11 @@ export const ListSearch = withCurrentWorkspace()(
               <span style={{fontWeight: 'bold'}}>{row.code}</span> {row.name}
             </div>
           </td>
-          <td style={columnStyle}>{row.type}</td>
-          <td style={columnStyle}>
+          <td style={styles.columnBody}>{row.type}</td>
+          <td style={styles.columnBody}>
             {row.count > -1 && row.count.toLocaleString()}
           </td>
-          <td style={{...columnStyle, padding: '0.2rem'}}>
+          <td style={{...styles.columnBody, padding: '0.2rem'}}>
             {row.hasHierarchy &&
             <i className='pi pi-sitemap'
                style={styles.treeIcon}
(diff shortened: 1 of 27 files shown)


ui/src/app/cohort-search/list-search/list-search.component.tsx
```diff
@@ -51,6 +51,11 @@ const styles = reactStyles({
     opacity: 0.4,
     cursor: 'not-allowed'
   },
+  brandIcon: {
+    marginRight: '0.4rem',
+    color: '#9a9a9a',
```
blrubenstein: Please use the colors file if possible. Lou should be able to help with what the right color is. (And we're avoiding adding new colors to that file, if possible it should be a different intensity version of one of the colors in our palette)

ui/src/app/styles/colors.ts
```diff
@@ -24,6 +24,8 @@ export default {
     notebook: '#8BC990'
   },
   white: '#fff',
+  purple: 'purple',
```
blrubenstein: We are avoiding adding new colors to this file to avoid file explosion. Purple or Blue should probably exist as one of our existing colors in our palette. (Maybe primary and accent is what you are looking for?)

If you're interested, I'm going to be giving rationale at the Eng Sync today.

ui/src/app/cohort-search/list-search/list-search.component.tsx
dolbeew: Okay. I'm going to do a follow-up PR to try to refactor all CB colors to use the new file.

ui/src/app/styles/colors.ts
dolbeew: This was not an intentional change. I couldn't deploy the UI locally, so just added these to prevent compile errors till the fix came through. Removed.

ui/src/app/cohort-search/list-search/list-search.component.tsx
blrubenstein: Sounds good. I've been going through as well, but if you can handle the CB, that would be a great help

PR #2306, opened 2019-07-03
[risk=low][RW-2709]: Add UI flow for Playground Mode in new notebooks page

ericsong:
This adds a new pattern of "Do not show again". This is implemented by tracking this value on the users cookies.

Screen Shot 2019-07-03 at 3 44 35 PM
Screen Shot 2019-07-03 at 3 48 46 PM


ui/src/app/views/confirm-playground-mode-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export class ConfirmPlaygroundModeModal extends React.Component<Props, State> {
+
+  public static DO_NOT_SHOW_AGAIN = 'DO_NOT_SHOW_AGAIN';
+
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      checked:  Cookies.get(ConfirmPlaygroundModeModal.DO_NOT_SHOW_AGAIN) === String(true)
```
gjuggler: Couple of quick drive-by points here:

1) Shouldn't the cookie string (currently DO_NOT_SHOW_AGAIN) have some prefix that clarifies that this is unique to the confirm playground modal component?

2) We do have some sort of server-side storage that might be shifted to support this case as well. We already use it to prevent showing the quick tour modal to a user after they've seen it. See https://github.com/all-of-us/workbench/blob/master/api/src/main/resources/workbench.yaml#L313

I was running into this API call while working on the homepage recently, and I thought with the right client-side mapping it could be a nice way of tracking certain small bits of user-centric state like this. Just food for though.

ericsong: 1. Good point, I'll change that to something more unique.

2. Cool! I didn't know about that. I think I'm inclined to stick with the cookies solution for now since its simpler and Karthik since its fine from a UX perspective. But I'll let him know there's a reasonable path forward to implement this with back end persistence.

ui/src/app/views/confirm-playground-mode-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+        </p>
+        <p style={{color: colors.primary}}>
+          To save your work, choose <b>Make a Copy</b> from the <b>File Menu</b>
+          to make your own version.
+        </p>
+      </ModalBody>
+      <ModalFooter style={{alignItems: 'center'}}>
+        <CheckBox checked={this.state.checked}
+                  onChange={() => { this.toggleChecked(); }} />
+        <label style={{marginLeft: '8px', marginRight: 'auto', color: colors.primary}}>
```
gjuggler: It's a pet peeve of mine when labels can't be used to toggle a checkbox. Maybe it's possible to do something like

<CheckBox id="foo" ... />
<label for="foo" ... >

ui/src/app/views/confirm-playground-mode-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+  render() {
+    return <Modal onRequestClose={() => { this.props.onCancel(); }}>
+      <ModalTitle>Playground Mode</ModalTitle>
+      <ModalBody>
+        <p style={{color: colors.primary}}>
+          Playground mode allows you to explore, change and run the code,
+          but your edits will not be saved.
+        </p>
+        <p style={{color: colors.primary}}>
+          To save your work, choose <b>Make a Copy</b> from the <b>File Menu</b>
```
gjuggler: Nitpick: based on the PR screenshot, you may need an &nbsp; after the </b> to add a space after File Menu and before the next word.

ericsong: good catch

ui/src/app/views/confirm-playground-mode-modal.tsx
ericsong: done!

PR #2309, opened 2019-07-05
[RW-2927][Risk=no] Hide DUA if not enabled in config

blrubenstein:

### DIFF
--- a/ui/src/app/views/registration-dashboard.tsx
+++ b/ui/src/app/views/registration-dashboard.tsx
@@ -78,7 +78,7 @@ async function redirectToTraining() {
 // This needs to be a function, because we want it to evaluate at call time,
 // not at compile time, to ensure that we make use of the server config store.
 // This is important so that we can feature flag off registration tasks.
-export const RegistrationTasks = () => serverConfigStore.getValue() ? [
+export const getRegistrationTasks = () => serverConfigStore.getValue() ? [
   {
     key: 'twoFactorAuth',
     title: 'Turn on Google 2-Step Verification',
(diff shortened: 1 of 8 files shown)


ui/src/app/views/registration-dashboard.tsx
```diff
@@ -110,7 +113,7 @@ export const RegistrationTasks = [
       return !!profile.eraCommonsCompletionTime || !!profile.eraCommonsBypassTime;
     },
     onClick: redirectToNiH
-  }, {
+  }, ...(serverConfigStore.getValue().enableDataUseAgreement ? [{
```
ericsong: I don't think its that necessary but I thought of a cute trick that looks cleaner to me. We can add another property to each task called `isEnabled` and we can filter on that value to return the correct list of tasks.

The default can be `() => true` but for values backed by serverConfig, we can do the API call.

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -116,18 +119,23 @@ export const RegistrationTasks = [
...
-];
+] : (() => {
+  throw new Error('Cannot load registration tasks before config loaded');
+})();
 
-export const RegistrationTasksMap = RegistrationTasks.reduce((acc, curr) => {
-  acc[curr.key] = curr;
-  return acc;
-}, {});
+export const getRegistrationTasksMap = () => getRegistrationTasks()
```
ericsong: I like the idea of using the feature flag to avoid adding boilerplate values to the rest of the tasks.

One note about the api change, `getRegistrationTasks` and `getRegistrationTasksMap` now return different values which I think is a little confusing. It should return the same values but in different data structures (list vs map).

In order to do that, I think we will need to define the list outside of the `getRegistrationTasks` or `getRegistrationTasksMap` methods. Another function will work but if we want to keep it as a compiled array, we can set `featureFlag` to the feature flag key or even a method that checks the store. We can still keep the logic of ignoring undefined values to avoid adding it to other tasks.

PR #2310, opened 2019-07-05
[RW-2798][Risk=no] Make data tab default

blrubenstein:


calbach:
Should the workspace name breadcrumb be updated as well to match this?

calbach:
Sorry, I meant to expand on that comment: clicking the workspace name within the breadcrumb still takes you to the "about" page. Should it go to the data tab now?

PR #2311, opened 2019-07-05
Adds disease ontology search to workspace edit page

bopwerks:

### DIFF
--- a/ui/src/app/views/workspace-edit.tsx
+++ b/ui/src/app/views/workspace-edit.tsx
@@ -11,9 +11,8 @@ import {TwoColPaddedTable} from 'app/components/tables';
 import {cdrVersionsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
 import {reactStyles, ReactWrapperBase, sliceByHalfLength, withCurrentWorkspace, withRouteConfigData} from 'app/utils';
-import {currentWorkspaceStore, navigate, userProfileStore} from 'app/utils/navigation';
+import {currentWorkspaceStore, navigate, serverConfigStore, userProfileStore} from 'app/utils/navigation';
 import {CdrVersion, DataAccessLevel, SpecificPopulationEnum, Workspace} from 'generated/fetch';
-import {serverConfigStore} from 'app/utils/navigation';
 import * as fp from 'lodash/fp';
 import * as React from 'react';
 
--- a/ui/src/app/views/workspace-edit.tsx
+++ b/ui/src/app/views/workspace-edit.tsx
@@ -388,17 +388,19 @@ export interface WorkspaceEditState {
   diseaseEntryState: number;
 }
 
+// States for disease search interaction
+enum DiseaseSearchState {
+    DS_START,   // Input box lacks focus, drop-down invisible
+    DS_ACTIVE,  // Input box focused, drop-down invisible
+    DS_SUGGEST, // Input box focused, search results visible
+    DS_HOVER,   // Input box focused, search results visible, item activated
+}
+
 export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace())(
   class WorkspaceEditCmp extends React.Component<WorkspaceEditProps, WorkspaceEditState> {
     // Declare this ahead of time so the typescript compiler won't choke on the following
     // this.searchTermChangedEvent = ...
     searchTermChangedEvent;
-
-    // States for disease search interaction
-    DS_START;   // Input box lacks focus, drop-down invisible
-    DS_ACTIVE;  // Input box focused, drop-down invisible
-    DS_SUGGEST; // Input box focused, search results visible
-    DS_HOVER;   // Input box focused, search results visible, item activated
 
     constructor(props: WorkspaceEditProps) {
       super(props);
@@ -451,7 +453,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
 
     get showSearchResults(): boolean {
       const state = this.state.diseaseEntryState;
-      return state === this.DS_SUGGEST || state === this.DS_HOVER;
+      return state === DiseaseSearchState.DS_SUGGEST || state === DiseaseSearchState.DS_HOVER;
     }
 
     diseaseSearch(value: string): void {
@@ -469,7 +471,8 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
         const diseases = fp.map((elt) => elt['label'])(labeledMatches);
         this.setState({
           diseaseList: diseases,
-          diseaseEntryState: diseases.length > 0 ? this.DS_SUGGEST : this.DS_ACTIVE,
+          diseaseEntryState: diseases.length > 0 ?
+            DiseaseSearchState.DS_SUGGEST : DiseaseSearchState.DS_ACTIVE,
           diseaseHover: diseases.map(() => false),
         });
       });
@@ -738,13 +741,14 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                           disabled={this.state.workspace.researchPurpose.diseaseFocusedResearch}>
                         <TextInput value={this.state.workspace.researchPurpose.diseaseOfFocus}
                           style={{width: '90%', border: '1px solid #9a9a9', borderRadius: '5px'}}
-                          onFocus={() => this.setState({diseaseEntryState: this.DS_ACTIVE})}
+                          onFocus={() =>
+                            this.setState({diseaseEntryState: DiseaseSearchState.DS_ACTIVE})}
                           onBlur={() => {
                             const state = this.state.diseaseEntryState;
-                            if (state === this.DS_ACTIVE  ||
-                                state === this.DS_SUGGEST ||
-                                state === this.DS_HOVER) {
-                              this.setState({diseaseEntryState: this.DS_START});
+                            if (state === DiseaseSearchState.DS_ACTIVE  ||
+                                state === DiseaseSearchState.DS_SUGGEST ||
+                                state === DiseaseSearchState.DS_HOVER) {
+                              this.setState({diseaseEntryState: DiseaseSearchState.DS_START});
                             }
                           }}
                           onChange={e => {
@@ -769,13 +773,13 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                               styles.boxHover : styles.box}
                                 onMouseOver={() =>
                                   this.setState({
-                                    diseaseEntryState: this.DS_HOVER,
+                                    diseaseEntryState: DiseaseSearchState.DS_HOVER,
                                     diseaseHover: this.toggleDiseaseHover(j)
                                   })
                                 }
                                 onMouseOut={() =>
                                   this.setState({
-                                    diseaseEntryState: this.DS_SUGGEST,
+                                    diseaseEntryState: DiseaseSearchState.DS_SUGGEST,
                                     diseaseHover: this.toggleDiseaseHover(j)
                                   })
                                 }
--- a/ui/src/app/views/workspace-edit.tsx
+++ b/ui/src/app/views/workspace-edit.tsx
@@ -13,6 +13,7 @@ import colors from 'app/styles/colors';
 import {reactStyles, ReactWrapperBase, sliceByHalfLength, withCurrentWorkspace, withRouteConfigData} from 'app/utils';
 import {currentWorkspaceStore, navigate, userProfileStore} from 'app/utils/navigation';
 import {CdrVersion, DataAccessLevel, SpecificPopulationEnum, Workspace} from 'generated/fetch';
+import {serverConfigStore} from 'app/utils/navigation';
 import * as fp from 'lodash/fp';
 import * as React from 'react';
 
@@ -460,8 +461,8 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
       if (!searchTerm) {
         return;
       }
-      const url = 'https://firecloud-orchestration.dsde-dev.broadinstitute.org' +
-        '/duos/autocomplete/' + searchTerm;
+      const baseurl = serverConfigStore.getValue().firecloudURL;
+      const url = baseurl + '/duos/autocomplete/' + searchTerm;
       fetch(encodeURI(url)).then((response) => {
         return response.json();
       }).then((matches) => {
--- a/ui/src/app/views/workspace-edit.tsx
+++ b/ui/src/app/views/workspace-edit.tsx
@@ -10,7 +10,7 @@ import {SpinnerOverlay} from 'app/components/spinners';
 import {TwoColPaddedTable} from 'app/components/tables';
 import {cdrVersionsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
-import {isBlank, reactStyles, ReactWrapperBase, sliceByHalfLength, withCurrentWorkspace, withRouteConfigData} from 'app/utils';
+import {reactStyles, ReactWrapperBase, sliceByHalfLength, withCurrentWorkspace, withRouteConfigData} from 'app/utils';
 import {currentWorkspaceStore, navigate, userProfileStore} from 'app/utils/navigation';
 import {CdrVersion, DataAccessLevel, SpecificPopulationEnum, Workspace} from 'generated/fetch';
 import * as fp from 'lodash/fp';
@@ -291,7 +291,7 @@ const styles = reactStyles({
     paddingLeft: '0.2rem',
   },
   boxHoverElement: {
-    color: '#262262',
+    color: colors.primary,
     margin: 0,
     padding: 0,
     fontFamily: 'Montserrat',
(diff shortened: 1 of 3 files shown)


ui/src/app/views/workspace-edit.tsx
```diff
@@ -268,7 +268,45 @@ const styles = reactStyles({
...
+    color: '#262262',
+  },
+  boxHover: {
+    background: 'rgb(234, 243, 250)',
+    color: '#262262',
+    paddingTop: '0.2rem',
+    paddingLeft: '0.2rem',
+  },
+  boxHoverElement: {
+    color: '#262262',
```
NehaBroad: instead of colors: #262262 we should uses colors.primary

ui/src/app/views/workspace-edit.tsx
```diff
@@ -390,8 +434,45 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+    get showSearchResults(): boolean {
+      return !this.state.autocompleteLoading &&
+        this.state.autocompleteDiseases.length > 0 &&
+        !isBlank(this.state.searchTerm) && this.state.dropDown;
+    }
+
+    diseaseSearch(value: string): void {
+      this.setState({autocompleteLoading: true, autocompleteDiseases: [], searchTerm: value});
+      if (!value.trim()) {
+        this.setState({autocompleteLoading: false, dropDown: false});
```
NehaBroad: how about writing this as         this.setState({autocompleteLoading: value.trim.....});

ui/src/app/views/workspace-edit.tsx
```diff
@@ -390,8 +434,45 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+      const self = this;
+      const url = 'https://firecloud-orchestration.dsde-dev.broadinstitute.org' +
+        '/duos/autocomplete/' + value;
+      fetch(encodeURI(url)).then((response) => {
+        return response.json();
+      }).then((matches) => {
+        const labeledMatches = fp.filter((elt) => elt.hasOwnProperty('label'))(matches);
+        const labels = fp.map((elt) => elt['label'])(labeledMatches);
+        const diseaseHover = labels.map(() => false);
+        self.setState({
```
blrubenstein: Curious why you need to bind self above? Because you're just using lambdas, the `this` should be preserved in here.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -390,8 +434,45 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+    }
+
+    diseaseSearch(value: string): void {
+      this.setState({autocompleteLoading: true, autocompleteDiseases: [], searchTerm: value});
+      if (!value.trim()) {
+        this.setState({autocompleteLoading: false, dropDown: false});
+        return;
+      }
+      const self = this;
+      const url = 'https://firecloud-orchestration.dsde-dev.broadinstitute.org' +
```
ericsong: This will not work for all environments since its hard coded to the dev environment. We should be able to use the generated client for this which will take care of the request configuration.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -390,8 +434,45 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-        showStigmatizationDetails: false
+        showStigmatizationDetails: false,
+        autocompleteLoading: false,
+        autocompleteDiseases: [],
+        diseaseHover: [],
+        dropDown: false,
+        searchTerm: '',
+        hovering: false
       };
+      this.searchTermChangedEvent = fp.debounce(300, this.diseaseSearch);
```
NehaBroad: this. searchTermChangedEvent should be defined as a class variable searchTermChangedEvent: Function;

ui/src/app/views/workspace-edit.tsx
NehaBroad: can we define this as const

NehaBroad: Question: since the url ends with -dev.broadinstitute.org, will there be a different url for production?

ui/src/app/views/workspace-edit.tsx
```diff
@@ -505,6 +586,11 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
       return fp.includes(populationEnum, this.state.workspace.researchPurpose.populationDetails);
     }
 
+    toggleDiseaseHover(diseaseIndex) {
+      this.state.diseaseHover[diseaseIndex] = !this.state.diseaseHover[diseaseIndex];
```
blrubenstein: In general, we avoid directly setting state using `=`, and instead use the `setState` method, because it does the re-rendering under the hood. And both times you use this method, it is used within a setState call. I think you could even do something pretty like:
`this.setState({diseaseHover: fp.setWith(Array, '['+diseaseIndex+']', !this.state.diseaseHover[diseaseIndex], this.state.diseaseHover)` if you wanted to one-line it. (Not sure this is as readable? Also, not sure on syntax. But generally avoid direct `=` setting on state.)

ui/src/app/views/workspace-edit.tsx
bopwerks: yep! const variables are limited to the scope in which they're defined, afaik.

blrubenstein: I don't know if we have a generated firecloud client, but if you want to use fetch, we should at least pull down the FC base URL from the config. Otherwise could go through the firecloud generated client. https://github.com/all-of-us/workbench/blob/master/api/config/config_local.json#L4

blrubenstein: This PR: https://github.com/all-of-us/workbench/pull/2262/files is an example of how to pull down server config information to the UI.

ui/src/app/views/workspace-edit.tsx
bopwerks: good idea. i was able to remove a bunch of duplicate state.

ui/src/app/views/workspace-edit.tsx
bopwerks: It seems that `fp.setWith(Array, '['+diseaseIndex+']', !this.state.diseaseHover[diseaseIndex], this.state.diseaseHover)` and `toggleDiseaseHover` are functionally equivalent, except that with `fp.setWith` the array indexing code is interpreted rather than compiled. If there's no simpler way to express assignment then I don't care to change this. Re-rendering will happen either way since `toggleDiseaseHover()` is evaluated within the call to `setState`.

blrubenstein: I would rather use the setState in here rather than using a `state = {new state}` then. Because nothing with this function says that it needs to be called within a setState, and I'd rather have this in code. It seems fine not to use the setWith (I think its less readable, and this is fine), but definitively doing a setState here ensures it can't be misused.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -641,23 +715,51 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+                          {this.state.autocompleteDiseases.map((disease, j) => {
+                            return (
+                            <div key={j} style={this.state.diseaseHover[j] ?
+                              styles.boxHover : styles.box}
+                                onMouseOver={() =>
+                                  this.setState({diseaseHover: this.toggleDiseaseHover(j)})}
+                                onMouseOut={() =>
+                                  this.setState({diseaseHover: this.toggleDiseaseHover(j)})}
+                                onClick={() => {
+                                  this.setState(fp.set([
```
gjuggler: Parsing this line, it looks like this will evaluate to a call to setState like:

this.setState({ workspace: { researchPurpose: { diseaseOfFocus: 'my-disease' } } });

I'm interested in the semantics here of how React handles setting state for a deeply-nested object. The docs say "shallow merge" (https://reactjs.org/docs/react-component.html#setstate), which suggests to me that this will cause the entire workspace object to be overwritten. That doesn't seem to be what we want to happen, but it also seems like it doesn't break the overall flow of this component.

If we wanted to be on the safe side, you could use the setState updater syntax to modify the latest version of the state object:

this.setState((prevState, props) => {
  fp.set('workspace.researchPurpose.diseaseOfFocus', disease, prevState);
  return {
    workspace: prevState.workspace,
    autocompleteDiseases: [],
  };
});

ui/src/app/views/workspace-edit.tsx
```diff
@@ -348,6 +386,9 @@ export interface WorkspaceEditState {
   loading: boolean;
   showUnderservedPopulationDetails: boolean;
   showStigmatizationDetails: boolean;
+  autocompleteDiseases: Array<string>;
+  diseaseHover: Array<boolean>;
+  searchTermChangedEvent: Function;
```
gjuggler: This state variable is unused. In the constructor, the searchTermChangedEvent debounced function is assigned directly as an instance property (which is fine, IMO since it's not involved in the rendering of the component).

ui/src/app/views/workspace-edit.tsx
bopwerks: Not needed anymore, will use `this`

ui/src/app/views/workspace-edit.tsx
```diff
@@ -259,7 +260,45 @@ const styles = reactStyles({
...
+    maxHeight: '12rem',
+    minHeight: '30px',
+    visibility: 'visible',
+    overflowY: 'scroll',
+    width: '35%',
+    marginTop: '.25rem',
+    zIndex: 100,
+    border: '1px solid #979797',
+    borderRadius: '5px',
+    backgroundColor: '#FFFFFF',
```
jacmrob: nit: use scout's colors

ui/src/app/views/workspace-edit.tsx
```diff
@@ -259,7 +260,45 @@ const styles = reactStyles({
...
+  },
+  box: {
+    borderRadius: '5px',
+    paddingTop: '0.2rem',
+    paddingLeft: '0.2rem',
+    color: '#262262',
+  },
+  boxHover: {
+    background: 'rgb(234, 243, 250)',
+    color: '#262262',
```
jacmrob: nit: colors

ui/src/app/views/workspace-edit.tsx
bopwerks: The following experiment seems to indicate that the state is preserved.
```
    				onMouseDown={() => {
                                  this.setState((oldState, props) => {
                                    console.log("oldState:", oldState);
                                    const newState = fp.set([
                                      'workspace',
                                      'researchPurpose',
                                      'diseaseOfFocus'
                                    ], disease)(oldState);
                                    console.log("newState:", newState);
                                    return newState;
                                  });
```
Results:
```
oldState: 
{cdrVersionItems: Array(1), workspace: {…}, workspaceCreationConflictError: false, workspaceCreationError: false, workspaceCreationErrorMessage: "", …}
cdrVersionItems: [{…}]
cloneUserRole: false
diseaseEntryState: 8
diseaseHover: (20) [false, false, true, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false]
diseaseList: (20) ["lepromatous leprosy", "larynx leiomyosarcoma", "large bowel leiomyoma", "larynx leiomyoma", "leptospirosis", "laryngostenosis", "legionellosis", "lactocele", "labyrinthitis", "lagophthalmos", "leech infestation", "leptomeninges sarcoma", "Lafora disease", "leiomyomatosis", "Legionnaires' disease", "Leber congenital amaurosis", "cutaneous leishmaniasis", "LEOPARD syndrome", "Lambert-Eaton myasthenic syndrome", "leishmaniasis"]
loading: false
showStigmatizationDetails: false
showUnderservedPopulationDetails: false
workspace: {name: "", dataAccessLevel: "registered", namespace: undefined, cdrVersionId: "1", researchPurpose: {…}}
workspaceCreationConflictError: false
workspaceCreationError: false
workspaceCreationErrorMessage: ""
__proto__: Object
workspace-edit.tsx:791 newState: 
{cdrVersionItems: Array(1), workspace: {…}, workspaceCreationConflictError: false, workspaceCreationError: false, workspaceCreationErrorMessage: "", …}
cdrVersionItems: [{…}]
cloneUserRole: false
diseaseEntryState: 8
diseaseHover: (20) [false, false, true, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false]
diseaseList: (20) ["lepromatous leprosy", "larynx leiomyosarcoma", "large bowel leiomyoma", "larynx leiomyoma", "leptospirosis", "laryngostenosis", "legionellosis", "lactocele", "labyrinthitis", "lagophthalmos", "leech infestation", "leptomeninges sarcoma", "Lafora disease", "leiomyomatosis", "Legionnaires' disease", "Leber congenital amaurosis", "cutaneous leishmaniasis", "LEOPARD syndrome", "Lambert-Eaton myasthenic syndrome", "leishmaniasis"]
loading: false
showStigmatizationDetails: false
showUnderservedPopulationDetails: false
workspace: {name: "", dataAccessLevel: "registered", namespace: undefined, cdrVersionId: "1", researchPurpose: {…}}
workspaceCreationConflictError: false
workspaceCreationError: false
workspaceCreationErrorMessage: ""
__proto__: Object
```

ui/src/app/views/workspace-edit.tsx
```diff
@@ -381,8 +438,42 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
         loading: false,
         showUnderservedPopulationDetails: false,
-        showStigmatizationDetails: false
+        showStigmatizationDetails: false,
+        diseaseList: [],
+        diseaseHover: [],
+        diseaseEntryState: 0,
       };
+      this.searchTermChangedEvent = fp.debounce(300, this.diseaseSearch);
+      this.DS_START = 1;
```
jacmrob: if these are just enums i don't quite understand why they have to be defined with values?  could you just initialize them as 
`const diseaseSearchStates = {DS_START, DS_ACTIVE, DS_SUGGEST, DS_HOVER}`
?

ui/src/app/views/workspace-edit.tsx
```diff
@@ -381,8 +438,42 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+    }
+
+    diseaseSearch(value: string): void {
+      this.setState({diseaseList: []});
+      const searchTerm = value.trim();
+      if (!searchTerm) {
+        return;
+      }
+      const baseurl = serverConfigStore.getValue().firecloudURL;
+      const url = baseurl + '/duos/autocomplete/' + searchTerm;
```
jacmrob: nit - just call the above line in this line rather than defining it in a const? more concise since we don't use `baseUrl` again

ui/src/app/views/workspace-edit.tsx
bopwerks: Spoke with Eric and we agreed that using the config would be the faster choice for now, but created https://precisionmedicineinitiative.atlassian.net/browse/RW-3057 to use the generated client in the future.

ui/src/app/views/workspace-edit.tsx
gjuggler: +1, we have a pretty strong pattern to define constants like this as a Typescript enum, rather than as member variables, e.g. https://github.com/all-of-us/workbench/blob/1edcbd0eeeacf2eb3e3264be662702a53057ee39/ui/src/app/views/workspace-edit.tsx#L323

gjuggler:
Reiterating my comment about tests – I'm OK with this going in w/ manual testing for Bedford purposes, but we need to have a follow-up ticket filed & on track for completion in the next sprint.

ui/src/app/views/workspace-edit.tsx
bopwerks: Fixed

ui/src/app/views/workspace-edit.tsx
bopwerks: Fixed

ui/src/app/views/workspace-edit.tsx
bopwerks: Fixed

ui/src/app/views/workspace-edit.tsx
```diff
@@ -339,10 +378,28 @@ export interface WorkspaceEditState {
...
+
+  // The current state of the disease search state machine.
+  diseaseEntryState: number;
 }
 
 export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace())(
   class WorkspaceEditCmp extends React.Component<WorkspaceEditProps, WorkspaceEditState> {
+    // Declare this ahead of time so the typescript compiler won't choke on the following
+    // this.searchTermChangedEvent = ...
+    searchTermChangedEvent;
```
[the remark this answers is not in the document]
bopwerks: Fixed

ui/src/app/views/workspace-edit.tsx
```diff
@@ -381,8 +438,42 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+        return response.json();
+      }).then((matches) => {
+        const labeledMatches = fp.filter((elt) => elt.hasOwnProperty('label'))(matches);
+        const diseases = fp.map((elt) => elt['label'])(labeledMatches);
+        this.setState({
+          diseaseList: diseases,
+          diseaseEntryState: diseases.length > 0 ? this.DS_SUGGEST : this.DS_ACTIVE,
+          diseaseHover: diseases.map(() => false),
+        });
+      });
```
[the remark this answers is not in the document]
bopwerks: In the event of failure, no search results will be shown. I added a catch() under the assumption that things will blow up otherwise.

PR #2312, opened 2019-07-05
[RW-2688][Risk=no] Srubenst/attempt to fix deadline exceeded

blrubenstein:
See following log for the bug this attempts to fix: https://console.cloud.google.com/logs/viewer?project=all-of-us-workbench-test&minLogLevel=0&expandAll=false&timestamp=2019-07-05T17%3A50%3A11.587000000Z&customFacets&limitCustomFacetWidth=true&interval=NO_LIMIT&resource=gae_app%2Fmodule_id%2Fapi%2Fversion_id%2Fcircle-ci-test&logName=projects%2Fall-of-us-workbench-test%2Flogs%2Fappengine.googleapis.com%252Frequest_log&scrollTimestamp=2019-07-01T18%3A27%3A45.329697000Z&advancedFilter=resource.type%3D%22gae_app%22%0Aresource.labels.module_id%3D%22api%22%0Aresource.labels.zone%3D%22us17%22%0Aresource.labels.project_id%3D%22all-of-us-workbench-test%22%0Aresource.labels.version_id%3D%22circle-ci-test%22%0Atimestamp%3D%222019-06-27T17%3A59%3A01.398156000Z%22%0AinsertId%3D%225d1504270008c2438d9a7c78%22

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -71,6 +71,8 @@ public class DataSetController implements DataSetApiDelegate {
   private final WorkspaceService workspaceService;
 
   private static int NO_OF_PREVIEW_ROWS = 20;
+  private static int PREVIEW_RETRY_LIMIT = 2;
+  private static long APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC = 55000l;
   private static String CONCEPT_SET = "conceptSet";
   private static String COHORT = "cohort";
 
@@ -246,7 +248,6 @@ public class DataSetController implements DataSetApiDelegate {
       String workspaceNamespace, String workspaceId, DataSetRequest dataSet) {
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
-    int retryLimit = 2;
     DataSetPreviewResponse previewQueryResponse = new DataSetPreviewResponse();
     Map<String, QueryJobConfiguration> bigQueryJobConfig = dataSetService.generateQuery(dataSet);
     bigQueryJobConfig.forEach(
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -260,9 +260,15 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+              /* Google appengine has a 60 second timeout, we want to make sure this endpoint completes
+               * before that limit is exceeded, or we get a 500 error with the following type:
+               * com.google.apphosting.runtime.HardDeadlineExceededError
+               * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
+               */
               TableResult queryResponse =
                   bigQueryService.executeQuery(
-                      bigQueryService.filterBigQueryConfig(queryJobConfiguration), 60000l);
+                      bigQueryService.filterBigQueryConfig(queryJobConfiguration),
+                      50000l / retryLimit + 1);
```
NehaBroad: why is this dependent on retry limit and not hard coded as 60 sec?

blrubenstein: Because its the total time limit for the entire API endpoint. So if we're retrying, then the query could easily go over 60 seconds and then we throw a HardDeadlineExceededError

gjuggler: Can you convert retryLimit to a class-level constant with ALL_CAPS? Right now it's statically defined at the start of this function and never reassigned – would be easier to read if it's clearly a constant.

(By the same token – 50000l is a bit of a magic number, which could also be a constant APP_ENGINE_HARD_TIMEOUT_MSEC or something

blrubenstein: Will do. This bug was occuring relatively rarely, but happened more often with large value sets, and with all participants. It is hard to reliably repro locally, because we aren't running an actual appengine server, so we don't get this error.

PR #2313, opened 2019-07-05
[RW-2688][Risk=low] handle no concepts in set per domain

blrubenstein:


blrubenstein:
Talked offline, it isn't possible to recreate this situation. Only old concept sets with no concepts will throw this exception. We can't create concept sets with no concepts any more.

PR #2314, opened 2019-07-05
[RW-2983] [risk=no] Load user roles before opening the workspace share modal.

gjuggler:
See the bug description with some add'l info on the breakage.

With this PR, users will see a brief spinner on the workspace card itself while we're loading the user roles:

Screen Shot 2019-07-05 at 3 41 26 PM

I tried to add a minimal unit test that checks that we're correctly loading data and populating the workspace-share props. And some add'l docs on the API to clarify that you need to include the full desired ACL, not just newly-shared users.

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -1050,7 +1050,12 @@ paths:
         - workspaces
       consumes:
         - application/json
-      description: Shares a workspace with users
+      description: >
+        Shares a workspace with the given users and permissions. Note that the entire ACL must
+        be provided to this method (including all existing users and roles), not just the additional
+        roles desired. Clients should first fetch existing ACLs via the getFirecloudWorkspaceUserRoles
+        method, make any changes desired, then pass the final list to this method.
+
       operationId: shareWorkspace
       parameters:
           - in: body
--- a/ui/src/app/components/buttons.tsx
+++ b/ui/src/app/components/buttons.tsx
@@ -117,13 +117,17 @@ const computeStyle = ({style = {}, hover = {}, disabledStyle = {}}, {disabled}) 
 
 // Set data test id = '' in the child to prevent it from propegating to all children
 export const Clickable = ({as = 'div', disabled = false, onClick = null, ...props}) => {
+  // Don't propagate test IDs to the rendered child component.
+  delete props['data-test-id'];
   return <Interactive
-    as={as} {...props} data-test-id=''
+    as={as} {...props}
     onClick={(...args) => onClick && !disabled && onClick(...args)}
   />;
 };
 
 export const Button = ({type = 'primary', style = {}, disabled = false, ...props}) => {
+  // Don't propagate test IDs to the rendered child component.
+  delete props['data-test-id'];
   return <Clickable
     disabled={disabled} {...props}
     {...fp.merge(computeStyle(buttonVariants[type], {disabled}), {style})}
--- a/ui/src/app/components/popups.tsx
+++ b/ui/src/app/components/popups.tsx
@@ -26,11 +26,15 @@ const styles = {
   }
 };
 
+interface WithDynamicPositionProps {
+  target: string;
+}
+
 export const withDynamicPosition = () => WrappedComponent => {
-  const Wrapper = class extends React.Component {
+  const Wrapper = class WithDynamicPosition extends React.Component {
     static displayName = `withDynamicPosition()`;
 
-    props: any;
+    props: WithDynamicPositionProps;
     state: any;
     element: any;
     animation: number;
@@ -202,6 +206,8 @@ export const Popup = fp.flow(
   onClickOutside,
   withDynamicPosition()
 )(class PopupComponent extends React.Component {
+  static displayName = `Popup`;
+
   static readonly defaultProps = {
     side: 'right'
   };
--- a/ui/src/app/views/workspace-list.spec.tsx
+++ b/ui/src/app/views/workspace-list.spec.tsx
@@ -2,7 +2,6 @@ import {mount} from 'enzyme';
 import * as React from 'react';
 
 import {
-  dataSetApi,
   registerApiClient, workspacesApi,
 } from 'app/services/swagger-fetch-clients';
 import {navigate, serverConfigStore, userProfileStore} from 'app/utils/navigation';
@@ -40,7 +39,7 @@ describe('WorkspaceList', () => {
     registerApiClient(WorkspacesApi, new WorkspacesApiStub());
 
     // mocking because we don't have access to the angular service
-    reload.mockImplementation(async() => {
+    reload.mockImplementation(async () => {
       const newProfile = await profileApi.getMe();
       userProfileStore.next({profile: newProfile as unknown as Profile, reload, updateCache});
     });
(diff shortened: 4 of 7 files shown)


ui/src/app/components/buttons.tsx
```diff
@@ -117,13 +117,17 @@ const computeStyle = ({style = {}, hover = {}, disabledStyle = {}}, {disabled})
 
 // Set data test id = '' in the child to prevent it from propegating to all children
 export const Clickable = ({as = 'div', disabled = false, onClick = null, ...props}) => {
+  // Don't propagate test IDs to the rendered child component.
```
gjuggler: This change seemed to help with testing – we simply don't propagate the data-test-id prop to any child components, because it will be findable at this component.

ui/src/app/components/popups.tsx
```diff
@@ -26,11 +26,15 @@ const styles = {
   }
 };
 
+interface WithDynamicPositionProps {
```
gjuggler: These changes helped me with debugging the popup trigger issue... having non-Any props makes it much easier to follow what's going on, and the named class helps improve stack traces.

ui/src/app/views/workspace-list.spec.tsx
```diff
@@ -28,47 +32,67 @@ describe('WorkspaceList', () => {
   const updateCache = jest.fn();
 
   const component = () => {
-    return mount(<WorkspaceList/>);
+    return mount(<WorkspaceList/>, {attachTo: document.getElementById('root')});
```
gjuggler: This is what seemed to fix the popup trigger issue for me. I'm still not... quite sure why it works, but it did. I commented on RW-2625, this needs more investigation.

ui/src/app/views/workspace-list.tsx
```diff
@@ -83,66 +84,108 @@ const WorkspaceCardMenu: React.FunctionComponent<{
...
   }
 
   async deleteWorkspace() {
     const {wp} = this.props;
-    workspacesApi().deleteWorkspace(wp.workspace.namespace, wp.workspace.id).then(() => {
-      this.setState({confirmDeleting: false});
-      this.props.reload();
-    }).catch(() => {
-      this.setState({workspaceDeletionError: true});
+    this.setState({
```
gjuggler: This was kind of a drive-by fix... deleting a workspace is now a bit cleaner, and we use the same spinner on the workspace card while the deletion API call is pending.

ui/src/app/views/workspace-list.tsx
```diff
@@ -83,66 +84,108 @@ const WorkspaceCardMenu: React.FunctionComponent<{
...
+  async handleShareDialogClose() {
     // Share workspace publishes to current workspace,
     // but here we aren't in the context of a workspace
     // so we need to clear it.
     currentWorkspaceStore.next(undefined);
-    this.setState({sharing: false});
-    this.props.reload();
+    this.setState({
+      sharing: false,
+      userRoles: null});
```
blrubenstein: Opt nit: I don't think it actually matters if you clear userRoles, because we always reload above, right?

ui/src/app/views/workspace-list.tsx
```diff
@@ -152,20 +195,27 @@ export class WorkspaceCard extends React.Component<
...
+  }
+
   render() {
-    const {wp, userEmail} = this.props;
-    const {bugReportOpen, bugReportError, confirmDeleting,
-      sharing, workspaceDeletionError} = this.state;
+    const {userEmail, wp} = this.props;
+    const {bugReportError, bugReportOpen, confirmDeleting, loadingData,
+      sharing, userRoles, workspaceDeletionError} = this.state;
     const permissionBoxColors = {'OWNER': '#4996A2', 'READER': '#8F8E8F', 'WRITER': '#92B572'};
```
blrubenstein: Opt: Existing error, but we should probably update these to use the colors variables if we can.

api/src/main/resources/workbench.yaml
```diff
@@ -1050,7 +1050,12 @@ paths:
         - workspaces
       consumes:
         - application/json
-      description: Shares a workspace with users
+      description: >
+        Shares a workspace with the given users and permissions. Note that the entire ACL must
```
ericsong: This seems non intuitive to me and doesn't really reflect the intent of the user IMO. "I want to add a user to the share list" vs "I want to set the share list to this + someone new".

Not suggesting we change this now but do you think it would be worth changing in the future? Aside from the benefit of (IMO)  a simpler API, it looks like we'll gain a functional improvement in decreased loading times by not requiring the fetch ACL call.

ui/src/app/components/popups.tsx
ericsong: I definitely agree with using an interface over `any` but I'm curious where the prop's interface name shows up in stack traces.

At some point, we had made a collective decision to name Props/State interfaces within a file to just `Prop` and `State` since their scope can be inferred from the file it resides in.

But it looks like there may be a reason to reconsider?

api/src/main/resources/workbench.yaml
gjuggler: I mostly agree!

The challenge is that this API and the share-workspace UI component are both used in two places: here in the workspace-list page, but also in the workspace about page. In the context of the workspace list, it makes sense to be a "quickly add one collaborator" type action. But from the about page, it's explicitly about editing / removing the ACLs across the whole workspace.

If we split out a new API, then we'll be maintaining two highly-related API endpoints instead of one. It may be a net gain, but is certainly beyond the scope of this bugfix.

I'm not sure what the right motivation might be to follow-up on a decision here... it's not enough of an obvious win for me to want to immediately track a tech debt ticket, but I could be convinced otherwise.

(I think we'll bump up against this API again when we start allowing users to self-publish workspaces, since it will involve making sure these share actions are round-trip safe even when we have a "all users" entry in the Terra ACL. At that point we may want to make our APIs less of a pass-through to Firecloud.

ui/src/app/components/popups.tsx
gjuggler: The props rename didn't have anything to do with stack traces... I just wanted to leave some breadcrumb that this component actually expects a 'target' property. And the reason I didn't name it "Props" is that this file has a bunch of exported components, so the next time someone tries to create a typed props interface, they'd end up colliding in name.

(The 'target' prop was especially confusing to me, because 'target' is used elsewhere in this file with an entirely different meaning and type. TBH I found the code style in this file frustrating, because it's all super loosely-typed, has zero comments and no unit tests. In an ideal world our core components would have the most tests, type-safety, and well-documented interfaces – but we sort of have the inverse here :\)

ui/src/app/views/workspace-list.tsx
gjuggler: No it doesn't matter for functionality, but for cleanliness I figured userRoles should only be populated during the time when it's up-to-date and meant to actually be used. It might become stale e.g. if the workspaces list has been updated but we haven't explicitly reloaded the userRoles state variable yet.

ui/src/app/views/workspace-list.tsx
gjuggler: Heh, I'll leave that as part of the broader colors cleanup. I don't want to introduce new styling bugs as part of this fix.

PR #2316, opened 2019-07-08
[RW-2993][risk=no] fixing broken BQ test.

freemabd:
fixing broken BQ test
adding BadRequest exception around subtype
and adding a extra dao test case for criteria


api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/util/CriteriaLookupUtil.java
```diff
@@ -183,6 +184,10 @@ public CriteriaLookupUtil(CBCriteriaDao cbCriteriaDao) {
       putLeavesOnParent(byParent, parents, leaves);
     }
     for (CriteriaLookupUtil.FullTreeType treeType : childrenByTreeType.keySet()) {
+      if (treeType.subType == null) {
+        throw new BadRequestException(
+            "Bad Request: Please provide a valid criteria subtype. null is not valid.");
```
ericsong: nit: I think throwing a BadRequestException here may violate the layered architecture principle. It looks like this class is mostly just business logic and is higher in the architecture stack so it shouldn't be concerned with a lower level concept like HTTP.

My preference would be for this class to throw an IllegalArgumentException here and the calling class can catch it and throw a BadRequestException. This keeps HTTP related code separate from this business/persistence class.

freemabd: Moved BadRequestException to calling class.

PR #2319, opened 2019-07-08
[risk=no][RW-2905] Text and URL updates for the user registration flow

jmthibault79:
Steps 1-7 of RW-2905.  Step 8 will require more effort, and can be done independently.

### DIFF
--- a/ui/src/app/views/registration-dashboard.tsx
+++ b/ui/src/app/views/registration-dashboard.tsx
@@ -51,7 +51,7 @@ const styles = reactStyles({
 });
 
 function redirectToGoogleSecurity(): void {
-  let url = 'https://myaccount.google.com/security';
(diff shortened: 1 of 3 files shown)


ui/src/app/views/registration-dashboard.tsx
```diff
@@ -51,7 +51,7 @@ const styles = reactStyles({
 });
 
 function redirectToGoogleSecurity(): void {
-  let url = 'https://myaccount.google.com/security';
+  let url = 'https://myaccount.google.com/u/2/signinoptions/two-step-verification/enroll';
```
jacmrob: we need to add the authed user option here so that it will go directly to that user's 2FA set up and not ask the person to sign into their google identity again

jacmrob: should add `?authuser=<username>` where username is `profile.username`

PR #2322, opened 2019-07-08
[RW-2998][risk=low] Use an API token rather than a password for Jira auth

calbach:
Already generated API token per https://confluence.atlassian.com/cloud/api-tokens-938839638.html and pushed to all `jira-login.json` files. After the next release, will update these JSON files to omit the password.


calbach:
UI test failure is unrelated.

PR #2325, opened 2019-07-09
RW-2953  [Risk=no] Read-only behavior for Cohorts, Datasets

NehaBroad:
@blrubenstein : For user who have read access:
1) the save/update/analyze button will be disabled with tooltip
2) The plus icon for cohort and concept will be disabled with tooltip
3) User can SELECT Cohort/concept and value and can still PREVIEW data

Screen Shot 2019-07-09 at 11 53 57 AM

Screen Shot 2019-07-09 at 11 54 05 AM


ui/src/app/views/dataset-page.spec.tsx
```diff
@@ -5,13 +5,18 @@ import {Button} from 'app/components/buttons';
 import {dataSetApi, registerApiClient} from 'app/services/swagger-fetch-clients';
 import {currentWorkspaceStore, NavStore, urlParamsStore} from 'app/utils/navigation';
 import {DataSetPage} from 'app/views/dataset-page';
-import {CohortsApi, ConceptsApi, ConceptSetsApi, DataSetApi} from 'generated/fetch';
+import {
+  CohortsApi, ConceptsApi, ConceptSetsApi, DataSetApi, WorkspaceAccessLevel} from 'generated/fetch';
```
blrubenstein: Nit: Can you put all these on their own line if there's too many for one line?

ui/src/app/views/dataset-page.tsx
```diff
@@ -162,12 +159,14 @@ export const ValueListItem: React.FunctionComponent <
...
-  return <a data-test-id={dataTestId} href={path}
+const plusLink = (dataTestId: string, path: string, disable: boolean) => {
+  return <TooltipTrigger data-test-id='plus-icon-tooltip' disabled={!disable}
+                         content='Requires Owner or Writer permission'>
+    <Clickable disabled={disable} data-test-id={dataTestId} href={path}
             onClick={e => {navigateAndPreventDefaultIfNoKeysPressed(e, path); }}>
     <ClrIcon shape='plus-circle' class='is-solid' size={16}
-             style={styles.selectBoxHeaderIconLinks}/>
-  </a>;
+             style={{fill: disable ? colorWithWhiteness(colors.dark, 0.4) : colors.accent}}/>
```
blrubenstein: Opt: You could make the style object for this class be a function that takes disabled and returns the right color? (Would avoid getting rid of the object.)
I'm good either way on this, but I know we have a preference for JS objects.

ui/src/app/views/dataset-page.tsx
```diff
@@ -162,12 +159,14 @@ export const ValueListItem: React.FunctionComponent <
...
-const plusLink = (dataTestId: string, path: string) => {
-  return <a data-test-id={dataTestId} href={path}
+const plusLink = (dataTestId: string, path: string, disable: boolean) => {
+  return <TooltipTrigger data-test-id='plus-icon-tooltip' disabled={!disable}
+                         content='Requires Owner or Writer permission'>
+    <Clickable disabled={disable} data-test-id={dataTestId} href={path}
             onClick={e => {navigateAndPreventDefaultIfNoKeysPressed(e, path); }}>
     <ClrIcon shape='plus-circle' class='is-solid' size={16}
-             style={styles.selectBoxHeaderIconLinks}/>
-  </a>;
```
blrubenstein: Does moving from `<a>` to `<Clickable>` have any screen reader/href/link implications? If you cmd click the icon, does it open in a new tab still? I think we can keep the Clickable as an `<a>`

ui/src/app/views/dataset-page.tsx
```diff
@@ -371,11 +370,15 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
       }
     }
 
+    canWrite() {
```
blrubenstein: I'd make this a get method, which lets you use it as a property.
```
get canWrite() {
...
}
```

ui/src/app/views/dataset-page.tsx
```diff
@@ -162,12 +159,14 @@ export const ValueListItem: React.FunctionComponent <
     </div>;
   };
 
-const plusLink = (dataTestId: string, path: string) => {
-  return <a data-test-id={dataTestId} href={path}
+const plusLink = (dataTestId: string, path: string, disable: boolean) => {
```
blrubenstein: I think disable can probably be an optional property?

ui/src/app/views/dataset-page.tsx
NehaBroad: It opens it in the same tab. I wanted to keep it consistent  with the data tab page.

ui/src/app/views/dataset-page.tsx
NehaBroad: So since the other style in wrapped under reactStyle i had to create a new const

ui/src/app/views/dataset-page.tsx
blrubenstein: Okay, that seems fine for now. We should audit our components and see what makes sense moving forwards though.

ui/src/app/views/dataset-page.tsx
```diff
@@ -596,12 +607,15 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
               </div>
-              <Button data-test-id='save-button'
-                onClick ={() => this.setState({openSaveModal: true})}
-                disabled={this.disableSave()}>
-                {this.editing ? !dataSetTouched ? 'Analyze' :
+              <TooltipTrigger data-test-id='save-tooltip'
+                           content='Requires Owner or Writer permission' disabled={this.canWrite}>
+                <Button data-test-id='save-button'
+                  onClick ={() => this.setState({openSaveModal: true})}
+                  disabled={this.disableSave() || !this.canWrite}>
```
blrubenstein: Opt: You might be able to fold the canWrite bit into disableSave, since I think the intent is the same?

PR #2327, opened 2019-07-09
[RW-3000][risk=low] Resource Card Refactoring - Notebooks

ericsong:
Putting up a WIP pull request so I can get some early input and see if this is pattern that people like.

This is very much WIP and only has a working but unpolished implementation for NotebookResourceCard. I am looking for feedback on the high level design and not the nitty gritty atm.

This would be replacing the code in [resource-card](https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/resource-card.tsx) and [resources](https://github.com/all-of-us/workbench/blob/master/ui/src/app/components/resources.tsx).

Some problems I had with our old pattern

- resource-card is too big and contains the logic+HTML for all actions across all resource types

- Adding new actions requires adding to the shared ResourceCardMenu which is why we have a base class with 10 optional handlers in its Props

- Related to above - having a bloated class with many optional parameters is counterproductive to the type system. It's hard to read a class and know which parameters are being used where, what parameters are actually required for this resource type, etc.

What's different here

- Every resource type will have its own component which contains all code specific to its type. Shared code is in a base ResourceCard class. 

- Adding new actions just involves adding to the `actions` list in the desired resource type.

- More specific types make it easier to understand how to create a resource card and how its parameters are used.


ui/src/app/views/notebook-resource-card.tsx
```diff
@@ -0,0 +1,181 @@
...
+  onUpdate: Function; // TODO eric: this could use a better name
+  onDuplicateResource: Function;
+}
+
+interface State {
+  showCopyNotebookModal: boolean;
+  renaming: boolean;
+}
+
+export class NotebookResourceCard extends React.Component<Props, State> {
```
jmthibault79: Not sure if this makes sense: but could this extend `ResourceCardTemplate` ?

jmthibault79:
I really like this approach.

ui/src/app/views/resource-card-template.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+});
+
+const defaultProps = {
+  marginTop: '1rem'
+};
+
+export interface Action {
+  displayName: string;
+  onClick: Function;
```
gjuggler: Overall: I like the approach of splitting things up! Whatever you can do to decompose the monolith seems sensible to me.

One thing stood out to me, though: down in L135, we're passing in a pretty funky object structure to this onClick handler. It's a slightly surprising pattern since onClick usually takes an event or no argument. I guess the goal here is for concrete "subclasses" to be able to call centrally-defined methods from the "superclass" over here?

On the one hand, it feels a bit like we're rubbing up against React's preference for composition over inheritance by doing inheritance without calling it inheritance... on the other hand, this is already much cleaner than the single component. With a bit of careful commenting & type-safety, this can be totally workable.

But I'd also be interested to see whether a bit of inheritance (as much as React tries to avoid it) might help here, e.g. by allowing the resourceCardFns to be somewhat-awkwardly passed as a function argument...

ericsong: Agreed, the resourceCardFns object is pretty funky. I can look into some react design patterns to see what other people have done.

ericsong: @gjuggler I found another pattern through using higher level components that allows us to reuse state related code. Lmk what you think!

ui/src/app/views/notebook-resource-card.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    async duplicateNotebook() {
+      workspacesApi().cloneNotebook(
+        this.props.resourceCard.workspaceNamespace,
+        this.props.resourceCard.workspaceFirecloudName,
+        this.props.resourceCard.notebook.name)
+        .then(() => {
+          this.props.onUpdate();
+        }).catch(e => {
+          this.props.onDuplicateResource(false);
```
gjuggler: What exactly is the purpose of this callback? It's only ever called with a 'false' argument here. Should we call it with a 'true' value at L138 before making the API call?

Actually... it seems like this may be redundant, as we already call the onUpdate callback to cause the notebook list to reload altogether. Is there some special handling that the notebook-list needs to do when a notebook is duplicated?

ui/src/app/views/notebook-resource-card.tsx
```diff
@@ -0,0 +1,188 @@
...
+                     onRename={(newName) => this.receiveNotebookRename(newName)}
+                     onCancel={() => this.setState({renaming: false})}
+                     hideDescription={true}
+                     oldName={this.props.resourceCard.notebook.name}
+                     nameFormat={(name) => this.fullNotebookName(name)}/>
+        }
+
+        <ResourceCardTemplate
+          actions={this.actions}
+          actionsDisabled={!this.writePermission}
```
gjuggler: I wonder if it might be more flexible for each action to have its own logic for deciding if it should be enabled or not. There are some scenarios where only a subset of actions should be enabled, right?

ui/src/app/views/notebook-resource-card.tsx
```diff
@@ -0,0 +1,188 @@
...
+        },
+        {
+          displayName: 'Copy to another Workspace',
+          onClick: () => this.setState({showCopyNotebookModal: true}),
+        },
+        {
+          displayName: 'Delete',
+          onClick: () => {
+            this.props.showConfirmDeleteModal(this.displayName, 'Notebook',
+              (onComplete) => this.deleteNotebook(onComplete));
```
gjuggler: Instead of having an onComplete callback passed through here, could you have the receiveDelete function be typed to always return a Promise? That may be a slightly-cleaner control flow (but only marginally cleaner, so not a huge deal if you leave as-is).

ui/src/app/views/notebook-resource-card.tsx
```diff
@@ -0,0 +1,188 @@
...
+import {Action, ResourceCardTemplate} from 'app/views/resource-card-template';
+import {RecentResource} from 'generated/fetch';
+import * as React from 'react';
+import * as fp from 'lodash';
+import {withErrorModal} from "app/views/with-error-modal";
+import {withConfirmDeleteModal} from "app/views/with-confirm-delete-modal";
+
+interface Props {
+  resourceCard: RecentResource; // Destructure this into used parameters only
+  onUpdate: Function; // TODO eric: this could use a better name
```
gjuggler: Obviously super-early nitpick, but these functions should eventually have tighter type-definitions. I've been seeing a fair amount of "Function" type defs elsewhere in our codebase, which leave us open to all sorts of confusion and type-related bugs :)

ui/src/app/views/resource-card-template.tsx
gjuggler: I think this looks a lot better! Thanks for following up.

ui/src/app/views/notebook-resource-card.tsx
ericsong: good catch. It got copied over while I was refactoring but its no longer necessary.

ui/src/app/views/notebook-resource-card.tsx
ericsong: tried it out and it looks good. first time actually creating my own promises

ui/src/app/views/notebook-resource-card.tsx
ericsong: sounds good. I actually didnt know you could specify function types. I'll swap it out wherever I can

ui/src/app/views/notebook-resource-card.tsx
ericsong: That would be cool. I'll add that refactoring when I run into those actions.

ericsong:
@gjuggler @jmthibault79 @blrubenstein 

Reviving this. I fixed it up and implemented the changes for notebook resource cards. I'll add the rest in follow up PRs

ui/tsconfig.json
```diff
@@ -16,6 +16,7 @@
     ],
     "lib": [
       "es2016",
+      "es2018.promise",
```
jmthibault79: where is this used?

ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -0,0 +1,187 @@
...
+  fullNotebookName(name) {
+    return !name || /^.+\.ipynb$/.test(name) ? name : `${name}.ipynb`;
+  }
+
+  resourceUrl(jupyterLab = false): string {
+    const {workspaceNamespace, workspaceFirecloudName, notebook} =
+      this.props.resourceCard;
+    const workspacePrefix = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}`;
+
+    const queryParams = new URLSearchParams([
```
jmthibault79: `queryParams` are no longer used, and can be removed here.

This also means that...
* the `jupyterLab` function param can be removed
* the notebook resource card Action "Open in Jupyter Lab" can be removed, and
* the feature flag `enableJupyterLab` can be removed

... but it might be a good idea to check with Product on this.

At a minimum, disable the "Open in Jupyter Lab" Action if the feature flag `enableJupyterLab` is false.

ui/tsconfig.json
ericsong: it gives us `.finally` on promises - https://github.com/all-of-us/workbench/pull/2327/files#diff-86ce5d10b2dc782d34328e59e16c73cfR120

ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -0,0 +1,187 @@
...
+import {workspacesApi} from 'app/services/swagger-fetch-clients';
+import colors from 'app/styles/colors';
+import {formatRecentResourceDisplayDate} from 'app/utils';
+import {navigateByUrl} from 'app/utils/navigation';
+import {RecentResource} from 'generated/fetch';
+import * as fp from 'lodash';
+import * as React from 'react';
+
+interface Props extends WithConfirmDeleteModalProps, WithErrorModalProps, WithSpinnerOverlayProps {
+  resourceCard: RecentResource;
```
gjuggler: Optional: maybe this would be too big of a change, but I'd propose changing this property name to "resource" since it's just data, not a card-like UI element.

ericsong: sounds good to me!

ui/src/app/components/with-spinner-overlay.tsx
```diff
@@ -0,0 +1,43 @@
+import {SpinnerOverlay} from 'app/components/spinners';
```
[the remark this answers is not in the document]
ericsong: I decided to rename all of the implementations within the `with-x` components to just be `show/hide`. I think that shows the distinction between the props and the impl without being too wordy.

ui/src/app/pages/analysis/notebook-resource-card.tsx
ericsong: we decided to just remove it altogether since it's currently unsupported

ui/src/app/pages/analysis/notebook-redirect/component.ts
```diff
@@ -241,11 +242,11 @@ export class NotebookRedirectComponent implements OnInit, OnDestroy {
     if (nbName.endsWith('.ipynb')) {
```
jmthibault79: I think this check can be removed now, since drop/append are idempotent.  Starting from 240, we can do something like:
```
const decodedName = decodeURIComponent(nbName);
this.notebookName = dropNotebookFileSuffix(decodedName);
this.fullNotebookName = appendNotebookFileSuffix(decodedName);
```

ui/src/app/pages/analysis/util.ts
```diff
@@ -0,0 +1,15 @@
+export function dropNotebookFileSuffix(filename: string) {
```
jmthibault79: I like it!  Could you also add `const notebookExtension = '.ipynb'` for internal use here?

ui/src/app/pages/analysis/notebook-redirect/component.ts
ericsong: oo. good catch!

ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  get displayName(): string {
+    return dropNotebookFileSuffix(this.props.resource.notebook.name);
+  }
+
+  get resourceUrl(): string {
+    const {workspaceNamespace, workspaceFirecloudName, notebook} =
+      this.props.resource;
+
+    return `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}
```
calbach: Newlines in backticks do not work as you might expect - this is injecting escaped spaces into the URL.

PR #2331, opened 2019-07-10
[RW-2688][Risk=no] Add error messaging to data set page

blrubenstein:

### DIFF
--- a/ui/src/app/views/dataset-page.tsx
+++ b/ui/src/app/views/dataset-page.tsx
@@ -418,30 +418,34 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
       } catch (ex) {
         const exceptionResponse = await ex.json() as unknown as ErrorResponse;
         let errorText: string;
-        if (exceptionResponse.statusCode === 400) {
-          if (exceptionResponse.message ===
-            'Data Sets must include at least one cohort and concept.') {
-            errorText = exceptionResponse.message;
-          } else if (exceptionResponse.message ===
-            'Concept Sets must contain at least one concept') {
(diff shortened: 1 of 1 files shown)


ui/src/app/views/dataset-page.tsx
```diff
@@ -410,6 +416,34 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
           selectedPreviewDomain: dataSetPreviewResp.domainValue[0].domain
         });
       } catch (ex) {
+        const exceptionResponse = await ex.json() as unknown as ErrorResponse;
+        let errorText: string;
+        if (exceptionResponse.statusCode === 400) {
```
NehaBroad: will it be neater to have a switch case on statusCode. its just difficult to read

ui/src/app/views/dataset-page.tsx
```diff
@@ -410,6 +416,34 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          if (exceptionResponse.message ===
+            'Timeout while querying the CDR to pull preview information.') {
+            errorText = 'Query to load data from the All of Us Database timed out. ' +
+              'Please either try again or export data set to a notebook to try there';
+          }
+        } else {
+          errorText = 'An unexpected error has occurred. ' +
+            'Please submit a bug using the contact support button';
+        }
+        this.setState({previewError: true, previewErrorText: errorText});
```
NehaBroad: do we need to reset this to false, in the scenario where we get an error and clicked preview button again

blrubenstein: it is reset to false when we close the error modal. The error text we don't need to reset because it changes next time you get an error anyway.

PR #2332, opened 2019-07-10
[risk=low] moving CDR versions for Bedford release.

freemabd:
Jira ticket describing work is here: https://precisionmedicineinitiative.atlassian.net/browse/RW-3008
moving CDR versions for Bedford release.

### DIFF
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -8,7 +8,7 @@
     "creationTime": "2017-12-26 00:00:00Z",
     "releaseNumber": 1,
     "numParticipants": 946237,
-    "cdrDbName": "cdr20181213"
+    "cdrDbName": "synth_r_2019q3_1"
   },
   {
     "cdrVersionId": 2,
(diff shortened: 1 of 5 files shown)


api/config/cdr_versions_prod.json
```diff
@@ -19,7 +19,7 @@
     "creationTime": "2018-11-13 15:09:27Z",
     "releaseNumber": 1,
     "numParticipants": 116157,
-    "cdrDbName": "r_2019q1_2"
+    "cdrDbName": "r_2019q1_3"
```
calbach: Is the new CDR `r_2019q2_1` already on the latest schema?

freemabd: I had to generate another version from R2018Q4R1 that includes the new criteria table. The prior version r_2019q1_2 only had the old criteria table and r_2019q2_1 was generated from R2019Q2R1.

calbach: Makes sense - my question was just to double-check whether the **other** CDR already has the new tables (All of Us Dataset v2 below).

freemabd: Ah.. sorry.. misunderstood your question. Yes that is correct that r_2019q2_1 has the new criteria tables.

PR #2337, opened 2019-07-10
[RW-2762][risk=no] Publish featured workspaces

jacmrob:


api/src/main/resources/workbench.yaml
```diff
@@ -1938,6 +1951,24 @@ definitions:
...
+        items:
+          $ref: "#/definitions/FeaturedWorkspace"
+
+  FeaturedWorkspace:
+    type: object
+    properties:
+      name:
+        type: string
+        description: the name of the workspace
+      id:
```
calbach: Is this sufficient? Shouldn't we need the namespace and the FC name? Maybe that's what is being referenced here, either way please name these more clearly

api/tools/src/main/java/org/pmiops/workbench/tools/ConfigLoader.java
```diff
@@ -69,6 +72,7 @@ public CommandLineRunner run(ConfigDao configDao) {
       String marshalledJson = gson.toJson(configObj, configClass);
       JsonNode marshalledNode = jackson.readTree(marshalledJson);
       JsonNode marshalledDiff = JsonDiff.asJson(newJson, marshalledNode);
+      log.info(marshalledDiff.toString());
```
calbach: The diff gets logged below as well - was this line for debugging?

ui/src/app/views/workspace-library.tsx
```diff
@@ -95,15 +98,26 @@ export const WorkspaceLibrary = withUserProfile()
...
     }
   }
 
-  async reloadPublishedWorkspaces() {
-    this.setState({workspacesLoading: true});
+  async loadFeaturedWorkspaces() {
+    try {
+      const resp = await featuredWorkspacesConfigApi().getFeaturedWorkspacesConfig();
+      this.setState({featuredWorkspaceIds: resp.featuredWorkspacesList.map(ws => ws.id)});
+      this.setState({workspacesLoading: false});
```
calbach: nit:combine with above `setState` call?

common-api/src/main/java/org/pmiops/workbench/config/FeaturedWorkspacesConfig.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * A class representing the featured workspaces configuration; parsed from JSON stored in the
+ * database. See {@link CacheSpringConfiguration}. This should be kept in sync with files in the
+ * config/ directory.
+ */
+public class FeaturedWorkspacesConfig {
+  public ArrayList<FeaturedWorkspace> featuredWorkspaces;
+
+  public static FeaturedWorkspacesConfig createEmptyConfig() {
```
calbach: unused?

common-api/src/main/resources/common_api.yaml
```diff
@@ -146,6 +146,16 @@ definitions:
         type: integer
         format: int32
 
+  FeaturedWorkspace:
```
calbach: You added this to both YAML files, it should only be in one place. workbench.yaml can reference models defined in this file

api/config/featured_workspaces_local.json
```diff
@@ -0,0 +1,8 @@
+{
+  "featuredWorkspaces": [
```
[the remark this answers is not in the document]
jacmrob: it won't cause issues but i will still revert

api/src/main/resources/workbench.yaml
jacmrob: "id" is the fc name; I don't think we need the namespace?

api/tools/src/main/java/org/pmiops/workbench/tools/ConfigLoader.java
jacmrob: yes good catch

ui/src/app/views/workspace-library.tsx
jacmrob: 👍

common-api/src/main/resources/common_api.yaml
jacmrob: 👍

api/src/main/resources/workbench.yaml
jacmrob: i will add tho just in case

ui/src/app/views/workspace-library.tsx
```diff
@@ -95,22 +99,41 @@ export const WorkspaceLibrary = withUserProfile()
...
-  async reloadPublishedWorkspaces() {
-    this.setState({workspacesLoading: true});
+  async loadFeaturedWorkspaces() {
+    try {
+      const resp = await featuredWorkspacesConfigApi().getFeaturedWorkspacesConfig();
+      const idToNamespace = new Map();
+      resp.featuredWorkspacesList.map(fw => idToNamespace.set(fw.id, fw.namespace));
+
+      this.setState({
+        featuredWorkspaces: this.state.workspaceList.filter(ws => idToNamespace.get(ws.workspace.id)
```
gjuggler: Not an issue for this PR, but I would have hoped / thought we could start ignoring workspace namespaces for logic like this, at least in a post-1ppw world (since the namespace should be equal to the project ID).

But maybe this isn't the case? @ericsong maybe you have clarity on this one... do we need to ~forever support projects which were created pre-1PPW, or will we eventually be able to drop the "namespace" concept?

ui/src/app/views/workspace-library.tsx
```diff
@@ -70,19 +71,22 @@ export const WorkspaceLibrary = withUserProfile()
...
+      featuredWorkspaces: [],
       workspaceList: [],
       workspacesLoading: true
     };
   }
 
-  componentDidMount() {
-    this.reloadPublishedWorkspaces();
+  async componentDidMount() {
+    this.loadPublishedWorkspaces();
```
gjuggler: I think we should be awaiting on each of these method calls so they run in sequence, right?

ui/src/app/views/workspace-library.tsx
```diff
@@ -95,22 +99,41 @@ export const WorkspaceLibrary = withUserProfile()
...
+  async loadFeaturedWorkspaces() {
+    try {
+      const resp = await featuredWorkspacesConfigApi().getFeaturedWorkspacesConfig();
+      const idToNamespace = new Map();
+      resp.featuredWorkspacesList.map(fw => idToNamespace.set(fw.id, fw.namespace));
+
+      this.setState({
+        featuredWorkspaces: this.state.workspaceList.filter(ws => idToNamespace.get(ws.workspace.id)
+          && idToNamespace.get(ws.workspace.id) === ws.workspace.namespace),
+        workspacesLoading: false
```
gjuggler: [Spinner nitpick]

I'm getting a little confused on the exact expected behavior of 'workspacesLoading' and display of the spinner.

Just thinking through the loading paths:

1) On component mount, we do [loadPublished --> loadFeatured]

2) When the user clicks refresh, we do [loadPublished].

For (1), I would expect the spinner to show until both calls are complete. But I think the spinner will be hidden once loadPublished is complete, ignoring loadFeatured. Is that correct?

For (2), I would expect some spinner to show while the API call is being made. But right now I think the refresh will happen without user feedback.

One way to possibly handle this would be to have loadPublished setState to {loadingWorkspaces: true} when first called, which would at least make case (2) show a spinner while loading.

jacmrob: that's what i was doing with it, but the workspaces load so quickly upon reload that the spinner basically just flashes on the screen.  isn't a great UX, although maybe it is better to have it implemented that way. thoughts?

ui/src/app/views/workspace-library.tsx
jacmrob: actually since we are starting on the published workspaces page, and we load featured upon click of the link, i'm thinking that maybe we should just load published on init

ui/src/app/views/workspace-library.tsx
jacmrob: it was unclear to me. my impression is that workspace ids were only unique across namespaces

ui/src/app/views/workspace-library.tsx
gjuggler: Oh – good point. I actually thought we were only going to show "featured" workspaces for the Bedford release (I missed this detail)... when we launch next week with a few DRC-approved featured workspaces, will they show up in both the published & featured tab, or only in the featured tab?

We should avoid an "empty screen" scenario where users have to click through to another tab before seeing anything meaningful... if that's how things are currently, then we're all good!

ui/src/app/views/workspace-library.tsx
gjuggler: I'm surprised workspaces are loading so quickly on load! In the user-specific workspace page it's been taking a few seconds regularly.

I agree, a super-quick spinner isn't something we should keep. Feel free to close this thread out then, & we can always follow-up with a UX tweak later on if needed.

ui/src/app/views/workspace-library.tsx
jacmrob: both tabs will be shown

ui/src/app/views/workspace-library.tsx
calbach: We discussed doing this but there are complications and it was not in scope for 1ppw. As of today, you need both namespace and name to create a link. We don't need to forever support old project/workspaces - after https://precisionmedicineinitiative.atlassian.net/browse/RW-2705 we will more freedom to change this behavior if desired, but it needs more thought.

My current position is to keep the current behavior, though we could discuss further:
- Just using name as-is in the URL doesn't work, it's not unique across all projects
- Just using namespace is possible, but it removes the user-generated "id" from the URL, which I'd consider a net loss, or neutral change. When factoring in the extra lookups / client code needed to support that, I'd argue against it.
- (most likely path) Forcing uniqueness on names OR creating a new third globally unique namespace I don't consider to be worth the effort as it puts us in a position of policing a global namespace; it's possible but it's an annoying constraint to enforce which would only be applicable within AoU. It would involve a lot of code changes as we wouldn't be able to directly call into any Terra APIs without first going through our own lookup layer to determine the project.

PR #2341, opened 2019-07-11
[RW-3009][risk=low] Workspace OWNER grants billing project user

calbach:
Will add tests and more docs, pushing up for initial review.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
@@ -53,6 +53,7 @@ import org.springframework.transaction.annotation.Transactional;
 @Service
 public class WorkspaceServiceImpl implements WorkspaceService {
 
+  private static final String FC_OWNER_ROLE = "OWNER";
   private static final Logger log = Logger.getLogger(WorkspaceService.class.getName());
 
   // Note: Cannot use an @Autowired constructor with this version of Spring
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -314,6 +318,24 @@ public Workspace updateWorkspaceAcls(
...
+
+    // Finally, keep OWNER and billing project users in lock-step. In Rawls, OWNER does not grant
+    // canCompute on the workspace / billing project, nor does it grant the ability to grant
+    // canCompute to other users.
+    for (String user : Sets.union(updatedAclsMap.keySet(), aclsMap.keySet())) {
+      String fromAccess =
+          aclsMap.getOrDefault(user, new WorkspaceAccessEntry().accessLevel("")).getAccessLevel();
+      WorkspaceAccessLevel toAccess =
+          updatedAclsMap.getOrDefault(user, WorkspaceAccessLevel.NO_ACCESS);
+      if ("OWNER".equals(fromAccess) && WorkspaceAccessLevel.OWNER != toAccess) {
```
blrubenstein: Will that equivalence work on the right? Shouldn't it be the same format as the left? `!"OWNER".equals(toAccess)`

If so, can we make the string "OWNER" a constant at the top of the file or method?

(Even if we don't need to use owner there, it may make sense to abstract "OWNER" anyway, since we're using it in multiple places)

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -314,6 +318,24 @@ public Workspace updateWorkspaceAcls(
       }
       throw new BadRequestException(usersNotFound);
     }
+
+    // Finally, keep OWNER and billing project users in lock-step. In Rawls, OWNER does not grant
+    // canCompute on the workspace / billing project, nor does it grant the ability to grant
+    // canCompute to other users.
+    for (String user : Sets.union(updatedAclsMap.keySet(), aclsMap.keySet())) {
```
blrubenstein: Nit: Can we make this `String username`? Just because I tend to associate user with an object.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Strings need to use `.equals()`; enums can use `==`. Extracted to an constant though

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Changed to `email` for consistency with the rest of the method.

calbach:
Tests fixed and added.

PR #2342, opened 2019-07-11
[RW-3042][risk=no] Don't block "create workspace" on free tier billing project

calbach:
The per user free tier project goes away with 1ppw, and this was only really relevant in local/test, so should be fine to remove the guard entirely.

### DIFF
--- a/ui/src/app/views/signed-in/component.html
+++ b/ui/src/app/views/signed-in/component.html
@@ -62,7 +62,7 @@
       <button class="btn btn-primary nav-link"
           [class.active]="createWorkspaceActive"
           routerLink="/workspaces/build"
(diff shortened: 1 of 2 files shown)


calbach:
@freemabd if possible please ensure this makes the release cut tomorrow AM.

ui/src/app/views/signed-in/component.html
```diff
@@ -62,7 +62,7 @@ <h6 style="padding: 0.5rem;">
       <button class="btn btn-primary nav-link"
           [class.active]="createWorkspaceActive"
           routerLink="/workspaces/build"
-          [disabled]="(hasDataAccess ? null : true) || !billingProjectInitialized"
```
jmthibault79: Should remove parens here for consistency with the other button.  OK to merge without this.

PR #2345, opened 2019-07-11
[RW-2500] [risk=low] Add a button to self-bypass access modules when unsafeSelfBypass is true

gjuggler:
This adds a button to the registration dashboard, only when the server config value is enabled:

Screen Shot 2019-07-11 at 9 24 09 AM

When you click bypass, it will send out bypass HTTP requests and ask you to reload the page.

Screen Shot 2019-07-11 at 9 24 25 AM

An un-bypass button will be shown if everything is bypassed already:

Screen Shot 2019-07-10 at 8 10 28 PM

### DIFF
--- a/ui/src/app/views/registration-dashboard.tsx
+++ b/ui/src/app/views/registration-dashboard.tsx
@@ -216,6 +216,9 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
   render() {
     const {taskCompletionMap, trainingWarningOpen} = this.state;
     const {betaAccessGranted, eraCommonsError, trainingCompleted} = this.props;
+    // todo: move this to the state
+    const canUnsafeSelfBypass = serverConfigStore.getValue().getUnsafeSelfBypassEnabled;
+
     return <div style={styles.registrationPage}
                 data-test-id='registration-dashboard'>
       <div style={styles.mainHeader}>Researcher Workbench</div>
--- a/ui/src/app/views/registration-dashboard.tsx
+++ b/ui/src/app/views/registration-dashboard.tsx
@@ -188,7 +188,7 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
     this.setState({showRefreshButton: false});
   }
 
-  getTaskCompletionList(): Array<boolean> {
+  get taskCompletionList(): Array<boolean> {
     const list = getRegistrationTasks().map((config) => {
       return this.props[config.completionPropsKey] as boolean;
     });
@@ -196,11 +196,11 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
   }
 
   allTasksCompleted(): boolean {
-    return this.getTaskCompletionList().reduce((acc, val) => acc && val);
+    return this.taskCompletionList.reduce((acc, val) => acc && val);
   }
 
   isEnabled(i: number): boolean {
-    const taskCompletionList = this.getTaskCompletionList();
+    const taskCompletionList = this.taskCompletionList;
 
     if (i === 0) {
       return !taskCompletionList[i];
@@ -227,7 +227,7 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
     }
   }
 
-  async setAllModulesBypassed(isBypassed: boolean) {
+  async setAllModulesBypassState(isBypassed: boolean) {
     this.setState({bypassInProgress: true});
 
     // TypeScript enum iteration is nonfunctional
@@ -254,7 +254,6 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
   render() {
     const {bypassActionComplete, bypassInProgress, trainingWarningOpen} = this.state;
     const {betaAccessGranted, eraCommonsError, trainingCompleted} = this.props;
-    // todo: move this to the state
     const canUnsafeSelfBypass = serverConfigStore.getValue().unsafeAllowSelfBypass;
 
     const anyBypassActionsRemaining = !(this.allTasksCompleted() && betaAccessGranted);
@@ -276,16 +275,18 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
           {!bypassActionComplete && <span>
             [Test environment] Self-service bypass is enabled:&nbsp;
             {anyBypassActionsRemaining &&
-              <Button onClick={() => this.setAllModulesBypassed(true)}
+              <Button style={{marginLeft: '0.5rem'}}
+                      onClick={() => this.setAllModulesBypassState(true)}
                       disabled={bypassInProgress}>Bypass all</Button>}
             {!anyBypassActionsRemaining &&
-              <Button onClick={() => this.setAllModulesBypassed(false)}
+              <Button style={{marginLeft: '0.5rem'}}
+                      onClick={() => this.setAllModulesBypassState(false)}
                       disabled={bypassInProgress}>Un-bypass all</Button>}
           </span>
           }
         </div>
       }
-      {!betaAccessGranted && !canUnsafeSelfBypass &&
+      {!betaAccessGranted &&
         <div data-test-id='beta-access-warning'
              style={{...baseStyles.card, ...styles.warningModal}}>
           <ClrIcon shape='warning-standard' class='is-solid'
@@ -304,7 +305,7 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
             </div>
             {!this.allTasksCompleted() &&
             <div style={styles.cardDescription}>{card.description}</div>}
-            {this.getTaskCompletionList()[i] ?
+            {this.taskCompletionList[i] ?
               <Button disabled={true} data-test-id='completed-button'
                       style={{backgroundColor: colors.success,
                         width: 'max-content',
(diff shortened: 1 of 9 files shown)


ui/src/app/views/registration-dashboard.tsx
```diff
@@ -151,25 +159,34 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
     super(props);
     this.state = {
       trainingWarningOpen: !props.firstVisitTraining,
-      taskCompletionMap: new Map<number, boolean>(),
```
gjuggler: I bumped this out of the state, since it's directly calculated from props values. The logic is now within the getTaskCompletionList function.

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -78,6 +81,7 @@ async function redirectToTraining() {
 export const RegistrationTasks = [
   {
     key: 'twoFactorAuth',
+    completionPropsKey: 'twoFactorAuthCompleted',
```
gjuggler: This explicitly connects each registration task to its location within the props passed into this component.

(Note: in the longer term, we should probably refactor this component to directly refer to the user profile. As-is, there's a lot of boilerplate coordination between homepage and registration-dashboard that could probably be gutted. But that's a refactor for another time...)

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -195,24 +208,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
+    // so just copy the whole list
+    const modules = [
+      AccessModule.COMPLIANCETRAINING,
+      AccessModule.ERACOMMONS,
+      AccessModule.TWOFACTORAUTH,
+      AccessModule.DATAUSEAGREEMENT,
+      AccessModule.BETAACCESS
+    ];
+
+    for (const module of modules) {
```
gjuggler: Side note: I learned that in order to run a series of promises in serial (rather than parallel), you need to use a for-loop and 'await' calls rather than array.forEach. I'll do a quick follow-up change to make our user admin UI do the same thing when it runs its bypass loop (https://github.com/all-of-us/workbench/blob/master/ui/src/app/views/admin-user-bypass.tsx#L57)

ui/src/app/views/registration-dashboard.tsx
jacmrob: 👍

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
     }
   }
 
+  async setAllModulesBypassed(isBypassed: boolean) {
+    this.setState({bypassInProgress: true});
+
+    // TypeScript enum iteration is nonfunctional
```
jacmrob: so sad :(

ui/src/app/views/registration-dashboard.tsx
jacmrob: 👍

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
     }
   }
 
+  async setAllModulesBypassed(isBypassed: boolean) {
```
jacmrob: nit: rename to `setAllModulesBypass` (not to necessarily assume we are setting bypassed=true)

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
+
+    // Call the callback (which refreshes the user's profile).
+    this.setState({bypassInProgress: false, bypassActionComplete: true});
+  }
+
   render() {
-    const {taskCompletionMap, trainingWarningOpen} = this.state;
+    const {bypassActionComplete, bypassInProgress, trainingWarningOpen} = this.state;
     const {betaAccessGranted, eraCommonsError, trainingCompleted} = this.props;
+    // todo: move this to the state
```
jacmrob: remove comment or do todo? or is this for a follow up PR?

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -169,25 +178,34 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
 
+  getTaskCompletionList(): Array<boolean> {
+    const list = getRegistrationTasks().map((config) => {
+      return this.props[config.completionPropsKey] as boolean;
+    });
+    return list;
+  }
+
+  allTasksCompleted(): boolean {
+    return this.getTaskCompletionList().reduce((acc, val) => acc && val);
```
jacmrob: nit: i think for a getter you don't need to use parens in function invocation

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
-
+      {canUnsafeSelfBypass &&
+        <div data-test-id='self-bypass'
+             style={{...baseStyles.card, ...styles.warningModal}}>
+          {bypassActionComplete &&
+            <span>Bypass action is complete. Reload the page to continue.</span>}
+          {!bypassActionComplete && <span>
+            [Test environment] Self-service bypass is enabled:&nbsp;
+            {anyBypassActionsRemaining &&
+              <Button onClick={() => this.setAllModulesBypassed(true)}
```
jacmrob: nit: add a `style={{marginLeft: '0.5rem'}}` to this button so it's not squished against text?

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,71 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
+                      disabled={bypassInProgress}>Bypass all</Button>}
+            {!anyBypassActionsRemaining &&
+              <Button onClick={() => this.setAllModulesBypassed(false)}
+                      disabled={bypassInProgress}>Un-bypass all</Button>}
+          </span>
+          }
+        </div>
+      }
+      {!betaAccessGranted && !canUnsafeSelfBypass &&
```
jacmrob: would we still wanna show this even if the user *can* selfbypass?  i am of the opinion that we should not hide user flow pieces for users with adminish privileges, in case someone wants to develop this piece or see it in action

ui/src/app/views/registration-dashboard.tsx
calbach: Typescript enum iteration normally works, but it appears to have relatively broken behavior when the value types are `any`, and this is how ts fetch codegen generates its enums for some reason. The normal way to write the enum would  be to use a string value.

gjuggler: Thanks – I can try that out as a follow-up (I'd rather not touch the AccessModule enum as part of this PR).

gjuggler:
OK @jacmrob this is ready for a re-review. I resolved all convo's where suggestions were directly incorporated into the latest commit.

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -169,25 +178,34 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
-    this.state.taskCompletionMap.set(2, props.eraCommonsLinked);
-    this.state.taskCompletionMap.set(3, props.dataUseAgreementCompleted);
   }
 
   componentDidMount() {
     this.setState({showRefreshButton: false});
   }
 
+  get taskCompletionList(): Array<boolean> {
+    const list = getRegistrationTasks().map((config) => {
```
calbach: nit: `list` doesn't appear to be serving a purpose, would just return here

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -169,25 +178,34 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
+      return this.props[config.completionPropsKey] as boolean;
+    });
+    return list;
+  }
+
+  allTasksCompleted(): boolean {
+    return this.taskCompletionList.reduce((acc, val) => acc && val);
```
calbach: opt_nit: I'm not sure I like it any better than what you have, but one alternative I came up with:

`this.taskCompletionList.every(v => v);` or `this.taskCompletionList.every(_.identity)`

ui/src/app/views/registration-dashboard.tsx
```diff
@@ -213,24 +227,72 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
...
+    ];
+
+    for (const module of modules) {
+      await profileApi().unsafeSelfBypassAccessRequirement({
+        moduleName: module,
+        isBypassed: isBypassed
+      });
+    }
+
+    // Call the callback (which refreshes the user's profile).
```
calbach: Is this stale? I don't see a callback here

gjuggler:
Thanks CH for the look-over. I applied your comments & resolved all threads.

PR #2347, opened 2019-07-11
[risk=no] turn on universal search in staging, stable and prod.

freemabd:
turn on universal search in staging, stable and prod.


dolbeew:
The failure was looking for a redux action which is no longer used with the feature flag enabled, so I've removed that test.

PR #2348, opened 2019-07-11
[RW-3058][RISK=NO] Selecting Survey Concept set on dataset page results in exception while retrieving value

NehaBroad:
If we select a survey in  data set it results in error while getting value for surveys. 
If we do end up showing the value , the big query will have to change to include ds_survey rather than observation. For bedford release i do not think its best to change the big query so close to release rather filter out survey in concept set for data set page.


gjuggler:
Neha, would you mind creating a ticket to track this issue (if one doesn't already exist)? It will help to have this PR associated w/ the ticket in the description, for better tracking of follow-up on the fix.

blrubenstein:
Good call @gjuggler

PR #2351, opened 2019-07-11
[risk=no] turning off elastic.

freemabd:
turning off elastic.


calbach:
Are the universal search tickets still outstanding?

PR #2353, opened 2019-07-11
[No ticket] [risk=no] Adds an 'opacity' prop to the spinner component

gjuggler:
While working on https://github.com/all-of-us/workbench/pull/2345, I found myself wishing I were able to make the spinner overlay semi-opaque to make it clearer to the user which part of the screen was currently "loading", by graying it out a little bit.

This PR adds a simple prop to our core Spinner component, which should allow that to happen. By default opacity is zero, which should make this change a no-op for all existing usage.

### DIFF
--- a/ui/src/app/components/spinners.tsx
+++ b/ui/src/app/components/spinners.tsx
@@ -5,7 +5,7 @@ import * as React from 'react';
 
 const styles = reactStyles({
   overlay: {
-    backgroundColor: 'rgba(200, 200, 200, 0)',
+    backgroundColor: new Color(colors.light).alpha(0).toString(),
     position: 'absolute', top: 0, left: 0, bottom: 0, right: 0,
     display: 'flex', justifyContent: 'center', alignItems: 'center',
     zIndex: 1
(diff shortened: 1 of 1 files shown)


ui/src/app/components/spinners.tsx
```diff
@@ -1,10 +1,11 @@
 import colors from 'app/styles/colors';
 import {reactStyles} from 'app/utils/index';
+import * as Color from 'color';
 import * as React from 'react';
 
 const styles = reactStyles({
   overlay: {
-    backgroundColor: 'rgba(0, 0, 0, 0)',
+    backgroundColor: 'rgba(200, 200, 200, 0)',
```
blrubenstein: Could we use `colors.dark` here? I think that would give a similar feel to this.

gjuggler: colors.dark is equivalent to rgb(74, 74, 74)... which is pretty far off from this. Closer would be colors.light (rgb(233, 236, 239)) -- I'll go with that, and can follow-up with Lou if / when this starts coming into use.

blrubenstein: That works for me. I noticed when working with Lou that he tended to use colors.dark with a bunch of whiteness mixed in over colors.light. I suspect that is because if he wanted to hue our backgrounds, he'd want to equivalently hue our overlays and disables?

PR #2354, opened 2019-07-11
[No ticket] [risk=no] Update tsconfig files to point to git-ignored directories

gjuggler:
I started trying to use IntelliJ's built-in support for Typescript auto-compilation the other week. It worked quite well, except many of the output files were getting dumped to a directory that wasn't in our .gitignore.

Maybe I'm misunderstanding the purpose of these tsconfig files, but this PR attempts to change our configs to more consistently point compiled output JS files to a single directory.

I'm very open to dropping this if we think it might break anyone's flow!


ui/src/tsconfig.app.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "outDir": "../out-tsc/app",
+    "outDir": "../dist/out-tsc/app",
```
calbach: What files are being put here? We should be careful with `dist` since we deploy this directory as static assets: https://github.com/all-of-us/workbench/blob/75c5a78792cc2f65969c5bf282b7f30098b317c1/ui/app.yaml#L27

I don't think subdirs are accessible (I'm not 100% sure about that), but it could mean more files to upload.

gjuggler: Ah – yikes, I didn't realize the purpose of that folder. I'd just been trying to follow the pattern of the root ui/tsconfig.json, which has "outDir": "./dist/out-tsc" (maybe that's incorrect as well?).

Maybe the most conservative approach would be to direct typescript compiler output to a root-level out-tsc folder which is .gitignore'd. I took a stab at something that feels reasonable (and brings this to a place where IntelliJ Typescript service works well), but I'm not sure how dangerous it might be re: our deployment setup. PTAL latest push.

Separately, I'm still trying to understand how each tsconfig file is used. Please correct me if I'm wrong here:

 1) ui/src/tsconfig.app.json is used by Angular to build the production app.
 2) ui/src/tsconfig.jest.json and ui/src/tsconfig.spec.json are for tests, referred to explicitly by the Jest and Angular build configs.
 3) ui/tsconfig.json is the "root" build config, defining common settings for all builds.
 4) ui/src/tsconfig.json I don't fully understand – is this ever used as the canonical tsconfig for any build process? Would you expect any impact if we removed this file?

calbach: I'm not sure what files are generated at that location; if it's the actual compiled JS, then that may be where we want the output to wind up.

Your description sounds right, and AFAICT you're right that (4) can be removed. For reference I generated a fresh project using the Angular CLI - it only output a single config file at `src/tsconfig.json` :man_shrugging:

gjuggler: When the project is compiled with IntelliJ's TypeScript service, a bunch of individual compiled .js files are what ends up in the outDir. I haven't been able to decide whether Angular or App Engine compile, or make use of, the outDir that was pointed at dist/out-tsc via the root /ui/tsconfig.json file. I did try running a deploy-ui command and the package was successfully built with my current PR snapshot, so I suspect it was non-essential.

I'll go ahead with this change – it should be easy to tell if something breaks on test when it gets deployed.

PR #2356, opened 2019-07-11
[RW-2705][risk=med] Bulk Clone Workspaces for 1PPW migration

ericsong:
Adding a `project.rb` tool which can be used to clone workspaces and move them from the old billing structure to 1PPW.

It works by iterating through our list of known workspaces, impersonating the creator of the workspace, and issuing a clone request on their behalf.


api/tools/src/main/java/org/pmiops/workbench/tools/BulkCloneWorkspacesTool.java
```diff
@@ -0,0 +1,328 @@
...
+import org.springframework.context.annotation.Primary;
+import org.springframework.context.annotation.Scope;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.security.access.method.P;
+
+@SpringBootApplication
+@EnableConfigurationProperties
+@EnableJpaRepositories("org.pmiops.workbench.db.dao")
+@EntityScan({"org.pmiops.workbench.db.model"})
+public class BulkCloneWorkspacesTool {
```
ericsong: Not the cleanest code but I think it's OK given its mostly one time use. It was hard to factor out functions given that the dependencies are only provided to `run`.

More looking for reviews on correctness and behavior here.

api/tools/src/main/java/org/pmiops/workbench/tools/BulkCloneWorkspacesTool.java
```diff
@@ -0,0 +1,329 @@
...
+                  + " seconds");
+          Thread.sleep(sleepIntervalSeconds * 1000);
+        }
+
+        providedUser = userDao.findUserByEmail(workspaceResponse.getWorkspace().getCreatedBy());
+        if (providedUser == null || providedUser.getDisabled() || !checkUserExistsInGoogle(directoryService, providedUser.getEmail())) {
+          continue;
+        }
+
+        impersonateUser(fireCloudService.getApiClientWithImpersonation(providedUser.getEmail()));
```
calbach: Note: One follow-up, if we want to establish this as a pattern in our tools we'll need to figure out a way to do this without mutating shared state like this. The next API call made by any dependency will now utilize this impersonated user cred, which is scary and could easily be unintentional.

PR #2357, opened 2019-07-11
[RW-2952][risk=low] Enable local/test welder

calbach:


calbach:
All known blocking bugs have now been addressed in Leo dev - this is good to go.

PR #2358, opened 2019-07-11
[RW-2924][risk=low] Sanitize notebook previews

calbach:
@gjuggler @jmthibault79 FYI

See code comments for known caveats here. Open to suggestions on how to better cover the `<style>` case.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceImplTest.java
+++ b/api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceImplTest.java
@@ -52,7 +52,7 @@ public class NotebooksServiceImplTest {
   @Autowired private CloudStorageService cloudStorageService;
 
   @Test
-  public void testBasicContent() throws Exception {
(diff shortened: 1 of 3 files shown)


api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+      return null;
+    }
+  }
+
+  @Autowired private NotebooksService notebooksService;
+  @Autowired private FireCloudService firecloudService;
+  @Autowired private CloudStorageService cloudStorageService;
+
+  @Test
+  public void testBasicContent() throws Exception {
```
ericsong: I would prepend all the tests with `readOnlyHtml_` to add some context. Otherwise looks good

calbach: Oops, good catch - done.

PR #2363, opened 2019-07-12
[RW-2599][risk=low] Save DUA version signed

andy7i:
DB, UI, API changes and tests for saving the DUA version that a user has signed.

### DIFF
--- a/api/db/changelog/db.changelog-master.xml
+++ b/api/db/changelog/db.changelog-master.xml
@@ -96,4 +96,7 @@
   <include file="changelog/db.changelog-86-drop-tos.xml"/>
   <include file="changelog/db.changelog-87-add-survey-concept-set.xml"/>
   <include file="changelog/db.changelog-88-published-workspaces.xml"/>
+  <include file="changelog/db.changelog-89-add-data-use-agreement-signed-version.xml"/>
+
+  <!-- Note: to update the DB locally, run the following: todo: add commands here-->
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-master.xml
+++ b/api/db/changelog/db.changelog-master.xml
@@ -98,5 +98,10 @@
   <include file="changelog/db.changelog-88-published-workspaces.xml"/>
   <include file="changelog/db.changelog-89-add-data-use-agreement-signed-version.xml"/>
 
-  <!-- Note: to update the DB locally, run the following: todo: add commands here-->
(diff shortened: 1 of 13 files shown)


ui/src/app/components/buttons.tsx
```diff
@@ -117,19 +117,17 @@ const computeStyle = ({style = {}, hover = {}, disabledStyle = {}}, {disabled})
 
 // Set data test id = '' in the child to prevent it from propegating to all children
 export const Clickable = ({as = 'div', disabled = false, onClick = null, ...props}) => {
-  // Don't propagate test IDs to the rendered child component.
-  delete props['data-test-id'];
   return <Interactive
-    as={as} {...props}
+    // `fp.omit` used to prevent propagation of test IDs to the rendered child component.
+    as={as} {...fp.omit(['data-test-id'], props)}
```
andy7i: I want to make a general function like `removeDataTestId(props)` to make this a bit more clear.

api/db/changelog/db.changelog-master.xml
```diff
@@ -96,4 +96,12 @@
   <include file="changelog/db.changelog-86-drop-tos.xml"/>
   <include file="changelog/db.changelog-87-add-survey-concept-set.xml"/>
   <include file="changelog/db.changelog-88-published-workspaces.xml"/>
+  <include file="changelog/db.changelog-89-add-data-use-agreement-signed-version.xml"/>
+
+  <!--
```
gjuggler: Nice addition! I really appreciate little breadcrumbs like this.

ui/src/app/components/buttons.tsx
gjuggler: That would be reasonable – but actually the fp.omit is pretty clear & concise as it stands.

ui/src/testing/stubs/profile-api-stub.ts
```diff
@@ -111,4 +111,9 @@ export class ProfileApiStub extends ProfileApi {
       resolve({});
     });
   }
+
+  public submitDataUseAgreement(dataUseAgreementVersion: number) {
+    return Promise.resolve(this.profile);
```
gjuggler: For consistency with the other stubs which actually modify the profile object, maybe you can add a line:

this.profile.dataUseAgreementSignedVersion = dataUseAgreementVersion;

PR #2365, opened 2019-07-15
[RW-1960][Risk=moderate] Fix angular production builds for staging, s…

blrubenstein:
…table, prod


blrubenstein:
@calbach, yes, these are the defaults. This doesn't solve the problem, but I am narrowing into solving it, and meeting with Isaac in two minutes to dig in with him on the react side of things.

PR #2373, opened 2019-07-16
[RW-2718][RW-2715][risk=no] Convert CB overview component to React

dolbeew:
Converts `list-overview` component to react. Everything should work the same as the current version, no design or functionality changes made. Also includes conversion of cohort save dialog (RW-2715)


ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+              </TooltipTrigger>
+              <TooltipTrigger content={<div>Review participant level data</div>}>
+                <Clickable style={{...styles.actionIcon, ...(disableIcon ? styles.disabled : {})}}
+                  onClick={() => this.navigateTo('review')}>
+                  <ClrIcon shape='copy' className='is-solid' size={30} />
+                </Clickable>
+              </TooltipTrigger>
+            </div>
+            <h2 style={styles.totalCount}>
+              Total Count: &nbsp;
```
jacmrob: I would change this logic to 
```
{this.hasErrors ? <span>error tooltip</span> : 
   {loading ? <Spinner /> : <span> totals count</span>}
}
```
more concise and avoids having to keep checking the error logic.  can you move the check of if total is undefined into `this.hasErrors`?   and what is the actual case where that will happen, if neither of the other error conditions are met?

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+  apiCallCheck: number;
+}
+
+export const ListOverview = withCurrentWorkspace()(
+  class extends React.Component<Props, State> {
+    dropdown: any;
+    constructor(props: Props) {
+      super(props);
+      this.state = {
+        error: false,
```
jacmrob: can the name of this var be more specific?  it's referring to errors that occur when making api calls right, not input errors.  it gets confusing below with `this.hasErrors` and other error checking logic

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+                -- <TooltipTrigger content={this.hasTemporalError ?
+                    'Please complete criteria selections before saving temporal relationship.' :
+                    `All criteria are suppressed. Un-suppress criteria to update the total count
+                     based on the visible criteria.`}>
+                  <ClrIcon style={{color: '#F57600'}} shape='warning-standard' size={18} />
+                </TooltipTrigger>
+              </span>}
+            </h2>
+          </div>
+          {error && !this.hasErrors && <div style={styles.totalError}>
```
jacmrob: see my naming comment where state is defined.  the whole "error && not error" phrasing is a bit confusing to me; it would be nicer if it were clear what kind of error this was

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+                  onClick={() => this.toggleChartMode()} />
+              </div>
+              <div style={{padding: '0.5rem 0.75rem'}}>
+                {chartData.size &&
+                  <ComboChart mode={stackChart ? 'stacked' : 'normalized'} data={chartData} />}
+              </div>
+            </div>
+          </div>}
+        </div>
+        {saveModal && <Modal>
```
jacmrob: nit: i would rename this state piece to "saveModalOpen" or something that better conveys what it is

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+            </div>}
+            {invalid && <div style={styles.invalid}>Cohort name is required</div>}
+            <TextInput style={{marginBottom: '0.5rem'}} value={name} placeholder='COHORT NAME'
+              onChange={(v) => this.setState({name: v, nameTouched: true})}
+              disabled={saving} />
+            <TextArea value={description} placeholder='DESCRIPTION' disabled={saving}
+              onChange={(v) => this.setState({description: v})}/>
+          </ModalBody>
+          <ModalFooter>
+            <Button style={{color: '#262262'}} type='link' onClick={() => this.setState({
```
jacmrob: nit: use colors

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+              saveModal: false, name: undefined, description: undefined, saveError: false,
+              nameTouched: false
+            })} disabled={saving}>CANCEL</Button>
```
jacmrob: nit: button styling automatically capitalizes; this doesn't need to be capitalized

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+            <Button type='primary' disabled={!name || saving} onClick={() => this.submit()}>
+              {saving && <Spinner style={{marginRight: '0.25rem'}} size={18} />}
+               SAVE
```
jacmrob: nit: see above

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
dolbeew: You're right, much cleaner this way. I had the undefined check on total because I was getting errors on component init, but I changed the initial value of loading to true and that error goes away.

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+    padding: '3px 5px'
+  },
+});
+
+interface Props {
+  searchRequest: any;
+  update: number;
+}
+
+interface State {
```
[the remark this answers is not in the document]
dolbeew: Done

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
dolbeew: Good point, it is confusing. I changed `error` to `apiError` and `hasErrors` to `definitionErrors`

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
```diff
@@ -0,0 +1,426 @@
...
+      if (this.props.update > prevProps.update && !this.hasErrors) {
+        this.setState({loading: true, error: false});
+        this.getTotalCount();
+      }
+    }
+
+    getTotalCount() {
+      try {
+        const {searchRequest} = this.props;
+        const localCheck = this.state.apiCallCheck + 1;
```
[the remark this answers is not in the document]
dolbeew: Yeah, this is an interesting one, not sure if there's a better way to do it. I'm using this variable to check if more api calls are made before this call returns. If `apiCallCheck` has changed in state since `localCheck` was set, another call has been made and I don't update state.

The problem occurs when you make several changes quickly that trigger a count calculation (toggle the temporal switch or suppress/enable search items). The first call would return and set loading to false and show an outdated count while to other calls are still pending. Eventually, the right count shows up, but it was confusing when the spinner goes away then the count changes again (possibly multiple times) several seconds later.

This is my attempt of making sure the state only gets updated with the most recent api call. Seems to work well, but there may be a better way to do it.

ui/src/app/cohort-search/list-overview/list-overview.component.tsx
dolbeew: Done.

PR #2374, opened 2019-07-16
[RW-2060] [risk=no] Add feature flag and code to enable VPC flow logs. Enable in test.

gjuggler:
Although we're likely going to roll this out rather quickly, I opted to keep this under control of a feature flag so it doesn't affect our prod project buffer without active enrollment in a separate PR.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
@@ -207,6 +207,13 @@ public class FireCloudServiceImpl implements FireCloudService {
     BillingApi billingApi = billingApiProvider.get();
     CreateRawlsBillingProjectFullRequest request = new CreateRawlsBillingProjectFullRequest();
     request.setBillingAccount("billingAccounts/" + configProvider.get().firecloud.billingAccountId);
+    if (configProvider.get().featureFlags.enableVpcFlowLogs) {
+      request.setHighSecurityNetwork(true);
+      request.setEnableFlowLogs(true);
+    } else {
+      request.setHighSecurityNetwork(false);
+      request.setEnableFlowLogs(false);
+    }
     request.setProjectName(projectName);
     retryHandler.run(
         (context) -> {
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -207,6 +207,13 @@ public void createAllOfUsBillingProject(String projectName) {
     BillingApi billingApi = billingApiProvider.get();
     CreateRawlsBillingProjectFullRequest request = new CreateRawlsBillingProjectFullRequest();
     request.setBillingAccount("billingAccounts/" + configProvider.get().firecloud.billingAccountId);
+    if (configProvider.get().featureFlags.enableVpcFlowLogs) {
```
calbach: opt: Could chain this request build

```
request = new CreateRawlsBillingProjectFullRequest()
  .billingAccount(...)
  .projectName(...)
  .highSecurityNetwork(enableVpcFlowLogs)
  .enableFlowLogs(enableVpcFlowLogs);
```

api/src/main/resources/fireCloud.yaml
```diff
@@ -791,6 +791,15 @@ definitions:
       billingAccount:
         type: string
         description: the id of the billing account to use, must start with 'billingAccounts/'
+      highSecurityNetwork:
+        type: boolean
+        description: Optional, false if not specified. If true, spin up all compute in a VPC network.
+      enableFlowLogs:
+        type: boolean
+        description: Optional, false if not specified. If true, enable flow logs within the high security network. Requires highSecurityNetwork = true.
+      servicePerimeter:
```
calbach: This has probably already been discussed elsewhere, but out of curiosity: do you have a plan for how this value will be provided? Single value per environment? Is there another API to create the perimeter, or is that a manual process?

gjuggler: Each AoU env will have a named perimeter (https://docs.google.com/document/d/1r2bRy8CRul5JUJ5k9PjhyEfy9zx-cmXXiNmAemRMixQ/edit#heading=h.8695vxj8hxsq). Perimeter setup is semi-automated, I think CA team wrote some scripts to document / streamline the process, but we can bucket it under "one-time Terra environment setup" for now. We decided against doing a full-blown API for perimeter management, since it felt like overkill (at least for now).

Doing the servicePerimeter work will be the next step here, gated on getting a synthetic CDR housed in test.firecloud.org.

PR #2375, opened 2019-07-16
[ticket=no][risk=no]remove deprecated anyListOf()

jmthibault79:


jmthibault79:
`anyListOf()` only existed for pre-Java-8 compatibility.  `anyList()` does the same thing.

PR #2379, opened 2019-07-16
[RW-2274][risk=low] Convert admin table to react

jacmrob:
Screen Shot 2019-07-16 at 4 58 47 PM

This PR also cleans up an annoying behavior in the user table that would cause the whole table to reload when a user is enabled/disabled.  Instead, I added an endpoint for fetching a single user (gated behind the same admin authority) so that the table only reloads the user that has been modified. 

The tooltips from the original table are not included as I couldn't find a way to set them as column headers in a primereact table.  However, as bypass has been released for a while I don't think we really need them anymore.


ui/src/app/views/admin-user.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+describe('AdminUser', () => {
+  let props: {profile: Profile};
+
+  const component = () => {
+    return mount(<AdminUser {...props}/>);
+  };
+
+  beforeEach(() => {
+    serverConfigStore.next({
```
gjuggler: Are all of these config properties actually relevant to the admin-user test setup? If not, it might be really useful to create a helper method somewhere that can be used by multiple tests to create a bog-standard ServerConfig for testing. This will help reduce the # of file edits required if we e.g. remove one of these config flags in the future.

PR #2380, opened 2019-07-17
[RW-3059] [RISK=NO] DataSet : Show preview data and values for Survey Concept SetNsaxena/get survey values

NehaBroad:
For survey the idea is to access ds_survey for joins and links directly.


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -327,14 +328,21 @@ private String getColumnName(CdrBigQuerySchemaConfig.TableConfig config, String
   @VisibleForTesting
   public ValuesLinkingPair getValueSelectsAndJoins(List<DomainValuePair> valueSetList, Domain d) {
     List<String> values =
-        valueSetList.stream().map(valueSet -> valueSet.getValue()).collect(Collectors.toList());
+        valueSetList.stream()
+            .map(valueSet -> valueSet.getValue().toUpperCase())
+            .collect(Collectors.toList());
     values.add(0, "CORE_TABLE_FOR_DOMAIN");
+
+    if (d == Domain.OBSERVATION) {
```
blrubenstein: Are all observation domains in survey? If so, is there a reason not to send up Domain.SURVEY from the UI?

ui/src/app/views/dataset-page.tsx
```diff
@@ -687,7 +696,7 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
                               justifyContent: 'center',
                               lineHeight: '32px',
                             }}>
-                       {previewRow.domain}
+                       {previewRow.domain === 'OBSERVATION' ? 'SURVEY' : previewRow.domain}
```
blrubenstein: Same thing, should we just move all 'OBSERVATION' to be 'SURVEY'? Are they identical?

NehaBroad: Ok so Concept set with domain Survey creation was failing as it was not able to find Survey corresponding OMOP table. Will check with Jon if we can use maybe observation table corresponding to Domain Survey? If yes will update accordingly

PR #2382, opened 2019-07-17
[RW-2275][risk=no] Convert admin review workspace to react

jacmrob:
This component had a note that the UI was a rough implementation that needed further design, so I went ahead and gave it the same style design as the other Admin table with a `DataTable`.  I also included some research purpose details as a field in the table as it may be relevant to reviewers (note that the absence of info in the screenshot is from legacy workspaces that don't have all of research purpose filled out; all "real" workspaces will have to have all of those fields filled out to be created successfully).  

Once approve/reject is selected a spinner appears in the place of the buttons and then when the api call finishes successfully the workspace row disappears. 

Screen Shot 2019-07-17 at 11 50 44 AM


jacmrob:
Just for comparison this is what the OG page looks like: 
Screen Shot 2019-07-17 at 1 40 32 PM

ui/src/app/views/admin-review-workspace.tsx
```diff
@@ -0,0 +1,130 @@
...
+    fontSize: 12
+  }
+});
+
+/**
+ * Review Workspaces. Users with the REVIEW_RESEARCH_PURPOSE permission use this
+ * to view other users' workspaces for which a review has been requested, and approve/reject them.
+ */
+export const AdminReviewWorkspace = withUserProfile()(class extends React.Component<
+  {profileState: {profile: Profile, reload: Function, updateCache: Function}},
```
NehaBroad: nit: we should use interface Props, and State like other files

PR #2383, opened 2019-07-17
[RW-2501] [risk=no] Turn on enforceRegistered in test.

gjuggler:
This PR flips the test config to begin enforcing the registered state. We think this will help us catch regressions, and perform more realistic performance testing, in our push-on-green test environment.

There should be two main impacts:

(1) The UI should now force-redirect new users to the "registration tasks" page in test. Use the self-bypass button to allow yourself entry.

(2) New workspaces should be created with an authorization domain (https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java#L133).

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -7,7 +7,7 @@
     "clusterMaxAgeDays": 7,
     "clusterIdleMaxAgeDays": 3,
     "debugEndpoints": false,
-    "enforceRegistered": false,
+    "enforceRegistered": true,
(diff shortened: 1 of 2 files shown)


gjuggler:
> This should also be enabled locally - right?
> 
> If yes - I think we should have a ticket to remove the `enforceRegistered` flag entirely.

Very reasonable – added to local config and filed https://precisionmedicineinitiative.atlassian.net/browse/RW-3093

PR #2385, opened 2019-07-18
[RW-3070][risk=low] Moves disease search into new component.

bopwerks:

### DIFF
--- a/ui/src/app/components/search-input.tsx
+++ b/ui/src/app/components/search-input.tsx
@@ -1,0 +1,225 @@
+import {TextInput} from 'app/components/inputs';
+import {TooltipTrigger} from 'app/components/popups';
+import colors, {colorWithWhiteness} from 'app/styles/colors';
+import {reactStyles} from 'app/utils';
+import * as fp from 'lodash/fp';
+import * as React from 'react';
+
+const styles = reactStyles({
+  dropdownMenu: {
+    display: 'block',
+    maxHeight: '12rem',
+    minHeight: '30px',
+    visibility: 'visible',
+    overflowY: 'scroll',
+    width: '35%',
+    marginTop: '.25rem',
+    zIndex: 100,
+    border: '1px solid #979797',
+    borderRadius: '5px',
+    backgroundColor: colors.white,
+  },
+  open: {
+    position: 'absolute',
+    backgroundColor: colors.white,
+    border: '1px solid'
+  },
+  box: {
+    borderRadius: '5px',
+    paddingTop: '0.2rem',
+    paddingLeft: '0.2rem',
+    color: colors.primary,
+  },
+  boxHover: {
+    background: 'rgb(234, 243, 250)',
+    color: '#262262',
+    paddingTop: '0.2rem',
+    paddingLeft: '0.2rem',
+  },
+  boxHoverElement: {
+    color: colors.primary,
+    margin: 0,
+    padding: 0,
+    fontFamily: 'Montserrat',
+    fontSize: '14px',
+    lineHeight: '32px',
+  },
+});
+
+// Fifo accepts a function 'fn' that returns a promise, and returns a
+// new function that returns a promise. Promises returned by the
+// resulting function return their result of the last-instantiated
+// promise. For example, if a0, a1 are the start and end times of a
+// computation represented by promise A, and b0, b1 are the start and end
+// times of a computation represented by promise B, and a0 < b0 < b1 <
+// a1, then the value returned by promise A at time a1 will be the result
+// of promise B, *not* promise A, because B was instantiated after A.
+function fifo(fn) {
+  let nextCallID = 1;
+  let lastCompletedCallID = 0;
+  let lastCompletedResult = null;
+  return (...args) => {
+    const thisCallID = nextCallID++;
+    return new Promise((accept, reject) => {
+      fn.apply(null, args).then((newResult) => {
+        if (thisCallID > lastCompletedCallID) {
+          lastCompletedCallID = thisCallID;
+          lastCompletedResult = newResult;
+        }
+        accept(lastCompletedResult);
+      }).catch((e) => {
+        reject(e);
+      });
+    });
+  };
+}
+
+export interface SearchInputState {
+  matches: Array<string>;
+  hover: Array<boolean>;
+  state: number;
+}
+
+export interface SearchInputProps {
+  enabled: boolean;
+  placeholder: string;
+  value: string;
+  onSearch: Function; // Returns a promise
+  tooltip: string;
+  onChange: Function; // Accepts the new input as a parameter
+}
+
+export class SearchInput extends React.Component<SearchInputProps, SearchInputState> {
+  // Forward declarations
+  _searchTermChangedEvent: Function;
+  _onSearch: Function;
+  _START: number;
+  _ACTIVE: number;
+  _SUGGEST: number;
+  _HOVER: number;
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      matches: [],
+      hover: [],
+      state: this._START,
+    };
+    this._onSearch = fifo(props.onSearch);
+    this._searchTermChangedEvent = fp.debounce(300, this._search);
+
+    // State machine states
+    this._START = 0;   // Input box lacks focus, drop-down invisible
+    this._ACTIVE = 1;  // Input box focused, drop-down invisible
+    this._SUGGEST = 2; // Input box focused, search results visible
+    this._HOVER = 3;   // Input box focused, search results visible, item activated
+  }
+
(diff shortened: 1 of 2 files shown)


ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+
+  render() {
+    return (
+      <div style={{position: 'relative'}}>
+        <TooltipTrigger
+          content={this.props.tooltip}
+          disabled={this.props.enabled}>
+          <TextInput
+            value={this.props.value}
+            style={{width: '90%', border: '1px solid #9a9a9', borderRadius: '5px'}}
```
blrubenstein: For the border, we should use the colors object. I think you could do this with: `border: {backtick}1px solid ${colorWithWhiteness(colors.dark, 0.7)}{backtick}` (Or something)

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+            placeholder={this.props.placeholder}
+            disabled={!this.props.enabled}/>
+        </TooltipTrigger>
+        {this._suggest() &&
+          <div data-test-id='drop-down'
+            style={{...styles.dropdownMenu, ...styles.open, minWidth: '90%'}}>
+            {this.state.matches.map((disease, j) => {
+              return (
+                <div key={j} style={this.state.hover[j] ? styles.boxHover : styles.box}
+                  onMouseOver={this._onMouseOver.bind(this, j)}
```
blrubenstein: If its not going from angular to react, you may be able to do this as `(j) => this._onMouseOver(j)` (And similar to below)

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+                  onMouseOut={this._onMouseOut.bind(this, j)}
+                  onMouseDown={this._onMouseDown.bind(this, disease)}>
```
blrubenstein: I'd remove specific references to disease here and below.

bopwerks: good catch, will fix

ui/src/app/views/workspace-edit.tsx
bopwerks: `j` is bound, though i suppose i could do `() => this._onMouseOver(j)`. tomayto/tomahto

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+      return;
+    }
+    this.props.onSearch(keyword).then((matches) => {
+      this.setState({
+        matches: matches,
+        state: matches.length > 0 ?
+          this.SUGGEST : this.ACTIVE,
+        hover: matches.map(() => false),
+      });
+    }).catch((e) => {});
```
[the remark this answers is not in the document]
bopwerks: i'll `console.error` it. i think the behavior -- not showing the drop down if there's no results -- is reasonable, though i'm not married to it.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+  }
+
+  _toggleHover(j) {
+    return this.state.hover.map((elt, i) => {
+      return (i === j) ? !elt : elt;
+    });
+  }
+
+  _search(keyword: string): void {
+    this.setState({matches: []});
```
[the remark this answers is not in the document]
bopwerks: it might be worth it, but since i'm still a react newbie i'll leave this for future-me to consider.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
 
 export enum WorkspaceEditMode { Create = 1, Edit = 2, Duplicate = 3 }
 
+export interface SearchInputState {
+  matches: Array<string>;
+  hover: Array<boolean>;
+  state: number;
+}
+
+function getDiseaseNames(keyword) {
```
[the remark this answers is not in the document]
bopwerks: good question. i could go either way. it's a useful function on its own, but it's currently only used by one class. i think the former outweighs the latter so i'll leave it for now but not married to the decision.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+  return fetch(encodeURI(url)).then((response) => {
+    return response.json();
+  }).then((matches) => {
+    const labeledMatches = fp.filter((elt) => elt.hasOwnProperty('label'))(matches);
+    const diseases = fp.map((elt) => elt['label'])(labeledMatches);
+    return diseases;
+  });
+}
+
+class SearchInput extends React.Component<any, SearchInputState> {
```
ericsong: Can we declare the `Props` explicitly? So we know what inputs this component takes

ericsong: And how about moving this into its own file? We have been trying to stick to 1 component per file where possible. Less stuff to digest per file and better enforces separation between components.

ui/src/app/views/workspace-edit.tsx
```diff
@@ -360,6 +360,148 @@ export const LabeledCheckBox = (props) => {
...
+
+  _search(keyword: string): void {
+    this.setState({matches: []});
+    keyword = keyword.trim();
+    if (!keyword) {
+      return;
+    }
+    this.props.onSearch(keyword).then((matches) => {
+      this.setState({
+        matches: matches,
```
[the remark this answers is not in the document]
bopwerks: Good catch. Fixed

ui/src/app/views/workspace-edit.tsx
bopwerks: Fixed

ui/src/app/views/workspace-edit.tsx
bopwerks: Done

ui/src/app/components/search-input.tsx
```diff
@@ -0,0 +1,225 @@
...
+  matches: Array<string>;
+  hover: Array<boolean>;
+  state: number;
+}
+
+export interface SearchInputProps {
+  enabled: boolean;
+  placeholder: string;
+  value: string;
+  onSearch: Function; // Returns a promise
```
ericsong: I believe you can type function parameters and returns, so you can turn that comment into code!

ui/src/app/components/search-input.tsx
```diff
@@ -0,0 +1,225 @@
...
+        }
+        accept(lastCompletedResult);
+      }).catch((e) => {
+        reject(e);
+      });
+    });
+  };
+}
+
+export interface SearchInputState {
```
ericsong: we try not to export `State` since its usually unnecessary for outside components to know how this component is implemented.

ui/src/app/components/search-input.tsx
```diff
@@ -0,0 +1,225 @@
...
+  },
+  box: {
+    borderRadius: '5px',
+    paddingTop: '0.2rem',
+    paddingLeft: '0.2rem',
+    color: colors.primary,
+  },
+  boxHover: {
+    background: 'rgb(234, 243, 250)',
+    color: '#262262',
```
ericsong: @blrubenstein will be very happy if you use the shared colors class :)

ui/src/app/components/search-input.tsx
```diff
@@ -0,0 +1,225 @@
...
+  dropdownMenu: {
+    display: 'block',
+    maxHeight: '12rem',
+    minHeight: '30px',
+    visibility: 'visible',
+    overflowY: 'scroll',
+    width: '35%',
+    marginTop: '.25rem',
+    zIndex: 100,
+    border: '1px solid #979797',
```
ericsong: here as well (colors class)

ui/src/app/components/search-input.tsx
bopwerks: Fixed

ui/src/app/components/search-input.tsx
bopwerks: Fixed

ui/src/app/components/search-input.tsx
bopwerks: Fixed

ui/src/app/components/search-input.tsx
bopwerks: Fixed

PR #2386, opened 2019-07-18
[RW-2102][RISK=LOW]Create Demographics visual selector Data Set builder

NehaBroad:
1) Show one on data Set page a concept set with id : -1 name Demographics and domain person.
2) If user selects such concept set, create a concept set with name "Demographics" and concept Id: first from domain: gender/race/ethnicity.
3) Once the concept set is created step 1 will not be required as it will list out all the created concept set on data set page which will include Demographics as well
3)However Demographics concept set is filtered out in the concept set list on Data page

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -12,9 +12,11 @@ import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -26,10 +28,12 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.pmiops.workbench.cdr.dao.ConceptDao;
+import org.pmiops.workbench.cdr.model.Concept;
 import org.pmiops.workbench.db.dao.CohortDao;
 import org.pmiops.workbench.db.dao.ConceptSetDao;
 import org.pmiops.workbench.db.dao.DataSetDao;
 import org.pmiops.workbench.db.dao.DataSetService;
+import org.pmiops.workbench.db.model.CommonStorageEnums;
 import org.pmiops.workbench.db.model.DataSetValues;
 import org.pmiops.workbench.db.model.User;
 import org.pmiops.workbench.db.model.Workspace;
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -547,6 +561,33 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+      List<Concept> genderRaceConcept = conceptDao.findGenderRaceEthnicityFromConcept();
+      if (genderRaceConcept.size() > 0) {
+        org.pmiops.workbench.db.model.ConceptSet demographicsCS =
+            new org.pmiops.workbench.db.model.ConceptSet();
+        demographicsCS.setName("Demographics");
+        demographicsCS.setDomain(CommonStorageEnums.domainToStorage(Domain.PERSON));
+        demographicsCS.setCreationTime(now);
+        demographicsCS.setLastModifiedTime(now);
+        demographicsCS.setWorkspaceId(workspaceId);
+        Set<Long> conceptSet = new HashSet<Long>();
```
blrubenstein: Nit: Since we're already using concept set to refer to the object type, can we name this something like: `conceptsInSet`?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -547,6 +561,33 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+    if (request.getConceptSetIds().size() > 0 && request.getConceptSetIds().contains(-1l)) {
+      request.getConceptSetIds().remove(-1l);
+      // -1 Represent Demographics concepts set, Check for GENDER OR RACE OR ETHNICITY DOMAIN
+      // concept
+      // and add it as a dummy concept under the Demographics concept set
+      List<Concept> genderRaceConcept = conceptDao.findGenderRaceEthnicityFromConcept();
+      if (genderRaceConcept.size() > 0) {
+        org.pmiops.workbench.db.model.ConceptSet demographicsCS =
+            new org.pmiops.workbench.db.model.ConceptSet();
+        demographicsCS.setName("Demographics");
```
blrubenstein: What happens if a user has a concept set named "Demographics"?

NehaBroad: Yes this will restrict user to have a concept set named Demographics. 

However i am hoping this will go away once and if we allow user to add/have Demographic Domain concept sets under the concept set page.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
NehaBroad: How about conceptIdSets

blrubenstein: Sure, that seems fine.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Okay, we should make sure the messaging on that is reasonable. And block the concept set creation API from allowing that to be submitted for now.

PR #2387, opened 2019-07-18
[risk=low][RW-2707]notebook locking metadata endpoint

jmthibault79:
Also refactor storage.get() to return a better error

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -763,8 +763,8 @@ public class WorkspacesController implements WorkspacesApiDelegate {
     try {
       MessageDigest sha256 = MessageDigest.getInstance("SHA-256");
       byte[] hash = sha256.digest(toHash.getBytes(StandardCharsets.UTF_8));
-      final String hex = BaseEncoding.base16().lowerCase().encode(hash);
-      return hex;
+      // convert to printable hex text
+      return BaseEncoding.base16().lowerCase().encode(hash);
     } catch (final NoSuchAlgorithmException e) {
       throw new RuntimeException(e);
     }
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -4,7 +4,8 @@ import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
-import java.nio.charset.Charset;
+import com.google.common.io.BaseEncoding;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.sql.Timestamp;
@@ -761,15 +762,9 @@ public class WorkspacesController implements WorkspacesApiDelegate {
     String toHash = String.format("%s:%s", bucket, email);
     try {
       MessageDigest sha256 = MessageDigest.getInstance("SHA-256");
-      byte[] hash = sha256.digest(toHash.getBytes(Charset.forName("UTF-8")));
-
-      // convert to printable hex text
-      StringBuilder sb = new StringBuilder();
-      for (byte b : hash) {
-        sb.append(String.format("%02x", b));
-      }
-
-      return sb.toString();
+      byte[] hash = sha256.digest(toHash.getBytes(StandardCharsets.UTF_8));
+      final String hex = BaseEncoding.base16().lowerCase().encode(hash);
+      return hex;
     } catch (final NoSuchAlgorithmException e) {
       throw new RuntimeException(e);
     }
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -705,7 +705,7 @@ public class WorkspacesController implements WorkspacesApiDelegate {
   }
 
   @Override
-  public ResponseEntity<NotebookLockingMetadataResponse> notebookLockingMetadata(
+  public ResponseEntity<NotebookLockingMetadataResponse> getNotebookLockingMetadata(
       String workspaceNamespace, String workspaceName, String notebookName) {
 
     // Retrieving the workspace is done first, which acts as an access check.
--- a/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
@@ -4,6 +4,9 @@ import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import java.nio.charset.Charset;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.HashMap;
@@ -11,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
+import java.util.Set;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -46,6 +50,7 @@ import org.pmiops.workbench.model.CloneWorkspaceResponse;
 import org.pmiops.workbench.model.CopyNotebookRequest;
 import org.pmiops.workbench.model.EmptyResponse;
 import org.pmiops.workbench.model.FileDetail;
+import org.pmiops.workbench.model.NotebookLockingMetadataResponse;
 import org.pmiops.workbench.model.NotebookRename;
 import org.pmiops.workbench.model.ReadOnlyNotebookResponse;
 import org.pmiops.workbench.model.ResearchPurpose;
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -690,13 +695,86 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
 
   @Override
   public ResponseEntity<ReadOnlyNotebookResponse> readOnlyNotebook(
-      String workspace, String workspaceName, String notebookName) {
+      String workspaceNamespace, String workspaceName, String notebookName) {
```
jmthibault79: not really related - but I happened to notice this was wrong

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -71,8 +73,17 @@ String getImagesBucketName() {
   }
 
   private String readToString(String bucketName, String objectPath) {
+    return new String(getBlob(bucketName, objectPath).getContent()).trim();
+  }
+
+  // wrapper for storage.get() which throws NotFoundException instead of NullPointerException
```
jmthibault79: previously we used `storage.get().whatever()` without null checking.  it's still an error now, but a nicer one

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -127,10 +138,15 @@ public GoogleCredential getDefaultServiceAccountCredentials() throws IOException
   }
 
   @Override
-  public JSONObject getFileAsJson(String bucketName, String fileName) throws IOException {
+  public JSONObject getFileAsJson(String bucketName, String fileName) {
```
jmthibault79: unrelated misc change

api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceImplTest.java
```diff
@@ -52,7 +52,7 @@ User user() {
   @Autowired private CloudStorageService cloudStorageService;
 
   @Test
-  public void testGetReadOnlyHtml_basicContent() throws Exception {
+  public void testGetReadOnlyHtml_basicContent() {
```
jmthibault79: unrelated misc cleanup

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+    final String lastLockedUser = LOGGED_IN_USER_EMAIL;
+    final Long lockExpirationTime = Instant.now().plus(Duration.ofMinutes(1)).toEpochMilli();
+
+    final Map<String, String> gcsMetadata =
+        new HashMap<String, String>() {
+          {
+            put("lockExpirationTime", lockExpirationTime.toString());
+            put(
+                "lastLockedBy",
+                WorkspacesController.notebookLockingEmailHash(BUCKET_NAME, lastLockedUser));
```
calbach: opt: Hardcoding the current output of this function might be preferable, so you can see a concrete input rather than depending on the implementation within the test. This can help in sanity checking, e.g. if Welder was returning the hash with capital hex letters and this was expected lowercase.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -690,13 +695,86 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+  static String notebookLockingEmailHash(String bucket, String email) {
+    String toHash = String.format("%s:%s", bucket, email);
+    try {
+      MessageDigest sha256 = MessageDigest.getInstance("SHA-256");
+      byte[] hash = sha256.digest(toHash.getBytes(Charset.forName("UTF-8")));
+
+      // convert to printable hex text
+      StringBuilder sb = new StringBuilder();
+      for (byte b : hash) {
+        sb.append(String.format("%02x", b));
```
calbach: String.format is pretty slow and this would need to run 32*(num users in the acl); since we already pull in Guava - I might prefer an approach like this: https://stackoverflow.com/a/36477814

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+    org.pmiops.workbench.firecloud.model.Workspace fcWorkspace =
+        createFcWorkspace(testWorkspaceNamespace, testWorkspaceName, LOGGED_IN_USER_EMAIL);
+    fcWorkspace.setBucketName(BUCKET_NAME);
+    stubGetWorkspace(fcWorkspace, WorkspaceAccessLevel.OWNER);
+    stubFcGetWorkspaceACL(acl);
+
+    final String testNotebookPath = "notebooks/" + testNotebook;
+    doReturn(gcsMetadata).when(cloudStorageService).getMetadata(BUCKET_NAME, testNotebookPath);
+
+    // this is what I want to do, but it's broken somehow (never fails)
```
calbach: Looks like the paren is misplaced so this doesn't currently assert anything, should probably be:

```
assertThat(...getBody()).isEqualTo(expected);
```

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+          {
+            put("lockExpirationTime", lockExpirationTime.toString());
+            put(
+                "lastLockedBy",
+                WorkspacesController.notebookLockingEmailHash(BUCKET_NAME, lastLockedUser));
+            put("extraMetadata", "is not a problem");
+          }
+        };
+
+    // I can see that I have locked it myself, and when
```
calbach: incomplete?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+    assertNotebookLockingMetadata(gcsMetadata, expectedResponse, workspaceACL);
+  }
+
+  @Test
+  public void testNotebookLockingMetadataUnknownUser() {
+    final String lastLockedUser = "a-stranger@fake-research-aou.org";
+    final Long lockExpirationTime = Instant.now().plus(Duration.ofMinutes(1)).toEpochMilli();
+
+    final Map<String, String> gcsMetadata =
+        new HashMap<String, String>() {
```
calbach: Can this be `HashMap<>`, or does the anonymous class shenanigans below prevent that?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+          {
```
calbach: This is really interesting - I've never seen this pattern used. I generally use an immutable map builder where I can... Does this anonymous class have an unfavorable side-effects? e.g. performance overhead or debug implications?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2318,4 +2324,179 @@ public void testGetWorkspacesDoesNotGetsPublishedIfReader() {
...
+            put("lockExpirationTime", lockExpirationTime.toString());
+            // store directly in plaintext, to show that this does not work
+            put("lastLockedBy", lastLockedUser);
+            put("extraMetadata", "is not a problem");
+          }
+        };
+
+    // in case of accidentally storing the user email in plaintext
```
calbach: Nice test case

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jmthibault79: meaning "and when I locked it" - does this need clarification?

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jmthibault79: Agree that this is useful.  I added notebookLockingEmailHashTest() as an equivalent

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
jmthibault79: Thanks, updated

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jmthibault79: thanks, fixed

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jmthibault79: anon-class doesn't allow it.  removed regardless.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
jmthibault79: ImmutableMap.Builder looks better so I updated it to use that.

I should really have a look around Guava to see what else I'm missing.

api/src/main/resources/workbench.yaml
```diff
@@ -1303,6 +1303,31 @@ paths:
...
+    get:
+      tags:
+        - workspaces
+      summary: Get the locking metadata for a notebook
+      description: |
+        Get the locking metadata for a notebook, if available.  If the notebook is not locked
+        (e.g. at first creation) this may be empty.  Additionally, the caller of this endpoint
+        may not have the appropriate permissions to determine who is locking this notebook.
+        When this is the case, the lastLockedBy field is set to UNKNOWN.
+      operationId: "notebookLockingMetadata"
```
calbach: nit: most of our operation ids are verbs, since they translate into method names in swagger codegen. Of course, the endpoint immediately above this one violates this convention as well, so it could be classified as an existing issue.

api/src/main/resources/workbench.yaml
```diff
@@ -3441,3 +3466,11 @@ definitions:
       notebookName:
         type: string
 
+  NotebookLockingMetadataResponse:
+    type: object
+    properties:
+      lastLockedBy:
+        type: string
+      lockExpirationTime:
+        type: integer
```
calbach: It's probably worth adding a description here mentioning this is ms from the unix epoch (I assume it's ms).

api/src/main/resources/workbench.yaml
jmthibault79: ok, updated

PR #2390, opened 2019-07-22
[RW-2864][Risk=moderate] Srubenst/tracing

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
@@ -41,7 +41,6 @@ public class TracingInterceptor extends HandlerInterceptorAdapter {
     } catch (IOException e) {
       log.log(Level.WARNING, "Failed to setup tracing");
     }
-
   }
 
   /**
--- a/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
@@ -1,0 +1,79 @@
+package org.pmiops.workbench.interceptors;
+
+import io.opencensus.common.Scope;
+import io.opencensus.exporter.trace.stackdriver.StackdriverTraceConfiguration;
+import io.opencensus.exporter.trace.stackdriver.StackdriverTraceExporter;
+import io.opencensus.trace.SpanBuilder;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.samplers.Samplers;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.inject.Provider;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.firecloud.FirecloudApiClientTracer;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+// Interceptor to create a trace of the lifecycle of api calls.
+@Service
+public class TracingInterceptor extends HandlerInterceptorAdapter {
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final Logger log = Logger.getLogger(FirecloudApiClientTracer.class.getName());
+
+  private final Provider<WorkbenchConfig> workbenchConfigProvider;
+
+  private Map<Integer, Scope> spanMap;
+
+  @Autowired
+  public TracingInterceptor(Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.workbenchConfigProvider = workbenchConfigProvider;
+    this.spanMap = new HashMap<>();
+    try {
+      StackdriverTraceExporter.createAndRegister(StackdriverTraceConfiguration.builder().build());
+    } catch (IOException e) {
+      log.log(Level.WARNING, "Failed to setup tracing");
+    }
+
+  }
+
+  /**
+   * @param handler The Swagger-generated ApiController. It contains our handler as a private
+   *     delegate.
+   */
+  @Override
+  public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler)
+      throws Exception {
+
+    SpanBuilder requestSpanBuilder =
+        tracer.spanBuilder(
+            workbenchConfigProvider.get().server.shortName + request.getRequestURI());
+
+    if (workbenchConfigProvider.get().server.alwaysTrace) {
+      requestSpanBuilder.setSampler(Samplers.alwaysSample());
+    }
+
+    Scope requestSpan = requestSpanBuilder.startScopedSpan();
+    this.spanMap.put(request.hashCode(), requestSpan);
+    return true;
+  }
+
+  @Override
+  public void postHandle(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      Object handler,
+      ModelAndView modelAndView)
+      throws Exception {
+    if (spanMap.get(request.hashCode()) != null) {
+      spanMap.get(request.hashCode()).close();
+    }
+  }
+}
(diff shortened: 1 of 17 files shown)


api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Override
+  public void postHandle(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      Object handler,
+      ModelAndView modelAndView)
+      throws Exception {
+    if (spanMap.get(request.hashCode()) != null) {
+      spanMap.get(request.hashCode()).close();
```
calbach: This map will grow infinitely large, if you keep the map this should also remove the entry.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public <T> ApiResponse<T> execute(Call call, Type returnType) throws ApiException {
+    Response response;
+    T data;
+    try (Scope ss =
+        tracer
+            .spanBuilderWithExplicitParent("NotebooksApiCall", tracer.getCurrentSpan())
+            .startScopedSpan()) {
+      response = call.execute();
```
calbach: Can you just replace all this logic with `return super.execute()`? Wrapped with the session block?

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.squareup.okhttp.Call;
+import com.squareup.okhttp.Response;
+import io.opencensus.common.Scope;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.io.IOException;
+import java.lang.reflect.Type;
+
+/** Created by brubenst on 7/22/19. */
```
calbach: See other comment

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksConfig.java
```diff
@@ -86,7 +86,7 @@ public ClusterApi serviceClusterApi(@Qualifier(NOTEBOOKS_SERVICE_CLIENT) ApiClie
   }
 
   private ApiClient buildApiClient(WorkbenchConfig workbenchConfig) {
-    return new ApiClient()
+    return new NotebooksApiClientTracer()
```
calbach: We also have Mandrill and Jira API clients - may as well update those as well for consistency.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
```diff
@@ -0,0 +1,38 @@
...
+      response = call.execute();
+      data = handleResponseWithTracing(response, returnType);
+    } catch (IOException e) {
+      throw new ApiException(e);
+    }
+    return new ApiResponse<>(response.code(), response.headers().toMultimap(), data);
+  }
+
+  private <T> T handleResponseWithTracing(Response response, Type returnType) throws ApiException {
+    String targetUrl = response.request().httpUrl().encodedPath();
```
calbach: Is this not available on the `call`? If you can just get the operation/method name, I think that would probably be sufficient for tracing purposes (e.g. put this on the parent span).

If you need the full URL, then an annotation might be more appropriate. Opening and immediately closing a span seems wrong to me.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -87,6 +87,7 @@ public static WorkbenchConfig createEmptyConfig() {
     public String projectId;
     public String shortName;
     public String oauthClientId;
+    public boolean alwaysTrace;
```
gjuggler: Naming / docs nitpicks:

1) Please add a short comment describing the use of this flag (just enough so anyone on the team can quickly understand its usage / impact without further searching the codebase).

2) For some reason "alwaysTrace" sounds a bit misleading to me, as if there's a discrete time when we're "not tracing" when this flag is false. Maybe "traceAllRequests" is slightly clearer? Or a float-valued "traceSamplingRatio" to provide control over the sampling rate, though that's probably overkill.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
blrubenstein: I can't, because I need to manually open and close the trace for the URL.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
blrubenstein: It is not available on `call`, turns out the request is there, but it's just protected and not accessible. I tried doing it as an annotation, but it was actually harder to get to the information on the Stackdriver Trace URL, because it wasn't visible by default, and was just sorta off to the side. This was the cleanest I could get it on the UI.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksConfig.java
blrubenstein: Doing Mandrill, I don't see Jira here anymore. (at least, not under `api/src/main/java/org/pmiops/workbench`)

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
blrubenstein: I can make it `traceAllRequests`. I think the sampling ratio is overkill, especially cause I couldn't find documentation on what the default is. Will add a comment.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
blrubenstein: Screen Shot 2019-07-23 at 12 26 04 PM
Screen Shot 2019-07-23 at 12 35 09 PM

blrubenstein: The top one is with the span opened and closed, the bottom one is with the annotation. It requires opening the 'Show Events' button to see the URL. 

The other reason I like the span marker, is that that is where the response comes back, so it gives us the true length of the API call.

blrubenstein:
@gjuggler @calbach I think all comments have been addressed or replied to. PTAL

api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public void postHandle(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      Object handler,
+      ModelAndView modelAndView)
+      throws Exception {
+    ((Scope)request.getAttribute(TRACE_ATTRIBUTE_KEY)).close();
```
calbach: opt_nit: generally I'd put a space after a cast like this: `(Scope) request`

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
calbach: OK, lets go with this for now. The placement of the response in that trace raises a lot of questions already.

How about adding some additional text here to make the event clearer, e.g. `Response received: /api/workspaces`

api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private final Provider<WorkbenchConfig> workbenchConfigProvider;
+
+  @Autowired
+  public TracingInterceptor(Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.workbenchConfigProvider = workbenchConfigProvider;
+    try {
+      StackdriverTraceExporter.createAndRegister(StackdriverTraceConfiguration.builder().build());
+    } catch (IOException e) {
+      log.log(Level.WARNING, "Failed to setup tracing");
```
calbach: please include the exception here `, e)`

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksApiClientTracer.java
blrubenstein: Works for me. I will add response received. (I like that wording too)

PR #2393, opened 2019-07-22
[RW-2875][risk=no] refactor ElasticFilters to use CriteriaLookupUtil

freemabd:
This PR addresses updating the ElasticFilters to use the CriteriaLookupUtil. 
CriteriaLookupUtil was created to consolidate criteria lookup logic between BQ and Elastic.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -137,14 +137,12 @@ public final class ElasticFilters {
           filter.should(b);
         } else {
           // "should" gives us "OR" behavior so long as we're in a filter context, which we are.
-          // This
-          // translates to N occurrences of criteria 1 OR N occurrences of criteria 2, etc.
+          // This translates to N occurrences of criteria 1 OR N occurrences of criteria 2, etc.
           filter.should(
               QueryBuilders.functionScoreQuery(
                       QueryBuilders.nestedQuery(
                           // We sum a constant score for each matching document, yielding the total
-                          // number
-                          // of matching nested documents (events).
+                          // number of matching nested documents (events).
                           "events", QueryBuilders.constantScoreQuery(b), ScoreMode.Total))
                   .setMinScore(occurredAtLeast));
         }
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
@@ -22,14 +22,13 @@ import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.model.AttrName;
 import org.pmiops.workbench.model.Attribute;
 import org.pmiops.workbench.model.CriteriaType;
+import org.pmiops.workbench.model.DomainType;
 import org.pmiops.workbench.model.Modifier;
 import org.pmiops.workbench.model.ModifierType;
 import org.pmiops.workbench.model.SearchGroup;
 import org.pmiops.workbench.model.SearchGroupItem;
 import org.pmiops.workbench.model.SearchParameter;
 import org.pmiops.workbench.model.SearchRequest;
-import org.pmiops.workbench.model.TreeSubType;
-import org.pmiops.workbench.model.TreeType;
 
 /**
  * Utility for conversion of Cohort Builder request into Elasticsearch filters. Instances of this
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
```diff
@@ -44,43 +39,29 @@
...
-    ElasticFilters f = new ElasticFilters(criteriaDao);
+  public static QueryBuilder fromCohortSearch(CBCriteriaDao cbCriteriaDao, SearchRequest req) {
+    ElasticFilters f = new ElasticFilters(cbCriteriaDao);
     return f.process(req);
   }
 
-  /**
-   * A hand-maintained mapping of which criteria trees map to standard vs source concept IDs. This
-   * indicates which concept ID field we should be querying against for a SearchRequest on a given
-   * tree. TODO(RW-2249): Document or encode this more centrally.
```
calbach: Does universal search obsolete this ticket? https://precisionmedicineinitiative.atlassian.net/browse/RW-2249

If so, do you mind closing it?

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
```diff
@@ -75,147 +85,123 @@ public void setUp() {
...
+            .domain(DomainType.CONDITION.toString())
+            .type(CriteriaType.ICD9CM.toString())
+            .ancestorData(false)
+            .standard(false)
             .group(false);
   }
 
+  @After
+  public void tearDown() {
+    jdbcTemplate.execute("drop table cb_criteria_ancestor");
```
calbach: This test has `@DirtiesContext` - is this statement necessary?

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
```diff
@@ -800,13 +828,15 @@ public void testAgeAndVisitQuery() {
     SearchParameter ethParam =
         new SearchParameter()
             .conceptId(Long.parseLong(conceptId))
-            .type(TreeType.DEMO.toString())
-            .subtype(TreeSubType.ETH.toString())
```
calbach: Is subtype deprecated with universal search? Are we tracking removal somewhere?

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
```diff
@@ -782,15 +808,17 @@ public void testAgeQuery() {
...
     Object left = now.minusYears(34).minusYears(1).toLocalDate();
     Object right = now.minusYears(20).toLocalDate();
     SearchParameter ageParam =
         new SearchParameter()
-            .type(TreeType.DEMO.toString())
-            .subtype(TreeSubType.AGE.toString())
+            .domain(DomainType.PERSON.toString())
+            .type(CriteriaType.AGE.toString())
             .group(false)
+            .ancestorData(false)
```
calbach: I don't see this value being read in ElasticFilters; should it be? Or is that a different ticket?

api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticFilters.java
freemabd: Yes, there is a now a standard flag in the criteria table, so now need for the codebase to handle hand maintained source/standard mappings. Closed ticket.

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
freemabd: jdbcTemplate is used to create/insert data into the cb_criteria_ancestor table since the codebase currently doesn't have a need to implement a DAO for this table. Therefore the @DirtiesContext annotation seems to only recognized Hibernate persisted entities and their related tables. So the following tear down method needs to drop to table to keep from colliding with other test classes that also use this approach to create the cb_criteria_ancestor table. Not implementing this drop causes subsequent test suite runs to fail with a BadSqlGrammarException: StatementCallback; bad SQL grammar [create table cb_criteria_ancestor(ancestor_id integer, descendant_id integer)]; nested exception is org.h2.jdbc.JdbcSQLException: Table "CB_CRITERIA_ANCESTOR" already exists. Another approach could be to try catch the create of cb_criteria_ancestor and do nothing on BadSqlGrammarException. Added this comment to the tear down method.

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
freemabd: ancestor data property is referenced several times by the CriteriaLookupUtil which ElasticFilters is now using. 
https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/util/CriteriaLookupUtil.java#L107

api/src/test/java/org/pmiops/workbench/elasticsearch/ElasticFiltersTest.java
freemabd: subtype is still relevant for some trees(Measurement, Physical Measurement and Surveys(PPI)). Referencing it here: https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/cohortbuilder/querybuilder/util/CriteriaLookupUtil.java#L226
TreeType and TreeSubType are being replace with CriteriaType and CriteriaSubType in the new CB search. Tracking removal here: https://github.com/all-of-us/workbench/blob/master/api/src/main/resources/cb_search_api.yaml#L780

PR #2394, opened 2019-07-23
[RW-2906][RISK=NO] Replace Google docs in DUA page

NehaBroad:
Screen Shot 2019-07-23 at 12 10 07 AM

Note: Pending Lou Approval

### DIFF
--- a/ui/src/app/views/data-use-agreement.tsx
+++ b/ui/src/app/views/data-use-agreement.tsx
@@ -5,6 +5,7 @@ import {
   styles as headerStyles
 } from 'app/components/headers';
 import {TextInput} from 'app/components/inputs';
+import {Modal, ModalBody, ModalFooter, ModalTitle} from 'app/components/modals';
 import {TooltipTrigger} from 'app/components/popups';
 import {SpinnerOverlay} from 'app/components/spinners';
 import {profileApi} from 'app/services/swagger-fetch-clients';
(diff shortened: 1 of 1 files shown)


ui/src/app/views/data-use-agreement.tsx
```diff
@@ -233,28 +248,86 @@ const InitialsAgreement = (props) => {
...
+            implement the following sanctions if it is determined that an investigator has violated
+            the Code of Conduct:
+            </label>
+            <IndentedUl>
+              <IndentedLi>
+                Determine whether any action by the investigator is required to remedy the
+                violation.
+              </IndentedLi>
+              <IndentedLi>
+                Determine whether any action by the investigator is required to remedy the
```
jacmrob: this is a repeat of the above point

ui/src/app/views/data-use-agreement.tsx
```diff
@@ -233,28 +248,86 @@ const InitialsAgreement = (props) => {
...
+              <IndentedLi>
+                Prosecute the violator for breach of a contract with the federal government
+              </IndentedLi>
+            </IndentedUl>
+          </li>
+        </IndentedUl>
+      </label>
+    </ModalBody>
+    <ModalFooter>
+      <Button type='primary' onClick={props.onClose}>CLOSE</Button>
```
jacmrob: button will automatically capitalize; "Close" does not need to be in caps (and shouldn't, in case we decide to change the button format)

PR #2395, opened 2019-07-23
[RW-2872][risk=no] Add value_source_concept_id to elastic index for ppi data

freemabd:
Add value_source_concept_id to elastic index for ppi data from the observation OMOP table. When CB searches for any categorical data for surveys data(PPI) we now need to match against value_source_concept_id instead of value_as_concept_id.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
+++ b/api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
@@ -47,7 +47,7 @@ public class ElasticDocument {
           .put("visit_concept_id", 4)
           .put("value_as_number", 5)
           .put("value_as_concept_id", 6)
-          .put("value_source_concept_id", 7)
+          .put("value_as_source_concept_id", 7)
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/elasticsearch/ElasticDocument.java
```diff
@@ -62,6 +63,7 @@ String lower() {
                   .put("visit_concept_id", esType(ElasticType.KEYWORD))
                   .put("value_as_number", esType(ElasticType.FLOAT))
                   .put("value_as_concept_id", esType(ElasticType.KEYWORD))
+                  .put("value_source_concept_id", esType(ElasticType.KEYWORD))
```
calbach: nit: should this be `value_as_source_concept_id` to match the other naming?

freemabd: Renamed to value_as_source_concept_id and also added a test case to cover this case.

PR #2396, opened 2019-07-23
[RW-1919][risk=no] Convert workspace about to react

jacmrob:


jacmrob:
I made some changes to component names to better reflect what things are.  

`Workspace -> WorkspaceAbout`
Since this component is no longer the default home tab of workspaces, and it is not a wrapper around the workspace (there is a separate component for that).  What it *actually* is is the about page.
`WorkspaceAbout -> ResearchPurpose`
Since this component is *actually* just the research purpose part of the about page.  (It was given the name "WorkspaceAbout" to distinguish it when we updated research purpose; should have been called Research Purpose from the get-go)

ui/src/app/views/research-purpose.tsx
```diff
@@ -0,0 +1,153 @@
...
+      </FadeBox>;
+    }
+  }
+);
+
+@Component({
+  selector: 'app-workspace-about',
+  template: '<div #root></div>'
+})
+export class WorkspaceAboutComponent extends ReactWrapperBase {
```
blrubenstein: Should this be here? Do we need this? If not, we should remove this? If we do need it, we should update the selector and class name.

ui/src/app/views/research-purpose.tsx
```diff
@@ -0,0 +1,153 @@
...
+  sectionText: {
+    fontSize: '14px', lineHeight: '24px', color: colors.primary, marginTop: '0.3rem'
+  }
+});
+
+
+export const ResearchPurpose = fp.flow(withCurrentWorkspace(), withUserProfile())(
+  class extends React.Component<
+    {profileState: { profile: Profile, reload: Function }, workspace: WorkspaceData},
+    {publishing: boolean, workspacePermissions: WorkspacePermissions}> {
```
blrubenstein: Semantically, I wouldn't expect publishing to be owned by the Research Purpose component. Should this be moved into the WorkspaceAbout component?

ui/src/app/views/research-purpose.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <div style={styles.sectionHeader}>Anticipated findings from this study</div>
+        <div style={styles.sectionText}>{workspace.researchPurpose.anticipatedFindings}</div>
+        {workspace.researchPurpose.population && <div style={{marginBottom: '1rem'}}>
+          <div style={styles.sectionHeader}>Population area(s) of focus</div>
+          <TwoColPaddedTable header={false} style={{marginTop: '0.3rem'}}
+                             contentLeft={this.getSelectedPopulationsSlice(true)}
+                             contentRight={this.getSelectedPopulationsSlice(false)}/>
+        </div>
+        }
+        {profile.authorities.includes(Authority.FEATUREDWORKSPACEADMIN) &&
```
blrubenstein: I think this should be in the workspace about page, and isn't necessarily tied to research purpose?

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
-    display: 'flex', flexDirection: 'row'
+  mainPage: {
+    display: 'flex', justifyContent: 'space-between', alignItems: 'stretch',
+    height: 'calc(100% - 60px)'
   },
-  sectionHeader: {
-    fontSize: '16px', fontWeight: 600, color: colors.primary, marginTop: '1rem'
+  rightSidebar: {
+    backgroundColor: colorWithWhiteness(colors.primary, 0.85), marginRight: '-0.6rem',
+    paddingLeft: '0.5rem', paddingTop: '1rem', width: '22%', display: 'flex',
```
blrubenstein: How does this resize? Does it work well? (Just looking at the comparison of this and the above, I know its what it was before, but it may not resize well?)

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
+import {CdrVersion, Cohort, Profile, UserRole, WorkspaceAccessLevel} from 'generated/fetch';
+import {WorkspaceShare} from './workspace-share';
+
+
+interface WorkspaceState {
+  sharing: boolean;
+  cdrVersion: CdrVersion;
+  useBillingProjectBuffer: boolean;
+  freeTierBillingProject: string;
+  cohortsLoading: boolean;
```
blrubenstein: Are we loading cohorts anymore? It may be worth a quick state audit to make sure we're still using all of these? (I think `cohortList` is unused, for example)

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
+import {InfoIcon} from 'app/components/icons';
+import {Modal, ModalBody, ModalFooter, ModalTitle} from 'app/components/modals';
+import {TooltipTrigger} from 'app/components/popups';
+import {Spinner} from 'app/components/spinners';
+import colors, {colorWithWhiteness} from 'app/styles/colors';
+import {reactStyles, ReactWrapperBase, withUrlParams, withUserProfile} from 'app/utils';
+import {ResearchPurpose} from 'app/views/research-purpose';
+import {ResetClusterButton} from 'app/views/reset-cluster-button';
+import {CdrVersion, Cohort, Profile, UserRole, WorkspaceAccessLevel} from 'generated/fetch';
+import {WorkspaceShare} from './workspace-share';
```
blrubenstein: Nit: We should be consistent with view imports. I'd either make this `'app/views/workspace-share'` or change the above imports to be relative. (Preference for the absolute path, over the relative)

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
-  sectionText: {
-    fontSize: '14px', lineHeight: '24px', color: colors.primary, marginTop: '0.3rem'
+  shareHeader: {
+    display: 'flex', flexDirection: 'row', justifyContent: 'space-between', paddingRight: '0.5rem',
+    paddingBottom: '0.5rem'
+  },
+  infoBox: {
+    backgroundColor: colorWithWhiteness(colors.primary, 0.75), height: '2rem', width: '6rem',
+    borderRadius: '5px', padding: '0.4rem', marginRight: '0.5rem', marginBottom: '0.5rem',
+    color: colors.black, lineHeight: '14px'
```
blrubenstein: I think the color should probably be `colors.primary`? If it is text?

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
+            individual-level data) outside of the <i>All of Us</i> research environment without
+            approval from All of Us Resource Access Board (RAB).<br/>
+            Notebooks should rarely be downloaded directly from Google Cloud Console, as output
+            cells in Notebooks may contain sensitive individual level data.
+        </ModalBody>
+        <ModalFooter>
+          <Button type='secondary'
+                  onClick={() => this.setState({googleBucketModalOpen: false})}>Cancel</Button>
+          <Button onClick={() => this.openGoogleBucket()}
+                  style={{marginLeft: '0.5rem'}}>Continue</Button>
```
blrubenstein: It would be good if we could make this an actual <a> with an href for accessibility reasons (especially since its just opening a link) Is that doable?

ui/src/app/views/workspace-about.tsx
```diff
@@ -2,151 +2,276 @@ import {Component} from '@angular/core';
...
+          </React.Fragment> :
+          <Spinner size={50} style={{display: 'flex', alignSelf: 'center'}}/>}
+        <div>
+          <h3 style={{marginBottom: '0.5rem'}}>Workspace Information:
+            <TooltipTrigger content={WorkspaceInfoTooltipText()}>
+              <InfoIcon style={{margin: '0 0.3rem'}}/>
+            </TooltipTrigger>
+          </h3>
+          <div style={styles.infoBox} data-test-id='cdrVersion'>
+            <div style={styles.infoBoxHeader}>Dataset</div>
```
blrubenstein: Not sure whether this is an actionable PR comment, but now that we have data sets, this language makes me feel weird... might be worth checking with Lou

ui/src/app/app-routing.module.ts
```diff
@@ -91,7 +91,7 @@ const routes: Routes = [
             children: [
               {
                 path: '',
```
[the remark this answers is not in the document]
jacmrob: good idea; i did it and it wasn't an issue

ui/src/app/styles/colors.ts
```diff
@@ -24,6 +24,7 @@ export default {
     notebook: '#8BC990'
   },
   white: '#fff',
+  black: '#000',
```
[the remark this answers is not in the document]
jacmrob: I use it for a font

ui/src/app/views/research-purpose.tsx
jacmrob: we don't; I will get rid of it.  missed that during my naming update.

ui/src/app/views/research-purpose.tsx
jacmrob: yeah good thought.  although this is not user-facing yet and is a pretty rough sketch

ui/src/app/views/workspace-about.tsx
jacmrob: good catch

ui/src/app/styles/colors.ts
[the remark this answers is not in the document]
jacmrob: nevermind i scrapped it and used primary instead

ui/src/app/views/workspace-about.tsx
jacmrob: 👍

ui/src/app/views/workspace-about.tsx
jacmrob: i'll check with lou, gonna leave it for this PR though bc it is not a regression

ui/src/app/views/workspace-about.tsx
jacmrob: I'm not sure... I mean i don't see how to do that and have it also be a `Button` at the same time

blrubenstein: I think it's possible to make it a link, and simply style as a Button? Alternatively, wrap the continue within the button with an anchor link.

blrubenstein:
As a note, with this PR, you may be able to delete the angular TopBox component.

jacmrob:
I'm doing that in the follow-up PR

PR #2398, opened 2019-07-23
[RW-3115][risk=no] Don't copy blob metadata

calbach:
Verified via local manual testing. 

Specifically this avoids copying lock metadata when cloning a notebook. We currently have no other need to copy GCS metadata on any blob copy, so just turn it off everywhere.


calbach:
Feel free to merge on my behalf in the AM if you have no requested changes.

PR #2399, opened 2019-07-24
[risk=no][ticket=no]Add one notebookLockingEmailHashTest case

jmthibault79:


jmthibault79:
I think that's a good discussion to have.  In this particular case, it was not done with any meaningful intent.

PR #2400, opened 2019-07-25
Clean up deadcode from react conversion

jacmrob:


ui/src/app/styles/colors.ts
```diff
@@ -24,6 +24,7 @@ export default {
     notebook: '#8BC990'
   },
   white: '#fff',
+  black: '#000',
```
NehaBroad: Do we need black color?

PR #2401, opened 2019-07-25
[No Ticket][Risk=no] File Restructuring

blrubenstein:


ui/src/app/pages/analysis/notebook-list.tsx
```diff
@@ -8,13 +8,13 @@ import {ClrIcon, InfoIcon} from 'app/components/icons';
 import {TooltipTrigger} from 'app/components/popups';
```
calbach: Looks like this is implementing a change suggested in a comment thread around adding a `pages` subdir. Please update doc to include that so this is made clear to all reviewers

ui/src/app/app.module.ts
```diff
@@ -21,43 +21,42 @@ import {StatusCheckService} from './services/status-check.service';
...
+import {SignedInComponent} from './pages/signed-in/component';
+import {TopBoxComponent} from './pages/top-box/component';
+import {StigmatizationPageComponent} from './pages/workspace/stigmatization-page';
+import {WorkspaceAboutComponent} from './pages/workspace/workspace-about';
+import {WorkspaceEditComponent} from './pages/workspace/workspace-edit';
+import {WorkspaceLibraryComponent} from './pages/workspace/workspace-library';
+import {WorkspaceListComponent} from './pages/workspace/workspace-list';
+import {WorkspaceNavBarComponent} from './pages/workspace/workspace-nav-bar';
+import {WorkspaceShareComponent} from './pages/workspace/workspace-share';
+import {BugReportComponent} from './widgets/bug-report';
```
calbach: "widgets/" with this single component feels like a bit of a cop-out... There are other fairly specific modals living in `components/` - any reason not to put it there too?

ui/src/app/pages/analysis/notebook-redirect/component.spec.ts
```diff
@@ -9,11 +9,11 @@ import {Observable} from 'rxjs/Observable';
 
 import {NotebookComponent} from 'app/icons/notebook/component';
 import {ReminderComponent} from 'app/icons/reminder';
+import {TopBoxComponent} from 'app/pages/top-box/component';
```
calbach: nit: I wouldn't consider a top-box to really be a page; I might put this in `components` (I think having subdirs there should be fine)

ui/src/app/pages/analysis/notebook-list.tsx
blrubenstein: Done.

ui/src/app/app.module.ts
blrubenstein: Sure, will do.

ui/src/app/pages/analysis/notebook-redirect/component.spec.ts
blrubenstein: It isn't really, but it is also going away shortly (With Jackie's PR) so I wasn't too worried about it.

calbach: OK - sounds fine then

PR #2404, opened 2019-07-26
[RW-2921][Risk=no] Display appropriate message on preview page

blrubenstein:


calbach:
Thank you! Pleasantly surprised to see this change show up in test since I'd missed the PR.

PR #2405, opened 2019-07-26
[RW-1960][risk=no] Revert spammy content security policy reporting

calbach:
Will come up with a more lenient policy and fix forward. Right now it's generating 100s of console errors per page


blrubenstein:
Ouch, it wants to block all inline styles :'(

calbach:
Sending a PR to remove all inline styles next.

/s

PR #2409, opened 2019-07-29
[RW-4108] move cohorts/concept set routes behind /data route

jacmrob:
This also moves all cohort/concept set routes behind the /data route, so that the breadcrumbs will route back to the Data tab.  Additionally, I added a breadcrumb and header for "Search Concepts". 

Follow-up: the concepts/sets route seems unused now with new Data tab design?  Also some of the headers on those pages could be cleaned up to better reflect with things are.


ui/src/app/components/breadcrumb.tsx
```diff
@@ -61,29 +61,41 @@ export const getTrail = (type: BreadcrumbType, data): {label: string, url: strin
...
+          url: `${prefix}/data/cohorts/${cid}/review/participants/${pid}`}
       ];
     case BreadcrumbType.CohortAdd:
       return [
-        ...getTrail(BreadcrumbType.Workspace, data),
-        {label: 'Build Cohort Criteria', url: `${prefix}/cohorts/build`}
+        ...getTrail(BreadcrumbType.Data, data),
+        {label: 'Build Cohort Criteria', url: `${prefix}/data/cohorts/build`}
+      ];
+    case BreadcrumbType.SearchConcepts:
```
jacmrob: Note: I added this breadcrumb because it seemed weird that there wasn't one for Concept search.  will need a little more clean up with concepts/concept sets IMO, see my note in the description.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -547,7 +547,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
               accessLevel: ws.accessLevel
             }));
         }
-        navigate(['workspaces', workspace.namespace, workspace.id, 'data']);
+        navigate(['workspaces', workspace.namespace, workspace.id, 'about']);
```
jacmrob: IMO, workspace edit should route back to the about page because that is where a user starts when they go to edit their research purpose.  plus, that is where they will be able to see the results of their changes.  it doesn't make sense to me for them to be routed to the data tab

blrubenstein: They can also edit via the nav-bar, so technically we don't know where they started. But this seems reasonable to me.

blrubenstein:
@gjuggler, I am reviewing, Jackie wanted whomever of us got to it first to review, so you are good on this.

blrubenstein:
@calbach found a place where the change was not included, removing approval to fix that. The risk inherent in this though has swayed me over to trying the simpler fix first, and taking more time on this one.

ui/src/app/app-routing.module.ts
```diff
@@ -209,33 +171,72 @@ const routes: Routes = [
...
+                    data: {
+                      title: 'Search Concepts',
+                      breadcrumb: BreadcrumbType.SearchConcepts
+                    }
                   },
-                }, ]
+                  {
+                    path: 'concepts/sets',
+                    children: [{
+                      path: '',
```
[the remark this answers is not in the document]
jacmrob: We don't, but we still have the component, so it's unclear if we want to remove the route altogether.  I think that is work for a follow-up PR

PR #2417, opened 2019-07-30
[RW-1960][risk=low] Refactor sketchy Google Analytics initialization

calbach:
Move GA initialization into `index.html`, rather than constructing a Javascript script string piecemeal at runtime.

This allows us to define a tighter Content Security Policy (RW-1960), i.e. disallowing unsafe-inline scripts. Tested via manual deploy to GAE.

### DIFF
--- a/ui/src/app/pages/app/component.ts
+++ b/ui/src/app/pages/app/component.ts
@@ -1,4 +1,4 @@
-import {Component, Inject, OnInit} from '@angular/core';
+import {Component, OnInit} from '@angular/core';
 import {Title} from '@angular/platform-browser';
 import {
   ActivatedRoute,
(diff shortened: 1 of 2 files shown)


calbach:
Functionally, it should be a no-op. The relevant code here is just the initialization code block for Google Analytics (/ Google Tag Manager). We can't copy-paste the setup snippet exactly because we need to substitute in the analytics ID for the environment (`environment.gaId`), which I can't directly replace into index.html with our current build tooling: https://developers.google.com/analytics/devguides/collection/gtagjs

I'm working on RW-1960, which defines a security policy for the site - part of which restricts what scripts can be run on the page. The previous method of injecting a dynamically constructed script tag requires a more lenient policy to work. It was also overly complex and involved unnecessarily risky construction of a Javascript code string for execution.

gjuggler:
+1, thanks for the cleanup here CH. I was just about to start doing some analytics build-out in coordination w/ Helen and the CB team, so this was convenient timing.

PR #2419, opened 2019-07-31
[RW-3147][P2] reload collaborators after sharing

jacmrob:

### DIFF
--- a/ui/src/app/pages/workspace/workspace-about.tsx
+++ b/ui/src/app/pages/workspace/workspace-about.tsx
@@ -29,6 +29,7 @@ interface WorkspaceState {
   workspaceUserRoles: UserRole[];
   googleBucketModalOpen: boolean;
   publishing: boolean;
+  loadingWorkspaceUserRoles: boolean;
 }
 
 const styles = reactStyles({
@@ -99,7 +100,8 @@ export const WorkspaceAbout = fp.flow(withUserProfile(), withUrlParams())
       workspace: undefined,
       workspaceUserRoles: [],
       googleBucketModalOpen: false,
-      publishing: false
+      publishing: false,
+      loadingWorkspaceUserRoles: true
     };
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/workspace/workspace-about.tsx
```diff
@@ -226,7 +237,7 @@ export const WorkspaceAbout = fp.flow(withUserProfile(), withUrlParams())
                   data-test-id='workspaceShareButton'
                   onClick={() => this.setState({sharing: true})}>Share</Button>
         </div>
-        {workspaceUserRoles.length > 0 ?
+        {(!loadingWorkspaceUserRoles && workspaceUserRoles.length > 0) ?
```
bopwerks: If I'm reading this right, this fragment will only render when there are roles to show, regardless of whether the user roles are being loaded. I wonder -- could we set `workspaceUserRoles = []` in `onShare()` or `loadUserRoles()`, and delete the `loadingWorkspaceUserRoles` variable?

ui/src/app/pages/workspace/workspace-about.tsx
```diff
@@ -133,9 +135,12 @@ export const WorkspaceAbout = fp.flow(withUserProfile(), withUrlParams())
...
+    this.setState({loadingWorkspaceUserRoles: true});
     workspacesApi().getFirecloudWorkspaceUserRoles(workspace.namespace, workspace.id).then(
       resp => {
-        this.setState({workspaceUserRoles: fp.sortBy('familyName', resp.items)});
+        this.setState({
+          workspaceUserRoles: fp.sortBy('familyName', resp.items),
+          loadingWorkspaceUserRoles: false});
       }
     ).catch(error => {
       console.error(error);
```
bopwerks: See my comment below about possibly removing the `loadingWorkspaceUserRoles` variable, but if you decide to keep it, do you think we should set it to `false` here in the event of an error?

ui/src/app/pages/workspace/workspace-about.tsx
jacmrob: well technically it will be not be shown when roles are not "loading" AND there are roles to show, because we want to show a spinner in case of reload even when we already have a set of roles (rather than have the change suddenly pop up when the api call finishes).  I'm just personally not a fan of "erasing" a variable so to speak when it gets reloaded - it feels inaccurate.  because in the interim while we are loading the updated user roles the current roles still exist, so to unset them even just as logic for displaying a spinner feels semantically wrong to me.

ui/src/app/pages/workspace/workspace-about.tsx
jacmrob: I don't think so, because if something went wrong fetching the roles i think we still want to display a spinner (rather than out-of-date roles or no roles).  Something to implement in the future is a better UX for that failure case (we have a bad habit of quietly swallowing a lot of our api call failures, and a seemingly-arbitrary subset where we show a popup or error banner.  If there is not already a ticket for auditing that I will make one).

PR #2421, opened 2019-08-01
[RW-1960][risk=no] Add a content security policy in report-only mode

calbach:
CSP primer: https://developers.google.com/web/fundamentals/security/csp/
Spec: https://content-security-policy.com
tl;dr: this protects against XSS and injection of unexpected content on our site

- Enables a content security policy in report-only mode; this will log violations to a UI endpoint - we won't log the JSON report content, but we can track instances of violations via GAE logs
- Policy generated via the [CSP tester extension](https://chrome.google.com/webstore/detail/csp-tester/ehmipebdmhlmikaopdfoinmcjhhfadlf).
- A single policy is used for both dev and prod. This seems simplest to keep configuration understandable - there is not much risk in allowing connections across environment backends.
- This policy whitelists our exact inline script contents in `index.html`, this will need to be kept in sync manually. See comment in app.yaml - in the future this could instead be included in another JS file we load, for now while we're on a mixture of React/Angular this is non-trivial throwaway work and/or an extra browser request for another JS file.

Once enabled fully, this means that any requests from the Workbench to connect to sources that are not whitelisted will fail. This means we'll need to be diligent about updating this policy as we add/remove dependencies.

### DIFF
--- a/ui/app.yaml
+++ b/ui/app.yaml
@@ -12,6 +12,7 @@ handlers:
     Strict-Transport-Security: "max-age=31536000; includeSubDomains; preload"
     X-XSS-Protection: 1
     X-Content-Type-Options: "nosniff"
+    Content-Security-Policy-Report-Only: "default-src 'none'; report-uri /content-security-report"
 - url: /.*
   static_files: dist/index.html
   upload: dist/index.html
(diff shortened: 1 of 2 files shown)


gjuggler:
> * Enables a content security policy in report-only mode; this will log violations to a UI endpoint - we won't log the JSON report content, but we can track instances of violations via GAE logs

Thanks for the links to documentation – CSP has come a long way since I remembered dealing with it ~8 years ago :)

Can you give an example of tracking instances of violations via GAE logs? In particular, I'm having trouble understanding what setting the report-uri value to "/content-security-report" actually does. AFAICT, that isn't an actual endpoint in our API server – unless I'm missing something?

Otherwise this all looks good – thanks for the clear comments.

jaycarlton:
Is there ever a legitimate use case where a researcher might need an exception to this policy? Or does all this work at a lower level than any of the analysis?

gjuggler:
> Is there ever a legitimate use case where a researcher might need an exception to this policy? Or does all this work at a lower level than any of the analysis?

Good question. This is mostly unrelated; the connect-src and child-src entries for "https://leonardo.dsde-dev.broadinstitute.org" are what allows us to embed an iframe pointed at the Leonardo reverse proxy for a user's notebook; once they're inside that environment, their notebook has full access to the internet and isn't really subject to any of our CSP stuff.

calbach:
RE report URI: Yes - if we hit CSP violations, the browser should send a JSON report POST to our UI service at `/content-security-report`. As you noted, that path doesn't exist and the request will route to `index.html`, as all our UI paths do, thereby returning a 200 and a meaningless HTML response. The POST body will also be lost since it doesn't get logged by GAE.

However, we will see a request to `/content-security-report` in our GAE logs, which indicates someone somewhere hit a CSP violation, and we can investigate further manually. Initially I was just going to omit this, but figured having this limited signal would be better than nothing (and if you don't specify a report-uri in report-only mode, CSP logs an error which is annoying). With a log-based metric, this could eventually be used to drive an alert.

calbach:
And to be concrete about the usage in the short term, my plan is:
- roll out this PR to prod
- let it sit for a ~week
- search the logs for `/content-security-report` and confirm 0 requests
- enable CSP fully, i.e. drop the `-Report-Only`

PR #2425, opened 2019-08-01
[risk=low][RW-3161]Create projects in a Service Perimeter

jmthibault79:

### DIFF
--- a/api/config/config_test.json
+++ b/api/config/config_test.json
@@ -13,7 +13,7 @@
     "billingRetryCount": 2,
     "billingProjectBufferCapacity": 10,
     "xAppIdValue": "test-AoU-RW",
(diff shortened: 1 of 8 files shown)


api/config/config_test.json
```diff
@@ -69,6 +70,7 @@
     "enableLeoWelder": true,
     "useBillingProjectBuffer": true,
     "unsafeAllowDeleteUser": true,
-    "enableVpcFlowLogs": true
+    "enableVpcFlowLogs": true,
+    "enableVpcServicePerimeter": false
```
jmthibault79: should we also enable test?

jmthibault79:
FYI @jaycarlton

api/config/config_test.json
gjuggler: Yeah, I think we should go ahead and try it out. Better to find out sooner than later if it breaks something.

PR #2432, opened 2019-08-05
[RW-2705][risk=low] Add inactive status to workspaces

ericsong:
Also pulling in refactoring changes from https://github.com/all-of-us/workbench/pull/2356


api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
             .workspace(workspaceMapper.toApiWorkspace(savedWorkspace, toFcWorkspace)));
   }
 
+  // A retry period is needed because the permission to copy files into the cloned workspace is not granted transactionally
+  private void copyBlobWithRetries(String bucketName, Blob b, int retryPeriodSeconds) {
```
jaycarlton: If you use the `Duration` class here then you can guard more strongly against mis-uniting.
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    final int SLEEP_INTERVAL_SECONDS = 5;
+
+    try {
+      cloudStorageService.copyBlob(b.getBlobId(), BlobId.of(bucketName, b.getName()));
+    } catch (StorageException e) {
+      if (retryPeriodSeconds <= 0) {
```
jaycarlton: I think this name is misleading. I'd thought a retry period would be the interval of retrying. As used, it's really a total pool of time, and the interval is `retryPeriodSeconds / SLEEP_INTERVAL_SECONDS`. Likewise, `SLEEP_INTERVAL_SECONDS` is confusing.

I recommend having a separate parameter for `numRetries` and then just a fixed duration for the interval.

Even better would be to use a dedicated `Retryer` class like [this one](https://www.programcreek.com/java-api-examples/?api=com.github.rholder.retry.Retryer).

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
             .workspace(workspaceMapper.toApiWorkspace(savedWorkspace, toFcWorkspace)));
   }
 
+  // A retry period is needed because the permission to copy files into the cloned workspace is not granted transactionally
+  private void copyBlobWithRetries(String bucketName, Blob b, int retryPeriodSeconds) {
+    final int SLEEP_INTERVAL_SECONDS = 5;
+
+    try {
+      cloudStorageService.copyBlob(b.getBlobId(), BlobId.of(bucketName, b.getName()));
+    } catch (StorageException e) {
```
jaycarlton: Do you want to inspect the  `StorageException` you get back to make sure it's really an access problem?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+      if (retryPeriodSeconds <= 0) {
+        throw e;
+      }
+
+      log.info(
```
jaycarlton: I would think this should at least be a warning. Don't we try to make sure every account is provisioned with the right access programmatically? I.e. getting here is probably not just a user error or transient condition if I understand correctly.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -130,22 +155,32 @@ public Workspace get(String ns, String firecloudName) {
...
+    org.pmiops.workbench.firecloud.model.WorkspaceResponse fcResponse;
+    org.pmiops.workbench.firecloud.model.Workspace fcWorkspace;
+
+    WorkspaceResponse response = new WorkspaceResponse();
+
+    // This enforces access controls.
+    fcResponse = fireCloudService.getWorkspace(workspaceNamespace, workspaceId);
+    fcWorkspace = fcResponse.getWorkspace();
+
+    if (fcResponse.getAccessLevel().equals(WorkspaceService.PROJECT_OWNER_ACCESS_LEVEL)) {
```
jaycarlton: I'd make a dedicated mapping function for this. A function like `getWorkspace` shouldn't need to know these details.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -130,22 +155,32 @@ public Workspace get(String ns, String firecloudName) {
...
+      // We don't expose PROJECT_OWNER in our API; just use OWNER.
+      response.setAccessLevel(WorkspaceAccessLevel.OWNER);
+    } else {
+      response.setAccessLevel(WorkspaceAccessLevel.fromValue(fcResponse.getAccessLevel()));
+      if (response.getAccessLevel() == null) {
+        throw new ServerErrorException("Unsupported access level: " + fcResponse.getAccessLevel());
```
jaycarlton: Do you really want to throw here, or just put the error type in `WorkspaceResponse` and return it?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -326,29 +314,7 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-      response.setAccessLevel(WorkspaceAccessLevel.OWNER);
-    } else {
-      response.setAccessLevel(WorkspaceAccessLevel.fromValue(fcResponse.getAccessLevel()));
-      if (response.getAccessLevel() == null) {
-        throw new ServerErrorException("Unsupported access level: " + fcResponse.getAccessLevel());
-      }
-    }
-    response.setWorkspace(workspaceMapper.toApiWorkspace(dbWorkspace, fcWorkspace));
-    return ResponseEntity.ok(response);
+    return ResponseEntity.ok(workspaceService.getWorkspace(workspaceNamespace, workspaceId));
```
jaycarlton: OK, so this will just convert to a 500 if you throw inside `getWorkspace()`?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+      log.info(
+          "Service Account does not have access to bucket "
+              + bucketName
+              + ". Retrying for another "
+              + retryPeriodSeconds
+              + " seconds.");
+      try {
+        Thread.sleep(SLEEP_INTERVAL_SECONDS * 1000);
+      } catch (InterruptedException ex) {
+        e.printStackTrace();
```
jaycarlton: So here we're printing the exception, not logging an error or warning, and then just continuing?

I think we could use a simpler pattern here than this try/catch/try/throw logic.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -65,8 +71,8 @@
 import org.pmiops.workbench.cohorts.CohortMaterializationService;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.CdrVersionDao;
-import org.pmiops.workbench.db.dao.CohortCloningService;
-import org.pmiops.workbench.db.dao.ConceptSetService;
+import org.pmiops.workbench.cohorts.CohortCloningService;
```
jaycarlton: nit, and probably not fixable here: why is cohorts plural and conceptset singular?

api/src/main/java/org/pmiops/workbench/interceptors/InterceptorUtils.java
```diff
@@ -16,7 +23,14 @@ public static Method getControllerMethod(HandlerMethod handlerMethod) {
     Pattern apiControllerPattern = Pattern.compile("(.*\\.[^.]+)Api(Controller)");
     Method apiControllerMethod = handlerMethod.getMethod();
     String apiControllerName = apiControllerMethod.getDeclaringClass().getName();
-    String controllerName = apiControllerPattern.matcher(apiControllerName).replaceAll("$1$2");
+
+    String controllerName;
```
calbach: I assumed this was some temporary change for your oneoff branch - but I guess this actually is due to moving controllers out of the API directory. This deserves a comment

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -558,6 +521,33 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
             .workspace(workspaceMapper.toApiWorkspace(savedWorkspace, toFcWorkspace)));
   }
 
+  // A retry period is needed because the permission to copy files into the cloned workspace is not granted transactionally
```
calbach: Optional: could consider just leaving this functionality out in favor of https://precisionmedicineinitiative.atlassian.net/browse/RW-3187 being implemented soon (hopefully).

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -135,7 +136,8 @@
   private static final org.pmiops.workbench.cdr.model.Concept CONCEPT_2 =
       makeConcept(CLIENT_CONCEPT_2);
 
-  @Autowired WorkspacesController workspacesController;
+  @Autowired
```
calbach: nit: revert?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -144,6 +144,11 @@ public BillingProjectBufferEntry assignBillingProject(User user) {
     return entry;
   }
 
+  public double availableProportion() {
```
calbach: nit: would revert, since it isn't actually called from production code anywhere

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -18,6 +18,9 @@
   @Query("SELECT COUNT(*) FROM BillingProjectBufferEntry WHERE status IN (0, 2)")
   Long getCurrentBufferSize();
 
+  @Query("SELECT COUNT(*) FROM BillingProjectBufferEntry WHERE status=2")
+  Long numberOfAvailableEntries();
```
calbach: nit: see other revert comment

api/src/main/resources/workbench.yaml
```diff
@@ -2317,7 +2317,7 @@ definitions:
   WorkspaceActiveStatus:
     type: string
     description: Status of workspace
-    enum: [ACTIVE, DELETED]
+    enum: [ACTIVE, DELETED, INACTIVE]
```
calbach: We should document somewhere (probably here) what INACTIVE means

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -101,10 +102,34 @@ public FireCloudService getFireCloudService() {
 
   @Override
   public Workspace get(String ns, String firecloudName) {
-    return workspaceDao.findByWorkspaceNamespaceAndFirecloudNameAndActiveStatus(
+    Workspace workspace = workspaceDao.findByWorkspaceNamespaceAndFirecloudNameAndActiveStatus(
         ns,
         firecloudName,
         StorageEnums.workspaceActiveStatusToStorage(WorkspaceActiveStatus.ACTIVE));
```
calbach: Shouldn't this function `workspaceActiveStatusToStorage` need updating? I'm a bit disturbed that it seems our tests didn't fail here - I think active status may not have a test where others do - could you add that as well?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: I think I'm ok with casting a wider net by catching on `StorageException`. If its due to some other problem, we'll see the exception get bubbled up after the retries are over.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: Added. Looks like we already use Spring's `Retryable` in a few places but I decided to add the new library since there was some discussion of being less tied to Spring.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: The permission grant is eventually consistent so reaching this code block is part of expected behavior. I'm ok with bumping it up to a warning though since a flood of these would indicate a problem.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
ericsong: Yeah... This is something I'm working on for the hackathon! Going to try to make a clear translation layer between our API, DB, and Firecloud models.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
ericsong: I just moved this over from `WorkspaceController` so I'm a little hesitant to make functional changes. It could be that this condition is enforcing some access control.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: Should be. This line shouldn't execute if an exception is thrown inside of the `getWorkspace()` call. Some error handler should catch the exception and generate a 500.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
ericsong: We have a lot of naming issues to address 🙃

api/src/main/resources/workbench.yaml
gjuggler: +1 to this. This would also be a reasonable place to clarify what the expectations are around the DELETED state (e.g. is the workspace still "alive" in Terra? Has the underlying project been deleted? Curious minds want to know... we have comments here https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java#L312, but it should probably be documented at the API level.)

Re: naming... do we expect the INACTIVE state to be used by other migrations / workspaces in the future? Do we expect these workspaces to remain in this status indefinitely?

If this will effectively be a one-off status for migrations like this, it might be reasonable to make a more specific-sounding status like PENDING_DELETION_POST_MIGRATION.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
jaycarlton: Sounds good.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jaycarlton: Slick!

api/src/main/resources/workbench.yaml
jaycarlton: Would workspaces that had tripped the exfiltration alarm and been unplugged also be `INACTIVE` here? Or would that be a new status?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
jaycarlton: Nice. I _believe_ this was the one we used at another company.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jaycarlton: Might be a conversation for a later date. One way to look at it is that a 500 should really get someone's attention if they're monitoring things, so if this situation can happen in normal UI or API operations, then a 403 exception would be better I would think.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: I think I'd prefer to keep it since it is causing some failures already in our production. We can remove it once `RW-3187` lands

api/src/main/resources/workbench.yaml
ericsong: Probably a new status. It's cheap to add a new status and hard (at the moment) to figure out how a workspace ended up in a certain state.

I can give it the more specific name as @gjuggler  suggested.

Which should also take care of the need for a comment. I think the name is self explanatory.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: This line corresponds to a programming error or unexpected response from Firecloud, so 500 is probably appropriate. Any expected 403s would have propagated from line L162

ericsong:
@jaycarlton @calbach updated

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Not addressed I think

ericsong: Hmm, my tests didn't catch it again since the value is never actually read by our system. It reads off as a null which, for our purposes, behaves the same way `INACTIVE` should.

I'm not sure if I want to add a null check just to make the test fail but this could be something that gets covered by my hackathon project (adding Kotlin). It enforced non nullability in the type system which can be used in the StorageEnum functions.

calbach: Not asking for a null check, just want the new value to be included here: https://github.com/all-of-us/workbench/blob/b42c9d552f391a9127a7813fbddbf25e8be026e8/api/src/main/java/org/pmiops/workbench/db/model/StorageEnums.java#L188

(and test case added to ensure that conversion function is exhaustive)

ericsong: I know. I'm adding the new value. I'm just saying its hard to write a test case that actually makes it fail since the null behavior is identical to the INACTIVE behavior.

calbach: Ah sorry - RE test: to clarify, I was just referring to the missing test case at this level: https://github.com/all-of-us/workbench/blob/b42c9d552f391a9127a7813fbddbf25e8be026e8/api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java#L24

Though agree would be nice if it were caught at the service level as well.

ericsong: I rewrote the storage enums test to use reflection and auto check all fields. I think it covers everything the old test did.

https://github.com/all-of-us/workbench/pull/2432/commits/1bb93c7125da4be5f6f8b4641b0ca64327c0f30b#diff-dcf82ff88eec81c83dbafd1d56ec62fbR13

ericsong: Example of a test failure

```
name              : org.pmiops.workbench.model.AnnotationType:DATE
expected not to be: null

	at org.pmiops.workbench.db.model.StorageEnumsTest.noMissingMapEntries(StorageEnumsTest.java:37)
```

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jaycarlton: Thanks CH. So is Unsupported Access level still what we want?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
jaycarlton: Maybe add a TODO to remind us to do that

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -107,6 +107,24 @@ public Workspace get(String ns, String firecloudName) {
...
   }
 
+  @Override
+  public List<WorkspaceResponse> getWorkspaces() {
+    return getWorkspacesAndPublicWorkspaces().stream()
+        .filter(
+            workspaceResponse ->
+                !(workspaceResponse.getAccessLevel() != WorkspaceAccessLevel.OWNER
```
jaycarlton: Is there an easier way to read this if you factor the outermost `!` through the expression?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -130,22 +148,32 @@ public Workspace get(String ns, String firecloudName) {
...
 
-  @Override
-  public List<WorkspaceResponse> getPublishedWorkspaces() {
-    return getWorkspacesAndPublicWorkspaces().stream()
-        .filter(workspaceResponse -> workspaceResponse.getWorkspace().getPublished())
-        .collect(Collectors.toList());
+    org.pmiops.workbench.firecloud.model.WorkspaceResponse fcResponse;
+    org.pmiops.workbench.firecloud.model.Workspace fcWorkspace;
+
+    WorkspaceResponse response = new WorkspaceResponse();
```
jaycarlton: nit: I would almost want to see the fully qualified name for this `WorkspaceResponse` here as well to make it explicit that there are two of them now. IntelliJ might fight you on that though.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
jaycarlton: Oh right because you rethrow. Good deal.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
ericsong: sounds good to me

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: Added to the ticket

PR #2434, opened 2019-08-05
[RW-2459][Risk=no] Srubenst/copy concept sets

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
@@ -98,7 +98,9 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
                 throw new BadRequestException(
                     "Domain " + conceptSet.getDomain() + " is not allowed for concept sets");
               }
-              dbConceptSet.setVersion(Etags.toVersion(conceptSet.getEtag()));
+              if (conceptSet.getEtag() != null) {
+                dbConceptSet.setVersion(Etags.toVersion(conceptSet.getEtag()));
+              }
               dbConceptSet.setDescription(conceptSet.getDescription());
               dbConceptSet.setName(conceptSet.getName());
               return dbConceptSet;
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
@@ -15,6 +15,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
 import javax.persistence.OptimisticLockException;
+import org.pmiops.workbench.cdr.CdrVersionContext;
 import org.pmiops.workbench.cdr.ConceptBigQueryService;
 import org.pmiops.workbench.cdr.dao.ConceptDao;
 import org.pmiops.workbench.db.dao.ConceptSetDao;
@@ -97,6 +98,7 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
                 throw new BadRequestException(
                     "Domain " + conceptSet.getDomain() + " is not allowed for concept sets");
               }
+              dbConceptSet.setVersion(Etags.toVersion(conceptSet.getEtag()));
               dbConceptSet.setDescription(conceptSet.getDescription());
               dbConceptSet.setName(conceptSet.getName());
               return dbConceptSet;
@@ -373,14 +375,15 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
         workspaceService.get(
             copyRequest.getToWorkspaceNamespace(), copyRequest.getToWorkspaceName());
     Workspace fromWorkspace = workspaceService.get(fromWorkspaceNamespace, fromWorkspaceId);
-    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
+    workspaceService.enforceWorkspaceAccessLevel(
         toWorkspace.getWorkspaceNamespace(),
         toWorkspace.getFirecloudName(),
         WorkspaceAccessLevel.WRITER);
+    CdrVersionContext.setCdrVersionNoCheckAuthDomain(toWorkspace.getCdrVersion());
     if (toWorkspace.getCdrVersion().getCdrVersionId()
         != fromWorkspace.getCdrVersion().getCdrVersionId()) {
       throw new BadRequestException(
-          "New workspace does not have the same CDR version as current workspace");
+          "Target workspace does not have the same CDR version as current workspace");
     }
     org.pmiops.workbench.db.model.ConceptSet conceptSet =
         conceptSetDao.findOne(Long.valueOf(fromConceptSetId));
@@ -388,8 +391,7 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
       throw new NotFoundException(
           String.format(
               "Concept set %s does not exist",
-              generateConceptSetPathFromWorkspaceAndIdentifier(
-                  fromWorkspaceNamespace, fromWorkspaceId, fromConceptSetId)));
+              createResourcePath(fromWorkspaceNamespace, fromWorkspaceId, fromConceptSetId)));
     }
     org.pmiops.workbench.db.model.ConceptSet newConceptSet =
         new org.pmiops.workbench.db.model.ConceptSet(conceptSet);
@@ -403,24 +405,24 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
 
     try {
       newConceptSet = conceptSetDao.save(newConceptSet);
-      userRecentResourceService.updateConceptSetEntry(
-          toWorkspace.getWorkspaceId(),
-          userProvider.get().getUserId(),
-          newConceptSet.getConceptSetId(),
-          now);
     } catch (DataIntegrityViolationException e) {
       throw new ConflictException(
           String.format(
               "Concept set %s already exists.",
-              generateConceptSetPathFromWorkspaceAndIdentifier(
+              createResourcePath(
                   toWorkspace.getWorkspaceNamespace(),
                   toWorkspace.getFirecloudName(),
                   newConceptSet.getName())));
     }
+    userRecentResourceService.updateConceptSetEntry(
+        toWorkspace.getWorkspaceId(),
+        userProvider.get().getUserId(),
+        newConceptSet.getConceptSetId(),
+        now);
     return ResponseEntity.ok(toClientConceptSet(newConceptSet));
   }
 
-  private String generateConceptSetPathFromWorkspaceAndIdentifier(
+  private String createResourcePath(
       String workspaceNamespace, String workspaceFirecloudName, String identifier) {
     return String.format("\"/%s/%s/%s\"", workspaceNamespace, workspaceFirecloudName, identifier);
   }
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
@@ -385,7 +385,11 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
     org.pmiops.workbench.db.model.ConceptSet conceptSet =
         conceptSetDao.findOne(Long.valueOf(fromConceptSetId));
     if (conceptSet == null) {
-      throw new NotFoundException(String.format("Concept set %s does not exist", generateConceptSetPathFromWorkspaceAndIdentifier(fromWorkspaceNamespace, fromWorkspaceId, fromConceptSetId)));
+      throw new NotFoundException(
+          String.format(
+              "Concept set %s does not exist",
+              generateConceptSetPathFromWorkspaceAndIdentifier(
+                  fromWorkspaceNamespace, fromWorkspaceId, fromConceptSetId)));
     }
     org.pmiops.workbench.db.model.ConceptSet newConceptSet =
         new org.pmiops.workbench.db.model.ConceptSet(conceptSet);
@@ -416,10 +420,9 @@ public class ConceptSetsController implements ConceptSetsApiDelegate {
     return ResponseEntity.ok(toClientConceptSet(newConceptSet));
   }
 
-  private String generateConceptSetPathFromWorkspaceAndIdentifier(String workspaceNamespace, String workspaceFirecloudName, String identifier) {
-    return String.format("\"/%s/%s/%s\"", workspaceNamespace,
-        workspaceFirecloudName,
-        identifier);
+  private String generateConceptSetPathFromWorkspaceAndIdentifier(
+      String workspaceNamespace, String workspaceFirecloudName, String identifier) {
+    return String.format("\"/%s/%s/%s\"", workspaceNamespace, workspaceFirecloudName, identifier);
   }
 
   private org.pmiops.workbench.db.model.ConceptSet getDbConceptSet(
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
@@ -9,6 +9,7 @@ import com.google.common.collect.Streams;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Function;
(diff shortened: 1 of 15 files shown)


api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
     return ResponseEntity.ok(toClientConceptSet(dbConceptSet));
   }
 
+  @Override
+  public ResponseEntity<ConceptSet> copyConceptSet(
+      String fromWorkspaceNamespace,
+      String fromWorkspaceId,
+      String fromConceptSetId,
+      CopyRequest copyRequest) {
```
jaycarlton: Could the `CopyRequest` object contain the origin namespace, workspace and concept set?

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    workspaceService.enforceWorkspaceAccessLevel(
+        fromWorkspaceNamespace, fromWorkspaceId, WorkspaceAccessLevel.READER);
+    Workspace toWorkspace =
+        workspaceService.get(
+            copyRequest.getToWorkspaceNamespace(), copyRequest.getToWorkspaceName());
```
jaycarlton: aside: it might make sense to have a little class like `QualifiedWorkspaceId` to contain both the namespace and name.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
+        toWorkspace.getWorkspaceNamespace(),
+        toWorkspace.getFirecloudName(),
+        WorkspaceAccessLevel.WRITER);
+    if (toWorkspace.getCdrVersion().getCdrVersionId()
+        != fromWorkspace.getCdrVersion().getCdrVersionId()) {
+      throw new BadRequestException(
+          "New workspace does not have the same CDR version as current workspace");
+    }
+    org.pmiops.workbench.db.model.ConceptSet conceptSet =
```
jaycarlton: what if it can't be found? Does it throw? It's possible the concept set could have been deleted or had its access level changed just before we get here.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+        conceptSetDao.findOne(Long.valueOf(fromConceptSetId));
+    org.pmiops.workbench.db.model.ConceptSet newConceptSet =
+        new org.pmiops.workbench.db.model.ConceptSet();
+    newConceptSet.setDomainEnum(conceptSet.getDomainEnum());
+    newConceptSet.setSurveysEnum(conceptSet.getSurveysEnum());
+
+    newConceptSet.setDescription(conceptSet.getDescription());
```
jaycarlton: Is there anywhere in the model where we need to keep up with the fact that this concept set was copied from the old one at this timestamp? Either for auditing internally or UX purposes? Any requirement to have a prefix for the title like "Copy of Concept Set 3"?

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+    newConceptSet.setName(copyRequest.getNewName());
+    newConceptSet.setCreator(userProvider.get());
+    newConceptSet.setWorkspaceId(toWorkspace.getWorkspaceId());
+    newConceptSet.setCreationTime(now);
+    newConceptSet.setLastModifiedTime(now);
+    newConceptSet.setVersion(1);
```
jaycarlton: Please make a constant like `INITIAL_VERSION`.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+        toWorkspace.getFirecloudName(),
+        WorkspaceAccessLevel.WRITER);
+    if (toWorkspace.getCdrVersion().getCdrVersionId()
+        != fromWorkspace.getCdrVersion().getCdrVersionId()) {
+      throw new BadRequestException(
+          "New workspace does not have the same CDR version as current workspace");
+    }
+    org.pmiops.workbench.db.model.ConceptSet conceptSet =
+        conceptSetDao.findOne(Long.valueOf(fromConceptSetId));
+    org.pmiops.workbench.db.model.ConceptSet newConceptSet =
```
jaycarlton: I'd probably either give `ConceptSet` a copy constructor, or otherwise bust out a factory method that takes one in. Calling that form here will shorten this function and make that logic separately testable.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
...
+      newConceptSet = conceptSetDao.save(newConceptSet);
+      userRecentResourceService.updateConceptSetEntry(
+          toWorkspace.getWorkspaceId(),
+          userProvider.get().getUserId(),
+          newConceptSet.getConceptSetId(),
+          now);
+    } catch (DataIntegrityViolationException e) {
+      throw new BadRequestException(
+          String.format(
+              "Concept set \"/%s/%s/%s\" already exists.",
```
jaycarlton: It might help keep messaging consistent to make a method to print the path to the concept set given a workspace concept set name.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -180,14 +180,16 @@ public JSONObject getNotebookContents(String bucketName, String notebookName) {
   }
 
   private String withNotebookExtension(String notebookName) {
-    return notebookName.endsWith(".ipynb") ? notebookName : notebookName.concat(".ipynb");
+    return notebookName.endsWith(NOTEBOOK_EXTENSION)
```
jaycarlton: I'd make a new method somewhere to handle the file path and extension business logic here and at line 192.

api/src/main/resources/workbench.yaml
```diff
@@ -1256,7 +1256,7 @@ paths:
           name: copyNotebookRequest
           required: true
           schema:
-            $ref: '#/definitions/CopyNotebookRequest'
+            $ref: '#/definitions/CopyRequest'
```
jaycarlton: should the `name` attribute above at line 1256 be changed as well?

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,63 @@ private void addConceptsToSet(
     return ResponseEntity.ok(toClientConceptSet(dbConceptSet));
   }
 
+  @Override
+  public ResponseEntity<ConceptSet> copyConceptSet(
```
jaycarlton: Just thinking out loud, but it might be possible to have common code for copying objects (of all types) and have that call object-type-specific code. That might let you consolidate implementation and testing for several different copy APIs on the different controllers. I.e. if they all use the same namespace rules and name conventions, then just having an object type parameter might let you share a lot of code.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
```diff
@@ -648,17 +649,17 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
-              fromNotebookName,
-              copyNotebookRequest.getToWorkspaceNamespace(),
-              copyNotebookRequest.getToWorkspaceName(),
-              copyNotebookRequest.getNewName());
+              fromNotebookName.concat(NOTEBOOK_EXTENSION),
+              copyRequest.getToWorkspaceNamespace(),
+              copyRequest.getToWorkspaceName(),
+              copyRequest.getNewName().concat(NOTEBOOK_EXTENSION));
     } catch (BlobAlreadyExistsException e) {
       throw new BadRequestException("File already exists at copy destination");
```
jaycarlton: Instead of a 400, we could do a 409 here, since there's a conflict with an existing resource.

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -2056,8 +2056,8 @@ public void copyNotebook() {
     String newNotebookName = "new";
     String expectedNotebookName = newNotebookName + ".ipynb";
```
jaycarlton: use `NotebooksService.FILE_EXTENTION` here and below

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: I think so, but that may be a larger refactor than is in the scope of this PR.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: I believe this would throw an error. I will verify.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: I'll build out a copy constructor I think.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: None in the requirements specified. The name is user inputted, so we can't have a prefix of the title. (Since it could be arbitrary values and does not necessarily have the same name.)

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: Will do.

api/src/main/java/org/pmiops/workbench/api/WorkspacesController.java
blrubenstein: 👍

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
blrubenstein: Will do.

api/src/main/resources/workbench.yaml
blrubenstein: Unsure. I actually am happy with it as is, because it is still a request to copyNotebook? Do you have a preference?

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: It will return null. I am handling now.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
blrubenstein: That sounds reasonable to me. Might be a good cleanup down the line. I think there's some different requirements that could require design though. (If I copy a data set, does it copy all underlying resources? How about if I copy a cohort review? I think there's some product stuff to figure out, but I worry about fitting too tightly to our model.)

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+              "Concept set %s already exists.",
+              generateConceptSetPathFromWorkspaceAndIdentifier(
+                  toWorkspace.getWorkspaceNamespace(),
+                  toWorkspace.getFirecloudName(),
+                  newConceptSet.getName())));
+    }
+    return ResponseEntity.ok(toClientConceptSet(newConceptSet));
+  }
+
+  private String generateConceptSetPathFromWorkspaceAndIdentifier(
```
ericsong: i think we can generalize this function name to something like `createResourcePath` since the actual implementation isn't tied to concept sets

ui/src/app/components/copy-modal.spec.tsx
```diff
@@ -2,20 +2,22 @@ import {mount} from 'enzyme';
...
+import {Workspace, WorkspaceAccessLevel, WorkspacesApi} from 'generated/fetch';
 import {WorkspacesApiStub} from 'testing/stubs/workspaces-api-stub';
 import {waitOneTickAndUpdate} from 'testing/react-test-helpers';
 
-import { CopyNotebookModal, CopyNotebookModalProps, CopyNotebookModalState } from './copy-notebook-modal';
-import { TextInput } from 'app/components/inputs';
+import {CopyModal, CopyModalProps, CopyModalState} from './copy-modal';
+import {dropNotebookFileSuffix} from 'app/pages/analysis/util';
 
 describe('CopyNotebookModal', () => {
```
ericsong: `CopyModal`

api/src/main/resources/workbench.yaml
```diff
@@ -1704,6 +1704,31 @@ paths:
...
+      - $ref: '#/parameters/workspaceNamespace'
+      - $ref: '#/parameters/workspaceId'
+      - in: path
+        name: conceptSetId
+        required: true
+        type: string
+    post:
+      tags:
+        - conceptSets
+      description: Copy specified notebook in path to specified workspace in body
```
ericsong: `notebook` -> `concept set`

ui/src/app/components/copy-modal.tsx
```diff
@@ -4,22 +4,25 @@ import { Button } from 'app/components/buttons';
...
 export interface Props {
+  destinationTab: string;
   fromWorkspaceNamespace: string;
   fromWorkspaceName: string;
-  fromNotebook: FileDetail;
+  fromResourceName: string;
   onClose: Function;
   onCopy: Function;
+  resourceType: ResourceType;
+  saveFunction: (CopyRequest) => Promise<any>;
```
ericsong: can we type the `any`?

ericsong: also maybe `copy` or `copyFunction` might be a more appropriate name? I think I prefer dropping the `Function` suffix if we're good about applying types.

ui/src/app/components/copy-modal.tsx
```diff
@@ -63,57 +66,58 @@ class CopyNotebookModal extends React.Component<Props, State> {
...
   }
 
   goToDestinationWorkspace() {
     navigate(
-      ['workspaces', this.state.destination.namespace, this.state.destination.id, 'notebooks']
+      [
+        'workspaces',
+        this.state.destination.namespace,
+        this.state.destination.id,
+        this.props.destinationTab
```
ericsong: can we generate this off of the `ResourceType` prop? If we can, we can reduce the number of props this component requires.

ui/src/app/components/copy-modal.tsx
```diff
@@ -156,40 +161,37 @@ class CopyNotebookModal extends React.Component<Props, State> {
...
   }
 }
 
 export {
-  CopyNotebookModal,
-  Props as CopyNotebookModalProps,
-  State as CopyNotebookModalState,
+  CopyModal,
+  Props as CopyModalProps,
+  State as CopyModalState,
```
ericsong: JW, is it possible to remove the state export? I think I added it before we decided thats probably bad practice.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -144,7 +146,7 @@ public void setUserProvider(Provider<User> userProvider) {
     dbConceptSet.setWorkspaceId(workspace.getWorkspaceId());
     dbConceptSet.setCreationTime(now);
     dbConceptSet.setLastModifiedTime(now);
-    dbConceptSet.setVersion(1);
+    dbConceptSet.setVersion(INITIAL_VERSION);
```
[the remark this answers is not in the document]
blrubenstein: I'm actually not sure. If we clone a concept set, do we want to maintain version history? 

Probably not, tbh? 1 should be a default.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+      String fromConceptSetId,
+      CopyRequest copyRequest) {
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    workspaceService.enforceWorkspaceAccessLevel(
+        fromWorkspaceNamespace, fromWorkspaceId, WorkspaceAccessLevel.READER);
+    Workspace toWorkspace =
+        workspaceService.get(
+            copyRequest.getToWorkspaceNamespace(), copyRequest.getToWorkspaceName());
+    Workspace fromWorkspace = workspaceService.get(fromWorkspaceNamespace, fromWorkspaceId);
+    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
```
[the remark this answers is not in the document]
blrubenstein: I believe if access is denied it throws an exception which we are not catching here. 

I can move to separate functions though.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+        toWorkspace.getWorkspaceNamespace(),
+        toWorkspace.getFirecloudName(),
+        WorkspaceAccessLevel.WRITER);
+    if (toWorkspace.getCdrVersion().getCdrVersionId()
+        != fromWorkspace.getCdrVersion().getCdrVersionId()) {
+      throw new BadRequestException(
+          "New workspace does not have the same CDR version as current workspace");
```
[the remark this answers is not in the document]
blrubenstein: Target sounds good. I'll make it that.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+    newConceptSet.setName(copyRequest.getNewName());
+    newConceptSet.setCreator(userProvider.get());
+    newConceptSet.setWorkspaceId(toWorkspace.getWorkspaceId());
+    newConceptSet.setCreationTime(now);
+    newConceptSet.setLastModifiedTime(now);
+    newConceptSet.setVersion(INITIAL_VERSION);
+
+    try {
+      newConceptSet = conceptSetDao.save(newConceptSet);
+      userRecentResourceService.updateConceptSetEntry(
```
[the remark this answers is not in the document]
blrubenstein: Sure! Will move out

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+          toWorkspace.getWorkspaceId(),
+          userProvider.get().getUserId(),
+          newConceptSet.getConceptSetId(),
+          now);
+    } catch (DataIntegrityViolationException e) {
+      throw new ConflictException(
+          String.format(
+              "Concept set %s already exists.",
```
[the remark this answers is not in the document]
blrubenstein: I would much rather us try to figure out what other error cases might exist, and handle all of them. This is currently (as far as I know) the only cause that we can see from this, because it is copying a valid concept set and changing the name only. Can you think of any other errors we might see?

ui/src/app/components/copy-modal.tsx
blrubenstein: We can't, because it could return any number of objects. This is explicitly intended to be generic. 

Happy to drop the suffix and name it copy though.

ui/src/app/components/copy-modal.tsx
blrubenstein: Yes, with a map. I can build that out.

ui/src/app/components/copy-modal.tsx
blrubenstein: Probably. I will check.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -358,6 +360,71 @@ private void addConceptsToSet(
...
+        fromWorkspaceNamespace, fromWorkspaceId, WorkspaceAccessLevel.READER);
+    Workspace toWorkspace =
+        workspaceService.get(
+            copyRequest.getToWorkspaceNamespace(), copyRequest.getToWorkspaceName());
+    Workspace fromWorkspace = workspaceService.get(fromWorkspaceNamespace, fromWorkspaceId);
+    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
+        toWorkspace.getWorkspaceNamespace(),
+        toWorkspace.getFirecloudName(),
+        WorkspaceAccessLevel.WRITER);
+    if (toWorkspace.getCdrVersion().getCdrVersionId()
```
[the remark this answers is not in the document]
blrubenstein: cdrVersionId is a long.

ui/src/app/components/copy-modal.tsx
blrubenstein: Nope, we need it for the tests.

api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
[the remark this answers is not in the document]
jaycarlton: That's a product decision, but cloning with history is another story if we want it I think. The only reason we'd need to maintain that would be backtracking bad data or changes but I'm not sure whether the concept set would even be the way that works.

ui/src/app/components/copy-modal.tsx
jaycarlton: I guess I'm confused as to how we've agreed on the shape of a `CopyRequest` but still need to leave the return type unspecified. Could there not be a `CopyResponse` that has an array of the copies created, or whatever? Just thinking out loud.

blrubenstein: I actually figured out how to do the typing, so we're actually good on that! :D

PR #2436, opened 2019-08-05
[RW-3188][risk=low] Add billing project user backfill script

calbach:
Backfill template stolen from https://github.com/all-of-us/workbench/pull/2221.

This script may need to be run periodically until we perform the 1PPW migration.

CC @gjuggler @jmthibault79

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/BackfillBillingProjectUsers.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/BackfillBillingProjectUsers.java
@@ -5,9 +5,7 @@ import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import java.io.IOException;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
 import org.apache.commons.cli.CommandLine;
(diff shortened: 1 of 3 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/BackfillBillingProjectUsers.java
```diff
@@ -0,0 +1,166 @@
...
+    dryLog(dryRun, String.format("added %d users as billing project users", userUpgrades));
+  }
+
+  @Bean
+  public CommandLineRunner run() {
+    return (args) -> {
+      CommandLine opts = new DefaultParser().parse(options, args);
+      backfill(
+          newWorkspacesApi(opts.getOptionValue(fcBaseUrlOpt.getLongOpt())),
+          newBillingApi(opts.getOptionValue(fcBaseUrlOpt.getLongOpt())),
```
ericsong: I think you can grab the fcBaseUrl through the WorkbenchConfigProvider if you want to get it through DI instead of command line args.

api/tools/src/main/java/org/pmiops/workbench/tools/BackfillBillingProjectUsers.java
```diff
@@ -0,0 +1,166 @@
...
+          // Only owners should be granted billing project user.
+          continue;
+        }
+        dryLog(
+            dryRun,
+            String.format(
+                "granting billing project user on '%s' to '%s' (%s)",
+                w.getNamespace(), user, entry.getAccessLevel()));
+        if (!dryRun) {
+          billingApi.addUserToBillingProject(w.getNamespace(), "user", user);
```
ericsong: Should we be handling error cases or does this tool have a high probability on passing on every run?

One possible error is calling this API for a user that no longer exists in FC/Google.

Also, I'm guessing the call is idempotent but it looks like we would be running the same api calls if the tool is rerun.

calbach: Good call - warning is better here, added.

This call should be idempotent. It may or may not have converged on rerun depending on the issue/ FC implementation.

PR #2438, opened 2019-08-06
[RW-2388][risk=no] Add cohort review resource card

dolbeew:
- Also adds new color to `colors.ts` for the review card labels (we used the maroon from the cohort graphs, can change if needed)

Screen Shot 2019-08-06 at 10 46 17 AM


ui/src/app/pages/data/data-page.tsx
```diff
@@ -66,6 +66,7 @@ enum Tabs {
   SHOWALL = 'SHOW ALL',
   DATASETS = 'DATA SETS',
   COHORTS = 'COHORTS',
+  COHORTREVIEWS = 'COHORTREVIEWS',
```
blrubenstein: This should probably have a space in the string.

dolbeew: Nice catch, done.

dolbeew:
I'm merging this as is and will wait on updating the review resource cards to match the notebooks. I like the new approach, but looks like reviews will need to be implemented differently since they're in the data tab mixed with with three other resource types, where as notebooks are the only resource in the analysis tab.

PR #2441, opened 2019-08-06
[RW-2211][risk=low] Adjusts semantics of workspace list filter.

bopwerks:

### DIFF
--- a/ui/src/app/pages/workspace/workspace-list.tsx
+++ b/ui/src/app/pages/workspace/workspace-list.tsx
@@ -359,8 +359,8 @@ export const WorkspaceList = withUserProfile()
     // Maps each "Filter by" dropdown element to a set of access levels to display.
     const filters = [
       { label: 'Owner',  value: ['OWNER'] },
-      { label: 'Writer', value: ['OWNER', 'WRITER'] },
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/workspace/workspace-list.tsx
```diff
@@ -359,8 +359,8 @@ export const WorkspaceList = withUserProfile()
     // Maps each "Filter by" dropdown element to a set of access levels to display.
     const filters = [
       { label: 'Owner',  value: ['OWNER'] },
-      { label: 'Writer', value: ['OWNER', 'WRITER'] },
-      { label: 'Reader', value: ['OWNER', 'READER'] },
+      { label: 'Writer', value: ['WRITER'] },
+      { label: 'Reader', value: ['READER'] },
```
calbach: Note: Just followed the thread on this when seeing this in prod - the resolution here looks correct to me. But wanted to clarify a point: owner, writer, and reader are concentric roles in our system. As in, a "Writer" can also "Read" (i.e. in the initial implementation, I would have expected `Reader -> ['OWNER', 'WRITER' 'READER']`).

PR #2442, opened 2019-08-06
fixes for VS Code editor

jaycarlton:
When editing/viewing the UI code in Visual Studio Code, I noticed that paths weren't resolving correctly. It looks like this is because we're using webpack implied roots (like `app`). Adding paths for `app` and `generated` to `.tsconfig` fixed that.

Additionally, I want to ignore the `.vscode` directory this editor creates.

An open question is whether this can go in a shared `.editorconfig`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .classpath
 .project
 .externalToolBuilders
+.vscode
 .settings
 cron-emulator.log
 /api/?/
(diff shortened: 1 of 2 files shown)


jaycarlton:
@blrubenstein do the UI tests that failed because of this not get run by Github's pre-merge hooks?

blrubenstein:
It looks like there was a failure in the checks above. I don't know about pre-merge though.

PR #2448, opened 2019-08-07
[RISK=LOW] Concept Add modal refactor : Use same modal for both concept set and survey

NehaBroad:


ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -96,13 +109,39 @@ export const ConceptAddModal = withCurrentWorkspace()
     }
   }
 
+  isSurvey() {
```
blrubenstein: Nit: Make this a get function? (so you can call as parameter rather than method)

ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -96,13 +109,39 @@ export const ConceptAddModal = withCurrentWorkspace()
...
+      name, selectedConceptsInDomain
+    } = this.state;
     this.setState({saving: true});
-    const conceptIds = fp.map(selected => selected.conceptId, selectedConceptsInDomain);
+    let conceptIds = fp.map(selected => selected.conceptId, selectedConceptsInDomain);
+    let survey = Surveys.THEBASICS;
+    if (this.isSurvey()) {
+      conceptIds = selectedSurvey.map((surveys) => surveys.conceptId);
+
+      switch (surveyName) {
```
blrubenstein: This seems like its pretty likely to be reused elsewhere. (And maybe is?) 

Can we either make this a map or a common util? (Probably utils for enum->displayname, and vice versa)

ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -119,26 +158,32 @@ export const ConceptAddModal = withCurrentWorkspace()
...
-      const conceptSet: ConceptSet = {
+      let conceptSet: ConceptSet = {
         name: name,
         description: newSetDescription,
         domain: selectedDomain.domain
       };
+      if (this.isSurvey()) {
+        conceptSet = {
+          ...conceptSet,
+          survey: survey
```
blrubenstein: Could you add this above and keep the const? (Just add to the object `survey: (this.isSurvey() ? survey : null)`?)

ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -148,11 +193,19 @@ export const ConceptAddModal = withCurrentWorkspace()
     }
   }
 
+  get Title() {
+    const {selectedDomain, selectedSurvey, surveyName} = this.props;
+    const {selectedConceptsInDomain} = this.state;
+
+    return !this.isSurvey() ? 'Add ' + selectedConceptsInDomain.length + ' Concepts to ' +
```
blrubenstein: There's a lot of shared language here. The only different bits seem to be `Concepts` vs `Survey` and the name vs the number. Can we just parameterize that and keep the shared language?

PR #2449, opened 2019-08-07
[RW-2853][risk=no] Removing old CB search code.

freemabd:
Removing old CB search code.


api/db-cdr/changelog-data/db.changelog-master.xml
```diff
@@ -12,20 +12,7 @@
   <include file="changelog-data/db.changelog-2.xml"/>
   <include file="changelog-data/db.changelog-3.xml"/>
   <include file="changelog-data/db.changelog-4.xml"/>
+  <include file="changelog-data/db.changelog-5.xml"/>
```
calbach: Could you add some context on this? Won't this make it very challenging to perform a SQL migration on older CDR versions databases?

And going further in the other direction - is there a point in keep 7 files instead of just collapsing into 1?

freemabd: So locally we only have one version of the cloud mysql cdr: https://github.com/all-of-us/workbench/blob/master/api/config/cdr_versions_local.json. So the master file located in db-cdr/changelog-data is only used to generate local data for local development. Example: `./project.rb run-local-data-migrations`. These migrations should never run in any other environment except local. Migrations for cloud mysql cdr for all environments are handled from db-cdr/changelog-schema.

You are correct that we should just collapse this into a single file. Added a comment(from above) to change log files and confirmed that generating local data works with consolidation into 1 file. Done.

calbach: Thanks for the comment.

PR #2450, opened 2019-08-07
[RW-3140][risk=low] Makes compliance data load async and only when th…

bopwerks:
…e user lacks registered access.

### DIFF
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -172,7 +172,7 @@ export const Homepage = withUserProfile()(class extends React.Component<
       });
     }).catch(err => {
       this.setState({trainingCompleted: false});
-      console.error('error fetching moodle training status');
+      console.error('error fetching moodle training status:', err);
     });
     const twoFactorAuthStatus = profileApi().syncTwoFactorAuthStatus().then(result => {
       this.setState({
@@ -180,7 +180,7 @@ export const Homepage = withUserProfile()(class extends React.Component<
       });
     }).catch(err => {
       this.setState({twoFactorAuthCompleted: false});
-      console.error('error fetching two factor auth status');
+      console.error('error fetching two factor auth status:', err);
     });
     return Promise.all([complianceStatus, twoFactorAuthStatus]);
   }
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -193,7 +193,6 @@ export const Homepage = withUserProfile()(class extends React.Component<
         reload();
       }, 10000);
     } else {
-
       if (!profile.betaAccessRequestTime) {
         profileApi().requestBetaAccess();
       }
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -228,6 +227,7 @@ export const Homepage = withUserProfile()(class extends React.Component<
               accessTasksLoaded: true
             });
           } else {
+            await this.syncCompliance();
```
calbach: This isn't in the right place - this `else` clause is effectively dead code; `enforceRegistered` is a server side flag which we are in the process of removing, and should always be `true` at this point. This should be running as a function of the profile state - possibly using a check similar to L215.

Also we'd probably want `syncCompliance()` to complete before setting `accessTasksLoaded`

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -165,6 +165,26 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
   }
 
+  async syncCompliance() {
+    const complianceStatus = profileApi().syncComplianceTrainingStatus().then(result => {
+      this.setState({
+        trainingCompleted: getRegistrationTasksMap()['complianceTraining'].isComplete(result)
+      });
+    }).catch(err => {
+      this.setState({trainingCompleted: false});
+      console.error('error fetching moodle training status');
```
jaycarlton: Do you want to pull the error message out of `err` (here and below)?

bopwerks: yes, good idea 👍

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -194,37 +213,20 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
-      } catch (ex) {
-        this.setState({twoFactorAuthCompleted: false});
-        console.error('error fetching two factor auth status');
-      }
-
       this.setState({betaAccessGranted: !!profile.betaAccessBypassTime});
 
       const {workbenchAccessTasks} = queryParamsStore.getValue();
+      const hasRegisteredAccess = hasRegisteredAccessFetch(profile.dataAccessLevel);
+      if (!hasRegisteredAccess || parseInt(workbenchAccessTasks, 2) === 1) {
```
calbach: nit: this check is inconsistent with the check 3 lines down (L223). `|| workbenchAccessTasks` seems fine to me, I don't think we've specifically defined a contract here - just generally "if any value is specified". If you feel very strongly about `parseInt`, please factor this out so the same logic is applied in both situations.

bopwerks: tl;dr I'll use `|| workbenchAccessTasks` to be consistent. When I wrote this I meant to avoid the case where `?workbenchAccessTasks=0` would have the same effect as `?workbenchAccessTasks=1` (since non-empty strings evaluate to `true`). But I think it's reasonable for the empty string to denote `false` and the non-empty string to denote `true` as you imply.

PR #2451, opened 2019-08-08
[RW-3216][Risk=low] Interpolate values in dataset notebook export

blrubenstein:


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -282,10 +282,10 @@ public DataSet saveDataSet(
```
gjuggler: Is there no unit test for this class? It would be nice to have a test somewhere which demonstrates the interpolation happening correctly.

(After looking around... looks like there is some testing at the controller level. It may be easiest to add some example of interpolation over there. It's worth noting that we're trying to move more of our unit testing onto services in order to leave controllers as simple as possible... not for this PR, but having no test file for this service class is a bit of a code smell in that regard.)

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -442,13 +444,19 @@ public void setUp() throws Exception {
...
-    when(bigQueryService.filterBigQueryConfig(any()))
-        .thenReturn(
-            QueryJobConfiguration.newBuilder(
-                    "SELECT * FROM person_id from `" + TEST_CDR_TABLE + "` person")
-                .addNamedParameter(
-                    NAMED_PARAMETER_NAME, QueryParameterValue.string(NAMED_PARAMETER_VALUE))
-                .build());
+    // This is not great, but due to the interaction of mocks and bigquery, it is
+    // exceptionally hard to fix it so that it calls the real filterBitQueryConfig
+    // but _does not_ call the real methods in the rest of the bigQueryService
```
gjuggler: Is it not possible to do

when(bigQueryService.filterBigQueryConfig(any())).thenCallRealMethod()

?

blrubenstein: Talked offline. Tried doing that. I suspect it has to do with `MockBean` and the way that spring and mockito interact in our tests.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Yeah, this is something Eric has brought up a few times, that we tend to handle our testing at the controller level only. (There's many other cases where we do a bunch of service level tests from the controller tests.)

Might be worth looking into a refactor there to break our tests apart more.

PR #2457, opened 2019-08-09
[RW-2808][risk=no] Add route in review for cohort description

dolbeew:
- Moves cohort description page to a separate route to help with Google Analytics.

### DIFF
--- a/ui/src/app/cohort-review/page-layout/page-layout.ts
+++ b/ui/src/app/cohort-review/page-layout/page-layout.ts
@@ -1,4 +1,5 @@
 import {Component, OnDestroy, OnInit} from '@angular/core';
+import {Router} from '@angular/router';
 import {
   cohortReviewStore,
   filterStateStore,
@@ -21,6 +22,8 @@ export class PageLayout implements OnInit, OnDestroy {
   cohortLoaded = false;
   readonly = false;
 
+  constructor(private router: Router) {}
+
   ngOnInit() {
     const {ns, wsid, cid} = urlParamsStore.getValue();
     const {accessLevel, cdrVersionId} = currentWorkspaceStore.getValue();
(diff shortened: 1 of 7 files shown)


ui/src/app/cohort-review/page-layout/page-layout.ts
```diff
@@ -20,6 +23,20 @@ export class PageLayout implements OnInit, OnDestroy {
   reviewPresent: boolean;
   cohortLoaded = false;
   readonly = false;
+  subscription: Subscription;
+
+  constructor(private location: Location, private router: Router) {
+    this.subscription = router.events.subscribe(event => {
+      if (event instanceof NavigationStart) {
+        const oldRoute = router.url.split('/').pop();
```
freemabd: Can we determine the same conditions by possibly setting some state instead of explicitly checking the url? Feels like url/routing could change and this would be broken.

ui/src/app/cohort-review/page-layout/page-layout.ts
```diff
@@ -34,7 +51,7 @@ export class PageLayout implements OnInit, OnDestroy {
     }).then(review => {
       cohortReviewStore.next(review);
       this.reviewPresent = review.reviewStatus !== ReviewStatus.NONE;
-      if (this.reviewPresent) {
+      if (this.reviewPresent && this.router.url.split('/').pop() === 'review') {
```
freemabd: I would ask same question as above. Can we use state to determine this or do we have to check url strings?

ui/src/app/cohort-review/page-layout/page-layout.ts
dolbeew: I could potentially track the active component in state, but I'd still need to compare the urls to determine if the browser's back button was clicked.

ui/src/app/cohort-review/page-layout/page-layout.ts
dolbeew: The reason I added this condition was to prevent switching back to the `participants` route when you refresh the page on the `participants/:pid` route (participant detail page), in which case the state gets reset anyway.

It's definitely not the cleanest solution, but the only way you can get the current route in Angular is using the url.

PR #2460, opened 2019-08-09
[RW-3174][RISK=NO]Update Data Tab - Language and Images

NehaBroad:
- Remove the data header

- Remove concept set tile

- Add images to Cohort and data set tile and change language

Screen Shot 2019-08-09 at 4 24 58 PM


NehaBroad:
Have added a tech debt ticket (RW-3230) for the same, so that it can be discussed more during our Eng sync meetings

PR #2463, opened 2019-08-12
[RW-3236][Risk=no] Fix data page images

blrubenstein:


ui/src/app/pages/data/data-page.tsx
```diff
@@ -195,7 +195,7 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
               <div style={styles.cardText}>
                 {descriptions.cohorts}
               </div>
-              <div style={{width: '100%', paddingTop: '1rem'}}>
+              <div style={{width: '100%', maxWidth: '425px', paddingTop: '1rem'}}>
```
calbach: Where did 425px come from, is it somehow related to the width of the image? Could you add a comment?

blrubenstein: It is related to the aspect ratio of the image, and the height of the box. (From what I could tell) Will add comment.

PR #2464, opened 2019-08-13
[RW-3096][risk=no] Add source/standard limitation language to CB wizard

dolbeew:
Also includes updates to use new colors file and minor fixes to physical measurements tree.

### DIFF
--- a/ui/src/app/cohort-search/list-search/list-search.component.tsx
+++ b/ui/src/app/cohort-search/list-search/list-search.component.tsx
@@ -334,7 +334,7 @@ export const ListSearch = withCurrentWorkspace()(
     }
 
     render() {
-      const {wizard: {domain}} = this.props;
+      const {wizard: {domain, isStandard}} = this.props;
       const {data, error, ingredients, loading, results, sourceMatch} = this.state;
       const listStyle = domain === DomainType.DRUG ? {...styles.listContainer, marginTop: '3.75rem'}
         : styles.listContainer;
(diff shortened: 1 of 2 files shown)


ui/src/app/cohort-search/list-search/list-search.component.tsx
```diff
@@ -16,15 +17,15 @@ const styles = reactStyles({
     position: 'absolute',
```
jaycarlton: Is all of this styling unique to this component? Or could some of it be inherited or pulled in from a common place?

ui/src/app/cohort-search/list-search/list-search.component.tsx
```diff
@@ -371,7 +372,16 @@ export const ListSearch = withCurrentWorkspace()(
...
               Return to source code
-          </Clickable>.
+            </Clickable>.
+          </div>}
+          {isStandard !== undefined && <div style={{...styles.error, margin: '-0.5rem 0 0.5rem'}}>
+            <div style={{float: 'left', height: '1.5rem'}}>
+              <ClrIcon style={{margin: '0 0.5rem 0 0.25rem'}} className='is-solid'
+                shape='exclamation-triangle' size='22'/>
+            </div>
+            You have added a {isStandard ? 'standard' : 'source'} code. To add a
```
jaycarlton: this calculation is repeated. Maybe do `const codeType = isStandard ? 'standard' : 'source'` above and use that here and below.

ui/src/app/cohort-search/node-info/node-info.component.ts
```diff
@@ -73,7 +73,7 @@ export class NodeInfoComponent implements OnInit, OnDestroy, AfterViewInit {
 
   get paramId() {
     return `param${this.node.conceptId && !this.isPPI ?
```
jaycarlton: nit: can we spell out PPI instead of abbreviating it?

ui/src/app/cohort-search/list-search/list-search.component.tsx
dolbeew: Yes, could use a refactor on several components to separate out common styles. Will try to do a follow-up pr with that.

ui/src/app/cohort-search/list-search/list-search.component.tsx
dolbeew: It looks the same but actually switches strings in the second instance.
ln 382: `isStandard ? ' standard' : 'source'` 
ln 383: `isStandard ? ' source' : ' standard'`

ui/src/app/cohort-search/node-info/node-info.component.ts
dolbeew: Good idea. Actually, the way we handle domains has changed, so I'm renaming to `isSurvey` to better reflect what we're checking here.

PR #2465, opened 2019-08-13
[RW-3142][Risk=no] Disable playground fully when read-only

blrubenstein:

### DIFF
--- a/ui/src/app/icons/playground-mode-icon.tsx
+++ b/ui/src/app/icons/playground-mode-icon.tsx
@@ -1,30 +1,85 @@
 import * as React from 'react';
 
+import colors, {colorWithWhiteness} from 'app/styles/colors';
 
-export const PlaygroundModeIcon = () => <svg version='1.1'
-                                             id='Layer_1'
-                                             viewBox='0 0 14.1 12.3'>
-  <title>Group</title>
-  <desc>Created with Sketch.</desc>
-  <g id='ANAYLSIS'>
-    <g id='Notebook-PREVIEW' transform='translate(-363.000000, -85.000000)'>
-      <g id='Group-5' transform='translate(363.000000, 85.000000)'>
-        <g id='Group-3' transform='translate(0.889808, 0.297292)'>
-          <g id='Group' transform='translate(0.047675, 0.625027)'>
-            <path fill='#216FB4' d={`M11.8-0.9c0.7,0,1.3,0.6,1.3,1.3V10c0,0.7-0.6,1.3-1.3,1.3H0.4c
-            -0.7,0-1.3-0.6-1.3-1.3V0.4c0-0.7,0.6-1.3,1.3-1.3C0.4-0.9,11.8-0.9,11.8-0.9z`}/>
+export interface PlaygroundModeIconProps {
+  disabled: boolean;
+  style: object;
+  enableHoverEffect: boolean;
+}
+
+export interface PlaygroundModeIconState {
+  style: object;
+}
+
+const defaultStyle = {
+  height: 19,
+  width: 19,
+  marginLeft: '.5rem',
+  fill: colors.accent,
+  cursor: 'pointer'
+};
+
+const hoverStyle = {...defaultStyle, fill: colorWithWhiteness(colors.accent, 0.2)};
+
+const disabledStyle = {cursor: 'not-allowed',
+  fill: colorWithWhiteness(colors.dark, 0.6)};
+
+
+export class PlaygroundModeIcon extends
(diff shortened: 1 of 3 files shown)


blrubenstein:
This also changes icon color to gray when disabled, and fixes a style bug in our icons where disabled icons would override component-level set sizes

gjuggler:
Can you include a screenshot on the styling changes? I'm reading the code now, but it will be v. helpful to see the changes visually.

ui/src/app/icons/playground-mode-icon.tsx
```diff
@@ -1,30 +1,85 @@
 import * as React from 'react';
 
+import colors, {colorWithWhiteness} from 'app/styles/colors';
 
-export const PlaygroundModeIcon = () => <svg version='1.1'
```
gjuggler: Wow, I didn't know we're embedding SVG within some of our code. Is there a way to avoid doing this? It feels so wrong... :)

ui/src/app/icons/playground-mode-icon.tsx
```diff
@@ -1,30 +1,85 @@
...
-  <g id='ANAYLSIS'>
-    <g id='Notebook-PREVIEW' transform='translate(-363.000000, -85.000000)'>
-      <g id='Group-5' transform='translate(363.000000, 85.000000)'>
-        <g id='Group-3' transform='translate(0.889808, 0.297292)'>
-          <g id='Group' transform='translate(0.047675, 0.625027)'>
-            <path fill='#216FB4' d={`M11.8-0.9c0.7,0,1.3,0.6,1.3,1.3V10c0,0.7-0.6,1.3-1.3,1.3H0.4c
-            -0.7,0-1.3-0.6-1.3-1.3V0.4c0-0.7,0.6-1.3,1.3-1.3C0.4-0.9,11.8-0.9,11.8-0.9z`}/>
+export interface PlaygroundModeIconProps {
+  disabled: boolean;
+  style: object;
```
gjuggler: nitpick: I think you can add a React.CSSProperties typedef to this, like we do here https://github.com/all-of-us/workbench/blob/322e08140e18415d2cc7115fe919c922551f0e4b/ui/src/app/components/spinners.tsx#L41

ui/src/app/icons/playground-mode-icon.tsx
```diff
@@ -1,30 +1,85 @@
...
+  enableHoverEffect: boolean;
+}
+
+export interface PlaygroundModeIconState {
+  style: object;
```
gjuggler: Also here (add tighter type definition)

ui/src/app/icons/playground-mode-icon.tsx
blrubenstein: Maybe. I was noticing that it wouldn't override fill values without this, which is why I factored the playground icon this way. Not super happy with it though. I'm curious if Terra does something better.

blrubenstein:
Screen Shot 2019-08-13 at 10 44 59 AM

jaycarlton:
Visually, typically when you show a disabled menu option, a user thinks they can do something to re-enable it. If that's not possible in the current situation (i.e. the user would need to go request access or log out and back in), then it might be less confusing to hide the disabled options altogether.

ui/src/app/pages/analysis/interactive-notebook.tsx
```diff
@@ -245,19 +248,18 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
               <div style={{display: 'flex'}}>
                 <div style={
                   Object.assign({}, styles.navBarItem,
-                    this.canWrite() ? styles.clickable : styles.disabled)}
+                    this.canWrite ? styles.clickable : styles.disabled)}
```
jaycarlton: this bit of logic can also be a property like `this.theButtonStyle` so you don't have to repeat it on line 259.

ui/src/app/pages/analysis/interactive-notebook.tsx
```diff
@@ -187,7 +190,7 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
       this.navigateOldNotebooksPage(false);
     }
 
-    private canWrite() {
+    private get canWrite() {
```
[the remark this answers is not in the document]
blrubenstein: Writer access/write permission is what it is called throughout our codebase, so leaving for consistency.

ui/src/app/pages/analysis/interactive-notebook.tsx
blrubenstein: Will add.

PR #2471, opened 2019-08-14
[RW-3201][Risk=low] Update dockerfile for api build

blrubenstein:


blrubenstein:
Built from: https://github.com/pires/netty-tcnative-alpine/blob/master/Dockerfile
Which builds from: https://github.com/jeanblanchard/docker-java/blob/master/jdk/Dockerfile
Which builds from: https://github.com/jeanblanchard/docker-alpine-glibc/blob/master/Dockerfile

PR #2472, opened 2019-08-15
[RW-3146] [risk=no] Create analytics module for event tracking

gjuggler:
This PR creates a somewhat-trivial analytics module to house centrally-defined event tracking code.

I would imagine this module may be extended over time, e.g. to hold any more opinionated analytics-related functions (e.g. Ajax event timing, page-view tracking, etc.).

I also included a couple examples of usage from the workspace-list component. I ran this locally and verified with the Analytics debugger extension that the events were getting fired:

Screen Shot 2019-08-14 at 8 57 24 PM

If we wanted to be more "type-safe" with our analytics categories, we could consider creating a string-valued enum to store those within the analytics module. Possibly useful for follow-up work as we start to scale up the event reporting throughout our codebase.

### DIFF
--- a/ui/src/app/utils/analytics.ts
+++ b/ui/src/app/utils/analytics.ts
@@ -1,0 +1,22 @@
+declare let gtag: Function;
+
+/**
+ * Triggers a Google Analytics event.
+ *
+ * For guidance on how to use action names, categories, and labels effectively,
+ * see https://support.google.com/analytics/answer/1033068#Anatomy.
+ */
+export function triggerEvent(
+  category: string, action: string, label?: string, value?: number,
+  nonInteraction: Boolean = false) {
+  if (window['gtag']) {
+    gtag('event', action, {
+      'event_category': category,
+      'event_label': label,
+      'event_value': value,
+      'non_interaction': nonInteraction
+    });
+  } else {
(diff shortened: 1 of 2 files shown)


ui/src/app/utils/analytics.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Triggers a Google Analytics event.
+ *
+ * For guidance on how to use action names, categories, and labels effectively,
+ * see https://support.google.com/analytics/answer/1033068#Anatomy.
+ */
+export function triggerEvent(
+  category: string, action: string, label?: string, value?: number,
+  nonInteraction: Boolean = false) {
```
dolbeew: Just curious, for what kind of event would you set `nonInteraction` flag to `true`?

gjuggler: Good question – I'd thought it might be useful for some cases, but after re-reading the docs, the only functional impact it would have is on the bounce rate of certain pages (a pageview with a non-interaction event can be considered a bounce, but a pageview with an interaction event will not be considered a bounce). So I've removed it, since it won't likely be needed in the immediate future.

dolbeew: Makes sense. Thanks!

PR #2476, opened 2019-08-16
[no ticket, no risk] Fixes run-local-all-migrations

jmthibault79:


jmthibault79:
/bin/sh is no longer a real bash, after the Dockerfile update.  This script needs a real bash.

PR #2481, opened 2019-08-19
[No ticket] [risk=no] Set default jsx compiler option in tsconfig

als364:
When loading UI as a project in IntelliJ, the compiler threw a bunch of warnings that suggested something was wonky in the Angular configuration (i.e. `Cannot find module` for all modules, `--jsx is not set`). This change fixed all of the warnings.

### DIFF
--- a/ui/tsconfig.json
+++ b/ui/tsconfig.json
@@ -20,6 +20,7 @@
       "dom"
     ],
     "module": "es2015",
-    "baseUrl": "./"
+    "baseUrl": "./",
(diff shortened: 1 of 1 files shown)


jaycarlton:
Were you able to run all of the ui tests with this change?

als364:
I had always been able to run UI tests even without this change. IntelliJ just complained.

als364:
Local UI test run passes.

jaycarlton:
Great, ship it.

PR #2483, opened 2019-08-19
[risk=no] Drop "AoU" from "All Participants" cohort

calbach:


calbach:
Great catch - updated one test.

PR #2485, opened 2019-08-20
[risk=no][RW-3161]New workspace projects are created in the Service Perimeter for Staging, Stable, and Prod

jmthibault79:

### DIFF
--- a/api/config/config_prod.json
+++ b/api/config/config_prod.json
@@ -12,7 +12,7 @@
     "billingRetryCount": 4,
     "billingProjectBufferCapacity": 20,
     "xAppIdValue": "AoU-RW",
-    "vpcServicePerimeterName": ""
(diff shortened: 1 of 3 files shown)


api/config/config_prod.json
```diff
@@ -12,7 +12,7 @@
     "billingRetryCount": 4,
     "billingProjectBufferCapacity": 20,
     "xAppIdValue": "AoU-RW",
-    "vpcServicePerimeterName": ""
+    "vpcServicePerimeterName": "accessPolicies/357236995176/servicePerimeters/terra_prod_aou_prod"
```
jaycarlton: I guess we've already settled on this scheme, but I would've probably just had this field hold the name of the perimeter, e.g. `terra_prod_aou_prod` and filled in the policy value programmatically.

PR #2487, opened 2019-08-20
[RW-3029][risk=no] Actually open exported dataset notebooks in a new tab

calbach:
`window.open()` and `window.history.back()` calls previously created a race condition.

### DIFF
--- a/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
+++ b/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
@@ -129,7 +129,7 @@ class ExportDataSetModal extends React.Component<
     // Open notebook in a new tab and close the modal
     const notebookUrl = '/workspaces/' + workspaceNamespace + '/' + workspaceFirecloudName +
         '/notebooks/' + appendNotebookFileSuffix(encodeURIComponent(this.state.notebookName));
-    window.open(notebookUrl);
+    window.open(notebookUrl, '_blank');
     this.props.closeFunction();
   }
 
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -129,7 +129,7 @@ class ExportDataSetModal extends React.Component<
     // Open notebook in a new tab and close the modal
     const notebookUrl = '/workspaces/' + workspaceNamespace + '/' + workspaceFirecloudName +
         '/notebooks/' + appendNotebookFileSuffix(encodeURIComponent(this.state.notebookName));
-    window.open(notebookUrl);
+    window.open(notebookUrl, '_blank');
```
jaycarlton: Would it make sense to pull this and possibly related functions into a utility class for window/navigation management? Also, defining that string constant someplace would help.

ui/src/app/pages/data/data-set/new-dataset-modal.tsx
```diff
@@ -162,7 +162,7 @@ class NewDataSetModal extends React.Component<Props, State> {
         // Open notebook in a new tab and return back to the Data tab
         const notebookUrl = '/workspaces/' + workspaceNamespace + '/' + workspaceId +
```
jaycarlton: nit (and out of scope): It would be helpful to have a utility function to construct a notebook URL from the things that aren't const in it.

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
blrubenstein: Since this is a JS/web standard thing, as opposed to something in our code, I don't know if I would pull it out into a utility class/bother wrapping it in our own method. I don't have a strong feeling though, could go either way.

jaycarlton:
Fair enough. I guess I had also kinda expected React to encapsulate the
window object, but it sounds like it doesn't work that way. Feel free to
keep it the way it is.

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
calbach: Yes, it is a standard built into the Javascript browser API. Wrapping it will only result in misdirection/obfuscation, and would not be portable to any other code project.

PR #2488, opened 2019-08-21
[RW-3229] [risk=low] Enable copy notebook / concept set from readonly workspace

als364:
One can now copy notebooks and concept sets from a read-only workspace to another workspace without cloning the entire read-only workspace.

### DIFF
--- a/ui/src/app/components/resource-card.tsx
+++ b/ui/src/app/components/resource-card.tsx
@@ -162,7 +162,7 @@ export class ResourceCard extends React.Component<Props, State> {
   }
 
   get actionsDisabled(): boolean {
-    return !this.writePermission && !this.isConceptSet
+    return !this.writePermission && !this.isConceptSet;
   }
 
   get writePermission(): boolean {
--- a/ui/src/app/components/resource-card.tsx
+++ b/ui/src/app/components/resource-card.tsx
@@ -161,16 +161,12 @@ export class ResourceCard extends React.Component<Props, State> {
     return this.resourceType === ResourceType.DATA_SET;
   }
 
-  get actionsDisabled(): boolean {
-    return !this.writePermission && !this.isConceptSet;
-  }
-
-  get writePermission(): boolean {
+  get writerPermission(): boolean {
     return this.props.resourceCard.permission === 'OWNER'
       || this.props.resourceCard.permission === 'WRITER';
   }
 
-  get deletePermission(): boolean {
+  get ownerPermission(): boolean {
     return this.props.resourceCard.permission === 'OWNER';
   }
 
@@ -510,18 +506,18 @@ export class ResourceCard extends React.Component<Props, State> {
                         data-test-id='card'>
         <div style={{display: 'flex', flexDirection: 'column', alignItems: 'flex-start'}}>
           <div style={{display: 'flex', flexDirection: 'row', alignItems: 'flex-start'}}>
-            <ResourceCardMenu disabled={this.actionsDisabled}
+            <ResourceCardMenu disabled={false}
                               resourceType={this.resourceType}
                               onCloneResource={() => this.cloneResource()}
                               onCopyConceptSet={() => this.setState({copyingConceptSet: true})}
-                              canDelete={this.deletePermission}
+                              canDelete={this.ownerPermission}
                               onDeleteResource={() => this.openConfirmDelete()}
                               onRenameResource={() => this.renameResource()}
-                              canEdit={this.writePermission}
+                              canEdit={this.writerPermission}
                               onEdit={() => this.edit()}
                               onExportDataSet={() => this.exportDataSet()}
                               onReviewCohort={() => this.reviewCohort()}/>
-            <Clickable disabled={this.actionsDisabled}>
+            <Clickable>
               <a style={styles.cardName}
                    data-test-id='card-name'
                  href={this.getResourceUrl()}
--- a/ui/src/app/components/resource-card.tsx
+++ b/ui/src/app/components/resource-card.tsx
@@ -162,12 +162,16 @@ export class ResourceCard extends React.Component<Props, State> {
   }
 
   get actionsDisabled(): boolean {
-    return !this.writePermission;
+    return !this.writePermission && !this.isConceptSet
   }
 
   get writePermission(): boolean {
     return this.props.resourceCard.permission === 'OWNER'
       || this.props.resourceCard.permission === 'WRITER';
+  }
+
+  get deletePermission(): boolean {
+    return this.props.resourceCard.permission === 'OWNER';
   }
 
   get displayName(): string {
@@ -510,8 +514,10 @@ export class ResourceCard extends React.Component<Props, State> {
                               resourceType={this.resourceType}
                               onCloneResource={() => this.cloneResource()}
                               onCopyConceptSet={() => this.setState({copyingConceptSet: true})}
+                              canDelete={this.deletePermission}
                               onDeleteResource={() => this.openConfirmDelete()}
                               onRenameResource={() => this.renameResource()}
+                              canEdit={this.writePermission}
                               onEdit={() => this.edit()}
                               onExportDataSet={() => this.exportDataSet()}
                               onReviewCohort={() => this.reviewCohort()}/>
--- a/ui/src/app/components/resources.tsx
+++ b/ui/src/app/components/resources.tsx
@@ -9,12 +9,12 @@ import {ResourceType} from 'app/utils/resourceActionsReact';
 
 export const ResourceCardMenu: React.FunctionComponent<{
   disabled: boolean, resourceType: ResourceType, onRenameResource?: Function,
-  onCloneResource?: Function, onCopyConceptSet?: Function, onDeleteResource?: Function,
-  onEdit?: Function, onExportDataSet: Function, onReviewCohort?: Function,
+  onCloneResource?: Function, onCopyConceptSet?: Function, canDelete: boolean, onDeleteResource?: Function,
+  canEdit: boolean, onEdit?: Function, onExportDataSet: Function, onReviewCohort?: Function,
 }> = ({
         disabled, resourceType, onRenameResource = () => {}, onCloneResource = () => {},
-        onCopyConceptSet = () => {}, onDeleteResource = () => {}, onEdit = () => {},
-        onExportDataSet = () => {}, onReviewCohort = () => {}
+        onCopyConceptSet = () => {}, canDelete, onDeleteResource = () => {},
+        canEdit, onEdit = () => {}, onExportDataSet = () => {}, onReviewCohort = () => {}
       }) => {
   return <PopupTrigger
     data-test-id='resource-card-menu'
@@ -39,9 +39,9 @@ export const ResourceCardMenu: React.FunctionComponent<{
         }],
         ['conceptSet', () => {
           return <React.Fragment>
-            <MenuItem icon='pencil' onClick={onEdit}>Rename</MenuItem>
+            <MenuItem icon='pencil' onClick={onEdit} disabled={!canEdit}>Rename</MenuItem>
             <MenuItem icon='copy' onClick={onCopyConceptSet}>Copy to another workspace</MenuItem>
-            <MenuItem icon='trash' onClick={onDeleteResource}>Delete</MenuItem>
+            <MenuItem icon='trash' onClick={onDeleteResource} disabled={!canDelete}>Delete</MenuItem>
           </React.Fragment>;
         }],
         ['dataSet', () => {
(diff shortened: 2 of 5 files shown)


ui/src/app/components/resource-card.tsx
```diff
@@ -162,14 +162,18 @@ export class ResourceCard extends React.Component<Props, State> {
   }
 
   get actionsDisabled(): boolean {
-    return !this.writePermission;
+    return !this.writePermission && !this.isConceptSet;
```
blrubenstein: Is it worth calling this out specifically for concept sets? I think its okay for us to enable the dropdown and have nothing inside it enabled for other resources as well.

ui/src/app/components/resource-card.tsx
```diff
@@ -162,14 +162,18 @@ export class ResourceCard extends React.Component<Props, State> {
...
   }
 
   get writePermission(): boolean {
     return this.props.resourceCard.permission === 'OWNER'
       || this.props.resourceCard.permission === 'WRITER';
   }
 
+  get deletePermission(): boolean {
```
blrubenstein: Nit: I'd probably name this `ownPermission` or something to match above, or change `writePermission` to `editPermission` (or something similar). Just for semantic consistency. I lean towards `own/ownerPermission` because nothing stops us from adding other owner only actions.

ui/src/app/components/resources.tsx
```diff
@@ -54,6 +57,7 @@ export const ResourceCardMenu: React.FunctionComponent<{
         }]
       )
     }
+    disabled={resourceType !== ResourceType.CONCEPT_SET || !canEdit}
```
blrubenstein: Won't this be always disabled for something not a concept set? This should maybe be an &&?

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -79,29 +94,47 @@ const ConceptSetMenu: React.FunctionComponent<{
...
 };
 
-export const ConceptSetDetails =
-  fp.flow(withUrlParams(), withCurrentWorkspace())(class extends React.Component<{
-    urlParams: any, workspace: WorkspaceData}, {
-      conceptSet: ConceptSet, deleting: boolean, editName: string, editDescription: string,
-      editSaving: boolean, error: boolean, errorMessage: string, editing: boolean,
-      loading: boolean, removingConcepts: boolean, removeSubmitting: boolean,
-      selectedConcepts: Concept[], workspacePermissions: WorkspacePermissions}> {
+export interface ConceptSetProps {
```
blrubenstein: Nit: I'd name these `ConceptSetDetailsProps` and state to match with component name.

ui/src/app/components/resource-card.tsx
als364: I called it out for only concept sets because I was concerned that it would be confusing to have dropdowns that open but don't let you do anything. If you think it would be ok, I'll change it, then we don't have a magic conditional gating this.

ui/src/app/components/resource-card.tsx
als364: 👍

ui/src/app/components/resources.tsx
als364: oh, snap, you're right! thank you!

ui/src/app/pages/data/concept/concept-set-details.tsx
als364: 👍

ui/src/app/components/resource-card.tsx
```diff
@@ -161,15 +161,15 @@ export class ResourceCard extends React.Component<Props, State> {
     return this.resourceType === ResourceType.DATA_SET;
   }
 
-  get actionsDisabled(): boolean {
-    return !this.writePermission;
-  }
-
-  get writePermission(): boolean {
+  get writerPermission(): boolean {
```
jaycarlton: So some of the boolean properties begin with `'is`. I'd follow that is/has convention and make this `hasWriterPermission` and `hasOwnerPermission`. If I just see `foo.ownerPermission()` I don't know if it has a side effect or not, or what it returns.

ui/src/app/components/resource-card.tsx
```diff
@@ -506,16 +506,18 @@ export class ResourceCard extends React.Component<Props, State> {
                         data-test-id='card'>
         <div style={{display: 'flex', flexDirection: 'column', alignItems: 'flex-start'}}>
           <div style={{display: 'flex', flexDirection: 'row', alignItems: 'flex-start'}}>
-            <ResourceCardMenu disabled={this.actionsDisabled}
+            <ResourceCardMenu disabled={false}
                               resourceType={this.resourceType}
                               onCloneResource={() => this.cloneResource()}
                               onCopyConceptSet={() => this.setState({copyingConceptSet: true})}
+                              canDelete={this.ownerPermission}
                               onDeleteResource={() => this.openConfirmDelete()}
```
jaycarlton: nit/question: in Javascript, can you just pass a function name for a no-arg method where a lambda is expected? I.e. can you just do 
`onDeleteResource={this.openConfirmDelete}`?

ui/src/app/components/resources.tsx
```diff
@@ -9,12 +9,13 @@ import {ResourceType} from 'app/utils/resourceActionsReact';
 
 export const ResourceCardMenu: React.FunctionComponent<{
   disabled: boolean, resourceType: ResourceType, onRenameResource?: Function,
-  onCloneResource?: Function, onCopyConceptSet?: Function, onDeleteResource?: Function,
-  onEdit?: Function, onExportDataSet: Function, onReviewCohort?: Function,
+  onCloneResource?: Function, onCopyConceptSet?: Function, canDelete: boolean,
+  onDeleteResource?: Function, canEdit: boolean, onEdit?: Function,
+  onExportDataSet: Function, onReviewCohort?: Function,
```
jaycarlton: nit: I can't wrap my head around multiple lambdas in one line, particularly when they're mixed in with multiple assignments. I'd put one property per line. (Does the linter have an opinion on this?)

ui/src/app/components/resources.tsx
```diff
@@ -24,32 +25,48 @@ export const ResourceCardMenu: React.FunctionComponent<{
       switchCase(resourceType,
```
jaycarlton: nit: There's a bit of repetition here. It would be cleaner I think to externalize these fragments so the `switchCase` is more readable. Like I can read code with a bit of UI in it, or UI with a bit of code in it, but code containing UI that contains code seems error prone to me. Maybe a little function that converts icon, click handler, disabled, and text into a MenuItem, and another function that takes a list of MenuItems to make a React.Fragment. 

Or even have a little component just for this. There are opportunities for individual menu modes or menu items to be . inconsistent during maintenance with the current approach.

ui/src/app/components/resource-card.tsx
```diff
@@ -161,15 +161,15 @@ export class ResourceCard extends React.Component<Props, State> {
     return this.resourceType === ResourceType.DATA_SET;
   }
 
-  get actionsDisabled(): boolean {
-    return !this.writePermission;
-  }
-
-  get writePermission(): boolean {
+  get writerPermission(): boolean {
     return this.props.resourceCard.permission === 'OWNER'
```
jaycarlton: nit/idea: it's possible to implement permissions so that they support greater than, for example if you wanted owner to always be higher than (and include) writer. Then you could have `permission >= 'WRITER'`. Also, can we use an enum instead of string literals? I thought Swagger could define that for us from the API deinition.

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -79,29 +94,47 @@ const ConceptSetMenu: React.FunctionComponent<{
...
-      conceptSet: ConceptSet, deleting: boolean, editName: string, editDescription: string,
-      editSaving: boolean, error: boolean, errorMessage: string, editing: boolean,
-      loading: boolean, removingConcepts: boolean, removeSubmitting: boolean,
-      selectedConcepts: Concept[], workspacePermissions: WorkspacePermissions}> {
+export interface ConceptSetDetailsProps {
+  urlParams: any;
+  workspace: WorkspaceData;
+}
+
+export interface ConceptSetDetailsState {
```
jaycarlton: 👏 Much cleaner.

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -204,10 +254,11 @@ export const ConceptSetDetails =
         <div style={{display: 'flex', flexDirection: 'column'}}>
           <div style={styles.conceptSetHeader}>
             <div style={{display: 'flex', flexDirection: 'row'}}>
-              <ConceptSetMenu canDelete={workspacePermissions.isOwner}
-                              canEdit={workspacePermissions.canWrite}
+              <ConceptSetMenu canDelete={workspace.accessLevel === WorkspaceAccessLevel.OWNER}
```
jaycarlton: nit: I thought there was a `canDelete` helper somewhere around here. This comparison is more logic than I like to see inside html; one level of abstraction higher like `workspace.canDelete` would be more readable and future-proof I think.

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -313,6 +369,15 @@ export const ConceptSetDetails =
                   Remove concepts</Button>
           </ModalFooter>
         </Modal>}
+        {copying && <CopyModal
+          fromWorkspaceNamespace={workspace.namespace}
+          fromWorkspaceName={workspace.id}
+          fromResourceName={conceptSet.name}
+          resourceType={ResourceType.CONCEPT_SET}
+          onClose={() => this.setState({copying: false})}
+          onCopy={() => this.onCopy()}
```
jaycarlton: same question re: the fat arrow function vs just function name. Maybe it's a scope thing.

ui/src/app/components/resource-card.tsx
als364: Yes, you can do this. I suspect the reason it hasn't been done in the other cases is that Javascript functions are first class objects, so it's easy to screw up:
`onDeleteResource={() => this.openConfirmDelete}` would return the function `this.openConfirmDelete` rather than calling it.
See the `MenuItem` in concept-set-details for an example.
`onDeleteResource={this.openConfirmDelete()}` probably also does something weird, but I don't know offhand what it would do.

ui/src/app/components/resources.tsx
als364: The linter has no opinion on this aside from 'don't have lines longer than 100 characters' which is the reason I shuffled these.

ui/src/app/components/resources.tsx
als364: You're right. If this is <30m of effort in my update-all-the-tests PR, I'll break it out.

ui/src/app/pages/data/concept/concept-set-details.tsx
als364: there's a `WorkspacePermissions` class you can create to get this if you have the entire `Workspace` and `WorkspaceAccessLevel`. I could change it to only need the `WorkspaceAccessLevel` for this function and just use that.

ui/src/app/components/resource-card.tsx
jaycarlton: But does it expose a string-valued property like `WorkspaceAccessLevel.OWNER.name` or something so you don't have to spell it out?

I wouldn't rely on the existing numerical value for comparison unless it was explicitly designed that way b/c it's not future-proof.

ui/src/app/components/resource-card.tsx
blrubenstein: In general, I think that get methods should never have side effects. Happy to have it changed, but I think following that rule is the most important thing.

ui/src/app/components/resource-card.tsx
blrubenstein: Yeah, in general I tend towards something like `this.props.resourceCard.permission === WorkspaceAccessLevel.OWNER` and pass around the access levels. (even over using a string value property)

ui/src/app/components/resource-card.tsx
blrubenstein: You can, but it doesn't maintain `this` bindings. `() => this.method()` calls this.method within the current scope of this, whereas if you pass down the function, it will cal the function with `this` scoped to wherever you are.

There's another way to do it which is pass `this.method.bind(this)`, but in general that is less good, because IIRC it makes the function akin to a scoped global, as opposed to just providing a callback.

ui/src/app/components/resources.tsx
als364: This is going to be nontrivial - the best way to do it would be along the lines of #2327. @ericsong is making a backlog tech debt ticket to convert non-notebook resource card types to this paradigm.

jaycarlton: Awesome, thanks for the update.

PR #2490, opened 2019-08-21
[RW-2189][risk=low] Implement cronjob for alerting on per-project billing usage

ericsong:
Pretty straightforward.

- Grab (project, cost) data from BigQuery
- Join with Users through workspace.creator_id 
- Group by User to get per user total costs

Misc
- Created a billing config object to group billing fields
- Wrote unit tests by mocking the BQ query. The `TableResult` object returned by BQ is serializable so we can write it to disk and read from it for tests. We can use this method for some of BQ integration tests if we want to speed them up.


api/src/test/java/org/pmiops/workbench/billing/BillingAlertsServiceTest.java
```diff
@@ -0,0 +1,195 @@
...
+    billingAlertsService.alertUsersExceedingFreeTierBilling();
+    verifyZeroInteractions(notificationService);
+  }
+
+  @Test
+  public void alertUsersExceedingFreeTierBilling_combinedProjectsExceedsLimit() {
+    workbenchConfig.billing.defaultFreeCreditsLimit = 500.0;
+
+    User user = createUser("test@test.com");
+    createWorkspace(user, "aou-test-f1-26");
```
calbach: Hmm, the downside to this serialization approach is that I have no way of directly inspecting the test data in source code. e.g. what is the value for each of these workspaces? Presumably it sums to >500, but if this test broke I'd be a bit lost. Even something like JSON checked in I think would be a step up from this binary serialization.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -40,21 +41,29 @@ public static WorkbenchConfig createEmptyConfig() {
...
   }
 
+  public static class BillingConfig {
+
+    public Integer billingRetryCount;
+    public Integer billingProjectBufferCapacity;
+    public String billingProjectPrefix;
+    public String billingAccountId;
+    public String billingExportBigQueryTable;
+    public ArrayList<String> whitelistedUsers = new ArrayList<>();
```
calbach: Presumably this was from a previous iteration - should revert

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -40,21 +41,29 @@ public static WorkbenchConfig createEmptyConfig() {
     config.mandrill = new MandrillConfig();
     config.moodle = new MoodleConfig();
     config.server = new ServerConfig();
+    config.billing = new BillingConfig();
     return config;
   }
 
+  public static class BillingConfig {
+
+    public Integer billingRetryCount;
```
calbach: The billing prefix here and below is now probably redundant, would drop it

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -38,4 +38,10 @@
       "SELECT user FROM User user WHERE user.dataAccessLevel IN :dals AND ( lower(user.email) LIKE lower(concat('%', :term, '%')) OR lower(user.familyName) LIKE lower(concat('%', :term, '%')) OR lower(user.givenName) LIKE lower(concat('%', :term, '%')) )")
   List<User> findUsersByDataAccessLevelsAndSearchString(
       @Param("dals") List<Short> dataAccessLevels, @Param("term") String term, Sort sort);
+
+  // This query can return more than 1 row but we can grab any one of them because the creator value
```
calbach: Does Hibernate not complain in this situation? Should it have a `limit 1` or something?

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -0,0 +1,80 @@
...
+    QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder(
+                "SELECT project.id, SUM(cost) cost FROM `"
+                    + workbenchConfigProvider.get().billing.billingExportBigQueryTable
+                    + "` GROUP BY project.id ORDER BY cost desc;")
+            .build();
+
+    Map<User, Double> perUserSpend =
+        Streams.stream(bigQueryService.executeQuery(queryConfig).getValues())
+            .filter(fv -> !fv.get("id").isNull() && !fv.get("id").isNull())
```
calbach: I've spotted some redundancy in this filter, did you mean to check whether it was a string?

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -0,0 +1,80 @@
...
+            .map(
+                fv ->
+                    new AbstractMap.SimpleImmutableEntry<>(
+                        userDao.findCreatorByWorkspaceNamespace(fv.get("id").getStringValue()),
```
calbach: I think it will probably be easier to code and will scale better to instead start before this stream with a single map of  BP -> user. Ultimately, you'll likely need to fetch every user.

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -0,0 +1,80 @@
...
+                        fv.get("cost").getDoubleValue()))
+            .filter(spend -> spend.getKey() != null)
+            .filter(
+                spend ->
+                    !workbenchConfigProvider
+                        .get()
+                        .billing
+                        .whitelistedUsers
```
calbach: per previous discussion, the whitelisting can be removed

api/config/config_local.json
```diff
@@ -2,19 +2,24 @@
...
     "vpcServicePerimeterName": "accessPolicies/228353087260/servicePerimeters/terra_dev_aou_test"
   },
+  "billing": {
+    "billingAccountId": "014D91-FCB792-33D2C0",
+    "billingProjectPrefix": "aou-rw-local1-",
+    "billingExportBigQueryTable": "all-of-us-workbench-test-bd.billing_data.gcp_billing_export_v1_014D91_FCB792_33D2C0",
+    "billingRetryCount": 2,
+    "billingProjectBufferCapacity": 2,
+    "defaultFreeCreditsLimit": 300.0,
+    "whitelistedUsers": []
```
jaycarlton: Whitelisted for what exactly? I.e. you might want multiple whitelists for different billing-related things at some point.

Also, does keeping this here imply we need to redeploy the config to add users to the whitelist, or can that be done while we're running?

api/config/config_staging.json
```diff
@@ -1,19 +1,24 @@
...
     "enforceRegistered": true,
     "leoBaseUrl": "https:\/\/notebooks.firecloud.org",
     "registeredDomainName": "all-of-us-registered-staging",
-    "billingRetryCount": 2,
-    "billingProjectBufferCapacity": 10,
     "xAppIdValue": "staging-AoU-RW",
     "vpcServicePerimeterName": ""
   },
+  "billing": {
+    "billingAccountId": "014D91-FCB792-33D2C0",
```
jaycarlton: since these are all scoped under billing, do we need the 'billing' prefix as well?

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -0,0 +1,80 @@
...
+      Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.bigQueryService = bigQueryService;
+    this.notificationService = notificationService;
+    this.userDao = userDao;
+    this.workbenchConfigProvider = workbenchConfigProvider;
+  }
+
+  public void alertUsersExceedingFreeTierBilling() {
+    QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder(
```
jaycarlton: Would make sense to have a new Dao class to contain this query?

jaycarlton: Is the BQ table's cost always aligned with the current billing cycle? I.e. is there a way to run this job on historical data to see how many users exceeded the free tier two months ago?

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
ericsong: I ended up removing this when I changed the DAO to return the full bp -> user list in a single query.

But for knowledge sharing, it seems to just work. Also Hibernate does not seem to support limit queries? https://stackoverflow.com/a/1239745 . There is a workaround via using a native query but I think they should be avoided where possible since our tests do not use MySQL.

api/src/test/java/org/pmiops/workbench/billing/BillingAlertsServiceTest.java
ericsong: I looked into some methods to generate the TableResult from the JSON but couldn't find a quick way where I would trust the generated object's consistency with the real thing. I checked in the JSON as an intermediary measure.

PR #2491, opened 2019-08-22
[RW-3304][risk=low] Add fallback snippets behavior for large notebooks

calbach:
The session data I was depending on was null for large notebooks, i.e. while Jupyter UI was still initializing. Fallback to listening for notebook load.

### DIFF
--- a/api/cluster-resources/aou-snippets-menu.js.template
+++ b/api/cluster-resources/aou-snippets-menu.js.template
@@ -1,7 +1,9 @@
 define([
+  'jquery',
   'base/js/namespace',
+  'base/js/events',
   'nbextensions/snippets_menu/main'
-], (Jupyter, snippets_menu) => {
+], ($, Jupyter, events, snippets_menu) => {
   // Jupyter UI extension to customize the "Snippets Menu" extension with AoU
   // specific code snippets. See
   // https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/snippets_menu/readme.html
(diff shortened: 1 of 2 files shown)


api/cluster-resources/aou-snippets-menu.js.template
```diff
@@ -1,38 +1,58 @@
...
-    } else if (kernelName === "ir") {
-      customMenus = [rMenu];
-    }
-    snippets_menu.options['menus'] = snippets_menu.default_menus;
-    snippets_menu.options['menus'][0]['sub-menu'].splice(0, 0, ...customMenus);
+      // Include custom menu content from the nbconfigurator UI extension, if any.
+      // We primarily expect this to be used for manual testing of
+      // workbench-snippets code, so prefix it to differentiate from the menus
+      // installed by this extension.
+      const snippetsConfig = Jupyter.notebook.config.data.snippets || {};
```
jaycarlton: Are we restricted to plain javascript here instead of typeScript?

api/cluster-resources/aou-snippets-menu.js.template
```diff
@@ -1,38 +1,58 @@
...
 
-    console.log('loaded custom AoU snippet-menus');
+        // Kernel name should now be available.
+        handleKernel(Jupyter.notebook.session.kernel_model.name);
+        snippets_menu.menu_setup(
+            snippets_menu.options.menus, snippets_menu.options.sibling, /* insertBeforeSibling */ false);
+        console.log('loaded AoU snippets on event');
+      });
+    }
+    console.log('loaded custom AoU snippet-menus extension');
```
jaycarlton: is there a version of the extension or other metadata you might want here?

api/cluster-resources/aou-snippets-menu.js.template
```diff
@@ -1,38 +1,58 @@
...
+    // Session may be null at this point, e.g. for large notebooks, while the
+    // Jupyter UI is still doing initial rendering.
+    if (Jupyter.notebook.session && Jupyter.notebook.session.kernel_model) {
+      handleKernel(Jupyter.notebook.session.kernel_model.name);
+      console.log('loaded AoU snippets on initial load');
+    } else {
+      console.log('session data not yet available, listening for notebook load');
+      events.on('notebook_loaded.Notebook', () => {
+        // We can't avoid the initial Snippets menu insertion, so remove it here and render again.
+        $('.dropdown:contains("Snippets")').remove();
```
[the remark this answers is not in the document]
calbach: I agree, but there isn't a clear way. It's fair to call out that this is brittle and we should plan for it to fail. I've updated this to  have a reasonable fallback if we can't identify the snippets menu and log an error, and also added more comments.

api/cluster-resources/aou-snippets-menu.js.template
calbach: Not really. Best I could think of would be to plug in the git commit hash here, but I don't think it's worth the effort. There's only one source file here so it's easy enough to see which version you're looking at (and these logs go nowhere but the user's browser console).

PR #2492, opened 2019-08-22
[RW-3229] [risk=no] first pass on tests + refactors

als364:
Updated one unit test and a small refactor to make ResourceCardMenu inputs nicer.

Changes related to RW-2625 will go in another PR.

### DIFF
--- a/ui/src/app/components/resource-card.spec.tsx
+++ b/ui/src/app/components/resource-card.spec.tsx
@@ -122,21 +122,21 @@ describe('ResourceCardComponent', () => {
   // without a browser to define bounding boxes...
   it('should always have a clickable popup menu (cohort)', () => {
     const wrapper = component(ResourceCardWrapper.readonlyCohortCard);
-    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(false);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(undefined);
   });
 
   it('should always have a clickable popup menu (cohort review)', () => {
     const wrapper = component(ResourceCardWrapper.readonlyCohortReviewCard);
-    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(false);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(undefined);
   });
 
   it('should always have a clickable popup menu (concept set)', () => {
     const wrapper = component(ResourceCardWrapper.readonlyConceptSetCard);
-    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(false);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(undefined);
   });
 
   it('should always have a clickable popup menu (notebook)', () => {
     const wrapper = component(ResourceCardWrapper.readonlyNotebookCard);
-    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(false);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBe(undefined);
   });
 });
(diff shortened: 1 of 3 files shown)


ui/src/app/components/resource-card.spec.tsx
```diff
@@ -82,6 +118,25 @@ describe('ResourceCardComponent', () => {
...
+  // We can't test the popup menus themselves in this test because PopupTrigger doesn't know how to draw the modal
+  // without a browser to define bounding boxes...
+  it('should always have a clickable popup menu (cohort)', () => {
+    const wrapper = component(ResourceCardWrapper.readonlyCohortCard);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBeUndefined();
+  });
+
+  it('should always have a clickable popup menu (cohort review)', () => {
+    const wrapper = component(ResourceCardWrapper.readonlyCohortReviewCard);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBeUndefined();
```
jaycarlton: if the `find` fails, that doesn't also give you `undefined` that would make the test pass erroneously, does it?

ui/src/app/components/resources.tsx
```diff
@@ -3,79 +3,141 @@ import * as React from 'react';
...
+            </React.Fragment>;
+          }]
+        )
+      }
+    >
+      <Clickable data-test-id='resource-menu'>
+        <ClrIcon shape='ellipsis-vertical' size={21}
+                 style={{
+                   color: colors.accent,
+                   marginLeft: -9, cursor: 'pointer'
```
jaycarlton: could break after the -9

ui/src/app/components/resource-card.spec.tsx
als364: If `find` fails, then `props` will die; it's meant to be run on exactly one node.

ui/src/app/components/resource-card.spec.tsx
```diff
@@ -82,6 +118,25 @@ describe('ResourceCardComponent', () => {
     expect(lastModifiedDate(wrapper)).toBeTruthy();
   });
 
-  // Note: this spec is not testing the Popup menus on resource cards due to an issue using
-  //    PopupTrigger in the test suite.
+  // We can't test the popup menus themselves in this test because PopupTrigger doesn't know how to draw the modal
+  // without a browser to define bounding boxes...
+  it('should always have a clickable popup menu (cohort)', () => {
+    const wrapper = component(ResourceCardWrapper.readonlyCohortCard);
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBeUndefined();
```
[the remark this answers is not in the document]
als364: I removed the `disabled` property because we don't want to disable the triggers anymore. It would probably be clearer to word this test as 'disabled not in props'.

ui/src/app/components/resource-card.spec.tsx
```diff
@@ -122,21 +122,21 @@ describe('ResourceCardComponent', () => {
   // without a browser to define bounding boxes...
   it('should always have a clickable popup menu (cohort)', () => {
     const wrapper = component(ResourceCardWrapper.readonlyCohortCard);
-    expect(wrapper.find('[data-test-id="resource-menu"]').props()['disabled']).toBeUndefined();
+    expect(wrapper.find('[data-test-id="resource-menu"]').props()).not.toContain('disabled');
```
jaycarlton: 🏅 Much better. 

One other thing that might make it more readable is to pull out a helper method to fetch the resource menu (or its props), since you do that in several places.

ui/src/app/components/resources.tsx
```diff
@@ -135,7 +135,8 @@ export class ResourceCardMenu extends React.Component<ResourceCardMenuProps> {
         <ClrIcon shape='ellipsis-vertical' size={21}
                  style={{
                    color: colors.accent,
-                   marginLeft: -9, cursor: 'pointer'
+                   marginLeft: -9,
```
jaycarlton: nit, but I reserve the right to be more adamant next time 😄: where do these magic numbers for size and margin come from?

als364: ... in a different PR cause it's actually a lot of changes

als364:
rebasing off master.

PR #2495, opened 2019-08-22
[RW-2713][RW-3276][Risk=moderate] Srubenst/perf env

blrubenstein:


api/config/cdr_versions_perf.json
```diff
@@ -0,0 +1,27 @@
+[
+  {
+    "cdrVersionId": 1,
+    "name": "Synthetic Dataset v1",
+    "dataAccessLevel": 1,
+    "bigqueryProject": "fc-aou-cdr-synthetic-test",
```
calbach: I'd suggest pointing to all-of-us-ehr-dev for now, eventually we should probably create a separate CDR project for perf

api/config/config_perf.json
```diff
@@ -0,0 +1,73 @@
...
+    "clusterIdleMaxAgeDays": 3,
+    "debugEndpoints": false,
+    "enforceRegistered": true,
+    "leoBaseUrl": "https:\/\/leonardo.dsde-perf.broadinstitute.org",
+    "registeredDomainName": "all-of-us-registered-perf",
+    "registeredDomainGroup": "all-of-us-registered-perf@perf.test.firecloud.org",
+    "billingRetryCount": 4,
+    "billingProjectBufferCapacity": 20,
+    "xAppIdValue": "perf-AoU-RW",
+    "vpcServicePerimeterName": "accessPolicies/357236995176/servicePerimeters/terra_prod_aou_perf"
```
calbach: Does this exist? Also is "terra_prod_" the correct prefix here?

api/config/config_perf.json
```diff
@@ -0,0 +1,73 @@
...
+      "all-of-us-rw-perf@appspot.gserviceaccount.com"
+    ]
+  },
+  "cdr": {
+    "debugQueries": true
+  },
+  "googleCloudStorageService": {
+    "clusterResourcesBucketName": "all-of-us-rw-perf-cluster-resources",
+    "credentialsBucketName": "all-of-us-rw-perf-credentials",
+    "emailImagesBucketName": "all-of-us-workbench-test-email-images"
```
calbach: Existing issue: why is this always coming from test? Should we add this to the deploy process?

api/libproject/devstart.rb
```diff
@@ -1452,12 +1463,11 @@ def deploy_gcs_artifacts(cmd_name, args)
     # This file must be readable by all AoU researchers and the Leonardo service
     # account (https://github.com/DataBiosphere/leonardo/issues/220). Just make
```
calbach: Please update comment

api/libproject/devstart.rb
```diff
@@ -1452,12 +1463,11 @@ def deploy_gcs_artifacts(cmd_name, args)
...
     # these public since these assets are public in GitHub anyways.
-    run_inline_or_log(op.opts.dry_run, %W{
-      gsutil acl ch -u AllUsers:R
-      gs://#{gcc.project}-cluster-resources/setup_notebook_cluster.sh
-      gs://#{gcc.project}-cluster-resources/playground-extension.js
-      gs://#{gcc.project}-cluster-resources/aou-snippets-menu.js
-    })
+    if auth_domain_group.end_with?('firecloud.org')
```
calbach: What is the intent with this check?

ui/app.yaml
```diff
@@ -57,6 +57,7 @@ handlers:
         'self'
         https://api-dot-all-of-us-workbench-test.appspot.com
         https://api-dot-all-of-us-rw-staging.appspot.com
+        https://api-dot-all-of-us-rw-perf.appspot.com
```
calbach: Nice catch

api/config/cdr_versions_perf.json
blrubenstein: I can change that. Will do.

api/config/config_perf.json
blrubenstein: This does not exist. Current plan was to punt on VPC perimeter for perf for the time being. I'll change to TODO or something similar.

api/config/config_perf.json
blrubenstein: Not sure. I'm going to say out of scope for this though.

api/libproject/devstart.rb
blrubenstein: Will do.

api/libproject/devstart.rb
blrubenstein: This is intended to not run anything until the group exists, because we need to have deployed to create the group, and this is part of the deploy process. This sorta avoids circular requirements. (There's a step in the doc to temporarily make it TODO until you can create the auth domain.)

api/config/config_perf.json
blrubenstein: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-3317&epics=visible&selectedEpic=RW-1801

api/libproject/devstart.rb
calbach: OK, that sounds fine. I still don't understand the specific string check though. How about:

```
if !auth_domain_group.nil? and !auth_domain_group.empty?
```

api/config/config_perf.json
```diff
@@ -13,7 +13,7 @@
     "billingRetryCount": 4,
     "billingProjectBufferCapacity": 20,
     "xAppIdValue": "perf-AoU-RW",
-    "vpcServicePerimeterName": "accessPolicies/357236995176/servicePerimeters/terra_prod_aou_perf"
+    "vpcServicePerimeterName": "TODO"
```
calbach: nit: would just omit this parameter entirely and leave the TODO in a comment if needed

ui/src/environments/environment.perf.ts
```diff
@@ -0,0 +1,18 @@
+import {Environment} from 'environments/environment-type';
+
+export const environment: Environment = {
+  displayTag: 'Perf',
+  shouldShowDisplayTag: true,
+  allOfUsApiUrl: 'https://api-dot-all-of-us-rw-perf.appspot.com',
+  clientId: '63939010390-aj0r8hro7r8lkt7a45gissu3m73ietl2.apps.googleusercontent.com',
+  leoApiUrl: 'https://leonardo.dsde-dev.broadinstitute.org',
```
calbach: perf

api/config/config_perf.json
blrubenstein: I think the build will complain if its completely omitted.

calbach: I don't think it will. But if you're concerned about it, I'll also accept empty string. I just don't want "TODO" in a live code path on principle.

.circleci/config.yml
```diff
@@ -276,9 +276,10 @@ jobs:
               --key-file circle-sa-key.json \
               --promote
 
-  # In staging, we deploy all at once in order to track and control the release
-  # for a tagged version only. In "test", we maintain individual release jobs
-  # for API and UI to allow faster deploys and because we lack a version tag.
+  # In staging and perf, we deploy all at once in order to track and control
```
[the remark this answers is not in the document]
blrubenstein: I don't think we could easily do that via the circle config. I'm Also, this code changes very little, so I am going to push back and say I don't think its worth the time to do that.

api/config/cdr_versions_perf.json
```diff
@@ -0,0 +1,27 @@
...
+    "cdrVersionId": 1,
+    "name": "Synthetic Dataset v1",
+    "dataAccessLevel": 1,
+    "bigqueryProject": "all-of-us-ehr-dev",
+    "bigqueryDataset": "synthetic_cdr20180606",
+    "creationTime": "2017-12-26 00:00:00Z",
+    "releaseNumber": 1,
+    "numParticipants": 946237,
+    "cdrDbName": "synth_r_2019q3_2",
+    "elasticIndexBaseName": "synth_r_2019q1_1"
```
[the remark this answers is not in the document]
blrubenstein: This is copied over from our other environments, where these work. I have not manually verified, but I think I'd rather have the consistency. (I think these are also pointing at the same table, actually. Not entirely sure why we have it expressed twice, but I suspect its to test cdr version switching.)

api/config/config_perf.json
```diff
@@ -0,0 +1,73 @@
...
+    "clusterIdleMaxAgeDays": 3,
+    "debugEndpoints": false,
+    "enforceRegistered": true,
+    "leoBaseUrl": "https:\/\/leonardo.dsde-perf.broadinstitute.org",
+    "registeredDomainName": "all-of-us-registered-perf",
+    "registeredDomainGroup": "all-of-us-registered-perf@perf.test.firecloud.org",
+    "billingRetryCount": 4,
+    "billingProjectBufferCapacity": 20,
+    "xAppIdValue": "perf-AoU-RW",
+    "vpcServicePerimeterName": ""
```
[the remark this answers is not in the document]
blrubenstein: Will make one.

api/libproject/devstart.rb
```diff
@@ -1764,7 +1775,7 @@ def create_project_resources(gcc)
   common = Common.new
   common.status "Enabling APIs..."
```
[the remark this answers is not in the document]
blrubenstein: Will do. (Its an existing log though, so in general out of scope. Small enough change, but I think we should tend to avoid changing out of band things within the scope of a ticket unless its a refactor.)

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -51,6 +51,7 @@ public static WorkbenchConfig createEmptyConfig() {
     public Integer clusterMaxAgeDays;
     public Integer clusterIdleMaxAgeDays;
     public String registeredDomainName;
+    public String registeredDomainGroup;
```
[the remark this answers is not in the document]
blrubenstein: It is possible, I am going to add a ticket to look at breaking our config up more as tech debt. I think its probably bigger than just this set. (And might not be worth it, because it will add more boilerplate.)

common-ui/common-ui-devstart.rb
```diff
@@ -49,7 +49,7 @@ def build(cmd_name, ui_name, args)
   # and also uglifies the source.
   # See https://github.com/angular/angular-cli/wiki/build#--dev-vs---prod-builds.
   optimize = "--aot"
-  if Set['staging', 'stable', 'prod'].include?(options.env)
+  if Set['perf', 'staging', 'stable', 'prod'].include?(options.env)
```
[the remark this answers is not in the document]
blrubenstein: If its only used once (which I think it is), my preference is to leave as a constant and move to a variable as soon as it is used more than once. If you disagree, lets have that conversation as a follow up.

common-ui/common-ui-devstart.rb
```diff
@@ -273,6 +273,7 @@ def run
     validate_options
     environment_names = {
```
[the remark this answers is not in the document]
blrubenstein: Will update.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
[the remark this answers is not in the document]
jaycarlton: The idea is to raise the level of abstraction. Anytime you expect two or more pieces of data to appear together, I think you really want an object.

api/libproject/devstart.rb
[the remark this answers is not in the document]
jaycarlton: It's a slippery slope, but I think small, frequent refactors when you're in the neighborhood can be really helpful.

common-ui/common-ui-devstart.rb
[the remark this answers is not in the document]
jaycarlton: It's still a readability win, even if only used once. It's just like a string literal or number literal. It's a Set literal. It's an unnamed magic constant as written, just as if you had 3.14159 but only used it once.

And yes, I'm applying Java sensibilities to Ruby, which isn't necessarily fair.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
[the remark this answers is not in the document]
blrubenstein: I'm fine with that, making a tech debt ticket.

PR #2498, opened 2019-08-23
[RW-3134] [risk=no] close modal after choosing playground mode

als364:


ui/src/app/pages/analysis/interactive-notebook.tsx
```diff
@@ -286,7 +286,10 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
             onCancel={() => {
               this.setState({showInUseModal: false}); }}
             onCopy={() => {this.cloneNotebook(); }}
-            onPlaygroundMode={() => {this.startPlaygroundMode(); }}>
+            onPlaygroundMode={() => {
+              this.setState({showInUseModal: false});
```
calbach: Did the order matter here? I might have naturally written this in the reverse order (startPlayground, then hide), but I don't feel strongly.

als364: This ordering mirrors a similar set of commands a few lines further up in this module. The order shouldn't matter.

calbach: Ah, I didn't see that. +1 for consistency

PR #2499, opened 2019-08-23
[RW-3327][RISK=NO]/dataSet returning 500

NehaBroad:


api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
```diff
@@ -32,7 +32,7 @@
   private List<Long> conceptSetId;
   private List<Long> cohortSetId;
   private List<DataSetValues> values;
-  private short prePackagedConceptSet;
+  private Short prePackagedConceptSet;
```
jaycarlton: Why not store a `PrePackagedConceptSetEnum` here, and do the conversion on assignment?

Alternatively, you could use an `Optional<Short>` instead of just a boxed short. That would make things more obvious for the callers. Depends on whether 0 has special meaning in this enum that matches a notion of absense.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
```diff
@@ -180,16 +180,19 @@ public void setValues(List<DataSetValues> values) {
...
   }
 
-  public void setPrePackagedConceptSet(short prePackagedConceptSet) {
+  public void setPrePackagedConceptSet(Short prePackagedConceptSet) {
     this.prePackagedConceptSet = prePackagedConceptSet;
   }
 
   @Transient
   public PrePackagedConceptSetEnum getPrePackagedConceptSetEnum() {
+    if (this.prePackagedConceptSet == null) {
```
jaycarlton: We don't usually want `get` methods to mutate state, unless it's a lazy evaluation or other optimization. I'd probably move this check to the `set` method.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
NehaBroad: 0 has a meaning in the enum ('None' which results in none of the concept set selected on view) hence just keeping it Short solves the issue

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
NehaBroad: I actually do not need this check removing it

PR #2511, opened 2019-08-28
[RW-3289][risk=no] homepage changeup

andy7i:
Requirements from https://precisionmedicineinitiative.atlassian.net/browse/RW-3289

mocks: https://projects.invisionapp.com/d/#/console/12820961/363986673/preview

```
Change background and CSS changes to home page to match new mocks

UI mocks:

https://projects.invisionapp.com/d/#/console/12820961/363986673/preview

Dev Notes

Background color

Text changes (please work with Karthik on text)

Div structures

Feature Flag:

Yes - Put behind feature flag until product approval

AC:

Change background and CSS changes to home page to match new mocks


```

The new versions of Last Accessed Items / workspaces will be implemented in future tickets.


andy7i:
side by side comparisons
Screen Shot 2019-08-29 at 10 00 04 AM
Screen Shot 2019-08-29 at 10 01 05 AM

ui/src/environments/test-env-base.ts
```diff
@@ -12,4 +12,5 @@ export const testEnvironmentBase = {
   zendeskHelpCenterUrl: 'http://aousupporthelp.zendesk.com/hc',
   shibbolethUrl: 'http://mock-nih.dev.test.firecloud.org',
   shouldShowDisplayTag: true,
+  enableHomepageRestyle: true
```
andy7i: not sure if this is required here

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -302,124 +371,267 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+                style={{color: colors.accent, marginBottom: 17}}
+              />
+            </Clickable>
+          </div>
+          <video width='100%' controls autoPlay>
+            <source src={videoLink} type='video/mp4'/>
+          </video>
+        </Modal>}
+      </React.Fragment>;
+    } else { // delete this block below once enableHomepageRestyle is removed
```
andy7i: below this block is the old homepage, to be deleted when the entire homepage restyling epic is completed

ui/src/environments/test-env-base.ts
andy7i: seems like it essentially duplicates ui/src/environments/environment.test.ts and ui/src/environments/environment.local.ts (and gets overwritten by those files anyway)

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -302,124 +371,267 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
-                       src={thumbnail.src}/>
-                </Clickable>
-              </React.Fragment>;
-            })}
+    if (environment.enableHomepageRestyle) {
+      return <React.Fragment>
+        <div style={styles.pageWrapper}>
+          <div style={styles.welcomeMessage}>
+            <div style={{display: 'flex', flexDirection: 'column', width: '50%'}}>
+              <div style={{display: 'flex', flexDirection: 'row'}}>
```
NehaBroad: Since display: 'flex' and flexDirection is being used in so many style i wonder if we can extract it and use it in other styles as well as here

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -302,124 +371,267 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
-                        <div style={styles.footerText}>
-                          <div style={{color: colors.white, marginTop: '2%'}}>{col.title}</div>
-                          <ul style={{color: colors.secondary}}>
-                            {col.links.map((link, ii) => {
-                              return <li key={ii}>
-                                <a href='#' style={{color: colors.secondary}}>{link}</a>
-                              </li>;
-                            } )}
-                          </ul>
+            <div style={styles.quickTour}>
```
NehaBroad: Just to confirm this is something that was already there and there is no change to quicktour?

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -302,124 +371,267 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+              <div style={styles.quickTourLabel}>Quick Tour and Videos</div>
+              <div style={styles.quickTourCardsRow}>
+                {quickTourResources.map((thumbnail, i) => {
+                  return <React.Fragment key={i}>
+                    <Clickable onClick={thumbnail.onClick}
+                               data-test-id={'quick-tour-resource-' + i}>
+                      <img style={{maxHeight: '7rem', width: '9rem', marginRight: '0.5rem'}}
```
NehaBroad: The norm is , styles with 3 or less properties are inline else extract it.

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -18,22 +20,33 @@ import {getRegistrationTasksMap, RegistrationDashboard} from 'app/pages/homepage
...
     flexDirection: 'column', justifyContent: 'space-between'
   },
+  welcomeMessage: {
+    marginLeft: '3%', display: 'flex', flexDirection: 'row'
+  },
+  welcomeMessageIconRow: {
+    display: 'flex', flexDirection: 'row', alignItems: 'flex-end', marginLeft: '1rem'
+  },
+  welcomeMessageIcon: {
+    height: '2.25rem', width: '2.75rem'
```
NehaBroad: we can have this inline

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -18,22 +20,33 @@ import {getRegistrationTasksMap, RegistrationDashboard} from 'app/pages/homepage
...
-    backgroundRepeat: 'no-repeat', backgroundSize: 'cover', height: '100%',
+  pageWrapper: {
     marginLeft: '-1rem', marginRight: '-0.6rem', display: 'flex',
     flexDirection: 'column', justifyContent: 'space-between'
   },
+  welcomeMessage: {
+    marginLeft: '3%', display: 'flex', flexDirection: 'row'
+  },
+  welcomeMessageIconRow: {
+    display: 'flex', flexDirection: 'row', alignItems: 'flex-end', marginLeft: '1rem'
```
NehaBroad: it would be great to extract out display: 'flex' and flexDirection since its being in so many places

ui/src/app/pages/homepage/homepage.tsx
andy7i: Right, I didn't really change quicktour. Looks like github is a bit confused with this diff. This line is equivalent to line 350 in the original code: https://github.com/all-of-us/workbench/pull/2511/files#diff-64640e577f5207068021f36d41dfcedfL350

ui/src/app/pages/homepage/homepage.tsx
andy7i: I added this here because it was used for 3 icons and I thought it would be easier to set the styles for all 3 at once instead of one at a time

ui/src/app/pages/homepage/homepage.tsx
andy7i: I extracted it out so that we can see what it would look like. I didn't change it for the code that will be deleted.

andy7i: https://github.com/all-of-us/workbench/pull/2511/commits/05dbb8b3bc9c3e0063efc94b89bab19ee31f0e87

andy7i: thoughts?

ui/src/app/styles/flex.tsx
```diff
@@ -0,0 +1,27 @@
+import {reactStyles} from 'app/utils';
```
gjuggler: I'm not sure this file is in the right place – is there a reason this doesn't live within app/components/flex.tsx?

These are definitely React components being produced (even though they're lightweight wrappers largely intended to enforce some style invariant). It will probably be more consistent to have this live alongside all other common components, even if it has little to no interactivity.

I'm not wedded to this view, though – I'd be happy to hear a convincing argument in the other direction!

ui/src/environments/test-env-base.ts
gjuggler: The idea behind this file was that you can define env variables here, and not redefine them in the other files, to clarify exactly how each environment differs from the base "test" environment. If you want to follow that pattern, just remove the redefinition of this value from environment.ts and environment.test.ts

(We could also follow-up at some point with environment.local to have it "inherit" from the test-env-base, since it should share most of the same properties as the test env.)

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -17,78 +19,132 @@ import {RecentWork} from 'app/pages/homepage/recent-work';
...
+import {Column, Row} from 'app/styles/flex';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
+import {environment} from 'environments/environment';
 import {
   BillingProjectStatus,
   Profile,
 } from 'generated/fetch';
 
 export const styles = reactStyles({
   mainHeader: {
```
gjuggler: Meta-comment (no action required): whenever we start to have massive lists of object properties like we see here, I always, always favor alphabetizing the properties. Unless there's a very compelling reason to do otherwise, I've found alphabetization is the only sane way to organize lists like this.

This is a pretty pervasive style nitpick though, and one which could / should easily be cleaned up at some point in the future in a massive sweep of the codebase. I just can't help myself from commenting when I come across examples which I'd love to clean up someday :)

ui/src/app/styles/flex.tsx
```diff
@@ -0,0 +1,27 @@
...
+    flexDirection: 'column'
+  },
+  row: {
+    display: 'flex',
+    flexDirection: 'row'
+  }
+});
+
+
+export const Row = (props) => {
```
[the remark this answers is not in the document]
andy7i: Good point. I like FlexRow/FlexColumn! I'll make that change.

ui/src/app/styles/flex.tsx
andy7i: Makes sense. Moved it to app/components/flex.tsx

ui/src/app/pages/homepage/homepage.tsx
andy7i: They actually are alphabetized, but the linter has them alphabetized by file path rather than by object name.

ui/src/environments/test-env-base.ts
andy7i: So I'm a little confused but my understanding is that `src/environments/environment.ts` gets replaced anyway in all environments except dev. CH and I were discussing in https://pmi-engteam.slack.com/archives/CHRN2R51N/p1567115237109500

That means that the env variable in `environment.ts` can be removed, as you also mentioned.

I think we wanted this feature flag to be enabled for the test environment, which would mean leaving it in `environment.test.ts`. Does that sound right/reasonable?

gjuggler: Yes, environment.ts gets replaced wholesale depending on which environment is being built & deployed.

But if you look at environment.test.ts, all it does it splat out the variables from test-env-base.ts (see this line https://github.com/all-of-us/workbench/blob/master/ui/src/environments/environment.test.ts#L5) and then add any overrides necessary. I created this pattern in an attempt to minimize the needless duplication between our test-aligned environment configs, & so it's easier to clearly tell which variables are different from the "base" test config.

I'm not 100% sure my attempt has succeeded, but it should be an easy pattern to follow, e.g. if you leave "enableHomepageRestyle: true" in the test-env-base and don't overwrite it in either the local or test config files.

ui/src/app/pages/homepage/homepage.tsx
gjuggler: Ah, my comment was about the list of object entries within this reactStyles({ ... }) call, not the file imports themselves.

ui/src/environments/test-env-base.ts
andy7i: Got it. Thanks for explaining!

ui/src/app/pages/homepage/homepage.tsx
andy7i: I can see this being a better option for long term maintainability. I sorted the lines that we will be keeping (but not the lines that will be deleted when the feature flag is removed).

PR #2518, opened 2019-08-29
[RW-3206][risk=moderate] Update CDR project locations

calbach:
Rated this as moderate risk since it defines a new connection to the recently created firecloud.org prod CDR dataset (note: this is likely to be applied tomorrow, ahead of the next actual code release).

- I decided to proliferate the number of CDR versions in staging/stable, rather than upgrade in place, to minimize disruption to existing workspaces.
- Production now defaults to the new firecloud.org CDR project

### DIFF
--- a/api/config/cdr_versions_prod.json
+++ b/api/config/cdr_versions_prod.json
@@ -1,7 +1,7 @@
 [
   {
     "cdrVersionId": 1,
-    "name": "Synthetic Dataset v1",
+    "name": "Synthetic Dataset v1 (Legacy)",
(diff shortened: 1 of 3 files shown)


calbach:
To list all CDRs in an environment, list the BQ datasets in a CDR cloud project, e.g. `bq ls --project_id fc-aou-cdr-synth-staging` (authenticated as the right user).

Keeping these in sync is probably something that could be tested automatically, e.g. by the update_cdr_versions tool. Checking that the referenced datasets actually exist seems like a reasonable idea to me.

PR #2521, opened 2019-08-29
[risk=no, ticket=no] Bump set-value to 2.0.1 per dependabot recommendation

jmthibault79:
PR-ing myself because it could not be done automatically.

Here's the security alert: https://github.com/all-of-us/workbench/network/alert/ui/yarn.lock/set-value/open


jaycarlton:
Do we have a unit test for the functionality that relies on this package?

PR #2522, opened 2019-08-29
[RW-3306][Risk=moderate] Move to workspaceAcl call on FireCloud

blrubenstein:

### DIFF
--- a/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
+++ b/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.when;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import java.sql.Timestamp;
 import java.time.Clock;
@@ -711,12 +712,10 @@ public class CohortReviewControllerBQTest extends BigQueryBaseTest {
 
   private void stubMockFirecloudGetWorkspaceAcl() throws ApiException {
     WorkspaceACL workspaceAccessLevelResponse = new WorkspaceACL();
-    HashMap<String, WorkspaceAccessEntry> acl = new HashMap<>();
     WorkspaceAccessEntry accessLevelEntry =
         new WorkspaceAccessEntry().accessLevel(WorkspaceAccessLevel.WRITER.toString());
-    acl.put(userProvider.get().getEmail(), accessLevelEntry);
-    workspaceAccessLevelResponse.setAcl(acl);
+    Map<String, WorkspaceAccessEntry> userEmailToAccessEntry = ImmutableMap.of(userProvider.get().getEmail(), accessLevelEntry);
+    workspaceAccessLevelResponse.setAcl(userEmailToAccessEntry);
     when(mockFireCloudService.getWorkspaceAcl(NAMESPACE, NAME)).thenReturn(workspaceAccessLevelResponse);
-
   }
 }
--- a/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
+++ b/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
@@ -46,6 +46,8 @@ import org.pmiops.workbench.db.model.User;
 import org.pmiops.workbench.db.model.Workspace;
 import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.firecloud.FireCloudService;
+import org.pmiops.workbench.firecloud.model.WorkspaceACL;
+import org.pmiops.workbench.firecloud.model.WorkspaceAccessEntry;
 import org.pmiops.workbench.firecloud.model.WorkspaceResponse;
 import org.pmiops.workbench.model.AllEvents;
 import org.pmiops.workbench.model.CohortChartData;
@@ -80,6 +82,7 @@ import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Scope;
 
 @RunWith(BeforeAfterSpringTestRunner.class)
 @Import({TestJpaConfig.class})
@@ -117,6 +120,7 @@ public class CohortReviewControllerBQTest extends BigQueryBaseTest {
 
   private Cohort cohort;
   private CohortReview review;
+  private static User currentUser;
 
   @TestConfiguration
   @Import({
@@ -147,6 +151,12 @@ public class CohortReviewControllerBQTest extends BigQueryBaseTest {
     }
 
     @Bean
+    @Scope("prototype")
+    User user() {
+      return currentUser;
+    }
+
+    @Bean
     public Clock clock() {
       return CLOCK;
     }
@@ -164,13 +174,14 @@ public class CohortReviewControllerBQTest extends BigQueryBaseTest {
   }
 
   @Before
-  public void setUp() {
+  public void setUp() throws Exception {
     User user = new User();
     user.setEmail("bob@gmail.com");
     user.setUserId(123L);
     user.setDisabled(false);
     user.setEmailVerificationStatusEnum(EmailVerificationStatus.SUBSCRIBED);
     user = userDao.save(user);
+    currentUser = user;
     when(userProvider.get()).thenReturn(user);
     controller.setUserProvider(userProvider);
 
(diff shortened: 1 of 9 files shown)


api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
```diff
@@ -1463,4 +1424,15 @@ private void stubMockFirecloudGetWorkspace() throws ApiException {
     workspaceResponse.setAccessLevel(WorkspaceAccessLevel.WRITER.toString());
     when(mockFireCloudService.getWorkspace(NAMESPACE, NAME)).thenReturn(workspaceResponse);
   }
+
+  private void stubMockFirecloudGetWorkspaceAcl() throws ApiException {
+    WorkspaceACL workspaceAccessLevelResponse = new WorkspaceACL();
```
jaycarlton: nit: I generally like to make local vars final where possible.  Also, "stub mock" seems a bit redundant, or at least a mouthful. 

Can you use `ImmutableMap.of()` to make the acl map? I'd also name it to indicate the key and value, like `userEmailToAccessEntry`.

Finally, I'm surprised you don't need more than one access level response for the tests, but i haven't read closely. You might think about having a function to build that response object based on passed-in info, and the use different responses in different `when` methods to test things.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -394,8 +399,16 @@ public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to)
...
-        fireCloudService.getWorkspace(workspaceNamespace, workspaceId).getAccessLevel();
+    WorkspaceACL workspaceACL = fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId);
+    WorkspaceAccessEntry workspaceAccessEntry =
+        Optional.of(workspaceACL.getAcl().get(userProvider.get().getEmail()))
+            .orElseThrow(
+                () ->
+                    new NotFoundException(
+                        String.format(
+                            "Workspace %s/%s not found", workspaceNamespace, workspaceId)));
+    String userAccess = workspaceAccessEntry.getAccessLevel();
```
jaycarlton: nit: final

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -394,8 +399,16 @@ public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to)
...
-    String userAccess =
-        fireCloudService.getWorkspace(workspaceNamespace, workspaceId).getAccessLevel();
+    WorkspaceACL workspaceACL = fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId);
+    WorkspaceAccessEntry workspaceAccessEntry =
+        Optional.of(workspaceACL.getAcl().get(userProvider.get().getEmail()))
+            .orElseThrow(
+                () ->
+                    new NotFoundException(
+                        String.format(
+                            "Workspace %s/%s not found", workspaceNamespace, workspaceId)));
```
jaycarlton: It'd help to have a utility function for formatting a fully qualified workspace.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -394,8 +399,16 @@ public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to)
   @Override
   public WorkspaceAccessLevel getWorkspaceAccessLevel(
       String workspaceNamespace, String workspaceId) {
-    String userAccess =
-        fireCloudService.getWorkspace(workspaceNamespace, workspaceId).getAccessLevel();
+    WorkspaceACL workspaceACL = fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId);
+    WorkspaceAccessEntry workspaceAccessEntry =
+        Optional.of(workspaceACL.getAcl().get(userProvider.get().getEmail()))
```
jaycarlton: If the user has no email (or it's invalid), it would be good to know that as well. If that's impossible, then this is probably fine.

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -276,6 +289,17 @@ private void stubGetWorkspace(String ns, String name, String creator, WorkspaceA
     fcResponse.setWorkspace(fcWorkspace);
     fcResponse.setAccessLevel(access.toString());
     when(fireCloudService.getWorkspace(ns, name)).thenReturn(fcResponse);
+    stubGetWorkspaceAcl(ns, name, creator, access);
+  }
+
+  private void stubGetWorkspaceAcl(String ns, String name, String creator, WorkspaceAccessLevel access) {
+    WorkspaceACL workspaceAccessLevelResponse = new WorkspaceACL();
+    HashMap<String, WorkspaceAccessEntry> acl = new HashMap<>();
```
jaycarlton: same comments as above. It would be nice to have an immutable class for the ACL stuff with a builder method, but this is fine.

api/src/test/java/org/pmiops/workbench/api/ConceptSetsControllerTest.java
```diff
@@ -773,6 +785,17 @@ private void stubGetWorkspace(String ns, String name, String creator, WorkspaceA
     fcResponse.setWorkspace(fcWorkspace);
     fcResponse.setAccessLevel(access.toString());
     when(fireCloudService.getWorkspace(ns, name)).thenReturn(fcResponse);
+    stubGetWorkspaceAcl(ns, name, creator, access);
+  }
+
+  private void stubGetWorkspaceAcl(String ns, String name, String creator, WorkspaceAccessLevel access) {
```
jaycarlton: If you need this in several places, then maybe build a reusable `testAclImpl` or something that you can have in all these test classes.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
blrubenstein: That should be covered earlier on in the process. I think handling email validation should be handled by createUser. (All users must have an email, it is how they log in.)

api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
blrubenstein: "Stub Mock" is what is used already in the file, so I'm inclined to stick with existing file naming. It is somewhat redundant. 

Currently, one access level is sufficient, and I think makes a good default. You can always override this for specific tests. (These are mostly for the sake of setup functions)

api/src/test/java/org/pmiops/workbench/api/ConceptSetsControllerTest.java
blrubenstein: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-2480&epics=visible&selectedEpic=RW-1801

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
blrubenstein: I think this is all over the place. Might be worth a tech debt cleanup.

blrubenstein: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-3365&epics=visible&selectedEpic=RW-1801

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -395,7 +400,8 @@ public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to)
   public WorkspaceAccessLevel getWorkspaceAccessLevel(
       String workspaceNamespace, String workspaceId) {
     String userAccess =
-        fireCloudService.getWorkspace(workspaceNamespace, workspaceId).getAccessLevel();
+        fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId).getAcl().get(userProvider.get().getEmail()).getAccessLevel();
```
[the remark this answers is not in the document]
jaycarlton: This would be a good thing for a standards document at some point. You already have a story for that in the backlog IIRC.

api/src/bigquerytest/java/org/pmiops/workbench/api/CohortReviewControllerBQTest.java
jaycarlton: That file is the only place that uses it. I'd vote for just changing it to `mock`

Do we have test cases for when the access level is not sufficient? That's why I was curious this function always gave one level of access. If that's a follow-on story that's OK too.

PR #2526, opened 2019-08-29
[RW-3354] [risk=low] Scope all BigQuery service instances to the current CDR project

gjuggler:
This fixes an issue we encountered when trying to turn on VPC-SC BigQuery enforcement in test. By default, the BigQuery client library executes queries in the context of the current default project (in our case, the App Engine project, e.g. all-of-us-workbench-test).

With VPC-SC BigQuery enforcement enabled, a query that's executed in the context of a project outside of the perimeter is considered "cross-perimeter", even if the query itself only accesses resources within the perimeter and the acting service account is whitelisted for access.

This PR should allow us to turn on BigQuery enforcement in the test environment. We may want to prepare a cherry-pick onto the current release in order to gain business & security approval to push it out next week. This will be required to enable enforcement on staging, stable and prod.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
+++ b/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
@@ -2,6 +2,7 @@ package org.pmiops.workbench.api;
 
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryException;
+import com.google.cloud.bigquery.BigQueryOptions;
 import com.google.cloud.bigquery.Field;
 import com.google.cloud.bigquery.FieldList;
 import com.google.cloud.bigquery.FieldValue;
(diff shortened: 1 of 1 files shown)


gjuggler:
I tested this locally – it doesn't seem to *break* access to the BigQuery tables, but without vpc-sc enforcement turned on in test it's hard to positively verify the fix.

api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -36,10 +37,16 @@
 
   private static final Logger logger = Logger.getLogger(BigQueryService.class.getName());
 
-  @Autowired private BigQuery bigquery;
-
   @Autowired private Provider<WorkbenchConfig> workbenchConfigProvider;
 
+  private BigQuery getBigQueryService() {
+    CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
```
jaycarlton: if you're not testing `cdrVersion` then there's no harm in doing this inline below.

api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -151,6 +158,6 @@ public FieldList getTableFieldsFromDomain(Domain d) {
     TableId tableId =
         TableId.of(cdrVersion.getBigqueryProject(), cdrVersion.getBigqueryDataset(), tableName);
 
-    return bigquery.getTable(tableId).getDefinition().getSchema().getFields();
+    return getBigQueryService().getTable(tableId).getDefinition().getSchema().getFields();
```
jaycarlton: What do you want to happen if the table with id `tableId` isn't found? I don't think that will map to an empty field list.

PR #2529, opened 2019-08-29
[RW-3029][Risk=no] Get rid of race condition on new data set modal

blrubenstein:


ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -129,7 +129,7 @@ class ExportDataSetModal extends React.Component<
     // Open notebook in a new tab and close the modal
     const notebookUrl = '/workspaces/' + workspaceNamespace + '/' + workspaceFirecloudName +
         '/notebooks/' + appendNotebookFileSuffix(encodeURIComponent(this.state.notebookName));
-    window.open(notebookUrl, '_blank');
+    window.open(notebookUrl);
```
calbach: This should probably be updated to match `new-dataset-modal`

blrubenstein:
@calbach should be good for another look!

PR #2530, opened 2019-08-29
[RW-3368][risk=no] Autoselect the default CDR when cloning

calbach:


calbach:
Tested manually locally.

PR #2531, opened 2019-08-30
[RW-3333][risk=low] Add org unit for new researcher accounts

calbach:
This sub-organization will have 2FA enforcement disabled, allowing us to self-manage 2FA enforcement at the application layer.

@gjuggler FYI


api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -38,6 +38,8 @@
   private static final String ALLOWED =
       "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*()";
   private static final String APPLICATION_NAME = "All of Us Researcher Workbench";
```
jaycarlton: nit/aside: Is there an easy way to externalize these kinds of constants (like a `.properties file`), in case we or other OSS users want to use some of this code for another application, or somebody who's not a programmer needs to make changes?

calbach: I don't know of an easy way, but I also don't think I agree there's merit to doing that if we easily could. It increases complexity and we don't currently have the OSS use case you're describing.

PR #2533, opened 2019-08-30
[RW-3286] [risk=low] Revamp hamburger menu

als364:


ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -86,6 +86,7 @@ export const WorkspaceLibrary = withUserProfile()
 
   async componentDidMount() {
     this.loadPublishedWorkspaces();
```
als364: This is kind of bad. This function is what populates the `workspaceList`, which is what we use to determine whether or not we want to display anything in the library at all.

I could rename the function to `loadWorkspaceList` because that is what it actually does. 

We should probably also call this to refresh the `workspaceList` whenever we switch tabs instead of just whenever we load the Published Workspaces tab. The Featured Workspaces tab just filters down the Published Workspaces list by what we have decided to feature.

als364:
Before: 
all-of-us-workbench-test appspot com_

After:
localhost_4200_

ui/src/app/components/breadcrumb.tsx
```diff
@@ -1,20 +1,27 @@
...
+import {WorkspaceData} from 'app/utils/workspace-data';
+import {Cohort, ConceptSet} from 'generated/fetch';
 
 const styles = {
+  breadcrumb: {
+    marginLeft: '3.25rem',
+    display: 'inline-block',
+  },
+  primary: {
+    color: colors.primary
```
NehaBroad: not sure if this needs to be extracted since its just 1 property and that too call to colors class, normally we extract styles having more than 3 properties

ui/src/app/components/breadcrumb.tsx
```diff
@@ -27,75 +34,99 @@ const styles = {
...
-        {label: 'Dataset', url: `${prefix}/data/datasets`}
+        ...getTrail(BreadcrumbType.Data, workspace, cohort, conceptSet, urlParams),
+        new BreadcrumbData('Dataset', `${prefix}/data/datasets`)
       ];
     case BreadcrumbType.Data:
       return [
-        ...getTrail(BreadcrumbType.Workspaces, data),
-        {label: workspace ? workspace.name : '...', url: `${prefix}/data`}
+        ...getTrail(BreadcrumbType.Workspaces, workspace, cohort, conceptSet, urlParams),
+        new BreadcrumbData(workspace ? workspace.name : '...', `${prefix}/data`)
```
NehaBroad: How about creating a map with key as BreadCrumbType and values {label, url} that ways this switch case, can be easily replaced by .
..new BreadCrumbData(mapname.get(BreadCrumbtype).label,mapname.get(BreadCrumbtype).url)
Also that will help in case there is a new breadcrumbType too

ui/src/app/components/breadcrumb.tsx
```diff
@@ -111,42 +142,60 @@ const BreadcrumbLink = ({href, ...props}) => {
...
 
-@Component({
-  selector: 'app-breadcrumb',
-  template: '<div #root></div>'
-})
-export class BreadcrumbComponent extends ReactWrapperBase {
-  constructor() {
-    super(Breadcrumb, []);
+    first(): Array<BreadcrumbData> {
+      return fp.dropRight(1, this.trail());
```
NehaBroad: i like this, it would be nice to have it in a util file that takes in an array and returns first last. (right now its index.tsx file that has other array functions maybe we can add to that or create a new util file)

ui/src/app/components/side-nav.tsx
```diff
@@ -0,0 +1,333 @@
...
+    width: '100%',
+    margin: 0,
+    paddingLeft: '1rem',
+    textAlign: 'left',
+    textTransform: 'none',
+    height: '2rem',
+    color: colors.white,
+  },
+  sideNavItemActive: {
+    backgroundColor: '#4356a7',
```
NehaBroad: lets check with Lou we should use variation of color mentioned in colors  and use colorWithWhiteness or add a new color in color file

ui/src/app/pages/signed-in/signed-in-nav-bar.tsx
```diff
@@ -0,0 +1,223 @@
...
+        </div>
+        <div>
+          <a href={'/'}>
+            <img
+              src={this.props.headerImg}
+              style={styles.headerImage}
+            />
+          </a>
+          {
+            this.props.shouldShowDisplayTag
```
NehaBroad: this could be like {this.props.shouldShowDisplayTag && <div style={styles.displayTag}>...</div>}

ui/src/app/components/breadcrumb.tsx
als364: Because the labels / urls aren't 1:1 with only the `BreadcrumbType`, this map would end up doing basically the same thing as the switch statement.

NehaBroad:
The "User Admin" menu option for user with Authority.ACCESSCONTROLADMIN is missing (in mocks as well) had a talk with karthik it should be part of the hamburger menu option

ui/src/app/components/breadcrumb.tsx
NehaBroad: But since its variable in colors file, its already adding anything

ui/src/app/components/breadcrumb.tsx
```diff
@@ -27,75 +34,95 @@ const styles = {
...
     case BreadcrumbType.Workspaces:
       return [
-        {label: 'Workspaces', url: '/workspaces'}
+        new BreadcrumbData('Workspaces', '/workspaces')
       ];
     case BreadcrumbType.Workspace:
       return [
-        ...getTrail(BreadcrumbType.Workspaces, data),
-        {label: workspace ? workspace.name : '...', url: `${prefix}/data`}
+        ...getTrail(BreadcrumbType.Workspaces, workspace, cohort, conceptSet, urlParams),
```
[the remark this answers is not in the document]
als364: What do you mean?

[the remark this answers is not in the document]
als364: I've tried a few different things, including leaving the args as the original 'data' (compile time errors), changing it to a spread parameter (runtime errors), and defining consts outside the scope of this function that call this function (only slightly neater)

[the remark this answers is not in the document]
jaycarlton: Sorry, that wasn't clear at all. Not worth worrying about. I was wondering if this might be a place to try function [currying](https://wsvincent.com/javascript-currying/), but that would almost just be showing off. So a local function would capture the (local) arguments that are the same between all the calls, leaving only a one-arg function to operate on the remaining ones. No need to worry about it.

jaycarlton:
I like the pictures. Much vamping.

jaycarlton:
re: admin menu, good catch. We may want multiple admin pages, if that impacts the design. I think Cluster Admin is the one I'm going to add for shutting down nefarious users' clusters.

PR #2534, opened 2019-08-30
[risk=no] fixing BQ test cases to run inside Intellij

freemabd:
So it seems that the least intrusive way to solve this problem was to override BigQueryService#getBigQueryService().

### DIFF
--- a/api/src/bigquerytest/java/org/pmiops/workbench/api/BigQueryServiceTest.java
+++ b/api/src/bigquerytest/java/org/pmiops/workbench/api/BigQueryServiceTest.java
@@ -1,0 +1,17 @@
+package org.pmiops.workbench.api;
+
+import com.google.cloud.bigquery.BigQuery;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * This sole purpose of this class is to override the setting of projectId on the BigQuery service
+ * instance so test cases are runnable inside IntelliJ.
+ */
+public class BigQueryServiceTest extends BigQueryService {
+
+  @Autowired private BigQuery bigquery;
+
+  protected BigQuery getBigQueryService() {
+    return bigquery;
+  }
+}
--- a/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
+++ b/api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
@@ -62,7 +62,7 @@ import org.springframework.jdbc.core.JdbcTemplate;
 
 @RunWith(BeforeAfterSpringTestRunner.class)
 @Import({
-  BigQueryServiceTest.class,
+  BigQueryTestService.class,
   CloudStorageServiceImpl.class,
   CohortQueryBuilder.class,
   SearchGroupItemQueryBuilder.class,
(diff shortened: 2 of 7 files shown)


api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -39,7 +39,11 @@
 
   @Autowired private Provider<WorkbenchConfig> workbenchConfigProvider;
 
-  private BigQuery getBigQueryService() {
+  /**
+   * Changed access modifier to protected so that BigQueryServiceTest.java can override this method.
```
gjuggler: Wording nitpick: it's better if the comment states the way things are rather than tries to describe a change (since a reader doesn't easily know what it was changed from). I'd go with

"This method is protected, rather than private, so that..."

api/src/bigquerytest/java/org/pmiops/workbench/api/BigQueryServiceTest.java
```diff
@@ -0,0 +1,17 @@
+package org.pmiops.workbench.api;
+
+import com.google.cloud.bigquery.BigQuery;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * This sole purpose of this class is to override the setting of projectId on the BigQuery service
+ * instance so test cases are runnable inside IntelliJ.
+ */
+public class BigQueryServiceTest extends BigQueryService {
```
gjuggler: Naming nitpick: maybe BigQueryTestService would sound less like a test class?

api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
```diff
@@ -62,7 +62,7 @@
 
 @RunWith(BeforeAfterSpringTestRunner.class)
 @Import({
-  BigQueryService.class,
+  BigQueryServiceTest.class,
```
gjuggler: Just to clarify my understanding with what's going on here... this @Import list is what causes these classes to be autowired into the test case, so importing a subclass will cause BigQueryServiceTest to be injected anywhere that a BigQueryService is required in the execution test?

api/src/bigquerytest/java/org/pmiops/workbench/api/BigQueryServiceTest.java
freemabd: Done.

api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
freemabd: This is correct. We are explicitly loading only the classes we are interested in.

api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
freemabd: Done.

api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -39,7 +39,11 @@
 
   @Autowired private Provider<WorkbenchConfig> workbenchConfigProvider;
 
-  private BigQuery getBigQueryService() {
+  /**
+   * This method is protected, rather than private, so that it can be overridden for testing
```
jaycarlton: I'd replace this comment with `@VisibleForTesting`.

freemabd: I just replaced it. Thanks, Jay.

PR #2537, opened 2019-09-03
[RW-3341][risk=low] Inactivity Logout

ericsong:
See https://docs.google.com/document/d/1gBB8ZHHrL_f8FnqVE3_WUZzIWIk3Z2dMxPx3b4SAIB4/edit?ts=5d6927d2#heading=h.b5vt06vb56bp

for tech design.

**High level overview**
Several javascript events will be tracked to detect if a user is active or not. There are three sources of user activity events, the AoU web frame, read only notebook iframe, and leonardo iframe. All three will post event messages to the AoU web frame where a message handler will aggregate the events to determine overall user activity.

The leonardo iframe requires a notebook extension because the javascript running in the iframe must originate from the same source as the leonardo iframe.

There are two environment specified timeout windows. One which determines the length of time which must pass before a user is logged out and another which determines when the inactivity warning popup will be displayed.

Web API's localStorage is used to keep track of last known user activity in the case of the AoU application being closed.

Upon opening the AoU app (regardless of logged in status), the application will check if the time between now and the last active timestamp stored in localStorage is greater than the inactivity window. If so, the user will be sent to the Google logout page. Before being signed out, the application will mark the current time as the last active timestamp. This is needed to prevent an endless login loop.


ui/src/environments/environment.local.ts
```diff
@@ -14,5 +14,7 @@ export const environment: Environment = {
   gaUserAgentDimension: 'dimension1',
   zendeskHelpCenterUrl: 'http://aousupporthelp.zendesk.com/hc',
   shibbolethUrl: 'http://mock-nih.dev.test.firecloud.org',
+  inactivityTimeoutInSeconds: 20,
```
calbach: reminder to revert before merging

ui/src/app/utils/index.tsx
```diff
@@ -353,3 +353,25 @@ export const toggleIncludes = fp.curry((value, arr) => {
...
+      action();
+    }
+  }, sensitivityMs);
+
+  return () => {
+    t = Date.now();
+  }
+}
+
+export function resettableTimeout(f, timeoutInSeconds) {
```
calbach: Docs? The purpose of this function is not obvious to me from the name alone

ui/src/app/utils/index.tsx
```diff
@@ -353,3 +353,25 @@ export const toggleIncludes = fp.curry((value, arr) => {
 export function sliceByHalfLength(obj) {
   return Math.ceil(obj.length / 2);
 }
+
+export function debouncer(action, sensitivityMs) {
```
calbach: This could also use some documentation, the concept of "debouncing" is fairly common but I don't think the details can be assumed

ui/src/app/pages/signed-in/component.ts
```diff
@@ -128,9 +177,20 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
...
+  closeInactivityModal(): void {
+    this.showInactivityModal = false;
+  }
+
+  get inactivityModalText(): string {
+    const timeText = environment.inactivityWarningInSecondsBefore % 60 == 0 && environment.inactivityWarningInSecondsBefore > 60 ?
+      `${environment.inactivityWarningInSecondsBefore / 60} minutes` :
+      `${environment.inactivityWarningInSecondsBefore} seconds`;
+
+    return `You've been idle for sometime. You will be logged out in ${timeText} if no activity is detected.`;
```
calbach: `sometime` -> `some time`

calbach: I'm used to the UX here being clicking an "I'm here" button on the modal. I think this is preferred, otherwise the modal might flash in and disappear as soon as the user moves their mouse, which would be frustrating.

ui/src/app/pages/signed-in/component.ts
```diff
@@ -128,9 +177,20 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
   private navigateSignOut(): void {
     // Force a hard browser reload here. We want to ensure that no local state
     // is persisting across user sessions, as this can lead to subtle bugs.
-    window.location.assign('/');
+    window.location.assign('https://accounts.google.com/logout');
```
calbach: Per slack, I would use this redirect approach to send them back to the AoU login page: https://stackoverflow.com/questions/4202161/google-account-logout-and-redirect

ui/src/environments/environment-type.ts
```diff
@@ -41,6 +41,8 @@ export interface Environment {
   // Example value: https://aousupporthelp.zendesk.com/hc/
   zendeskHelpCenterUrl: string;
 
+  inactivityTimeoutInSeconds: number;
```
calbach: opt: personally I prefer to just directly concatenate the units on the normal variable name, rather than using "in" e.g. inactivityTimeoutSeconds

ui/src/app/pages/app/component.ts
```diff
@@ -64,6 +65,12 @@ export class AppComponent implements OnInit {
         };
         console.log('To override the API URLs, try:\n' +
           'setAllOfUsApiUrl(\'https://host.example.com:1234\')');
+
+        const lastActive = window.localStorage.getItem(INACTIVITY_CONFIG.LOCAL_STORAGE_KEY_LAST_ACTIVE);
+        if (lastActive == null || Date.now() - parseInt(lastActive) > environment.inactivityTimeoutInSeconds * 1000) {
```
calbach: Is there anything to prevent them from being forced to logout on their first sign in? Will we set the local storage value during the account creation flow?

ui/src/app/pages/signed-in/component.ts
```diff
@@ -103,6 +112,42 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    const signalUserActivity = debouncer(() => {
+      window.postMessage(INACTIVITY_CONFIG.MESSAGE_KEY, '*');
+    }, 1000);
+
+    INACTIVITY_CONFIG.TRACKED_EVENTS.forEach(eventName => {
+      window.addEventListener(eventName, () => signalUserActivity(), false);
+    });
+  }
+
+  startInactivityTimers() {
```
calbach: Could use a comment somewhere around here as to what the rules are of the local storage data, e.g. what it means if the data is unset and how that's handled

ui/src/app/pages/signed-in/component.ts
```diff
@@ -103,6 +112,42 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    const resetLogoutTimeout = resettableTimeout(() => {
+      this.signOut();
+    }, environment.inactivityTimeoutInSeconds * 1000);
+
+    const resetInactivityModalTimeout = resettableTimeout(() => {
+      this.showInactivityModal = true;
+    }, (environment.inactivityTimeoutInSeconds - environment.inactivityWarningInSecondsBefore) * 1000);
+
+    localStorage.setItem(INACTIVITY_CONFIG.LOCAL_STORAGE_KEY_LAST_ACTIVE, Date.now().toString());
```
calbach: One possible edge case here: there's nothing tying this local storage value to the current user login. Should we be somehow incorporating that into the key? This would only matter in a multilogin / shared machine situation

e.g. I could flip through multiple logins without having to reenter my password.

ui/src/app/pages/app/component.ts
```diff
@@ -66,11 +68,18 @@ export class AppComponent implements OnInit {
...
-        }
+        this.signInService.isSignedIn$.subscribe(signedIn => {
+          if (signedIn) {
+            const lastActive = window.localStorage
+              .getItem(INACTIVITY_CONFIG.LOCAL_STORAGE_KEY_LAST_ACTIVE);
+            if (lastActive == null ||
+              Date.now() - parseInt(lastActive, 10) > environment.inactivityTimeoutSeconds * 1000) {
+              localStorage
+                .setItem(INACTIVITY_CONFIG.LOCAL_STORAGE_KEY_LAST_ACTIVE, Date.now().toString());
+              SignedInComponent.prototype.navigateSignOut();
```
calbach: This is pretty sketchy and the dependency is questionable, would just move this into the navigation utils as a static helper function

ui/src/app/pages/signed-in/component.ts
calbach: Good point

ui/src/app/pages/signed-in/component.ts
```diff
@@ -42,9 +59,14 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
   // True if the user tried to open the Zendesk support widget and an error
   // occurred.
   zendeskLoadError = false;
+  showInactivityModal = false;
   private profileLoadingSub: Subscription;
   private subscriptions = [];
 
+  private getUserActivityTimer;
```
calbach: nit: types?

common-ui/tslint.json
```diff
@@ -25,7 +25,7 @@
     "max-line-length": [
       true,
       {
-        "limit": 100,
+        "limit": 140,
```
calbach: Was this change intentional?

ericsong: yeah. We had discussed in slack before and most people seemed on board with increasing the limit.

I asked around the office real quick and it seems like opinions haven't changed.

PR #2538, opened 2019-09-03
[RW-3375]Workspaces with concept sets cannot be duplicated

NehaBroad:
Concept set constructor being called at the start of method 'cloneConceptSetAndConceptIds' was already taking care of concepts for new concept set, hence the extra call to bulkcopy was causing sql constraint exceptions

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/conceptset/ConceptSetService.java
+++ b/api/src/main/java/org/pmiops/workbench/conceptset/ConceptSetService.java
@@ -1,5 +1,6 @@
 package org.pmiops.workbench.conceptset;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/conceptset/ConceptSetService.java
```diff
@@ -34,7 +34,6 @@ public ConceptSet cloneConceptSetAndConceptIds(
     c.setCreationTime(targetWorkspace.getCreationTime());
     c.setVersion(1);
     ConceptSet saved = conceptSetDao.save(c);
-    conceptSetDao.bulkCopyConceptIds(conceptSet.getConceptSetId(), saved.getConceptSetId());
```
calbach: Please add a test that fails without this change

PR #2539, opened 2019-09-03
[No Ticket][Risk=no] Remove deploy-api ruby endpoint

blrubenstein:

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1540,20 +1540,11 @@ def deploy_app(cmd_name, args, with_cron, with_gsuite_admin)
     %W{--project #{gcc.project} #{promote}} +
     (op.opts.quiet ? %W{--quiet} : []) +
     (op.opts.version ? %W{--version #{op.opts.version}} : []))
-end
 
-def deploy_api(cmd_name, args)
-  ensure_docker cmd_name, args
-  common = Common.new
-  common.status "Deploying api..."
-  deploy_app(cmd_name, args, true, true)
+  with_cloud_proxy_and_db_env(cmd_name, args) do |ctx|
+    load_config(ctx.project)
+  end
 end
-
-Common.register_command({
-  :invocation => "deploy-api",
-  :description => "Deploys the API server to the specified cloud project.",
-  :fn => ->(*args) { deploy_api("deploy-api", args) }
-})
 
 def create_workbench_db()
   run_with_redirects(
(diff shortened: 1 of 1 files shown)


calbach:
Why?

gjuggler:
> Why?

Sorry, I should have asked for more context in the description. Scout ran into an issue where they ran deploy-api to update the perf environment, assuming it would deploy the API and config changes. But it only deploys the API code and doesn't do the other stuff (update configs, CDR versions, migrations, etc.).

We discussed a bit offline, and I agreed that it was somewhat misleading to have deploy-api separated out as a distinct command-line tool if it only half-deploys an environment.

One might argue that it does have a valid use case: deploying API-only changes from a dev branch for others to be able to test. But it involves understanding the limitations (e.g. any config changes won't be included).

My instinct (which may be wrong!) was that deploy-ui is more frequently used, and generally less reliant on config or API related changes, so removing deploy-api would be a net win for simplicity.

If there's a strong desire to keep this around, I'd like to see at least some sort of documentation be added to clarify that "deploy" should be preferred when running a full-environment deploy, while deploy-api and deploy-ui are useful for dev-only deploys, with known limitations.

calbach:
Hmm... I'm a bit on the fence. deploy-ui is much more commonly used, but I've also used `deploy-api` somewhat frequently (every 1.5 months?). The reason `deploy` is not a great substitute is that it's not very hermetic, e.g. it will apply any random database upgrades, or new config values you might have on your local branch for everyone on the test environment, thereby likely breaking test temporarily or putting the database in a state which needs to be manually resolved.

Renaming `deploy-api` --> `deploy-app-engine` might make this clearer.

blrubenstein:
I would be happy to name it: `deploy-api-without-db-config` or something? I think it would be fine if it was explicit and clear (and it was similarily clear what command to run if you `do` want to update db and config.)

calbach:
I'd rather not list everything it's **not** doing in the command name..

How about:
`deploy-full`
`deploy-app-engine`
`deploy-server-config`
`deploy-gcs-artifacts`
`deploy-cdr-versions`

Or:
`deploy`
`update-app-engine`
`update-config`
`update-gcs-artifacts`
`update-cdr-versions`

blrubenstein:
We shouldn't have any running it as a commandline run, just as a ruby endpoint.

I like `update-app-engine`

blrubenstein:
I'm going to close this for now.

PR #2540, opened 2019-09-03
[RW-3366][risk=low] Switch to new VPC-SC-enabled synthetic test CDR project

calbach:
Drop the "etic", it's cleaner.

This is more similarly configured to the other projects, and is named consistently with our staging/stable names (which were mistakenly named this way).


calbach:
Manually tested locally

PR #2544, opened 2019-09-04
[RW-3400][Risk=no] Fix the text to say policy reminder and update tex…

blrubenstein:
…t color


blrubenstein:
New: 
Screen Shot 2019-09-04 at 11 03 19 AM

blrubenstein:
Old: 
Screen Shot 2019-09-04 at 11 03 39 AM

PR #2551, opened 2019-09-05
[RW-2517][RISK=NO]Create Account page

NehaBroad:
This is First step of new create account. The next step will be part of story RW-3284 | Create Account - User survey pagewhich will have the submit/save account information functionality, Because of this reason i have set the environment variable 'enableAccountPages'  set to false even on local/test
Screen Shot 2019-09-05 at 5 59 52 PM
Screen Shot 2019-09-05 at 6 00 00 PM
Screen Shot 2019-09-05 at 6 00 07 PM


ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+
+// The values will change once we have ENUM as part of DB work as part of create account 2
+const Options = {
+  roles: [
+    {label: `Undergraduate (Bachelor level) student`, value: 'bachelor'},
+    {label: `Graduate trainee (Current student in a Masters, PhD, or Medical school training
+        program)`, value: 'trainee'},
+    {label: `Research fellow (a post-doctoral fellow or medical resident in training)`,
+      value: 'research'},
+    {label: `Early career tenure-track researcher`, value: 'earylCarrer'},
```
blrubenstein: Nit: Spelling on value.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+    {label: `Non tenure-track researcher`, value: 'nontenure'},
+    {label: `Mid-career tenured researcher`, value: 'midCarrer'},
```
blrubenstein: Nit: Spelling (should be `midCareer`), also line below

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+import * as fp from 'lodash/fp';
+import {Dropdown} from 'primereact/dropdown';
+import * as React from 'react';
+import * as validate from 'validate.js';
+
+function isBlank(s: string) {
+  return (!s || /^\s*$/.test(s));
+}
+
+// The values will change once we have ENUM as part of DB work as part of create account 2
```
blrubenstein: :+1: to this

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+  ]
+};
+
+// The following will be part of Profile object once create account survey is in
+export interface Address {
+  streetAddress1: string;
+  streetAddress2: string;
+  city: string;
+  state: string;
+  zipcode: string;
```
blrubenstein: Should the zipcode be a number? Functionally it seems like number might be right.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+export interface AccountCreationProps {
+  invitationKey: string;
+  setProfile: Function;
+}
+
+export interface AccountCreationState {
+  creatingAccount: boolean;
+  errors: any;
+  invalidEmail: boolean;
+  rolesOptions: any;
```
blrubenstein: I think this could be more descriptive than `any`. Like maybe: `Option[]`? (and then create an Option interface?)

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+      .then((savedProfile) => {
+        this.setState({profile: savedProfile, creatingAccount: false});
+        setProfile(savedProfile); })
+      .catch(error => {
+        console.log(error);
+        this.setState({creatingAccount: false});
+      });
+  }
+
+  usernameValid(): boolean {
```
blrubenstein: Could this be a `get` method?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+    if (isBlank(this.state.profile.username) || this.state.usernameCheckInProgress) {
+      return undefined;
+    }
+    return !this.isUsernameValidationError();
+  }
+
+  isUsernameValidationError(): boolean {
```
blrubenstein: Could this be a `get` method?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+    return (this.state.usernameConflictError || this.usernameInvalidError());
+  }
+
+  usernameInvalidError(): boolean {
```
blrubenstein: Could this be a `get` method?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+    } else if (affiliation === 'freeText') {
+      this.setState({showAffiliationOther: true});
+      return;
+    }
+    this.selectAffiliationRoles(this.state.affiliationRole);
+  }
+
+  selectAffiliationRoles(role) {
+    this.setState({affiliationRole: role});
+    if (role === 'freeText') {
```
blrubenstein: Opt Nit: Could you put this above and have it only issue one call to setState?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+      [profile.givenName, profile.familyName, profile.username, profile.contactEmail,
+        profile.currentPosition, profile.organization, profile.areaOfResearch];
+    if (requiredFields.some(isBlank)) {
+      this.setState({showAllFieldsRequiredError: true});
+      return;
+    }
+  }
+
+  validateAccountCreation() {
+    const {profile: {
```
blrubenstein: Nit: For spacing, do you mind putting `profile` on a newline? This is sorta weirdly spaced with everything else in the object.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+      givenName: {
+        presence: {allowEmpty: false}
+      },
+      familyName: {
+        presence: {allowEmpty: false}
+      },
+      contactEmail: {
+        presence: {allowEmpty: false},
+        format: {
+          pattern: /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/,
```
blrubenstein: Could you use the same regex pattern on line 190 for email validation?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+            {summarizeErrors(this.state.errors)}
+          </ValidationError>
+        </div>}
+        <Section header='Create an All of Us username'>
+          <div>
+            <TextInput id='username' name='username' placeholder='New Username'
+                       value={username}
+                       onChange={v => this.usernameChanged(v)}
+                       invalid={this.state.usernameConflictError || this.usernameInvalidError()}
+                       style={{width: '12rem', marginRight: '0.5rem'}}/>
```
blrubenstein: I'm seeing `width: '12rem'` repeated a bunch. Can we parameterize that?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+            </div>
+            <TooltipTrigger content={<div>Usernames can contain only letters (a-z),
+              numbers (0-9), dashes (-), underscores (_), apostrophes ('), and periods (.)
+              (minimum of 3 characters and maximum of 64 characters).<br/>Usernames cannot
+              begin or end with a period (.) and may not contain more than one period (.) in a row.
+            </div>}>
+              <InfoIcon style={{'height': '22px', 'paddingLeft': '2px'}}/>
+            </TooltipTrigger>
+            {this.state.usernameConflictError &&
+            <div style={{height: '1.5rem'}}>
```
blrubenstein: I'm seeing this height repeated for most errors. Can we parameterize that?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+            </Error></div>}
+          </div>
+        </Section>
+        <Section header='About you'>
+          <div style={{display: 'flex', flexDirection: 'column'}}>
+            <div style={{paddingBottom: '0.5rem'}}>
+              <TextInput id='givenName' name='givenName' autoFocus
+                         placeholder='First Name'
+                         value={givenName}
+                         invalid={givenName.length > 80}
```
blrubenstein: Can we parameterize 80 as something like `nameLengthLimit`?

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,665 @@
...
+                         style={{width: '12rem', marginRight: '2rem'}}
+                         onChange={v => this.updateProfile('givenName', v)}/>
+              {givenName.length > 80 &&
+              <ErrorMessage id='givenNameError'>
+                First Name must be 80 characters or less.
```
blrubenstein: If we parameterize the name length limit, these error messages can also read text from that.

ui/src/app/pages/login/account-creation/account-creation.tsx
NehaBroad: Its the same regex, the other is going to be deleted once we have the enable flag set to true

ui/src/app/pages/login/account-creation/account-creation.tsx
NehaBroad: For zipcode i was planning to use this as string so as to not loose the leading zeros

ui/src/app/pages/login/account-creation/account-creation.tsx
blrubenstein: I think you could change 80 to `{{nameLength}}` (and below as well)

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -439,7 +439,7 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
                          onChange={v => this.updateProfile('givenName', v)}/>
               {givenName.length > nameLength &&
               <ErrorMessage id='givenNameError'>
-                First Name must be 80 characters or less.
+                First Name must be {nameLength} characters or less.
```
jaycarlton: nit: this should be fewer rather than less since it's countable in a reasonable period of time.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,668 @@
...
+
+import colors from 'app/styles/colors';
+import {summarizeErrors} from 'app/utils/index';
+import {environment} from 'environments/environment';
+import * as fp from 'lodash/fp';
+import {Dropdown} from 'primereact/dropdown';
+import * as React from 'react';
+import * as validate from 'validate.js';
+
+function isBlank(s: string) {
```
jaycarlton: is there a builtin or library function for this? If we need to roll our own, then we need to have unit tests I think.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,668 @@
...
+  return (!s || /^\s*$/.test(s));
+}
+
+// The values will change once we have ENUM as part of DB work as part of create account 2
+const Options = {
+  roles: [
```
jaycarlton: Can we externalize this into a JSON file for roleConfig or something?

ui/src/app/pages/login/account-creation/account-creation.tsx
jaycarlton: nit: if this is for international addresses as well, we need to be careful. One (minor) point is that I think the US is the only one that calls it a zip code. The bigger problem is that we need a corpus of test cases to make sure we're representing them fully. Again, it seems like there would be an off-the-shelf component for working with addresses.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,668 @@
...
+      affiliationOther: ''
+    };
+  }
+
+  // This method will be modified once the story  below will be done
+  // https://precisionmedicineinitiative.atlassian.net/browse/RW-3284
+  createAccount(): void {
+    const {invitationKey, setProfile} = this.props;
+    const profile = this.state.profile;
+    const emailValidRegex = new RegExp(/^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/);
```
jaycarlton: Please use a library to validate email addresses. There are tons of edge cases. For example, it looks like this won't support the plus character, which gmail uses for lightweight aliases.

We need lots of unit tests too.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,668 @@
...
+    this.setState({showAllFieldsRequiredError: false});
+    this.setState({invalidEmail: false});
+    const requiredFields =
+      [profile.givenName, profile.familyName, profile.username, profile.contactEmail,
```
jaycarlton: Not everybody has two names. There's a classic post about this at https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/.

Likewise, for citizen scientists, they might not have a current position, org, or research area. [this may be a future issue].

Sorry with hitting you with product concerns in a PR.

ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -0,0 +1,668 @@
...
+        username: '',
+        dataAccessLevel: DataAccessLevel.Unregistered,
+        givenName: '',
+        familyName: '',
+        contactEmail: '',
+        currentPosition: '',
+        organization: '',
+        areaOfResearch: '',
+      },
+      usernameCheckInProgress: false,
```
jaycarlton: I believe this state could be tracked in a more object-oriented way, for example by having a little component for each piece of data we're taking in and having it report errors (and be able to define what its own error state looks like). That way we could make them smarter later (e.g. allowing incremental feedback while you're typing, for characters remaining, forbidden characters, etc.). The way it is now, there's no real separation among all the input widgets' state, and the maintainer has to keep track fo whether each one is checked/flagged. 

This is just a quick look though; you could have reasons it needs to be centralize like this too.

jaycarlton:
@NehaBroad I don't really understand the state we're in now or why we can't use a feature branch or something, but go ahead if these files aren't new. I need to pay more attention.

blrubenstein:
@jaycarlton we have tended to avoid feature branches, since they can become stale quickly, and often lead to several awful rebases, in favor of smaller PRs behind flags. It does make unit testing hard though.

jaycarlton:
@blrubenstein OK. I don't know much about how the feature flags work, but at other places where we've used them, unit test coverage was one of the things we relied on before turning them on. Is there a way to mock the feature flag for testing purposes, or to un-flag a feature to only the test account? Sorry for thread-jacking.

gjuggler:
> @jaycarlton we have tended to avoid feature branches, since they can become stale quickly, and often lead to several awful rebases, in favor of smaller PRs behind flags. It does make unit testing hard though.

If we can't write unit tests for code that's behind a feature flag, this is an important bug / friction point in our infrastructure. I had a review yesterday with @als364 where we ran into the same thing. I filed https://precisionmedicineinitiative.atlassian.net/browse/RW-3457 to track this as tech debt.

PR #2554, opened 2019-09-06
[RW-3418][Risk=low] Remove DB fields for free tier billing

blrubenstein:


ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -360,7 +360,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
         workspace: {
           name: '',
           dataAccessLevel: DataAccessLevel.Registered,
-          namespace: userProfileStore.getValue().profile.freeTierBillingProjectName,
+          namespace: '',
```
jaycarlton: Should there not still be a namespace for this one? Or is it just never needed?

How did we decide to use empty strings instead of null?

blrubenstein: It is automatically filled, but I believe required. I can try it as null! /shrug

blrubenstein: Actually, only name is required! I'll get rid of it.

api/src/main/java/org/pmiops/workbench/api/AuditController.java
```diff
@@ -127,12 +124,6 @@ private static String auditSql(String cdrProjectId, List<String> tableSuffixes)
                       + "project jobUser",
                   cdrProjectId, total, email));
           numBad += total;
-        } else if (!whitelist.contains(project_id)) {
```
blrubenstein: @calbach this is the one thing I'm not confident with. We don't have whitelist anymore cause we don't have user projects. Is there another way to do this, or is it good to remove?

blrubenstein: Should I just get rid of the conditions and log all bad actors here?

api/db/changelog/db.changelog-96-drop-free-tier-billing.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="srubenst" id="db.changelog-86-drop-tos">
```
calbach: nit: update id

api/src/main/java/org/pmiops/workbench/api/AuditController.java
calbach: The equivalent replacement here would be to query the workspace dao for all workspace namespaces

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -242,17 +237,6 @@ export const Homepage = withUserProfile()(class extends React.Component<
     }));
   }
 
-  checkBillingProjectStatus() {
-    const {profileState: {profile, reload}} = this.props;
-    if (profile.freeTierBillingProjectStatus === BillingProjectStatus.Ready) {
-      this.setState({billingProjectInitialized: true});
```
calbach: This variable should no longer be used or referenced, rm

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceDao.java
```diff
@@ -24,6 +25,9 @@ Workspace findByWorkspaceNamespaceAndFirecloudNameAndActiveStatus(
   Workspace findByWorkspaceNamespaceAndNameAndActiveStatus(
       String workspaceNamespace, String name, short activeStatus);
 
+  @Query("SELECT distinct w.workspaceNamespace, w from Workspace w")
```
jaycarlton: nit: make DISTINCT and FROM all caps. Do we have a SQL linter?

jaycarlton: I couldn't get this query to work locally. Did you mean `workspace_namespace`. I don't think you need to make `w` an alias in this case. This worked for me
`SELECT DISTINCT workspace_namespace FROM workspace ORDER BY NULL`.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -409,10 +408,9 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
         if (this.isMode(WorkspaceEditMode.Duplicate)) {
           this.setState({workspace: {
             ...this.props.workspace,
-            // These are the only fields which are not automatically handled/differentiated
+            // This is the only field which is not automatically handled/differentiated
             // on the API level.
-            name: 'Duplicate of ' + this.props.workspace.name,
-            namespace: userProfileStore.getValue().profile.freeTierBillingProjectName
+            name: 'Duplicate of ' + this.props.workspace.name
```
jaycarlton: I don't think we want to do this. I guess the "duplicate of" stuff was there before, but I think you wan to kill this property completely and fixup any places using the duplicated one.

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceDao.java
```diff
@@ -24,6 +25,9 @@ Workspace findByWorkspaceNamespaceAndFirecloudNameAndActiveStatus(
...
+  Set<String> findAllWorkspaceNamespaces();
+
   @Query(
```
jaycarlton: this looks like it's using Java-ified column names like `workspaceNamespace` and `firecloudName` instead of the versions with the underscores.

`select workspace_namespace, firecloud_name, active_status from workspace LIMIT 3;` works for me. Were the columns renamed at some point?

calbach: Unless it says `native=true` in the query annotation, these are JPQL, not MySQL statements. https://docs.spring.io/spring-data/jpa/docs/current/reference/html/

jaycarlton: Thank you, I thought I was losing my mind there for a minute.

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceDao.java
jaycarlton: It's not MySQL, which makes sense.

calbach:
Snyk error looks unrelated - just an artifact of this PR not being rebased I believe.

PR #2555, opened 2019-09-06
[RW-2955][risk=no] Drop leo welder flag

calbach:
- Removes now unused aou playground mode extension
- Fix activity extension key name (this likely overwrote policy reminders @NehaBroad @ericsong )
- Whitelist some more keys from our configs


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -230,36 +225,23 @@
     String playgroundDir = "workspaces_playground/" + workspacePath;
```
jmthibault79: remove the "must be kept in sync" comment just above here

calbach: Done - nice catch.

PR #2556, opened 2019-09-06
[RW-3093][risk=no] Drop unused enforceRegistered flag

calbach:


.gitallowed
```diff
@@ -7,7 +7,11 @@
 657299777109-kvb5qafr70bl01i6bnpgsiq5nt6v1o8u.apps.googleusercontent.com
 56507752110-ovdus1lkreopsfhlovejvfgmsosveda6.apps.googleusercontent.com
 684273740878-d7i68in5d9hqr6n9mfvrdh53snekp79f.apps.googleusercontent.com
+63939010390-aj0r8hro7r8lkt7a45gissu3m73ietl2.apps.googleusercontent.com
```
gjuggler: Are the changes in this file related?

calbach: Sort of. The newest version of git-secrets is pickier and validates keys that look like this. Whenever I touch a config file it complains about any keys there. Handled in another PR though:
https://github.com/all-of-us/workbench/pull/2555, no longer in this one.

PR #2558, opened 2019-09-09
[no risk, no ticket]Remove enableJupyterLab feature flag (always false) and jupyterLabMode

jmthibault79:

### DIFF
--- a/ui/src/environments/environment-type.ts
+++ b/ui/src/environments/environment-type.ts
@@ -43,10 +43,6 @@ export interface Environment {
 
   inactivityTimeoutSeconds: number;
   inactivityWarningBeforeSeconds: number;
-  // Transient client-side flags.
+  // Add transient client-side flags below here
   //
-  // Whether users should be able to use JupyterLab as an option.
-  // See RW-1966 for details.
-  // Exit criteria: remove flag for Athens release.
-  enableJupyterLab: boolean;
 }
--- a/ui/src/environments/environment.prod.ts
+++ b/ui/src/environments/environment.prod.ts
@@ -15,9 +15,9 @@ export const environment: Environment = {
   trainingUrl: 'https://aou.nnlm.gov',
   inactivityTimeoutSeconds: 30 * 60,
   inactivityWarningBeforeSeconds: 5 * 60,
+  // Add transient client-side flags below here
   // Use care when changing these flags in prod!
   //
   // See environment-type.ts for more details on transient flags, including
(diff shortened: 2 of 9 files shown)


ui/src/environments/environment-type.ts
```diff
@@ -43,10 +43,6 @@ export interface Environment {
 
   inactivityTimeoutSeconds: number;
   inactivityWarningBeforeSeconds: number;
-  // Transient client-side flags.
+  // Add transient client-side flags below here
```
jmthibault79: Added this because there are currently no examples

ui/src/environments/environment.prod.ts
```diff
@@ -15,9 +15,9 @@ export const environment: Environment = {
   trainingUrl: 'https://aou.nnlm.gov',
   inactivityTimeoutSeconds: 30 * 60,
   inactivityWarningBeforeSeconds: 5 * 60,
+  // Add transient client-side flags below here
```
jmthibault79: b/c there are currently none

calbach:
FYI `[risk=XXX]` is a specific tag that the release script is looking for to generate the release notes, the format from this PR title will not get picked up so the oncall will have to manually fill this in at release time. `[No Ticket]` does nothing and isn't an official team pattern, though some devs seem to like adding this (personally I prefer to omit it entirely).

jaycarlton:
I'd love to see these upgraded into real fields with restricted values. You could even have it be a pre merge validation to fill them in opt out.

calbach:
I agree that would be nice. Putting these into the title is definitely not ideal, but I'm not sure anything heavier weight is justified. From when I last looked into this, anything more complex will require integrating our release tooling with the GitHub API - which is already a higher complexity cost than I'd be willing to pay for this functionality.

The risk/ticket fields are FYI-only on the release notes, and have very minimal consequence if forgotten, so while nice-to-have I don't consider it to be high value. Of course if you have ideas or know of services which could make this easy, I'm definitely interested in hearing.

jaycarlton:
It's not a near-term priority, especially compared to Jira, unless and
until we have N repos that want to do the same thing. It's just a pet peeve
of mine to have a system where syntax or protocol validation only happens
by co-workers reminding folks to do it the right way. I've had too many
dealings with folks that want to be the compiler.

OTOH, once we get started with something like the github API, future
enhancements will start to be easier.

jmthibault79:
OK - thx for process clarification.  👍

PR #2559, opened 2019-09-09
RW-3417 [risk=no] workspace library changes

als364:
- Split out `workspaceList` into distinct lists of `publishedWorkspaces` and `featuredWorkspaces`.
- Hide the `Published Workspaces` tab behind a feature flag.
- Take `WorkspaceLibrary` props and state definitions out of the type parameter...

### DIFF
--- a/ui/src/app/pages/workspace/workspace-library.tsx
+++ b/ui/src/app/pages/workspace/workspace-library.tsx
@@ -112,23 +112,24 @@ export const WorkspaceLibrary = withUserProfile()
     ];
 
   async componentDidMount() {
-    this.updateListedWorkspaces();
+    this.updateWorkspaces();
   }
 
   componentDidUpdate(prevProps, prevState) {
     // Reload libraries when switching tabs
     if (this.state.currentTab !== prevState.currentTab) {
-      this.updateListedWorkspaces();
+      this.updateWorkspaces();
     }
   }
 
-  async updateListedWorkspaces() {
-    const workspaceListLoaded = await this.loadWorkspaceList();
-    const featuredWorkspacesLoaded = await this.loadFeaturedWorkspaces(workspaceListLoaded);
-    this.updatePublishedWorkspaces(featuredWorkspacesLoaded);
+  async updateWorkspaces() {
+    const workspaceListLoaded = await this.getAllPublishedWorkspaces();
+    const featuredWorkspacesLoaded = await this.getFeaturedWorkspaces(workspaceListLoaded);
+    this.filterPublishedWorkspaces(featuredWorkspacesLoaded);
   }
 
-  async loadWorkspaceList() {
+  // Gets all published workspaces, including those configured as 'featured'
+  async getAllPublishedWorkspaces() {
     try {
       const workspacesReceived = await workspacesApi().getPublishedWorkspaces();
       workspacesReceived.items.sort(
@@ -144,8 +145,9 @@ export const WorkspaceLibrary = withUserProfile()
     }
   }
 
-
-  async loadFeaturedWorkspaces(workspaceListLoadedPromise) {
+  // Gets the 'featured workspaces' config and filters the list of published workspaces to
+  // find the 'featured' ones
+  async getFeaturedWorkspaces(workspaceListLoadedPromise) {
     try {
       const resp = await featuredWorkspacesConfigApi().getFeaturedWorkspacesConfig();
       const idToNamespace = new Map();
@@ -162,7 +164,8 @@ export const WorkspaceLibrary = withUserProfile()
     }
   }
 
-  updatePublishedWorkspaces(featuredWorkspacesLoadedPromise) {
+  // Filter the list of published workspaces to find the ones that are not configured as 'featured'
+  filterPublishedWorkspaces(featuredWorkspacesLoadedPromise) {
     const publishedWorkspaces = this.state.workspaceList.filter(ws =>
       !fp.contains(ws, this.state.featuredWorkspaces)
     );
@@ -219,7 +222,7 @@ export const WorkspaceLibrary = withUserProfile()
                   return <WorkspaceCard key={wp.workspace.name}
                                         wp={wp}
                                         userEmail={username}
-                                        reload={() => this.updateListedWorkspaces()}/>;
+                                        reload={() => this.updateWorkspaces()}/>;
                 })}
               </div>)}
           </div>}
@@ -233,7 +236,7 @@ export const WorkspaceLibrary = withUserProfile()
                     return <WorkspaceCard key={wp.workspace.name}
                                           wp={wp}
                                           userEmail={username}
-                                          reload={() => this.updateListedWorkspaces()}/>;
+                                          reload={() => this.updateWorkspaces()}/>;
                   })}
               </div>)}
           </div>}
--- a/ui/src/app/pages/workspace/workspace-library.tsx
+++ b/ui/src/app/pages/workspace/workspace-library.tsx
@@ -102,6 +102,11 @@ export const WorkspaceLibrary = withUserProfile()
     };
   }
 
+  // Defaulting to the environment variable value here, but allowing it to be overridden by props.
+  // This is to make our testing easier. We don't use environment variable injection anywhere in
+  // our unit testing framework, so we have to pass it in via props. We also don't create this
+  // component in any way except for legacy Angular magic, so we have to get this information via
+  // environment variable.
   libraryTabs = (this.props.enablePublishedWorkspaces || environment.enablePublishedWorkspaces)
     ? [
       libraryTabEnums.FEATURED_WORKSPACES,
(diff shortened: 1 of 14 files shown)


ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -140,19 +162,34 @@ export const WorkspaceLibrary = withUserProfile()
...
   render() {
     const {profile: {username}} = this.props.profileState;
-    const {currentTab, errorText, featuredWorkspaces, workspaceList,
-      workspacesLoading} = this.state;
+    const {
+      currentTab,
+      errorText,
+      featuredWorkspaces,
+      publishedWorkspaces,
+      workspaceList,
```
als364: The linter does not like that I'm declaring a variable here that's unused. I could:
- decide I don't care and merge anyways
- change the name to 'unused', decide I don't care, and merge anyways
- get rid of this entire declaration and access everything via `this.state`
- figure out a way to keep `workspaceList` out of the state despite it being variable
Thoughts?

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -67,55 +64,78 @@ const LibraryTab: React.FunctionComponent<{
     </Clickable>;
   };
 
+class ProfileState {
```
gjuggler: (Recommended, but open to pushback)

I see what you're doing here (and applaud the effort to factor out a real data definition!). But a few notes / thoughts:

- It's a bit confusing to have a data type which ends up in React props, but has a class suffix of "State". Some naming ideas:
  - ReloadableProfile
  - ProfilewithReload
  - ProfileProps
  - ProfileHook
  - ProfileWrapper
- It looks like we use the same data structure in a number of files (I searched "profileState" and found a bunch of results). Should we define the interface in a more central location and reduce our overall duplication? If you opt for this, I'd suggest splitting it out into a separate tech-debt ticket and PR.
- I think we have a pretty strong pattern of using interfaces, rather than classes, to define props & state data types.

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -67,55 +64,78 @@ const LibraryTab: React.FunctionComponent<{
...
       currentTab: libraryTabEnums.FEATURED_WORKSPACES,
       errorText: '',
       featuredWorkspaces: [],
+      publishedWorkspaces: [],
       workspaceList: [],
       workspacesLoading: true
     };
   }
 
+  libraryTabs = (this.props.enablePublishedWorkspaces || environment.enablePublishedWorkspaces)
```
gjuggler: Shouldn't this be "const libraryTabs"?

gjuggler: This boolean logic feels a bit awkward – why do we need to look at both the props *and* the environment to decide whether to enable showing published workspaces? I'd be inclined to choose one or the other. If (as I might guess) you're using props to enable easier testing, I'd recommend aligning entirely on that pattern.

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -67,55 +64,78 @@ const LibraryTab: React.FunctionComponent<{
...
-    switch (tabEnum) {
-      case libraryTabEnums.PUBLISHED_WORKSPACES:
-        this.loadPublishedWorkspaces();
-        break;
-      default:
-        this.loadFeaturedWorkspaces();
-        break;
-    }
+  async updateListedWorkspaces() {
+    const workspaceListLoaded = await this.loadWorkspaceList();
```
gjuggler: This pattern feels a bit unusual (unless I'm missing some detail or nuance in async / await!). My mental model is that it's meaningful to store the result of an "await this.foo()" expression only if you need to store the result of the async call. The code here just seems to want to ensure a series of two promises are executed successfully; I think a clearer way to do that may be to use try / catch:

try {
  await this.loadWorkspaceList();
  await this.loadFeaturedWorkspaces();
} catch (e) {
  // Handle an error encountered during either of the async calls.
}

(Note: I'm not super-fresh on React & async calls, so I could very well be wrong on any of these points!)

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -67,55 +64,78 @@ const LibraryTab: React.FunctionComponent<{
...
+    const featuredWorkspacesLoaded = await this.loadFeaturedWorkspaces(workspaceListLoaded);
+    this.updatePublishedWorkspaces(featuredWorkspacesLoaded);
   }
 
-  async loadFeaturedWorkspaces() {
+  async loadWorkspaceList() {
```
gjuggler: This method and "loadFeaturedWorkspaces" below may warrant a small method-level comment describing what exactly is being loaded. For example:

// Loads the set of all published workspaces.
async loadWorkspaceList

// Loads the config file containing system-anointed featured workspaces.
async loadFeaturedWorkspacesList

ui/src/app/pages/workspace/workspace-library.tsx
gjuggler: Maybe I'm missing something here, but this is just an expression which is extracting certain values from this.state using Javascript's destructuring syntax. Can you simply remove the workspaceList entry from this destructuring expression?

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -67,55 +64,78 @@ const LibraryTab: React.FunctionComponent<{
...
-  updateListedWorkspaces(tabEnum) {
-    switch (tabEnum) {
-      case libraryTabEnums.PUBLISHED_WORKSPACES:
-        this.loadPublishedWorkspaces();
-        break;
-      default:
-        this.loadFeaturedWorkspaces();
-        break;
-    }
+  async updateListedWorkspaces() {
```
gjuggler: Naming nit: While reading through the code I was getting pretty confused between this method and the "loadWorkspaceList" method. It would be super nice to have more easily distinguishable names. Some ideas for renaming this method:

- UpdateAllWorkspaces
- UpdateWorkspaces
- LoadWorkspacesAndConfig

ui/src/environments/environment.test.ts
```diff
@@ -5,4 +5,5 @@ export const environment: Environment = {
   ...testEnvironmentBase,
   displayTag: 'Test',
   debug: false,
+  enablePublishedWorkspaces: true,
```
gjuggler: nitpick: you should be able to set this to "true" in test-env-base and leave it out of this file. The test-env-base pattern was my (rather weak) attempt to reduce config duplication across the environments which are "test-like". However, I'm realizing this hasn't really caught on, and having each environment config be fully-specified may be the better path forward. Open to thoughts here (I had a round of comments w/ Andrew on a similar issue the other day... if you two decide this pattern isn't useful, please feel free to follow-up and remove it!)

ui/src/testing/stubs/workspaces-api-stub.ts
```diff
@@ -50,8 +50,15 @@ export const workspaceStubs = [
     },
     published: false,
     dataAccessLevel: DataAccessLevel.Registered
-  }
-];
+  };
+}
+
+export function buildWorkspaceStubs(suffixes: string[]) {
```
gjuggler: Can you annotate a return type here?

ui/src/app/pages/workspace/workspace-library.tsx
als364: Wouldn't that assign `this.state.workspaceList` to this scope's `workspacesLoading`?

ui/src/app/pages/workspace/workspace-library.tsx
gjuggler: Ah – I didn't realize that this is a class member variable, not a local var. My bad!

gjuggler: I meant, if props is needed for testing, just use the props value and don't refer to the environment at all from here.

ui/src/app/pages/workspace/workspace-library.tsx
gjuggler: Checking my understanding here – this stack overflow post is in line with my mental model: https://stackoverflow.com/questions/45479341/how-is-async-await-working-in-serial-and-parallel

And here are pointers to my prior experience on this: [this pattern](https://github.com/all-of-us/workbench/blob/327618a73f5e19e30b355fdf62d4f83c82e7d0bd/ui/src/app/pages/admin/admin-user-bypass.tsx#L56), from our user-admin page, triggers parallel requests (since the await calls are within a .forEach loop they execute in parallel). But [this pattern](https://github.com/all-of-us/workbench/blob/327618a73f5e19e30b355fdf62d4f83c82e7d0bd/ui/src/app/pages/homepage/registration-dashboard.tsx#L242) sends requests in serial, since execution pauses at each 'await' call in the for loop.

YMMV – if it's low effort, I'd recommend doing some logging as a sanity-check to make sure this works as expected!

ui/src/app/pages/workspace/workspace-library.tsx
als364: It looks like no, it will not, and instead this is syntactic magic for 'go find the value for the key with that name'

PR #2565, opened 2019-09-09
[risk=no] Create a new cell when inserting a snippet

calbach:
This feels more natural. Raised in feedback from @deflaux. Also approved by Karthik.

### DIFF
--- a/api/cluster-resources/aou-snippets-menu.js.template
+++ b/api/cluster-resources/aou-snippets-menu.js.template
@@ -25,6 +25,8 @@ define([
       const menus = snippets_menu.options['menus'];
       menus[0] = {
          ...menus[0],
+         // By default, code is inserted into the currently selected cell.
+         'insert_as_new_cell': true,
          'sub-menu': [...customMenus, ...menus[0]['sub-menu']]
       };
 
--- a/api/cluster-resources/aou-snippets-menu.js.template
+++ b/api/cluster-resources/aou-snippets-menu.js.template
@@ -44,6 +44,9 @@ define([
     };
 
     // By default, code is inserted into the currently selected cell.
+    if (!Jupyter.notebook.config.data.snippets) {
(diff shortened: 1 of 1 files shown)


jmthibault79:
Looks reasonable but I am struggling trying to test this.  Do you have suggestions?  My attempt so far:

1. Local changes here -> https://github.com/all-of-us/workbench/commit/995d7fe961b1f7c00d64844d10bc841ab3f2a3be
2. Create new bucket `all-of-us-workbench-test-cluster-resources-joel`
3. `project.rb deploy-gcs-artifacts --project all-of-us-workbench-test`
4. run local API + UI
5. recreate cluster

calbach:
That should have worked fine. Did you confirm or not-confirm whether you were picking up the updated Javascript?

Unfortunately my code wasn't working though... fixed.

jmthibault79:
No, it is not picking up the new JS.  What am I missing here?

calbach:
Sorry for the delay. Everything looks correct. I checked your bucket and it has the updated file.

The most likely issues here:
- However you are running your API server, it didn't apply the updated local config. dev-up will do this
- Verify your local UI server is actually talking to your local API server

jmthibault79:
Ah.  I was using `./gradlew appEngineRun`.  I'll try dev-up.

jmthibault79:
dev-up is now depolying the JS correctly.  I'm seeing this console error however: `Uncaught TypeError: Cannot set property 'insert_as_new_cell' of undefined`

calbach:
Interesting, thanks - I will take a look at this. May be failing in certain situations.

calbach:
Nice catch, was able to reproduce the issue on a fresh cluster and verify a fix; I'd already messed with the extension settings on my cluster which is why this code was working for me.

calbach:
Adding @als364 since @jmthibault79 is on vacation.

calbach:
Thanks Joel!

PR #2569, opened 2019-09-11
[risk=no]Describe BillingMigrationStatus

jmthibault79:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
@@ -62,10 +62,11 @@ public class Workspace {
     }
   }
 
+  // see https://precisionmedicineinitiative.atlassian.net/browse/RW-2705
   public enum BillingMigrationStatus {
-    OLD,
-    NEW,
-    MIGRATED
+    OLD, // pre-1PPW; this billing project may be associated with multiple workspaces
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/db/model/Workspace.java
```diff
@@ -62,10 +62,11 @@ public boolean equals(Object obj) {
     }
   }
 
+  // see https://precisionmedicineinitiative.atlassian.net/browse/RW-2705
   public enum BillingMigrationStatus {
-    OLD,
-    NEW,
-    MIGRATED
+    OLD, // pre-1PPW; this billing project may be associated with multiple workspaces
```
calbach: Also, this status is now being overloaded to track pre-VPC-SC projects, perhaps worth mentioning that here. @ericsong

jaycarlton: I would submit that OLD and NEW don't age well as identifiers. We should have separate enums for each migration (or a map from a migration enum to finished state).

calbach: Good point. The plan for this enum was to use it and remove the column after cleaning up after the migration. If it winds up being a permanent fixture though due to the nature of the migration, agreed we should rename these enums.

PR #2570, opened 2019-09-12
[RW-3350][Risk=no] Add request type to trace

s-rubenstein:
The change to `afterCompletion` from `postHandle` should also give us more accurate numbers, but also I am hoping it can fix https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&view=planning&selectedIssue=RW-3240&epics=visible&selectedEpic=RW-1801

Not marking that as in progress though, since I don't know for sure that will fix it.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
@@ -17,7 +17,6 @@ import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.firecloud.FirecloudApiClientTracer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
-import org.springframework.web.servlet.ModelAndView;
 import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 
 // Interceptor to create a trace of the lifecycle of api calls.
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -49,7 +48,10 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
 
     SpanBuilder requestSpanBuilder =
         tracer.spanBuilder(
-            workbenchConfigProvider.get().server.shortName + request.getRequestURI());
+            workbenchConfigProvider.get().server.shortName
```
jaycarlton: Please use `String.format("%s%s%s")` in preference to concatenation.

blrubenstein:
@jaycarlton to be clear, it is not possible to unit test it. I described how to do the manual test (after prefacing that it was a painful process), but opted to merge per team agreement because Andrew had approved the PR. I can link this PR tomorrow to the ticket I think it may have fixed, but it’s hard to be certain because it is a sporadic problem, so we just need to keep an eye on it and see if it still popping up, which I am doing.

I can do the string format in a follow up.

jaycarlton:
OK, thanks. There are a couple of things I want to talk about, and this PR
might not have been the place for it. I'll catch you sometime early this
week if that's ok.

-Jay

PR #2571, opened 2019-09-12
[RW-3451][risk=no]Replacing the beautiful hex background with the generic new plus background

andy7i:
See screenshots to see the new vs old backgrounds.
Hex:
Screen Shot 2019-09-12 at 3 38 59 PM
Plus:
Screen Shot 2019-09-12 at 3 39 01 PM

Get a review from Lou to see if we really do want the new background image.


ui/src/app/pages/signed-in/component.css
```diff
@@ -18,15 +18,12 @@ body {
 }
 
 .background-image {
-  background-image: url('/assets/images/hex-background@2x.png');
-  background-repeat: no-repeat;
+  background-image: url('/assets/images/BG-Pattern.png');
+  background-size: 80px; /* half the size of the image */
```
jaycarlton: I'd there a way to refer to the one side directly and have the calculation happen here?

ui/src/app/pages/app/component.css
```diff
@@ -92,16 +92,12 @@
 }
 
 .background-image {
-  background-image: url('/assets/images/hex-background@2x.png');
-  background-repeat: no-repeat;
-  z-index: -1;
-  background-size: cover;
-  right: 0rem;
-  top: 3.3rem;
+  background-image: url('/assets/images/BG-Pattern.png');
```
jaycarlton: The new image file has a different naming convention from the old one.

andy7i: That was the filename as given by the designer. Do you think we should rename it to something like `bg-pattern@2x.png`?

ui/src/app/pages/signed-in/component.css
andy7i: Not that I could find from a quick search on google. Do you think one exists?

ui/src/app/pages/app/component.css
jaycarlton: I don't know what the standard is. I probably wouldn't put an @ sign in the file name though.

ui/src/app/pages/signed-in/component.css
jaycarlton: Maybe, but I think you may have to hand-enter the image dimensions. https://stackoverflow.com/a/11841812

I was just curious. You don't need to spend a ton of time tracking it down. I really haven't done much css at all.

PR #2573, opened 2019-09-12
[RW-3512][risk=NO] Refactoring DataSetService

jaycarlton:
In order to fix RW-3247 cleanly, I'm refactoring things a bit. Most of this is using IntelliJ's draft refactoring.

Some highlights:

- Extracted several smaller methods
- moved `DomainConceptIdInfo` into a private static class
- used immutable collections with their builders, to highlight when the list is being built vs completed
- several renamings
- replaced some string concatenation with `StringBuilder`
- make almost every locally declared variable `final`

I made a few more edits after the Mob, especially in places where IntelliJ gave me a bloated argument list for extracted functions. 

*UPDATE* This [thread](https://github.com/all-of-us/workbench/pull/2573#discussion_r324254634) is important, but was shut off by a later commit. 

I don't want to have a huge `if` branch if we can avoid it. I think we can do it in a more function-oriented way by handling all the domains with no concept sets the same way. We do need a small conditional because of a limitation in the SQL `IN` clause that doesn't handle empty lists, but that should happen later in the SQL rendering phase, not in the gathering phase.

Is it reasonable to do that in this already large PR? I feel like I kinda need to, as it's not necessarily working now, so I can't justify organizing things around logic that's not known to be correct.

The larger plan to make this fluid involves a three-part pipeline.

Phase 1 is gathering up all the logic we have now to populate the things that go in each clause based on the business logic specific to the CDR and AoU Workbench selection logic. This is where anything dealing with `includeAllParticipants` and the `PERSON` domain's peculiarities should be handled. I'll call the class `DataSetRequestEvaulator`, which will be responsible for determining what to feed to the next phase.

For Phase 2, we then make a class called `UserSqlQueryBuilder` to hold all the parts of a SQL statement that we need and then just call toString() on it to generate the query. Critically, this class won't know anything about AoU, Concepts, Domains, Datasets, Persons, etc. It simply writes the query based on the input lists for the Select, From (including joins), and Where clauses. This logic may be replaced by a third-party Sql query driver as imagined in [RW-3446](https://precisionmedicineinitiative.atlassian.net/browse/RW-3466).

Phase 3 is just stuffing the queries into the two notebook types. I want start with a high-level interface like `UserNotebookCodeGenerator` and have an implementation for each language. Then we can inject that with a map from the language enum to `PythonNotebookCodeGenerator`, etc. The SQL query is the same for both (I hope). There's an existing ticket [RW-3340](https://precisionmedicineinitiative.atlassian.net/browse/RW-3340) as well to use a template engine to simplify this phase. Not sure if I'll do that in the first step or not.

For this branch in particular, we may want to check in quickly so my branch doesn't get too far from the mainline. I'm going to try to focus on unit tests to start with. Cutting this list down to the bare minimum, we have:
- [x] unit tests for `DataSetServiceImpl`. Inputs for different combinations of concept sets, domains, person/non-person, etc will be invaluable. I've heard things anecdotally about some of the edge cases, which I appreciate, but if anyone can supply hard inputs and pseudocode for how the edges are handled, that will help here.
- [x] More consistent error handling & reporting policy: unit tests for the controller want exceptions in some cases but a silent success for an empty query in another
- [x] Actually fix RW-3247 and put in a unit test for it
- [x] feedback from this PR

UPDATE: added a story for this refactoring at [RW-3512](https://precisionmedicineinitiative.atlassian.net/browse/RW-3512)

Follow on work for another PR will break up the `DataSetServiceImpl` into the components described above. Hopefully we can switch to a third-party sql library at the same time.


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -216,97 +190,140 @@ public DataSet saveDataSet(
...
+        + conceptSetListQuery2
+        + s3;
+  }
+
+  private void validateSelectedConceptSetsForDomain(List<ConceptSet> conceptSetsSelected,
+      Domain domain) {
+    conceptSetsSelected.stream()
+        .map(ConceptSet::getDomain)
+        .distinct()
+        .filter(cs -> domain != Domain.PERSON)
```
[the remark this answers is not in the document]
jaycarlton: To clarify, yes I suspect this wasn't intended, but it might still be load-bearing because of reasons I didn't want to chase down at the time. The nice thing about refactoring is that it makes bugs more obvious. But yeah, I can't think why you'd ever want to check a higher more broadly scoped condition in an inner loop or stream.

I think the moral is that we've just about extended this approach of hand-rolling SQL queries as far as we safely can. I want to sneak in work on https://precisionmedicineinitiative.atlassian.net/browse/RW-3466 at some point.

jaycarlton:
My current thinking after talking with @NehaBroad a bit is to split this class and its kin into horizontal layers. The Bottom layer will know how to create SQL queries based on a bit of structure and lists of columns, tables, IDs maybe, etc. Classes just above that should handle the business logic of dealing with concept sets, domains, and any person-specific behavior. Classes in this tier will know nothing about SQL syntax generally (though it might need to know something about our tables' structural assumptions).

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -121,42 +120,22 @@
...
-      throw new BadRequestException("Missing values");
-    }
-    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+    validateDataSetCreateRequest(dataSetRequest);
+    final Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
-    long wId = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
-    List<DataSetValues> dataSetValuesList =
+    final long workspaceId1 = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
```
jaycarlton: todo: rename this.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -216,101 +222,144 @@ public DataSet saveDataSet(
...
+          .append(" ")
+          .append(formatValuesLinkingPair(valuesLinkingPair));
+
+      validateSelectedConceptSetsForDomain(conceptSetsSelected, domain);
 
       String conceptSetQueries =
           conceptSetsSelected.stream()
-              .filter(cs -> d == cs.getDomainEnum())
-              .filter(cs -> d != Domain.PERSON)
+              .filter(cs -> domain == cs.getDomainEnum())
```
blrubenstein: Could this be done as one line with an `&&`?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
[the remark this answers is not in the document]
blrubenstein: I do think we could probably remove this. If we want to not run this on PERSON, we should do that as an if statement and then return if PERSON.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -132,185 +163,241 @@ public DataSet saveDataSet(
...
+          .filter(valueSet -> valueSet.getDomain() == domain)
+          .collect(Collectors.toList());
+
+      final ValuesLinkingPair valuesLinkingPair = getValueSelectsAndJoins(
+          valuePairsForCurrentDomain);
+
+      queryBuilder.append(String.join(", ", valuesLinkingPair.getSelects()))
+          .append(" ")
+          .append(formatValuesLinkingPair(valuesLinkingPair));
+      validateSelectedConceptSetsForDomain(conceptSetsSelected);
```
NehaBroad: We do want this to happen for Domain.PERSON

jaycarlton: Do you think you could write a unit tests (against master) demonstrating what is supposed to happen with PERSON? We suspect it's not currently working correctly based on the stream logic.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: The `(cs -> d != Domain.PERSON)` checks are getting moved external to this stream, since it doesn't depend on anything in the stream. I will probably need help with that logic though.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -132,185 +163,241 @@ public DataSet saveDataSet(
...
+      resultBuilder.put(domain.toString(), queryJobConfiguration);
     }
-    return dataSetUtil;
+    return resultBuilder.build();
+  }
+
+  private static String formatWhereClause(String conceptSetListQuery, String s,
+      String standardConceptIdColumn, String s2, String sourceConceptIdColumn,
+      String conceptSetListQuery2, String s3) {
+    return s
```
blrubenstein: Should this be string builder rather than concatenation?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -466,13 +530,8 @@ private static String convertQueryToString(
...
     }
 
-    return sqlSection
-        + "\n\n"
-        + namedParamsSection
-        + "\n\n"
-        + dataFrameSection
-        + "\n\n"
-        + displayHeadSection;
+    return formatWhereClause(namedParamsSection, sqlSection, "\n\n", "\n\n", dataFrameSection,
```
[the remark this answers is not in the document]
blrubenstein: This seems like something different than `formatWhereClause`? Maybe the naming on the function should be different?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: Yes. I wound up re-inlining this function. IntelliJ was too greedy in parameterizing it. This should be going into my phase 2 class, along with anything that tries to write sql.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -216,101 +222,144 @@ public DataSet saveDataSet(
...
         if (!includesAllParticipants) {
-          query = query.concat(" \nAND (PERSON_ID IN (" + cohortQueries + "))");
+          queryBuilder
+              .append(" \nAND (PERSON_ID IN (")
+              .append(cohortQueries)
+              .append("))");
         }
       } else if (!includesAllParticipants) {
-        query = query.concat(" \nWHERE PERSON_ID IN (" + cohortQueries + ")");
+        queryBuilder
```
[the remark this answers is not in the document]
jaycarlton: It's more complex than just this. This will all be revisited hopefully.

calbach:
You can remove reviewers, but I don't know how that affects email notifications (probably they'll still receive updates since they participated in the PR).

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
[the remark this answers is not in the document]
jaycarlton: Turns out the global filter's effects of making the stream degenerate were necessary, so I've tried to replicate their effects in this version. I've got one unit test on the person situation, and @NehaBroad has some more in the works from what I hear.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: I wrote a simple one against master in the controller test. Most of that class should move into the service test class, but I gave up on doing that for this PR.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -216,101 +222,144 @@ public DataSet saveDataSet(
...
 
       String conceptSetQueries =
           conceptSetsSelected.stream()
-              .filter(cs -> d == cs.getDomainEnum())
-              .filter(cs -> d != Domain.PERSON)
+              .filter(cs -> domain == cs.getDomainEnum())
+              .filter(cs -> domain != Domain.PERSON)
               .flatMap(cs -> cs.getConceptIds().stream().map(cid -> Long.toString(cid)))
               .collect(Collectors.joining(", "));
       String conceptSetListQuery = " IN (" + conceptSetQueries + ")";
```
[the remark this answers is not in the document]
jaycarlton: Solution is to get the ifs and Optionals right so this doesn't happen. In the new code I'm throwing `IllegalStateException` if list is empty.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -171,163 +205,309 @@ public DataSet saveDataSet(
...
+      throw new BadRequestException("Concept Sets must contain at least one concept");
+    }
+  }
+
+  private Optional<String> buildConceptIdListClause(
+      Domain domain, List<ConceptSet> conceptSetsSelected) {
+    final Optional<String> conceptSetSqlInClauseMaybe;
+    if (supportsConceptSets(domain)) {
+      conceptSetSqlInClauseMaybe =
+          Optional.of(buildConceptIdSqlInClause(domain, conceptSetsSelected));
```
gjuggler: Was it really worth breaking out this method? The name of that method is similar but slightly different and it's only ~5 lines of code, and AFAICT not used anywhere else or separately tested or overridden. It's an aesthetic thing, but as a reader I felt more confused than aided by the separation (especially as these methods have no comments, javadocs, or unit tests).

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -171,163 +205,309 @@ public DataSet saveDataSet(
...
-      queryMap.put(query, cohortParameters);
-      QueryJobConfiguration queryJobConfiguration =
-          bigQueryService.filterBigQueryConfig(
-              QueryJobConfiguration.newBuilder(query)
-                  .setNamedParameters(cohortParameters)
-                  .setUseLegacySql(false)
-                  .build());
-      dataSetUtil.put(d.toString(), queryJobConfiguration);
+  @VisibleForTesting
+  public boolean conceptSetSelectionIsNonemptyAndEachDomainHasAtLeastOneConcept(
```
gjuggler: nit: missing capital "e" in Empty for lowerCamelCase

[Optional] It looks like this method is effectively "isConceptSetValid," but with a very long and explicit method name. If I were bike-shedding this, I'd say maybe the two rules that show up in the method name ("selection is non-empty" and "each domain has nonzero concepts") should each be their own method, which would make for more modular logic and less monolithic method names. But this is so low priority I probably wouldn't bother...

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -358,11 +538,12 @@ private String getQualifiedColumnName(Domain currentDomain, String columnName) {
             "Kernel Type " + kernelTypeEnum.toString() + " not supported");
     }
     return queryJobConfigurationMap.entrySet().stream()
+        .filter(e -> e.getKey() != null)
```
gjuggler: This is an interesting addition – when do we expect to see a map entry with a null key? Seems worth a comment, since the reason for this feels unclear from context.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -373,55 +554,72 @@ private String getQualifiedColumnName(Domain currentDomain, String columnName) {
   private String getColumnName(CdrBigQuerySchemaConfig.TableConfig config, String type) {
     Optional<CdrBigQuerySchemaConfig.ColumnConfig> conceptColumn =
         config.columns.stream().filter(column -> type.equals(column.domainConcept)).findFirst();
+    // TODO: move this logic to new DataSetRequestValidator class. Goal is for the service class
```
gjuggler: That's an interesting design principle – why should service classes never throw exceptions? Or am I misinterpreting the comment?

Side note: we should really work towards having a place where we can document (and disagree on, and come to consensus on) over-arching code structure principles like this.

api/src/test/java/org/pmiops/workbench/db/dao/DataSetServiceTest.java
```diff
@@ -0,0 +1,224 @@
...
+    final DataSetRequest result = new DataSetRequest();
+    result.setCohortIds(ImmutableList.of(COHORT_ID_1));
+    result.setConceptSetIds(ImmutableList.of(CONCEPT_SET_ID_1));
+    result.setValues(Collections.emptyList());
+    result.setValues(ImmutableList.of(DOMAIN_VALUE_PAIR_PERSON_PERSON_ID));
+    return result;
+  }
+
+  @Test(expected = BadRequestException.class)
+  public void itThrowsForNoCohortOrConcept() throws Exception {
```
gjuggler: Consistency nit: I think most if not all of our other test classes use the "testFoo" form, not the "itFoo" form. Unless there's a strong functional reason to phrase things this way, I'd prefer consistency over personal style here.

api/src/test/java/org/pmiops/workbench/db/dao/DataSetServiceTest.java
```diff
@@ -0,0 +1,224 @@
...
+    final DataSetRequest invalidRequest = buildEmptyRequest();
+    ImmutableMap<String, QueryJobConfiguration> configurationsByDomain =
```
gjuggler: Is there a reason to include the ImmutableMap boilerplate here if the value isn't being used in any way? e.g. I might have thought

Map<String, QueryJobConfiguration> result = dataSetServiceImpl.generateQueryJobConfigurations...

would be fine.

api/src/test/java/org/pmiops/workbench/db/dao/DataSetServiceTest.java
```diff
@@ -0,0 +1,224 @@
...
+        ImmutableMap.copyOf(
+            dataSetServiceImpl.generateQueryJobConfigurationsByDomainName(invalidRequest));
+  }
+
+  @Test
+  public void itGetsCohortQueryStringAndCollectNamedParameters() throws Exception {
```
gjuggler: Grammar nit: collect -> collects

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -171,163 +205,309 @@ public DataSet saveDataSet(
...
+  // signatures.
+  private ImmutableList<ConceptSet> getExpandedConceptSetSelections(
+      PrePackagedConceptSetEnum prePackagedConceptSet,
+      List<Long> conceptSetIds,
+      List<Cohort> selectedCohorts,
+      boolean includesAllParticipants,
+      List<DomainValuePair> domainValuePairs) {
+    final ImmutableList<org.pmiops.workbench.db.model.ConceptSet> initialSelectedConceptSets =
+        ImmutableList.copyOf(this.conceptSetDao.findAllByConceptSetIdIn(conceptSetIds));
+    // TODO: can we not check this earlier?
```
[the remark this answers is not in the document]
jaycarlton: Most of them are there to make the reviewers scratch their heads to see if they agree, which may be worse still. Though making each one a Jira issue seems like overkill. But yeah, I'll kill them and put them into the round 2 ticket.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: Seems like we had that before due to some pre-stream map lookup into list logic that required a null check when going over the list. Good news is ImmutableList won't allow nulls.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: [Umm](https://pmi-engteam.slack.com/archives/CN4AX0K9V/p1569948603007000). _Whistles_

Yeah it wasn't a clear comment. I'm OK with them letting runtime exceptions pass through, but not having them throw doesn't feel very functional and makes reasoning about the logic harder.

So I don't feel like exceptions belong in implementation/logic classes, and I only want to see throws happening at the public API boundary. But when I do throw I want them to be checked exceptions so there are no excuses and the IDE forces you to catch and deal with them. I've never worked on anything low-level enough where I felt the need to throw a RuntimeException, since I feel those are much more exceptional.

But the comment should've said we shouldn't be doing request validation, and a 400 shouldn't happen because of complicated business logic giving the user a null result. Talked with @blrubenstein the other day on the strategy for this on the front end too. We really want a 200 with an empty map and a useful message to the end user. Currently it's doing 400s for missing cohort and concept set, but not for empty column/value set.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -169,11 +149,35 @@
     }
   }
 
+  private DataSetValues getDataSetValuesFromDomainValueSet(DomainValuePair domainValueSet) {
```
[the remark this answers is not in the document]
jaycarlton: I think it's a holdover from the previous [version](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java#L244). 

Looking at it closer, I don't understand why we need both `org.pmiops.workbench.model.DomainValuePair` and `org.pmiops.workbench.db.model.DataSetValues`. I guess one is a db model and one a swagger-generated model.

api/src/test/java/org/pmiops/workbench/db/dao/DataSetServiceTest.java
jaycarlton: They beat this into me at HubSpot. I think the reason is that it's an assertion in active voice, so if you see the failure in the log you know exactly what the test case is trying to do. It's harder to start a sentence with `test` that's as crisp. Both prefixes are redundant b/c the framework knows which are test methods based on the annotation.

I'll change these though.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: Good point, but remember this is merely the opening act for what I'd like to do with this class. I wanted to keep going but had to come up for air in the middle.

The key reason in this case is that I want to highlight that not every (valid) query has a list of concept IDs (thus the Optional). Previously this was implied by concatenating an empty list, but there was lots of confusion around that. I also want to move to a model that encodes this domain knowledge and query parts in a richer way.

Also, the parent method goes from line 367-447. I broke out this one to shrink it slightly, but more importantly, because I understood what it was actually doing and could give it a name. Additionally, lots of these methods are going to be rearranged and put into smaller services and classes, with their own unit tests and things.

All of this is an intermediate state, but I to break down some crazy long functions.

Unit tests are a great idea, as are comments.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: removed the comment.

api/src/test/java/org/pmiops/workbench/db/dao/DataSetServiceTest.java
jaycarlton: True, I don't even need the result. I think I had plans for it but forgot. Thanks.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: Added unit tests and consolidated a couple of helper methods.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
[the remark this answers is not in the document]
jaycarlton: Although OTOH, if the existing code is messy enough, at TODO with a description of improvements might be better than a small Jira ticket. There's really no in between, which is the problem, for things that are between 30 minutes and 2 hours.

PR #2574, opened 2019-09-13
[no ticket][risk=no] default background color update

andy7i:
Background color update and adding an additional comment on a feature flagged feature just to make sure we're deleting a file once we stop using it.

The color is a lighter grey.

before:

Screen Shot 2019-09-17 at 12 29 41 AM

after:

Screen Shot 2019-09-17 at 12 29 56 AM

### DIFF
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -93,6 +93,7 @@ export const styles = reactStyles({
     height: '2.25rem', width: '2.75rem'
   },
   // once enableHomepageRestyle is enabled, delete all styles ending in 'ToDelete'
+  // '/assets/images/AoU-HP-background.jpg' can also be deleted when the feature flag is removed
   mainHeaderToDelete: {
     color: colors.white, fontSize: 28, fontWeight: 400,
     display: 'flex', letterSpacing: 'normal'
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -93,6 +93,7 @@ export const styles = reactStyles({
     height: '2.25rem', width: '2.75rem'
   },
   // once enableHomepageRestyle is enabled, delete all styles ending in 'ToDelete'
+  // '/assets/images/AoU-HP-background.jpg' can also be deleted when the feature flag is removed
```
jaycarlton: This works, thanks.

ui/src/styles.css
```diff
@@ -39,7 +39,7 @@ a {
 }
 
 body {
-  background-color: #F1F1F1;
+  background-color: #FAFBFC;
```
jaycarlton: nit: it might help to have names for the colors used in the pallet, for easier coordination with UI/UX designers.

andy7i: I like the idea. I think when we convert this file to React we should put this backgroundGrey in `colors.ts`

andy7i:
@jaycarlton 
> It would be good to have before & after images, but not required.

The color is a lighter grey.

before:

Screen Shot 2019-09-17 at 12 29 41 AM

after:

Screen Shot 2019-09-17 at 12 29 56 AM

PR #2576, opened 2019-09-13
[risk=low][RW-2705] Bulk Clone Workspaces 

ericsong:
Adding the tool that will clone all workspaces in an environment with the default CDR version.

The initial version, https://github.com/all-of-us/workbench/pull/2356, has grown stale and merge conflicts were hard to resolve because part of the changes in that pull request got pulled in.

This pull request puts together the remaining changes and adds the CDR override.

Scheduled to run on production Monday night with the following command
`./project.rb bulk-clone-workspaces --n 500 --project all-of-us-rw-prod`

Test
`./project.rb bulk-clone-workspaces --dry-run --n 500 --project all-of-us-rw-stable`


ericsong:
@calbach I think the plan is to merge it so we abide by security protocols of having code reviewed and merged before it gets executed. I believe there were some concerns about what the correct procedure would be if we were to just run this from a branch.

@gjuggler may have more details.

api/tools/src/main/java/org/pmiops/workbench/tools/BulkCloneWorkspacesTool.java
```diff
@@ -258,6 +263,7 @@ public CommandLineRunner run(
               workspaceService.getWorkspace(
                   dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getFirecloudName());
           CloneWorkspaceRequest request = createCloneRequest(apiWorkspace);
+          request.getWorkspace().setCdrVersionId(String.valueOf(defaultCdr.getCdrVersionId()));
```
calbach: nit: probably cleaner to pass this version ID to `createCloneRequest`, or just set it in there

calbach:
Be sure to rebase or merge before running next week

calbach:
Also be careful if there were any database changes this week - running locally might cause issues or fail. If there are DB changes you can either wait for the release to run this tool, or rebase onto the most recent release tag and cherry-pick https://github.com/all-of-us/workbench/pull/2578

PR #2582, opened 2019-09-16
[No Ticket][Risk=no] Change to string format

blrubenstein:


blrubenstein:
The URI starts with a `/`, so it had one baked in automatically, whereas the other strings need their own separator.

PR #2584, opened 2019-09-16
[RW-3285][risk=low]Profile - Add completed dates and checkmark behind feature flag

andy7i:
This will add the completed (or bypassed) date to the UI and a checkmark.

New UI changes are behind feature flag. The old code is separated for easy cleanup when we're ready to remove the feature flag.

Before and after screenshots:

before:
Screen Shot 2019-09-16 at 4 52 17 PM

after:
Screen Shot 2019-09-16 at 4 42 09 PM


ui/src/app/pages/profile/profile-registration-step-status.tsx
```diff
@@ -56,42 +59,79 @@ const ProfileRegistrationStepStatus: React.FunctionComponent<Props> =
...
+            { title }
+          </div>
+          <FlexRow>
+            <div>
+              { isComplete ? (
+                <FlexRow>
+                  <ClrIcon style={{color: colors.success, width: '2rem', height: '2rem'}}
+                           shape='check-circle' className='is-solid' />
+                  <FlexColumn style={{marginLeft: '0.5rem'}}>
+                    <div>{ wasBypassed ? 'Bypassed By Admin on:' : completedButtonText + ' on:' }
```
NehaBroad: Add a space after on: In case of eraCommons there is information about username too which is not looking right 

Screen Shot 2019-09-16 at 5 13 51 PM

andy7i: nice catch! with the fix it looks like this:

Screen Shot 2019-09-16 at 11 24 58 PM

andy7i: When none are complete, it looks like this:

Screen Shot 2019-09-17 at 12 37 56 AM

ui/src/app/pages/profile/profile-registration-step-status.tsx
```diff
@@ -56,42 +59,79 @@ const ProfileRegistrationStepStatus: React.FunctionComponent<Props> =
...
+              { children }
+            </div>
+          </FlexColumn>
         </div>
-        <div style={{ display: 'flex' }}>
-          <div style={styles.buttonContainer}>
-            { isComplete ? (
-              <Button
+      );
+    } else { // delete this block below once enableProfileCapsFeatures is removed
```
blrubenstein: TOL: It might be valuable to add the Feature Flag ticket number on all these todo cleanups, so that we have a way to search for all of them.

ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -86,6 +86,7 @@ interface RegistrationTask {
   buttonText: string;
   completedText: string;
   isRefreshable?: boolean;
+  completionTimestamp: Function;
```
ericsong: specify the signature?

ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -128,6 +135,9 @@ export const getRegistrationTasks = () => serverConfigStore.getValue() ? ([
       'and tools.',
     buttonText: 'Login',
     completedText: 'Linked',
+    completionTimestamp: (profile: Profile) => {
```
ericsong: i think we can do away with `isComplete` (at least the implementation) by just checking for `null` in `completionTimestamp`

I think leaving the function but changing it to call into `completionTimestamp` would also be good.

Also, we might be approaching the point where its worth formally defining this object as a class

blrubenstein: I got rid of it. Might make sense to be a class though, that would have made self reference easier.

PR #2587, opened 2019-09-17
[RW-3492][risk=no] Fix main background to adjust to size of content 

dolbeew:
Background div is not adjusting to the size of the content. 

Cuts of on longer pages (Data page):
Screen Shot 2019-09-17 at 10 56 37 AM

Overflows on shorter pages:
Screen Shot 2019-09-17 at 10 57 01 AM


blrubenstein:
Just make sure it doesn’t move or do anything strange because of relative

jaycarlton:
Do we need to update anything in Rainforest?

dolbeew:
AFAIK it won't affect rainforest

PR #2590, opened 2019-09-18
[RW-3288][risk=low] Recently updated workspaces a la Google Docs

als364:
Summary of the conversation below:
Q: Aren't you just recreating the `user_workspaces` table that we deleted a while ago because it was a bug factory?
A: No - that tracked per-user ACLs, which we now track in FC. This is tracking the four most recent workspaces a user looked at so we can display them on the homepage as a shortcut.

Q: Why aren't you extending the `user_recent_resources` table instead of making a new table?
A: A few reasons:
- `user_recent_resources` tracks something conceptually different - resources such as cohorts, concept sets, and notebooks that are owned by a workspace.
- We have different limits on how many workspaces and resources we want to keep track of, which makes the business logic a bit complicated.
- We already have a `workspace_id` column in `user_recent_resources` that we use for keeping track of what workspace these resources belong to.

Q: How will you keep track of changing permissions on workspaces?
A: Check in with Rawls whenever we grab the recent workspaces and remove all the ones that we no longer have access to.

Q: Why are you putting these calls in the UI when we usually do these sorts of actions on the backend?
A: It's hard to differentiate between calls to `getSomeWorkspaceOrResource` for the purpose of rendering that workspace or resource and calls to said endpoint for the purpose of listing workspaces or resources. This way, updates to `user_recent_workspace` are triggered by specific user actions.


api/src/main/resources/workbench.yaml
```diff
@@ -1594,6 +1594,64 @@ paths:
...
+            required: true
+            schema:
+              $ref: "#/definitions/RecentWorkspaceRequest"
+        responses:
+          200:
+            description: Successfully added/updated
+            schema:
+              $ref: "#/definitions/RecentWorkspace"
+
+    /v1/workspaces/{workspaceNamespace}/{workspaceId}/user-recent-workspaces/delete:
```
jaycarlton: Why do we need to expose this?

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +183,30 @@ public void activeStatus() {
     EnumSet.allOf(WorkspaceActiveStatus.class)
         .forEach(
             status ->
-                assertThat(mockDbWorkspace("1", "1", status).getWorkspaceActiveStatusEnum())
+                assertThat(mockDbWorkspace(workspaceIdIncrementer.getAndIncrement(),"1", status).getWorkspaceActiveStatusEnum())
```
jaycarlton: nit: space after comma

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +183,30 @@ public void activeStatus() {
...
                     .isEqualTo(status));
   }
+
+  @Test
+  public void getRecentWorkspaces() {
+    for (int i = 1; i < WorkspaceServiceImpl.RECENT_WORKSPACE_COUNT; i++) {
```
jaycarlton: can you use a stream, foreach, or for-in loop here?

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +183,30 @@ public void activeStatus() {
...
+      org.pmiops.workbench.db.model.Workspace workspace = workspaces.get(i);
+      workspaceService.updateRecentWorkspaces(workspace.getWorkspaceId(), USER_ID, Timestamp.from(NOW));
+    }
+    List<UserRecentWorkspace> recentWorkspaces = workspaceService.getRecentWorkspaces(USER_ID);
+    for (int i = 0; i < recentWorkspaces.size(); i++) {
```
jaycarlton: You can do this without a loop using `assertThat(recentWorkspaces).containsAll(workspaces)` or something similar.

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -364,7 +365,7 @@ export const Homepage = withUserProfile()(class extends React.Component<
                                               twoFactorAuthCompleted={twoFactorAuthCompleted}
                                             dataUseAgreementCompleted={dataUseAgreementCompleted}/>
                       ) : (
-                        <FlexRow style={{paddingTop: '2rem'}}>
+                        /*<FlexRow style={{paddingTop: '2rem'}}>
```
jaycarlton: can you go ahead and delete this commented code?

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -384,10 +385,38 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+                            <FlexColumn>
+                              <SmallHeader>Recently Accessed Items</SmallHeader>
+                              <RecentWork/>
+                            </FlexColumn>
+                          </React.Fragment>
+                        )
                   ) :
                   <Spinner dark={true} style={{width: '100%', marginTop: '5rem'}}/>}
-              </FlexColumn>
+              {/*</FlexColumn>*/}
```
jaycarlton: remove commented code please.

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -0,0 +1,40 @@
...
+      this.setState({loading: false});
+    }
+  }
+
+
+
+  render() {
+    return <div>lol</div>;
+  }
+}
```
jaycarlton: I don't know if we care, but your IDE isn't putting a final newline at the end of text files.

api/src/main/resources/workbench.yaml
```diff
@@ -3549,6 +3607,36 @@ definitions:
...
+        type: integer
+        format: int64
+      userId:
+        type: integer
+        format: int64
+
+  RecentWorkspace:
+    type: object
+    properties:
+      workspaceId:
```
jaycarlton: FWIW, I would tend to leave off the prefix on the property names, since we know we're talking about a workspace. So just `id`, `namespace`, `firecloudName`, etc.

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentWorkspaceDao.java
```diff
@@ -0,0 +1,10 @@
+package org.pmiops.workbench.db.dao;
+
+import org.pmiops.workbench.db.model.UserRecentWorkspace;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.List;
+
+public interface UserRecentWorkspaceDao extends CrudRepository<UserRecentWorkspace, Long> {
+    List<UserRecentWorkspace> findTopByUserIdOrderByLastUpdateTime(long userId);
```
jaycarlton: This method doesn't appear to take in a max count, so it could be thousands of workspaces if the user was running a script potentially.

I think you want to order by update time descending maybe?

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +183,30 @@ public void activeStatus() {
...
+    }
+  }
+
+  @Test
+  public void updateRecentWorkspaces() {
+    workspaces.forEach(workspace -> {
+      workspaceService.updateRecentWorkspaces(workspace.getWorkspaceId(), USER_ID, Timestamp.from(NOW));
+    });
+    List<UserRecentWorkspace> recentWorkspaces = workspaceService.getRecentWorkspaces(USER_ID);
+    for (int i = 0; i < recentWorkspaces.size(); i++) {
```
jaycarlton: same trick with the assertion above. You should almost never need to use the good-ole C-style loops anymore.

gjuggler:
I think this needs a quick design check-in; let's grab a few minutes ad-hoc today and/or I'll set up some time for today or tomorrow if possible.

The warning sign for me here is that we're creating a new table which seems like a caching layer that mirrors the actual source of truth re: a user's workspaces and their last-accessed times.

That's not necessarily a bad design (if we manage it carefully and if the optimization is needed). However, we did have a very similar-sounding table in the recent past: it was created Aug 2018 (https://github.com/all-of-us/workbench/pull/1118) to aid in showing recently-used resources, but it was ultimately deleted June 2019 (https://github.com/all-of-us/workbench/pull/2243). See RW-1381 for the motivation for deletion – it was a high maintenance burden and source of data-consistency bugs in our system.

One could argue that this time we're using it for a different, more focused purpose, and we'll be more careful to not abuse / mis-use the construct... but there's a burden of proof given our team's past experience with managing this type of a cache in our database. (For example... I don't see code in this PR to handle actually updating the recent resources. Would this be handled in a future PR?)

A few questions at the top of my mind:

1) Was building a new table the only viable way to implement this? e.g. did we try making direct calls to Rawls and our workspace table, and it was just too slow? (I might hope / suggest that the Terra performance work may make those Rawls calls to list workspaces much, much faster in the near future.)
2) Would it be worth considering a client-side caching approach (e.g. using localStorage or just an in-memory cache) instead of a database layer?
3) What design or implementation decisions can we make to prevent this table from suffering the same fate as our old userRecentResources table.

Sorry to jump on this after this PR is almost up for review – I just want to make sure we're thinking this through fully before we commit to building another caching layer into our system!

gjuggler:
> I think this needs a quick design check-in; let's grab a few minutes ad-hoc today and/or I'll set up some time for today or tomorrow if possible.

We chatted offline – tl;dr my concerns were mostly / grossly over-stated. I feel much better about the approach after chatting for a few minutes. To summarize:

- The primary purpose of the new table is to track per-user last-accessed / last-edited time – this data isn't available either in Rawls or in our workspaces table (since a workspace is collaborative, its overall last-updated time may not be the same as each user's last-updated time).
- The table that was deleted in #2243 was mainly used to track per-user ACLs on workspaces, not last-accessed times. Ariel pointed out that we do still have the user_recent_resources table, which tracks per-user accessed times for notebooks, cohorts, etc. within a workspace. This table can be thought of as a workspace-level parallel to that, and unrelated to the table that's now gone.
- There's still a chance of this table having stale references to workspaces that are no longer accessible to a user. Ideally, we'd first check with Rawls to get the full list of accessible workspaces, and then combine that list w/ the last-updated times from the table in this PR. This may depend on Rawls' performance.
- It may be easier to keep this table updated if it's defined to be the user's last-accessed-time (e.g. the last time they viewed the workspace or anything inside it), rather than the last-edited time (e.g. the last time they edited and saved any item within the workspace).

jaycarlton:
Does Leonardo or Rawls have a notion of lifecycle event hooks? I.e. we
could give them an API to hit when a cluster gets birthed, sleeps, or is
killed permanently, and we could update our state of the world accordingly.

-Jay

api/src/main/java/org/pmiops/workbench/api/CohortAnnotationDefinitionController.java
```diff
@@ -136,8 +142,11 @@ private void validateColumnName(String columnName) {
               "Conflict: Cohort Annotation Definition name exists for: %s",
               request.getColumnName()));
     }
+
+    Workspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
```
jaycarlton: Can you rename `workspaceId` to `firecloudWorkspaceName`?

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -32,11 +32,7 @@
 import org.pmiops.workbench.cohortreview.ReviewQueryBuilder;
 import org.pmiops.workbench.cohortreview.util.ParticipantCohortStatusDbInfo;
 import org.pmiops.workbench.db.dao.UserRecentResourceService;
-import org.pmiops.workbench.db.model.Cohort;
-import org.pmiops.workbench.db.model.CohortReview;
-import org.pmiops.workbench.db.model.ParticipantCohortStatus;
-import org.pmiops.workbench.db.model.ParticipantCohortStatusKey;
-import org.pmiops.workbench.db.model.User;
+import org.pmiops.workbench.db.model.*;
```
jaycarlton: nit: check IJ settings. Some folks are militant about not importing *, but it's actually an area I don't have an opinion on.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -375,6 +384,9 @@ public void setUserProvider(Provider<User> userProvider) {
     cohortReviewService.deleteParticipantCohortAnnotation(
         annotationId, cohortReviewId, participantId);
 
+    Workspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
+    workspaceService.updateRecentWorkspaces(workspace.getWorkspaceId(), userProvider.get().getUserId());
```
jaycarlton: Should we maybe be updating this new table and the main workspace table in a single transaction?

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -705,6 +722,9 @@ public void setUserProvider(Provider<User> userProvider) {
     cohortReview.incrementReviewedCount();
     cohortReviewService.saveCohortReview(cohortReview);
 
+    Workspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
+    workspaceService.updateRecentWorkspaces(workspace.getWorkspaceId(), userProvider.get().getUserId());
```
jaycarlton: It looks like you do this in many places. I'd make a local helper function that just takes the workspace firecloud name and namespace, since it can fetch the workspace and userid itself. So you'd just have `this.updateRecentWorkspaces(workspaceNamespace, workspaceFirecloudName)` or some such.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -138,7 +138,7 @@
     Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
-    long wId = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
+    long dbWorkspaceId = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
```
jaycarlton: nit: I like to mark everything final except when you definitely need to reassign it. Reasonable folks have disagreed in the case of locals though, since `final` is such a long modifier.

ui/README.md
```diff
@@ -25,3 +25,7 @@ To run UI unit tests:
 or to run tests headless:
 
 `yarn install && yarn test --browsers=ChromeHeadless`
+
+To regenerate TypeScript APIs after you've changed them in Java:
+
+`yarn codegen`
```
jaycarlton: in the ui directory?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -503,19 +504,21 @@ public Workspace setPublished(
   }
 
   @Override
-  public UserRecentWorkspace updateRecentWorkspaces(long workspaceId, long userId, Timestamp lastAccessDate) {
+  public UserRecentWorkspace updateRecentWorkspaces(long workspaceId, long userId) {
```
jaycarlton: One thing to think about, which may not matter, is that if you don't pass in a timestamp here, the access times recorded somewhere else will probably not match what's put in the table exactly. For example, if the workspace had a last accessed time (regardless of user), it might be off by a few milliseconds from what's here. Probably ok for this application, but it made me scratch my head for a minute.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -526,4 +529,13 @@ public UserRecentWorkspace updateRecentWorkspaces(long workspaceId, long userId,
       return recentWorkspace;
     }
   }
+
+  @Override
+  public void maybeDeleteRecentWorkspace(long workspaceId, long userId) {
+    List<UserRecentWorkspace> maybeRecentWorkspaces = userRecentWorkspaceDao.findByWorkspaceIdAndUserId(workspaceId, userId);
```
jaycarlton: If you like streams, you can do this whole function in one statement.

jaycarlton: I don't understand why there'd ever be more than one such entry. I.e. I'd expect to see `Optional<UserRecentWorkspace>` for this function. (Maybe you need to limit 1 on the query for that to work).

calbach:
- We have an existing `user_recent_resource` table which is meant to track effectively the same thing at a resource level. Did you consider just grabbing distinct workspace IDs from there instead? This wouldn't exactly capture any access to a workspace, but it might suffice depending on the product requirements (which also - I'll bet are very squishy, following up on ticket).
- Injecting this into every API seems burdensome; is this something that could be either something that lives on a relatively central API used from all workspace UI screens, e.g. getWorkspace(), or alternatively is this something that could be explicitly posted from the UI where needed?

als364:
- `user_recent_resource` doesn't track the same thing. It tracks when all those resources were updated (but iirc claims it's tracking when they were last accessed). It also doesn't cover all resource types that are owned by a workspace. Further, it wouldn't capture viewing or editing workspace metadata, like the name, research purpose, etc., or just looking at the workspace data / analysis pages.

- Yeah, I'm going to move to calling this from the UI - we just don't have a pattern of doing that sort of thing and we really should

als364:
@calbach You may find [this context](https://github.com/all-of-us/workbench/pull/2590#issuecomment-533125383) useful.

calbach:
Thanks, I read that before posting.

- `user_recent_resource` is supposed to track last accessed. If it isn't doing that, that's a bug. Agreed it doesn't handle simply looking at a workspace, e.g. metadata. Is that important - do we have actual requirements that are not on the ticket? (the ticket says something entirely different)
- Calling from the UI: if this would supplant all the unrelated API changes, that addresses my concerns. I think product requirements are likely to evolve, so having the separate table (not using `user_recent_resource`) is fine - I just want some record of this alternative and why it was rejected.

als364:
- `user_recent_resource` calls the column `lastAccessDate` and then updates the table on resource update.
- The ticket says `See my recently accessed workspace (like Google Docs)`. The idea is to display the last four workspaces a user looked at, either directly or by jumping to one of the resources associated with that workspace.

calbach:
- I understand, I'm asserting this is not working as intended
- I get that's the gist of the ticket, was asking whether there are more concrete requirements than `"like Google Docs"` - there's certainly a lot of room for interpretation on that. (the ticket only expands on this by saying "last updated workspaces" - which I think is just a mistake).

als364:
We want to display the four most recently accessed workspaces, where 'accessed' is defined as:
- looking at the workspace, in the UI
- looking at any resource owned by the workspace, in the UI

We want to avoid updating the table upon calling some random GET endpoint that fetches that workspace or resource, unless the user specifically took action to do so.

calbach:
Thanks for that, those requirements sound great. :+1: to the "call from UI" approach replacing all of the server-side hooks.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +227,66 @@ public void activeStatus() {
...
+          workspaceService.updateRecentWorkspaces(
+              workspace.getWorkspaceId(),
+              USER_ID,
+              Timestamp.from(NOW.minusSeconds(workspaces.size() - workspace.getWorkspaceId())));
+        });
+    List<UserRecentWorkspace> recentWorkspaces =
+        workspaceService.getRecentWorkspacesByUser(USER_ID);
+    assertThat(recentWorkspaces.size()).isEqualTo(WorkspaceServiceImpl.RECENT_WORKSPACE_COUNT);
+    assertThat(
+            recentWorkspaces.stream()
```
jaycarlton: I'd make variables for these two lists like `expected` and `actual` and just do the assertion at the end so it's all on one line. Easier to read with fewer indentations plus it's simpler to step through in the debugger.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -159,7 +227,66 @@ public void activeStatus() {
...
+        sharedId,
+        "shared",
+        "shared_namespace",
+        WorkspaceAccessLevel.NO_ACCESS,
+        WorkspaceActiveStatus.ACTIVE);
+    workspaceService.updateRecentWorkspaces(sharedId, USER_ID, Timestamp.from(NOW));
+
+    List<UserRecentWorkspace> recentWorkspaces =
+        workspaceService.getRecentWorkspacesByUser(USER_ID);
+    assertThat(recentWorkspaces.size()).isEqualTo(1);
```
jaycarlton: I'd also assert that the one left after deleting has the ID you expect.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -471,7 +442,7 @@ private Workspace createWorkspace(String workspaceNameSpace, String workspaceNam
     researchPurpose.setReviewRequested(true);
     researchPurpose.setApproved(false);
     Workspace workspace = new Workspace();
-    workspace.setId(workspaceName);
+    workspace.setId("1");
```
jaycarlton: Is this really a number as a string (from firecloud)?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +496,89 @@ public Workspace setPublished(
...
+                .map(UserRecentWorkspace::getWorkspaceId)
+                .collect(Collectors.toList()));
+
+    // Map by primary key id bc it's illegal to string streams along in Java. not pretty enough.
+    Map<Long, Workspace> dbWorkspacesById =
+        dbWorkspaces.stream()
+            .collect(Collectors.toMap(Workspace::getWorkspaceId, dbWorkspace -> dbWorkspace));
+
+    // Unfortunately it doesn't look like there's a way to batch call this currently. At least we
+    // can make it parallel.
```
jaycarlton: Is this already parallel or is that the goal? From what I can tell `forEach` is sequential and single-threaded.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -739,4 +722,55 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+    recentWorkspace.setWorkspace(workspaceMapper.toApiWorkspace(dbWorkspace));
+    recentWorkspace.setAccessedTime(userRecentWorkspace.getLastAccessDate().toString());
+    recentWorkspace.setAccessLevel(accessLevel);
+    return recentWorkspace;
+  }
+
+  @Override
+  public ResponseEntity<RecentWorkspaceResponse> updateRecentWorkspaces(
+      String workspaceNamespace, String workspaceId) {
+    long userId = userProvider.get().getUserId();
```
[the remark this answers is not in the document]
als364: I can do this, but I'd want to overload the service function for ease of testing.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
jaycarlton: Looks like we don't have a convention on this, so star away.

calbach: There's a team convention around not using `*` imports. However, this wasn't written down anywhere so arguably it's not official.

api/db/changelog/db.changelog-97-user-recent-workspace.xml
```diff
@@ -0,0 +1,33 @@
...
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+    <changeSet author="als364" id="db.changelog-97-user-recent-workspace">
+        <createTable tableName="user_recent_workspace">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="user_id" type="bigint">
```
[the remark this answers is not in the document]
jaycarlton: Oh really? I'd think you could just have an ORM object with those fields and then a helper class for finding the oldest one (and its "slot" like 0, 1, 2, 3). I thought it'd be something like 
```
recentAccess.slot1workspaceId = currrent.workspaceId;
recentAccess.slot1timestamp = currrent.timestamp;
```
and you'd figure out which slot to overwrite by comparing the timestamps, maybe with a multimap from timestamp to slot(s).

But it is a funny pattern. It's just a slight smell that we try to maintain a fixed number of rows in a table matching a key (userId), since that kind of invariant isn't something SQL provides out of the box. OTOH your scheme is better if we wan to change the number of remembered access times.

Note: I don't believe anyone has asserted that the cap at 4 is for security or privacy reasons, because right now the plan is for all accesses to be stored in the workspace audit log.

api/src/main/java/org/pmiops/workbench/db/model/UserRecentWorkspace.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.sql.Timestamp;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
```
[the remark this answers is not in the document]
jaycarlton: No, I'm still learning all of this by reading this PR. I'm thinking there's probably an easier way to do this but it's not in reach at the moment, but don't settle if you're irritated. Irritation is a very important resource that we need to conserve. Sorry for the confusion.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +496,89 @@ public Workspace setPublished(
...
+        userRecentWorkspaceDao.findByUserIdOrderByLastAccessDateDesc(userId);
+    enforceFirecloudAclsInRecentWorkspaces(userRecentWorkspaces);
+    return userRecentWorkspaceDao.findByUserIdOrderByLastAccessDateDesc(userId);
+  }
+
+  private void enforceFirecloudAclsInRecentWorkspaces(List<UserRecentWorkspace> recentWorkspaces) {
+    // Map recentWorkspaces by id because we'll want to access them like that later when we delete
+    Map<Long, UserRecentWorkspace> recentWorkspacesById =
+        recentWorkspaces.stream()
+            .collect(Collectors.toMap(UserRecentWorkspace::getWorkspaceId, rw -> rw));
```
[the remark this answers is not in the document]
jaycarlton: Yep. `ImmutableList.[toImmutableList](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html#toImmutableList--)()` and others for Set & Map.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +496,89 @@ public Workspace setPublished(
...
+
+    // Grab the dbWorkspaces.
+    List<Workspace> dbWorkspaces =
+        workspaceDao.findAllByWorkspaceIdIn(
+            recentWorkspaces.stream()
+                .map(UserRecentWorkspace::getWorkspaceId)
+                .collect(Collectors.toList()));
+
```
jaycarlton: Should you take some action if you can't find all the workspace you were looking for? I.e. should we maybe also remove from the recent access table all workspaces (for all users) that aren't in the DB anymore? There a cron job or cascade that's supposed to handle that, so maybe log an error or warning in this case (i.e. this is a referential integrity problem).

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +496,89 @@ public Workspace setPublished(
...
+    dbWorkspacesById.forEach(
+        (dbWorkspaceId, dbWorkspace) -> {
+          // Grab the ACLs.
+          Map<String, WorkspaceAccessEntry> aclsByEmail =
+              getFirecloudWorkspaceAcls(
+                  dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getFirecloudName());
+
+          // If this user's email isn't something that has access, remove the corresponding
+          // entry from userRecentResources.
+          if (!aclsByEmail.containsKey(userProvider.get().getEmail())
```
jaycarlton: We can fetch the email once before entering the loop.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +496,89 @@ public Workspace setPublished(
...
+              || aclsByEmail
+                  .get(userProvider.get().getEmail())
+                  .getAccessLevel()
+                  .equals(WorkspaceAccessLevel.NO_ACCESS.toString())) {
+            userRecentWorkspaceDao.delete(recentWorkspacesById.get(dbWorkspaceId));
```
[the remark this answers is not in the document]
jaycarlton: I took a swing at it. It looks prettier than it deserves too b/c of the repeated firecloud api calls hidden so neatly inside the stream: https://github.com/all-of-us/workbench/pull/2624

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
jaycarlton: I propose adding it to the Spotless rules and/or IntelliJ settings file if we want to make it a norm.

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentWorkspaceDao.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.List;
+import java.util.Optional;
+import org.pmiops.workbench.db.model.UserRecentWorkspace;
+import org.springframework.data.repository.CrudRepository;
+
+public interface UserRecentWorkspaceDao extends CrudRepository<UserRecentWorkspace, Long> {
+  List<UserRecentWorkspace> findByUserIdOrderByLastAccessDateDesc(long userId);
+
+  Optional<UserRecentWorkspace> findFirstByWorkspaceIdAndUserId(long workspaceId, long userId);
+}
```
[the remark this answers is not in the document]
jaycarlton: For some reason I thought we would be hitting this with a cron job. I don't have any timing data on whether it's worth it, but it would make things a bit easier to see in database logs, firecloud service logs, etc, if it were 1:1. I just have a general aversion to being chatty with the DB.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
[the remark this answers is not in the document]
jaycarlton: I'm slightly confused. I'm saying put a public function on the service. You may already have that.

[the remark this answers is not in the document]
als364: Now I'm confused. What exactly are you envisioning here?

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
als364: Ugh I thought I already got IntelliJ to stop doing this

api/db/changelog/db.changelog-97-user-recent-workspace.xml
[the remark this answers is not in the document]
als364: I'd have to make separate functions for updating and deleting each one (or each combination of them if I wanted to batch it). You can't really create queries on the fly with Spring.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
als364: I don't know about a cron job, but there's certainly not a cascade that's supposed to handle this. We don't have a DB table for who has access to what DB. FireCloud is the source of truth for all workspace permissions.

I don't think this is a referential integrity problem per se. I do delete the referenced `user_recent_workspace` entry when a `workspace` or `user` is deleted. The only stale data in here is a recent workspace that had been shared with the user that is unshared. That's what I'm handling here. If I were to change it, I would handle it at the point where it's being unshared, but we don't have control over all those points so I'd still need to keep this function.

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentWorkspaceDao.java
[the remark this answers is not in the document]
als364: What do you mean by 'hitting this with a cron job' and what were you imagining would be 1:1?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
als364: No way around that, there isn't a batch API for this in FC.

api/src/main/resources/workbench.yaml
```diff
@@ -3549,6 +3577,21 @@ definitions:
...
 
+  RecentWorkspace:
+    type: object
+    properties:
+      workspace:
+        $ref: '#/definitions/Workspace'
+      accessLevel:
+        $ref: '#/definitions/WorkspaceAccessLevel'
+      accessedTime:
+        type: string
```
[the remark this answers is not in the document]
als364: Can you show me an example? I saw a lot of `string` and specifying ISO-8601 (which I can certainly do in my endpoint)

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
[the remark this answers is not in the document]
als364: Ah, I see. I'm not sure how often we build `Response` objects in the Service layer, lemme check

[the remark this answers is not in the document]
als364: Actually, come to think of it, I only want the Service layer returning models, cause services need to call into other services sometimes, and I don't want them to be munging response objects when doing so.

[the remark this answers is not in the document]
jaycarlton: Good point. I guess you could still return a `RecentWorkspace` and build the response in the controller if you want.

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -0,0 +1,59 @@
...
+import {WorkspaceCard} from 'app/pages/workspace/workspace-card';
+import {workspacesApi} from 'app/services/swagger-fetch-clients';
+import {withUserProfile} from 'app/utils';
+
+interface State {
+  loading: boolean;
+  recentWorkspaces: RecentWorkspace[];
+}
+
+export const RecentWorkspaces = withUserProfile()
```
calbach: The API changes are substantial. The UI piece feels like a second PR - I would definitely recommend breaking it out into a second PR for the sake of reviewers and your own sanity. I'm guessing this may have already been discussed elsewhere so forgive me if I missed it.

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -138,7 +138,8 @@ export const ConceptSetDetails = fp.flow(withUrlParams(), withCurrentWorkspace()
     }
 
     componentDidMount() {
-      this.getConceptSet();
+      workspacesApi().updateRecentWorkspaces(this.props.workspace.namespace, this.props.workspace.id)
```
calbach: Did you consider adding this logic centrally instead, e.g. here? https://github.com/all-of-us/workbench/blob/900231bdd366a3143d3c51f030f380521055e8f4/ui/src/app/pages/workspace/workspace-wrapper/component.ts#L23

Any workspace-related component is supposed to be rendered via that codepath. It's in Angular, but we'll likely have a logically corresponding React component once we complete the migration. Alternatively, as part of the withCurrentWorkspace() HOC.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -21,12 +22,9 @@
...
 import org.pmiops.workbench.conceptset.ConceptSetService;
 import org.pmiops.workbench.db.dao.UserDao;
+import org.pmiops.workbench.db.dao.UserRecentWorkspaceDao;
 import org.pmiops.workbench.db.dao.WorkspaceDao;
-import org.pmiops.workbench.db.model.Cohort;
-import org.pmiops.workbench.db.model.ConceptSet;
-import org.pmiops.workbench.db.model.StorageEnums;
-import org.pmiops.workbench.db.model.User;
-import org.pmiops.workbench.db.model.Workspace;
+import org.pmiops.workbench.db.model.*;
```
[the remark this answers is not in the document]
als364: Ugh I thought I got IntelliJ to stop doing that. You can't turn it all the way off unfortunately but you can set the limit at which point it does that to 99

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -67,4 +70,20 @@ WorkspaceACLUpdate updateFirecloudAclsOnUser(
...
+  List<UserRecentWorkspace> getRecentWorkspaces();
+
+  UserRecentWorkspace updateRecentWorkspaces(
+      long workspaceId, long userId, Timestamp lastAccessDate);
+
+  UserRecentWorkspace updateRecentWorkspaces(long workspaceId);
+
+  boolean maybeDeleteRecentWorkspace(long workspaceId);
+
+  List<RecentWorkspace> buildRecentWorkspaceList(List<UserRecentWorkspace> userRecentWorkspaces);
```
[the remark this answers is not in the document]
als364: My understanding was that we wanted Controllers to be very thin wrappers around service calls. Is your reasoning for saying these belong in the controller because they construct the actual objects that get returned to the client? I can certainly move this up in any case

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
     return this.saveWithLastModified(workspace);
   }
+
+  @Override
+  public List<UserRecentWorkspace> getRecentWorkspaces() {
+    long userId = userProvider.get().getUserId();
+    List<UserRecentWorkspace> userRecentWorkspaces =
+        userRecentWorkspaceDao.findByUserIdOrderByLastAccessDateDesc(userId);
+    enforceFirecloudAclsInRecentWorkspaces(userRecentWorkspaces);
+    return userRecentWorkspaceDao.findByUserIdOrderByLastAccessDateDesc(userId);
```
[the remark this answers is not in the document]
als364: What do you mean by 'risky' in this context?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
+  }
+
+  private void enforceFirecloudAclsInRecentWorkspaces(List<UserRecentWorkspace> recentWorkspaces) {
```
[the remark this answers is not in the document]
als364: Thank for the tip, I'll change that

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
+  @Override
+  public UserRecentWorkspace updateRecentWorkspaces(
+      long workspaceId, long userId, Timestamp lastAccessDate) {
+    Optional<UserRecentWorkspace> maybeRecentWorkspace =
+        userRecentWorkspaceDao.findFirstByWorkspaceIdAndUserId(workspaceId, userId);
+    final UserRecentWorkspace matchingRecentWorkspace =
+        maybeRecentWorkspace
+            .map(
+                recentWorkspace ->
+                    new UserRecentWorkspace(
```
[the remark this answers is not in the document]
als364: I'm not sure what you're asking. The point of looking up the row to begin with is to make sure we're not updating instead of duplicating entries in `user_recent_workspace` if, say, you're flipping back and forth between two workspaces for some reason

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
+            .filter(workspace -> userHasWorkspaceAccess(email, workspace))
+            .map(Workspace::getWorkspaceId)
+            .collect(ImmutableList.toImmutableList());
+
+    if (!idsToDelete.isEmpty()) {
+      userRecentWorkspaceDao.deleteByWorkspaceIdIn(idsToDelete);
+    }
+  }
+
+  private boolean userHasWorkspaceAccess(String email, Workspace workspace) {
```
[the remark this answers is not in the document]
als364: I don't think I noticed that function before, thank you

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
+  }
+
+  @Override
+  public List<RecentWorkspace> buildRecentWorkspaceList(
+      List<UserRecentWorkspace> userRecentWorkspaces) {
+    return userRecentWorkspaces.stream()
+        .map(
+            userRecentWorkspace -> {
+              org.pmiops.workbench.db.model.Workspace dbWorkspace =
+                  this.findByWorkspaceId(userRecentWorkspace.getWorkspaceId());
```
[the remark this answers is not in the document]
als364: I didn't know this existed, I'll look into it. Thank you!

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +498,122 @@ public Workspace setPublished(
...
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public RecentWorkspace buildRecentWorkspace(
+      org.pmiops.workbench.db.model.Workspace dbWorkspace,
+      UserRecentWorkspace userRecentWorkspace,
+      WorkspaceAccessLevel accessLevel) {
+    RecentWorkspace recentWorkspace = new RecentWorkspace();
+    recentWorkspace.setWorkspace(workspaceMapper.toApiWorkspace(dbWorkspace));
```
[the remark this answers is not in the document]
als364: How can this work? Those methods are void

ui/src/app/pages/homepage/recent-workspaces.tsx
als364: Not discussed at all... with this many commits it would be infeasible to break out the UI pieces at this point in time.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
calbach: I understand what this was trying to do, I'm asserting it doesn't work. As written currently there's effectively no difference between the if/else cases of this optional handling.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
calbach: They aren't - perhaps you are looking at the `setX` variants, note the lack of `set` in my example code

ui/src/app/pages/homepage/recent-workspaces.tsx
als364: It occurs to me that I didn't word this well - I don't know a good way of telling git to revert or cherrypick only these files (vs only these changesets). If I was only adding new files I'd be fine pulling it out but I have new files and significant changes to existing files.

calbach: I really don't think it's infeasible, there are many ways you could go about doing this and I don't think number of commits has much bearing on it. Here's one way you could do it:

```
# Or merge master
(as/recent-workspaces)$ git rebase master
(as/recent-workspaces)$ git checkout -b as/recent-workspace-ui
(as/recent-workspaces-ui)$ git checkout -
(as/recent-workspace)$ git checkout master ui
# send separate PR for as/recent-workspace-ui based on top of this PR, or just after this PR has merged
```

als364: Cool, didn't know checkout could do individual paths.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
als364: Gotcha, thank you.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
als364: Gotcha - writing a test to cover the flip flopping case now.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
als364: Argh yep that's exactly what I was doing

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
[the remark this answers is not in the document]
calbach: Sorry I missed this one. I don't really share that world view - but I think the comment stands independently.

If a method would only conceivably be used in one place, it makes sense to localize it there, I think that's the case for the controller which calls this method. Alternatively, you could just update the method signature to return the API model instead of the DB model (and make these methods private to the service) - right now the interface has a mix of DB and API.

ui/src/app/pages/homepage/recent-workspaces.tsx
calbach: Thanks!

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
als364: magic! ✨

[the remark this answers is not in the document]
als364: maybe slightly less and/or more magic than I thought, taking this to slack.

als364:
This is ready for another review.

api/src/main/java/org/pmiops/workbench/db/model/UserRecentWorkspace.java
```diff
@@ -0,0 +1,80 @@
...
+  public int hashCode() {
+    return Objects.hash(workspaceId, userId);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof UserRecentWorkspace)) {
+      return false;
+    }
+    UserRecentWorkspace that = (UserRecentWorkspace) obj;
```
calbach: id?

api/src/main/java/org/pmiops/workbench/db/model/UserRecentWorkspace.java
```diff
@@ -0,0 +1,80 @@
...
+    return lastAccessDate;
+  }
+
+  public void setLastAccessDate(Timestamp lastAccessDate) {
+    this.lastAccessDate = lastAccessDate;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(workspaceId, userId);
```
calbach: id? lastAccess?

calbach: Note: I generally just use Intellij's code dropdown for this "generate hashCode/Equals", though it would be nice if we had a library which just handled this for us on value types

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
calbach: Not addressed?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -495,4 +502,98 @@ public Workspace setPublished(
...
+
+  private List<UserRecentWorkspace> enforceFirecloudAclsInRecentWorkspaces(
+      List<UserRecentWorkspace> recentWorkspaces) {
+    List<Workspace> dbWorkspaces =
+        workspaceDao.findAllByWorkspaceIdIn(
+            recentWorkspaces.stream()
+                .map(UserRecentWorkspace::getWorkspaceId)
+                .collect(Collectors.toList()));
+
+    ImmutableList<Long> idsToDelete =
```
calbach: Should collect into a `Set` instead, to avoid the n^2 behavior at L545

PR #2592, opened 2019-09-19
[RW-2796][risk=no] Add support footer to help sidebar

dolbeew:
Last ticket for the help tip sidebar! Positioned at the bottom of the sidebar container, gives the user two options for support:
- FAQs - opens a new tab to the [Research Hub FAQ page](https://www.researchallofus.org/frequently-asked-questions/)
- Contact us - opens the Zendesk widget

Screen Shot 2019-09-19 at 12 46 20 PM

### DIFF
--- a/ui/src/app/components/help-sidebar.tsx
+++ b/ui/src/app/components/help-sidebar.tsx
@@ -101,7 +101,7 @@ const styles = reactStyles({
   footer: {
     position: 'absolute',
     bottom: 0,
-    padding: '1rem 1.5rem 1.5rem 1rem',
+    padding: '1.25rem 0.75rem',
     background: colorWithWhiteness(colors.primary, .8),
   },
   link: {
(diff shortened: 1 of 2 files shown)


ui/src/app/components/help-sidebar.tsx
```diff
@@ -97,10 +98,16 @@ const styles = reactStyles({
     backgroundColor: 'transparent',
     outline: 'none',
   },
-  highlighted: {
-    color: colorWithWhiteness(colors.success, -0.4),
-    backgroundColor: colorWithWhiteness(colors.success, 0.7),
-    display: 'inline-block'
+  footer: {
+    position: 'absolute',
```
blrubenstein: Thinking out Loud: I wonder if we could do this with flex as opposed to absolute/relative positions. Probably not important, just a thought exercise.

dolbeew: For the pages with enough help content that it scrolls within the sidebar, we want the support footer to stay fixed at the bottom and the content to scroll behind it. Not sure flex would do that, although it's possible as I haven't used it very extensively. 
Screen Shot 2019-09-26 at 8 07 49 AM

PR #2605, opened 2019-09-25
[RW-3401][risk=no] Fix incremental builds - don't hardcode java location

calbach:
@blrubenstein @jaycarlton fyi


jaycarlton:
Excellent, thanks.

PR #2606, opened 2019-09-25
[RW-3467][risk=no] adding deceased to elastic person index.

freemabd:
Adding a Boolean elastic type to the person index that is driven by existence of person id in the OMOP death table.

### DIFF
--- a/api/tools/src/main/resources/bigquery/es_person.sql
+++ b/api/tools/src/main/resources/bigquery/es_person.sql
@@ -18,6 +18,7 @@ SELECT
   measurement_concept_ids,
   measurement_source_concept_ids,
   visit_concept_ids,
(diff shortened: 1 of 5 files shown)


api/tools/src/main/resources/bigquery/es_person.sql
```diff
@@ -18,6 +18,7 @@ SELECT
   measurement_concept_ids,
   measurement_source_concept_ids,
   visit_concept_ids,
+  case when death.person_id is null then false else true end as is_deceased,
```
calbach: Can't this be rewritten as `death.person_id is null as is_deceased` ?

freemabd: Good catch. Actually it needs to be death.person_id is not null as is_deceased.

PR #2607, opened 2019-09-25
[RW-3525][Risk=no] Preview per domain

blrubenstein:
The reason to do this is to reduce the chance we would hit the appengine timeout, and make it more likely we don't hit unexpected timeouts. Also, this allows us to display queries that return more quickly to the user before all queries have returned.

Attached is a screenshot of a domain loading:
Screen Shot 2019-09-25 at 11 57 35 AM

For Mocks, those are here:
https://projects.invisionapp.com/d/main?origin=v7#/console/12820961/385557370/inspect

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -49,7 +49,6 @@ import org.pmiops.workbench.model.DataSet;
 import org.pmiops.workbench.model.DataSetCodeResponse;
 import org.pmiops.workbench.model.DataSetExportRequest;
 import org.pmiops.workbench.model.DataSetListResponse;
-import org.pmiops.workbench.model.DataSetPreviewList;
 import org.pmiops.workbench.model.DataSetPreviewRequest;
 import org.pmiops.workbench.model.DataSetPreviewResponse;
 import org.pmiops.workbench.model.DataSetPreviewValueList;
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -264,16 +264,22 @@ public class DataSetController implements DataSetApiDelegate {
   }
 
   // TODO (srubenst): Delete this method and make generate query take the composite parts.
-  private DataSetRequest generateDataSetRequestFromPreviewRequest(DataSetPreviewRequest dataSetPreviewRequest) {
+  private DataSetRequest generateDataSetRequestFromPreviewRequest(
+      DataSetPreviewRequest dataSetPreviewRequest) {
     return new DataSetRequest()
         .name("Does not matter")
         .conceptSetIds(dataSetPreviewRequest.getConceptSetIds())
         .cohortIds(dataSetPreviewRequest.getCohortIds())
         .prePackagedConceptSet(dataSetPreviewRequest.getPrePackagedConceptSet())
         .includesAllParticipants(dataSetPreviewRequest.getIncludesAllParticipants())
-        .values(dataSetPreviewRequest.getValues().stream()
-            .map(value -> new DomainValuePair().domain(dataSetPreviewRequest.getDomain()).value(value))
-            .collect(Collectors.toList()));
+        .values(
+            dataSetPreviewRequest.getValues().stream()
+                .map(
+                    value ->
+                        new DomainValuePair()
+                            .domain(dataSetPreviewRequest.getDomain())
+                            .value(value))
+                .collect(Collectors.toList()));
   }
 
   @Override
@@ -283,10 +289,12 @@ public class DataSetController implements DataSetApiDelegate {
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
     DataSetPreviewResponse previewQueryResponse = new DataSetPreviewResponse();
     DataSetRequest dataSetRequest = generateDataSetRequestFromPreviewRequest(dataSetPreviewRequest);
-    Map<String, QueryJobConfiguration> bigQueryJobConfig = dataSetService.generateQuery(dataSetRequest);
+    Map<String, QueryJobConfiguration> bigQueryJobConfig =
+        dataSetService.generateQuery(dataSetRequest);
 
     if (bigQueryJobConfig.size() > 1) {
-      throw new BadRequestException("There should never be a preview request with more than one domain");
+      throw new BadRequestException(
+          "There should never be a preview request with more than one domain");
     }
     List<DataSetPreviewValueList> valuePreviewList = new ArrayList<>();
 
@@ -397,7 +405,8 @@ public class DataSetController implements DataSetApiDelegate {
 
           Collections.sort(
               valuePreviewList,
-              Comparator.comparing(item -> dataSetPreviewRequest.getValues().indexOf(item.getValue())));
+              Comparator.comparing(
+                  item -> dataSetPreviewRequest.getValues().indexOf(item.getValue())));
         });
 
     previewQueryResponse.setValues(valuePreviewList);
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -283,7 +283,7 @@ public class DataSetController implements DataSetApiDelegate {
   }
 
   @Override
-  public ResponseEntity<DataSetPreviewResponse> generateDataSetPreview(
+  public ResponseEntity<DataSetPreviewResponse> previewDataSetByDomain(
       String workspaceNamespace, String workspaceId, DataSetPreviewRequest dataSetPreviewRequest) {
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
@@ -297,118 +297,110 @@ public class DataSetController implements DataSetApiDelegate {
           "There should never be a preview request with more than one domain");
     }
     List<DataSetPreviewValueList> valuePreviewList = new ArrayList<>();
+    QueryJobConfiguration queryJobConfiguration =
+        bigQueryJobConfig.get(dataSetPreviewRequest.getDomain().toString());
 
-    bigQueryJobConfig.forEach(
-        (domain, queryJobConfiguration) -> {
-          int retry = 0, rowsRequested = NO_OF_PREVIEW_ROWS;
-          String originalQuery = queryJobConfiguration.getQuery();
-          do {
-            try {
-              String query = originalQuery.concat(" LIMIT " + rowsRequested);
+    int retry = 0, rowsRequested = NO_OF_PREVIEW_ROWS;
+    String originalQuery = queryJobConfiguration.getQuery();
+    do {
+      try {
+        String query = originalQuery.concat(" LIMIT " + rowsRequested);
 
-              queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
+        queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
 
-              /* Google appengine has a 60 second timeout, we want to make sure this endpoint completes
-               * before that limit is exceeded, or we get a 500 error with the following type:
-               * com.google.apphosting.runtime.HardDeadlineExceededError
-               * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
-               */
-              TableResult queryResponse =
-                  bigQueryService.executeQuery(
-                      bigQueryService.filterBigQueryConfig(queryJobConfiguration),
-                      APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC / PREVIEW_RETRY_LIMIT + 1);
-              queryResponse
-                  .getSchema()
-                  .getFields()
-                  .forEach(
-                      fields -> {
-                        valuePreviewList.add(new DataSetPreviewValueList().value(fields.getName()));
-                      });
+        /* Google appengine has a 60 second timeout, we want to make sure this endpoint completes
+         * before that limit is exceeded, or we get a 500 error with the following type:
+         * com.google.apphosting.runtime.HardDeadlineExceededError
+         * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
+         */
+        TableResult queryResponse =
+            bigQueryService.executeQuery(
+                bigQueryService.filterBigQueryConfig(queryJobConfiguration),
+                APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC / PREVIEW_RETRY_LIMIT + 1);
+        queryResponse
+            .getSchema()
+            .getFields()
+            .forEach(
+                fields -> {
+                  valuePreviewList.add(new DataSetPreviewValueList().value(fields.getName()));
+                });
 
-              queryResponse
-                  .getValues()
-                  .forEach(
-                      fieldValueList -> {
-                        IntStream.range(0, fieldValueList.size())
-                            .forEach(
-                                columnNumber -> {
-                                  try {
-                                    valuePreviewList
-                                        .get(columnNumber)
-                                        .addQueryValueItem(
-                                            fieldValueList.get(columnNumber).getValue().toString());
-                                  } catch (NullPointerException ex) {
-                                    log.severe(
-                                        String.format(
-                                            "Null pointer exception while retriving value for query: Column %s ",
-                                            columnNumber));
-                                    valuePreviewList.get(columnNumber).addQueryValueItem("");
-                                  }
-                                });
-                      });
-              queryResponse
-                  .getSchema()
-                  .getFields()
-                  .forEach(
-                      fields -> {
-                        DataSetPreviewValueList previewValue =
-                            valuePreviewList.stream()
-                                .filter(
-                                    preview ->
-                                        preview.getValue().equalsIgnoreCase(fields.getName()))
-                                .findFirst()
-                                .get();
-                        if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
-                          List<String> queryValues = new ArrayList<String>();
-                          DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
-                          previewValue
-                              .getQueryValue()
-                              .forEach(
-                                  value -> {
-                                    try {
-                                      Double fieldValue = Double.parseDouble(value);
-                                      queryValues.add(
-                                          dateFormat.format(new Date(fieldValue.longValue())));
-                                    } catch (NumberFormatException ex) {
-                                      queryValues.add("");
-                                    }
-                                  });
-                          previewValue.setQueryValue(queryValues);
-                        }
-                      });
-              break;
-            } catch (Exception ex) {
-              if ((ex.getCause() != null
-                  && ex.getCause().getMessage() != null
-                  && ex.getCause().getMessage().contains("Read timed out"))) {
-                rowsRequested = (rowsRequested / 2);
-                if (rowsRequested == 0) {
-                  throw new GatewayTimeoutException(
-                      "Timeout while querying the CDR to pull preview information.");
-                }
-                retry++;
-              } else {
-                throw ex;
-              }
-            }
-          } while (retry < PREVIEW_RETRY_LIMIT);
-
-          if (retry == PREVIEW_RETRY_LIMIT) {
+        queryResponse
+            .getValues()
+            .forEach(
+                fieldValueList -> {
+                  IntStream.range(0, fieldValueList.size())
+                      .forEach(
+                          columnNumber -> {
+                            try {
+                              valuePreviewList
+                                  .get(columnNumber)
+                                  .addQueryValueItem(
+                                      fieldValueList.get(columnNumber).getValue().toString());
+                            } catch (NullPointerException ex) {
+                              log.severe(
+                                  String.format(
+                                      "Null pointer exception while retriving value for query: Column %s ",
+                                      columnNumber));
+                              valuePreviewList.get(columnNumber).addQueryValueItem("");
+                            }
+                          });
+                });
+        queryResponse
+            .getSchema()
+            .getFields()
+            .forEach(
+                fields -> {
+                  DataSetPreviewValueList previewValue =
+                      valuePreviewList.stream()
+                          .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+                          .findFirst()
+                          .get();
+                  if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
+                    List<String> queryValues = new ArrayList<String>();
+                    DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+                    previewValue
+                        .getQueryValue()
+                        .forEach(
+                            value -> {
+                              try {
+                                Double fieldValue = Double.parseDouble(value);
+                                queryValues.add(
+                                    dateFormat.format(new Date(fieldValue.longValue())));
+                              } catch (NumberFormatException ex) {
+                                queryValues.add("");
+                              }
+                            });
+                    previewValue.setQueryValue(queryValues);
+                  }
+                });
+        break;
+      } catch (Exception ex) {
+        if ((ex.getCause() != null
+            && ex.getCause().getMessage() != null
+            && ex.getCause().getMessage().contains("Read timed out"))) {
+          rowsRequested = (rowsRequested / 2);
+          if (rowsRequested == 0) {
             throw new GatewayTimeoutException(
                 "Timeout while querying the CDR to pull preview information.");
           }
+          retry++;
+        } else {
+          throw ex;
+        }
+      }
+    } while (retry < PREVIEW_RETRY_LIMIT);
 
-          final List<String> requestValues =
-              dataSet.getDomainValuePairs().stream()
-                  .map(DomainValuePair::getValue)
-                  .collect(Collectors.toList());
+    if (retry == PREVIEW_RETRY_LIMIT) {
+      throw new GatewayTimeoutException(
+          "Timeout while querying the CDR to pull preview information.");
+    }
 
-          Collections.sort(
-              valuePreviewList,
-              Comparator.comparing(
-                  item -> dataSetPreviewRequest.getValues().indexOf(item.getValue())));
-        });
+    Collections.sort(
+        valuePreviewList,
+        Comparator.comparing(item -> dataSetPreviewRequest.getValues().indexOf(item.getValue())));
 
+    previewQueryResponse.setDomain(dataSetPreviewRequest.getDomain());
     previewQueryResponse.setValues(valuePreviewList);
     return ResponseEntity.ok(previewQueryResponse);
   }
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -272,7 +272,7 @@ public class DataSetController implements DataSetApiDelegate {
         .cohortIds(dataSetPreviewRequest.getCohortIds())
         .prePackagedConceptSet(dataSetPreviewRequest.getPrePackagedConceptSet())
         .includesAllParticipants(dataSetPreviewRequest.getIncludesAllParticipants())
-        .values(
+        .domainValuePairs(
             dataSetPreviewRequest.getValues().stream()
                 .map(
                     value ->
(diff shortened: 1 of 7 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -2964,19 +2964,45 @@ definitions:
...
+          All the selected values in the data set
+        items:
+          type: string
 
-  DataSetPreviewList:
+
+  DataSetPreviewResponse:
     type: object
     properties:
-      domain:
```
blrubenstein: Semantically, I'm not sure if this should return a domain. It should be deterministic from the request sent up, so should be redundant information.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -134,7 +135,11 @@ export const styles = reactStyles({
     display: 'flex',
     flexDirection: 'column',
     alignItems: 'center',
-    height: '10rem'
+    height: '10rem',
+    marginTop: '2rem',
```
blrubenstein: Moving these styles in with this style class, because it is only used in one place.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -604,16 +628,24 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          {filteredPreviewData.values.map(value =>
+            <Column key={value.value} header={this.getHeaderValue(value)}
+                    headerStyle={{textAlign: 'left', width: '5rem'}} style={{width: '5rem'}}
+                    field={value.value}/>
+          )}
+        </DataTable> :
+        <div style={styles.warningMessage}>
+          {filteredPreviewData.isLoading ?
+            <div>Generating preview for {domainDisplayed}</div> :
+            <div>No values found for {domainDisplayed}</div>
```
blrubenstein: This is likely an error case.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -604,16 +628,24 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-        {filteredPreviewData.values.map(value =>
-          <Column key={value.value} header={this.getHeaderValue(value)}
-                  headerStyle={{textAlign: 'left', width: '5rem'}} style={{width: '5rem'}}
-                  field={value.value}/>
-        )}
-      </DataTable>;
+            preview => preview.domain === this.state.selectedPreviewDomain)[0];
+      const domainDisplayed = this.state.selectedPreviewDomain.toString().toLowerCase();
+      return filteredPreviewData.values.length > 0 ?
+        <DataTable ref={el => this.dt = el} key={this.state.selectedPreviewDomain}
```
ericsong: i think this will be quicker to read if we pull out this rendering logic into a function that takes in a `filteredPreviewData` object and returns the JSX.

 I think it would make the overview of this render function easy to read at a glance since it simplifies down to 
- find your data
- if it exists, render it
  - if not, loading message

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -781,35 +813,38 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-                              fontWeight: (selectedPreviewDomain === previewRow.domain)
-                               ? 600 : 400,
-                              fontSize: '18px',
-                              display: 'flex',
-                              justifyContent: 'center',
-                              lineHeight: '32px',
-                            }}>
-                       {previewRow.domain === 'OBSERVATION' ? 'SURVEY' : previewRow.domain}
-                     </Clickable>
+                     <TooltipTrigger content={'Preview for domain ' + previewRow.domain.toString()
```
ericsong: we call `toLowerCase()` in the above render function. Should we also do it here?

ericsong: nit - if so, could warrant a small domain -> displayText function

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-              errorText = 'Query to load data from the All of Us Database timed out. ' +
-                'Please either try again or export data set to a notebook to try there';
-            }
-            break;
-          default:
-            errorText = 'An unexpected error has occurred. ' +
-              'Please submit a bug using the contact support button';
+      const domains = fp.uniq(this.state.selectedValues.map(domainValue => domainValue.domain));
+      this.setState({
+        previewList: domains.map(domain => {
```
ericsong: I don't know if the requirements may change in the future where multiple domains would make sense but it seems like a map is the more natural data structure for the current access patterns

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -224,6 +229,12 @@ const BoxHeader = ({text= '', header =  '', subHeader = '', style= {}, ...props}
   </div>;
 };
 
+interface DataSetPreviewList {
```
ericsong: is this used anywhere? I was going to ask about value being a list of lists but I couldn't find usages of this type.

api/src/main/resources/workbench.yaml
```diff
@@ -2036,15 +2036,15 @@ paths:
         - dataSet
       description: >
         Preview data set
-      operationId: "previewQuery"
+      operationId: "generateDataSetPreview"
```
NehaBroad: since this is done per domain, should the name also specify that like generatePreviewByDomain?

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -224,6 +229,12 @@ const BoxHeader = ({text= '', header =  '', subHeader = '', style= {}, ...props}
...
+  domain: Domain;
+  isLoading: boolean;
+  values?: Array<DataSetPreviewValueList>;
+}
+
 interface Props {
```
NehaBroad: nit: can we change the name of Props and state as per the naming conventions DataSetProps, data setstate

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          domain: domain,
+          conceptSetIds: this.state.selectedConceptSetIds,
+          includesAllParticipants: this.state.includesAllParticipants,
+          cohortIds: this.state.selectedCohortIds,
+          prePackagedConceptSet: this.getPrePackagedConceptSet(),
+          values: this.state.selectedValues.map(domainValue => domainValue.value)
+        };
+        try {
+          const dataSetPreviewResp = await dataSetApi().generateDataSetPreview(namespace, id, request);
+          const domainPreviewIndex = '[' + fp.findIndex(['domain', domain], this.state.previewList) + ']';
```
NehaBroad: Can you add comments to explain this

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          this.setState({previewList: fp.set(domainPreviewIndex + '.isLoading', false,
+            fp.set(domainPreviewIndex + '.values', dataSetPreviewResp.values, this.state.previewList))});
+        } catch (ex) {
+          const exceptionResponse = await ex.json() as unknown as ErrorResponse;
+          let errorText: string;
+          switch (exceptionResponse.statusCode) {
+            case 400:
+              if (exceptionResponse.message ===
+                'Data Sets must include at least one cohort and concept.') {
```
NehaBroad: nit: can all these strings be const

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -781,35 +813,38 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+                     + ' is still loading. It may take up to one minute'}
+                                     disabled={!previewRow.isLoading}
+                                     side='top'>
+                       <Clickable key={previewRow.domain.toString()}
+                            disabled={previewRow.isLoading}
+                            onClick={() =>
+                              this.setState({selectedPreviewDomain: previewRow.domain})}
+                              style={{
+                                marginLeft: '0.2rem', color: colors.accent,
+                                marginRight: '0.25rem', paddingBottom: '0.25rem',
```
NehaBroad: More than 3 attributes, can we move this to style

ui/src/app/pages/data/data-set/dataset-page.tsx
NehaBroad: Can we rename the interface name to DataSetPageProps

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -254,17 +254,44 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+                        new DomainValuePair()
+                            .domain(dataSetPreviewRequest.getDomain())
+                            .value(value))
+                .collect(Collectors.toList()));
+  }
+
   @Override
-  public ResponseEntity<DataSetPreviewResponse> previewQuery(
-      String workspaceNamespace, String workspaceId, DataSetRequest dataSet) {
+  public ResponseEntity<DataSetPreviewResponse> generateDataSetPreview(
```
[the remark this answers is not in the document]
blrubenstein: Sure, will rename.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -254,17 +254,44 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
+    Map<String, QueryJobConfiguration> bigQueryJobConfig =
+        dataSetService.generateQuery(dataSetRequest);
+
+    if (bigQueryJobConfig.size() > 1) {
+      throw new BadRequestException(
+          "There should never be a preview request with more than one domain");
+    }
+    List<DataSetPreviewValueList> valuePreviewList = new ArrayList<>();
+
     bigQueryJobConfig.forEach(
```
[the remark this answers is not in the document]
blrubenstein: I can just make it `.get(0)`. It is technically a list but that would make it more clear at least?

api/src/main/resources/workbench.yaml
blrubenstein: I would prefer it specify data set. I like `previewDataSetByDomain` though. Will change to that.

api/src/main/resources/workbench.yaml
blrubenstein: The domain is not in `DataSetPreviewValueList`, so I will add domain back in.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: This is used in this class. Since its no longer generated (because it is just client side structuring now), I redefined the interface to also include useful information (like isLoading and possibly isErrored)

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: I thought the convention was to name it Props, and then export as `DataSetPageProps`? Is that not the case?

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: The nice thing about a list is that it is ordered. I don't know if we care about ordering here, but that could be a valuable thing to maintain. Otherwise I agree with you 100%. Let's chat faces?

(Also, for what it's worth, we do support multiple domains. They are just loaded individually from the API as opposed to all at once)

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          conceptSetIds: this.state.selectedConceptSetIds,
+          includesAllParticipants: this.state.includesAllParticipants,
+          cohortIds: this.state.selectedCohortIds,
+          prePackagedConceptSet: this.getPrePackagedConceptSet(),
+          values: this.state.selectedValues.map(domainValue => domainValue.value)
+        };
+        try {
+          const dataSetPreviewResp = await dataSetApi().generateDataSetPreview(namespace, id, request);
+          const domainPreviewIndex = '[' + fp.findIndex(['domain', domain], this.state.previewList) + ']';
+          this.setState({previewList: fp.set(domainPreviewIndex + '.isLoading', false,
```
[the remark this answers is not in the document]
blrubenstein: I actually tried using that, the problem is it also has to pick out the right domain from the list, which is what the above code is doing. I can add a comment, and poke around a bit, but the finding the right object is the trickier part here, since previewList is a list. This could be an argument to move to a map.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Can you explain what you mean? They are currently constants. Do you mean pulling them out to be top level const variables?

I would push back a bit against that, because they are currently used only here for error handling, and I would worry that dong that would add a layer of unnecessary abstraction.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+                errorText = `Query to load data from the All of Us Database timed out for domain:
+                  ${domain}. Please either try again or export data set to a notebook to try
+                  there`;
+              }
+              break;
+            default:
+              errorText = `An unexpected error has occurred loading domain: ${domain}.
+              Please submit a bug using the contact support button`;
+          }
+          this.setState({previewError: true, previewErrorText: errorText});
```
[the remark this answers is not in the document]
blrubenstein: Will do.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          console.error(ex);
```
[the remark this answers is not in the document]
blrubenstein: I don't know, and I think it would be worth removing this. It is mostly useful in debugging to see the error on the page. Mixed feelings.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+        } finally {
+          this.setState({previewDataLoading: false});
```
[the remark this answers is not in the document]
blrubenstein: Yes, but on thinking more about this, we may not even need this piece of state anymore, because it is derivable from the `isLoading` bool on each domain.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -604,16 +628,24 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
     renderPreviewDataTable() {
       const filteredPreviewData =
           this.state.previewList.filter(
```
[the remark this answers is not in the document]
blrubenstein: Will change.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: No, because there is still a large chance of getting errors on this page. Just because the bigquery call can time out. That said, maybe the right thing to do instead of an isLoading is making an `isErrored` flag for that case? I don't have too many strong opinions there.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Yes. Will add the function to do that.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Will move.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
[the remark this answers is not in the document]
blrubenstein: Having done the get, it is objectively better! :D

ui/src/app/pages/data/data-set/dataset-page.tsx
jaycarlton: This error message (in US English) is NOT to be construed as an inviolable part of the contract. If we need fine-grained validation, let's make a separate return structure that includes that information. We could easily change the wording on the back end without finding this.

NehaBroad: > Can you explain what you mean? They are currently constants. Do you mean pulling them out to be top level const variables?
> 
> I would push back a bit against that, because they are currently used only here for error handling, and I would worry that dong that would add a layer of unnecessary abstraction.

I meant pulling at the top level it was more for readability.

ui/src/app/pages/data/data-set/dataset-page.tsx
NehaBroad: Had a chat about this offline, we are going with the way this is.

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
blrubenstein: Changing to leave it for now, if we decide it would be worth removing, that's fine, but right now I think it is a reasonable indicator and relatively standard in our codebase.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: @jaycarlton there's a ticket to build out that structure (with a clear error code that we can parse) That's something that may warrant a bit of a design before doing though.

@NehaBroad I think that pulling it out to the top level would reduce some of the readability here, but am happy to discuss either with faces or more here. Since if we pulled it out it would need to be clear that these are the various preview data error messages, whereas if we keep them here then it is clear via variable placement.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-            if (exceptionResponse.message ===
-              'Timeout while querying the CDR to pull preview information.') {
-              errorText = 'Query to load data from the All of Us Database timed out. ' +
-                'Please either try again or export data set to a notebook to try there';
-            }
-            break;
-          default:
-            errorText = 'An unexpected error has occurred. ' +
-              'Please submit a bug using the contact support button';
+      const domains = fp.uniq(this.state.selectedValues.map(domainValue => domainValue.domain));
```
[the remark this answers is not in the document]
blrubenstein: The `selectedValues` object is generated, although we could create a wrapper. I have no strong objections to doing that, but it would add a layer of abstraction around the DomainValuePair object.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-3006&epics=visible&search=errorcode

@jaycarlton FYI

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
ericsong: `previewList = new Map(this.state.selectedValues.map(selection => [selection.domain, {isLoading: true, values: []}]))`

do you know if this would work? Based off of this - https://stackoverflow.com/questions/26264956/convert-object-array-to-hash-map-indexed-by-an-attribute-value-of-the-object

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,59 +542,75 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+        selectedPreviewDomain: domains[0]
+      });
+      domains.forEach(async domain => {
+        const request: DataSetPreviewRequest = {
+          domain: domain,
+          conceptSetIds: this.state.selectedConceptSetIds,
+          includesAllParticipants: this.state.includesAllParticipants,
+          cohortIds: this.state.selectedCohortIds,
+          prePackagedConceptSet: this.getPrePackagedConceptSet(),
+          values: this.state.selectedValues.map(domainValue => domainValue.value)
```
ericsong: not something we can fix in this PR but making a note here that I think `Value` as a name for this concept is kind of confusing since it doesn't really provide any context. I actually thought it was another name for domain until I got to this line.

I'm wondering if our users also get confused by this. Maybe we can work with Lou and Karthik to come up with a better name.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -781,41 +834,34 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
               </Clickable>
             </div>
-            {previewDataLoading && <div style={styles.warningMessage}>
-              <Spinner style={{position: 'relative', top: '2rem'}} />
-              <div style={{top: '3rem', position: 'relative'}}>
-                It may take up to few minutes to load the data
-              </div>
-            </div>}
-            {previewList.length > 0 &&
+            {fp.toPairs(previewList).length > 0 &&
```
ericsong: nit: is it necessary to convert to `toPairs` to check the size? Would a `!previewList.isEmpty()` guarantee the non emptiness of `toPairs`?

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -781,41 +834,34 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
-                              justifyContent: 'center',
-                              lineHeight: '32px',
-                            }}>
-                       {previewRow.domain === 'OBSERVATION' ? 'SURVEY' : previewRow.domain}
-                     </Clickable>
-                  )}
+                  {fp.toPairs(previewList).map((value) => {
+                    const domain: string = value[0];
+                    const previewRow: DataSetPreviewInfo = value[1];
+                    return <TooltipTrigger content={'Preview for domain ' + fp.capitalize(domain)
```
ericsong: do you want to use your new `formatDomain` function here?

ui/src/app/pages/data/data-set/dataset-page.tsx
jaycarlton: Right, we've chatted about that one. 

In this case, I believe what we need is a separate dataset request validation API that you call before the dataset generation, and which could have detailed error information. Then the main API could just return 400. There are lots of special rules here that don't apply broadly, so I feel like that other ticket is not quite what we need.

Regarding making a constant, I 100% agree we should not have naked string literals in the middle of business logic. It makes translation much harder, for one thing. It also makes the code less reusable. But the real problem here is that you're trying to scan the error text (intended for humans) and take action on it, and that's at the very least a strong code smell. Either the message is good enough for the UI and we just show it verbatim, or we add finer-grained structure to give you enough info to take the right actions. Ultimately, I think driving the UI logic via positive api responses is much more stable; error handling changes frequently (even unintentionally), and it can be more art than science determining which 4xx or 5xx is appropriate.

ui/src/app/pages/data/data-set/dataset-page.tsx
jaycarlton: +1 Any variable that ends in `data`, `value`, `object` needs a better name or better abstraction. Likewise any function that starts with `process`, though I've used that to highlight things that need refactoring.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+        } catch (ex) {
+          const exceptionResponse = await ex.json() as unknown as ErrorResponse;
+          let errorText: string;
+          switch (exceptionResponse.statusCode) {
+            case 400:
+              if (exceptionResponse.message ===
+                'Data Sets must include at least one cohort and concept.') {
+                errorText = exceptionResponse.message;
+              } else if (exceptionResponse.message ===
+                'Concept Sets must contain at least one concept') {
```
jaycarlton: We could easily build this message (in either style) on the backend and the UI's job is just to print the message.

ui/src/app/pages/data/data-set/dataset-page.tsx
jaycarlton: The object-oriented scheme would be to have one abstract widget to go in this div with three subclasses for `loading`, `error`, and `loadedCorrectly`.

But yeah, if a domain has no values, I think that's a programming error, so we can throw a scary generic operation failed message and log an error. We should also have tests to verify that can't happen.

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
jaycarlton: Hang on. Let's meet and discuss the way we return this information first. The UI knows entirely too much here. Changing the response structure on the data set controller to give everything you need (in a computer-friendly structure) is not difficult; I'll volunteer as tribute to tackle that.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Got it, I am happy to move this to a constant string. Will do.

Trying to scan the error text is 100% the wrong thing. I agree with you completely. I would much prefer us to move to an error code based parsing, as opposed to error text. 

I think I strongly prefer having the API call itself do the validation, as it moves us more into making this API endpoint a reusable piece of code in other places. If another institute decided to build a UI on top of our API (I find this very unlikely, but it is theoretically possible), and used our Data Set Builder, I would hate for them to have to remember to call a separate API endpoint before calling preview. But they should get back a clear and concise immutable value that is part of the API contract, and not error message text that they parse on. That is where I think we should be improving our code, and it contains the logic for validation within the actual endpoint.

blrubenstein: Talked to Jay in a call. We're punting on this for now, but have a ticket to follow up on error responses throughout our codebase.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: `isEmpty()` appears to not exist on either `Map` or `Array` for some reason, going off the MDN documentation. I could do something else, like maybe `previewList.entries().length > 0` but it is functionally equivalent.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Agreed, but definitely not going to try to fix that in this PR

blrubenstein: https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-3596&epics=visible&selectedEpic=RW-3443

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -519,60 +530,73 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+            domain: domain,
+            isLoading: true,
+            values: []
+          };
+        }),
+        previewDataLoading: true,
+        selectedPreviewDomain: domains[0]
+      });
+      domains.forEach(async domain => {
+        const request = {
```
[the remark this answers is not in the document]
blrubenstein: This should be typeable as `DataSetPreviewRequest`. Will see if I can change type.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Removed the string interpretation. I think the text should be fine in all cases still/user-friendly.

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
blrubenstein: So, I tried this, and I am getting some truly obnoxious typing errors:
```ERROR in src/app/pages/data/data-set/dataset-page.tsx(549,17): error TS2345: Argument of type '(Domain | DataSetPreviewInfo)[][]' is not assignable to parameter of type 'ReadonlyArray<[Domain, DataSetPreviewInfo]>'.
  Type '(Domain | DataSetPreviewInfo)[]' is not assignable to type '[Domain, DataSetPreviewInfo]'.
    Property '0' is missing in type '(Domain | DataSetPreviewInfo)[]'```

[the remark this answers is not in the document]
blrubenstein: Its not entirely clear to me why map did it that way, where when I use the explicit constructor for `new Array` it doesn't do it that way, but spits out some other errors.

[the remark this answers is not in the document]
jaycarlton: this would be a useful mob session depending on time. I don't quite follow what we're doing.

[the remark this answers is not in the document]
blrubenstein: Got it. I am happy to do a knowledge sharing session at some point.

PR #2608, opened 2019-09-25
[RW-3435][risk=medium] Billing Project Garbage Collection cron endpoint

jmthibault79:
Cron endpoint for reassigning the Billing Projects associated with deleted Workspaces to "garbage collection" Service Accounts as a workaround to https://precisionmedicineinitiative.atlassian.net/browse/RW-3069

These GCSAs need to be created manually in the appropriate project for each environment and their credentials are copied to the environments' credentials buckets under a `garbage-collection` subdirectory.  When all the GCSAs for an environment become associated with too many groups, we will need to add more of them.

### DIFF
--- a/api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
+++ b/api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="changelog-101-billing-project-garbage-collection">
+
+    <createTable tableName="billing_project_garbage_collection">
+      <column name="fireCloudProjectName" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="owner" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+    </createTable>
+  </changeSet>
+</databaseChangeLog>
--- a/api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
+++ b/api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
@@ -7,7 +7,7 @@
   <changeSet author="thibault" id="changelog-101-billing-project-garbage-collection">
 
     <createTable tableName="billing_project_garbage_collection">
-      <column name="fireCloudProjectName" type="varchar(80)">
+      <column name="firecloud_project_name" type="varchar(80)">
         <constraints nullable="false"/>
       </column>
       <column name="owner" type="varchar(80)">
--- a/api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
@@ -46,7 +46,6 @@ public class FireCloudIntegrationTest {
   @Mock private BillingApi billingApi;
   @Mock private WorkspacesApi workspacesApi;
   @Mock private GroupsApi allOfUsGroupsApi;
-  @Mock private GroupsApi endUserGroupsApi;
   @Mock private WorkspacesApi workspaceAclsApi;
   @Mock private StaticNotebooksApi staticNotebooksApi;
   @Mock private ProfileApi profileApi;
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
@@ -19,8 +19,8 @@ import org.pmiops.workbench.db.dao.BillingProjectGarbageCollectionDao;
 import org.pmiops.workbench.db.model.BillingProjectBufferEntry;
 import org.pmiops.workbench.db.model.BillingProjectBufferEntry.BillingProjectBufferStatus;
 import org.pmiops.workbench.db.model.BillingProjectGarbageCollection;
-import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.exceptions.NotFoundException;
+import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.firecloud.FireCloudServiceImpl;
 import org.pmiops.workbench.google.CloudStorageService;
@@ -126,7 +126,11 @@ public class BillingGarbageCollectionService {
     try {
       final GoogleCredential gcsaCredential =
           garbageCollectionSACredentials.get(garbageCollectionSA);
-      fireCloudService.removeOwnerFromBillingProject(appEngineSA, gcsaCredential, projectName);
+
+      gcsaCredential.refreshToken();
+
+      fireCloudService.removeOwnerFromBillingProject(
+          projectName, appEngineSA, gcsaCredential.getAccessToken());
     } catch (final ExecutionException e) {
       final String msg =
           String.format(
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
@@ -72,7 +72,7 @@ public class BillingGarbageCollectionService {
             });
   }
 
-  // Is the App Engine SA a member of this FireCloud project?
+  // Checks whether the AppEngine service account is a member of this FireCloud project
   private boolean appSAIsMemberOfProject(String billingProject) {
     try {
       fireCloudService.getBillingProjectStatus(billingProject);
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
@@ -150,7 +150,7 @@ public class BillingGarbageCollectionService {
 
   BillingProjectGarbageCollectionResponse deletedWorkspaceGarbageCollection() {
     final List<String> billingProjects =
-        billingProjectGarbageCollectionDao.findBillingProjectsForGarbageCollection();
+        billingProjectBufferEntryDao.findBillingProjectsForGarbageCollection();
 
     final List<GarbageCollectedProject> garbageCollectedProjects =
         billingProjects.stream()
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
@@ -19,7 +19,7 @@ import org.pmiops.workbench.db.dao.BillingProjectGarbageCollectionDao;
 import org.pmiops.workbench.db.model.BillingProjectBufferEntry;
 import org.pmiops.workbench.db.model.BillingProjectBufferEntry.BillingProjectBufferStatus;
 import org.pmiops.workbench.db.model.BillingProjectGarbageCollection;
-import org.pmiops.workbench.exceptions.BadRequestException;
+import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.firecloud.FireCloudServiceImpl;
@@ -96,7 +96,7 @@ public class BillingGarbageCollectionService {
             "No available Garbage Collection Service Accounts.  "
                 + "These GCSAs exceed the configured capacity limit of %d: %s",
             config.garbageCollectionUserCapacity, String.join(", ", config.garbageCollectionUsers));
-    throw new BadRequestException(msg);
+    throw new ServerErrorException(msg);
   }
 
   private GarbageCollectedProject recordGarbageCollection(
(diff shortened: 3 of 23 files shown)


jmthibault79:
Still TODO for story completion: setting up nightly Cron runs, adding GCSAs for non-test environments.

api/src/main/java/org/pmiops/workbench/billing/BillingController.java
```diff
@@ -11,13 +12,23 @@
...
-      BillingProjectBufferService billingProjectBufferService) {
+      BillingProjectBufferService billingProjectBufferService,
+      BillingGarbageCollectionService billingGarbageCollectionService) {
     this.billingAlertsService = billingAlertsService;
     this.billingProjectBufferService = billingProjectBufferService;
+    this.billingGarbageCollectionService = billingGarbageCollectionService;
+  }
+
+  @Override
+  public ResponseEntity<BillingProjectGarbageCollectionResponse> billingProjectGarbageCollection() {
```
jmthibault79: alternately: this could go into a new `BillingProjectGarbageCollectionController` ?

api/src/main/java/org/pmiops/workbench/db/model/BillingProjectBufferEntry.java
```diff
@@ -25,12 +25,18 @@
   private User assignedUser;
 
   public enum BillingProjectBufferStatus {
-    CREATING, // Sent a request to FireCloud to create a BillingProject. Status of BillingProject is
-    // TBD
+    // Sent a request to FireCloud to create a BillingProject. Status of BillingProject is TBD
```
jmthibault79: Unrelated, but this felt clearer to me

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -55,12 +61,15 @@
   private final Provider<StaticNotebooksApi> staticNotebooksApiProvider;
   private final FirecloudRetryHandler retryHandler;
   private final Provider<GoogleCredential> fcAdminCredsProvider;
+  private final Provider<CloudStorageService> cloudStorageServiceProvider;
   private final ServiceAccounts serviceAccounts;
+  private final LoadingCache<String, GoogleCredential> garbageCollectionSACredentials;
```
jmthibault79: In Perf we may need to get these credentials hundreds of times per run.  Let's not hit GCS and auth APIs every time.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -93,13 +102,16 @@ public FireCloudServiceImpl(
       Provider<BillingApi> billingApiProvider,
       Provider<GroupsApi> groupsApiProvider,
       Provider<NihApi> nihApiProvider,
-      @Qualifier("workspacesApi") Provider<WorkspacesApi> workspacesApiProvider,
-      @Qualifier("workspaceAclsApi") Provider<WorkspacesApi> workspaceAclsApiProvider,
+      @Qualifier(FireCloudConfig.END_USER_WORKSPACE_API)
```
jmthibault79: Unrelated, just reusing proper consts

api/src/integration/java/org/pmiops/workbench/FireCloudIntegrationTest.java
```diff
@@ -46,7 +46,6 @@
   @Mock private BillingApi billingApi;
   @Mock private WorkspacesApi workspacesApi;
   @Mock private GroupsApi allOfUsGroupsApi;
-  @Mock private GroupsApi endUserGroupsApi;
```
jmthibault79: Unrelated - was not used

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,164 @@
...
+    this.workbenchConfigProvider = workbenchConfigProvider;
+    this.cloudStorageServiceProvider = cloudStorageServiceProvider;
+    this.clock = clock;
+
+    this.garbageCollectionSACredentials = initializeCredentialCache();
+  }
+
+  private LoadingCache<String, GoogleCredential> initializeCredentialCache() {
+    return CacheBuilder.newBuilder()
+        .expireAfterWrite(24, TimeUnit.HOURS)
```
jmthibault79: arbitrary - but it's a daily cron so it makes sense

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,164 @@
...
+              appEngineSA, projectName, garbageCollectionSA);
+      throw new BadRequestException(msg, e);
+    }
+
+    return recordGarbageCollection(projectName, garbageCollectionSA);
+  }
+
+  BillingProjectGarbageCollectionResponse deletedWorkspaceGarbageCollection() {
+    final List<String> billingProjects =
+        billingProjectGarbageCollectionDao.findBillingProjectsForGarbageCollection();
```
jmthibault79: alt: could stream this call directly instead of saving a local var

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectGarbageCollectionDao.java
```diff
@@ -0,0 +1,26 @@
...
+public interface BillingProjectGarbageCollectionDao
+    extends CrudRepository<BillingProjectGarbageCollection, Long> {
+
+  // get Billing Projects which are ASSIGNED to Workspaces which have been DELETED and have NEW
+  // Billing Migration Status.  These are ready to be garbage-collected
+  @Query(
+      "SELECT p.fireCloudProjectName "
+          + "FROM BillingProjectBufferEntry p "
+          + "JOIN Workspace w "
+          + "ON w.workspaceNamespace = p.fireCloudProjectName "
```
ericsong: we shouldn't make this change here but making a note that I think we should probably add workspace as a foreign key to the buffer table so we avoid making joins w/o the foreign key integrity.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -259,6 +266,32 @@ public void removeUserFromBillingProject(String email, String projectName) {
...
+  public void addOwnerToBillingProject(String ownerEmail, String projectName) {
+    addRoleToBillingProject(ownerEmail, projectName, OWNER_FC_ROLE);
+  }
+
+  @Override
+  public void removeOwnerFromBillingProject(
+      String ownerEmailToRemove, GoogleCredential retainingOwnerCredential, String projectName)
+      throws IOException {
+
+    retainingOwnerCredential.refreshToken();
```
ericsong: nit: Maybe pass in the access token instead of the GoogleCredential so we can push all of the credentials logic into the caller versus having it split between the caller and this function.

api/src/main/resources/workbench.yaml
```diff
@@ -1202,6 +1203,27 @@ paths:
...
             $ref: '#/definitions/ErrorResponse'
 
+  /v1/cron/billingProjectGarbageCollection:
+    get:
+      security: []
+      tags:
+        - offlineBilling
+        - cron
+      description: >
+        Endpoint meant to be called offline to trigger a bulk transfer of ownership of billing
```
ericsong: nit: offline usage is implied from the `offlineBilling` tag and cron usage from `cron`

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,164 @@
...
+              public GoogleCredential load(String saEmail) throws IOException {
+                return cloudStorageServiceProvider
+                    .get()
+                    .getGarbageCollectionServiceAccountCredentials(saEmail)
+                    .createScoped(FireCloudServiceImpl.FIRECLOUD_API_OAUTH_SCOPES);
+              }
+            });
+  }
+
+  // Is the App Engine SA a member of this FireCloud project?
```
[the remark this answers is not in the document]
jmthibault79: describing the function.  is there a better way to say this?  (or should I remove the comment?)

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,164 @@
...
+  private Boolean appSAIsMemberOfProject(String billingProject) {
```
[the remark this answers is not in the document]
jmthibault79: yes

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,171 @@
...
+        return sa;
+      }
+    }
+
+    final String msg =
+        String.format(
+            "No available Garbage Collection Service Accounts.  "
+                + "These GCSAs exceed the configured capacity limit of %d: %s",
+            config.garbageCollectionUserCapacity, String.join(", ", config.garbageCollectionUsers));
+    throw new BadRequestException(msg);
```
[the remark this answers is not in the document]
jmthibault79: I'm not super familiar with our error handling / error code response norms.  These exception classes seemed like a good way to provide flexibility: calling code can choose to catch this or not, and bubble up to the HTTP response if not.

500 makes sense.  I'll update the others as well.

api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="changelog-101-billing-project-garbage-collection">
+
+    <createTable tableName="billing_project_garbage_collection">
```
als364: does mysql let you have column names longer than 31 characters?

api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
```diff
@@ -0,0 +1,18 @@
...
+      <column name="fireCloudProjectName" type="varchar(80)">
```
als364: snake_case instead of camelCase please

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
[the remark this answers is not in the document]
als364: 'Checks whether the AppEngine service account is a member of this FireCloud project'

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectGarbageCollectionDao.java
```diff
@@ -0,0 +1,26 @@
...
+          + "AND p.status = 4 " // BillingProjectBufferStatus.ASSIGNED
```
als364: Is it common for us to use magic enum numbers in queries like this?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
```diff
@@ -46,11 +47,26 @@
   void addUserToBillingProject(String email, String projectName);
 
   /**
-   * Removes the specified user from the specified billing project. Only used for errored billing
-   * projects
+   * Removes the specified user from the specified billing project.
+   *
+   * <p>Only used for errored billing projects
```
als364: Why `<p>`?  It these are automagically parsed to print html-y documentation, does the parser close the tags for you?

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -0,0 +1,171 @@
...
+
+    final BillingProjectGarbageCollection gc = new BillingProjectGarbageCollection();
+    gc.setFireCloudProjectName(projectName);
+    gc.setOwner(garbageCollectionSA);
+    billingProjectGarbageCollectionDao.save(gc);
+
+    final BillingProjectBufferEntry entry =
+        billingProjectBufferEntryDao.findByFireCloudProjectName(projectName);
+    entry.setStatusEnum(
+        BillingProjectBufferStatus.GARBAGE_COLLECTED,
```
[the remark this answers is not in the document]
jmthibault79: There's a tradeoff here between state-redundancy and consistency.  These workspaces are in the `ASSIGNED` state at the beginning of the process, which is no longer accurate.  So changing that state would be my preference.

api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
jmthibault79: Looks like 64: https://dev.mysql.com/doc/refman/8.0/en/identifier-length.html

api/db/changelog/db.changelog-101-billing-project-garbage-collection.xml
jmthibault79: Good catch, thanks.  Oddly enough, this still works.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
jmthibault79: This was the linter's doing, and the linter will actually remove a closing `</p>`.  I can remove this if I don't insist on making this a separate line.

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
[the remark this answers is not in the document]
ericsong: Fair 👍

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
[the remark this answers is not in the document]
ericsong: I'm not sure if we've reached a team wide consensus on this but my preference is to leave all HTTP related code in the controllers in the interest of layered architecture. Writing our business logic service classes to be platform agnostic makes the code more flexible. Ex. what if we want to invoke this function from a background task and not as a HTTP response? 

In this design, the controllers would receive a more generic error response like an IllegalStateException and then decide what kind of response it wants to return.

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectGarbageCollectionDao.java
jmthibault79: Dunno - but yeah I should do this without magic

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectGarbageCollectionDao.java
jmthibault79: Moved this query to `BillingProjectBufferEntryDao` btw

PR #2610, opened 2019-09-26
[risk=no][no ticket] A few small changes I made in the course of doing actual work

jmthibault79:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
@@ -129,29 +129,29 @@ public class CloudStorageServiceImpl implements CloudStorageService {
     return readToString(getCredentialsBucketName(), objectPath);
   }
 
-  private GoogleCredential getBucketCredential(final String objectPath) throws IOException {
-    String json = readCredentialsBucketString(objectPath);
+  private GoogleCredential getCredential(final String objectPath) throws IOException {
+    final String json = readCredentialsBucketString(objectPath);
     return GoogleCredential.fromStream(new ByteArrayInputStream(json.getBytes()));
   }
 
   @Override
   public GoogleCredential getGSuiteAdminCredentials() throws IOException {
-    return getBucketCredential("gsuite-admin-sa.json");
+    return getCredential("gsuite-admin-sa.json");
   }
 
   @Override
   public GoogleCredential getFireCloudAdminCredentials() throws IOException {
-    return getBucketCredential("firecloud-admin-sa.json");
+    return getCredential("firecloud-admin-sa.json");
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -727,6 +727,62 @@ paths:
           schema:
             $ref: '#/definitions/EmptyResponse'
 
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/share:
```
jmthibault79: These were grouped with the Cron endpoints for some reason

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -112,38 +111,47 @@ public void writeFile(String bucketName, String fileName, byte[] bytes) {
...
+  private GoogleCredential getBucketCredential(final String objectPath) throws IOException {
+    String json = readCredentialsBucketString(objectPath);
+    return GoogleCredential.fromStream(new ByteArrayInputStream(json.getBytes()));
   }
 
   @Override
   public GoogleCredential getGSuiteAdminCredentials() throws IOException {
-    String json = readToString(getCredentialsBucketName(), "gsuite-admin-sa.json");
-    return GoogleCredential.fromStream(new ByteArrayInputStream(json.getBytes()));
+    return getBucketCredential("gsuite-admin-sa.json");
```
jaycarlton: This is a big improvement.

The only other thing I'd do is to put these constant JSON file names into a .properties file that you could swap out per environment if needed. This will make the Java code more easily portable.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -112,38 +111,47 @@ public void writeFile(String bucketName, String fileName, byte[] bytes) {
...
   public JSONObject getElasticCredentials() {
-    return new JSONObject(readToString(getCredentialsBucketName(), "elastic-cloud.json"));
+    return getCredentialsBucketJSON("elastic-cloud.json");
+  }
+
+  private String readCredentialsBucketString(String objectPath) {
+    return readToString(getCredentialsBucketName(), objectPath);
+  }
+
+  private GoogleCredential getBucketCredential(final String objectPath) throws IOException {
```
jaycarlton: I'd just call this `getCredential`. I typically don't put `final` in argument lists. I just hunt down anyone who reassigns a passed-in argument and hit them with a rolled-up newspaper.

api/src/main/resources/workbench.yaml
jaycarlton: I believe @als364 has a tool to rearrange the swagger file. Alphabetizing the paths and definitions would be awesome.

als364: I don't have a tool per se I just want to alphabetize it. Maybe next Friday in between all the meetings.

PR #2611, opened 2019-09-26
[RW-3474][risk=low] Enable VPC-SC in perf

calbach:


calbach:
Manually applied this to the perf env `./project.rb update-cloud-config --project all-of-us-rw-perf`, verified by creating a new workspace.

PR #2612, opened 2019-09-26
[RW-3560][risk=no] minIdleInstances over minInstances

calbach:
Critical distinction:

- minInstances: all versions will maintain this instance count, regardless of whether they are actively serving traffic
- minIdleInstances: only the traffic-serving version will maintain this instance count

https://cloud.google.com/appengine/docs/standard/python/config/appref

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -27,7 +27,7 @@ DRY_RUN_CMD = %W{echo [DRY_RUN]}
 
 def make_gae_vars(min, max)
   {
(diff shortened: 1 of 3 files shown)


calbach:
Tested this by manually deploying a non-traffic-serving "calbach" version to the test env.

api/libproject/devstart.rb
```diff
@@ -27,7 +27,7 @@
 
 def make_gae_vars(min, max)
```
jaycarlton: nit: rename `min` argument to `min_idle`.

PR #2613, opened 2019-09-26
[risk=low][RW-3452] Inactivity timeout should work across tabs

ericsong:
The previous iteration has a bug where the user will be signed out if any of their tabs has not seen activity in the specified timeout window.

The solution was pretty simple by relying on the Web Message API. Every write to local storage will trigger a message event to all other tabs that can potentially read it (same domain).

All that had to be done was listening to that message and executing the existing logic of resetting the timers when activity within the current tab is detected.

I also added a line to hide the warning modal when activity is detected so that the modals across all tabs close when activity is detected. A consequence of this is that the warning modal automatically closes when activity is detected. The previous behavior was that the user had to click "Extend Session" to close the modal.

I discussed with Karthik and we decided this might actually be a useful feature so we're going to keep it in for now to get some user feedback.


gjuggler:
> Looks good and it works well when I tested locally.
> 
> I don't like having an "Extend Session" button which you can't reach because moving your mouse to click it makes the modal go away. But dismissing the modal on activity is pretty nice. We can iterate on what that looks like more specifically.

It would be reasonable to remove "mousemove" from the activity tracker. Tracking mousedown / clicks is probably sufficient for a reasonable definition of "activity", and it would make the extend session button feel more useful.

ericsong:
I'll ping Karthik on those points. Either
1. Remove the "Extend Session" button
2. Remove "mousemove" as a tracked event.

ericsong:
We've decided to remove "mousemove" as an event.

api/cluster-resources/activity-checker-extension.js
```diff
@@ -32,7 +32,7 @@ define([
     }, 1000);
 
     // Events array should be in sync with INACTIVITY_CONFIG.TRACKED_EVENTS from signed-in/component.ts
-    ['mousemove', 'mousedown', 'keypress', 'scroll', 'click'].forEach(eventName => {
+    ['mousedown', 'keypress', 'scroll', 'click'].forEach(eventName => {
```
jaycarlton: I see this is specified in two places. Can they not pull the events array from the same place?

PR #2622, opened 2019-09-30
[risk=no] Fix returning user creation

calbach:
This codepath gets hit when a user visits the site and gets lazily created, rather than getting created by the account creation page. This happens primarily during local development.

Aside from this, the NPE is clearly visible within the UserService, since another method passes `userAffiliations = null`.


calbach:
Ignoring BQ integration test flake - this doesn't interact with BigQuery at all.

PR #2626, opened 2019-10-01
[risk=low][RW-3435]Follow up: record ownership transfer by logging instead of return value

jmthibault79:
We can't retrieve the results of cron endpoints from Cloud Console, but we can see logs.

Since our cron endpoints don't return anything, they should be 204 No Content instead of 200 OK

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
@@ -14,7 +14,6 @@ import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.google.CloudResourceManagerService;
 import org.pmiops.workbench.model.DataAccessLevel;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
@@ -105,7 +104,7 @@ public class OfflineUserController implements OfflineUserApiDelegate {
           String.format("%d errors encountered during compliance training sync", errorCount));
     }
 
-    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+    return ResponseEntity.noContent().build();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -1080,7 +1080,7 @@ paths:
       description: sync moodle badge/training status for all users.
       operationId: bulkSyncComplianceTrainingStatus
       responses:
-        200:
+        204:
```
jmthibault79: These first 4 were already returning 204s, in addition to the 5 I changed

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -105,7 +104,7 @@ private boolean timestampsEqual(Timestamp a, Timestamp b) {
           String.format("%d errors encountered during compliance training sync", errorCount));
     }
 
-    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+    return ResponseEntity.noContent().build();
```
jaycarlton: 👏

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -114,10 +112,13 @@ private GarbageCollectedProject recordGarbageCollection(
         () -> new Timestamp(clock.instant().toEpochMilli()));
     billingProjectBufferEntryDao.save(entry);
 
-    return gc.toGarbageCollectedProject();
+    log.info(
```
jaycarlton: Is there any thought to a queryable table for this information, like `project_gcp_id, owner, lifecycle_stage?

Also, if you want these values to be a separate (searchable) field, you can use the Logging client with a jsonPayload.

api/src/main/java/org/pmiops/workbench/billing/OfflineBillingController.java
```diff
@@ -25,33 +23,32 @@
   }
 
   @Override
-  public ResponseEntity<BillingProjectGarbageCollectionResponse> billingProjectGarbageCollection() {
-    BillingProjectGarbageCollectionResponse response =
-        billingGarbageCollectionService.deletedWorkspaceGarbageCollection();
-    return ResponseEntity.ok(response);
+  public ResponseEntity<Void> billingProjectGarbageCollection() {
+    billingGarbageCollectionService.deletedWorkspaceGarbageCollection();
+    return ResponseEntity.noContent().build();
```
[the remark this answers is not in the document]
jaycarlton: Is there any possibility that could break something? I.e. do we have code looking for 200 exactly? 

I'm in favor of this change, but it's possible we've got tendrils growing around the old status codes.

[UPDATE] from chatting offline it sounds like we're good.

PR #2628, opened 2019-10-01
Switch flag for account creation on test

NehaBroad:


ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -131,9 +131,9 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
         },
         institutionalAffiliations: [
           {
-            institution: undefined,
+            institution: '',
             nonAcademicAffiliation: undefined,
```
gjuggler: All other fields in this object are either strings or boolean values... is there a reason why we believe 'undefined' is safe for the nonAcademicAffiliation field? If so, we should add a comment explaining why it's different from all the other fields.

ui/src/environments/environment.test.ts
```diff
@@ -6,5 +6,5 @@ export const environment: Environment = {
   displayTag: 'Test',
   debug: false,
   enablePublishedWorkspaces: true,
-  enableAccountPages: false
+  enableAccountPages: true
```
gjuggler: For simplicity in cherry-picking a change for this release, I'd prefer if the environment change were separated from the code fix.

ui/src/app/pages/login/account-creation/account-creation.tsx
NehaBroad: nonAcademicAffiliation is ENUM field while institution is string since it is an existing field i can add comment

PR #2629, opened 2019-10-01
[RW-3601] [risk=low] Fix database errors when creating user accounts

NehaBroad:
with the feature flag set to NO users were not able to create account because of institution is required error. This should fix that

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
@@ -6,7 +6,6 @@ import java.sql.Timestamp;
 import java.time.Clock;
 import java.time.Instant;
 import java.util.List;
-import java.util.Objects;
 import java.util.Random;
 import java.util.function.Function;
 import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -268,11 +269,20 @@ public User createUser(
     if (demographicSurvey != null) demographicSurvey.setUser(user);
     if (institutionalAffiliations != null) {
       final User u = user;
-      institutionalAffiliations.forEach(
-          affiliation -> {
-            affiliation.setUser(u);
-            u.addInstitutionalAffiliation(affiliation);
-          });
+      institutionalAffiliations.stream()
+          .filter(Objects::nonNull)
```
calbach: In my view, the bug is that currently the client code is sending:

```
/createAccount {
  institutionalAffiliations: [{
    institution: undefined,
    role: undefined
  }]
}
```

client side I would either filter that out when calling `createAccount`, or maybe just always filter it out if the flag is off.

gjuggler:
> I'm a little confused here. Is this PR meant to cover the fix that had previously been part of #2628? The code looks completely different... did you decide on an alternative approach?
> 
> If we can get clarity & confidence with this fix by later this morning, I'll be tentatively OK pushing forward with a fix (although it's almost never best practice to roll forward). If the breakage were simply due to a code change, I would immediately rollback, but since it involves a Liquibase db migration it may be tricky to safely rollback (since we haven't yet implemented CH's database rollback strategy).

We chatted offline – this is an improved version of the prior fix. I'll try patching this in & confirming things work okay.

PR #2630, opened 2019-10-02
[RW-3458][risk=no] Sort based on DataSet "sub-value" rather than value

jmthibault79:
Sneaky bug here: the "values" member of DataSet does not actually contain values we want to compare against.  Instead it refers to `DomainValuePair`s which must then be dereferenced to get their actual values.

This is confusing!  I'm going to fix that situation, but in PR #2631 since that is a much larger change.  This PR simply fixes the bug.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -362,9 +362,14 @@ public class DataSetController implements DataSetApiDelegate {
                 "Timeout while querying the CDR to pull preview information.");
           }
 
+          final List<String> requestValues =
+              dataSet.getValues().stream()
+                  .map(DomainValuePair::getValue)
+                  .collect(Collectors.toList());
+
           Collections.sort(
               valuePreviewList,
-              Comparator.comparing(item -> dataSet.getValues().indexOf(item.getValue())));
+              Comparator.comparing(item -> requestValues.indexOf(item.getValue())));
 
           previewQueryResponse.addDomainValueItem(
               new DataSetPreviewList().domain(domain).values(valuePreviewList));
@@ -581,10 +586,7 @@ public class DataSetController implements DataSetApiDelegate {
     }
     DataSetListResponse dataSetResponse =
         new DataSetListResponse()
-            .items(
-                dbDataSets.stream()
-                    .map(dbDataSet -> TO_CLIENT_DATA_SET.apply(dbDataSet))
-                    .collect(Collectors.toList()));
+            .items(dbDataSets.stream().map(TO_CLIENT_DATA_SET).collect(Collectors.toList()));
     return ResponseEntity.ok(dataSetResponse);
   }
 


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -581,10 +586,7 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
     }
     DataSetListResponse dataSetResponse =
         new DataSetListResponse()
-            .items(
-                dbDataSets.stream()
-                    .map(dbDataSet -> TO_CLIENT_DATA_SET.apply(dbDataSet))
-                    .collect(Collectors.toList()));
+            .items(dbDataSets.stream().map(TO_CLIENT_DATA_SET).collect(Collectors.toList()));
```
jmthibault79: couldn't help myself

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -362,9 +362,14 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
...
 
+          final List<String> requestValues =
+              dataSet.getValues().stream()
+                  .map(DomainValuePair::getValue)
+                  .collect(Collectors.toList());
+
           Collections.sort(
               valuePreviewList,
-              Comparator.comparing(item -> dataSet.getValues().indexOf(item.getValue())));
+              Comparator.comparing(item -> requestValues.indexOf(item.getValue())));
```
jmthibault79: No `DomainValuePair` matched the `String` so `indexOf()` was always -1 and therefore no sorting happened here.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: silence of the lambdas

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: so this is something I was wondering aloud the other day: should the queries have a canonical form so they're mostly repeatable, or do we hold the position that you could get a different, equivalent query the next time you use the builder. Might be amusing for workshops if everybody's query was sorted differently.

jaycarlton: Let's chat about refactoring the nomenclature here. I hate using value as a name for anything, especially a column. If nothing else, we can keep it in the UI but make the code base easier to unersdand.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: So I'd argue that this method knows too much. I want controller methods paper thin, with no business logic if possible. I.e. I'd push this all into a service method that returns your list of dataSets to just be packaged into the response.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -362,9 +362,14 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
                 "Timeout while querying the CDR to pull preview information.");
           }
 
+          final List<String> requestValues =
```
jaycarlton: A unit test for this logic would be nice.

PR #2631, opened 2019-10-02
[risk=no][no ticket]Rename confusing "values" in DataSet and DataSetRequest

jmthibault79:


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -362,9 +356,14 @@ public DomainValuePair apply(DataSetValues dataSetValue) {
                 "Timeout while querying the CDR to pull preview information.");
           }
 
+          final List<String> requestValues =
```
jmthibault79: This is #2630 which will get merged first so this diff will go away

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -111,13 +111,13 @@ class ExportDataSetModal extends React.Component<
   async exportDataSet() {
     this.setState({loading: true});
     const {dataSet, workspaceNamespace, workspaceFirecloudName} = this.props;
-    const request = {
+    const request: DataSetRequest = {
```
jmthibault79: Always set the type, where possible

jmthibault79:
Note to self: rebase after #2630 and #2573 have merged since this will conflict with them

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -132,23 +132,17 @@
...
-    } else if (dataSetRequest.getValues() == null || dataSetRequest.getValues().size() == 0) {
+    } else if (dataSetRequest.getDomainValuePairs() == null
+        || dataSetRequest.getDomainValuePairs().isEmpty()) {
       throw new BadRequestException("Missing values");
     }
     Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
     long wId = workspaceService.get(workspaceNamespace, workspaceId).getWorkspaceId();
     List<DataSetValues> dataSetValuesList =
```
jmthibault79: NOT updating `DataSetValues`.  That is a whole other can of worms.  Aiming for minimal in this PR.

api/src/main/java/org/pmiops/workbench/db/model/DataSetValues.java
```diff
@@ -34,10 +40,6 @@ public Domain getDomainEnum() {
     return CommonStorageEnums.domainFromStorage(Short.parseShort(domainId));
   }
 
-  public void setDomainEnum(Domain domain) {
```
jmthibault79: replaced by new constructor

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -247,7 +248,7 @@ interface State {
   selectedCohortIds: number[];
   selectedConceptSetIds: number[];
   selectedPreviewDomain: string;
-  selectedValues: DomainValuePair[];
+  selectedValuePairs: DomainValuePair[];
```
ericsong: curious why you went with just `ValuePairs` in the front end and `DomainValuePairs` in the API.

jmthibault79: dunno - so the variable names wouldn't be overly long?  That's probably not a great reason.  I'll update to `selectedDomainValuePairs`

ericsong: I think thats a fair reason if you want to keep it as is.

ericsong: I don't think so BUT I am also a strong proponent for our naming being consistent between our code and our UI. Our UI currently calls these "Values" so I think our code should reflect that.

I do think there is a better name out there but I'd like product/design input and buy in.

jmthibault79: Yeah, my thought was that was non-ideal but unsafe to fix quickly

PR #2632, opened 2019-10-02
[risk=no][no ticket] Increase billing buffer capacity to 500 in perf env

aweng98:
For nightly perf test runs in Perf env.


jaycarlton:
Awesome. It'd be great to get some of these knobs exposed for live tuning, but that takes some doing.

PR #2634, opened 2019-10-02
[no ticket] [risk=low] fix confusingly named frontend config files

als364:
rename `environment.ts` and comment other environment files

### DIFF
--- a/ui/angular.json
+++ b/ui/angular.json
@@ -42,7 +42,7 @@
             "local": {
               "fileReplacements": [
                 {
-                  "replace": "src/environments/environment.ts",
+                  "replace": "src/environments/environment.localtest.ts",
                   "with": "src/environments/environment.local.ts"
                 }
               ]
@@ -50,7 +50,7 @@
             "test": {
               "fileReplacements": [
                 {
-                  "replace": "src/environments/environment.ts",
+                  "replace": "src/environments/environment.localtest.ts",
                   "with": "src/environments/environment.test.ts"
                 }
               ]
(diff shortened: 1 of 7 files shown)


ui/angular.json
```diff
@@ -42,23 +42,23 @@
             "local": {
               "fileReplacements": [
                 {
-                  "replace": "src/environments/environment.ts",
+                  "replace": "src/environments/environment.localtest.ts",
```
calbach: This should probably not be changed, you probably need to keep the environment.ts file (perhaps has an empty stub) and in the `dev` block above, replace it with `localtest`

als364:
This is odd - this is failing because the 'new' Account Creation page is getting rendered. Which means the test is reading a truthy `enableAccountPages` from somewhere. I think what's ending up getting pulled in is `environment.ts`. I have noooo idea where this is getting pulled in to jest, especially since I can't actually access `environment` as a variable in tests.

There are commented out tests in `account-creation.spec.tsx` intended for the new Account Creation page. There's two tests there that are failing for the above reason. The rest are passing. The commented out tests pass when uncommented (with `enableAccountPages` truthy and with some typo fixes). I could do any of the following:
- leave those tests commented out, put `enableAccountPages: false` in `environment.ts`
- uncomment those tests, delete the 'old' tests
- uncomment those tests, leave the old tests, and condition both on the value of `enableAccountPages` in `environment.ts`, which I can import explicitly

@calbach What do you think about those options?

calbach:
Also the Snyk error is unrelated to this PR. You can feel free to ignore it - or if you feel more comfortable having it go green, I suspect if you rebase your PR against master and push it up again, that would do the trick.

als364:
I'll rebase.

Sorry about the title, I was trying to get _something_ up before I had a meeting so I wouldn't completely lose continuity

PR #2636, opened 2019-10-03
[risk=no][no ticket]Flaky test fix suggestion

jmthibault79:


api/src/test/java/org/pmiops/workbench/db/dao/CohortAnnotationDefinitionDaoTest.java
```diff
@@ -22,6 +24,7 @@
 @Import({LiquibaseAutoConfiguration.class})
 @AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
 @Transactional
+@DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD)
```
jaycarlton: aside: If this fixes it, can we push it into a parent class, like `WorkbenchDaoTest`?

freemabd: Seems fine if everyone understands that any doa jpa test needs to extend this parent class. I would chose another name for the parent class since we have a WorkspaceDaoTest.

PR #2639, opened 2019-10-03
[RW-3623][RISK=no] domains must be unique and we need actual title case

jaycarlton:
The map error was caused by passing in all domains from the domainvaluepairs, not just the unique ones. Aside: could we just use a map for domainValuePairs.

Another issue was the library method StringUtils.capitalize() didn't convert already uppercase letters to lowercase.

This lets me get through the dataset request stuff, but I'm seeing permissions errors locally 

```
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
{
  "code" : 403,
  "errors" : [ {
    "domain" : "global",
    "message" : "all-of-us-workbench-test@appspot.gserviceaccount.com does not have storage.objects.create access to fc-secure-3f9af1df-24a5-44b2-82de-4c72e800a06d/notebooks/bb.ipynb.",
    "reason" : "forbidden"
  } ],
  "message" : "all-of-us-workbench-test@appspot.gserviceaccount.com does not have storage.objects.create access to fc-secure-3f9af1df-24a5-44b2-82de-4c72e800a06d/notebooks/bb.ipynb."
}
```

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -590,6 +590,9 @@ public class DataSetServiceImpl implements DataSetService {
     return new ValuesLinkingPair(valueSelects, valueJoins);
   }
 
+  // Capitalizes the first letter of a string and lowers the remaining ones.
+  // Assumes a single word, so you'd get "A tale of two cities" instead of
+  // "A Tale Of Two Cities"
   private static String toTitleCase(String name) {
     if (name.isEmpty()) {
       return name;
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -594,7 +594,7 @@ public class DataSetServiceImpl implements DataSetService {
     if (name.isEmpty()) {
       return name;
     } else if (name.length() == 1) {
-      return name;
+      return name.toUpperCase();
     } else {
       return String.format("%s%s", name.charAt(0), name.substring(1).toLowerCase());
     }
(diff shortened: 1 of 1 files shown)


jaycarlton:
I know these are both necessary, but I haven't confirmed they're sufficient.

jaycarlton:
Confirmed Preview is working but we fail with the access error on export. @blrubenstein says this may indicate an SA quota problem.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -591,6 +590,16 @@ private ValuesLinkingPair getValueSelectsAndJoins(List<DomainValuePair> domainVa
     return new ValuesLinkingPair(valueSelects, valueJoins);
   }
 
+  private static String toTitleCase(String name) {
+    if (name.isEmpty()) {
+      return name;
+    } else if (name.length() == 1) {
+      return name;
+    } else {
+      return String.format("%s%s", name.charAt(0), name.substring(1).toLowerCase());
```
blrubenstein: Nit: I would explicitly set the first character to uppercase, just for documentation reasons/input safety

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -591,6 +590,16 @@ private ValuesLinkingPair getValueSelectsAndJoins(List<DomainValuePair> domainVa
     return new ValuesLinkingPair(valueSelects, valueJoins);
   }
 
+  private static String toTitleCase(String name) {
+    if (name.isEmpty()) {
+      return name;
+    } else if (name.length() == 1) {
+      return name;
```
calbach: Why would it be correct to leave a single letter uncapitalized here?

calbach:
RE notebook error: oversight on my part. Looking into it

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: Good catch. We don't have any single-letter names at the moment.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: good point.

PR #2640, opened 2019-10-04
unit tests for string util function

jaycarlton:
This tiny helper function, when I tried to replace it with a library function, broke the whole data set service. So it's got a more descriptive name and unit tests.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -593,7 +593,8 @@ public class DataSetServiceImpl implements DataSetService {
   // Capitalizes the first letter of a string and lowers the remaining ones.
   // Assumes a single word, so you'd get "A tale of two cities" instead of
   // "A Tale Of Two Cities"
-  private static String toTitleCase(String name) {
+  @VisibleForTesting
+  public static String toTitleCase(String name) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -559,7 +559,7 @@ private ValuesLinkingPair getValueSelectsAndJoins(List<DomainValuePair> domainVa
             .collect(Collectors.toList()));
 
     final String domainName = domainMaybe.get().toString();
-    final String domainTitleCase = toTitleCase(domainName);
+    final String domainTitleCase = capitalizeFirstCharacterOnly(domainName);
```
jmthibault79: also rename `domainTitleCase`

jaycarlton: Since the domain is always a single word, technically it's still title case. The reason for the rename was if this method were moved or exposed more generally.

Still, `domainFirstCharacterCapitalized` is fine too.

jmthibault79: OK then either is fine

jaycarlton: done

PR #2643, opened 2019-10-04
[risk=no] Add API interface mapping for DatasetController

ericsong:


jmthibault79:
What if we did something like `WorkspacesController.getClass().toString()` so that we would catch references when we refactored?

api/src/main/java/org/pmiops/workbench/interceptors/InterceptorUtils.java
```diff
@@ -12,8 +12,10 @@
       ImmutableMap.of(
           "org.pmiops.workbench.api.WorkspacesApiController",
               "org.pmiops.workbench.workspaces.WorkspacesController",
-          "org.pmiops.workbench.api.BillingApiController",
-              "org.pmiops.workbench.billing.BillingController");
+          "org.pmiops.workbench.api.OfflineBillingApiController",
```
calbach: I don't know if pushing controllers down into subpackages is worth the effort of maintaining this obscure map. Alternatively, we should probably fix this entire class as it is omega-hacky - there's probably a way to get the bound instance class via Spring boot instead to eliminate the sketchy naming-convention-based reflection lookup below.

Obviously not for this PR, but I think we should track this in a ticket if we haven't already.

ericsong:
great suggestion, adding that now

jmthibault79:
Closing this old PR in favor of #3710

PR #2646, opened 2019-10-04
[draft] cron job logging for checkFreeTierBillingUsage

jaycarlton:
Added some logging around error conditions and general tidbits.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
@@ -1,21 +1,34 @@
 package org.pmiops.workbench.billing;
 
+import com.google.cloud.bigquery.FieldValue;
+import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.TableResult;
 import com.google.common.collect.Streams;
 import java.util.AbstractMap;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
 import org.pmiops.workbench.api.BigQueryService;
 import org.pmiops.workbench.config.WorkbenchConfig;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -43,26 +55,69 @@ public void checkFreeTierBillingUsage() {
...
+    if (queryResults.getTotalRows() < numWorkspaces) {
+      logger.warn("Expected %d rows but found %d", numWorkspaces, queryResults.getTotalRows());
+    }
+
+    Streams.stream(queryResults.getValues())
+        .filter(fv -> fv.get("id").isNull())
+        .forEach(
+            fv -> logger.error(String.format("Query result has no project ID: %s", fv.toString())));
 
     Map<User, Double> perUserSpend =
```
calbach: This entire stream should probably be written differently per my above comment - perhaps starting with the namespaceToCreator map and then joining against the query results.

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -43,26 +55,69 @@ public void checkFreeTierBillingUsage() {
...
     Map<String, User> namespaceToCreator = workspaceDao.namespaceToCreator();
+    final int numWorkspaces = namespaceToCreator.keySet().size();
+    final long numUniqueUsers = namespaceToCreator.values().stream().distinct().count();
+    logger.info(
+        String.format("found %d workspaces with %d unique users", numWorkspaces, numUniqueUsers));
+
+    final TableResult queryResults = bigQueryService.executeQuery(queryConfig);
+    logger.info("Query retrieved %d results", queryResults.getTotalRows());
+
+    if (queryResults.getTotalRows() < numWorkspaces) {
```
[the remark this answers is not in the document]
jaycarlton: OK, I'll try to document that.

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
jaycarlton: Is it worth me doing that on master, or would you rather pick what's useful off this branch?

calbach: Thanks - I can just pick things off this branch as needed

PR #2649, opened 2019-10-07
[draft] bean question

jaycarlton:


api/src/main/java/org/pmiops/workbench/audit/ActionAuditService.java
```diff
@@ -0,0 +1,15 @@
+package org.pmiops.workbench.audit;
+
+import java.util.Collection;
+import org.elasticsearch.common.UUIDs;
+import org.springframework.context.annotation.Bean;
+import org.springframework.stereotype.Service;
+
+@Service
```
jaycarlton: this is the thing I want to inject

PR #2650, opened 2019-10-07
[RW-3528][Risk=no] Retry Preview Table from Front End only

blrubenstein:
This is to avoid the AppEngine hard timeout limit, giving us the full time for each retry. 

Open question for reviewers: should I add a 'requested rows' parameter to the request that shrinks with each retry? This would be an optional parameter with some reasonable default.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -77,7 +77,6 @@ public class DataSetController implements DataSetApiDelegate {
   private final WorkspaceService workspaceService;
 
   private static int NO_OF_PREVIEW_ROWS = 20;
-  private static int PREVIEW_RETRY_LIMIT = 2;
   // See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
   private static long APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC = 55000l;
   private static String CONCEPT_SET = "conceptSet";
@@ -300,101 +299,88 @@ public class DataSetController implements DataSetApiDelegate {
     QueryJobConfiguration queryJobConfiguration =
         bigQueryJobConfig.get(dataSetPreviewRequest.getDomain().toString());
 
-    int retry = 0, rowsRequested = NO_OF_PREVIEW_ROWS;
     String originalQuery = queryJobConfiguration.getQuery();
-    do {
-      try {
-        String query = originalQuery.concat(" LIMIT " + rowsRequested);
+    TableResult queryResponse;
+    try {
+      String query = originalQuery.concat(" LIMIT " + NO_OF_PREVIEW_ROWS);
 
-        queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
+      queryJobConfiguration = queryJobConfiguration.toBuilder().setQuery(query).build();
 
-        /* Google appengine has a 60 second timeout, we want to make sure this endpoint completes
-         * before that limit is exceeded, or we get a 500 error with the following type:
-         * com.google.apphosting.runtime.HardDeadlineExceededError
-         * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
-         */
-        TableResult queryResponse =
-            bigQueryService.executeQuery(
-                bigQueryService.filterBigQueryConfig(queryJobConfiguration),
-                APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC / PREVIEW_RETRY_LIMIT + 1);
-        queryResponse
-            .getSchema()
-            .getFields()
-            .forEach(
-                fields -> {
-                  valuePreviewList.add(new DataSetPreviewValueList().value(fields.getName()));
-                });
-
-        queryResponse
-            .getValues()
-            .forEach(
-                fieldValueList -> {
-                  IntStream.range(0, fieldValueList.size())
-                      .forEach(
-                          columnNumber -> {
-                            try {
-                              valuePreviewList
-                                  .get(columnNumber)
-                                  .addQueryValueItem(
-                                      fieldValueList.get(columnNumber).getValue().toString());
-                            } catch (NullPointerException ex) {
-                              log.severe(
-                                  String.format(
-                                      "Null pointer exception while retriving value for query: Column %s ",
-                                      columnNumber));
-                              valuePreviewList.get(columnNumber).addQueryValueItem("");
-                            }
-                          });
-                });
-        queryResponse
-            .getSchema()
-            .getFields()
-            .forEach(
-                fields -> {
-                  DataSetPreviewValueList previewValue =
-                      valuePreviewList.stream()
-                          .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
-                          .findFirst()
-                          .get();
-                  if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
-                    List<String> queryValues = new ArrayList<String>();
-                    DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
-                    previewValue
-                        .getQueryValue()
-                        .forEach(
-                            value -> {
-                              try {
-                                Double fieldValue = Double.parseDouble(value);
-                                queryValues.add(
-                                    dateFormat.format(new Date(fieldValue.longValue())));
-                              } catch (NumberFormatException ex) {
-                                queryValues.add("");
-                              }
-                            });
-                    previewValue.setQueryValue(queryValues);
-                  }
-                });
-        break;
-      } catch (Exception ex) {
-        if ((ex.getCause() != null
-            && ex.getCause().getMessage() != null
-            && ex.getCause().getMessage().contains("Read timed out"))) {
-          rowsRequested = (rowsRequested / 2);
-          if (rowsRequested == 0) {
-            throw new GatewayTimeoutException(
-                "Timeout while querying the CDR to pull preview information.");
-          }
-          retry++;
-        } else {
-          throw ex;
-        }
+      /* Google appengine has a 60 second timeout, we want to make sure this endpoint completes
+       * before that limit is exceeded, or we get a 500 error with the following type:
+       * com.google.apphosting.runtime.HardDeadlineExceededError
+       * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
+       */
+      queryResponse =
+          bigQueryService.executeQuery(
+              bigQueryService.filterBigQueryConfig(queryJobConfiguration),
+              APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC);
+    } catch (Exception ex) {
+      if ((ex.getCause() != null
+          && ex.getCause().getMessage() != null
+          && ex.getCause().getMessage().contains("Read timed out"))) {
+        throw new GatewayTimeoutException(
+            "Timeout while querying the CDR to pull preview information.");
+      } else {
+        throw ex;
       }
-    } while (retry < PREVIEW_RETRY_LIMIT);
-
-    if (retry == PREVIEW_RETRY_LIMIT) {
-      throw new GatewayTimeoutException(
-          "Timeout while querying the CDR to pull preview information.");
     }
+    queryResponse
+        .getSchema()
+        .getFields()
+        .forEach(
+            fields -> {
+              valuePreviewList.add(new DataSetPreviewValueList().value(fields.getName()));
+            });
+
+    queryResponse
+        .getValues()
+        .forEach(
+            fieldValueList -> {
+              IntStream.range(0, fieldValueList.size())
+                  .forEach(
+                      columnNumber -> {
+                        try {
+                          valuePreviewList
+                              .get(columnNumber)
+                              .addQueryValueItem(
+                                  fieldValueList.get(columnNumber).getValue().toString());
+                        } catch (NullPointerException ex) {
+                          log.severe(
+                              String.format(
+                                  "Null pointer exception while retriving value for query: Column %s ",
+                                  columnNumber));
+                          valuePreviewList.get(columnNumber).addQueryValueItem("");
+                        }
+                      });
+            });
+    queryResponse
+        .getSchema()
+        .getFields()
+        .forEach(
+            fields -> {
+              DataSetPreviewValueList previewValue =
+                  valuePreviewList.stream()
+                      .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+                      .findFirst()
+                      .get();
+              if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
+                List<String> queryValues = new ArrayList<String>();
+                DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+                previewValue
+                    .getQueryValue()
+                    .forEach(
+                        value -> {
+                          try {
+                            Double fieldValue = Double.parseDouble(value);
+                            queryValues.add(dateFormat.format(new Date(fieldValue.longValue())));
+                          } catch (NumberFormatException ex) {
+                            queryValues.add("");
+                          }
+                        });
+                previewValue.setQueryValue(queryValues);
+              }
+            });
 
     Collections.sort(
         valuePreviewList,
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -44,7 +44,6 @@ import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.exceptions.ConflictException;
 import org.pmiops.workbench.exceptions.GatewayTimeoutException;
 import org.pmiops.workbench.exceptions.NotFoundException;
-import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.firecloud.model.WorkspaceResponse;
 import org.pmiops.workbench.model.ConceptSet;
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -340,18 +340,12 @@ public class DataSetController implements DataSetApiDelegate {
               IntStream.range(0, fieldValueList.size())
                   .forEach(
                       columnNumber -> {
-                        try {
-                          valuePreviewList
-                              .get(columnNumber)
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+       * See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
+       */
+      queryResponse =
+          bigQueryService.executeQuery(
+              bigQueryService.filterBigQueryConfig(queryJobConfiguration),
+              APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC);
+    } catch (Exception ex) {
+      if ((ex.getCause() != null
+          && ex.getCause().getMessage() != null
+          && ex.getCause().getMessage().contains("Read timed out"))) {
```
jaycarlton: Does BigQuery not document a more specific exception in this situation? Looks like we can catch [BigQueryException](https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/bigquery/BigQueryException.html#getError--) and then call `getError` on that to get a `BigQueryError`, which exposes [getReason](https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/bigquery/BigQueryError.html#getReason--), which is a "short error code", which might be more stable than using the message.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
-    } while (retry < PREVIEW_RETRY_LIMIT);
-
-    if (retry == PREVIEW_RETRY_LIMIT) {
-      throw new GatewayTimeoutException(
-          "Timeout while querying the CDR to pull preview information.");
     }
+    queryResponse
+        .getSchema()
+        .getFields()
+        .forEach(
```
jaycarlton: nit: Instead of foreach, you can map/collect into a list, and then do an addAll onto `valuePreviewList`. The main difference for me is that readers generally expect `forEach` loops to have side effects, and thus be less functional than the stream style.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+              IntStream.range(0, fieldValueList.size())
+                  .forEach(
+                      columnNumber -> {
+                        try {
+                          valuePreviewList
+                              .get(columnNumber)
+                              .addQueryValueItem(
+                                  fieldValueList.get(columnNumber).getValue().toString());
+                        } catch (NullPointerException ex) {
+                          log.severe(
```
jaycarlton: So in this situation, why do we keep going? It seems like the user will get a query with fewer values than requested? I don't see how the client is going to know anything is amiss.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+                  valuePreviewList.stream()
+                      .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+                      .findFirst()
+                      .get();
+              if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
+                List<String> queryValues = new ArrayList<String>();
+                DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
+                previewValue
+                    .getQueryValue()
+                    .forEach(
```
jaycarlton: I think it would be more readable to make a private helper method to do the stuff inside the foreach and return a string, and then just collect that into a list.

Also, using empty strings makes me nervous, but maybe that's how it has to work.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+                        value -> {
+                          try {
+                            Double fieldValue = Double.parseDouble(value);
+                            queryValues.add(dateFormat.format(new Date(fieldValue.longValue())));
+                          } catch (NumberFormatException ex) {
+                            queryValues.add("");
```
jaycarlton: I would log an error here, and include the value and any other context we have handy. It means the front end fed us bad information or we have a programming error on the back end. I might even make it fatal.

ui/src/app/utils/index.tsx
```diff
@@ -463,3 +463,18 @@ export function highlightSearchTerm(searchTerm: string, stringToHighlight: strin
       {word}
     </span>);
 }
+
+const BACKOFF_DEFAULT_WAIT_TIME_IN_MS = 1000;
```
jaycarlton: What's the plan for testing this retry code? Do we need a test api that can timeout after a configurable amount of time?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+        .forEach(
+            fields -> {
+              DataSetPreviewValueList previewValue =
+                  valuePreviewList.stream()
+                      .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+                      .findFirst()
+                      .get();
+              if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
+                List<String> queryValues = new ArrayList<String>();
+                DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
```
jaycarlton: Pleas pull the date format out of the loop. and into a constant.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -300,101 +299,88 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+            });
+    queryResponse
+        .getSchema()
+        .getFields()
+        .forEach(
+            fields -> {
+              DataSetPreviewValueList previewValue =
+                  valuePreviewList.stream()
+                      .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+                      .findFirst()
```
jaycarlton: This will blow up if it's not found, as it's calling get on an empty optional. Also, it might be faster and simpler to make a map from lowercased field name to previewValue.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Let me see what I can do.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Will change.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: We've determined that null is a valid response for many of these columns where there is no data. That said, I don't really like catching a null pointer. So I can poke at trying to fix that. 

Whether we keep going or not here is a product question though, and I would argue is outside the scope of this ticket.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Will move to map. I can poke at the optional and see if we can get it to throw something rather than blow up though.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Will do.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Agreed. I will log an error. I don't think we want to make it fatal, because in this case some data is still valuable to a user, but we should log something in our system

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Let me poke at the foreach's, because those might be changing anyway.

ui/src/app/utils/index.tsx
blrubenstein: I had been thinking about that, and proposed something, but people didn't like it. In this case I just manually tested, but creating some sort of unit test for it would be good. I'll see what I can do.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -552,30 +552,36 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
+          () => dataSetApi().previewDataSetByDomain(namespace, id, request));
+        const newPreviewInformation = {
+          isLoading: false,
+          values: dataSetPreviewResp.values
+        };
+        this.setState({previewList: this.state.previewList.set(dataSetPreviewResp.domain, newPreviewInformation)});
+      } catch (ex) {
+        const exceptionResponse = await ex.json() as unknown as ErrorResponse;
+        const errorText = this.generateErrorTextFromPreviewException(exceptionResponse, domain);
+        this.setState({previewError: true, previewErrorText: errorText});
```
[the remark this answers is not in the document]
blrubenstein: I've put some thought into this, and I think that we want to return data to users as soon as its available. I want to rethink the way we display errors on this page to not blow away all other error messages. Maybe as a little warning icon with a tooltip that pops up when you mouse over it with the data.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: I think that the filter is probably clear, and avoids an additional step, since we would need to filter the map anyway. I am moving to make it not blow up.

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
jaycarlton: Between this conversation and the multiple tickets and ongoing conversations we have on the dataset api and contract, maybe we need little story to get it all straightened out.

On the UI side, maybe we need a separate spinner or progress (and error) indication for each domain.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: Discussed offline, and this is just a try/catch used in place of a null check. Just a style/readability problem.

ui/src/app/pages/data/data-set/dataset-page.tsx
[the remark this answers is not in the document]
blrubenstein: We do have a separate spinner right now for each domain. I just updated it to also display a separate error per domain. There's a little bit of cleanup I want to do now, and it isn't styled/passed by Lou, but it is better in that there's no longer a popup for all data sets.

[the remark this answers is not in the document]
gjuggler: Thanks! Mea culpa, I'd missed that there were already separate spinners per-domain. Sounds like a reasonable improvement over the status quo.

blrubenstein:
@jaycarlton I think I have everything other than the bigquery exception updates, which I can't really replicate on my local dev reliably... Do you have any other changes you want made. (Noting the request changes flag right now) @gjuggler if you could take another look when you get the chance, it should be good for re-review.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -405,6 +365,43 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+  private void addFieldValuesFromBigQueryToPreviewList(List<DataSetPreviewValueList> valuePreviewList, FieldValueList fieldValueList) {
+    IntStream.range(0, fieldValueList.size())
+        .forEach(
+            columnNumber -> {
+              valuePreviewList
+                  .get(columnNumber)
+                  .addQueryValueItem(
+                      Optional.ofNullable(fieldValueList.get(columnNumber).getValue())
+                          .orElse("")
+                          .toString());
```
jaycarlton: For clarity, the `toString()` should go probably immediately after `getValue()` inside of the `ofNullable` expression, so that the optional expression will be a string in every case. (The empty string in the orElse doesn't need to be converted).

I'm surprised it also works the way it is, since it's an `Optional<Object>`

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -405,6 +365,43 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+            });
+  }
+
+  private void formatTimestampValues(List<DataSetPreviewValueList> valuePreviewList, Field fields) {
```
jaycarlton: nit: A short comment describing what this function does would help here. It could also be unit tested if there's time.

Also, why is `fields` plural here?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -405,6 +365,43 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+    DataSetPreviewValueList previewValue =
+        valuePreviewList.stream()
+            .filter(preview -> preview.getValue().equalsIgnoreCase(fields.getName()))
+            .findFirst()
+            .orElseThrow(() -> new ServerErrorException("Value should be present when it is not in dataset preview request"));
```
jaycarlton: nit: maybe `IllegalStateException`

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -405,6 +365,43 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
+    if (fields.getType() == LegacySQLTypeName.TIMESTAMP) {
+      List<String> queryValues = new ArrayList<>();
+      DateFormat dateFormat = new SimpleDateFormat(dateFormatString);
+      previewValue
+          .getQueryValue()
+          .forEach(
+              value -> {
+                if (!value.equals(emptyCellMarker)) {
+                  Double fieldValue = Double.parseDouble(value);
+                  queryValues.add(dateFormat.format(new Date(fieldValue.longValue())));
```
jaycarlton: nit: I think this works, but it seems like we'd want all the controllers to use the same data formatting function so the UI knows what to expect.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -77,12 +80,14 @@
   private final WorkspaceService workspaceService;
 
   private static int NO_OF_PREVIEW_ROWS = 20;
-  private static int PREVIEW_RETRY_LIMIT = 2;
   // See https://cloud.google.com/appengine/articles/deadlineexceedederrors for details
   private static long APP_ENGINE_HARD_TIMEOUT_MSEC_MINUS_FIVE_SEC = 55000l;
   private static String CONCEPT_SET = "conceptSet";
   private static String COHORT = "cohort";
 
+  private static final String dateFormatString = "yyyy/MM/dd HH:mm:ss";
```
gjuggler: Silly / newbie Java naming question: what's the difference between these constants and the constants above that makes them differently-cased?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: This is currently the only place we are using date formatting. I'd push back on abstracting until we have a second date formatting location.

ui/src/app/utils/index.tsx
```diff
@@ -463,3 +463,32 @@ export function highlightSearchTerm(searchTerm: string, stringToHighlight: strin
...
+ * This method will only error
+ * Parameters:
+ *    apiCall: Lambda that will run an API call, in the form of () => apiClient.apiCall(args)
+ *    maxRetries: The amount of retries the system will take before erroring
+ *    defaultWaitTime: How long the base exponential backoff is, in milliseconds
+ *      For example, if 1000 is passed in, it will wait 1s for the first retry,
+ *      2s for the second, etc.
+ */
+export async function apiCallWithGatewayTimeoutRetries<T>(
+  apiCall: () => Promise<T>, maxRetries = 3, defaultWaitTime = 1000): Promise<T> {
```
gjuggler: Sorry for another naming nit: this isn't technically a "default" but rather an "initial" value. Maybe "initialWaitTime" or "startingWaitTime"

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: That's interesting. Is this really the only place in all our UI we need to get date strings from the backend? That's a different smell, but for another day I suppose. Next time we're looking at the request structure, maybe we can just pass epoch millis as a long and let the FE deal with it.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: For posterity, I poked around at this, but we do some abstraction over the BigQueryException, which means that isn't what we get back from the `executeQuery` method. I tried to see if I could consistently get an error from bigquery in order to resolve this, but I couldn't. Talked offline with Jay, who said if I spent too long I can just punt on this for now, because it is an existing problem in the codebase.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: I think maybe swagger does some work on this, but when I searched for `.format` this was all I found (other than string formatting)

ui/src/app/utils/index.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+
+describe('IndexUtils', () => {
+  it('should not retry if successful', async() => {
+    const successfulFunctionSpy = spyOn(functionStub, 'successfulFunction').and.callThrough();
+    await apiCallWithGatewayTimeoutRetries(() => functionStub.successfulFunction());
+    expect(successfulFunctionSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('should retry three times by default', async() => {
```
gjuggler: Nice test!

Super nitpick: maybe qualify the test description with "apiCallWithGatewayTimeout" since index.tsx has a bunch of other stuff which might eventually get tested too.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: Renamed already :D

PR #2652, opened 2019-10-08
[RW-3288] [risk=low] Recent workspaces UI

als364:
The UI changes for RW-3288. On all UI accesses of workspaces or workspace resources, calls the API to update `user_recent_workspaces`.

### DIFF
--- a/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
+++ b/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
@@ -5,7 +5,7 @@ import {
   OnInit,
   ViewChild,
 } from '@angular/core';
-import {cohortsApi} from 'app/services/swagger-fetch-clients';
+import {cohortsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import {Observable} from 'rxjs/Observable';
 
 import {idsInUse, initExisting, searchRequestStore} from 'app/cohort-search/search-state.service';
@@ -47,16 +47,19 @@ export class CohortSearchComponent implements OnInit, OnDestroy {
       const cohortId = params.cohortId;
       if (cohortId) {
         this.loading = true;
-        cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId)
-          .then(cohort => {
-            this.loading = false;
-            this.cohort = cohort;
-            currentCohortStore.next(cohort);
-            if (cohort.criteria) {
-              initExisting.next(true);
-              searchRequestStore.next(parseCohortDefinition(cohort.criteria));
-            }
-          });
+        Promise.all([
+          workspacesApi().updateRecentWorkspaces(workspace.namespace, workspace.id),
+          cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId)
+        ]).then(values => {
+          const cohort = values[1];
+          this.loading = false;
+          this.cohort = cohort;
+          currentCohortStore.next(cohort);
+          if (cohort.criteria) {
+            initExisting.next(true);
+            searchRequestStore.next(parseCohortDefinition(cohort.criteria));
+          }
+        });
       } else {
         this.cohort = {criteria: '{"includes":[],"excludes":[]}'};
       }
--- a/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
+++ b/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
@@ -5,7 +5,7 @@ import {
   OnInit,
   ViewChild,
 } from '@angular/core';
-import {cohortsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
+import {cohortsApi} from 'app/services/swagger-fetch-clients';
 import {Observable} from 'rxjs/Observable';
 
 import {idsInUse, initExisting, searchRequestStore} from 'app/cohort-search/search-state.service';
@@ -47,16 +47,16 @@ export class CohortSearchComponent implements OnInit, OnDestroy {
       const cohortId = params.cohortId;
       if (cohortId) {
         this.loading = true;
-        cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId).then(values => {
-          const cohort = values[1];
-          this.loading = false;
-          this.cohort = cohort;
-          currentCohortStore.next(cohort);
-          if (cohort.criteria) {
-            initExisting.next(true);
-            searchRequestStore.next(parseCohortDefinition(cohort.criteria));
-          }
-        });
+        cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId)
+          .then(cohort => {
+            this.loading = false;
+            this.cohort = cohort;
+            currentCohortStore.next(cohort);
+            if (cohort.criteria) {
+              initExisting.next(true);
+              searchRequestStore.next(parseCohortDefinition(cohort.criteria));
+            }
+          });
       } else {
         this.cohort = {criteria: '{"includes":[],"excludes":[]}'};
       }
--- a/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
+++ b/ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
@@ -47,10 +47,7 @@ export class CohortSearchComponent implements OnInit, OnDestroy {
       const cohortId = params.cohortId;
       if (cohortId) {
         this.loading = true;
-        Promise.all([
-          workspacesApi().updateRecentWorkspaces(workspace.namespace, workspace.id),
-          cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId)
-        ]).then(values => {
+        cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId).then(values => {
           const cohort = values[1];
           this.loading = false;
           this.cohort = cohort;
--- a/ui/src/app/components/card.tsx
+++ b/ui/src/app/components/card.tsx
@@ -11,9 +11,7 @@ export const baseStyles = reactStyles({
     backgroundColor: colors.white,
     border: `1px solid ${cardBorderColor}`,
     display: 'flex',
-    flexDirection: 'column'
-  },
-  cardPadding: {
+    flexDirection: 'column',
     padding: '1rem',
   }
 });
@@ -21,24 +19,22 @@ export const baseStyles = reactStyles({
 export const styles = reactStyles({
   workspaceCard: {
     ...baseStyles.card,
+    padding: '0px',
     minWidth: '300px', maxWidth: '300px',
     minHeight: '223px', maxHeight: '223px',
   },
   domainCard: {
     ...baseStyles.card,
-    ...baseStyles.cardPadding,
     minWidth: '300px', maxWidth: '300px',
     minHeight: '223px', maxHeight: '223px',
   },
   resourceCard: {
     ...baseStyles.card,
-    ...baseStyles.cardPadding,
     minWidth: '200px', maxWidth: '200px',
     minHeight: '223px', maxHeight: '223px'
   },
   cohortActionCard: {
     ...baseStyles.card,
-    ...baseStyles.cardPadding,
     width: '30%',
     height: '12rem'
   }
--- a/ui/src/app/components/card.tsx
+++ b/ui/src/app/components/card.tsx
@@ -5,14 +5,16 @@ const cardBorderColor = colorWithWhiteness(colors.dark, 0.6);
 
 export const baseStyles = reactStyles({
   card: {
-    padding: '1rem',
+    margin: '0 1rem 1rem 0',
     borderRadius: '0.2rem',
     boxShadow: `0 0.125rem 0.125rem 0 ${cardBorderColor}`,
     backgroundColor: colors.white,
     border: `1px solid ${cardBorderColor}`,
     display: 'flex',
-    flexDirection: 'column',
-    margin: '0 1rem 1rem 0'
+    flexDirection: 'column'
+  },
+  cardPadding: {
+    padding: '1rem',
   }
 });
 
@@ -22,18 +24,27 @@ export const styles = reactStyles({
     minWidth: '300px', maxWidth: '300px',
     minHeight: '223px', maxHeight: '223px',
   },
+  domainCard: {
+    ...baseStyles.card,
+    ...baseStyles.cardPadding,
+    minWidth: '300px', maxWidth: '300px',
+    minHeight: '223px', maxHeight: '223px',
+  },
   resourceCard: {
     ...baseStyles.card,
+    ...baseStyles.cardPadding,
     minWidth: '200px', maxWidth: '200px',
     minHeight: '223px', maxHeight: '223px'
   },
   cohortActionCard: {
     ...baseStyles.card,
+    ...baseStyles.cardPadding,
     width: '30%',
     height: '12rem'
   }
 });
 
 export const WorkspaceCardBase = withStyle(styles.workspaceCard)('div');
+export const DomainCardBase = withStyle(styles.domainCard)('div');
 export const ResourceCardBase = withStyle(styles.resourceCard)('div');
 export const ActionCardBase = withStyle(styles.cohortActionCard)('div');
--- a/ui/src/app/components/popups.tsx
+++ b/ui/src/app/components/popups.tsx
@@ -1,3 +1,4 @@
+import colors from 'app/styles/colors';
 import {switchCase} from 'app/utils';
 import * as fp from 'lodash/fp';
 import * as React from 'react';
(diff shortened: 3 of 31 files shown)


ui/src/app/components/card.tsx
```diff
@@ -5,14 +5,16 @@ const cardBorderColor = colorWithWhiteness(colors.dark, 0.6);
...
     borderRadius: '0.2rem',
     boxShadow: `0 0.125rem 0.125rem 0 ${cardBorderColor}`,
     backgroundColor: colors.white,
     border: `1px solid ${cardBorderColor}`,
     display: 'flex',
-    flexDirection: 'column',
-    margin: '0 1rem 1rem 0'
+    flexDirection: 'column'
+  },
+  cardPadding: {
```
blrubenstein: Since `workspaceCard` is the only exception to the rule on this, I'd probably rather have that override the padding value, and leave the default padding as is.

ui/src/app/components/popups.tsx
```diff
@@ -6,7 +7,7 @@ import onClickOutside from 'react-onclickoutside';
 
 const styles = {
   tooltip: {
-    background: 'black', color: 'white',
+    background: 'black', color: colors.white,
```
blrubenstein: Opt: While you're changing this, should this be black? Or `colors.dark` or something similar?

ui/src/app/styles/colors.ts
```diff
@@ -25,6 +25,7 @@ export default {
     dataSet: '#6CACE4',
     notebook: '#8BC990'
   },
+  border: '#b7b7b7',
```
blrubenstein: Have we checked with Lou before adding this to the color palette? In general, I think we should tend to use colorsWithWhiteness to avoid ending up with a large number of slight variations in `colors.tsx`. It will make it easier if we want to change the theming on the site to a different color palette.

ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
```diff
@@ -48,16 +48,19 @@ export class CohortSearchComponent implements OnInit, OnDestroy {
...
-            this.loading = false;
-            this.cohort = cohort;
-            currentCohortStore.next(cohort);
-            if (cohort.criteria) {
-              initExisting.next(true);
-              searchRequestStore.next(parseCohortDefinition(cohort.criteria));
-            }
-          });
+        Promise.all([
+          workspacesApi().updateRecentWorkspaces(workspace.namespace, workspace.id),
```
blrubenstein: If this errors, I'm not sure we want to count all API endpoints as having failed here. This will mean that it rejects if any of the promises fail. I'm not sure we want to do that?

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -216,7 +216,8 @@ export const ConceptHomepage = withCurrentWorkspace()(
     }
 
     componentDidMount() {
-      this.loadDomainsAndSurveys();
+      workspacesApi().updateRecentWorkspaces(this.props.workspace.namespace, this.props.workspace.id)
```
blrubenstein: Should updateRecentWorkspaces be called with any visit? Or just with edits? If its just with edits, I don't think we want this here, and if it is with any visit, could we just update on fetch to `getWorkspace`?

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -367,30 +367,36 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+                                  See all Workspaces
+                                </span>
+                              </FlexRow>
+                              <RecentWorkspaces />
                             </FlexColumn>
-                          </FlexColumn>
-                        </FlexRow>)
+                            <FlexColumn>
+                              <SmallHeader>Recently Accessed Items</SmallHeader>
+                              <RecentWork/>
```
blrubenstein: Is this `RecentAccessedObjects` (e.g., cohorts, concept sets, etc) or `RecentWorkspaces`? If `RecentWorkspaces`, I'd probably lean towards full spelling there. If not, it would probably be good to rename this to be clearer.

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -0,0 +1,59 @@
...
+    this.loadWorkspaces();
+  }
+
+  async loadWorkspaces() {
+    try {
+      this.setState({loading: true});
+      const recentWorkspaces = await workspacesApi().getUserRecentWorkspaces();
+      this.setState({recentWorkspaces: recentWorkspaces});
+    } catch (error) {
+      console.error(error);
```
blrubenstein: What error cases do we expect we might see here? I think it is within the scope of this to consider how we can display errors to the user, or if we should. If not, it is presumably being captured by the API logs, but we should figure out what gets displayed here in the workspace stead/if we want to hide it.

ui/src/app/pages/workspace/workspace-card.tsx
```diff
@@ -0,0 +1,316 @@
+import * as React from 'react';
```
blrubenstein: 👍 To this refactor. Thank you for pulling this out.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -395,50 +395,70 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+    }
 
-      const selectedCdrIsLive = liveCdrVersions.some(cdr => cdr.cdrVersionId === this.state.workspace.cdrVersionId);
-      if (this.isMode(WorkspaceEditMode.Create) || (
-        this.isMode(WorkspaceEditMode.Duplicate) && !selectedCdrIsLive)) {
-        // We preselect the default CDR version when a new workspace is being
-        // created (via create or duplicate) with one exception: cloning a
-        // workspace which references a non-default, non-archived CDR version.
-        this.setState(fp.set(['workspace', 'cdrVersionId'], cdrResp.defaultCdrVersionId));
+    async maybeUpdateRecentWorkspaces() {
```
blrubenstein: Another place on this page where we may want to updateRecentWorkspaces is when you clone a workspace, we probably want to add the clone onto the set of recentWorkspaces. Is that handled by the API?

ui/src/app/styles/colors.ts
als364: I did check with him. I can try and do the math on how to get `b7` out of `colorsWithWhiteness`.

ui/src/app/pages/data/concept/concept-homepage.tsx
als364: With any visit. I would strongly prefer to update only on explicit UI actions so that some developer doesn't come along six months from now and not realize `getWorkspace` has a side effect.

ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
als364: Do you know of another way to force all the promises to complete before executing the code in the `.then`?

ui/src/app/pages/data/concept/concept-homepage.tsx
blrubenstein: Talked offline. This is probably doable with a HOC or wrapper component, but we are spinning that off into a separate tech debt ticket.

ui/src/app/pages/homepage/recent-workspaces.tsx
blrubenstein: Yeah, a default state might be reasonable here. 👍

ui/src/app/styles/colors.ts
blrubenstein: Yeah, I'm pushing for us to run a pretty tight ship on this file, because the refactor when we had 12 different blues was a pain in the neck, and I very much want to avoid doing that again.

ui/src/app/pages/data/concept/concept-homepage.tsx
calbach: What's the rationale for filing a tech debt ticket instead of just putting this into workspace-wrapper right now? It seems pretty directly linked to what we're attempting to track here, and it would be a much smaller change.

ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
calbach: I think the point (which I agree with), is that the result of `updateRecentWorkspaces()` is irrelevant for rendering of the page. It shouldn't be on the critical path at all, I don't think there's anything we need to do in response to that call completing, so I would make the call independently and keep `getCohort()` as it was.

ui/src/app/pages/data/concept/concept-homepage.tsx
calbach: Based on the current UI stack, if there's a workspace page which is NOT part of workspace-wrapper, that's a bug. That's because workspace-wrapper is what controls setting `currentWorkspaceStore`, which is where all the React components get workspace data (e.g. `withCurrentWorkspace()`).

Worth auditing if it can be done quickly, but also reasonable for the purposes of this PR to assume that everything can or will be covered.

als364: Looks like the thing we thought was outside the workspace wrapper component just minimizes the header is the only thing the workspace wrapper component renders. I'll move the call there.

ui/src/app/pages/homepage/recent-workspaces.tsx
als364: Added a 'Create new workspace' card when the section is not loading and there are no recent workspaces.

ui/src/app/cohort-search/cohort-search/cohort-search.component.ts
```diff
@@ -47,16 +47,16 @@ export class CohortSearchComponent implements OnInit, OnDestroy {
...
-          .then(cohort => {
-            this.loading = false;
-            this.cohort = cohort;
-            currentCohortStore.next(cohort);
-            if (cohort.criteria) {
-              initExisting.next(true);
-              searchRequestStore.next(parseCohortDefinition(cohort.criteria));
-            }
-          });
+        cohortsApi().getCohort(workspace.namespace, workspace.id, cohortId).then(values => {
```
calbach: revert file?

ui/src/app/pages/data/cohort-review/cohort-review.spec.tsx
```diff
@@ -1,18 +1,21 @@
+
 import {shallow} from 'enzyme';
 import * as React from 'react';
 
 import {registerApiClient} from "app/services/swagger-fetch-clients";
 import {currentWorkspaceStore} from 'app/utils/navigation';
-import {CohortBuilderApi, CohortReviewApi, CohortsApi, WorkspaceAccessLevel} from 'generated/fetch';
+import {CohortBuilderApi, CohortReviewApi, CohortsApi, WorkspaceAccessLevel, WorkspacesApi} from 'generated/fetch';
```
calbach: revert?

ui/src/app/pages/data/cohort-review/cohort-review.tsx
```diff
@@ -6,7 +6,7 @@ import {Modal, ModalFooter, ModalTitle} from 'app/components/modals';
 import {SpinnerOverlay} from 'app/components/spinners';
 import {CreateReviewModal} from 'app/pages/data/cohort-review/create-review-modal';
 import {cohortReviewStore, visitsFilterOptions} from 'app/services/review-state.service';
-import {cohortBuilderApi, cohortReviewApi, cohortsApi} from 'app/services/swagger-fetch-clients';
+import {cohortBuilderApi, cohortReviewApi, cohortsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
```
calbach: revert?

ui/src/app/pages/analysis/interactive-notebook.tsx
```diff
@@ -101,20 +101,19 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
     }
 
     componentDidMount(): void {
-      workspacesApi().readOnlyNotebook(this.props.urlParams.ns,
-        this.props.urlParams.wsid, this.props.urlParams.nbName)
-        .then(html => {
-          this.setState({html: html.html});
-          this.loadNotebookActivityTracker();
-        });
+      const {ns, wsid, nbName} = this.props.urlParams;
```
calbach: opt: revert? I like the change, though it's now unrelated to the PR

ui/src/app/pages/data/concept/concept-set-details.tsx
```diff
@@ -16,7 +16,7 @@ import {PopupTrigger, TooltipTrigger} from 'app/components/popups';
 import {SpinnerOverlay} from 'app/components/spinners';
 import {EditComponentReact} from 'app/icons/edit';
 import {ConceptTable} from 'app/pages/data/concept/concept-table';
-import {conceptSetsApi} from 'app/services/swagger-fetch-clients';
+import {conceptSetsApi, workspacesApi} from 'app/services/swagger-fetch-clients';
```
calbach: revert

ui/src/app/pages/data/data-page.tsx
```diff
@@ -11,7 +11,13 @@ import {ResourceCard} from 'app/components/resource-card';
 import {SpinnerOverlay} from 'app/components/spinners';
 import {NotebookResourceCard} from 'app/pages/analysis/notebook-resource-card';
 import {CohortResourceCard} from 'app/pages/data/cohort/cohort-resource-card';
-import {cohortReviewApi, cohortsApi, conceptSetsApi, dataSetApi} from 'app/services/swagger-fetch-clients';
+import {
```
calbach: revert file changes

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -10,14 +10,16 @@ import {
   ConceptsApi,
   ConceptSetsApi,
   DataSetApi,
-  WorkspaceAccessLevel
+  WorkspaceAccessLevel,
```
calbach: revert

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -16,7 +16,7 @@ import {
   cohortsApi,
   conceptsApi,
   conceptSetsApi,
-  dataSetApi
+  dataSetApi, workspacesApi
```
calbach: revert

ui/src/app/pages/homepage/recent-work.spec.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import {registerApiClient} from 'app/services/swagger-fetch-clients';
 import {currentWorkspaceStore} from 'app/utils/navigation';
-import {RecentWork} from 'app/pages/homepage/recent-work';
+import {RecentWork} from 'src/app/pages/homepage/recent-resources';
```
calbach: This import looks wrong

ui/src/app/pages/homepage/recent-work.spec.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
```
calbach: Update this filename as well?

als364: noticed most of these immediately after pressing the re-request review button. On it, thank you

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -355,8 +356,7 @@ export const Homepage = withUserProfile()(class extends React.Component<
           <FadeBox style={styles.fadeBox}>
             {/* The elements inside this fadeBox will be changed as part of ongoing
             homepage redesign work*/}
-            <div style={{display: 'flex', justifyContent: 'center'}}>
-              <FlexColumn style={styles.singleCard}>
+            <div style={{display: 'flex', justifyContent: 'flex-start', flexDirection: 'column'}}>
```
blrubenstein: Nit: This can be a <FlexColumn> with just `justifyContent: 'flex-start'`

ui/src/app/pages/workspace/workspace-card.tsx
```diff
@@ -0,0 +1,317 @@
...
+                backgroundColor: colors.workspacePermissionsHighlights[accessLevel]
+              }}
+             data-test-id='workspace-access-level'
+            >
+              {accessLevel}
+            </div>
+            <div
+              style={{
+                fontSize: 12,
+                lineHeight: '17px'
```
blrubenstein: Nit: I think you can just make this `lineHeight: 17` to match above.

ui/src/app/pages/analysis/interactive-notebook.tsx
als364: keeping this

PR #2658, opened 2019-10-09
[RW-3651][Risk=no] Move from toString to map

blrubenstein:
This addresses the bug in RW-3651, by using map as opposed to directly calling `toString`, which would NPE when there was no object in the optional.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -7,6 +7,7 @@ import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.QueryJobConfiguration;
 import com.google.cloud.bigquery.TableResult;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import java.sql.Date;
@@ -85,7 +86,7 @@ public class DataSetController implements DataSetApiDelegate {
   private static String COHORT = "cohort";
 
   private static final String DATE_FORMAT_STRING = "yyyy/MM/dd HH:mm:ss";
-  private static final String EMPTY_CELL_MARKER = "";
+  public static final String EMPTY_CELL_MARKER = "";
 
   private static final Logger log = Logger.getLogger(DataSetController.class.getName());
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -368,7 +368,8 @@ private void addFieldValuesFromBigQueryToPreviewList(
               valuePreviewList
                   .get(columnNumber)
                   .addQueryValueItem(
-                      Optional.ofNullable(fieldValueList.get(columnNumber).getValue().toString())
+                      Optional.ofNullable(fieldValueList.get(columnNumber).getValue())
```
ericsong: May be my lack of familiarity with Optionals but this reads off a bit weirdly to me.

Is it applying the map function to the non null value from Optional? If not -> ""?

In my head, I think what makes more sense to me is using Optional to get the non null value and then applying `toString` to it.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -368,7 +368,8 @@ private void addFieldValuesFromBigQueryToPreviewList(
...
+                          .map(Object::toString)
```
ericsong: is this something we can easily add a test for? If its low effort because of existing testing infrastructure, I think it's worth adding.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: That's basically what it does. The `Optional.map(Object::toString)` turns an Optional<Object> to an Optional<String>. If the first one is empty, then the second one stays empty, and the `orElse` provides the default value of an empty string.

The most confusing thing is that `getValue()` is the thing that can be null.

It might read better if you declare `final Object value = fieldValueList.get(columnNumber).getValue();` and then pack that into the `ofNullable`.

Overall, I'd say Optional.map() is one of the most powerful and functional things you get from Optional.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: +1 I think . `FieldValueUtils` class to do some of these mappings would be more readable and testable than doing it inline in this controller.

blrubenstein: It isn't particularly low effort, unfortunately. :( We have some basic bigquery stubbing, but not a ton, and not with much detail. I can poke around at improving that, but want to get this fix in to un-break things in the time being.

blrubenstein: I may be able to test just this function by constructing a fieldvaluelist? Which seems less valuable but faster to implement. Any thoughts on that?

ericsong:
Looks good. 👍 

In general, I prefer to test at a higher level to avoid getting into implementation details in tests but this is a good low effort solution.

blrubenstein:
Yeah, this at the very least should prevent this bug from happening again. I think we have some higher level tests for this api, but they don't have nullable values. It's hard to catch all cases at the top level.

PR #2660, opened 2019-10-09
[RW-3611][Risk=no] Add fields parameter to listWorkspaces

blrubenstein:
I am taking the same approach we took to parameterizing getWorkspace to say that the tests for this should be covered under rainforest/puppeteer, since it is ultimately an integration style change.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
@@ -352,10 +352,10 @@ public class FireCloudServiceImpl implements FireCloudService {
   }
 
   @Override
-  public List<WorkspaceResponse> getWorkspaces() {
+  public List<WorkspaceResponse> getWorkspaces(List<String> fields) {
     return retryHandler.run(
         (context) ->
-            workspacesApiProvider.get().listWorkspaces(FIRECLOUD_LIST_WORKSPACES_REQUIRED_FIELDS));
+            workspacesApiProvider.get().listWorkspaces(fields));
   }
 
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -350,7 +353,9 @@ public WorkspaceResponse getWorkspace(String projectName, String workspaceName)
 
   @Override
   public List<WorkspaceResponse> getWorkspaces() {
```
calbach: I'd prefer to have this take the field list as a parameter. Right now this specifies a seemingly arbitrary (looking just at this class) subset of fields which in reality needs to be the supserset of all fields which are used currently throughout the application. At a conceptual level, the service shouldn't know this. Tying this instead to the callsite which knows what fields are needed for that context would be much more explicit.

blrubenstein: That seems reasonable to me. I was mirroring what we did for getWorkspace, but since there's only one callsite for this, it makes it lot easier to avoid duplicated code.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -87,6 +87,9 @@
           "workspace.bucketName",
           "workspace.createdBy");
 
+  public static final List<String> FIRECLOUD_LIST_WORKSPACES_REQUIRED_FIELDS =
```
calbach: rm

PR #2662, opened 2019-10-09
[RW-3657][risk=no] Record Billing Buffer entries before FC request

ericsong:


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -103,8 +104,15 @@ public void syncBillingProjectStatus() {
           default:
             break;
         }
+      } catch (NotFoundException e) {
+        log.log(Level.WARNING, "Get BillingProjectStatus call failed for " + entry.getCreationTime() + ". Project not found.");
+
+        if ((getCurrentTimestamp().getTime() - entry.getCreationTime().getTime()) > minutesToMs(CREATING_TIMEOUT_MINUTES)) {
```
calbach: Is this necessary? I think `cleanBillingBuffer` already covers this without modification.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -103,8 +104,15 @@ public void syncBillingProjectStatus() {
           default:
             break;
         }
+      } catch (NotFoundException e) {
```
calbach: opt: One thing that I found would have been useful when debugging this issue is if L140 had some associated logging. One theory I had is we were hitting the Creating timeout.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -103,8 +104,15 @@ public void syncBillingProjectStatus() {
...
+        log.log(Level.WARNING, "Get BillingProjectStatus call failed for " + entry.getCreationTime() + ". Project not found.");
```
calbach: I would also ensure you log the original exception, i.e. as the third parameter to `log()`

PR #2664, opened 2019-10-10
[no ticket]Decrease billing project buffer size to 300 in perf env

aweng98:


jmthibault79:
@jaycarlton No - this is the max size of the billing buffer.  Larger numbers here mean we go thru quota faster, but we can create workspaces faster too.  This does nothing to resolve quota concerns.

PR #2665, opened 2019-10-10
[RW-3518][Risk=no] Srubenst/data set variable names

blrubenstein:
The product ideal that Karthik and I talked about was to have the dataset variables qualified by the id, and then a comment at the top describing what dataset it corresponded to. ID ended up falling through, because there were no guarantees the dataset was created in the DB at this point, so I moved to random eight numeric characters. Eight was an arbitrary number I chose as likely enough to avoid conflicts, but few enough that it doesn't create too much clutter in the notebook.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
@@ -1,15 +1,7 @@
 package org.pmiops.workbench.api;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.eq;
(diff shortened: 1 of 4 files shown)


api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -736,8 +751,8 @@ public void testGetQueryTwoDomains() {
                 workspace.getNamespace(), WORKSPACE_NAME, KernelTypeEnum.PYTHON.toString(), dataSet)
             .getBody();
     verify(bigQueryService, times(2)).executeQuery(any());
-    assertThat(response.getCode()).contains("blah_condition_df");
-    assertThat(response.getCode()).contains("blah_drug_df");
+    assertThat(response.getCode()).contains("condition_df");
+    assertThat(response.getCode()).contains("drug_df");
```
andy7i: these names seem more reasonable.

PR #2666, opened 2019-10-10
[RW-3496][risk=no] Data Dictionary API - Part 2 - Add GET endpoint

ericsong:
Pretty straightforward. 

Most notable change is the addition of Mapstruct which I used to do the db -> api translation for DataDictionaryEntry


ericsong:
@calbach I accidentally addressed your comments for Part 1 in this PR.

My plan, once both parts are approved, is to merge Part 2 into Part 1 and then Part 1+2 into master.

Note - merging this PR will add some fake data to our various environments. I think this is OK given that it doesn't impact the real data set and we can also clean it up fairly easily with some delete statements.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -609,6 +624,32 @@ private void formatTimestampValues(List<DataSetPreviewValueList> valuePreviewLis
     return ResponseEntity.ok(dataSetResponse);
   }
 
+  @Override
+  public ResponseEntity<DataDictionaryEntry> getDataDictionaryEntry(
+      Long cdrVersionId, String domain, String domainValue) {
```
calbach: Enum instead of string for `domain`?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -609,6 +624,32 @@ private void formatTimestampValues(List<DataSetPreviewValueList> valuePreviewLis
...
+    CdrVersion cdrVersion = cdrVersionDao.findByCdrVersionId(cdrVersionId);
+
+    if (cdrVersion == null) {
+      throw new BadRequestException("Invalid CDR Version");
+    }
+
+    String omopTable = conceptSetDao.DOMAIN_TO_TABLE_NAME.get(Domain.fromValue(domain));
+
```
calbach: ultranit: visually I think it is more readable to group logical statements together, i.e. dropping the newline between the lookup and error condition checks - right now there's basically just one newline between every statement, which doesn't help the eye.

api/src/test/java/org/pmiops/workbench/api/DataDictionaryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    User user() {
+      return currentUser;
+    }
+  }
+
+  private DataSetController dataSetController;
+
+  @Before
+  public void setUp() {
+    dataSetController =
```
calbach: Is this line necessary? Can you just `@Autowire` L91?

api/src/test/java/org/pmiops/workbench/api/DataDictionaryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    dataDictionaryEntry.setFieldType("C");
+    dataDictionaryEntry.setDataProvenance("D");
+    dataDictionaryEntry.setSourcePpiModule("E");
+    dataDictionaryEntry.setTransformedByRegisteredTierPrivacyMethods(true);
+
+    dataDictionaryEntryDao.save(dataDictionaryEntry);
+
+    org.pmiops.workbench.model.DataDictionaryEntry response =
+        dataSetController
+            .getDataDictionaryEntry(cdrVersion.getCdrVersionId(), domain.toString(), domainValue)
```
calbach: nit: this isn't really exercising the matching functionality since there's only one entry in the database, may want to insert another entry

api/src/test/java/org/pmiops/workbench/api/DataDictionaryTest.java
```diff
@@ -0,0 +1,156 @@
...
+            dataSetMapper,
+            dataSetService,
+            fireCloudService,
+            notebooksService,
+            userProvider,
+            workspaceService);
+  }
+
+  @Test
+  public void testGetDataDictionaryEntry() {
```
calbach: May be worth covering a few boundary situations, namely domainValue not found

calbach:
Why would this load bad data outside of local dev? I don't see anything that is autloading this into other environments. Related: do you intend to have this feature flagged?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
ericsong: I was unable to use object types in the parameter if I wanted to keep these as query parameters. I think it is possible if I change it to be a request body parameter but it seemed wrong to change the look of the API because of a swagger limitation.

FYI - OpenAPI 3.0 supports this. One more reason to upgrade.

I did end up changing the query parameters to path parameters because that felt more correct.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
```diff
@@ -0,0 +1,15 @@
...
+import org.mapstruct.Mapping;
+import org.pmiops.workbench.db.model.DataDictionaryEntry;
+import org.pmiops.workbench.utils.CommonMappers;
+
+@Mapper(
+    componentModel = "spring",
+    uses = {CommonMappers.class})
+public interface DataSetMapper {
+
+  @Mapping(target = "cdrVersionId", source = "dbModel.cdrVersion.cdrVersionId")
```
jaycarlton: Is it possible to fully qualify both the source and the target? That might make the line easier to scan in Find results, etc.

api/src/main/java/org/pmiops/workbench/db/model/DataDictionaryEntry.java
```diff
@@ -133,4 +136,40 @@ public void setTransformedByRegisteredTierPrivacyMethods(
       Boolean transformedByRegisteredTierPrivacyMethods) {
     this.transformedByRegisteredTierPrivacyMethods = transformedByRegisteredTierPrivacyMethods;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
jaycarlton: I would've thought that one of our frameworks could define this method for us. This looks good, but how do we remember to update it if we add a new field?

api/src/test/java/org/pmiops/workbench/api/DataDictionaryTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Autowired CdrVersionDao cdrVersionDao;
+  @Autowired DataDictionaryEntryDao dataDictionaryEntryDao;
+
+  private static final Instant NOW = Instant.now();
+  private static final FakeClock CLOCK = new FakeClock(NOW, ZoneId.systemDefault());
+  private static User currentUser;
+
+  @TestConfiguration
+  @Import({DataSetMapperImpl.class})
```
jaycarlton: I couldn't find this class in either PR. Is it generated somehow?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
calbach: Unfortunate, thanks - sounds good.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
ericsong: What would that look like? I'm not sure if there's something we can preprend to `target = cdrVersionId`.

api/src/test/java/org/pmiops/workbench/api/DataDictionaryTest.java
ericsong: Yep! Mapstruct generates this, fairly readable too. It's nice to be able to see what it produces for debugging.

api/src/main/java/org/pmiops/workbench/db/model/DataDictionaryEntry.java
ericsong: We don't 🙃. I think this is a +1 reason to adopt some tooling like Kotlin or Immutables.

jaycarlton: Yeah, the savings in developer time and bug opportunities are huge.

ericsong:
@calbach @jaycarlton is this good to merge?

PR #2667, opened 2019-10-10
[RW-3653] [risk=no] Fix bug in Hibernate table mapping and add missing @Transactional to getRecentWorkspaces

als364:
In `UserRecentWorkspace`, the `@Column` annotation for `getLastAccessDate` was in camel case and should have been in snake case.

In `WorkspaceServiceImpl`, calls to `getRecentWorkspaces` that resulted in `user_recent_workspaces` needing to be pruned threw an error on the `.deleteByWorkspaceIdIn` because Spring thinks it should have been part of a transaction and it was not. So added a `@Transactional` to `getRecentWorkspaces` since that's where we get the data we pass in to `pruneInaccessibleRecentWorkspaces`.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/model/UserRecentWorkspace.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/UserRecentWorkspace.java
@@ -45,7 +45,7 @@ public class UserRecentWorkspace {
     this.workspaceId = workspaceId;
   }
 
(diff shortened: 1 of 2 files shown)


als364:
It fixed a `NeedsTransactionException` I was getting whenever `pruneInaccessibleRecentWorkspaces` was called.

PR #2668, opened 2019-10-10
[risk=no][no ticket]Use FlexColumn and FlexRow where possible

jmthibault79:
💪


jmthibault79:
Do you mean other instances of `<div style={{foo...}}/>` with shared patterns like those replaced here?  I would guess that there are.

PR #2669, opened 2019-10-10
[RW-3612][RISK=NO] Action Audit Service & Other Goodies - Part 1

jaycarlton:
Introduce the `ActionAuditService` and related classes. Several enum types as described in the design doc, along with an adapter service especially for the `WorkspacesController`, so as not to clutter that class and be independently testable.

Noticed many calls to `WorkspaceMapper` were static calls masquerading as instance calls. Made all those methods static and removed the service-ness of that class. I'm hoping MapStruct may make most of that class moot.

There are some unit tests. More are planned as part of another subtask, but this is a reasonable stopping point I think.


api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
```diff
@@ -233,7 +233,7 @@
   private static User currentUser;
 
   @TestConfiguration
-  @Import({WorkspaceServiceImpl.class, WorkspaceMapper.class})
+  @Import({WorkspaceServiceImpl.class, WorkspaceConversionUtils.class})
```
ericsong: This no longer needs to be auto wired right?

api/src/test/java/org/pmiops/workbench/audit/ActionAuditServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+            .map(p -> (JsonPayload) p)
+            .collect(ImmutableList.toImmutableList());
+
+    assertThat(
+            payloads.stream()
+                .map(JsonPayload::getDataAsMap)
+                .map(entry -> entry.get(AuditColumn.ACTION_ID.name()))
+                .distinct()
+                .count())
+        .isEqualTo(1);
```
ericsong: is the goal of this test to ensure that duplicate events are not recorded? If so, I think that could be useful info to append to the test name

api/src/test/java/org/pmiops/workbench/audit/ActionAuditServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  @Test
+  public void testNullPayloadDoesNotThrow() {
+    actionAuditService.send((Collection<ActionAuditEvent>) null);
+  }
+}
```
ericsong: nit: I think another useful test would be simple one that hard codes the serialized enum names and verifies they are the same.

This adds another step to changing the enum fields and reduces the likelihood of accidental changes.

api/src/test/java/org/pmiops/workbench/audit/AuditDataGeneratorTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.pmiops.workbench.audit.synthetic.AuditDataGenerator;
+
+public class AuditDataGeneratorTest {
+
+  @Test
+  public void itGetsLongInRange() {
+    final long low = 1000L;
+    final long high = 2000L;
+    final long random = AuditDataGenerator.randomLongInRange(low, high);
+    assertThat(low < high).isTrue();
```
ericsong: what is this asserting? `1000L` will always be less than `2000L` right?

api/src/test/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+    dbWorkspace1 = WorkspaceConversionUtils.toDbWorkspace(workspace1);
+    dbWorkspace1.setWorkspaceId(WORKSPACE_1_DB_ID);
+    dbWorkspace1.setLastAccessedTime(new Timestamp(now));
+    dbWorkspace1.setLastModifiedTime(new Timestamp(now));
+    dbWorkspace1.setCreationTime(new Timestamp(now));
+  }
+
+  @Test
+  public void testFiresCreateWorkspaceEvents() {
+    workspaceAuditAdapterService.fireCreateAction(workspace1, WORKSPACE_1_DB_ID);
```
ericsong: can we test one level higher by creating workspaces through our controllers? I think that adds a bit more coverage .

api/src/test/java/org/pmiops/workbench/api/ConceptsControllerTest.java
jaycarlton: Exactly. I have changes for that locally.

api/src/test/java/org/pmiops/workbench/audit/ActionAuditServiceTest.java
jaycarlton: Renamed to `testSendsMultipleEventsAsSingleAction`. We still have multiple events, but they're not considered duplicated. Just grouped with the actionId column.

api/src/test/java/org/pmiops/workbench/audit/ActionAuditServiceTest.java
jaycarlton: That reminds me I need to update the enums to give snake_case titles (as this happens whether I want it or not when the JsonPayload is constructed).

api/src/test/java/org/pmiops/workbench/audit/AuditDataGeneratorTest.java
jaycarlton: Ugh, good catch. I'm pulling this whole data generator and test out in the next PR though. At least until I have a need to write a command to invoke it.

api/src/test/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceTest.java
ericsong: Adding a few tests to `WorkspacesControllerTest` works too. I just want to make sure the whole flow is captured through our tests. It's fine if it's split up.

api/src/main/java/org/pmiops/workbench/audit/AbstractAuditableEvent.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.logging.Payload.JsonPayload;
+import com.google.cloud.logging.Severity;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public abstract class AbstractAuditableEvent {
+  private static final MonitoredResource MONITORED_RESOURCE =
+      MonitoredResource.newBuilder("global").build();
+  private static final String LOG_NAME = "action-audit-test-3";
```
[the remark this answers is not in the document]
jaycarlton: 👍

api/src/main/java/org/pmiops/workbench/audit/AbstractAuditableEvent.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public abstract long timestamp();
```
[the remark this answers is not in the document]
jaycarlton: I tried that initially, but the Logging library choked on it when building the `JsonPayload`. So it would just be an exercise in converting it and immediately switching back to `long`. 

I'm debating even keeping it, since there are at least two other timestamps associated with the LogEntry that I get for free.

api/src/main/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterService.java
```diff
@@ -0,0 +1,9 @@
+package org.pmiops.workbench.audit.adapters;
+
+import org.pmiops.workbench.model.Workspace;
+
+public interface WorkspaceAuditAdapterService {
+  void fireCreateAction(Workspace createdWorkspace, long dbWorkspaceId);
+
+  void fireDeleteAction(org.pmiops.workbench.db.model.Workspace dbWorkspace);
```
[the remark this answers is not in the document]
jaycarlton: I'd love to ideally. The problem is, the api request doesn't have the long ID (which I need), and the delete call doesn't use the API model. I don't trust the mapping utilities either, so this seemed like the least intrusive way to go. I can revisit that when I do the update instrumentation. 

Since I don't want to depend on the `WorkspacesController` (as it depends on me), any improvements in signature consistency will come at the cost of extra grungy code.

api/src/main/java/org/pmiops/workbench/audit/synthetic/AuditDataGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.function.Supplier;
+import java.util.stream.IntStream;
+import org.elasticsearch.common.UUIDs;
+import org.pmiops.workbench.audit.ActionAuditEvent;
+import org.pmiops.workbench.audit.ActionAuditEvent.Builder;
+import org.pmiops.workbench.audit.ActionType;
+import org.pmiops.workbench.audit.AgentType;
+import org.pmiops.workbench.audit.TargetType;
+
+public class AuditDataGenerator {
```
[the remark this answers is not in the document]
jaycarlton: yes, but it's coming out

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -269,8 +263,25 @@ private FirecloudWorkspaceId generateFirecloudWorkspaceId(String namespace, Stri
...
+    workspaceAuditAdapterService.fireCreateAction(createdWorkspace, dbWorkspace.getWorkspaceId());
+    return ResponseEntity.ok(createdWorkspace);
+  }
+
+  private void setDbWorkspaceFields(
+      org.pmiops.workbench.db.model.Workspace dbWorkspace,
+      User user,
+      FirecloudWorkspaceId workspaceId,
+      org.pmiops.workbench.firecloud.model.Workspace fcWorkspace,
+      Timestamp now) {
```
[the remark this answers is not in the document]
jaycarlton: This was an auto-extraction where previously we had repeated code. In both cases the timestamp is getting reused later, so I'll just rename the argument.

api/src/test/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceTest.java
jaycarlton: All there was to do was verify we're calling the adapter service.

api/src/main/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterService.java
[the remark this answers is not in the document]
jaycarlton: I'll revisit this on the next go-round. The main goal of this interface design is minimal intrusion into the controller code.

api/src/main/java/org/pmiops/workbench/audit/targetproperties/WorkspaceTargetProperty.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.collect.ImmutableMap;
+import java.util.AbstractMap;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Function;
+import org.pmiops.workbench.model.Workspace;
+
+public enum WorkspaceTargetProperty {
```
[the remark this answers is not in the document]
jaycarlton: thanks

PR #2670, opened 2019-10-11
[RW-3666][RW-3670][Risk=no] Clone datasets with workspace

blrubenstein:

### DIFF
--- a/api/db/changelog/db.changelog-104-dataset-name-unique-in-workspace.xml
+++ b/api/db/changelog/db.changelog-104-dataset-name-unique-in-workspace.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="srubenst" id="db.changelog-104-dataset-name-unique-in-workspace">
+    <dropUniqueConstraint tableName="data_set"
+                          uniqueColumns="name"
+                          constraintName="name"/>
+    <addUniqueConstraint tableName="data_set"
+                         columnNames="name, workspace_id"
+                         constraintName="data_set_unique_name_per_workspace"/>
+  </changeSet>
+</databaseChangeLog>
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
@@ -4,8 +4,10 @@ import com.google.cloud.bigquery.QueryJobConfiguration;
 import java.sql.Timestamp;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import org.pmiops.workbench.db.model.DataSet;
 import org.pmiops.workbench.db.model.DataSetValues;
+import org.pmiops.workbench.db.model.Workspace;
 import org.pmiops.workbench.model.DataSetRequest;
 import org.pmiops.workbench.model.KernelTypeEnum;
 import org.pmiops.workbench.model.PrePackagedConceptSetEnum;
@@ -32,4 +34,7 @@ public interface DataSetService {
       KernelTypeEnum kernelTypeEnum,
       String dataSetName,
       Map<String, QueryJobConfiguration> queryJobConfigurationMap);
+
+  DataSet cloneDataSetToWorkspace(
+      DataSet fromDataSet, Workspace toWorkspace, Set<Long> cohortIds, Set<Long> conceptSetIds);
 }
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
@@ -6,7 +6,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import org.pmiops.workbench.db.model.DataSet;
-import org.pmiops.workbench.db.model.DataSetValues;
+import org.pmiops.workbench.db.model.DataSetValue;
 import org.pmiops.workbench.db.model.Workspace;
 import org.pmiops.workbench.model.DataSetRequest;
 import org.pmiops.workbench.model.KernelTypeEnum;
@@ -22,7 +22,7 @@ public interface DataSetService {
       long workspaceId,
       List<Long> cohortIdList,
       List<Long> conceptIdList,
-      List<DataSetValues> values,
+      List<DataSetValue> values,
       PrePackagedConceptSetEnum prePackagedConceptSetEnum,
       long creatorId,
       Timestamp creationTime);
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
@@ -37,4 +37,6 @@ public interface DataSetService {
 
   DataSet cloneDataSetToWorkspace(
       DataSet fromDataSet, Workspace toWorkspace, Set<Long> cohortIds, Set<Long> conceptSetIds);
+
+  List<DataSet> getDataSets(Workspace workspace);
 }
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -30,7 +30,7 @@ import org.pmiops.workbench.db.model.Cohort;
 import org.pmiops.workbench.db.model.CommonStorageEnums;
 import org.pmiops.workbench.db.model.ConceptSet;
 import org.pmiops.workbench.db.model.DataSet;
-import org.pmiops.workbench.db.model.DataSetValues;
+import org.pmiops.workbench.db.model.DataSetValue;
 import org.pmiops.workbench.db.model.Workspace;
 import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.exceptions.NotFoundException;
@@ -166,7 +166,7 @@ public class DataSetServiceImpl implements DataSetService {
       long workspaceId,
       List<Long> cohortIdList,
       List<Long> conceptIdList,
-      List<DataSetValues> values,
+      List<DataSetValue> values,
       PrePackagedConceptSetEnum prePackagedConceptSetEnum,
       long creatorId,
       Timestamp creationTime) {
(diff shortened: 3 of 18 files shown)


api/db/changelog/db.changelog-104-dataset-name-unique-in-workspace.xml
```diff
@@ -0,0 +1,15 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="srubenst" id="db.changelog-104-dataset-name-unique-in-workspace">
+    <dropUniqueConstraint tableName="data_set"
+                          uniqueColumns="name"
+                          constraintName="name"/>
+    <addUniqueConstraint tableName="data_set"
```
jaycarlton: So this won't require changing any existing data since it's a relaxation of the existing constraint.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -534,6 +537,28 @@ private String getQualifiedColumnName(Domain currentDomain, String columnName) {
...
   }
 
+  @Override
+  @Transactional
+  public DataSet cloneDataSetToWorkspace(
+      DataSet fromDataSet, Workspace toWorkspace, Set<Long> cohortIds, Set<Long> conceptSetIds) {
+    DataSet toDataSet = new DataSet(fromDataSet);
+    toDataSet.setWorkspaceId(toWorkspace.getWorkspaceId());
+    toDataSet.setCreatorId(toWorkspace.getCreator().getUserId());
+    toDataSet.setLastModifiedTime(toWorkspace.getLastModifiedTime());
```
jaycarlton: Why are you using the destination workspace's modified time instead of the source dataset's? Same question for creation time.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
```diff
@@ -47,12 +51,28 @@ public DataSet(
...
+    setName(dataSet.getName());
+    setVersion(DataSet.INITIAL_VERSION);
+    setDescription(dataSet.getDescription());
+    setInvalid(dataSet.getInvalid());
+    setIncludesAllParticipants(dataSet.getIncludesAllParticipants());
+    setValues(
+        dataSet.getValues().stream()
+            .map(
+                dataSetValue ->
+                    new DataSetValues(dataSetValue.getDomainId(), dataSetValue.getValue()))
```
jaycarlton: You could also make a copy constructor for `DataSetValues`, which would simplify the lambda in the `map` call to `DatasetValues::new`.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
```diff
@@ -18,6 +19,7 @@
 @Entity
 @Table(name = "data_set")
 public class DataSet {
```
jaycarlton: This value class doesn't have `equals()` or `hashCode()` defined, which could bite you if you try to use it as a map key or set element.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -391,17 +396,36 @@ public Workspace updateWorkspaceAcls(
 
   @Override
   @Transactional
-  public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to) {
+  public Workspace saveAndCloneCohortsConceptSetsAndDataSets(Workspace from, Workspace to) {
```
jaycarlton: nit: it might be time to come up with a collective term for workspace "stuff", that is top level items like datasets, cohorts, and concept sets. Then the function would be `saveAndCloneWorkspaceStuff()` or whatever.

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -185,17 +187,18 @@
   })
   @MockBean({
     BillingProjectBufferService.class,
+    CdrVersionService.class,
```
jaycarlton: 👏 🔤

api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
```diff
@@ -32,4 +34,9 @@ DataSet saveDataSet(
       KernelTypeEnum kernelTypeEnum,
       String dataSetName,
       Map<String, QueryJobConfiguration> queryJobConfigurationMap);
+
+  DataSet cloneDataSetToWorkspace(
+      DataSet fromDataSet, Workspace toWorkspace, Set<Long> cohortIds, Set<Long> conceptSetIds);
+
+  List<DataSet> getDataSets(Workspace workspace);
```
jaycarlton: aside: This feels a bit like feature envy. I.e. it feels like it should live on the `Workspace` model. I understand that's not feasible in any of the workspace classes we have though.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -391,17 +396,36 @@ public Workspace updateWorkspaceAcls(
...
+          conceptSetService
+              .cloneConceptSetAndConceptIds(fromConceptSet, to, cdrVersionChanged)
+              .getConceptSetId());
+    }
+    for (DataSet dataSet : dataSetService.getDataSets(from)) {
+      dataSetService.cloneDataSetToWorkspace(
+          dataSet,
+          to,
+          toCohortIdByFromCohortId.values().stream()
+              .filter(cohortId -> dataSet.getCohortSetId().contains(cohortId))
```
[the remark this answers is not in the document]
ericsong: I just tested this out locally with your branch and I believe cohortIds and conceptSetIds are not being copied over.
 Can you verify?

If so, I think a test and variable name changes like the one in my previous comment would be good changes.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
blrubenstein: What Eric said :D 

I actually would have preferred it on the workspace model, so objects can fetch all of their information, but this seemed reasonable to me, and I did not have a strong opinion, so went with Eric's thoughts on it.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Yup! I thought about recreating `now` here, but decided to follow the existing pattern.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
blrubenstein: Let me add a `.equals()`

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
blrubenstein: Sure! I'll make a copy constructor.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -391,17 +396,36 @@ public Workspace updateWorkspaceAcls(
...
   @Override
   @Transactional
-  public Workspace saveAndCloneCohortsAndConceptSets(Workspace from, Workspace to) {
+  public Workspace saveAndCloneCohortsConceptSetsAndDataSets(Workspace from, Workspace to) {
     // Save the workspace first to allocate an ID.
     Workspace saved = workspaceDao.save(to);
     CdrVersionContext.setCdrVersionNoCheckAuthDomain(saved.getCdrVersion());
     boolean cdrVersionChanged =
         from.getCdrVersion().getCdrVersionId() != to.getCdrVersion().getCdrVersionId();
+    Map<Long, Long> toCohortIdByFromCohortId = new HashMap<>();
```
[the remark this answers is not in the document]
blrubenstein: Sure, I can move it to `fromCohortIdToToCohortId

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
blrubenstein: I don't have a good word for that that is descriptive yet. Maybe `workspaceObjects`? I can think on it while doing the other edits, but if you have any ideas, happy to take them.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
blrubenstein: Let me check about those. There are some weird variable names here, so I want to change those as well.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: I think it would be clearer to declare a `now` explicitly and use it. As written, it's definitely suggesting we want the workspace's times used for the dataset's times, and looks like a bug. So maybe a comment and/or declaration that this is deliberate (and why it's desired) would help.

It's also confusing that the new dataset's creator is the same as the old *workspace's* creator. In the case where the fromDataset was itself cloned into the fromWorkspace, this may well not be correct. But I guess we need to define (and document) the semantics: is the cloning user the creator, or the original dataset's creator, or the original workspace's creator?

Also, I'd consider the case where I take a year-old dataset and clone it to a new workspace (or an even older workspace). I certainly wouldn't expect a cloned workspace to have timestamps older than the creation time of the source dataset.

api/src/main/java/org/pmiops/workbench/db/model/DataSet.java
jaycarlton: IJ can give you one for free.

Additionally, would a copy constructor in this class help you with implementing `cloneDataSetToWorkspace()`?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -534,6 +537,28 @@ private String getQualifiedColumnName(Domain currentDomain, String columnName) {
         .collect(Collectors.toList());
   }
 
+  @Override
+  @Transactional
+  public DataSet cloneDataSetToWorkspace(
```
jaycarlton: fwiw @jmthibault79 tells me that we're not calling this `clone` anymore, but that it's `duplicate`.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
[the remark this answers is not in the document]
blrubenstein: Filed https://precisionmedicineinitiative.atlassian.net/secure/RapidBoard.jspa?rapidView=30&projectKey=RW&view=planning&selectedIssue=RW-3692&epics=visible

PR #2671, opened 2019-10-11
[no ticket][risk=no]use ResourceTypeDisplayNames to simplify code

andy7i:
Extra refactoring work found during another PR.

This change removes the various forms of converting the `ResourceType` enum to a display name, in favor of using `ResourceTypeDisplayNames` which is defined once.


ui/src/app/components/confirm-delete-modal.tsx
```diff
@@ -35,15 +33,15 @@ export interface ConfirmDeleteModalState {
...
-  }
-
   constructor(props: ConfirmDeleteModalProps) {
     super(props);
     this.state = {loading: false};
   }
 
+  get resourceTypeDisplayName(): any {
+    return ResourceTypeDisplayNames.get(this.props.resourceType);
+  }
```
andy7i: Ideally I'd prefer to have this programmatically generated in the props, but I'm not sure if that's possible. Open to any feedback for any ideas!

jaycarlton: Like you want `this.props.resourceType.displayName` defined, or you want a property named `resourceTypeDisplayName`? I guess you already have the latter, which is probably cleaner anyway.

ui/src/app/utils/resourceActions.ts
```diff
@@ -27,6 +27,10 @@ export const ResourceTypeDisplayNames = new Map()
   .set(ResourceType.INVALID, 'Invalid')
   .set(ResourceType.WORKSPACE, 'Workspace');
 
+export function getResourceTypeDisplayName(resourceType: ResourceType) {
```
jaycarlton: It'd be cool if the enum values themselves supported a method to get the name. Then you could do `myResourceType.getDisplayName()`. 

This is great too though. Don't spend more than a couple of minutes experimenting if you want to try that.

andy7i: so.. after looking at it with fresh eyes, I don't think I see any reason why we need a separate Map or an enum with different values from the display name. So now the enum contains the display names. Any thoughts/objections to this approach?

andy7i:
Jay re-reviewed and thumbed on slack

PR #2673, opened 2019-10-13
[RW-3672][risk=no] Ingest Data Dictionary exports

ericsong:
This PR adds the real data dictionary export files as well as a parser that can read off entries into the database.

I think there is still an issue with the Data Dictionary export format regarding how to reference the CDR Version but I decided to make the change to the imported files as we need to get this in soon. We can update the parser in a follow up after we discuss with the Data Dictionary team.

The given CDR version references the `bigquery_dataset` field while I was working with the assumption that it would reference a unique column such as `name` or `cdr_version_id`.

I decided to bring in Kotlin with this PR as I had a need to make several POJO classes and wanted to cut down on the amount of boiler plate I had to write. I think we can still have a discussion on the extent to which we use Kotlin but using it as a POJO generator seemed fairly non controversial to me.


api/tools/src/main/java/org/pmiops/workbench/tools/MetaData.kt
```diff
@@ -1,10 +1,8 @@
 package org.pmiops.workbench.tools
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import java.sql.Timestamp
 
-@JsonIgnoreProperties(ignoreUnknown = true)
-data class MetaData(
+internal data class MetaData(
```
jaycarlton: nit: I'd spell it as one word: `Metadata`

.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - run:
           name: Integration tests compile
           working_directory: ~/workbench/api
-          command: ./project.rb gradle compileBigQuery compileIntegration
+          command: ./project.rb gradle compileBigquerytestJava compileIntegrationJava
```
jaycarlton: why is the t in test lowercase? Likewise, why not have a `Test` after `Integration`?

api/build.gradle
```diff
@@ -197,12 +198,15 @@ buildscript {    // Configuration for building
     classpath 'org.owasp:dependency-check-gradle:5.1.0'
     classpath 'com.diffplug.spotless:spotless-plugin-gradle:3.23.0'
     classpath 'net.ltgt.gradle:gradle-apt-plugin:0.21'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$KOTLIN_VERSION"
```
jaycarlton: Nit: can these be alphabetized, or is the order important?  Sometimes default versions and implicit dependencies mean the latter but accidental dependence on ordering can be avoided by adopting a canonical (alphabetical) order. Or at least that was my understanding of how it worked with Maven.

api/tools/src/main/java/org/pmiops/workbench/tools/AvailableField.kt
```diff
@@ -0,0 +1,12 @@
...
+internal data class AvailableField(
+        val field_name: String,
+        val relevant_omop_table: String,
+        val omop_cdm_standard_or_custom_field: String,
+        val description: String,
+        val field_type: String,
+        val data_provenance: String,
+        val source_ppi_module: String?,
+        val transformed_by_registered_tier_privacy_methods: Boolean
+)
```
jaycarlton: nit: I think most of us have IJ configured to end on a newline

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -30,7 +32,7 @@
   private Resource[] getDataDictionaryExportFiles() throws IOException {
```
jaycarlton: I'd name this `getDataDictionaryResources` so you don't give callers the knowledge that it's implemented with flat files.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -41,41 +43,48 @@ public CommandLineRunner run(
 
       for (Resource resource : getDataDictionaryExportFiles()) {
         InputStream is = resource.getInputStream();
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
```
jaycarlton: Why are we declaring a new mapper for every resource? It looks like you could reuse one (and its format, module, etc) and just pass in the resource's `InputStream` at line 51.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -41,41 +43,48 @@ public CommandLineRunner run(
...
 
-        BufferedReader reader = new BufferedReader(new InputStreamReader(is));
-        Timestamp definedTime = new Timestamp(Long.parseLong(reader.readLine()) * 1000);
+        mapper.registerModule(new KotlinModule());
+        DataDictionary dd = mapper.readValue(is, DataDictionary.class);
 
-        String line;
-        while ((line = reader.readLine()) != null) {
-          String[] fields = line.split(":");
+        Timestamp newEntryDefinedTime = dd.getMeta_data()[0].getCreated_time();
```
jaycarlton: I think you should guard against an empty metadata array here and below.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -41,41 +43,48 @@ public CommandLineRunner run(
...
+          if (entry != null && newEntryDefinedTime.before(entry.getDefinedTime())) {
             continue;
           }
 
           if (entry == null) {
             entry = new DataDictionaryEntry();
-            entry.setRelevantOmopTable(relevantOmopTable);
-            entry.setFieldName(fieldName);
-            entry.setCdrVersion((defaultCdrVersion));
+            entry.setRelevantOmopTable(field.getRelevant_omop_table());
```
jaycarlton: nit: is there a way to chain these or define a builder for DDE?

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -41,41 +43,48 @@ public CommandLineRunner run(
...
+        if (cdrVersion == null) {
+          // Skip over Data Dictionaries for CDR Versions not in the current environment
+          continue;
+        }
+
+        for (AvailableField field : dd.getTransformations()[0].getAvailable_fields()) {
           DataDictionaryEntry entry =
               dataDictionaryEntryDao.findByRelevantOmopTableAndFieldNameAndCdrVersion(
-                  relevantOmopTable, fieldName, defaultCdrVersion);
+                  field.getRelevant_omop_table(), field.getField_name(), cdrVersion);
```
jaycarlton: nit: can we not set up the find methods in the dao class to return an `Optional<DataDictionaryEntry>`? Seems like that could consolidate both null checks.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
```diff
@@ -41,41 +43,48 @@ public CommandLineRunner run(
...
 
-          entry.setDefinedTime(definedTime);
-          entry.setOmopCdmStandardOrCustomField(fields[2]);
-          entry.setDescription(fields[3]);
-          entry.setFieldType(fields[4]);
-          entry.setDataProvenance(fields[5]);
-          entry.setSourcePpiModule(fields[6]);
-          entry.setTransformedByRegisteredTierPrivacyMethods("true".equals(fields[7]));
+          entry.setDefinedTime(newEntryDefinedTime);
+          entry.setOmopCdmStandardOrCustomField(field.getOmop_cdm_standard_or_custom_field());
```
jaycarlton: Can your `ObjectMapper` help with some of this boilerplate?

.circleci/config.yml
ericsong: not sure. These tasks were auto generated. It did worry me a little that these had to be changed but the builds are passing.

api/build.gradle
ericsong: Not sure. I thiiiink the order might matter in cases where you have version conflicts if two packages import the same package.

Given that, I don't think I want to reorder what's already existing but I can move the two new imports into a location that is "ordered".

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
ericsong: I am OK with failing the script here if this throws an exception. It indicates a problem with the import file and it should be looked into before deploying.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
ericsong: I'm inclined to say it's not worth the effort for 1 time use.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
ericsong: I looked into it but I don't think it'll save us any code in this case because of the name mismatches. It is possible to override name mismatches but we would then be looking at at least an equal number of lines from override statements.

.circleci/config.yml
jaycarlton: Do you mind double checking what changed? There may be things that silently broke if name conventions got switched.

api/build.gradle
jaycarlton: that works for now. I need to learn more about gradle.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
jaycarlton: Yeah, but those lines would live in the mapper where it can be reused and not spilled out in this unrelated function, so I think it's still a win.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
jaycarlton: I thought we got that for free with Kotlin. But from what I'm reading, only Kotlin clients can take advantage of named parameters.

api/tools/src/main/java/org/pmiops/workbench/tools/DataDictionary.kt
```diff
@@ -0,0 +1,3 @@
+package org.pmiops.workbench.tools
+
+internal class DataDictionary(val meta_data: Array<Metadata>, val transformations: Array<Transformations>)
```
als364: I dunno whether we have strong style opinions on stuff like this but I would much rather see each arg on its own line, it's easier for me to parse that way

als364: nit: also, metadata should be all one word. Is the snake case because that is what Kotlin prefers for variables?

api/build.gradle
```diff
@@ -657,3 +664,14 @@ task loadDataDictionary(type: JavaExec) {
...
+compileKotlin {
+  kotlinOptions {
+    jvmTarget = "1.8"
+  }
+}
+compileTestKotlin {
+  kotlinOptions {
+    jvmTarget = "1.8"
+  }
+}
```
blrubenstein: Nit: Should end in a newline

api/tools/src/main/java/org/pmiops/workbench/tools/DataDictionary.kt
blrubenstein: Do you know if the spotless linter works on kotlin files? If so, this is the line format is probably enforced by spotless. If not, we should probably find and decide on a kotlin linter.

api/tools/src/main/resources/data_dictionary_exports/AllOfUsDatasetV2.yaml
```diff
@@ -0,0 +1,6392 @@
+meta_data:
```
blrubenstein: Is there a strong reason to host this in our codebase, as opposed to static files in GCS? (Maybe GCS, unsure. There's several ways to host it probably)

Do we expect to be maintaining these yamls? Or have them handed to us?

blrubenstein: I'd especially worry as we roll out N cdr versions, we end up with N data dictionary export files here.

jaycarlton: +1. If we do need it in github, maybe a separate repo, but GCS is very easy in terms of setup, automation, and interactive use for non-programmers.  This is more content than code.

Thinking more broadly, if we could collect all the auxiliary information we need with each CDR release (including ds_linking and and other schema-ish knowledge) into one "place" or "category" apart from the main line of source control, that would be helpful.

.circleci/config.yml
ericsong: Looked into it.

Looks like the gradle Java plugin auto generates `compile___Java` tasks for all source sets, `bigquerytest` being one of ours. https://github.com/all-of-us/workbench/blob/master/api/build.gradle#L273
https://docs.gradle.org/current/userguide/java_plugin.html#java_source_set_tasks

This generates the task `compileBigquerytestJava` but because gradle supports auto completion of tasks, it was able to map `compileBigQuery` -> `compileBigquerytestJava` even though `compileBigQuery` technically does not exist. 

Adding the Kotlin plugin generates another set of `compile___Kotlin` tasks for all source sets. At this point, gradle can no longer auto complete `compileBigQuery` since there are two possibilities with `compileBigquerytestJava` and `compileBigquerytestKotlin`.


I looked through the rest of our codebase and I couldn't find other instances where the addition of new gradle tasks would cause auto completion ambiguity.

api/build.gradle
```diff
@@ -194,6 +195,8 @@ buildscript {    // Configuration for building
     classpath 'com.google.cloud.tools:appengine-gradle-plugin:1.3.5'
     classpath 'gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.12.0'
     classpath 'io.swagger:swagger-codegen:2.2.3'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$KOTLIN_VERSION"
```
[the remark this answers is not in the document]
ericsong: https://groovy-lang.org/syntax.html#_double_quoted_string

double quotes support string interpolation, which I am using at the end.

We should probably just use double quotes everywhere.

api/tools/src/main/java/org/pmiops/workbench/tools/DataDictionary.kt
ericsong: it does! I added one. thanks for the suggestion

api/tools/src/main/resources/data_dictionary_exports/AllOfUsDatasetV2.yaml
ericsong: Nope. This was taken as the path of least resistance. I have filed https://precisionmedicineinitiative.atlassian.net/browse/RW-3705 as the follow up to address this.

api/build.gradle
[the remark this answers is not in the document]
als364: Aha, thanks for the clarification.

api/tools/src/main/resources/data_dictionary_exports/AllOfUsDatasetV2.yaml
jaycarlton: Sounds good. It's not hurting anything yet.

api/build.gradle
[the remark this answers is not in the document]
jaycarlton: So for Ruby, JavaScript, and I believe Python, where there's the same distinction, the general style guidance seems to be single quotes in every case except where you are doing interpolation. The main reason I like that approach is that when I see a single quoted string, I know it's just a literal, and don't have to use too many neurons on it.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
jaycarlton: Anyway, I'd argue builders are worth it for single-time use, especially if you can enforce required parameters. Not necessary at this point in the game for this PR though.

api/tools/src/main/java/org/pmiops/workbench/tools/AvailableField.kt
jaycarlton: 🏆

.circleci/config.yml
jaycarlton: Thanks for chasing that down. This is an odd convention having a prefix and suffix like that. Is it possible to have something horrendous like `compileBigQuery_test_Java` or something to distinguish the parts?

api/libproject/devstart.rb
```diff
@@ -1685,6 +1685,8 @@ def deploy(cmd_name, args)
     versions_file = get_cdr_versions_file(ctx.project)
     update_cdr_versions_for_project("config/#{versions_file}", op.opts.dry_run)
 
+    common.run_inline %W{gradle loadDataDictionary}
```
calbach: Should support a dry run flag, as the other commands here do

PR #2676, opened 2019-10-14
RW-3536 renaming alias to match alias names in ds_linking table.

freemabd:
renaming alias to match alias names in ds_linking table.


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -198,7 +198,7 @@ public DataSet saveDataSet(
           .put(Domain.OBSERVATION, "observation")
           .put(Domain.PERSON, "person")
           .put(Domain.PROCEDURE, "procedure")
-          .put(Domain.SURVEY, "survey")
+          .put(Domain.SURVEY, "answer")
```
jaycarlton: Good catch. I had to guess a few of these because we didn't seem to be using all the available domains yet (or have them all in the ds linking table).

PR #2677, opened 2019-10-14
[RW-3394][risk=low] Support non-CDR queries via BigQueryService

calbach:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
+++ b/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
@@ -39,10 +39,17 @@ public class BigQueryService {
   private static final Logger logger = Logger.getLogger(BigQueryService.class.getName());
 
   @Autowired private Provider<WorkbenchConfig> workbenchConfigProvider;
+  @Autowired private BigQuery defaultBigQuery;
 
   @VisibleForTesting
   protected BigQuery getBigQueryService() {
     CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
+    if (cdrVersion == null) {
+      return defaultBigQuery;
+    }
+    // If a query is being executed in the context of a CDR, it must be run within that project as
+    // well. By default, the query would run in the Workbench App Engine project, which would
+    // violate VPC-SC restrictions.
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/BigQueryService.java
```diff
@@ -39,10 +39,17 @@
   private static final Logger logger = Logger.getLogger(BigQueryService.class.getName());
 
   @Autowired private Provider<WorkbenchConfig> workbenchConfigProvider;
+  @Autowired private BigQuery defaultBigQuery;
 
   @VisibleForTesting
   protected BigQuery getBigQueryService() {
     CdrVersion cdrVersion = CdrVersionContext.getCdrVersion();
+    if (cdrVersion == null) {
```
jaycarlton: This should work, but it looks a bit like an afterthought.

So if there's a "CdrVersionContext", you can't use the default BQ service? I feel like there's a more Springy way to do this bifurcation where you inject the instance of the BigQueryService you need via a name or some other way. Like you'd have a `CdrBigQueryService` and a `DefaultBigQueryService`, and get a BigQueryServiceProvider or some such to give you the right one.

You could even go so far as to rename it and arrange things such that clients of the `CdrQueryService` don't need to know that it uses BigQuery internally.

The other question is, how does a caller know which flavor of the service she got?

calbach: Yes, I agree. Honestly this service is a bit of a wart anyways so I didn't want to spend too much time yak shaving here, but short version: the caller cares that their query executes successfully, they don't really care about which version they got - this achieves that, resulting in a 403 failure if misconfigured. This configuration is very secondary to the purpose of what the client is trying to do - setting the billing project for executing the query.

- I considered binding two versions of the service, but this seemed like a bit too much Spring magic to get working. 
- I also considered splitting this into two service versions with a common base. This wasn't awful, but it would require some untangling of a grab-bag of utility methods which are mostly here for testing isolation. I'd really like to have a better story for BigQuery fake/mocking before I invest too much more in this service.

PR #2678, opened 2019-10-14
[RW-3684][risk=no] Tune background image transition points on account creation

calbach:
The new demographics survey has larger controls, where the overlap is more noticeable. Though this was already a problem on existing creation pages.

Transition point (increased window size a few pixels between screenshots):
image

image


calbach:
No, I tried installing an extension to capture a GIF but it didn't capture the resizing properly. I've captured either side of the transition point with the above screenshots.

calbach:
Checked with Karthik on slack to verify this is what he was looking for.

PR #2680, opened 2019-10-15
[RW-3691][risk=low] Bump prod project buffer size from 20 -> 100

calbach:


jaycarlton:
Since this is in preparation for the workshop, do we have a follow-on ticket to restore the value afterwards?

calbach:
I don't think it's necessary to reduce it, we'd probably want this number or higher for public launch. 20 was chosen somewhat arbitrarily.

calbach:
That's higher than expected, lets follow-up in slack.

PR #2684, opened 2019-10-16
[RW-2812][RW-3579][risk=no] Fix participant detail table styles

dolbeew:
- Get rid of extra white space on rows with single-line names ([RW-2812](https://precisionmedicineinitiative.atlassian.net/browse/RW-2812))
- Adjust spacing in column headers ([RW-3579](https://precisionmedicineinitiative.atlassian.net/browse/RW-3579))

Before:
Screen Shot 2019-10-16 at 12 15 43 PM

After:
Screen Shot 2019-10-16 at 12 15 55 PM


ericsong:
@dolbeew Your api-local-test failure is due to an issue we had with Circle when the new MySQL image got pushed onto DockerHub. It should be fixed if you merge in master.

PR #2686, opened 2019-10-16
[risk=no] patch version of mysql in circle

ericsong:


.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
   api-local-test:
     docker:
       - image: allofustest/workbench:buildimage-0.0.13
-      - image: mysql:5.7
+      - image: mysql:5.7.27
```
calbach: opt: May want a comment about why this is pinned to a minor version, something like `# TODO: Try reverting to 5.7 when newer releases are published, mysql:5.7.28 results in startup failures`

PR #2688, opened 2019-10-16
[risk=no] lint before tests in circle

ericsong:
My preference would be to run the lint check BEFORE our tests because it runs much faster and it will exit early if the PR can't merge due to linting issues. 

Example scenario:
- Upload PR / respond to feed back / prepare to merge
- Tests run for ~20 min
- Linter fails 
- Fix and rerun Circle ~20 min

Proposed scenario:
- Upload PR / respond to feed back / prepare to merge
- Linter fails ~1 m
- Fix and rerun Circle ~20 min

Time saved - ~20m

Downside - linter failures are less important than test failures when evaluating a branch. Could negatively impact workflows where people run circle before lunch/meetings and come back to linter failures instead of tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,11 +39,6 @@ jobs:
           working_directory: ~/workbench/api
           command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
       - run:
-          # Lint before tests because it's much faster
-          name: Java linting
-          working_directory: ~/workbench/api
-          command: ./gradlew spotlessCheck
-      - run:
           name: Integration tests compile
           working_directory: ~/workbench/api
           command: ./project.rb gradle compileBigQuery compileIntegration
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -241,16 +236,24 @@ jobs:
...
+  lint:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+        name: Lint Typescript
+        working_directory: ~/workbench/ui
+        command: yarn run lint
+      - run:
+	name: Java/Kotlin linting
```
jaycarlton: Looks like some tabs snuck in 👻

calbach:
With some newer Circle functionality, we can achieve this same result without the downside of skipping unit test runs when lint failures, or the overhead of another Circle job (since linting is really only a small fraction of the runtime). Specifically we can move linting up, then add the condition `when:  always` to the unit test steps:

https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute

PR #2692, opened 2019-10-17
[risk=no] updating  _mapping_* tables name changes to _ext.

freemabd:
updating  _mapping_* tables name changes to _ext.

### DIFF
--- a/api/db-cdr/generate-cdr/make-bq-data.sh
+++ b/api/db-cdr/generate-cdr/make-bq-data.sh
@@ -460,7 +460,7 @@ and (c.count_value > 0 or c.source_count_value > 0)
 group by c.domain_id) c
 where d.domain_id = c.domain_id"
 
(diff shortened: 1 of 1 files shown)


api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -460,17 +460,17 @@ and (c.count_value > 0 or c.source_count_value > 0)
 group by c.domain_id) c
 where d.domain_id = c.domain_id"
 
-if [[ "$tables" == *"_mapping_"* ]]; then
+if [[ "$tables" == *"_ext" ]]; then
```
calbach: Could you also add a comment here which explains why we only want EHR counts here? And when/why we expect the mapping tables to be present/missing?

freemabd: Done.

PR #2694, opened 2019-10-17
[RW-2866][Risk=no] Srubenst/data dictionary ui

blrubenstein:

### DIFF
--- a/ui/src/app/components/popups.tsx
+++ b/ui/src/app/components/popups.tsx
@@ -242,14 +242,25 @@ export const Popup = fp.flow(
   }
 });
 
-export class PopupTrigger extends React.Component {
+interface PopupTriggerProps {
+  children: any;
+  closeOnClick?: boolean;
+  content: any;
+  closeOnClickOutside?: boolean;
+  onOpen?: () => void;
+  onClose?: () => void;
+  side?: string;
+}
+
+export class PopupTrigger extends React.Component<PopupTriggerProps> {
   static readonly defaultProps = {
     closeOnClick: false,
+    closeOnClickOutside: true,
     onOpen: () => {},
     onClose: () => {}
   };
 
-  props: any;
+  props: PopupTriggerProps;
   state: any;
   id: string;
 
@@ -287,7 +298,7 @@ export class PopupTrigger extends React.Component {
       })}
       {open && <Popup
           target={this.id}
-          handleClickOutside={() => this.close()}
+          handleClickOutside={this.props.closeOnClickOutside ? () => this.close() : () => {}}
           outsideClickIgnoreClass={this.id}
           onClick={closeOnClick ? () => this.close() : undefined}
           {...props}
--- a/ui/src/app/components/popups.tsx
+++ b/ui/src/app/components/popups.tsx
@@ -67,7 +67,7 @@ export const withDynamicPosition = () => WrappedComponent => {
       this.animation = requestAnimationFrame(() => this.reposition());
       const newDimensions = {
         element: fp.pick(['width', 'height'], this.element.current.getBoundingClientRect()),
-        target: fp.pick(['top', 'top-left', 'top-right', 'bottom', 'bottom-left', 'bottom-right', 'left', 'right'],
+        target: fp.pick(['top', 'bottom', 'left', 'right'],
           document.getElementById(target).getBoundingClientRect()),
         viewport: {width: window.innerWidth, height: window.innerHeight}
       };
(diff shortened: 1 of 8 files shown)


blrubenstein:
I have marked this ticket as blocked because some of the domains/values are failing in local, and I don't think we should expose the UI until those are fixed/ready. @ericsong FYI.

Still looking for review because this code is/should be independent of those API changes.

ui/src/app/components/popups.tsx
```diff
@@ -242,11 +244,20 @@ export const Popup = fp.flow(
   }
 });
 
-export class PopupTrigger extends React.Component {
+interface PopupTriggerProps {
```
jmthibault79: IIRC we started a norm of naming these "Props" and only getting more specific on export

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -210,16 +227,107 @@ const Subheader = (props) => {
...
+      parseInt(currentWorkspaceStore.getValue().cdrVersionId, 10),
+      domain.toString(),
+      domainValue.value).then(dataDictionaryEntry => {
+        this.setState({dataDictionaryEntry});
+      }).catch(e => {
+        this.setState({dataDictionaryEntryError: true});
+      });
+  }
+
+  showDataDictionaryEntry() {
```
jmthibault79: I prefer `toggle` DDE for this name

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -781,14 +889,24 @@ const DataSetPage = fp.flow(withCurrentWorkspace(), withUrlParams())(
...
                               selectedDomainValuePairs)}/>
                         )}
                       </div>)
                     }
                   </div>
+                  {valueSets.length > 0 && <FlexRow style={{
+                    width: '100%', height: '1.375rem', backgroundColor: colorWithWhiteness(colors.dark, 0.9),
+                    color: colors.primary, paddingLeft: '0.4rem', fontSize: '13px', lineHeight: '16px',
+                    alignItems: 'center'}}>
+                    <a href={'https://aousupporthelp.zendesk.com/hc/en-us/articles/' +
```
jmthibault79: This link does not work for me

ui/src/app/components/popups.tsx
blrubenstein: That makes sense, however, this is a file with many different components, and so there would be name collisions. I am not sure I would consider PopupTrigger the 'main' component of this this file? (it's a little less clear in shared component files)

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -102,12 +102,16 @@ describe('DataSet', () => {
         .find('input').first().simulate('change');
 
       await waitOneTickAndUpdate(wrapper);
+      await waitOneTickAndUpdate(wrapper);
```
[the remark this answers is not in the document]
blrubenstein: Will remove. Should be unneeded.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -52,6 +53,22 @@ import {Column} from 'primereact/column';
...
+    fontSize: '16px',
+    color: colors.primary,
+    textTransform: 'uppercase'
+  } as React.CSSProperties,
+  dataDictionarySubheader: {
+    fontSize: '13px',
+    fontWeight: 600,
+    color: colors.primary,
+    paddingTop: '0.5rem'
+  },
```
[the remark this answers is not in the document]
blrubenstein: Sure, will do.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: Will change.

ui/src/app/pages/data/data-set/dataset-page.tsx
blrubenstein: It shouldn't. I talked to Karthik, we don't have a zendesk for any environment but prod, and SSO will error out the site if you aren't coming from prod. This is problematic because it means we can't easily test these links from the site, but this link should work if you are coming from a web browser authed as a researchallofus account

ui/src/app/components/popups.tsx
jmthibault79: OK - let's keep it like this

ui/src/app/pages/data/data-set/dataset-page.tsx
jmthibault79: Link is correct when viewed as my Prod persona, thanks

PR #2695, opened 2019-10-17
[RW-3710][RISK=NO] Fix foreign key error in clone 

jaycarlton:
A call to `save` was ignoring the result, and thus not capturing the cohort ID (leaving it at zero). Later, the foreign key between dataset and cohort tables could not be satisfied b/c cohort 0 did not exist.

Unit test updates to follow.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohorts/CohortCloningService.java
+++ b/api/src/main/java/org/pmiops/workbench/cohorts/CohortCloningService.java
@@ -26,15 +26,16 @@ public class CohortCloningService {
 
   @Transactional
   public Cohort cloneCohortAndReviews(Cohort fromCohort, Workspace targetWorkspace) {
-    Cohort toCohort =
+    final Cohort duplicatedCohort =
         cohortFactory.duplicateCohort(
             fromCohort.getName(), targetWorkspace.getCreator(), targetWorkspace, fromCohort);
-    cohortDao.save(toCohort);
+    final Cohort toCohort = cohortDao.save(duplicatedCohort);
     copyCohortAnnotations(fromCohort, toCohort);
 
     for (CohortReview fromReview : fromCohort.getCohortReviews()) {
-      CohortReview toReview = cohortFactory.duplicateCohortReview(fromReview, toCohort);
-      toReview = cohortReviewDao.save(toReview);
+      final CohortReview duplicatedReview =
+          cohortFactory.duplicateCohortReview(fromReview, toCohort);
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/cohorts/CohortCloningService.java
```diff
@@ -26,15 +26,16 @@
 
   @Transactional
   public Cohort cloneCohortAndReviews(Cohort fromCohort, Workspace targetWorkspace) {
-    Cohort toCohort =
+    final Cohort duplicatedCohort =
         cohortFactory.duplicateCohort(
             fromCohort.getName(), targetWorkspace.getCreator(), targetWorkspace, fromCohort);
-    cohortDao.save(toCohort);
+    final Cohort toCohort = cohortDao.save(duplicatedCohort);
```
ericsong: nit: maybe we can pass the call of `duplicateCohort` directly into this save call? We shouldn't be using `duplicatedCohort` after this point right?

same comment for the save below.

jaycarlton: I rarely find nested calls to be more readable. In this case, with multiple arguments that themselves already call methods, I don't think that would be a win. 

The other reason is for step-ability, although IJ is getting really good at that and making it less of an issue.

ericsong: I don't think its more readable either. I think the benefit is that it decreases the likelihood of misuse. I'm guessing we should only be using `toCohort` and any calls made on `duplicatedCohort` can be faulty.

jaycarlton: The usual solution to avoiding variable reuse is keeping scopes small. There are plenty of ways to misuse any number of variables. I don't love the convention of using the same structure for both unsaved and saved models. Ultimately, I think the confusion you get from nesting easily outweighs the potential to use the wrong variable. They're both in scope only a short time, and I felt that calling out that the return value is important could help with avoiding the situation that caused this bug.

jaycarlton: OK, I just now realized that the [save](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html#save-S-) method can mutate its parameter in arbitrary ways. So yes, that variable is more of a live wire than I gave it credit for. In the particular case of APIs that feel the need to alter their arguments in unpredictable ways, avoiding keeping that parameter around is definitely better.

PR #2696, opened 2019-10-17
[no ticket][risk=no] add @Transactional to updateRecentWorkspaces

als364:
It's throwing `TransactionRequiredExceptions` because we're making two db calls in the same function

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
@@ -574,6 +574,7 @@ public class WorkspaceServiceImpl implements WorkspaceService {
   }
 
   @Override
+  @Transactional
   public UserRecentWorkspace updateRecentWorkspaces(
       Workspace workspace, long userId, Timestamp lastAccessDate) {
     Optional<UserRecentWorkspace> maybeRecentWorkspace =


api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -574,6 +574,7 @@ public Workspace setPublished(
   }
 
   @Override
+  @Transactional
```
calbach: Try temporarily reverting this and adding `@Transactional(propagation = Propagation.NOT_SUPPORTED)` to WorkspaceServiceTest, see if that reproduces the issue.

als364: It reproduces the issue even if I keep this annotation.

calbach: Unfortunate.. maybe, have you verified via manual testing whether this change fixes the issue?

als364: Still waiting for a server to spin up. This way it runs all the tests on someone else's computer

als364:
This is now verified via manual testing (and one new unit test)

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -270,6 +270,46 @@ public void updateRecentWorkspaces() {
...
+        workspace -> {
+          // Need a new 'now' each time or else we won't have lastAccessDates that are different
+          // from each other
+          workspaceService.updateRecentWorkspaces(
+              workspace,
+              USER_ID,
+              Timestamp.from(NOW.minusSeconds(mockWorkspaces.size() - workspace.getWorkspaceId())));
+        });
+    long OTHER_USER_ID = 2L;
+    workspaceService.updateRecentWorkspaces(
```
calbach: To reproduce the bug, shouldn't this happen before the other updates?

als364: Argh, yes, you're right.

api/src/main/java/org/pmiops/workbench/db/dao/UserRecentWorkspaceDao.java
```diff
@@ -11,5 +11,5 @@
 
   Optional<UserRecentWorkspace> findFirstByWorkspaceIdAndUserId(long workspaceId, long userId);
 
-  void deleteByWorkspaceIdIn(Collection<Long> workspaceIds);
+  void deleteByUserIdAndWorkspaceIdIn(long userId, Collection<Long> ids);
```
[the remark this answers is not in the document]
als364: I could do that, but in the Service I have to look at Workspace entities to check Rawls for permissions changes, so I would have to match those up with the UserRecentWorkspace entities in that function. I'd rather do the slightly more complex query and the moderately less complex business logic.

PR #2698, opened 2019-10-18
[RW-3719][risk=low] cron xml -> yaml, support per-env alternate configs

calbach:
- cron_default.yaml specifies the default cron configuration
- cron_{env}.yaml specifies an alternate cron configuration
- Disable all perf crons for now (later, we could choose an alternate more limited schedule if desired)
- Switch to yaml wasn't strictly necessary, but is more easily future extensible, in case we want to allow merge/override of yaml in the future instead


api/db/vars.env
```diff
@@ -15,6 +15,8 @@ WORKBENCH_DB_PASSWORD=wb-notasecret
 CDR_DB_USER=workbench
 CDR_DB_PASSWORD=wb-notasecret
 
+WORKBENCH_ENV=local
```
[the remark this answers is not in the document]
calbach: Done

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,13 @@
+require 'fileutils'
```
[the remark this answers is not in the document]
calbach: Done

api/src/main/webapp/WEB-INF/cron_perf.yaml
```diff
@@ -0,0 +1 @@
+cron:
```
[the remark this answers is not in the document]
calbach: RE separate PR: I think I'd prefer to leave it in, since this PR doesn't really do anything without at least one environment override.

RE cron disablement design: I filed a ticket and put my rationale in there, PTAL.

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# Future work: it will likely be desirable to support extension rather than
+# wholesale file replacement, e.g. to alter just the schedule or exclude a
+# particular cron for a given environment.
+require 'fileutils'
+
+env = ARGV[0]
+
+cron_dir = "src/main/webapp/WEB-INF"
+to_cron_path = ->(env) { "#{cron_dir}/cron_#{env}.yaml" }
```
jmthibault79: Codacy doesn't like this line.  I don't know Ruby so I have no personal opinion.

appengine-cron-emulator/build.gradle
```diff
@@ -13,7 +13,8 @@ repositories {
 
 dependencies {
     compile group: 'com.squareup.okhttp', name: 'okhttp', version: '2.7.5'
-    compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-xml', version: '2.9.9'
+    compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-yaml', version: '2.9.9'
```
jmthibault79: Is there any way for these to use the variables like `JACKSON_VERSION` in `build.gradle`?

api/libproject/generate_appengine_cron_yaml.rb
calbach: Thanks - I agree in this case, fixed.

appengine-cron-emulator/build.gradle
calbach: Good call. I refactored them to use a variable here.

I'm not sure if you were asking if I could reference the values we already have in `api/build.gradle`; I'll probably keep them separate for now as I don't know if there's much benefit and it would create a circular dep (at least, logically).

api/src/main/webapp/WEB-INF/cron_perf.yaml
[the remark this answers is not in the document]
calbach: I'm backtracking and going with your original suggestion to split out the perf change. It's Friday and I don't know that we've established we definitely want crons off in perf

appengine-cron-emulator/build.gradle
jmthibault79: Didn't realize the dependency.  This works, thanks.

gjuggler:
Thanks for the separate ticket & rationale – more than enough to convince me of the wisdom of keeping the configurability in the crons.yaml.

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
+# Generates the App Engine cron.yaml file for a given environment. This is meant
```
jaycarlton: Sorry for the late review, but this is easier than trying to cross-ref stuff from Slack.

First question I had was, why does this need to be a Ruby script? If it's just doing file operations, there's support for that natively in gradle. Also, it looks like Gradle supports Kotlin now in addition to Groovy, so maybe there's an opportunity for tighter integration. It doesn't look like this script will be doing much outside of gradle tasks.

I realize there are future plans here, and if Ruby has a better Yaml parser than you get out of the boxk, then that makes sense.

Also, anecdotally, we're seeing problems with gradle tasks that call ruby. With this commit I can't run `gradlew appengineRun` anymore on my Mac.

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# Future work: it will likely be desirable to support extension rather than
+# wholesale file replacement, e.g. to alter just the schedule or exclude a
+# particular cron for a given environment.
+require 'fileutils'
+
+env = ARGV[0]
+
+cron_dir = "src/main/webapp/WEB-INF"
+to_cron_path = ->(suffix) { "#{cron_dir}/cron_#{suffix}.yaml" }
```
jaycarlton: I'm curious about this convention of declaring "named lambdas" and then invoking them with `.call`. I don't feel like it's a readability win, and if we need it for any kind of functional magic (passing them to algorithms and whatnot), we don't need it yet. 

Same question in Java code btw: now that we have method references, it doesn't seem like it should be necessary (unless you need captures).

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+
+path = to_cron_path.call("default")
+if File.file?(to_cron_path.call(env))
```
jaycarlton: Does Ruby have any concept similar to `Optional.orElse()` that would work here?

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+  path = to_cron_path.call(env)
+end
+
+FileUtils.cp(path, "#{cron_dir}/cron.yaml")
```
jaycarlton: I'd make another method that computes the destination path. It's also good to use `File.join` for clarity instead of string manipulation.

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+# wholesale file replacement, e.g. to alter just the schedule or exclude a
+# particular cron for a given environment.
+require 'fileutils'
+
+env = ARGV[0]
+
+cron_dir = "src/main/webapp/WEB-INF"
+to_cron_path = ->(suffix) { "#{cron_dir}/cron_#{suffix}.yaml" }
+
+path = to_cron_path.call("default")
```
jaycarlton: A more descriptive name would help here, since we're dealing with more than one path. LIke `app_engine_cron_source_path`.

api/libproject/generate_appengine_cron_yaml.rb
```diff
@@ -0,0 +1,20 @@
...
+# For now, this picks cron_{env}.yaml if it exists, cron_default.yaml otherwise.
+#
+# Future work: it will likely be desirable to support extension rather than
+# wholesale file replacement, e.g. to alter just the schedule or exclude a
+# particular cron for a given environment.
+require 'fileutils'
+
+env = ARGV[0]
+
+cron_dir = "src/main/webapp/WEB-INF"
```
jaycarlton: Can you make this a constant like `CRON_DIR`?

api/libproject/generate_appengine_cron_yaml.rb
calbach: The rationale for using Ruby here is for consistency. There's a highly similar script in this same directory called for a similar purpose by Gradle. Moving both of them to Groovy might be fine, I haven't evaluated it so I don't know whether that would be worth the effort. I think in the future we'll want to do more munging of the yaml in this script, so we should ensure that this process is straightforward in whichever scripting language. 

My fault, sorry for the breakage. Per slack, please run `source api/db/load_vars.sh` to fix.

api/libproject/generate_appengine_cron_yaml.rb
calbach: I don't see how that pattern is relevant here, we're checking whether a file exists; we could use a ternary or something here but I don't think that makes this any more readable.

api/libproject/generate_appengine_cron_yaml.rb
calbach: Done.

api/libproject/generate_appengine_cron_yaml.rb
calbach: I'd thought it was needed for the closure, but apparently not.

jaycarlton: Oh cool. I last used Ruby ~10 years ago so I'm interested in learning more.

api/libproject/generate_appengine_cron_yaml.rb
jaycarlton: Right. Unless you wanted your path conversion to do the file check and return null if it didn't exist.

PR #2700, opened 2019-10-18
[risk=no][no ticket]When we can't retrieve a project from Firecloud/Terra, mark it as Error

jmthibault79:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -111,11 +111,13 @@ public class BillingProjectBufferService {
                 + entry.getFireCloudProjectName()
                 + ". Project not found.",
             e);
+        entry.setStatusEnum(ERROR, this::getCurrentTimestamp);
       } catch (WorkbenchException e) {
         log.log(
             Level.WARNING,
             "Get BillingProjectStatus call failed for " + entry.getFireCloudProjectName(),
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -111,11 +111,13 @@ public void syncBillingProjectStatus() {
                 + entry.getFireCloudProjectName()
                 + ". Project not found.",
             e);
+        entry.setStatusEnum(ERROR, this::getCurrentTimestamp);
```
calbach: The previous behavior was intentional, see https://pmi-engteam.slack.com/archives/CHRN2R51N/p1570652916047400 ; if a project fails repeatedly this way, it will eventually transition into the error state after an hour by the other cron process (pending CREATING state for too long).

What issue are you trying to solve?

jmthibault79: This is not true in Perf - see the billing buffer there

calbach: Ah.. cleanBillingBuffer should probably run much more frequently, it's set to run every 24h currently

calbach:
I merged https://github.com/all-of-us/workbench/pull/2706 , I think this PR can now be dropped.

jmthibault79:
Agreed

PR #2704, opened 2019-10-18
[RW-3355][risk=no] Use Docker-sync to improve Docker runtime on OSX

calbach:
See Jira for background & notes on potential speedup timings: https://precisionmedicineinitiative.atlassian.net/browse/RW-3355

To test:

```
# First, update your Docker OSX settings to use all CPUs and ~70-80% of available memory & swap

# Install docker-sync
$ gem install --user-install docker-sync
# Not shown: add ruby gems to PATH, instructions here https://docker-sync.readthedocs.io/en/latest/getting-started/installation.html

# Checkout branch
$ git checkout master && git pull
$ git checkout ch/docker-sync

# Docker-clean & cold dev-up
$ cd api
# Yes, a full docker-clean wipe is necessary – sorry!
$ ./project.rb docker-clean
# Docker-sync will be started & stopped as part of the dev script
$ ./project.rb dev-up

# Warm dev-up for realistic timings
$ ./project.rb dev-up
```


gjuggler:
I tried this out on my Google-corp macbook. I wasn't quite able to get to a successful dev-up, but it may be something unique / incorrect set up on my machine. A few notes:

1) I had to install Docker-sync in "user mode" on my corp laptop (https://docker-sync.readthedocs.io/en/latest/getting-started/installation.html#installation-as-non-root), otherwise I'm pretty sure I was running into some issues.
2) It wasn't 100% clear to me what should be happening when I ran "docker-sync start". I would see a message like

Screen Shot 2019-10-20 at 11 02 27 AM

and then it would hang there. I never saw a message like in the PR description.

3) When running "project.rb dev-up" I kept running into Gradle errors. The top-level error message would look like this:

Screen Shot 2019-10-20 at 11 05 01 AM

and I managed to change some flags to get a stack trace revealing a file permissioning issue:

Screen Shot 2019-10-20 at 11 06 12 AM

I didn't get any further than that, though. I confirmed that a fresh docker-clean & dev-up would run fine on master for me, so I don't *think* my machine is in a bad state, though I'm not at all certain.

I'm mostly posting these here as corroboration, in case another Macbook user encounters the same error messages.

calbach:
@gjuggler thanks! If you have time in the background, could you let docker-sync hang for at least an hour or so, see if it finishes (with a `time` as well ideally)? Apparently for large codebases setting up the initial index may take a while: https://docker-sync.readthedocs.io/en/latest/advanced/sync-strategies.html#unison-osx-windows-linux .

gjuggler:
> @gjuggler thanks! If you have time in the background, could you let docker-sync hang for at least an hour or so, see if it finishes (with a `time` as well ideally)? Apparently for large codebases setting up the initial index may take a while: https://docker-sync.readthedocs.io/en/latest/advanced/sync-strategies.html#unison-osx-windows-linux .

Sorry for the delay – I did let this run, and it did eventually complete at some point. I still wasn't able to get dev-up to complete successfully after that, though. So I'm not exactly sure what was causing my errors above...

als364:
belatedly: I can't even get docker-sync to start, is there another step I need or a dependency I'm missing?
```aschwart@failboat:~/Documents/dev/workbench/api$ docker-sync start
          ok  Starting native_osx for sync default-sync
/Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_strategy/native_osx.rb:41:in `realpath': No such file or directory @ realpath_rec - /Users/aschwart/Documents/dev/workbench/api/docker-sync-data (Errno::ENOENT)
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_strategy/native_osx.rb:41:in `start_container'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_process.rb:105:in `start_container'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_manager.rb:103:in `block in start_container'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_manager.rb:102:in `each'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/lib/docker-sync/sync_manager.rb:102:in `start_container'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/tasks/sync/sync.thor:163:in `daemonize'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/tasks/sync/sync.thor:47:in `start'
	from /Library/Ruby/Gems/2.3.0/gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
	from /Library/Ruby/Gems/2.3.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
	from /Library/Ruby/Gems/2.3.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	from /Library/Ruby/Gems/2.3.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
	from /Library/Ruby/Gems/2.3.0/gems/docker-sync-0.5.11/bin/docker-sync:14:in `<top (required)>'
	from /usr/local/bin/docker-sync:22:in `load'
	from /usr/local/bin/docker-sync:22:in `<main>'```

**Comment by calbach** [comment]
Sorry for the delay.

@als364 the issue you hit was just me accidentally leaving some unused stuff in the docker-sync config. That issue should now be fixed if you pull again.

@gjuggler is there some additional setup needed for docker volumes on Mac? The new volume is called "src-sync": https://docs.docker.com/docker-for-mac/troubleshoot/#volume-mounting-requires-file-sharing-for-any-project-directories-outside-of-users

If that doesn't do anything, would be interested in the following output:

**Comment by calbach** [comment]
Also note:
If you are trying to pick up / test any changes in the `docker-sync.yml`, you should do the following:
```
docker-sync stop
docker-sync clean
docker-sync start
```

File: api/docker-compose.yaml [hunk 1]
```diff
@@ -36,11 +36,11 @@ services:
...
     build:
       context: ./src/dev/server
-    user: ${UID}
+    user: ${USER_ID}
     working_dir: /w/api
     volumes:
-      - ..:/w:cached
-      - gradle-cache:/.gradle
+      - src-sync:/w:nocopy
+      - gradle-cache:/.gradle:cached
```
jmthibault79: What does this do, and why should we keep half of these lines as `gradle-cache:/.gradle` ?

README.md
```diff
@@ -21,6 +21,8 @@ For local development, also install:
   * [yarn](https://yarnpkg.com/lang/en/docs/install/#mac-stable)
   * [Node.js](https://nodejs.org/en/)
   * [Java 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+  * [Docker sync](https://docker-sync.io)
+    * > gem install docker-sync
```
jmthibault79: Add `--user-install` and the add-to-path link: https://docker-sync.readthedocs.io/en/latest/getting-started/installation.html

README.md
```diff
@@ -21,6 +21,8 @@ For local development, also install:
   * [yarn](https://yarnpkg.com/lang/en/docs/install/#mac-stable)
   * [Node.js](https://nodejs.org/en/)
   * [Java 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+  * [Docker sync](https://docker-sync.io)
```
jmthibault79: Should we update the Docker config suggestion on line 11?

calbach:
@jmthibault79 what do the docker config changes refer to "After Docker config changes ONLY:" - and why was this slower than your original setup? Am I understanding that correctly?

jmthibault79:
Correct: the Docker config testing suggestion here slowed down my dev-up:
`First, update your Docker OSX settings to use all CPUs and ~70-80% of available memory & swap`

My previous (faster) settings were much smaller: something like 1/2 CPUs and 1/4 memory/swap.  I can attempt to find a pattern here.

jmthibault79:
Also, the first time running `dev-up` after docker clean fails in this way ([full gist](https://gist.github.com/jmthibault79/ce3509dea659cce6795d06a025e23346)):
```
Step 23/23 : ENTRYPOINT ["with-uid.sh"]
 ---> Using cache
 ---> c650a8a82466
Successfully built c650a8a82466
Successfully tagged api_db-scripts:latest
WARNING: Image for service db-scripts was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
wait-for-it: waiting 15 seconds for db:3306
wait-for-it: timeout occurred after waiting 15 seconds for db:3306
+ IFS='
        '
+ activity=-PrunList=main
+ '[' -z ']'
+ context=
+ CREATE_DB_FILE=/tmp/create_db.sql
+ trap finish EXIT
+ envsubst
++ dirname /w/api/db/run-migrations.sh
+ echo 'Creating database if it does not exist...'
Creating database if it does not exist...
+ mysql -h db --port 3306 -u root -proot-notasecret
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 2003 (HY000): Can't connect to MySQL server on 'db' (111)
+ finish
+ rm -f /tmp/create_db.sql
+ docker-compose down
Stopping api_db_1 ... done
Removing api_db-scripts_run_69f2c701e220 ... done
Removing api_db_1                        ... done
Removing network api_default
+ docker-sync stop
```

The second run continues past this step and eventually succeeds:
```
Database init & migrations...
+ docker-compose run db-scripts ./run-migrations.sh main
Starting api_db_1 ... done
wait-for-it: waiting 15 seconds for db:3306
wait-for-it: db:3306 is available after 0 seconds
+ IFS='
        '
+ activity=-PrunList=main
+ '[' -z ']'
+ context=
+ CREATE_DB_FILE=/tmp/create_db.sql
+ trap finish EXIT
+ envsubst
++ dirname /w/api/db/run-migrations.sh
+ echo 'Creating database if it does not exist...'
Creating database if it does not exist...
+ mysql -h db --port 3306 -u root -proot-notasecret
mysql: [Warning] Using a password on the command line interface can be insecure.
+ echo 'Upgrading database...'
Upgrading database...
++ dirname /w/api/db/run-migrations.sh
+ cd /w/api/db
+ ../gradlew update -PrunList=main
```

Not sure if this is a pre-existing issue.

jmthibault79:
Never mind, fixed by increasing the timeout: https://github.com/all-of-us/workbench/pull/2871
```
wait-for-it: waiting 30 seconds for db:3306
wait-for-it: db:3306 is available after 14 seconds
```

jmthibault79:
[Docker config knob turning experiments](https://gist.github.com/jmthibault79/08b1fadef418c0b151f0bbe26bb901b8)

I can't seem to find any particular config setting that works better than any other.  OTOH it was not a major burden for me to run dev-up many times, so that's a win in itself.

api/docker-compose.yaml
gjuggler: Thanks for checking on this one – I did a bit more benchmarking and it has no impact on startup time. Removed.

README.md
gjuggler: Done.

README.md
gjuggler: Done.

gjuggler:
OK, I think this one is (finally!) ready for review. Adding a few folks who I think are working tomorrow: @calbach @jaycarlton @freemabd

My last lingering concern was that the new docker-sync requirement might break our CI builds or deployments. Aside from the empirical evidence that CircleCI is green, I've convinced myself that the codepath that triggers "docker-compose up api" is only ever used during local development.

The closest CircleCI gets to this is during the "api-local-test" workflow task, which uses a different set of commands and executes from within the CircleCI Docker image, which presumably maps a clean Git repo to the 'workbench' root dir. (@calbach please correct me if I'm grossly wrong on any of these details.)

I'd love for reviews with an eye especially towards any places where I'm adding entropy to our local dev setup rather than removing it. Missing comments, opportunity for a short separate .md doc, etc.

At the earliest, this won't go in before Thursday when most of the team is back. I'll probably aim for pushing Friday morning, and will be available that day to help debug, answer questions, or rollback.

gjuggler:
@calbach technically you can't review your own PR, but consider this an informal request for re-review :)

api/.env
```diff
@@ -0,0 +1,6 @@
+# These environment variables are referneced from within api/docker-compose.yaml to
```
calbach: nit: referenced

api/.env
```diff
@@ -0,0 +1,6 @@
...
+# set the user ID and group for files created within the container. A non-empty user
+# and group ID is necessary when running with docker-sync, since the sync tool is actively
+# trying to access files created by the Docker user.
+USER_ID=1001
```
calbach: Hm... are these right? The previous approach would use your current `$UID` as the user. I don't think 1001 or 82 refers to anything specific I could fine in the code - is this your own UID from your machine?

jmthibault79:
The currently branch initially failed for me, but worked on re-run: `/usr/bin/wait-for-it: line 174: /w/api/libproject/load_local_data_and_configs.sh: No such file or directory`

jaycarlton:
@gjuggler If you could talk about all this at an eng sync, I think that would be helpful.

api/.env
gjuggler: Sorry, I forgot a key bit of documentation here. This file was a workaround to avoid the "permission denied" errors that I was encountering when trying out the earliest version of this branch. Docker-sync needs to use the same UID as the Docker container, which doesn't happen if we leave the 'sync_userid' field blank in docker-sync.yml. This workaround leveraged the ability to have docker-sync and docker-compose share the same .env file (https://docker-sync.readthedocs.io/en/latest/getting-started/configuration.html#environment-variables-support), so they can be guaranteed to use the same user_id this way. The actual numbers were random... I can't remember how I landed on those, I think I saw them commonly in examples in the wild.

This morning I did a bit more digging into docker-sync options, and there is an option sync_userid:'from_host' that causes docker-sync to attach to the current user's userid. I tested this end-to-end and it works – so I'm inclined to switch to that simpler setup, as long as it also works on Linux.

Question for @calbach : is it beneficial / useful / necessary to have docker-compose run as the current user's UID? I see we have some hooks in the code to make this work (I think they trace back to https://github.com/all-of-us/utils/blob/6a6dc7b4bb07c4661b4f6721c1c70cf24ed82042/workbench.rb#L5-L5 where the UID env variable is set, there's also the with-uid.sh script which traces back to https://github.com/all-of-us/workbench/commit/74daf87dd99bea57c568a8a43a892549ce701f67#diff-b3b218df651d0cc4ce869e2a9c36fff5), but I don't know enough historically (or haven't found any docs indicating what strategy we're following) to understand whether or why this might be important.

jaycarlton: So there's the "user of docker" and the "docker user". Are these different entities?

api/.env
jaycarlton: We could use `.env` files in a number of places, for Cloud Functions, unit testing, etc. Does this one have to live at the top level like this, or could it go in a subdirectory just for docker?

api/docker-compose.yaml
```diff
@@ -36,10 +36,10 @@ services:
       - elastic
     build:
       context: ./src/dev/server
-    user: ${UID}
+    user: ${USER_ID}
```
jaycarlton: Is there a way to reuse some of this config info so you don't have to re-specify it each time?

gjuggler: Maybe it could have, but repetition in configuration isn't too detrimental (and can sometimes be beneficial for operations, requiring less head-scratching to understand the state of things).

Anyway, this may be a moot point as I think I found an alternative which didn't require any changes here. (See my longer reply to CH's comment.)

api/.env
gjuggler: [Caveat: see my reply to CH's comment, and a soon-to-be-pushed change which removes this file altogether.]

Re: user of docker vs. docker user. I think they can be – but we have code elsewhere which tries to align the Docker user ID with the user ID who called the dev script.

Re: .env file location, it looks like that's possible by specifying an 'env_file' location in the docker-compose.yaml (https://docs.docker.com/compose/environment-variables/#the-env-file).

api/.env
calbach: @gjuggler from_host sounds ideal; I can test this out later today. The decision to use the host UID predates me but I can more or less surmise why it was done from some comments and past experience and in my assessment it's a fairly reasonable approach. There are many articles I've seen talking about this issue and no single best practice I've found for handling it.

The issue being mitigated here is that when your docker container creates files within a volume shared with the host filesystem, the default owner of those files will be the docker user. This is a pretty big pain if that's a different UID from your current user. For example, if you run codegen within the container, you'll wind up with a bunch of generated files that the host user cannot remove (without sudo). Often you'll need to run sudo rm -rf ... to get rid of these.

@jaycarlton a docker container defines its own user ID space. This is rarely the same as the host system you're running docker on, as a docker image is defined once and different host systems will have different user spaces. When executing a command within docker, you will specify a user id to act as, or inherit the default for that image.

gjuggler:
@freemabd I pulled you into this review as I'm especially curious whether this branch breaks any of your IntelliJ-related development flow. In particular, by excluding many of our generated output directories from the docker-sync process, this means that running a project.rb dev-up will *not* create those files in your host filesystem. All source files are still synced between the host and container.

You can confirm this w/ this branch, for example, by running a './gradlew clean' to remove all api/src/generated files, then run a './project.rb dev-up' to kick off a server in the container. No new api/src/generated files will be generated on your host filesystem.

This may be a good thing or a bad thing, depending how you look at it, but it would definitely be a change in behavior.

gjuggler:
> @gjuggler I will confirm this branch works for me tonight. I'm trying to catch up with everything after being out for a week.

Thanks – to be clear, there's no huge rush on this one. I'd like for us to realize benefits sooner than later, but I'd rather take our time to flush out any deal-breakers before merge rather than rolling back after the fact.

gjuggler:
> @gjuggler @calbach So installing docker-sync and dev-up seem to be working fine. Intellij seems to be happy with everything. But I have noticed that most `./project.rb` commands just hang when running.

Thanks Brian – I realized this one last night too and have been thinking through solutions a bit. I suspect we'll have to do something like make an easy way to add docker-sync start/stop hooks into each of these commands... unless there's a better way? (An alternative approach would be to have each dev be in control of their own docker-sync process, starting / stopping manually... but that feels less clean somehow.)

calbach:
Confirmed it's working and hot-swapping on my machine. Thanks!

jaycarlton:
I forgot to approve the design doc. I don't have any objections to the plan.

PR #2710, opened 2019-10-21
[RW-3519][risk=no] Prevent illegal characters in notebook name

andy7i:
Going straight to the source, Jupyter, I tried to rename the file. The characters not allowed by jupyter were:

```
:\/
```

Going back to AoU, I tested creating a notebook with the other "standard" special characters excluding the ones Jupyter explicitly disallows, and this also failed. The notebook name was:
```
`-=~!@#$%^&*()_+~!@#$%^&*()_+[] ;'"{}|,.<>
```

I checked in Saturn/Terra to see how it was handled and I found that it has a larger list of disallowed/illegal characters:
NAME CAN'T CONTAIN THESE CHARACTERS: @ # $ % * + = ? , [ ] : ; / \

Other notebook names tested include:

```
`¡™£¢∞§¶•ªº––≠«‘“πøˆ¨¥†®´∑œåß∂ƒ©˙∆˚¬…æ÷≥≤µ˜∫√ç≈Ω
```

```
!^&()_-{}|'".><`~
```

```
`⁄€‹›ﬁﬂ‡°·‚—±»’”∏Øˆ¨Áˇ‰´„ŒÅÍÎÏ˝ÓÔÒÚÆ¿˘¯Â˜ı◊Ç˛¸
```

// ^ these are all of the special characters that macs provide using the keyboard with shift or option (minus the ones specified by jupyter and saturn/terra)

As an aside, it seems like for other resource types such as cohorts and datasets, they are not bound by the same name restrictions. Names with special characters (excluding `/`) worked fine. Therefore, there are 2 versions of the name validator - one for notebooks and one for all other resource types.

Tested XSS? Renaming a Dataset to: ` <script>alert('test') ` did not do anything unexpected, although it did require a refresh to view the change. I guess that was unexpected. I think that was due to a very stale tab, since subsequent tests could not replicate the issue.

Issues:

Bug / missing feature (fixed, mostly (see Other Issue #4 below)): duplicate/existing name validation does not happen at all for notebooks. This was the case for new notebooks and renaming notebooks.

2+ areas where notebook name is validated. Unifying them (DRY) is key to prevent bugs similar to this ticket in the future. I believe this is done.

Open question on security: are we sanitizing inputs on the back end? We should disallow these characters in the API call as well. This may be a separate task since it may require a larger refactoring on the back end to sanitize all inputs. (Ticket TBD)

--------
Other Issues found:

1.
Character limit

I also found a character limit between 200-250 chars for notebooks:
It fails ungracefully, with an infinite spinner.

```
{"message": "Unexpected error while saving file: workspaces/duplicateoftest1231021/250mualjohjmgiihiwxcbeqyqngtgzdbjbwhjfvoovzszgxjrtkbhrzlogbjhcfjixdiozgardrvijoclklyfjpvkopvyfctfotiwdvqkgogxzyzhadavmbqorcdkzjyeykwypenuvywoijoseqctgckvgyybgwnyropyjtohtcejdpbmpeylmquetevzmzzdfmstuvgfhkmsmirmuwvduhwkbhauzphrdeveectguocrmrfxwbmskwnlr.ipynb [Errno 36] File name too long: '/home/jupyter-user/notebooks/workspaces/duplicateoftest1231021/250mualjohjmgiihiwxcbeqyqngtgzdbjbwhjfvoovzszgxjrtkbhrzlogbjhcfjixdiozgardrvijoclklyfjpvkopvyfctfotiwdvqkgogxzyzhadavmbqorcdkzjyeykwypenuvywoijoseqctgckvgyybgwnyropyjtohtcejdpbmpeylmquetevzmzzdfmstuvgfhkmsmirmuwvduhwkbhauzphrdeveectguocrmrfxwbmskwnlr.ipynb'", "reason": null}
```
We can add a length limit using `validate` (https://validatejs.org/#validators-length). I initially suspected that the length of the notebook file includes the name of the workspace, but it seems like that is not the case. A workspace name that was 53 characters longer was able to create a 200 character notebook successfully. I can add it in this PR or in a separate PR. Your choice. I lean toward a separate PR so that we can make the limit for the other resource types as well.

1b.
 Speaking of, we don't give a helpful error message for workspace names that are too long. The UI says:

```
Error:
Could not create workspace.
```

though the API response says:

```
{"message":"Workspace name must be 80 characters or less","statusCode":400,"errorClassName":"org.pmiops.workbench.exceptions.BadRequestException","errorCode":null}
```

We should try to bubble that message up or validate in the UI. Now where have I seen a validate function before...

2.
Copying resources (notebooks, cohorts, etc.) does not validate name.

```
Copy to Workspace
Successfully copied Duplicate of += to testlocalcreation10102019 . Do you want to view the copied Notebook?
```

3.
`Esc`aping out of some modals does not work correctly. For example, the rename modals on the data tab.

```
ERROR TypeError: Cannot read property 'setPreviousFocus' of undefined
```

4.
Duplicate/existing notebook name is checked against notebooks outside of the current WS. Error message in the UI says: `NAME ALREADY EXISTS`.


ui/src/app/components/rename-modal.tsx
```diff
@@ -24,9 +25,45 @@ const styles = reactStyles({
...
+ *
+ * Other AoU resource types do not have the same name restriction and only block slashes.
+ */
+export const nameValidationFormat = (existingNames, type) =>
+  type === ResourceType.NOTEBOOK ?
+    ({
+      presence: {allowEmpty: false},
+      format: {
+        pattern: /^[^#[\]*?:;@$%+=\\,/]*$/,
+        message: 'can\'t contain these characters: \r \r  @ # $ % * + = ? , [ ] : ; / \\ '
```
als364: Did you mean `\r \n`?

als364: Also, nit: could you please put these in the same order as is in the regex if possible?

ui/src/app/pages/analysis/new-notebook-modal.tsx
```diff
@@ -6,16 +6,17 @@ import {styles as headerStyles} from 'app/components/headers';
...
+import {nameValidationFormat} from 'app/components/rename-modal';
 import {userMetricsApi} from 'app/services/swagger-fetch-clients';
 import {summarizeErrors} from 'app/utils';
 import {navigate} from 'app/utils/navigation';
 import {Kernels} from 'app/utils/notebook-kernels';
+import {ResourceType} from 'app/utils/resourceActions';
 
-
-import {FileDetail, Workspace} from 'generated/fetch';
+import {Workspace} from 'generated/fetch';
```
als364: Huh, where was this coming from before? I would think that the `workspace: Workspace` on line 19 wouldn't work without this

ui/src/app/pages/analysis/notebook-list.tsx
```diff
@@ -58,8 +59,10 @@ export const NotebookList = withCurrentWorkspace()(class extends React.Component
     try {
       const {workspace: {namespace, id}} = this.props;
       this.setState({loading: true});
-      const notebooks = await workspacesApi().getNoteBookList(namespace, id);
-      this.setState({notebooks});
+      const notebookList = await workspacesApi().getNoteBookList(namespace, id);
+      this.setState({notebookList});
+      const notebookNameList = notebookList.map(fd => fd.name.slice(0, -6)); // remove '.ipynb'
```
als364: Is the suffix the same for R notebooks?

ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -160,7 +161,7 @@ export const NotebookResourceCard = fp.flow(
         fromWorkspaceNamespace={this.props.resource.workspaceNamespace}
         fromWorkspaceName={this.props.resource.workspaceFirecloudName}
         fromResourceName={dropNotebookFileSuffix(this.props.resource.notebook.name)}
-        resourceType={ResourceType.NOTEBOOK}
+        resourceType={this.resourceType}
```
als364: Do you happen to know why this was hard-coded before?

ui/src/app/components/rename-modal.tsx
andy7i: Good question. I copied that over from Saturn code but I'll remove it.

ui/src/app/pages/analysis/new-notebook-modal.tsx
andy7i: i think it previously existed on line 15.

`import {FileDetail, Workspace} from 'generated/fetch';	`

ui/src/app/pages/analysis/notebook-resource-card.tsx
andy7i: I think it was hardcoded since it's specifically a `notebook-resource-card` and so the resource type in this case is always Notebook.

ui/src/app/pages/analysis/notebook-list.tsx
andy7i: good point! I just checked and it is the same suffix.

r-notebook-name.ipynb

ui/src/app/pages/analysis/new-notebook-modal.tsx
als364: Aha! Thank you!

ui/src/app/pages/analysis/notebook-resource-card.tsx
als364: It's currently hardcoded to `Notebook`, but `ResourceType.NOTEBOOK` is set to `notebook`. Do you know if there is any behavior that depends on this?

andy7i: as of this PR (merged today) those should all be referring to ResourceType.NOTEBOOK which has the value `Notebook` https://github.com/all-of-us/workbench/pull/2671/files so we should be covered in the behavior here.

als364: Sweet, I think my `master` is just behind then

ui/src/app/pages/analysis/notebook-list.tsx
dmohs: Tip: You can get the same documentation without the comment like so:
```javascript
fd.name.slice(0, -('.ipynb'.length))
```

PR #2712, opened 2019-10-21
[RW-3722][risk=no]

jaycarlton:
Added configuration for log name. This is necessary especially in the local environment in order to prevent logged events from mixing with those of the test environment (as they're in the same GCP project and thus Logging namespace).

Add collaborate actions for workspaces, which must capture access information all users. This requires one action with multiple events (with separate targets).

Workspace clone/duplicate events are likewise composed of multiple events in a single action. In this case, we also cheat and use different action types for the source and destination workspaces.


api/src/main/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceImpl.java
```diff
@@ -73,23 +81,135 @@ private void logAndSwallow(RuntimeException e) {
...
   }
+
+  @Override
+  public void fireDuplicateAction(
+      org.pmiops.workbench.db.model.Workspace sourceWorkspaceDbModel,
+      org.pmiops.workbench.db.model.Workspace destinationWorkspaceDbModel) {
+    try {
+      // We represent the duplication as a single action with events with different
+      // ActionTypes: DUPLICATE_FROM and DUPLICATE_TO. The latter action generates many events
+      // (similar to how CREATE is handled) for all the properties. I'm not (currently) filtering
```
jmthibault79: So... even though it's a single workspace, each auditable property set on the new workspace is its own event?

api/src/main/java/org/pmiops/workbench/audit/targetproperties/WorkspaceTargetProperty.java
```diff
@@ -50,4 +54,25 @@ public String extract(Workspace workspace) {
...
+
+  public static Map<String, PreviousNewValuePair> getChangedValuesByName(
+      Workspace previousWorkspace, Workspace newWorkspace) {
+    ImmutableMap.Builder<String, PreviousNewValuePair> resultBuilder = new Builder<>();
+    for (WorkspaceTargetProperty property : WorkspaceTargetProperty.values()) {
+      final String previousValue = property.extract(previousWorkspace);
+      final String newValue = property.extract(newWorkspace);
+      if (previousValue != null || newValue != null) {
+        if (previousValue != null && !previousValue.equals(newValue)
+            || !newValue.equals(previousValue)) {
```
jmthibault79: why two equals checks?

PR #2714, opened 2019-10-21
[RW-3736][risk=low] Move cluster defaults to config, increase size

calbach:
- Keep `n1-standard-2` for perf to save some costs.
- Increase all other environments to `n1-standard-4`, from n1-standard-2
- Increase all environments boot disk size from `40GB` -> `50GB`

### DIFF
--- a/api/config/config_prod.json
+++ b/api/config/config_prod.json
@@ -3,7 +3,7 @@
     "baseUrl": "https:\/\/api.firecloud.org",
     "clusterMaxAgeDays": 14,
     "clusterIdleMaxAgeDays": 7,
-    "clusterDefaultMachineType": "n1-standard-2",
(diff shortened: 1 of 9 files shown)


api/config/config_prod.json
```diff
@@ -3,6 +3,8 @@
     "baseUrl": "https:\/\/api.firecloud.org",
     "clusterMaxAgeDays": 14,
     "clusterIdleMaxAgeDays": 7,
+    "clusterDefaultMachineType": "n1-standard-2",
```
jmthibault79: swapped Prod and Perf

calbach: :man_facepalming:  Great catch.

That would have rendered this PR pretty useless.

PR #2715, opened 2019-10-21
[risk=low][RW-3594]Store free tier billing costs in the DB

jmthibault79:

### DIFF
--- a/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
+++ b/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
@@ -11,10 +11,10 @@
         <constraints primaryKey="true" nullable="false"/>
       </column>
       <column name="user_id" type="bigint">
-        <constraints nullable="false"/>
+        <constraints nullable="false" foreignKeyName="fk_free_tier_usage_user_id" references="user(user_id)"/>
       </column>
       <column name="workspace_id" type="bigint">
-        <constraints nullable="false"/>
+        <constraints nullable="false" foreignKeyName="fk_free_tier_usage_workspace_id" references="workspace(workspace_id)"/>
       </column>
       <column name="cost" type="double">
         <constraints nullable="false"/>
@@ -23,18 +23,6 @@
         <constraints nullable="false"/>
       </column>
     </createTable>
-
-    <addForeignKeyConstraint baseColumnNames="user_id"
-      baseTableName="workspace_free_tier_usage"
-      constraintName="fk_free_tier_usage_user_id"
-      referencedColumnNames="user_id"
-      referencedTableName="user" onDelete="CASCADE"/>
-
-    <addForeignKeyConstraint baseColumnNames="workspace_id"
-      baseTableName="workspace_free_tier_usage"
-      constraintName="fk_free_tier_usage_workspace_id"
-      referencedColumnNames="workspace_id"
-      referencedTableName="workspace" onDelete="CASCADE"/>
 
   </changeSet>
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
+++ b/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
@@ -7,7 +7,7 @@
   <changeSet author="thibault" id="changelog-105-workspace-free-tier-usage">
 
     <createTable tableName="workspace_free_tier_usage">
-      <column name="id" type="bigint" autoIncrement="true">
+      <column name="workspace_free_tier_usage_id" type="bigint" autoIncrement="true">
         <constraints primaryKey="true" nullable="false"/>
       </column>
       <column name="user_id" type="bigint">
--- a/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
+++ b/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
@@ -26,13 +26,13 @@
 
     <addForeignKeyConstraint baseColumnNames="user_id"
       baseTableName="workspace_free_tier_usage"
-      constraintName="fk_user_id"
+      constraintName="fk_free_tier_usage_user_id"
       referencedColumnNames="user_id"
       referencedTableName="user" onDelete="CASCADE"/>
 
     <addForeignKeyConstraint baseColumnNames="workspace_id"
       baseTableName="workspace_free_tier_usage"
-      constraintName="fk_workspace_id"
+      constraintName="fk_free_tier_usage_workspace_id"
       referencedColumnNames="workspace_id"
       referencedTableName="workspace" onDelete="CASCADE"/>
 
--- a/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
+++ b/api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="changelog-105-workspace-free-tier-usage">
+
+    <createTable tableName="workspace_free_tier_usage">
+      <column name="id" type="bigint" autoIncrement="true">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="user_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
+      <column name="workspace_id" type="bigint">
+        <constraints nullable="false"/>
+      </column>
+      <column name="cost" type="double">
+        <constraints nullable="false"/>
+      </column>
+      <column name="last_update_time" type="datetime">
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
```diff
@@ -20,49 +26,77 @@
...
-    QueryJobConfiguration queryConfig =
+    final Map<Workspace, Double> workspaceCosts = getFreeTierWorkspaceCosts();
+
+    final Set<User> expiredCreditsUsers =
+        workspaceCosts.entrySet().stream()
+            .collect(
+                Collectors.groupingBy(
+                    e -> e.getKey().getCreator(), Collectors.summingDouble(Entry::getValue)))
+            .entrySet()
+            .stream()
```
jmthibault79: I'm not super happy about `collect()` and then `stream()` again, but I didn't see an obvious better solution.

blrubenstein: Could you do it with a:
```
.map(entry -> entry.setKey(entry.getKey().getCreator()); entry.setValue(_something_);
```

blrubenstein: Recognizing that it isn't that simple, but maybe calling a constructor or a duplicator for an entry in there.

api/src/test/java/org/pmiops/workbench/billing/BillingAlertsServiceTest.java
```diff
@@ -96,75 +122,246 @@ public void checkFreeTierBillingUsage_disabledUserNotIgnored() {
...
+    assertThat(dbEntry.getUserId()).isEqualTo(user.getUserId());
+    assertThat(dbEntry.getWorkspaceId()).isEqualTo(workspace.getWorkspaceId());
+    assertThat(dbEntry.getCost())
+        .isCloseTo(SINGLE_WORKSPACE_TEST_COST * 2, withinPercentage(0.001));
+
+    Timestamp t1 = dbEntry.getLastUpdateTime();
+    assertThat(t1).isAfter(t0);
+  }
+
+  private TableResult mockTableResult(Map<String, Double> costMap) {
```
als364: It took me way too long to figure out that this is mocking a BigQuery thing instead of a SQL thing. This is certainly because I'm not familiar with talking to BigQuery in Java, but it would still be helpful for me if this was called something like `mockBigQueryTableResult` instead.

api/src/main/java/org/pmiops/workbench/billing/BillingAlertsService.java
jmthibault79: @als364 I need both `workspaceCosts` and `expiredCreditsUsers`.  Since the former contains everything needed for the latter, my preference is to get both of these via a single BigQuery query and this transformation.

jmthibault79: Here's a StackOverflow with ideas but none of them *quite* fit so I'm inclined to keep this version: https://stackoverflow.com/questions/44101297/java-stream-filter-sum-of-values

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
```diff
@@ -0,0 +1,77 @@
...
+  public long getId() {
+    return id;
+  }
+
+  public void setId(long id) {
+    this.id = id;
+  }
+
+  @Column(name = "user_id")
+  public long getUserId() {
```
ericsong: Can we reference the models here instead of the ID?

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
```diff
@@ -0,0 +1,77 @@
...
+    return userId;
+  }
+
+  public void setUserId(Long userId) {
+    this.userId = userId;
+  }
+
+  @Column(name = "workspace_id")
+  public long getWorkspaceId() {
```
ericsong: re: above

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
```diff
@@ -0,0 +1,77 @@
...
+  public void setCost(double cost) {
+    this.cost = cost;
+  }
+
+  @Column(name = "last_update_time")
+  public Timestamp getLastUpdateTime() {
+    return lastUpdateTime;
+  }
+
+  public void setLastUpdateTime(Timestamp lastUpdateTime) {
```
ericsong: if we are cool with setting this from `setCost`, we should be able to make this private

api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="changelog-105-workspace-free-tier-usage">
+
+    <createTable tableName="workspace_free_tier_usage">
+      <column name="id" type="bigint" autoIncrement="true">
```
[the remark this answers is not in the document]
als364: Is that the main reason we do this? I've noticed that JPA/Hibernate Does Not Like That At All, to an extent that some of its built in features do not work (`@OneToOne` / `@ManyToOne` type annotations for instance) if you break the convention that the pk of a table is called ID.

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
als364: How come you want it to be private?

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceFreeTierUsageDao.java
```diff
@@ -0,0 +1,22 @@
...
+
+  WorkspaceFreeTierUsage findOneByWorkspaceId(long workspaceId);
+
+  default void updateCost(Workspace workspace, double cost) {
+    WorkspaceFreeTierUsage usage = findOneByWorkspaceId(workspace.getWorkspaceId());
+    if (usage == null) {
+      usage = new WorkspaceFreeTierUsage(workspace);
+    }
+    usage.setCost(cost);
+    usage.setLastUpdateTime();
```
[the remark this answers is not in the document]
jmthibault79: sure

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
```diff
@@ -0,0 +1,77 @@
...
+  private long id;
+  private long userId;
+  private long workspaceId;
+  private double cost;
+  private Timestamp lastUpdateTime;
+
+  public WorkspaceFreeTierUsage(Workspace workspace) {
+    this.userId = workspace.getCreator().getUserId();
+    this.workspaceId = workspace.getWorkspaceId();
+    setLastUpdateTime();
```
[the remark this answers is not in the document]
jmthibault79: ok

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
jmthibault79: I like that idea - that will allow us to control it more tightly

api/db/changelog/db.changelog-106-workspace-billing-account-fields.xml
```diff
@@ -0,0 +1,17 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="changelog-106-billing-account-fields">
+    <addColumn tableName="workspace">
+      <column name="billing_status" type="tinyint" defaultValueNumeric="0">
+        <constraints nullable="false"/>
+      </column>
+      <column name="billing_account_type" type="tinyint" defaultValueNumeric="0">
+        <constraints nullable="false"/>
```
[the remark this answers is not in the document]
jmthibault79: We don't have a concept of non-free-tier yet, so all will start out as FREE_TIER.
The cron job will ultimately determine whether a workspace is ACTIVE/INACTIVE so it doesn't really matter which is the default here.

api/db/changelog/db.changelog-105-workspace-free-tier-usage.xml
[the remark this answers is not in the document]
jmthibault79: I made the change and it appears to be working.

api/src/main/java/org/pmiops/workbench/db/model/WorkspaceFreeTierUsage.java
ericsong: discourages usage. No need to let people set the last update time and possibly introduce a bug if it should only be updated with the cost field.

PR #2720, opened 2019-10-22
[RW-2818][risk=no] Delay workspace create/clone until ACLs are synced

calbach:
- Due to a Sam consistency issue, sometimes newly created workspaces are not accessible immediately after creation. Anecdotally, this seems to subside within a few seconds.
- If this occurs, poll for up to a minute until access propagates
- If access never propagates, show a modal and allow the user to continue in read-only mode.

Also:
- Add React-callable interface for Stackdriver logging
- Add real test coverage for workspace-edit
- Add necessary workspaceApi stub support

(Note: following GIF has outdated copy)

qqb71TeeK0Y


ui/package.json
```diff
@@ -23,6 +23,7 @@
     "coverage": "ng test --source-map=false --watch=false --code-coverage"
   },
   "jest": {
+    "clearMocks": true,
```
ericsong: do you need to merge in master? I think this change already went in

ui/src/app/pages/workspace/workspace-edit.spec.tsx
```diff
@@ -1,32 +1,137 @@
...
+    const wrapper = component();
+    await waitOneTickAndUpdate(wrapper);
+    expect(wrapper).toBeTruthy();
+  });
+
+  it('displays workspaces edit page', async () => {
+    routeConfigDataStore.next({mode: WorkspaceEditMode.Edit});
+    const wrapper = component();
+    await waitOneTickAndUpdate(wrapper);
+    expect(wrapper).toBeTruthy();
```
ericsong: since we're testing three different rendering of this page, can we assert something that verified that the right page was rendered?

Anything small and unique like a word in the title.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -552,7 +562,34 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-        navigate(['workspaces', workspace.namespace, workspace.id, 'data']);
+
+        const navigateToWorkspace = () => navigate(['workspaces', workspace.namespace, workspace.id, 'data']);
+        if (this.isMode(WorkspaceEditMode.Create) || this.isMode(WorkspaceEditMode.Duplicate)) {
+          let accessLevel = null;
+          let pollTimedOut = false;
+          setTimeout(() => pollTimedOut = true, NEW_ACL_DELAY_POLL_TIMEOUT_MS);
+          while (!pollTimedOut) {
+            ({workspace, accessLevel} = await workspacesApi().getWorkspace(workspace.namespace, workspace.id));
+            console.log(accessLevel);
```
ericsong: remove log?

ui/src/app/pages/workspace/workspace-edit.spec.tsx
calbach: Thanks, meant to come back to this but I forgot - done.

PR #2722, opened 2019-10-23
[RW-3309][risk=no] Tolerate missing workspace when fetching recent resources

jaycarlton:
There are, I believe, several root causes here:

- We handle deletion of composite objects inconsistently
- Some of the ServiceImpl classes are doing too much work, by handling both persistence and business logic for the associated objects/concepts. I propose establishing things like a WorkspaceManager to handle only CRUD and maintaining referential integrity
- The patterns in the recent resource and recent workspace code are a bit different, which means for every object deletion we have to let both services know to update the appropriate object.
- There's no unifying, generic concept of Item that can be in a workspace and/or be recent, but they don't necessarily work that way.
- Workspace delete is soft, but many of the items are CASCADING in the database schema. Since workspaces are never deleted, we don't generally have good integrity
- Finally, the WorkspaceService, even though it works with objects that are soft deleted, throws a NotFoundException rather than return an inactive workspace. Throwing that exception at the controller level might be helpful, but it's very awkward and dangerous in service-service calls. I generally only expect exceptions in very rare circumstances, and an inactive workspace is not that.

There's just a lot more complication than we need or can keep up with in what are effectively CRUD endpoints in most cases. I'm going to start thinking about some patterns that can help. We may already have such patterns established elsewhere that I can adapt.

In the end, and on the advice of @gjuggler and @ericsong, I focused on making this endpoint tolerate inconsistencies in the source data, as migrating to a model where references to unreachable assets are discarded is a larger conversation.

There were a few changes necessary to the test class to make this work, since now we need to synthesize additional behavior from the workspace service. I also added a new MapStruct, which may need to move.


api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -298,10 +302,27 @@ private void setDbWorkspaceFields(
...
     dbWorkspace = workspaceService.saveWithLastModified(dbWorkspace);
     workspaceService.maybeDeleteRecentWorkspace(dbWorkspace.getWorkspaceId());
+
+    deleteRecentResourcesForWorkspace(dbWorkspace);
+
     workspaceAuditAdapterService.fireDeleteAction(dbWorkspace);
     return ResponseEntity.ok(new EmptyResponse());
   }
 
+  private void deleteRecentResourcesForWorkspace(
```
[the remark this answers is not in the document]
jaycarlton: Oh you're right. I was thinking of clone workspace. So are those items just orphaned and unreachable in the system, awaiting the day the workspace is reactivated?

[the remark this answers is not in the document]
als364: That's what it looks like right now...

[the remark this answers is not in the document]
jaycarlton: After some back and forth, I was convinced to remove this fixup on delete, at least for the purposes of this bug.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -21,6 +22,8 @@
 
   DbWorkspace findByWorkspaceId(long workspaceId);
 
+  Optional<DbWorkspace> getWorkspaceMaybe(long workspaceId);
```
jaycarlton: You could argue that `findByWorkspaceId()` should be renamed `findByWorkspaceIdOrThrow()` and the optional-returning variant should have a simpler name like `findActiveWorkspace()`.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -343,6 +344,11 @@ public WorkspaceResponse getWorkspace(String projectName, String workspaceName)
                 projectName, workspaceName, FIRECLOUD_GET_WORKSPACE_REQUIRED_FIELDS));
   }
 
+  @Override
```
jaycarlton: A purist might argue that I'm breaking a glass wall by introducing a dependency on a data model class into this service. Seems really convenient though, as I can never remember what the arguments are to this function.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
```diff
@@ -0,0 +1,27 @@
...
+import org.pmiops.workbench.db.model.DbCohort;
+import org.pmiops.workbench.db.model.DbUser;
+import org.pmiops.workbench.model.Cohort;
+
+@Mapper(
+    componentModel = "spring",
+    uses = {CommonMappers.class},
+    unmappedTargetPolicy = ReportingPolicy.IGNORE)
+public interface CohortMapper {
+  DbCohort clientToDbModel(Cohort source);
```
ericsong: we've been using `api` to refer to the client/api models.

Also, I kind of like just naming it as `toDb` or `toApi`. It's short and captures most of the context I care about and makes sense even when overloaded.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
```diff
@@ -0,0 +1,27 @@
...
+
+  Cohort dbModelToClient(DbCohort destination);
+
+  default DbUser DbUserToCreatorEmail(String creator) {
+    final DbUser result = new DbUser();
+    result.setEmail(creator);
+    return result;
```
ericsong: Something seems off about creating incorrect models like this.  I can very easily see this leading to a bug where a someone calls any non email field like `dbCohort.creator.id` and result in an NPE.

We can use DAOs to do this mapping.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  default String DbUserToCreatorEmail(DbUser creator) {
```
ericsong: The inverse is fine since we are going from "lot of data" -> "less data"

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
ericsong: I am OK with using `Optional`s but having to suffix all of our function names with `Maybe` seems like a bit much. I think it will be easy to distinguish an optional from not since the type signature is so different.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
```diff
@@ -0,0 +1,27 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.ReportingPolicy;
+import org.pmiops.workbench.db.model.DbCohort;
+import org.pmiops.workbench.db.model.DbUser;
+import org.pmiops.workbench.model.Cohort;
+
+@Mapper(
+    componentModel = "spring",
+    uses = {CommonMappers.class},
+    unmappedTargetPolicy = ReportingPolicy.IGNORE)
```
jaycarlton: This was necessary because DbCohort and Cohort each have fields the other lacks.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -176,22 +180,26 @@ public void setDistinctWorkspaceLimit(int limit) {
...
+        logger.log(
+            Level.INFO,
+            String.format("Workspace ID %d still in recent list but not found", workspaceId));
+        continue;
+      }
+      final DbWorkspace dbWorkspace = workspaceMaybe.get();
+      final WorkspaceResponse workspaceResponse = fireCloudService.getWorkspace(dbWorkspace);
+
+      Optional.ofNullable(workspaceResponse)
+          .ifPresent(r -> liveWorkspacesByIdBuilder.put(workspaceId, r));
```
ericsong: Can we keep the functional style by using a filter? I believe Optionals supports this easily.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -261,6 +282,27 @@ public void testGetUserRecentResource() {
...
   }
 
+  @Test
+  public void testGetUserRecentResources_missingWorkspace() {
+    resource1.setWorkspaceId(9999); // missing workspace
+    when(mockUserRecentResourceService.findAllResourcesByUser(user.getUserId()))
+        .thenReturn(ImmutableList.of(resource1, resource2));
+    RecentResourceResponse recentResources =
+        userMetricsController.getUserRecentResources().getBody();
+    assertEquals(1, recentResources.size());
```
ericsong: nit: I'm a fan of `AssertJ` for more readable assertions.

`assertThat(recentResources).hasSize(1)`

ericsong: another nit: I don't think its necessarily a bad thing but I prefer testing minimally by only asserting what would fail.

In this case, this test would have failed with a 404 on the `getUserRecentResources()` call so I would be content with stopping there.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -261,6 +282,27 @@ public void testGetUserRecentResource() {
...
+    RecentResource foundResource = recentResources.get(0);
+    final Cohort cohort1 = foundResource.getCohort();
+    final Cohort cohort2 = cohortMapper.dbModelToClient(resource2.getCohort());
```
ericsong: If the `cohortMapper` is only used in test, I would bring it into the test package.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
```diff
@@ -152,19 +171,21 @@ public void setUp() {
...
-            clock);
+            mockUserProvider,
+            mockUserRecentResourceService,
+            mockWorkspaceService,
+            mockFireCloudService,
+            mockCloudStorageService,
+            fakeClock);
     userMetricsController.setDistinctWorkspaceLimit(5);
+
+    cohortMapper = new CohortMapperImpl();
```
ericsong: this line should be unnecessary with autowired

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -176,22 +180,26 @@ public void setDistinctWorkspaceLimit(int limit) {
...
-                Collectors.toMap(
-                    id -> id,
-                    id -> {
-                      DbWorkspace workspace = workspaceService.findByWorkspaceId(id);
-                      WorkspaceResponse workspaceResponse =
-                          fireCloudService.getWorkspace(
-                              workspace.getWorkspaceNamespace(), workspace.getFirecloudName());
-                      return workspaceResponse;
-                    }));
+    ImmutableMap.Builder<Long, WorkspaceResponse> liveWorkspacesByIdBuilder = new Builder<>();
```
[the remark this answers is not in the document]
jaycarlton: Excellent point about multimaps. So if we have singular `keyToValue` that sounds really nice. @blrubenstein had recommended `valuesByKeys` for consistency on the plurality, but the singular makes the most sense.

It'll take me a while to switch over. HubSpot really hammered that one into me.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -176,22 +180,26 @@ public void setDistinctWorkspaceLimit(int limit) {
...
+    for (long workspaceId : workspaceIdList) {
+      final Optional<DbWorkspace> workspaceMaybe = workspaceService.getWorkspaceMaybe(workspaceId);
+      if (!workspaceMaybe.isPresent()) {
+        logger.log(
+            Level.INFO,
+            String.format("Workspace ID %d still in recent list but not found", workspaceId));
```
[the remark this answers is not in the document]
jaycarlton: But is it really something to be tolerated and expected?  I was certainly surprised it works this way, which is why I took so long on this PR.

`<rant>`
I still don't understand why we'd want to store archival/log information in our main application transaction database. (This is the only reason I've heard why we don't throw away unreachable URR entries or other workspace contents). I'd much rather have the relationships we want enforced by the RDBMS and spend programming time on building cool stuff instead of emulating the DB's constraint integrity and features in lots of places (but not everywhere).
`</rant>`

Obviously I'm in the minority, so I'll remove the statement.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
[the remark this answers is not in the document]
jaycarlton: Oh, I see. Definitely. Plus a unit test for good measure.

jaycarlton: replaced with explicit ignore rules for each one.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
jaycarlton: Optional is the next best thing to Kotlin. I don't understand why you're only OK with it.

Except you can't overload based on return type.
Screen Shot 2019-11-08 at 5 04 27 PM

My preference is to name the new optional-returning one `findActiveWorkspaceById()` or `findWorkspaceByIdIfActive` and the old one `findActiveWorkspaceByIdOrThrow()`. In other words, returning optional for something that frequently won't be there should be the norm, and throwing should be really exceptional (no pun intended).

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
jaycarlton: Ehh, there are APIs all over the place. Whose api are we talking about? It's like saying "service" or "computer"; it's kind of useless. (BTW, I have a plan to rename all those classes with a prefix, so we'd have `ApiCohort` or `ClientApiCohort`, which may help.

Also, maybe propose a convention and do it globally? Most of CH's methods are using client. I'd prefer to make sure the whole team is behind a convention before switching, and I don't mind helping with big PRs that align conventions.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
jaycarlton: Yes, we definitely can, and i just happened to do it before reading this. I originally didn't do it all in a huge stream because there were exceptions in play. It is a bit to chew on if you're not familiar with the constructs.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
jaycarlton: Re: assertion libraries, please see [this ticket](RecentResourceResponse recentResources =
        userMetricsController.getUserRecentResources().getBody()) about unifying them. It's much more important that we settle on one of them than that we pick the best one. This file is using JUnit assertions, so I just went with what we had, but I agree with you that AssertJ and Truth are much better.

jaycarlton: I'm not following this comment: 
_In this case, this test would have failed with a 404 on the getUserRecentResources() call so I would be content with stopping there._

The whole intention of the PR was to have it not do a 404 here but return those resources whose workspaces are actually reachable.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
jaycarlton: I definitely intended it to be used in more places than the test.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
jaycarlton: oh right, thanks.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
jaycarlton: If we can use the DAO, then why do we need the mapper?

We don't have very much checking at all for required parameters on any of our models. I have some thoughts about that, but are you suggesting the mapper class should make a Dao call?

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
jaycarlton: Are you thinking I should just mark the Creator field as non-mappable? I can do that.

It's still not clear where I would go to find out what fields are required on any of these classes. That's one of the things I'm going to miss about Kotlin if we take it out.

jaycarlton:
@ericsong It's fine to have blocking comments, but please indicate which of your concerns you consider to be actual errors or blockers.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
ericsong: I like the additional safety it brings but the tradeoff, to me, is that I find the interface for using it to be a bit clunky and verbose.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
ericsong: Yeah. MapStruct is integrated into Spring's DI so you can pull in the UserDao and I think it will know how to convert String -> User. If not automatically, it would be a small function or annotation to add the mapping.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
```diff
@@ -86,7 +87,7 @@ WorkspaceACLUpdateResponseList updateWorkspaceACL(
    */
   WorkspaceResponse getWorkspace(String projectName, String workspaceName);
 
-  WorkspaceResponse getWorkspace(DbWorkspace dbWorkspace);
+  Optional<WorkspaceResponse> getWorkspace(DbWorkspace dbWorkspace);
```
ericsong: I don't think this actually ever returns an empty Optional. It will throw an exception if the object DNE.

api/src/test/java/org/pmiops/workbench/api/UserMetricsControllerTest.java
jaycarlton: Moving it is no problem for now.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
jaycarlton: Excellent catch.

api/src/main/java/org/pmiops/workbench/utils/CohortMapper.java
jaycarlton: Yeah, it just worked. And I noticed it's smart enough to tell me when there's more than one function that would do the job, so it's a win-win.

jaycarlton:
This looks ready to go, but since I just checked in another jumbo PR, I'll wait until the morning.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -173,84 +174,112 @@ public void setDistinctWorkspaceLimit(int limit) {
             .limit(distinctWorkspacelimit)
```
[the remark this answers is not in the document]
jaycarlton: Oh yeah! Thanks for the reminder.

My main concern here is that we don't (yet) really know the distribution of items to workspaces. For example, 80% of our users could have all 5 recent resources in the same workspace, or maybe 75% of them have no valid/showable resources in their first `distinctWorkspaceLimit` workspaces.

I'm going to keep pondering that (separately), but I don't believe anything should be done about it on this PR (particularly if there's a possibility workspaces will become hard-deleted soon).

PR #2727, opened 2019-10-23
[RW-3539][risk=no] New landing page for users with no workspaces

ericsong:
New
Screen Shot 2019-10-23 at 12 35 02 PM

Old
Screen Shot 2019-10-23 at 12 36 11 PM

### DIFF
--- a/ui/src/app/components/CustomBulletList.tsx
+++ b/ui/src/app/components/CustomBulletList.tsx
@@ -6,6 +6,6 @@ export const CustomBulletList = ({children}) => {
 
 export const CustomBulletListItem = ({bullet, children}) => {
   return <React.Fragment>
-    <div style={{position: 'absolute', left: -2}}> {bullet} </div> <li> {children} </li>
+    <div style={{position: 'absolute', left: '-.2em'}}> {bullet} </div> <li> {children} </li>
   </React.Fragment>;
 };
--- a/ui/src/app/components/CustomBulletList.tsx
+++ b/ui/src/app/components/CustomBulletList.tsx
@@ -1,0 +1,11 @@
+import * as React from 'react';
+
+export const CustomBulletList = ({children}) => {
+  return <ul style={{position: 'relative', listStyle: 'none', marginLeft: 0, paddingLeft: '1.2em'}}>{children}</ul>;
+};
+
+export const CustomBulletListItem = ({children}) => {
+  return <React.Fragment>
+    <div style={{position: 'absolute', left: -2}}> → </div> <li> {children} </li>
+  </React.Fragment>;
+};
--- a/ui/src/app/components/CustomBulletList.tsx
+++ b/ui/src/app/components/CustomBulletList.tsx
@@ -4,8 +4,8 @@ export const CustomBulletList = ({children}) => {
   return <ul style={{position: 'relative', listStyle: 'none', marginLeft: 0, paddingLeft: '1.2em'}}>{children}</ul>;
 };
 
-export const CustomBulletListItem = ({children}) => {
+export const CustomBulletListItem = ({bullet, children}) => {
   return <React.Fragment>
-    <div style={{position: 'absolute', left: -2}}> → </div> <li> {children} </li>
+    <div style={{position: 'absolute', left: -2}}> {bullet} </div> <li> {children} </li>
   </React.Fragment>;
 };
--- a/ui/src/app/components/CustomBulletList.tsx
+++ b/ui/src/app/components/CustomBulletList.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react';
 
 export const CustomBulletList = ({children}) => {
-  return <ul style={{position: 'relative', listStyle: 'none', marginLeft: 0, paddingLeft: '1.2em'}}>{children}</ul>;
+  return <ul style={{position: 'relative', listStyle: 'none', marginLeft: 0}}>{children}</ul>;
 };
 
 export const CustomBulletListItem = ({bullet, children}) => {
   return <React.Fragment>
-    <div style={{position: 'absolute', left: '-.2em'}}> {bullet} </div> <li> {children} </li>
+    <div style={{position: 'absolute', left: '-.2em'}}> {bullet} </div> <li style={{marginLeft: '1.2em'}}> {children} </li>
   </React.Fragment>;
 };
--- a/ui/src/app/components/headers.tsx
+++ b/ui/src/app/components/headers.tsx
@@ -19,6 +19,9 @@ export const styles = {
     fontWeight: 800,
     fontSize: '1rem',
   },
+  lessBold: {
+    fontWeight: 600
+  },
   h4: {
     fontWeight: 300,
     color: colors.primary,
@@ -42,5 +45,6 @@ export const styles = {
 export const BolderHeader = withStyle(styles.h1)('h1');
 export const BoldHeader = withStyle(styles.h2)('h2');
 export const Header = withStyle(styles.h3)('h3');
+export const LessBoldHeader = withStyle({...styles.h3, ...styles.lessBold})('h3');
 export const SmallHeader = withStyle(styles.h4)('h4');
 export const ListPageHeader = withStyle(styles.listPageHeader)('h3');
--- a/ui/src/app/components/headers.tsx
+++ b/ui/src/app/components/headers.tsx
@@ -19,7 +19,7 @@ export const styles = {
     fontWeight: 800,
     fontSize: '1rem',
   },
-  lessBold: {
+  semiBold: {
     fontWeight: 600
   },
   h4: {
@@ -45,6 +45,6 @@ export const styles = {
 export const BolderHeader = withStyle(styles.h1)('h1');
 export const BoldHeader = withStyle(styles.h2)('h2');
 export const Header = withStyle(styles.h3)('h3');
-export const LessBoldHeader = withStyle({...styles.h3, ...styles.lessBold})('h3');
+export const SemiBoldHeader = withStyle({...styles.h3, ...styles.semiBold})('h3');
 export const SmallHeader = withStyle(styles.h4)('h4');
 export const ListPageHeader = withStyle(styles.listPageHeader)('h3');
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -9,6 +9,7 @@ import {
   Clickable,
 } from 'app/components/buttons';
 import {FadeBox} from 'app/components/containers';
+import {CustomBulletList, CustomBulletListItem} from 'app/components/CustomBulletList';
 import {FlexColumn, FlexRow} from 'app/components/flex';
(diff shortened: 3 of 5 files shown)


ui/src/app/styles/colors.ts
```diff
@@ -11,6 +11,7 @@ export default {
   success: '#8BC990',
   // Used for tables
   light: '#E9ECEF',
+  lightBackground: '#E6E4Ed',
```
andy7i: Is this in our official list of colors? If not, we should ask Lou to add it in. Or maybe `light` is the color Lou wanted?

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -390,24 +402,47 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
-                              <SmallHeader>Recently Accessed Items</SmallHeader>
-                              <RecentResources/>
+                              {this.state.userHasWorkspaces ? (<React.Fragment>
+                                <SmallHeader>Recently Accessed Items</SmallHeader>
+                                <RecentResources/>
+                              </React.Fragment>) : <div style={{backgroundColor: colors.lightBackground,
+                                color: colors.primary,
+                                borderRadius: 10,
+                                padding: '10px 30px',
+                                margin: '30px 0px'}}>
```
andy7i: Looks like the sides of the fadebox are drawn over

andy7i: Recommend using rem instead of px. 

further info
https://engageinteractive.co.uk/blog/em-vs-rem-vs-px

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,22 +41,30 @@ export const RecentWorkspaces = withUserProfile()
...
-        })
-      }
-      {this.state.loading && <SpinnerOverlay dark={true} />}
-    </FlexRow>;
+    return <React.Fragment>
+        {!this.state.loading && this.state.recentWorkspaces.length === 0 ? (
+          <div style={{color: colors.primary, padding: '10px 30px'}}>
+            <h2 style={{fontWeight: 600}}>Create your first Workspace</h2>
+            <div>As you create your workspaces, this area will store your most recent workspaces.</div>
+            <div>To see all workspaces created, click on <b>See all Workspaces</b> to the right.</div>
```
andy7i: The mock shows the `See all Workspaces` link greyed out, but the code leaves it enabled. *However* I think it's okay to leave it enabled. It might be worth asking Lou though.

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -390,24 +402,47 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+                                <h2 style={{fontWeight: 600, marginTop: 0}}>Here are some tips to get you started:</h2>
+                                <CustomBulletList>
+                                  <CustomBulletListItem bullet='→'>
+                                    Create a <a href='https://support.google.com/chrome/answer/2364824'>Chrome Profile </a>
```
andy7i: The text looks a bit small compared to the mock, but maybe that's just the screenshot. The mocks have the text at 14px, which I think is the smallest we go.
This also applies to the text:
```
As you create your workspaces, this area will store your most recent workspaces. 
To see all workspaces created, click on See all Workspaces to the right.
```

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -390,24 +402,47 @@ export const Homepage = withUserProfile()(class extends React.Component<
                               <RecentWorkspaces />
                             </FlexColumn>
                             <FlexColumn>
-                              <SmallHeader>Recently Accessed Items</SmallHeader>
-                              <RecentResources/>
+                              {this.state.userHasWorkspaces ? (<React.Fragment>
+                                <SmallHeader>Recently Accessed Items</SmallHeader>
+                                <RecentResources/>
+                              </React.Fragment>) : <div style={{backgroundColor: colors.lightBackground,
```
andy7i: nit: for readability, could we format the ternary a bit differently? Something like:

```
{this.state.userHasWorkspaces ?

(<React.Fragment><SmallHeader>Recently Accessed Items</SmallHeader>
     <RecentResources/></React.Fragment>) : 

<div style={{backgroundColor: colors.lightBackground ..........
```

formatting multi-line ternary https://eslint.org/docs/rules/multiline-ternary

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -390,24 +402,47 @@ export const Homepage = withUserProfile()(class extends React.Component<
...
+                                margin: '30px 0px'}}>
+                                <h2 style={{fontWeight: 600, marginTop: 0}}>Here are some tips to get you started:</h2>
+                                <CustomBulletList>
+                                  <CustomBulletListItem bullet='→'>
+                                    Create a <a href='https://support.google.com/chrome/answer/2364824'>Chrome Profile </a>
+                                    with your All of Us Researcher workbench Google account. This will keep your Workbench
+                                    browser sessions isolated from your other Google accounts.
+                                  </CustomBulletListItem>
+                                  <CustomBulletListItem bullet='→'>
+                                    Check out <a onClick={() => navigate(['library'])}> Featured Workspaces </a>
```
andy7i: not your code or within the scope of this PR -- this link doesn't play nicely with command+click to open in a new tab -- something we need to refactor in the app.

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,22 +41,30 @@ export const RecentWorkspaces = withUserProfile()
...
+          </div>) : (<div>
```
andy7i: same comment about multi-line ternary formatting as before

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,22 +41,30 @@ export const RecentWorkspaces = withUserProfile()
...
+              })
+            }
+          </FlexRow>
+        </div>)
+        }
+    </React.Fragment>;
   }
 });
+
+//       {this.state.loading && <SpinnerOverlay dark={true} />}
```
andy7i: did you mean to leave this as a comment? / any reason for removing this?

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,22 +41,30 @@ export const RecentWorkspaces = withUserProfile()
...
-            accessLevel={recentWorkspace.accessLevel}
-            reload={() => this.loadWorkspaces()}
-          />;
-        })
-      }
-      {this.state.loading && <SpinnerOverlay dark={true} />}
-    </FlexRow>;
+    return <React.Fragment>
+        {!this.state.loading && this.state.recentWorkspaces.length === 0 ? (
+          <div style={{color: colors.primary, padding: '10px 30px'}}>
```
andy7i: maybe this isn't the best place for putting the padding? We can add padding in the parent div of RecentWorkspaces so that the parent is aware of how much padding it needs.

ui/src/app/components/headers.tsx
```diff
@@ -19,6 +19,9 @@ export const styles = {
     fontWeight: 800,
     fontSize: '1rem',
   },
+  lessBold: {
```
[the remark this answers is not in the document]
ericsong: Do I need to be invited to this? The link isn't working for me.

ui/src/app/pages/homepage/homepage.tsx
ericsong: yeah... this is what prompted my discussion at the eng-sync yesterday.

ui/src/app/pages/homepage/recent-workspaces.tsx
ericsong: forgot! Adding it back

ui/src/app/components/headers.tsx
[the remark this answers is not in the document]
calbach: ^ same. If we actually had an AoU style guide this entire time that several team members did not have access to I'll have some follow-up questions...

ui/src/app/pages/homepage/recent-workspaces.tsx
ericsong: The parent div won't work in this case unless I add a new div in between. The parent div contains the Workspace header which does not need the padding.

Is there an anti pattern with padding your own content?

ui/src/app/components/CustomBulletList.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export const CustomBulletList = ({children}) => {
+  return <ul style={{position: 'relative', listStyle: 'none', marginLeft: 0, paddingLeft: '1.2em'}}>{children}</ul>;
+};
+
+export const CustomBulletListItem = ({bullet, children}) => {
+  return <React.Fragment>
+    <div style={{position: 'absolute', left: -2}}> {bullet} </div> <li> {children} </li>
```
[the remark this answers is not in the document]
ericsong: I had trouble getting it to work with flexbox. My guess is that flexbox can be awkward in "
hacky" css tricks like this.

ui/src/app/pages/homepage/homepage.tsx
calbach: Note: here is the react conversion context I was mentioning: https://github.com/all-of-us/workbench/pull/1893#r265816213

ui/src/app/pages/homepage/homepage.tsx
ericsong: What do you mean? I don't think I'm seeing this.

ui/src/app/components/headers.tsx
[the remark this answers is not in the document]
andy7i: sent!

ui/src/app/pages/homepage/recent-workspaces.tsx
andy7i: Yeah I think you still want margin here. I'm realizing it probably doesn't need to be in a parent element.

More info: https://stackoverflow.com/questions/2189452/when-to-use-margin-vs-padding-in-css

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,29 +43,34 @@ export const RecentWorkspaces = withUserProfile()
...
-          </FlexRow>
-        </div>)
-        }
+      {
+        this.state.loading ?
+          <SpinnerOverlay dark={true} /> :
+
+          this.state.recentWorkspaces.length === 0 ?
+
+            <div style={{color: colors.primary, padding: '.5em 2em'}}>
```
andy7i: tl;dr from https://stackoverflow.com/questions/2189452/when-to-use-margin-vs-padding-in-css

>TL;DR: By default I use margin everywhere, except when I have a border or background and want to increase the space inside that visible box.

I think we should use margins here.

ericsong: gotcha. will do

ericsong:
poof. no more padding @andy7i

PR #2729, opened 2019-10-23
[RW-3754][Risk=no] Remove getWorkspace from create

blrubenstein:

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
@@ -29,6 +29,7 @@ import org.pmiops.workbench.firecloud.model.ManagedGroupWithMembers;
 import org.pmiops.workbench.firecloud.model.Me;
 import org.pmiops.workbench.firecloud.model.NihStatus;
 import org.pmiops.workbench.firecloud.model.Profile;
+import org.pmiops.workbench.firecloud.model.Workspace;
 import org.pmiops.workbench.firecloud.model.WorkspaceACL;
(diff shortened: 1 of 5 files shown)


blrubenstein:
Already done! It appears to work fine and creates a new workspace.

jaycarlton:
@blrubenstein if you don't mind, next time it would help on non-trivial PRs to put a description discussing what the problem was, any context around it, what the solution is, and any testing done.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -290,17 +291,13 @@ public void removeOwnerFromBillingProject(
...
     WorkspaceIngest workspaceIngest = new WorkspaceIngest();
     workspaceIngest.setName(workspaceName);
     workspaceIngest.setNamespace(projectName);
     checkAndAddRegistered(workspaceIngest);
-    retryHandler.run(
-        (context) -> {
-          workspacesApi.createWorkspace(workspaceIngest);
-          return null;
-        });
+    return retryHandler.run((context) -> workspacesApi.createWorkspace(workspaceIngest));
```
jaycarlton: maybe name your lambda parameter `unusedContext` or similar.

blrubenstein:
Sure, I can throw that context on there in the future. Although I'd rather details around problem, context, and solution be on the ticket (which I also made, and could have put more details on). In this case, there was an extra get call that wasn't necessary which was skewing our perf results. This involved removing that get call, since the create call had all the information we needed in the response.

jaycarlton:
👍 For most PRs I don't go back to look up the ticket, since it's usually a discussion of the problem only. Also our github is public but Jira is private, fwiw.

PR #2730, opened 2019-10-23
[no ticket][risk=no] push @Transactional all the way up the ServiceImpl call stack

als364:
Apparently if a method without a `@Transactional` calls a method with a `@Transactional` in the same classfile you will get an error. This has to do with how Spring handles objects - for things like this it wraps the entire instance in a proxy that intercepts calls to annotated methods. If the first method called on the instance is not a `@Transactional` method, the proxy won't catch the call.

I _could_ set `@Transactional` on the class itself, which says 'by default, run everything in this class in a transaction with propagation required'. Bringing that one up in eng sync.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
@@ -593,12 +593,12 @@ public class WorkspaceServiceImpl implements WorkspaceService {
   }
 
   @Override
+  @Transactional
   public UserRecentWorkspace updateRecentWorkspaces(Workspace workspace) {
     return updateRecentWorkspaces(
         workspace, userProvider.get().getUserId(), new Timestamp(clock.instant().toEpochMilli()));
   }
 
-  @Transactional
   private void handleWorkspaceLimit(long userId) {
     List<UserRecentWorkspace> userRecentWorkspaces =
         userRecentWorkspaceDao.findByUserIdOrderByLastAccessDateDesc(userId);


api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -593,12 +593,12 @@ public UserRecentWorkspace updateRecentWorkspaces(
   }
 
   @Override
+  @Transactional
```
freemabd: Do we need a test case that proves rollback when this method fails?

als364:
I'm using `@Transactional` because I'm getting a `TransactionRequiredException` when I don't.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
als364: I've had trouble getting a unit test to replicate the exception I was seeing that prompted me to add an `@Transactional` in the first place.

freemabd: Yeah, these types of test cases are usually tough to implement. Thanks for trying.

jaycarlton:
Should this be done everywhere?

als364:
We should consider putting `@Transactional` on all our Services if we think about them that way anyways

calbach:
`@Transactional` translates to start/commit of a SQL transaction. You can imagine that for performance or scoping reasons you would not want to do this 100% of the time, though it would be useful to understand exactly what the impact would be (e.g. exactly what this translates to in MySQL, and how locking would work). In general for transactions you'd want to be fairly explicit about starting/ending them. I don't think it makes sense on our services as currently defined. If we had something like Jay's proposed "entity manager" layer (which could just be a type of service), then putting `@Transactional` on every method might make sense.

jaycarlton:
Shameless plug https://github.com/all-of-us/workbench/pull/2740

PR #2732, opened 2019-10-23
initial version of prefix script

jaycarlton:
There's a handy Swagger Codegen [option](https://github.com/all-of-us/workbench/pull/2733/files#diff-dd8fd6d45c4e2eccf0f278e121d714d8R51) to add a prefix to generated model names. Script to fix up references to classes that have had a prefix applied. E.g Employee -> FooEmployee.

There are several combinations of importing and use with qualification that we need to sort out. I'm wondering if I can use other tricks and hacks in IntelliJ to make this script moot. [Example output](https://github.com/all-of-us/workbench/pull/2733)

### DIFF
--- a/api/tools/prefix_generated_classes.rb
+++ b/api/tools/prefix_generated_classes.rb
@@ -1,0 +1,45 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+require 'find'
+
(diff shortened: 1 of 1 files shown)


jaycarlton:
closing in favor of https://github.com/all-of-us/workbench/pull/2752

PR #2735, opened 2019-10-23
[risk=no][RW-3714] Fix Data Dictionary loading bug

ericsong:
We were getting a duplicate upload error because it was fetching via cdr name but saving with the default cdr.

Also adds an export file for the new prod CDR.

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -198,6 +198,9 @@ def dev_up()
     ./gradlew loadConfig -Pconfig_key=featuredWorkspaces -Pconfig_file=config/featured_workspaces_local.json
   }
 
+  common.status "Loading Data Dictionary..."
+  common.run_inline %W{docker-compose run api-scripts ./gradlew loadDataDictionary -PappArgs=false}
+
   run_api()
 end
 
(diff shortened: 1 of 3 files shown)


api/libproject/devstart.rb
```diff
@@ -198,6 +198,9 @@ def dev_up()
     ./gradlew loadConfig -Pconfig_key=featuredWorkspaces -Pconfig_file=config/featured_workspaces_local.json
   }
 
+  common.status "Loading Data Dictionary..."
+  common.run_inline %W{docker-compose run api-scripts ./gradlew loadDataDictionary -PappArgs=false}
```
andy7i: It looks like we're adding data dictionary to the deploy process? Is this part of the ticket or something extra?

api/libproject/devstart.rb
```diff
@@ -1665,6 +1668,8 @@ def deploy(cmd_name, args)
     versions_file = must_get_env_value(gcc.project, :cdr_versions_json)
     update_cdr_versions_for_project("config/#{versions_file}", op.opts.dry_run)
 
+    common.run_inline %W{gradle loadDataDictionary -PappArgs=#{op.opts.dry_run ? true : false}}
```
andy7i: I'm not clear on the difference between `deploy()` and `dev-up` and why dev-up uses docker and deploy does not.

api/libproject/devstart.rb
andy7i: I realized that this is part of 
`Revert "Remove data dictionary tasks from dev scripts. (#2693)"` which reverts reverted code.

api/libproject/devstart.rb
ericsong: I believe it does run in docker actually. The first line of `deploy()` is `ensure_docker cmd_name, args` which will run the rest of the function in a docker environment.

api/libproject/devstart.rb
ericsong: 👍 I'm reverting the commit that removed it from our dev-up and deploy tasks. It should be there IF it wasn't buggy.

PR #2736, opened 2019-10-23
[RW-3621][Risk=no] Srubenst/data set quick tour

blrubenstein:
This PR adds datasets to the quick tour modal, updates the images in the quick tour to the latest images, properly aligns the quick tour selector region, and puts all images for the quick tour into a subfolder within our assets folder.


ui/src/app/pages/homepage/quick-tour-modal.tsx
```diff
@@ -325,7 +337,7 @@ export class QuickTourReact extends React.Component<QuickTourReactProps, QuickTo
           <div style={styles.breadcrumbs}>
             {panels.map((p, i) => {
               return <React.Fragment key={i}>
-                <div style={{width: '128px'}}>
+                <div style={{width: i === panels.length - 1 ? null : '128px'}}>
```
blrubenstein: I really dislike this area of code, but it blows up the scope to try to fix it, because the container magic would get complicated.

blrubenstein:
I added a folder to our quick tour images, because our image asset folder was getting a lot of static files and it was hard to navigate. It is probably worth doing a quick look to make sure we still use all the images/files there.

PR #2738, opened 2019-10-23
[risk=no] Increase staging tasks to support one-time load test

calbach:
This matches perf. We don't want instance scaling to bias the load testing results. This should be reverted once the staging test has been completed.

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -61,7 +61,7 @@ ENVIRONMENTS = {
     :config_json => "config_staging.json",
     :cdr_versions_json => "cdr_versions_staging.json",
     :featured_workspaces_json => "featured_workspaces_staging.json",
-    :gae_vars => TEST_GAE_VARS
+    :gae_vars => make_gae_vars(20, 20)
   },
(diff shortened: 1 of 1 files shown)


calbach:
@gjuggler

aweng98:
@calbach Karthik said don't run perf test in prod (staging). So I thought we won't need to run test in staging anymore.

calbach:
Karthik said don't run "today". We can hold off on merging this until we confirm we actually need to run it. I still want the PR ready so people don't forget or misunderstand the change I am suggesting.

jaycarlton:
Have we done tests at several instance counts to track how the scaling improves performance?

calbach:
No. As of now we don't yet have any indication that CPU/memory on our side will be a limiting factor. This effectively fixes one parameter to a sufficiently high value so that we can focus on establishing baseline performance against Terra. Based on the current numbers, trying to tune our instance counts would be premature, though I think ideally we'd get there ahead of launch. Testing with autoscaling would also be good, once we're satisfied with the other parts of the system.

calbach:
Looks like we won't have to do this. Dropping.

PR #2741, opened 2019-10-24
[No Ticket][Risk=no] Remove unused images

blrubenstein:
I don't believe we are using any of these image files. This PR deletes them.


jaycarlton:
Awesome. How did you verify they are unused?

Is there a tool that can check the build for unreferenced resources?

ericsong:
I couldn't find any references either. Approved

blrubenstein:
@jaycarlton I used find in path on the UI repo on the filename.

PR #2744, opened 2019-10-24
[RW-3688][risk=no] Sections under Featured workspaces

ericsong:
- Adds categories for featured workspaces
- Removes original featured workspaces tab
- Fixes bug where we were indicating that the page has loaded after only 1 or 2 requests has completed.

Screen Shot 2019-10-24 at 4 12 36 PM
Screen Shot 2019-10-24 at 4 12 43 PM

### DIFF
--- a/common-api/src/main/resources/common_api.yaml
+++ b/common-api/src/main/resources/common_api.yaml
@@ -93,6 +93,10 @@ definitions:
     type: string
     enum: &Role [UNDERGRADUATE, TRAINEE, FELLOW, EARLY_CAREER, NON_TENURE, MID_CAREER, LATE_CAREER, PROJECT_PERSONNEL]
 
+  FeaturedWorkspaceCategory:
+    type: string
+    enum: [PHENOTYPE_LIBRARY, TUTORIAL_WORKSPACES]
+
   CdrVersionListResponse:
     type: object
     required:
@@ -206,7 +210,7 @@ definitions:
         type: string
         description: the Firecloud id of the workspace
       category:
-        type: string
+        $ref: '#/definitions/FeaturedWorkspaceCategory'
         description: The category that this workspace belongs to
 
   ErrorResponse:
--- a/common-api/src/main/resources/common_api.yaml
+++ b/common-api/src/main/resources/common_api.yaml
@@ -205,6 +205,9 @@ definitions:
       id:
         type: string
         description: the Firecloud id of the workspace
+      category:
+        type: string
+        description: The category that this workspace belongs to
 
   ErrorResponse:
     type: object
--- a/ui/src/app/pages/workspace/workspace-library.tsx
+++ b/ui/src/app/pages/workspace/workspace-library.tsx
@@ -14,7 +14,7 @@ import colors, {colorWithWhiteness} from 'app/styles/colors';
 import {reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {WorkspacePermissions} from 'app/utils/workspace-permissions';
 import {environment} from 'environments/environment';
-import {ErrorResponse, FeaturedWorkspace, Profile} from 'generated/fetch';
+import {ErrorResponse, FeaturedWorkspace, FeaturedWorkspaceCategory, Profile} from 'generated/fetch';
 
 const styles = reactStyles({
   navPanel: {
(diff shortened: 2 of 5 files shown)


ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -40,17 +39,40 @@ const styles = reactStyles({
...
-  FEATURED_WORKSPACES: {
-    title: 'Featured Workspaces',
-    icon: 'star'
-  },
   PUBLISHED_WORKSPACES: {
     title: 'Published Workspaces',
-    icon: 'library'
+    icon: 'bookmark',
+    filter: (publishedWorkspaces: WorkspacePermissions[], featuredWorkspaces: FeaturedWorkspace[]) => {
+      return publishedWorkspaces.filter(ws => !featuredWorkspaces.find(fws =>
```
gjuggler: Optional nit: I'm generally averse to overly-short abbreviations, because they can end up being difficult to grok and chars are cheap. But I'm happy to acknowledge this battle is lost if we have a strong pattern in our codebase to use 1-3 letter variable names. Do you have a clear sense on this?

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -40,17 +39,40 @@ const styles = reactStyles({
...
+  libraryTabDivider: {
+    width: '100%',
+    margin: '0.5rem 0',
+    backgroundColor: colorWithWhiteness(colors.dark, .5),
+    border: '0 none',
+    height: 1
   }
 });
 
 const libraryTabEnums = {
```
gjuggler: Optional: not your code, but these aren't really enums... maybe libraryTabConfig?

common-api/src/main/resources/common_api.yaml
```diff
@@ -205,6 +205,9 @@ definitions:
       id:
         type: string
         description: the Firecloud id of the workspace
+      category:
+        type: string
```
gjuggler: Can you add an "enum: [PHENOTYPE_LIBRARY, TUTORIAL_WORKSPACES]" here, or defined separately as a top-level enum?

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -40,17 +39,40 @@ const styles = reactStyles({
...
+      return publishedWorkspaces.filter(ws => !featuredWorkspaces.find(fws =>
+        ws.workspace.id === fws.id && ws.workspace.namespace === fws.namespace));
+    }
+  },
+  PHENOTYPE_LIBRARY: {
+    title: 'Phenotype Library',
+    icon: 'dna',
+    filter: (publishedWorkspaces: WorkspacePermissions[], featuredWorkspaces: FeaturedWorkspace[]) => {
+      return publishedWorkspaces.filter(ws => !!featuredWorkspaces.find(fws =>
+        ws.workspace.id === fws.id && ws.workspace.namespace === fws.namespace && fws.category === 'PHENOTYPE_LIBRARY'));
```
gjuggler: Opt: Raw strings that are really enums is a tiny red flag. If you define the enum as a named entity in Swagger, you can probably refer to that enum value here. Not a huge deal, just a minor convenience / typo safety.

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -110,17 +131,23 @@ export const WorkspaceLibrary = withUserProfile()
...
-      libraryTabEnums.FEATURED_WORKSPACES
+      libraryTabEnums.PHENOTYPE_LIBRARY,
+      libraryTabEnums.TUTORIAL_WORKSPACES
     ];
 
   async componentDidMount() {
     this.updateWorkspaces();
   }
 
+  areWorkspacesLoading() {
```
gjuggler: Nice as-you-go UX fix, thanks

ui/src/app/pages/workspace/workspace-library.tsx
```diff
@@ -131,19 +158,21 @@ export const WorkspaceLibrary = withUserProfile()
   async updateWorkspaces() {
     await this.getAllPublishedWorkspaces();
     await this.getFeaturedWorkspaces();
-    this.filterPublishedWorkspaces();
   }
 
   // Gets all published workspaces, including those configured as 'featured'
   async getAllPublishedWorkspaces() {
+    this.setState({
```
gjuggler: React nit (for here and below): setState calls which rely on the current state should be done with the "updater" function parameter. From https://reactjs.org/docs/react-component.html#setstate: "If you need to set the state based on the previous state, read about the updater argument below."

See https://github.com/all-of-us/workbench/blob/f796b9bc2e343bb251fc4373707025741c4e4935/ui/src/app/components/side-nav.tsx#L111 for an example.

ui/src/app/pages/workspace/workspace-library.tsx
ericsong: I'm all for longer variable names. The few times I break that rule is if I'm writing a small-ish inline function but I'm happy to change to their full names

ui/src/app/pages/workspace/workspace-library.tsx
ericsong: Agreed. I had it as a string at first because I was toying with the idea of generated categories dynamically with new configs but this is better since I didn't follow through with that idea.

ui/src/app/pages/workspace/workspace-library.tsx
jaycarlton: Yeah, +1 for tiny var names if they're going to be extremely short-lived (i.e. 1-line lambda or equivalent) and the value/context is obvious. Or in situations where we need to access it 10 times in a row in a conversion or factory method, such as when doing what ObjectMapper does.

ui/src/app/pages/workspace/workspace-library.tsx
jaycarlton: I bet we have the same issues with impoverished generated enums here as on the backend.

PR #2747, opened 2019-10-24
[RW-3773][risk = no] Capture names, initials with which DUA was signed

als364:
Still need to figure out how to view as if I'm a user with a signed DUA, but this is mostly done

### DIFF
--- a/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
+++ b/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
@@ -26,7 +26,7 @@
       <column name="data_use_agreement_signed_version" type="integer">
         <constraints nullable="false"/>
       </column>
-      <column name="completion_time">
+      <column name="completion_time" type="timestamp">
         <constraints nullable="false"/>
       </column>
     </createTable>
--- a/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
+++ b/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
@@ -22,7 +22,7 @@
       <column name="user_initials" type="varchar(6)">
         <constraints nullable="false"/>
       </column>
-      <column name="user_name_out_of_date" type="tinyint" />
+      <column name="user_name_out_of_date" type="tinyint" defaultValue="0"/>
       <column name="data_use_agreement_signed_version" type="integer">
         <constraints nullable="false"/>
       </column>
--- a/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
+++ b/api/db/changelog/db.changelog-108-user-data-use-agreement.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="aschwart" id="changelog-108-user-data-use-agreement">
+
+    <createTable tableName="user_data_use_agreement">
+      <column name="user_data_user_agreement_id" type="bigint" autoIncrement="true">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="user_id" type="bigint">
+        <constraints nullable="false" foreignKeyName="fk_dua_usage_user_id" references="user(user_id)"/>
+      </column>
+      <column name="user_given_name" type="varchar(80)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="user_family_name" type="varchar(80)">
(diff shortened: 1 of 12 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -287,14 +292,34 @@ public User createUser(
...
-        (user) -> {
-          user.setDataUseAgreementCompletionTime(timestamp);
-          user.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
-          return user;
+    UserDataUseAgreement dataUseAgreement = new UserDataUseAgreement();
+    dataUseAgreement.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
+    dataUseAgreement.setUserId(userProvider.get().getUserId());
+    dataUseAgreement.setUserFamilyName(userProvider.get().getFamilyName());
+    dataUseAgreement.setUserGivenName(userProvider.get().getGivenName());
+    dataUseAgreement.setUserInitials(initials);
```
jaycarlton: Please test with periods in initials, like "O.J.C". Also, with only 6 chars, I can't put spaces if I want.

api/src/main/java/org/pmiops/workbench/db/model/UserDataUseAgreement.java
```diff
@@ -0,0 +1,129 @@
...
+  @Column(name = "user_initials")
+  public String getUserInitials() {
+    return userInitials;
+  }
+
+  public void setUserInitials(String userInitials) {
+    this.userInitials = userInitials;
+  }
+
+  @Column(name = "user_name_out_of_date")
```
jaycarlton: Please put some comments about what this means. Does it cover both soft and "hard" name changes?

api/src/main/java/org/pmiops/workbench/db/model/UserDataUseAgreement.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "user_data_use_agreement")
+public class UserDataUseAgreement {
```
jaycarlton: I *think* you can write this in Kotlin now and drop 3/4 of the code. @ericsong do you know if it's supported for Entity classes?

als364: Tried it - having trouble figuring out how to make it support autoincrementing primary keys. Doing something like the following, it won't represent the PK as a `Long` instead of a `long` and so I can't null it out

```@Entity
@Table(name = "user_data_use_agreement")
data class UserDataUseAgreement(
        @Id
        @GeneratedValue(strategy = GenerationType.IDENTITY)
        @Column(name = "user_data_use_agreement_id")
        var userDataUseAgreementId: Long,

        @Column(name="user_id")
        var userId: Long,

        @Column(name = "user_given_name")
        var userGivenName: String,

        @Column(name = "user_family_name")
        var userFamilyName: String,

        @Column(name = "user_initials")
        var userInitials: String,

        // This is set to 'true' whenever user_given_name or user_family_name are not the same as the
        // given_name or family_name on the User entry referenced by user_id.
        @Column(name = "user_name_out_of_date")
        var userNameOutOfDate: Boolean,

        @Column(name = "data_use_agreement_signed_version")
        var dataUseAgreementSignedVersion: Int,

        @Column(name = "completion_time")
        var completionTime: Timestamp
)```

Comment (als364) [reply api/src/main/java/org/pmiops/workbench/db/model/UserDataUseAgreement.java]: (this may still be doable, but I don't think it's worth doing now, the day of the deadline)

Comment (ericsong) [reply api/src/main/java/org/pmiops/workbench/db/model/UserDataUseAgreement.java]: 👍 I think we can hold off on using it for now. It's a little tricky with entity models. I'll run a Kotlin primer soon™

**Comment by jaycarlton** [comment]
> Still need to figure out how to view as if I'm a user with a signed DUA, but this is mostly done

In your local or test db (needs startproxy cmd), find yourself in the user table and set the dua acceptance timestamp to something > 0.

**Comment by als364** [comment]
Yeah, looks like we still don't display anything different on the 'show current agreement' page if you've signed. Maybe we should

File: api/db/changelog/db.changelog-108-user-data-use-agreement.xml [hunk 4]
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="aschwart" id="changelog-108-user-data-use-agreement">
```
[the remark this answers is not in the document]
als364: Hardly any of our changelogs have verbs afaict

PR #2749, opened 2019-10-24
[RW-3776][risk=no] Remove howtouse section from homepage

andy7i:
Remove the section “How to Use the All of Us Researcher Workbench” from home page.

Styles and HTML were removed. Margin was added to the bottom between the main page and the bottom banner.

The goal is to re-add this in the future. I opted to delete the code rather than commenting the code out in order to keep the code base a bit cleaner. We can pull in whatever we need at that point.

Code for the old page (without feature flag) was removed in its own commit so that we don't have to pull it back in later.

Before
Screen Shot 2019-10-24 at 5 44 00 PM

After

Screen Shot 2019-10-24 at 5 44 25 PM
Screen Shot 2019-10-24 at 5 55 23 PM

### DIFF
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -13,7 +13,6 @@ import {FlexColumn, FlexRow} from 'app/components/flex';
 import {Header, SmallHeader} from 'app/components/headers';
 import {ClrIcon} from 'app/components/icons';
 import {Modal} from 'app/components/modals';
-import {TooltipTrigger} from 'app/components/popups';
 import {Spinner} from 'app/components/spinners';
 import {QuickTourReact} from 'app/pages/homepage/quick-tour-modal';
 import {RecentResources} from 'app/pages/homepage/recent-resources';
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -107,30 +83,9 @@ export const styles = reactStyles({
...
-  footerTextToDelete: {
-    height: '176px', opacity: 0.87, color: colors.secondary, fontSize: '16px',
-    fontWeight: 400, lineHeight: '30px', display: 'flex', width: '100%',
-    flexDirection: 'column', flexWrap: 'nowrap', overflowY: 'auto'
-  },
-  linksBlockToDelete: {
-    display: 'flex', marginBottom: '1.2rem', marginLeft: '1.4rem',
-    flexDirection: 'column', flexShrink: 1, minWidth: 0
-  },
   bottomBannerToDelete: {
```
jmthibault79: Are these other "ToDelete"s ready to delete as well?

andy7i: Not yet! Those are for when we delete the old code for the `enableHomepageRestyle` feature flag.

PR #2733, opened 2019-10-23
Demo output of swagger class with prefix fixup script.

jaycarlton:
Current state of the fixer-upper script's output. Needs a few more iterations to fixup missing imports, missed renames, etc.

PR #2750, opened 2019-10-25
[RW-3426][RISK=NO] Data Model Beauty Pageant 

jaycarlton:
The data models in the `org.pmiops.workbench.db.model` package have names that are the same as API models (and sometimes other things too, like vendor API models). They are distinguishable by package, and by cmd+click in the IDE, but for reading code at a glance without context (like in PRs or in the Find window), it's really tricky to tell which Workspace is which. This PR is a chance to see several schemes (or add your own).

Some ideas I had are here:
- `Workspace -> DbWorkspace`: a succinct prefix that gets to the point, but isn't idiomatic Java
- `Address -> AddressEntity`: Probably my favorite: visually distinctive, easy to pronounce, no abbreviations. However, not all these classes are titled `@Entity`
- `Cohort -> CohortDataModel`: also a classic choice, but not all data models are database data models. A bit long to read and say, but probably worth it for `DatasetDataModel`


calbach:
How about `Foo` -> `StorageFoo`?

jaycarlton:
@calbach I forgot that one. The rules of the game are you rename one of them the way you like it and merge it into this branch.

My personal opinion of a `Storage` prefix is that there are many kinds of storage, like buckets, Big Query, the File System, so I don't think it's especially descriptive.

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -50,7 +50,7 @@
   @Query(
       "SELECT p.fireCloudProjectName "
           + "FROM BillingProjectBufferEntry p "
-          + "JOIN Workspace w "
+          + "JOIN DbWorkspace w "
```
jaycarlton: This is a manual fix. Apparently this kind of annotation doesn't read the table name of the `@Table` annotation

jaycarlton:
If anyone's wondering "Why can't we also rename the Swagger-generated classes?", I'm working on that too. The rename is very easy, but the fixup of the usages takes a bit of finesse.


https://github.com/all-of-us/workbench/pull/2732 for the script
https://github.com/all-of-us/workbench/pull/2733 for the output [WIP]

gjuggler:
Another entrant: Foo -> FooDbModel

jmthibault79:
`WorkspaceTable`, `UserTable`, etc

blrubenstein:
I think I slightly prefer `{prefix}Object` over `Object{suffix}` because I read the prefix as a descriptor of the object type. For example `DbWorkspace` vs `WorkspaceDb`. The first one is an object of type workspace in the db, and the second is a DB that stores a workspace.

That leans me to land more on preferring `Db{object}` over `{Object}DataModel`, but I could be convinced.

jaycarlton:
The other thing about `CohortDbModel` aesthetically is that if we're using an abbreviation, I can read it a lot easier if it's at the beginning. Some of that is coming from experience with other languages though. It's also tough to go word-abbrev-word easily.

On Scout's point, I haven't found a good rule that works 100% of the time. I think it can have as much to do with the ear as anything else, and what kind of function/class/enum we're talking about. For me it's easier to say `CohortDataModel` than `DataModelCohort`, but conversely `DbCohort` is easier than `CohortDb`. (In the special case of "db", we're used to hearing about InnoDB, DynamoDB, etc, so I agree it's confusing at the end.)

What I like about an `Entity` suffix is that it's fun to say, exactly describes the classes we're talking about (in most cases), and is both short but visually distinctive enough to find quickly.

jaycarlton:
> Rename the DAOs as `WorkspaceTable`, `UserTable` etc as a companion to what we choose here?

Maybe...but those aren't confusing anyone right now AFAICT. BTW @jmthibault79 I never got your vote and we need a tiebreaker.

calbach:
Where are votes being tallied and how?

+1 DbWorkspace
+0 WorkspaceEntity
-1 WorkspaceDataModel

jmthibault79:
I think we need to have "Db" in there somewhere, to make it explicit.  "Data" "Entity" and "Model" could mean a variety of things, depending on context.  I also like Scout's prefix semantics argument.

+1 DbWorkspace
+1 WorkspaceDbModel
-1 WorkspaceEntity
-1 WorkspaceDataModel

jaycarlton:
I'll throw my support behind `DbWorkspace`. Thanks!

PR #2755, opened 2019-10-25
[risk=low][no ticket]Update vulnerable Jackson Databind version ... again

jmthibault79:


api/build.gradle
```diff
@@ -179,6 +179,7 @@ buildscript {    // Configuration for building
     GAE_VERSION = '1.9.64'
     SPRING_BOOT_VERSION = '1.5.22.RELEASE'
     JACKSON_VERSION = '2.9.10'
```
calbach: nit: any reason not to just bump it across the board?

jmthibault79: The `.1` bump is a Databind-only security release.

PR #2756, opened 2019-10-25
[risk=no][no ticket] Add run-local-rw-migrations to project.rb

jmthibault79:

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -634,10 +634,9 @@ end
 
 Common.register_command({
   :invocation => "run-local-all-migrations",
-  :description => "Runs local data/schema migrations for cdr/workbench schemas.",
+  :description => "Runs local data/schema migrations for the cdr and workbench schemas.",
   :fn => ->() { run_local_all_migrations() }
 })
-
 
 def run_local_data_migrations()
   init_new_cdr_db %W{--cdr-db-name cdr --run-list data --context local}
@@ -645,8 +644,19 @@ end
 
 Common.register_command({
   :invocation => "run-local-data-migrations",
-  :description => "Runs local data migrations for cdr/workbench schemas.",
+  :description => "Runs local data migrations for the cdr schema.",
   :fn => ->() { run_local_data_migrations() }
+})
+
+def run_local_rw_migrations()
+  common = Common.new
+  common.run_inline %W{docker-compose run db-scripts ./run-migrations.sh main}
+end
+
+Common.register_command({
+  :invocation => "run-local-rw-migrations",
+  :description => "Runs local migrations for the workbench schema.",
+  :fn => ->() { run_local_rw_migrations() }
 })
 
 def make_bq_denormalized_tables(*args)


api/libproject/devstart.rb
```diff
@@ -488,10 +487,22 @@ def connect_to_db()
...
+  common = Common.new
+
+  cmd = "MYSQL_PWD=root-notasecret mysql --database=cdr"
+  common.run_inline %W{docker-compose exec db sh -c #{cmd}}
+end
+
+Common.register_command({
+  :invocation => "connect-to-cdr-db",
+  :description => "Connect to the running CDR database via mysql.",
+  :fn => ->() { connect_to_cdr_db() }
```
calbach: I think a flag would be better to avoid command proliferation. Or just use the existing command:

```
./project connect-to-db
USE cdr;
```

api/libproject/devstart.rb
```diff
@@ -634,21 +645,31 @@ def run_local_all_migrations()
...
 })
 
+def run_local_rw_migrations()
+  common = Common.new
+  common.run_inline %W{docker-compose run db-scripts ./run-migrations.sh main}
+end
+
+Common.register_command({
+  :invocation => "run-local-rw-migrations",
+  :description => "Runs local data migrations for the workbench schema.",
```
calbach: nit: I would drop the word "data" here, I think it makes more sense for the CDR migrations since they are actually mostly just populating data.

api/libproject/devstart.rb
```diff
@@ -634,21 +645,31 @@ def run_local_all_migrations()
...
   :fn => ->() { run_local_data_migrations() }
 })
 
+def run_local_rw_migrations()
+  common = Common.new
+  common.run_inline %W{docker-compose run db-scripts ./run-migrations.sh main}
+end
+
+Common.register_command({
+  :invocation => "run-local-rw-migrations",
```
calbach: To be clear, `dev-up` already does just this. It might be nice to have this option without a full dev-up, but under normal circumstances in the dev flow you shouldn't need to use `run-local-all-migrations` unless you're developing cohort builder.

api/libproject/devstart.rb
jmthibault79: ok added a param

api/libproject/devstart.rb
jmthibault79: Right: this is only to speed up the non-dev-up by making it possible to run workbench liquibase updates without waiting ~6 minutes for an unnecessary CDR migration

api/libproject/devstart.rb
jmthibault79: ok

api/libproject/devstart.rb
```diff
@@ -478,21 +478,30 @@ def run_gradle(cmd_name, args)
   :fn => ->(*args) { run_gradle("gradle", args) }
 })
 
+def connect_to_db(*args)
+  if args.empty?
```
calbach: This breaks as soon as you add a second flag, here is the preferred way to set a default: https://github.com/all-of-us/workbench/blob/0ca9ec7874214717a3e083b9c4c5b4faaf76c729/api/libproject/devstart.rb#L1042

api/libproject/devstart.rb
```diff
@@ -478,21 +478,30 @@ def run_gradle(cmd_name, args)
...
+    args = "--database=workbench"
+  end
 
-def connect_to_db()
-  common = Common.new
+  op = WbOptionsParser.new("connect-to-db", *args)
+  op.add_option(
+      "--database [database]",
```
calbach: Is this better than just typing `use cdr;` after getting into the SQL prompt?

api/libproject/devstart.rb
```diff
@@ -478,21 +478,30 @@ def run_gradle(cmd_name, args)
...
+      ->(opts, v) { opts.database = v},
+      "Database to connect to"
```
calbach: Probably list values here and/or validate incoming values

api/libproject/devstart.rb
jmthibault79: Maybe not.  Think I'll just drop this idea.

api/libproject/devstart.rb
jmthibault79: thanks

PR #2759, opened 2019-10-25
[risk=low][RW-3780] Fix Dataset Export to Notebook bug

jmthibault79:

### DIFF
--- a/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
+++ b/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
@@ -129,9 +129,8 @@ class ExportDataSetModal extends React.Component<
           kernelType: this.state.kernelType
         });
       // Open notebook in a new tab and close the modal
-      const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}` +
-        `/notebooks/preview/${appendNotebookFileSuffix(
-          encodeURIComponentStrict(this.state.notebookName))}`;
+      const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}/notebooks/preview/` +
+        appendNotebookFileSuffix(encodeURIComponentStrict(this.state.notebookName));
       navigateByUrl(notebookUrl);
       this.props.closeFunction();
     } catch (error) {
--- a/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
+++ b/ui/src/app/pages/data/data-set/export-data-set-modal.tsx
@@ -128,8 +128,9 @@ class ExportDataSetModal extends React.Component<
         kernelType: this.state.kernelType
       });
     // Open notebook in a new tab and close the modal
-    const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}/notebooks/
-      preview/${appendNotebookFileSuffix(encodeURIComponentStrict(this.state.notebookName))}`;
+    const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}` +
+      `/notebooks/preview/${appendNotebookFileSuffix(
+        encodeURIComponentStrict(this.state.notebookName))}`;
     navigateByUrl(notebookUrl);
     this.props.closeFunction();
   }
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -119,19 +119,28 @@ class ExportDataSetModal extends React.Component<
...
-    await dataSetApi().exportToNotebook(
-      workspaceNamespace, workspaceFirecloudName,
-      {
-        dataSetRequest: request,
-        notebookName: this.state.notebookName,
-        newNotebook: this.state.newNotebook,
-        kernelType: this.state.kernelType
-      });
-    // Open notebook in a new tab and close the modal
-    const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}/notebooks/
```
jmthibault79: This was the bug: a linebreak inside the URL string

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -119,19 +119,28 @@ class ExportDataSetModal extends React.Component<
...
+          newNotebook: this.state.newNotebook,
+          kernelType: this.state.kernelType
+        });
+      // Open notebook in a new tab and close the modal
+      const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}` +
+        `/notebooks/preview/${appendNotebookFileSuffix(
+          encodeURIComponentStrict(this.state.notebookName))}`;
+      navigateByUrl(notebookUrl);
+      this.props.closeFunction();
+    } catch (error) {
```
jmthibault79: This is a different bug: https://precisionmedicineinitiative.atlassian.net/browse/RW-3782 caused an infinite spinner in an uncancellable modal here.  The bug remains but now the user is not prevented from cancelling out of this modal and going back to other work

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -119,19 +119,28 @@ class ExportDataSetModal extends React.Component<
...
+      // https://precisionmedicineinitiative.atlassian.net/browse/RW-3782
+      // The exportToNotebook call can fail with a 400.  Catch that error to ensure the user can exit the modal.
+      // TODO: better failure UX
```
jmthibault79: Some quick ideas at EOW, for further discussion:

We don't have a real workaround for the user, but we could give them more information with some error text.

We could also close the modal.  But maybe there is a possibility of success later?  Keeping it open lets the user try again.

calbach: Keeping the modal open and displaying an error message somewhere on screen (e.g. at the top) is probably the ideal flow here. We'd want to develop standard components / UI patterns to support that though.

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -119,19 +119,28 @@ class ExportDataSetModal extends React.Component<
...
+        workspaceNamespace, workspaceFirecloudName,
+        {
+          dataSetRequest: request,
+          notebookName: this.state.notebookName,
+          newNotebook: this.state.newNotebook,
+          kernelType: this.state.kernelType
+        });
+      // Open notebook in a new tab and close the modal
+      const notebookUrl = `/workspaces/${workspaceNamespace}/${workspaceFirecloudName}` +
+        `/notebooks/preview/${appendNotebookFileSuffix(
```
calbach: This makes me nervous, though I think it's probably fine. Would probably unwrap this line, or just use concatenation instead of templating in this case.

jmthibault79: I can clean this up a bit.

jmthibault79:
Snyk seems to be confused.  I actually just merged a PR to upgrade us out of that vulnerability: #2755

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
jmthibault79: Agreed, as a longer term improvement outside this PR.

PR #2760, opened 2019-10-28
[RW-3316][risk=no] Clear notebook load timer on navigate

calbach:


jmthibault79:
I'm not seeing where `runClusterTimer` is triggered - is that elsewhere?

calbach:
The timer just tracks a pending timeout, which gets invoked via `retry()`, which gets called here: https://github.com/all-of-us/workbench/pull/2760/files#diff-2ea8624017c9c5dbbed2d67d1d2e6c83L150-L154

calbach:
i.e. `runClusterTimer` is not a function, but an ID for a `setTimeout` so that we can clear it later.

jmthibault79:
got it, thx

PR #2761, opened 2019-10-28
[RW-3779][risk=no] Add a WorkbenchConfig field to control rate of buffer refill

gjuggler:
See the RW ticket for more context: https://precisionmedicineinitiative.atlassian.net/browse/RW-3779

This PR adjusts the cron endpoint to become pluralized and tunable via a WorkbenchConfig environment variable.

The intent is to give us a knob to be able to tune up the perf environment's refill rate, since we've been encountering times when testing during the day isn't possible due to an empty buffer.

This kind of tuning may also help us in a production scenario – we currently can't quickly refill the buffer if a need arose.

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
+++ b/common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
@@ -48,6 +48,7 @@ public class WorkbenchConfig {
   public static class BillingConfig {
     public Integer retryCount;
     public Integer bufferCapacity;
+    public Integer bufferRefillProjectsPerTask;
     public String projectNamePrefix;
     public String accountId;
     public String exportBigQueryTable;
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -20,6 +20,7 @@
     "exportBigQueryTable": "all-of-us-workbench-test-bd.billing_data.gcp_billing_export_v1_014D91_FCB792_33D2C0",
     "retryCount": 2,
     "bufferCapacity": 2,
+    "bufferRefillProjectsPerTask": 1,
     "defaultFreeCreditsLimit": 300.0,
     "garbageCollectionUserCapacity": 1000,
     "garbageCollectionUsers": [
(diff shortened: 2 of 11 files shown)


common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -45,13 +45,36 @@ public static WorkbenchConfig createEmptyConfig() {
...
+    // any new projects when the total number of in-progress & ready projects is at or above this
+    // level.
     public Integer bufferCapacity;
+    // The number of times to attempt project creation per cron task execution. This effectively
+    // controls the max rate of project refill. If the cron task is configured to run once per
+    // minute and this param is set to 5, then the buffer system will create up to 5 projects
+    // per minute.
+    //
+    // Per guidance from Google Cloud's project infrastructure team, we should limit our total rate
+    // of project creation to a number less than 1 per section. In practice, a reasonable aggressive
```
jmthibault79: Did you mean "second" here?

jmthibault79: Thanks for adding this guidance as a comment

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -45,13 +45,36 @@ public static WorkbenchConfig createEmptyConfig() {
...
   public static class BillingConfig {
+    // This config variable seems to be unused.
     public Integer retryCount;
+    // The total capacity of the GCP project buffer. The buffering system will not attempt to create
+    // any new projects when the total number of in-progress & ready projects is at or above this
+    // level.
     public Integer bufferCapacity;
+    // The number of times to attempt project creation per cron task execution. This effectively
+    // controls the max rate of project refill. If the cron task is configured to run once per
+    // minute and this param is set to 5, then the buffer system will create up to 5 projects
```
calbach: I looked up the GAE cron overrun policy here, looks like `every 1 minute` will wait 1 minute after each run, which is what we want (overruns would be bad) but requires adjusting your math a bit. It's not quite 5 / minute, more like N per (1m + N * mean_create_bp_latency). Would be OK just saying "approximately" here.

gjuggler: Good point – I added a small clarification to the comment.

PR #2762, opened 2019-10-28
[RW-3738][risk=no] Update prod featured workspaces

calbach:
Notes:
- I haven't published these in prod yet, looking at that next (this just needs to happen before this release goes to prod).
- "name" looks like it's probably unused, cleanup task: https://precisionmedicineinitiative.atlassian.net/browse/RW-3787


calbach:
Published all 4, will unpublish the two others after the release.

PR #2766, opened 2019-10-28
[RW-3784][risk=no] Remove extra "get" call from cloneWorkspace

gjuggler:
This makes a performance-focused optimization, to directly store & use the response object from a Terra cloneWorkspace call.

I had to finagle the API docs a bit, to pretend that get/list/clone/create Workspace all share the same underlying Workspace API model schema. As I understand from David An, this isn't strictly correct, but in practice the schemas are similar enough for our current usage.

David's team is tracking API doc fix-ups in https://broadworkbench.atlassian.net/browse/AS-21. I've set a watcher on that ticket, and we should do a one-off re-import of their Swagger once we're confident the API docs match reality.

TESTED: manually tested on a local env by duplicating a workspace as an owner. Seemed to work fine.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
@@ -69,7 +69,7 @@ public interface FireCloudService {
   /** Creates a new FC workspace. */
   Workspace createWorkspace(String projectName, String workspaceName);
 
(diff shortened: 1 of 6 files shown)


gjuggler:
> This is being reviewed from my phone, so I haven’t been able to run it, but it looks good. Just to confirm, have you manually tested duplicating a workspace on local?

Not yet – thanks for calling that out. I'll update description with a TESTED line once I've done that.

PR #2767, opened 2019-10-28
[RW-3438][risk=no] Close notebook activity monitor after navigating away

calbach:
Remove malfunctioning notebook preview activity monitor.

Two issues here:
1. This activity monitor is intended to handle user interactions with the notebooks preview, which lives in an iframe, to extend the activity timeout. When the user navigates away from this page, that element is no longer rendered, and yet the activity monitor listener persists.
2. In actuality, this implementation does not appear to function as expected. Instead of setting up a monitor on the iframe, it's actually adding another monitor to the top-level window. This is redundant to the work done at the top-level here: https://github.com/all-of-us/workbench/blob/master/ui/src/app/pages/signed-in/component.ts#L140 , but also accumulates error-generating event callbacks everytime the user visits a preview page. All callbacks get triggered on user input/interaction, causing massive numbers of errors, and actually slowing down the browser due to our stack trace generation attempt and send off to Stackdriver for reporting.

I strongly suspect injecting an activity monitor into the iframe itself would be a fairly tricky task to achieve in a non-hacky way; we'd need to execute Javascript within the iframe document source. So this PR removes the functionality for now since it causes lots of error spam and performance degradation.


calbach:
Bump - this is causing tons of logspam in production, and I think it actually causes significant browser slowdown.

calbach:
Thanks for asking. On consideration, I realize this is doing something a bit surprising, and it's probably not the right fix. The notebooks iframe is actually not on this page, so I think this is setting up a listener for later. Will sync with @ericsong on this, but this may require https://precisionmedicineinitiative.atlassian.net/browse/RW-1894 for the full fix.

calbach:
OK, discussed a bit with @ericsong , he is going to look at this on Monday for an easy fix-forward, else we should merge this in for next week's release. I was mistaken in my last comment, and found another critical issue here upon further investigation. See the PR description.

ericsong:
Going to merge to unblock Monday's release.

calbach:
I repurposed the attached ticket to track adding back the activity monitor in the preview frame: https://precisionmedicineinitiative.atlassian.net/browse/RW-3438

I think it's lower priority than these other activity issues though:
- https://precisionmedicineinitiative.atlassian.net/browse/RW-3554
- https://precisionmedicineinitiative.atlassian.net/browse/RW-3555

PR #2769, opened 2019-10-29
[RW-3586][risk=no] Populate CDR version as env vars in notebooks

als364:
Using the same variables that we use in code snippets - `CDR_VERSION_CLOUD_PROJECT` and `CDR_VERSION_BIGQUERY_DATASET`.
Screen Shot 2019-10-29 at 10 02 36 AM


api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -70,6 +80,18 @@ private ClusterRequest createFirecloudClusterRequest(
     nbExtensions.put(
         "aou-activity-checker-extension", gcsPrefix + "/activity-checker-extension.js");
 
+    Workspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceName);
+    Map<String, String> customClusterEnvironmentVariables = new HashMap<>();
+    // i.e. is NEW or MIGRATED
+    if (!workspace.getBillingMigrationStatusEnum().equals(BillingMigrationStatus.OLD)) {
```
als364: @calbach here is what I did for setting / not setting env vars based on migration status

PR #2781, opened 2019-10-30
[RW-3692][risk=no] Add test for clone data set 

ericsong:


api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
```diff
@@ -40,4 +42,8 @@ DataSet cloneDataSetToWorkspace(
       DataSet fromDataSet, Workspace toWorkspace, Set<Long> cohortIds, Set<Long> conceptSetIds);
 
   List<DataSet> getDataSets(Workspace workspace);
+
+  List<ConceptSet> getConceptSets(DataSet dataSet);
```
als364: Why is this in `DataSetService` instead of the `ConceptSetService`?

If must be in `DataSetService`, why is it not called `getConceptSetsForDataSet`?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
```diff
@@ -40,4 +42,8 @@ DataSet cloneDataSetToWorkspace(
...
+
+  List<Cohort> getCohorts(DataSet dataSet);
```
als364: Same question as above.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -1242,6 +1254,121 @@ public void testCloneWorkspaceWithConceptSetNewCdrVersionNewConceptSetCount() th
...
+
+    org.pmiops.workbench.db.model.Workspace clonedDbWorkspace =
+        workspaceDao.findByWorkspaceNamespaceAndFirecloudNameAndActiveStatus(
+            cloned.getNamespace(),
+            cloned.getId(),
+            StorageEnums.workspaceActiveStatusToStorage(WorkspaceActiveStatus.ACTIVE));
+
+    List<DataSet> dataSets = dataSetService.getDataSets(clonedDbWorkspace);
+    assertThat(dataSets).hasSize(1);
+    assertThat(dataSets.get(0).getName()).isEqualTo("data set name");
```
als364: Slight preference for factoring all the magic strings out to the top of the function

api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
als364: Even in `ConceptSetService` I would consider calling it `getConceptSetsForDataSet`, come to think of it, and the other gets `getConceptSetsForWhateverTheArgumentIs`

ericsong: My reasoning was that the `DataSet` model is the one that maintains the relationship to the `ConceptSet` and `Cohort` models so its service should be responsible for performing the fetch.

Neither `ConceptSet` or `Cohort` has any reference to `Dataset` so I view them as a base level of functionality on top of which `Dataset` has been built.

I opted for the shorter function names because I felt like the function signature captures the `ForDataset` context.

Ex. For me, I gain the same context from reading `getConceptSets(dataSet)` as I do from `getConceptSetsForDataset(dataSet)`.

I am cool with changing them to the longer version though.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
ericsong: Will do

api/src/main/java/org/pmiops/workbench/db/dao/DataSetService.java
als364: Thanks for the explanation, that makes sense.

PR #2784, opened 2019-10-30
fix spotless

jaycarlton:
forgot spotless


jaycarlton:
Since this is a build fix and since CircleCI is backed up, got the go-ahead from the TL to merge first and ask questions later.

PR #2786, opened 2019-10-30
[RW-3686][Risk=moderate] Upgrade Jupyter docker

blrubenstein:
This upgrades us to the new Jupyter Docker image. We are using the 'GATK' image, because it has up to date python and R. Manually tested in both R and Python by exporting a dataset, running both queries generated, and trying setup and a selection of other snippets. (All of the AoU specific snippets, unless I noticed a pattern that broke them, like with measurement_ext)

### DIFF
--- a/api/cluster-resources/setup_notebook_cluster.sh
+++ b/api/cluster-resources/setup_notebook_cluster.sh
@@ -39,12 +39,10 @@ apt-get update
 apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6
 apt-get install -y --no-install-recommends jq
 
-for v in "3"; do
-  "pip${v}" install --upgrade \
-    plotnine \
-    'https://github.com/all-of-us/pyclient/archive/pyclient-v1-17.zip#egg=aou_workbench_client&subdirectory=py' \
-    statsmodels==0.10.0rc2
-done
+"pip3" install --upgrade \
+  plotnine \
+  'https://github.com/all-of-us/pyclient/archive/pyclient-v1-17.zip#egg=aou_workbench_client&subdirectory=py' \
+  statsmodels==0.10.0rc2
 
 # Install wondershaper for basic egress throttling.
 apt-get install -y --no-install-recommends iproute2
--- a/api/cluster-resources/setup_notebook_cluster.sh
+++ b/api/cluster-resources/setup_notebook_cluster.sh
@@ -36,16 +36,17 @@ echo "Sys.setenv(RETICULATE_PYTHON = '$(which python3)')" >> ~/.Rprofile
 # TODO: These steps are quite slow, we could also consider pushing this into the
 # base Leo image or a custom AoU docker image.
 apt-get update
-apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6 jq
+apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6
+apt-get install -y --no-install-recommends jq
 
-for v in "2.7" "3"; do
+for v in "3"; do
   "pip${v}" install --upgrade \
     plotnine \
     'https://github.com/all-of-us/pyclient/archive/pyclient-v1-17.zip#egg=aou_workbench_client&subdirectory=py'
 done
 
 # Install wondershaper for basic egress throttling.
-apt-get install -y --no-install-recommends iproute
+apt-get install -y --no-install-recommends iproute2
 (cd /usr/local/share &&
  git clone https://github.com/magnific0/wondershaper.git &&
  cd wondershaper &&
--- a/api/cluster-resources/setup_notebook_cluster.sh
+++ b/api/cluster-resources/setup_notebook_cluster.sh
@@ -36,7 +36,7 @@ echo "Sys.setenv(RETICULATE_PYTHON = '$(which python3)')" >> ~/.Rprofile
 # TODO: These steps are quite slow, we could also consider pushing this into the
 # base Leo image or a custom AoU docker image.
 apt-get update
-apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6
+apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6 jq
 
 for v in "2.7" "3"; do
   "pip${v}" install --upgrade \
--- a/api/cluster-resources/setup_notebook_cluster.sh
+++ b/api/cluster-resources/setup_notebook_cluster.sh
@@ -42,7 +42,8 @@ apt-get install -y --no-install-recommends jq
 for v in "3"; do
   "pip${v}" install --upgrade \
     plotnine \
-    'https://github.com/all-of-us/pyclient/archive/pyclient-v1-17.zip#egg=aou_workbench_client&subdirectory=py'
+    'https://github.com/all-of-us/pyclient/archive/pyclient-v1-17.zip#egg=aou_workbench_client&subdirectory=py' \
+    statsmodels==0.10.0rc2
 done
 
 # Install wondershaper for basic egress throttling.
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -80,7 +81,7 @@ private ClusterRequest createFirecloudClusterRequest(
     nbExtensions.put(
         "aou-activity-checker-extension", gcsPrefix + "/activity-checker-extension.js");
 
-    Workspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceName);
+    Workspace workspace = workspaceService.getByName(workspaceNamespace, workspaceName);
```
blrubenstein: This is an issue I noticed, `getRequired` uses the FC name, but we were passing up the AoU name. This is a piece of tech debt that we definitely should tackle in the near future, because multiple things named almost the same is a pain in the neck.

blrubenstein:
In python the only error was the 'summarize a dataset' snippet, with this error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-10-9022e5298342> in <module>
      5 
      6 ## -----[ CHANGE THE DATAFRAME NAME(S) TO MATCH YOURS FROM DATASET BUILDER] -----
----> 7 dataset_16528939_person_df.loc[:10000,:].profile_report()  # Examine up to the first 10,000 rows. Larger
      8                                                             # dataframes can be profiled, but it takes more time.

/usr/local/lib/python3.7/dist-packages/pandas/core/generic.py in __getattr__(self, name)
   5177             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5178                 return self[name]
-> 5179             return object.__getattribute__(self, name)
   5180 
   5181     def __setattr__(self, name, value):

AttributeError: 'DataFrame' object has no attribute 'profile_report'
```

I also saw several unrelated errors for the `measurements of interest` snippets.

blrubenstein:
From R, I saw a number of other errors, unsure if related. Measurements of interest had many of the same errors as python, but I also saw this for the participant count snippets: 
```
Error: Table name "person" missing dataset while no default dataset is set in the request. [invalid]
Traceback:

1. bq_table_download(bq_project_query(BILLING_PROJECT_ID, page_size = 25000, 
 .     query = str_glue("\n-- Compute the count of unique participants in our All of Us cohort.\nSELECT\n  COUNT(DISTINCT person_id) AS total_number_of_participants\nFROM\n  `{CDR}.person`\nWHERE\n  person_id IN ({COHORT_QUERY})\n")))
2. as_bq_table(x)
3. bq_project_query(BILLING_PROJECT_ID, page_size = 25000, query = str_glue("\n-- Compute the count of unique participants in our All of Us cohort.\nSELECT\n  COUNT(DISTINCT person_id) AS total_number_of_participants\nFROM\n  `{CDR}.person`\nWHERE\n  person_id IN ({COHORT_QUERY})\n"))
4. bq_job_wait(job, quiet = quiet)
5. signal_reason(err$reason, err$message)
```

blrubenstein:
Found that the `jq` library isn't installed anymore, which is the cause of the R breakages. Looking into resolving that.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
jmthibault79: Yikes!  I'm seeing errors locally which might be caused by this, and are driving me nuts.  I'll take a poke at this.

calbach: Ah - that's not good. We should fix this in the other direction. We want to use the firecloud name here; the rest of the stack should pass the correct value down.

jmthibault79: Actually nope.  This can of worms is much larger than I'd guessed.

jmthibault79: https://precisionmedicineinitiative.atlassian.net/browse/RW-3890

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -36,6 +36,8 @@
   private static final String CLUSTER_LABEL_AOU = "all-of-us";
   private static final String CLUSTER_LABEL_CREATED_BY = "created-by";
   private static final String WORKSPACE_CDR = "WORKSPACE_CDR";
+  private static final String JUPYTER_DOCKER_IMAGE =
```
ericsong: nit: I can see this being a value that should be on our WorkbenchConfig. It would let us change the docker image without deploys if needed and there's also the benefit of making it easier to find. I can see this being a value that the researcher team or support team would be interested in knowing.

ericsong:
Also curious, how do we verify that this is actually using the new docker image?

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
blrubenstein: Agreed we should fix it in the other direction, I just picked the easiest possible fix. Thank you for filing Joel.

jmthibault79: I'm fixing this line in #2816 anyway

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
jaycarlton: +1 We may even eventually want to support multiple images that we decide among when launching a cluster based on user, free tier vs billed, etc.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -36,6 +36,8 @@
   private static final String CLUSTER_LABEL_AOU = "all-of-us";
   private static final String CLUSTER_LABEL_CREATED_BY = "created-by";
   private static final String WORKSPACE_CDR = "WORKSPACE_CDR";
```
jaycarlton: nit, and maybe a separate PR: I'd really like to pull out the environment variable names into an enum class, so we'd have `NotebookEnvironmentVariables.CDR_BQ_DATASET`. There's also a CDR_VERSION somewhere that could go in there. 

If you don't want to do that here I can do it in a separate PR.

blrubenstein: I'm going to say separate PR, because this is an existing problem. (I think it showed up in green because rebase.)

api/cluster-resources/setup_notebook_cluster.sh
```diff
@@ -37,13 +37,14 @@ echo "Sys.setenv(RETICULATE_PYTHON = '$(which python3)')" >> ~/.Rprofile
 # base Leo image or a custom AoU docker image.
 apt-get update
 apt-get -t stretch-cran35 install -y --no-install-recommends libmagick++-dev gfortran-6
+apt-get install -y --no-install-recommends jq
 
-for v in "2.7" "3"; do
+for v in "3"; do
```
ericsong: should we just drop the for loop since there's only 1 element now?

blrubenstein: I thought about that, sure. Sounds reasonable.

blrubenstein:
Removing the for loop maybe broke plot9, so I think I had the pip syntax wrong? (Did it need the quotes around it for some reason?).

Pushing this up in case someone else wants to bang on it for a while this evening. I am heading out for the evening.

PR #2787, opened 2019-10-30
[no ticket][risk=no] Change CDR env var name

als364:
Per conversation with Nicole Deflaux:
```
My recommendation would be to keep it simple: one new environment variable WORKSPACE_CDR where is value would be both the project id and the dataset name such as   'fc-aou-cdr-prod.R2019Q4R3'.

The only use case that comes to mind where I want the BigQuery project *only* is if I am programmatically trying to list all the BigQuery datasets in a particular GCP project. AoU curators might want to do this on occasion, but they know how to make that happen without the convenience of the environment variable.
```

Screen Shot 2019-10-30 at 4 57 35 PM

There's an extra change here because Spotless has some opinions.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -15,7 +15,6 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
 import org.json.JSONObject;
-import org.pmiops.workbench.WorkbenchConstants;
 import org.pmiops.workbench.annotations.AuthorityRequired;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.UserDao;
@@ -53,6 +52,8 @@ public class ClusterController implements ClusterApiDelegate {
   private static final String WORKSPACE_ID_KEY = "WORKSPACE_ID";
   private static final String API_HOST_KEY = "API_HOST";
   private static final String BUCKET_NAME_KEY = "BUCKET_NAME";
+  private static final String CDR_VERSION_CLOUD_PROJECT = "CDR_VERSION_CLOUD_PROJECT";
+  private static final String CDR_VERSION_BIGQUERY_DATASET = "CDR_VERSION_BIGQUERY_DATASET";
   // The billing project to use for the analysis.
   private static final String BILLING_CLOUD_PROJECT = "BILLING_CLOUD_PROJECT";
   private static final String DATA_URI_PREFIX = "data:application/json;base64,";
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -53,6 +52,8 @@
   private static final String WORKSPACE_ID_KEY = "WORKSPACE_ID";
   private static final String API_HOST_KEY = "API_HOST";
   private static final String BUCKET_NAME_KEY = "BUCKET_NAME";
+  private static final String CDR_VERSION_CLOUD_PROJECT = "CDR_VERSION_CLOUD_PROJECT";
```
jaycarlton: I think this is exactly right how you have it, but to double-check, should any of these be in a Config file?

api/src/main/java/org/pmiops/workbench/audit/targetproperties/WorkspaceTargetProperty.java
```diff
@@ -60,18 +60,19 @@ public String extract(Workspace workspace) {
       Workspace previousWorkspace, Workspace newWorkspace) {
     ImmutableMap.Builder<String, PreviousNewValuePair> resultBuilder = new Builder<>();
     for (WorkspaceTargetProperty property : WorkspaceTargetProperty.values()) {
-      final Optional<String> previousValue = Optional.ofNullable(property.extract(previousWorkspace));
+      final Optional<String> previousValue =
```
jaycarlton: nit: If you pull master and merge, this should go away

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -85,11 +85,10 @@ private ClusterRequest createFirecloudClusterRequest(
     // i.e. is NEW or MIGRATED
     if (!workspace.getBillingMigrationStatusEnum().equals(BillingMigrationStatus.OLD)) {
       customClusterEnvironmentVariables.put(
-          WorkbenchConstants.CDR_VERSION_CLOUD_PROJECT,
-          workspace.getCdrVersion().getBigqueryProject());
-      customClusterEnvironmentVariables.put(
-          WorkbenchConstants.CDR_VERSION_BIGQUERY_DATASET,
-          workspace.getCdrVersion().getBigqueryDataset());
+          WORKSPACE_CDR,
+          workspace.getCdrVersion().getBigqueryProject()
```
jaycarlton: nit: I feel like `String.format()`, or better yet a function to build the variable value, would be clearer.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -36,6 +35,7 @@
 
   private static final String CLUSTER_LABEL_AOU = "all-of-us";
   private static final String CLUSTER_LABEL_CREATED_BY = "created-by";
+  private static final String WORKSPACE_CDR = "WORKSPACE_CDR";
```
jaycarlton: nit: Can you make a little enum class for the env var names? Then you could have `NotebookEnvironmentVariables.WORKSPACE_CDR` etc for clarity. I'd do this even if there's only one and you're only using it once.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
als364: This was a reversion of a previous change where I'd factored these variables out for reuse.

jaycarlton: OK. There are also .properties files, if you need more than about this many. no worries.

PR #2788, opened 2019-10-30
[RW-3825][risk=no] Add legacy prod data dictionary files

ericsong:
Main thing to check is that the `meta_data.cdr_version` field matches the names in the `cdr_version` table in production.

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/LoadDataDictionary.java
@@ -8,8 +8,12 @@ import java.io.InputStream;
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


gjuggler:
> Main thing to check is that the `meta_data.cdr_version` field matches the names in the `cdr_version` table in production.

I wonder if this is something we could add as a unit test and/or deployment check – e.g. refuse to deploy unless we have dictionary data for all CDR versions in our cdr_versions config file for a given environment. Maybe not worth the effort though?

ericsong:
You're telling me you didn't read my 22,000 line PR ??

Jokes aside, testing locally wouldn't really work because the files are specific to the prod environment. I like the idea of adding a check though. Should be easy enough to add.

gjuggler:
> You're telling me you didn't read my 22,000 line PR ??
> 
> Jokes aside, testing locally wouldn't really work because the files are specific to the prod environment. I like the idea of adding a check though. Should be easy enough to add.

Nice addition – thanks for the follow-up!

PR #2794, opened 2019-10-31
Audit Code to Kotlin

jaycarlton:
Trying out Kotlin converter on my audit code. I figure this is a good opportunity b/c I know the code, it's not in the critical path, and I'm going to be working with it for a while.

I was able to get 85% or so automatically, and it would've been higher had I not had fancy things like ImmutableList and Mockito.

I'm bummed about the delete & rename semantics in the diffs. I tried re-staging the commits on another branch, and got them to show up locally as a rename initially, but on applying the changes it was just too much of a difference for most of them. If you want, I can make gists with versions for the before & after and then diff those, but we still won't have edit history.

I can set a diff threshold locally, though I'm not sure about applying it in github...


jaycarlton:
Rebase and force push to get around [this issue](https://pmi-engteam.slack.com/archives/C7JUTPQLE/p1572632580006200).

api/src/main/java/org/pmiops/workbench/audit/ActionAuditServiceImpl.kt
```diff
@@ -0,0 +1,70 @@
...
+        jsonValuesByColumn[AuditColumn.TARGET_TYPE.name] = auditEvent.targetType
+        jsonValuesByColumn[AuditColumn.TARGET_ID.name] = auditEvent.targetIdMaybe
+        jsonValuesByColumn[AuditColumn.AGENT_ID.name] = auditEvent.agentId
+        jsonValuesByColumn[AuditColumn.TARGET_PROPERTY.name] = auditEvent.targetPropertyMaybe
+        jsonValuesByColumn[AuditColumn.PREV_VALUE.name] = auditEvent.previousValueMaybe
+        jsonValuesByColumn[AuditColumn.NEW_VALUE.name] = auditEvent.newValueMaybe
+        return JsonPayload.of(jsonValuesByColumn)
+    }
+
+    companion object {
```
ericsong: i think these usually go to the top of the class?

api/src/main/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceImpl.kt
```diff
@@ -0,0 +1,183 @@
...
+import org.pmiops.workbench.audit.TargetType
+import org.pmiops.workbench.audit.targetproperties.AclTargetProperty
+import org.pmiops.workbench.audit.targetproperties.WorkspaceTargetProperty
+import org.pmiops.workbench.db.model.User
+import org.pmiops.workbench.model.Workspace
+import org.pmiops.workbench.workspaces.WorkspaceConversionUtils
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
+
+@VisibleForTesting
```
ericsong: why do we need this annotation?

api/src/main/java/org/pmiops/workbench/audit/targetproperties/PreviousNewValuePair.kt
```diff
@@ -0,0 +1,21 @@
...
+
+class PreviousNewValuePair constructor(var previousValue: String?, var newValue: String?) {
+
+    fun valueChanged(): Boolean {
+        return previousValue == null && newValue != null ||
+                newValue == null && previousValue != null ||
+                previousValue != newValue
+    }
+
+    override fun toString(): String {
```
ericsong: I think data class will give this to you for free

api/src/main/java/org/pmiops/workbench/audit/targetproperties/WorkspaceTargetProperty.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.function.Function
+import org.pmiops.workbench.model.Workspace
+
+internal data class WorkspaceTargetPropertyNameValue(val propertyName: String, val value: String)
+
+// N.B. entries will rarely be referred to by name, but are accessed via values(). I.e. they are
+// not safe to remove, even if an IDE indicates otherwise.
+enum class WorkspaceTargetProperty private constructor(val propertyName: String, private val extractor: Function<Workspace, String>) {
+    NAME(propertyName = "name", extractor = Function<Workspace, String> { it.name }),
```
ericsong: are the named parameters necessary? This class is a little hard to read since there's a lot going on. I think cutting that out could help.

api/src/main/java/org/pmiops/workbench/audit/ActionAuditEvent.kt
```diff
@@ -0,0 +1,22 @@
...
+    val actionType: ActionType,
+    val targetType: TargetType,
+    val targetPropertyMaybe: String? = null,
+    val targetIdMaybe: Long? = null,
+    val previousValueMaybe: String? = null,
+    val newValueMaybe: String? = null
+) {
+
+    companion object {
+        fun newActionId(): String = randomUUID().toString()
```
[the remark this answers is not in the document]
jaycarlton: I had it on the service at one point and moved it to the data class. I can just put it on the ActionAuditService interface. No static methods on interfaces, but I can implement it right on the interface as if it were an instance method, and the only contexts that need this already have that service injected.

api/src/main/java/org/pmiops/workbench/audit/ActionAuditServiceImpl.kt
jaycarlton: I'm not opposed to that. Do we have a style guide somewhere? I think spotless is doing things but I'm not actually sure.

api/src/main/java/org/pmiops/workbench/audit/adapters/WorkspaceAuditAdapterServiceImpl.kt
jaycarlton: oh, I made it open for testing purposes I think, and there was no `@OpenForTesting`. Which doesn't make enough sense to add more clarity than it removes, so I'll remove it.

api/src/main/java/org/pmiops/workbench/audit/ActionAuditEvent.kt
```diff
@@ -0,0 +1,22 @@
...
+    val agentType: AgentType,
+    val agentId: Long,
+    val agentEmailMaybe: String?,
+    val actionId: String,
+    val actionType: ActionType,
+    val targetType: TargetType,
+    val targetPropertyMaybe: String? = null,
+    val targetIdMaybe: Long? = null,
+    val previousValueMaybe: String? = null,
+    val newValueMaybe: String? = null
```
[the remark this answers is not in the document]
jaycarlton: I always used to name my optionals with maybe, but that hasn't caught on here. Kotlin will remind you at compile time if you forget I suppose, so I guess I should embrace it.

I'd like to do that rename on my follow-on branch to avoid clashes though.

jaycarlton:
Overriding Codacy checks b/c Kotlin config isn't what we want.

PR #2795, opened 2019-10-31
[RW-3782][RISK:NO]Cannot export Dataset to Notebook because "Concept Sets must contain at least one concept"

NehaBroad:
This was happening because export to notebook model was not sending dataset's Prepackaged concept set values. Therefore datasets with no concept sets apart from prepackaged started throwing error.

### DIFF
--- a/ui/src/app/pages/data/data-set/export-data-set-modal.spec.tsx
+++ b/ui/src/app/pages/data/data-set/export-data-set-modal.spec.tsx
@@ -1,4 +1,4 @@
-import {DataSet, DataSetApi, WorkspacesApi} from "../../../../generated/fetch/api";
+import {DataSet, DataSetApi, PrePackagedConceptSetEnum, WorkspacesApi} from 'generated/fetch/api';
 import {WorkspacesApiStub} from "../../../../testing/stubs/workspaces-api-stub";
 import {DataSetApiStub} from "../../../../testing/stubs/data-set-api-stub";
 import {registerApiClient} from "../../../services/swagger-fetch-clients";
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/data/data-set/export-data-set-modal.spec.tsx
```diff
@@ -1,4 +1,4 @@
-import {DataSet, DataSetApi, WorkspacesApi} from "../../../../generated/fetch/api";
+import {DataSet, DataSetApi, PrePackagedConceptSetEnum, WorkspacesApi} from 'generated/fetch/api';
 import {WorkspacesApiStub} from "../../../../testing/stubs/workspaces-api-stub";
```
als364: Can you get rid of these `../../`s and doublequotes while you're at it as well?

ui/src/app/pages/data/data-set/export-data-set-modal.tsx
```diff
@@ -82,15 +82,22 @@ class ExportDataSetModal extends React.Component<
...
+    return {
       name: dataSet.name,
+      includesAllParticipants: dataSet.includesAllParticipants,
+      description: dataSet.description,
       conceptSetIds: dataSet.conceptSets.map(cs => cs.id),
       cohortIds: dataSet.cohorts.map(c => c.id),
       domainValuePairs: dataSet.domainValuePairs,
-      includesAllParticipants: dataSet.includesAllParticipants,
+      prePackagedConceptSet: dataSet.prePackagedConceptSet
     };
```
jmthibault79: This change removes type info - we should keep that, maybe with `as DataSetRequest` ?

Good to see this consolidation of code.

PR #2796, opened 2019-10-31
[RW-3829][risk=no] Checking ACLs on featured workspaces breaks all the things

als364:
This fixes NPEs in:
- updateRecentWorkspaces
- getRecentWorkspaces
- getCohorts
- getCohortReviews
- getConceptSets
- getDataSets

ACLs for published workspaces don't work by adding an ACL for every user, they work by adding the registered user domain email as READER on the published workspace. We need to check that email as well as the user's email when checking ACLs.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
@@ -3,6 +3,8 @@ package org.pmiops.workbench.workspaces;
 import java.sql.Timestamp;
 import java.util.List;
 import java.util.Map;
+import javax.inject.Provider;
+import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.WorkspaceDao;
 import org.pmiops.workbench.db.model.UserRecentWorkspace;
 import org.pmiops.workbench.db.model.Workspace;
@@ -78,4 +80,8 @@ public interface WorkspaceService {
   UserRecentWorkspace updateRecentWorkspaces(Workspace workspace);
 
   boolean maybeDeleteRecentWorkspace(long workspaceId);
+
+  String getRegisteredUserDomainEmail();
+
+  void setWorkbenchConfigProvider(Provider<WorkbenchConfig> configProvider);
 }
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
@@ -1,5 +1,6 @@
 package org.pmiops.workbench.workspaces;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
 import com.google.gson.Gson;
@@ -22,6 +23,7 @@ import javax.inject.Provider;
 import org.pmiops.workbench.cdr.CdrVersionContext;
 import org.pmiops.workbench.cohorts.CohortCloningService;
 import org.pmiops.workbench.conceptset.ConceptSetService;
+import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.DataSetService;
 import org.pmiops.workbench.db.dao.UserDao;
 import org.pmiops.workbench.db.dao.UserRecentWorkspaceDao;
@@ -39,6 +41,7 @@ import org.pmiops.workbench.exceptions.ForbiddenException;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.firecloud.FireCloudService;
+import org.pmiops.workbench.firecloud.model.ManagedGroupWithMembers;
 import org.pmiops.workbench.firecloud.model.WorkspaceACL;
 import org.pmiops.workbench.firecloud.model.WorkspaceACLUpdate;
 import org.pmiops.workbench.firecloud.model.WorkspaceACLUpdateResponseList;
@@ -72,6 +75,7 @@ public class WorkspaceServiceImpl implements WorkspaceService {
   private DataSetService dataSetService;
   private UserDao userDao;
   private Provider<User> userProvider;
+  private Provider<WorkbenchConfig> workbenchConfigProvider;
   private UserRecentWorkspaceDao userRecentWorkspaceDao;
   private WorkspaceDao workspaceDao;
 
@@ -88,6 +92,7 @@ public class WorkspaceServiceImpl implements WorkspaceService {
       UserDao userDao,
       Provider<User> userProvider,
       UserRecentWorkspaceDao userRecentWorkspaceDao,
+      Provider<WorkbenchConfig> workbenchConfigProvider,
       WorkspaceDao workspaceDao) {
     this.clock = clock;
     this.cohortCloningService = cohortCloningService;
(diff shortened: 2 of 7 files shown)


api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -78,4 +80,8 @@ UserRecentWorkspace updateRecentWorkspaces(
   UserRecentWorkspace updateRecentWorkspaces(Workspace workspace);
 
   boolean maybeDeleteRecentWorkspace(long workspaceId);
+
+  String getRegisteredUserDomainEmail();
```
ericsong: does this have to be public? It looks like all of the usages can be narrowed down to `WorkspaceServiceImpl`.

It is called from WorkspaceController but all 4 call sites pass in the same value so we can refactor the arguments to remove the registered users group and fetch it directly from `WorkspaceServiceImpl`.

Reduces # of arguments and reduces interface scope so I think it's a win-win

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -78,4 +80,8 @@ UserRecentWorkspace updateRecentWorkspaces(
...
+
+  void setWorkbenchConfigProvider(Provider<WorkbenchConfig> configProvider);
```
ericsong: We actually found a new way to do this that doesn't require exposing a testing only method.


See `FreeTierBillingServiceTest.java` for an example. `Provider<User` within `WorkspaceControllerTest.java` also follows this method.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -434,8 +445,15 @@ public Workspace saveAndCloneCohortsConceptSetsAndDataSets(Workspace from, Works
   public WorkspaceAccessLevel getWorkspaceAccessLevel(
       String workspaceNamespace, String workspaceId) {
     WorkspaceACL workspaceACL = fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId);
+    Map<String, WorkspaceAccessEntry> workspaceAcls = workspaceACL.getAcl();
+    WorkspaceAccessEntry userAcl = workspaceAcls.get(userProvider.get().getEmail());
+    // Published / featured workspaces don't give ACLs to everyone per se, they give READER access
+    // to the registered user domain email, so we check for that one if the user doesn't have access
+    // themself.
+    WorkspaceAccessEntry actualAcl =
+        userAcl == null ? workspaceAcls.get(getRegisteredUserDomainEmail()) : userAcl;
```
ericsong: if a user's ACL is NO ACCESS, should they still get READER access through the registered domain?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
ericsong: I'll file a ticket to clean this up across our test files so there's no more outdated examples.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
als364: Oh, yeah, good point.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
als364: Ooh, sweet

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
als364: If it's NO_ACCESS, then `userAcl` wouldn't be null.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -434,8 +445,15 @@ public Workspace saveAndCloneCohortsConceptSetsAndDataSets(Workspace from, Works
   public WorkspaceAccessLevel getWorkspaceAccessLevel(
       String workspaceNamespace, String workspaceId) {
     WorkspaceACL workspaceACL = fireCloudService.getWorkspaceAcl(workspaceNamespace, workspaceId);
+    Map<String, WorkspaceAccessEntry> workspaceAcls = workspaceACL.getAcl();
+    WorkspaceAccessEntry userAcl = workspaceAcls.get(userProvider.get().getEmail());
+    // Published / featured workspaces don't give ACLs to everyone per se, they give READER access
+    // to the registered user domain email, so we check for that one if the user doesn't have access
+    // themself.
+    WorkspaceAccessEntry actualAcl =
```
calbach: This is also assuming the user is registered, which may not be true at this point.

Ugh - Sorry about this bug - this was really not even related to recent workspaces. This was introduced when we optimized the ACL check a few week's back - I was the reviewer and should have caught that. I'm not clear on how featured workspaces are currently functional, actually - this may have broader implications.

I think I'd prefer to just throw out this method, and replace it with the following: `getWorkspace(fields="accessLevel")` , this pushes the computation into Firecloud. I think this should be hopefully be equally performant.

als364: Sorry, do you mean replace this entire method with that call or replace the call to `fireCloudService` in this method with that call

calbach: Sorry, I meant: leave the method signature as-is for `getWorkspaceAccessLevel()`, replace the implementation with a call to `getWorkspace()` instead.

als364: I'm still missing something here. I don't know how that change would handle either the user not being registered or the workspace being a featured workspace.

als364: Also - how would a user that is not registered be able to even look at featured workspaces?

als364: I went digging through Rawls. It looks like, under the hood, the thing that gets called when when you do `getWorkspace(fields="accessLevel")` is `getMaximumAccessLevel`, which has an [all caps warning](https://github.com/broadinstitute/rawls/blob/7d26a387cf0267cf19d68921047b915a5dada2f8/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L314) to not use it for actual access decisions. Talking to cloud accounts team about what the actual best way to answer the question of 'should this person be allowed to see / touch this thing' is.

calbach: OK, thanks for digging - would be curious what they say on that.

RE: registered group, it should return the derived access level, which is what this method wants.
RE: how is an unregistered user able to look at featured workspaces? Exactly, they aren't supposed to be able to. So this method should behave consistently with that; it may even be used in places to make the decision about whether a user can look at them. This method is fairly generic and central - it should not be assuming that an incoming user has already been checked for registered access.

calbach: Also to expand on that first bullet which wasn't very clear: in the featured workspaces case, the registered tier group is a reader. Registered users are members of this group, so their effective access level should be reader. Rawls should be doing this when it returns the derived access level, though there may be a more direct way.

(Disclaimer for PR scope: we don't currently have any direct calls to Sam)

als364: Cloud Accounts said 'talk directly to SAM' - we should call [this endpoint](https://sam.dsde-prod.broadinstitute.org/#!/Resources/resourceAction) with the resourceId we get back from asking Rawls about the workspace and then ask about action 'read'. So that's going to blow this up a bit

PR #2797, opened 2019-10-31
[no ticket][risk=no] Don't try to cast WorkbenchExceptions to ApiExceptions

als364:
When working on #2796, I was getting exceptions like:
```
api_1                    | SEVERE: java.lang.ClassCastException
api_1                    | java.lang.ClassCastException: org.pmiops.workbench.exceptions.NotFoundException cannot be cast to org.pmiops.workbench.notebooks.ApiException
api_1                    |     at org.pmiops.workbench.notebooks.NotebooksRetryHandler.convertException(NotebooksRetryHandler.java:14)
api_1                    |     at org.pmiops.workbench.utils.RetryHandler.run(RetryHandler.java:38)
api_1                    |     at org.pmiops.workbench.notebooks.LeonardoNotebooksClientImpl.createCluster(LeonardoNotebooksClientImpl.java:125)
api_1                    |     at org.pmiops.workbench.api.ClusterController.listClusters(ClusterController.java:133)
api_1                    |     at org.pmiops.workbench.api.ClusterApiController.listClusters(ClusterApiController.java:47)
```
This is caused by our assumption in `RetryHandler` that every error that happens is an `ApiException`. We then try to cast those errors to `WorkbenchException`s (line 41), which causes the `ClassCastException`. If we get a `WorkbenchException` within a `RetryHandler.run`, we should just straight up throw it.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
@@ -64,9 +64,8 @@ public class LeonardoNotebooksClientImpl implements LeonardoNotebooksClient {
 
   private ClusterRequest createFirecloudClusterRequest(
       String userEmail,
-      String workspaceNamespace,
-      String workspaceName,
-      @Nullable ClusterConfig clusterOverride) {
+      @Nullable ClusterConfig clusterOverride,
+      Map<String, String> customClusterEnvironmentVariables) {
     if (clusterOverride == null) {
       clusterOverride = new ClusterConfig();
     }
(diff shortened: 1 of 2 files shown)


als364:
This API test failure is preexisting (reproducible on master). I'll take a look in a separate branch

calbach:
OK, had to stare at this for a bit. `WorkbenchException` is not expected for the NotebooksRetry handler, since those are throwing a different class of generated exceptions, from the Leo backend.

I think an unexpected thing here is that there's additional complex external calls happening inside the notebooks retry handler. I think separately we should avoid that, in particular I would avoid making a heavy backend call inside of an innocuous-looking helper method `createFirecloudClusterRequest`. This also avoids rerunning that call on every retry, i.e.  I would push this logic up the stack, pass what you need as a parameter to `createFirecloudClusterRequest`, determine the CDR version outside of the retry loop.https://github.com/all-of-us/workbench/blob/fbb276727d8ac4ca9d81e83340ed8c5578e31d26/api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java#L65 

Separately we can consider whether this change is worth making. I don't think there are scenarios where it **should** currently matter, though I could see an argument that this is the correct logic. Right now the retry handler is intended to retry/propagate errors from backend services.

calbach:
Codacy error was due to a shadow version of the Codacy project which was accidentally created (see the URL on the status link). API errors are unrelated and should be fixed if you merge or rebase master and push the branch again. Codacy should also run the right  version next time you push.

PR #2804, opened 2019-11-01
[RW-3843][risk=no] Part 1 - Refactoring UserRecentResource 

ericsong:
- Removes `VisibleForTesting` setters in favor of dependency injection
- remove timestamp argument from service calls that should always be set to `now` 
- misc test code cleanup. removing unnecessary calls, object instantiations, etc

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ClusterController.java
@@ -199,9 +199,7 @@ public class ClusterController implements ClusterApiDelegate {
         .forEach(
             notebook ->
                 userRecentResourceService.updateNotebookEntry(
-                    workspaceId,
-                    userProvider.get().getUserId(),
-                    gcsNotebooksDir + "/" + notebook));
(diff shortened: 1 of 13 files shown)


api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -199,10 +199,7 @@
         .forEach(
             notebook ->
                 userRecentResourceService.updateNotebookEntry(
-                    workspaceId,
-                    userProvider.get().getUserId(),
-                    gcsNotebooksDir + "/" + notebook,
-                    now));
+                    workspaceId, userProvider.get().getUserId(), gcsNotebooksDir + "/" + notebook));
```
jaycarlton: nits: did spotless do this? FWIW if it's more than two nontrivial arguments I prefer them one to a line.

Also seems like we should have a full notebook path function somewhere handy so this function doesn't need to know what separator to use or do any string manipulation in-place.

ericsong: yeah. it's enforcing this style so I think I'm going to stick with it.

PR #2813, opened 2019-11-05
[RW-3311][risk=no] Fix behavior when clearing concept set search box

andy7i:
https://precisionmedicineinitiative.atlassian.net/browse/RW-3311
Ticket description: 
> Summary
> clicking the x icon does not remove the keyword from the search box.
> 
> Steps to reproduce
> 1. Type “lung cancer” in the search box, and click enter.
> 2. Look for the results below.
> 3. To enter a new keyword, click the X icon in the search box.
> 
> Expected results
> The keyword will be removed from the search box and the table below refreshes to default (card display).
> 
> Actual results
> The keyword is not removed from search box.
> The keywords in the results/table below are not highlighted anymore.
> Also the results/table displayed below do not reset, unless the browser is refreshed.
> 
> Task
> Make the click on the X remove the search term.

To test: 
open a workspace -> new dataset -> new concept set
-OR- go to http://localhost:4200/workspaces/<BP>/<WS>/data/concepts

------

Notes

experimented briefly with wrapping the textinput with a form instead of handling submit (enter key) manually, but on form submit, the page would reload.
```
<form onSubmit={this.handleSubmit}>
<TextInput ....>
<input type='submit' style={{display: 'none'}} />
</form>
```

The better way to handle the value of an input field is with 1 source of truth (the state) as recommended in:
https://reactjs.org/docs/forms.html#controlled-components
A number of changes reflect this decision.

`e.key === 'Enter'` makes more sense than `e.keyCode === 13`

Fixed UX issue where the minimum character limit error message does not go away unless you click X. Now it will go away upon a search with more than 3 characters.

Use arrow function instead of binding the function to a component:
https://reactjs.org/docs/faq-functions.html#arrow-function-in-render
https://reactjs.org/docs/handling-events.html

Fixed typo

Fixed linter errors

3 spaces (sometimes) caused a critical error (app goes blank):
search for lung
then before results are fully loaded, replace "lung" with 3 spaces.
```
ERROR TypeError: Cannot read property 'toLowerCase' of undefined
    at concept-table.tsx:182
```
This checks that if there's a search term, it can't just be all whitespace.

------

Notes below didn't make it into the final code (due to ticket requirements requesting a different behavior) but is added as an FYI:

Clearing the search box (after a search) would clear the search box and the search results table would refresh to the search term again. I would expect it to do a blank search though. Turns out there's a race condition where setState is run async so we need to mimic a synchronous action by putting followup code into a callback function.
```
setState() does not immediately mutate this.state but creates a pending state transition. Accessing this.state after calling this method can potentially return the existing value.

   There is no guarantee of synchronous operation of calls to setState and calls may be batched for performance gains.
```
https://stackoverflow.com/questions/42018342/is-there-a-synchronous-alternative-of-setstate-in-reactjs

### DIFF
--- a/ui/src/app/pages/data/concept/concept-homepage.spec.tsx
+++ b/ui/src/app/pages/data/concept/concept-homepage.spec.tsx
@@ -24,7 +24,6 @@ import {
   WorkspacesApiStub,
   WorkspaceStubVariables
 } from 'testing/stubs/workspaces-api-stub';
-import {Key} from 'ts-key-enum';
 
 
 function isSelectedDomain(
@@ -46,7 +45,7 @@ function searchTable(searchTerm: string, wrapper) {
   wrapper.find('[data-test-id="concept-search-input"]')
     .find('input').simulate('change', {target: {value: searchTerm}});
   wrapper.find('[data-test-id="concept-search-input"]')
-    .find('input').simulate('keypress', {key: Key.Enter});
+    .find('input').simulate('keypress', {key: 'Enter'});
 }
 
 const defaultSearchTerm = 'test';
--- a/ui/src/app/pages/data/concept/concept-homepage.spec.tsx
+++ b/ui/src/app/pages/data/concept/concept-homepage.spec.tsx
@@ -49,6 +49,8 @@ function searchTable(searchTerm: string, wrapper) {
     .find('input').simulate('keypress', {key: Key.Enter});
 }
 
+const defaultSearchTerm = 'test';
+
 describe('ConceptHomepage', () => {
 
   beforeEach(() => {
@@ -78,17 +80,16 @@ describe('ConceptHomepage', () => {
   });
 
   it('should default to standard concepts only, and performs a full search', async() => {
-    const searchTerm = 'test';
     const spy = jest.spyOn(conceptsApi(), 'searchConcepts');
     const wrapper = mount(<ConceptHomepage />);
     await waitOneTickAndUpdate(wrapper);
-    searchTable(searchTerm, wrapper);
+    searchTable(defaultSearchTerm, wrapper);
     await waitOneTickAndUpdate(wrapper);
 
     DomainStubVariables.STUB_DOMAINS.forEach((domain) => {
       const includeDomainCounts = isSelectedDomain(domain, wrapper);
       const expectedRequest = {
-        query: searchTerm,
+        query: defaultSearchTerm,
         // Tests that it searches only standard concepts.
         standardConceptFilter: StandardConceptFilter.STANDARDCONCEPTS,
         domain: domain.domain,
@@ -122,7 +123,6 @@ describe('ConceptHomepage', () => {
 
   it('should change search criteria when standard only not checked', async() => {
     const spy = jest.spyOn(conceptsApi(), 'searchConcepts');
-    const searchTerm = 'test';
     const selectedDomain = DomainStubVariables.STUB_DOMAINS[1];
     const wrapper = mount(<ConceptHomepage />);
     await waitOneTickAndUpdate(wrapper);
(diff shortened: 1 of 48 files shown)


ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -266,14 +266,19 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
-    triggerSearch(e) {
+    handleChange(e) {
+      this.setState({currentSearchString: e});
+    }
+
+    handleSubmit(e) {
       // search on enter key
-      if (e.keyCode === 13) {
-        const searchTermLength = e.target.value.trim().length;
+      if (e.key === 'Enter' ) {
```
gjuggler: (Opt) Bare strings always scare me. Either here or as a follow-on, we could use https://github.com/nfriend/ts-key-enum to refer to these keys in a type-safe way.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -266,14 +266,19 @@ export const ConceptHomepage = withCurrentWorkspace()(
       }
     }
 
-    triggerSearch(e) {
+    handleChange(e) {
```
gjuggler: Naming nit: since this is handling a change on a specific subcomponent of the concept homepage, I'd qualify the name a bit, e.g.

handleSearchChange

here and below (for handleSubmit)

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -266,14 +266,19 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
+      this.setState({currentSearchString: e});
+    }
+
+    handleSubmit(e) {
```
gjuggler: Naming nit: since this method is actually called on every keypress, maybe a better name would be "maybeHandleSubmit"? Combined with my nitpick above, we end up with an awkwardly long, but very precise, name of "maybeHandleSearchSubmit". Your call whether that's any more readable than what's here.

ui/src/app/pages/data/concept/concept-homepage.spec.tsx
```diff
@@ -1,9 +1,9 @@
 import {mount} from 'enzyme';
 import * as React from 'react';
 
+import {ConceptHomepage} from 'app/pages/data/concept/concept-homepage';
```
gjuggler: (Opt) Is there a particular reason for shifting this line up a few lines? If not, I'd leave as-is to distract less from meaningful changes.

ui/src/app/pages/data/concept/concept-homepage.spec.tsx
```diff
@@ -1,9 +1,9 @@
 import {mount} from 'enzyme';
```
gjuggler: I was hoping to see a test case that verifies the fix here and would protect against regressions. Is there a reason it would be infeasible or prohibitively expensive  to write such a test?

ui/src/app/pages/data/concept/concept-homepage.spec.tsx
andy7i: the linter wanted this to be alphabetized

ui/src/app/pages/data/concept/concept-homepage.tsx
andy7i: running `npm audit fix` shows a few breaking changes (not yet applied)

```
+ @angular-devkit/build-angular@0.803.17
added 204 packages from 111 contributors, removed 164 packages, updated 76 packages and moved 6 packages in 33.769s
fixed 13 of 78 vulnerabilities in 67513 scanned packages
  1 vulnerability required manual review and could not be updated
  3 package updates for 64 vulnerabilities involved breaking changes
  (use `npm audit fix --force` to install breaking changes; or refer to `npm audit` for steps to fix these manually)

```

ui/src/app/pages/data/concept/concept-homepage.spec.tsx
andy7i: Seems like a good idea.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -266,14 +267,19 @@ export const ConceptHomepage = withCurrentWorkspace()(
       }
     }
 
-    triggerSearch(e) {
+    handleSearchStringChange(e) {
+      this.setState({currentSearchString: e});
+    }
+
+    searchOnEnterKeyPress(e) {
```
gjuggler: Nitpick redux: maybe this contradicts my previous comments, but this name reads somewhat awkwardly now. Maybe prefer consistency with the above method, for something like "handleSearchKeyPress"?

ui/src/app/pages/data/concept/concept-homepage.spec.tsx
```diff
@@ -196,4 +196,25 @@ describe('ConceptHomepage', () => {
     wrapper.find('[data-test-id="clear-search"]').first().simulate('click');
     expect(wrapper.find('[data-test-id="selectedConcepts"]').length).toEqual(0);
   });
+
+  it('should clear search box and reset page to default view', async() => {
```
gjuggler: Thanks for adding!

andy7i:
Verbal plan with Greg: revert the package changes, merge it (already thumbed), then open a separate PR to handle the package updates.

PR #2815, opened 2019-11-05
[RW-3052][risk=no] Add test that verifies all liquibase files are indexed for migration

ericsong:
See https://github.com/all-of-us/workbench/pull/2812 to see a failing example after I added a liquibase changelog file and "forgot" to add it to the changelog master file.

Screen Shot 2019-11-05 at 4 14 44 PM

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/db/LiquibaseTest.java
+++ b/api/src/test/java/org/pmiops/workbench/db/LiquibaseTest.java
@@ -1,0 +1,47 @@
+package org.pmiops.workbench.db;
+
+import static org.junit.Assert.fail;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import org.junit.Test;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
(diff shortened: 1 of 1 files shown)


jaycarlton:
So this looks handy, but it's a bit odd to see real files in a unit test. I'd think you might want this as a separate (tiny) build step. But that might be overkill.

gjuggler:
> So this looks handy, but it's a bit odd to see real files in a unit test. I'd think you might want this as a separate (tiny) build step. But that might be overkill.

We do a similar thing here: https://github.com/all-of-us/workbench/blob/master/common-api/src/test/java/org/pmiops/workbench/config/WorkbenchConfigTest.java#L12-L12

It is indeed a little surprising to have these as part of a unit test suite — I think they're here more out of convenience than a core feeling they they're unit tests. They're really config validation tests, which feel & act a lot like unit tests but test config contents rather than code paths. I agree they may be worth splitting out down the line if we start seeing more and more of these and it starts to get confusing thinking about which test is where.

(Side note: I had trouble finding the WorkbenchConfigTest file because it's still under common-api... if we don't already have a tech debt ticket to merge common-api and common-ui into the main folders , I'll file one.)

ericsong:
@jaycarlton I also found it a bit weird when I was first introduced to the concept but I've found it to be a big boon in enforcing small gotchas around development practices.

Another example where I've found it really useful in the context of liquibase migrations is to set up an integration test that verifies no backwards incompatible schema changes and other practices that the team has agreed upon like field naming conventions.

PR #2816, opened 2019-11-05
[risk=low][no ticket]Use the Workspace's firecloudName when interacting with Cluster APIs

jmthibault79:
Introduced in this release's commits, so we can fix it before it goes to Production

### DIFF
--- a/ui/src/app/pages/workspace/workspace-about.tsx
+++ b/ui/src/app/pages/workspace/workspace-about.tsx
@@ -254,7 +254,7 @@ export const WorkspaceAbout = fp.flow(withUserProfile(), withUrlParams(), withCd
           <Link disabled={!workspace}
                 onClick={() => this.setState({googleBucketModalOpen: true})}>Google Bucket</Link>
           {!!this.workspaceClusterBillingProjectId() &&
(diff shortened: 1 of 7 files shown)


ui/src/app/pages/workspace/workspace-about.tsx
```diff
@@ -254,7 +254,8 @@ export const WorkspaceAbout = fp.flow(withUserProfile(), withUrlParams(), withCd
           <Link disabled={!workspace}
                 onClick={() => this.setState({googleBucketModalOpen: true})}>Google Bucket</Link>
           {!!this.workspaceClusterBillingProjectId() &&
-            <ResetClusterButton billingProjectId={this.workspaceClusterBillingProjectId()} workspaceName={this.state.workspace.name}/>}
+            <ResetClusterButton billingProjectId={this.workspaceClusterBillingProjectId()}
+                                workspaceFirecloudName={this.state.workspace.id}/>}
```
jmthibault79: this is the only truly necessary change - the rest is to make this more explicit, a la https://precisionmedicineinitiative.atlassian.net/browse/RW-3890

PR #2818, opened 2019-11-06
[RW-3844][risk=no] convert to validate.js

als364:
We were handrolling validation for this page. I've switched it to `validation.js`. Everything is now more or less in the same place.

Relatedly, I can't believe validation.js doesn't have a truthiness validator. I guess it was originally intended for validating pretty simple forms and not whatever it is we're doing.

### DIFF
--- a/ui/src/app/pages/workspace/workspace-edit.tsx
+++ b/ui/src/app/pages/workspace/workspace-edit.tsx
@@ -662,9 +662,9 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
         intendedStudy: { presence: true },
         anticipatedFindings: {presence: true },
         reasonForAllOfUs: { presence: true },
-        primaryPurpose: { truthiness: { expected: true } },
-        specificPopulation: { truthiness: { expected: true } },
-        diseaseOfFocus: { truthiness: { expected: true } }
+        primaryPurpose: { truthiness: true },
+        specificPopulation: { truthiness: true },
+        diseaseOfFocus: { truthiness: true }
       });
       return <FadeBox  style={{margin: 'auto', marginTop: '1rem', width: '95.7%'}}>
         <div style={{width: '95%'}}>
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -490,26 +491,19 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-    get noSpecificPopulationSelected() {
-      return this.state.workspace.researchPurpose.population &&
-        (!this.state.workspace.researchPurpose.populationDetails ||
-          this.state.workspace.researchPurpose.populationDetails.length === 0);
-    }
-
-    get noDiseaseOfFocusSpecified() {
-      return this.state.workspace.researchPurpose.diseaseFocusedResearch &&
-        !this.state.workspace.researchPurpose.diseaseOfFocus;
+    get specificPopulationSelected() {
```
jmthibault79: 👍 for this PR generally - I made a half-hearted attempt at this once but got distracted.  Thanks for fixing.

But I prefer the previous semantics for these functions.  If `population` is not selected, this will validate correctly but the function name won't be accurate.

als364: That makes sense, I can revert it and invert the truthiness expectation.

ericsong: nit: how about adding `is` as a prefix to indicate the boolean return value?

als364: @jmthibault79 would you consider this to also solve your remark?

jmthibault79: Only adding "is" would not address my concern.  Renaming the original with "is" would be fine - maybe `isSpecificPopulationSelectionMissing()` ?

ericsong: oh, regarding the truthiness, I generally prefer `isTrue` version over the inverse unless there's a very intuitive reason for the inverse.

als364: That is why I changed it.

jmthibault79: That is my default preference as well - but I think in this case it would be slightly misleading.

als364: I am finding it confusing in this case.

PR #2823, opened 2019-11-06
[risk=low][RW-1894]Convert Notebook Redirect Component from Angular to React

jmthibault79:
Mostly based on prior work by @jacmrob and @andy7i


ui/src/app/icons/notebook-icon.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+export const NotebookIcon = ({style = {}}) =>
+  <svg style={{height: '45px', width: '36px', ...style}}
```
jaycarlton: Was this path hand-coded into this component, or was this code generated from something else? If the latter, then we should make sure to stash the input(s) somewhere.

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const wrapper = component();
+    expect(wrapper).toBeTruthy();
+  });
+
+  it('should show redirect display before showing notebook', async() => {
+    const wrapper = component();
+    expect(wrapper.exists('[data-test-id="notebook-redirect"]')).toBeTruthy();
+  });
+
+  it('redirect state for existing notebook should be "Initializing" until cluster is running', async() => {
```
jaycarlton: nit: it's nice if this can read like a sentence, like `it('puts existing notebook in redirect state')`

ui/src/app/services/notebooks-swagger-fetch-clients.ts
```diff
@@ -10,7 +10,7 @@ import {
   BaseAPI,
   ClusterApi,
   Configuration as FetchConfiguration,
-  FetchAPI
+  FetchAPI, JupyterApi, NotebooksApi
```
jaycarlton: nit: I'd keep these one per line. The way it is now it looks like they are set apart from the rest of the list somehow.

ui/src/app/utils/index.tsx
```diff
@@ -359,6 +360,11 @@ export const withCdrVersions = () => {
   return connectReplaySubject(cdrVersionStore, 'cdrVersionListResponse');
 };
 
+// HOC that provides a 'queryParams' prop with current query params
```
jaycarlton: nit: I don't know what HOC means, but if it's obvious to everyone else it's fine.

ui/src/testing/stubs/cluster-api-stub.ts
```diff
@@ -23,6 +24,7 @@ export class ClusterApiStub extends ClusterApi {
 
   listClusters(extraHttpRequestParams?: any): Promise<ClusterListResponse> {
     return new Promise<ClusterListResponse>(resolve => {
+      console.log(this.cluster);
```
jaycarlton: nit: if you want to keep this log statement, then it probably should say something about what's happening to this cluster or why it's important.

ui/src/testing/stubs/cluster-api-stub.ts
```diff
@@ -62,4 +63,11 @@ export class ClusterApiStub extends ClusterApi {
       resolve(cdrVersionList);
     });
   }
+
+  localize(projectName: string, clusterName: string, req: ClusterLocalizeRequest,
+    extraHttpRequestParams?: any): Promise<ClusterLocalizeResponse> {
+    return new Promise<ClusterLocalizeResponse>(resolve => {
+      resolve({clusterLocalDirectory: 'workspaces/${req.workspaceId}'});
```
jaycarlton: nit: can we put a function to generate the local directory on either ClusterRequest or some other handy place? This particular function may not be the place for this formatting knowledge (unless it really is the only place we need to do this).

ui/src/testing/stubs/jupyter-api-stub.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export class JupyterApiStub extends JupyterApi {
+
+  constructor() {
+    super(undefined, undefined, (..._: any[]) => {
+      throw Error('cannot fetch in tests');
+    });
+  }
+
+  public postContents(googleProject: string, clusterName: string, workspaceDir: string,
```
jaycarlton: nit: I'd probably name this `getPostContents()` so it doesn't look like this function is doing a POST. Likewise for `getPutContents()`

ui/src/testing/stubs/notebooks-api-stub.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class NotebooksApiStub extends NotebooksApi {
+
+  constructor() {
+    super(undefined, undefined, (..._: any[]) => {
+      throw Error('cannot fetch in tests');
+    });
+  }
+
+  public setCookie(googleProject: string, clusterName: string,
```
jaycarlton: nit: I'd put the arguments one per line, as it's odd to see the top line extend further than the next one.

ui/src/testing/stubs/notebooks-cluster-api-stub.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class NotebooksClusterApiStub extends ClusterApi {
+
+  constructor() {
+    super(undefined, undefined, (..._: any[]) => {
+      throw Error('cannot fetch in tests');
+    });
+  }
+
+  public startCluster(googleProject: string, clusterName: string,
```
jaycarlton: I don't see startCluster() or setCookie() called anywhere. Are they in use indirectly somehow?

ui/src/app/services/notebooks-swagger-fetch-clients.ts
jmthibault79: yep, accident

ui/src/app/utils/index.tsx
jmthibault79: https://reactjs.org/docs/higher-order-components.html but I'm mostly faking my knowledge

ui/src/testing/stubs/jupyter-api-stub.ts
jmthibault79: This overrides an API method which does a POST

ui/src/testing/stubs/notebooks-cluster-api-stub.ts
jmthibault79: They are both used by the NotebookRedirect component

ui/src/app/icons/notebook-icon.tsx
jmthibault79: Sorry, I don't know where this comes from.  This existed in the Angular version.

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: updated

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+        cluster.status === ClusterStatus.Stopped;
+    }
+
+    clearAndSetTimeout(timeoutCallback, timeoutMillis) {
+      clearTimeout(this.timeoutReference);
+      this.timeoutReference = setTimeout(() => timeoutCallback(), timeoutMillis);
+    }
+
+    async pollCluster(billingProjectId) {
+      const repoll = () => {
```
calbach: I don't think we need to carry over this repolling pattern. As currently written, this is more blunt than the Angular implementation in terms of error handling. Since we're in an async function, we can rewrite this method in a more straight-line fashion. Here's what I would suggest:

```
while (true) {
  const {cluster} = await retry(() => clusterApi().listClusters(), 10000, 5);
  ...
  if (cluster.status is RUNNING) {
    break;
  }
  // Effectively sleep 15s
  await new Promise(resolve => setTimeout(resolve, 15000));
}

await retry(() => this.initializeCookies(), 5000, 3);
...


async retry<T>(f: () => Promise<T>, backoff: number, maxRetries: number): T {
  let retries = 0;
  while (true) {
    try {
      return await f();
    } catch (e) {
      retries++;
      if (isAbortError(e) || retries >= maxRetries) throw e;
    }
    await new Promise(resolve => setTimeout(resolve, backoff));
  }
}
```

Would also need to address timeout leakage for such an implementation. retry accepting the `AbortSignal` parameter would be pretty neat.

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  it('should be "Resuming" until a Stopped cluster for a new notebook is running', async() => {
+    const wrapper = mountedComponent();
+
+    wrapper.setState({creatingNewNotebook: true});
+    clusterStub.cluster.status = ClusterStatus.Stopped;
+
+    await expectRedirectingAfterRunning(wrapper);
+
+    wrapper.unmount();
```
calbach: Interesting - I would not have thought this would be necessary. Does unmount not happen automatically at the end of these test cases?

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+    accessLevel: WorkspaceAccessLevel.OWNER,
+  };
+  const profile = ProfileStubVariables.PROFILE_STUB;
+  const reload = jest.fn();
+  const updateCache = jest.fn();
+
+  let clusterStub: ClusterApiStub;
+
+  /*
+  NotebookRedirect.progressCardStates:
```
calbach: rm commented out code - this will inevitably go out of sync

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+  {includes: [Progress.Unknown, Progress.Initializing, Progress.Resuming], icon: 'notebook'},
+  {includes: [Progress.Authenticating], icon: 'success-standard'},
+  {includes: [Progress.Creating, Progress.Copying], icon: 'copy'},
+  {includes: [Progress.Redirecting], icon: 'circle-arrow'}
+   */
+  const progressCardIds = {
```
calbach: DO you think defining an enum in the notebook redirect file rather than using index numbers would be an improvement?

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+    componentDidMount() {
+      this.setNotebookNames();
+      this.pollCluster(this.props.workspace.namespace);
+    }
+
+    componentWillUnmount() {
+      clearTimeout(this.timeoutReference);
+    }
+
+    isClusterInProgress(cluster: Cluster): boolean {
```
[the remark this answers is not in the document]
jmthibault79: done

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(),
+  withQueryParams())(class extends React.Component<Props, State> {
+
+    private timeoutReference: NodeJS.Timer;
+
+    constructor(props) {
+      super(props);
+      this.state = {
+        creatingNewNotebook: !!props.queryParams.creating,
+        fullNotebookName: undefined,
```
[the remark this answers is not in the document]
jmthibault79: Simply whether it includes the `.ipynb` suffix or not.  Added a comment.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+
+export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(),
+  withQueryParams())(class extends React.Component<Props, State> {
+
+    private timeoutReference: NodeJS.Timer;
+
+    constructor(props) {
+      super(props);
+      this.state = {
+        creatingNewNotebook: !!props.queryParams.creating,
```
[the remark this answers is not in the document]
jmthibault79: That is a good point.  Updated.

ui/src/app/pages/analysis/notebook-redirect.tsx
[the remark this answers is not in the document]
jmthibault79: done

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+          } else {
+            return 'Copying the notebook onto the server';
+          }
+        case 3:
+          return 'Redirecting to the notebook server';
+      }
+    };
+
+    // The last icon (Redirect) should be rotated 90 degrees
+    const rotateIcon = () => {
```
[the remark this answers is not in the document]
jmthibault79: No.  This is now captured as part of a `Map<ProgressCardState, Icon>`

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+    constructor(props) {
+      super(props);
+      this.state = {
+        creatingNewNotebook: !!props.queryParams.creating,
+        fullNotebookName: undefined,
+        initialized: false,
+        leoUrl: undefined,
+        localizationError: false,
+        notebookName: undefined,
+        playgroundMode: props.queryParams.playgroundMode === true,
```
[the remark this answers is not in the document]
jmthibault79: It did not work.  Thank you for the catch.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+        }
+      } catch (e) {
+        repoll();
+      }
+    }
+
+    incrementProgress(p: Progress): void {
+      this.setState({
+        progress: p,
+        progressComplete:  this.state.progressComplete.set(p, true)
```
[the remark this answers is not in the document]
jmthibault79: Fixed

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+const progressCardStates = [
+  {includes: [Progress.Unknown, Progress.Initializing, Progress.Resuming], icon: 'notebook'},
+  {includes: [Progress.Authenticating], icon: 'success-standard'},
+  {includes: [Progress.Creating, Progress.Copying], icon: 'copy'},
+  {includes: [Progress.Redirecting], icon: 'circle-arrow'}
+];
+
+const ProgressCard: React.FunctionComponent<{currentState: Progress, index: number,
+  progressComplete: Map<Progress, boolean>, creatingNewNotebook: boolean}> =
+  ({index, currentState, progressComplete, creatingNewNotebook}) => {
```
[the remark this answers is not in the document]
jmthibault79: fixed

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    async pollCluster(billingProjectId) {
+      const repoll = () => {
+        this.clearAndSetTimeout(() => this.pollCluster(billingProjectId), 15000);
+      };
+      const {workspace} = this.props;
+      const {initialized, creatingNewNotebook, fullNotebookName} = this.state;
+      try {
+        const resp = await clusterApi().listClusters(billingProjectId, workspace.id);
```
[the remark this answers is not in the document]
jmthibault79: fixed

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+    'mimetype': 'text/x-python',
+    'name': 'python',
+    'nbconvert_exporter': 'python',
+    'pygments_lexer': 'ipython3',
+    'version': '3.4.2'
+  }
+};
+
+const progressCardStates = [
+  {includes: [Progress.Unknown, Progress.Initializing, Progress.Resuming], icon: 'notebook'},
```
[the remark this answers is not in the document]
jmthibault79: Fixed.  This is now a `Map<ProgressCardState, Array<Progress>>`

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,425 @@
...
+  {includes: [Progress.Redirecting], icon: 'circle-arrow'}
+];
+
+const ProgressCard: React.FunctionComponent<{currentState: Progress, index: number,
+  progressComplete: Map<Progress, boolean>, creatingNewNotebook: boolean}> =
+  ({index, currentState, progressComplete, creatingNewNotebook}) => {
+    const includesStates = progressCardStates[index].includes;
+    const icon = progressCardStates[index].icon;
+    const isCurrent = includesStates.includes(currentState);
+    const isComplete = currentState.valueOf() > includesStates.slice(-1).pop().valueOf();
```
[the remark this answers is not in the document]
jmthibault79: Changed to your suggestion

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: Yes - added

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: This was intended as a comment to convey the same information that the enum now does

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: It does not seem to be necessary - removed

jmthibault79:
@calbach Addressed comments besides the polling refactoring - I'll need to think further about that one on Monday when I'm back at the Broad.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,453 @@
...
+export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(),
+  withQueryParams())(class extends React.Component<Props, State> {
+
+    private timeoutReference: NodeJS.Timer;
+    private aborter = new AbortController();
+
+    constructor(props) {
+      super(props);
+      this.state = {
+        initialized: false,
```
calbach: nit: Existing issue that I probably personally introduced originally, but rereading this I would probably make this slightly more specific: `clusterInitialized`. `initialized` at first glance could be misinterpreted as some kind of overall component initialization state.

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  async function expectRedirectingAfterRunning(wrapper) {
+    await waitOneTickAndUpdate(wrapper);
+
+    expect(wrapper
+      .exists(getCardSpinnerTestId(ProgressCardState.UnknownInitializingResuming)))
+      .toBeTruthy();
+
+    clusterStub.cluster.status = ClusterStatus.Running;
+    jest.runOnlyPendingTimers();
```
calbach: Hm... it does seem like you should be calling this above as well to get proper coverage

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+    expect(wrapper.exists('[data-test-id="notebook-redirect"]')).toBeTruthy();
+  });
+
+  it('should be "Initializing" until a Creating cluster for an existing notebook is running', async() => {
+    const wrapper = mountedComponent();
+
+    wrapper.setState({creatingNewNotebook: false});
+    clusterStub.cluster.status = ClusterStatus.Creating;
+
+    await expectRedirectingAfterRunning(wrapper);
```
calbach: We might want these tests to be slightly more focused, i.e. assert the before state, change the status, assert the after state. I think the after state here should care more about what the card shows after this step finishes.

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  it('should be "Resuming" until a Stopped cluster for an existing notebook is running', async() => {
+    const wrapper = mountedComponent();
+
+    wrapper.setState({creatingNewNotebook: false});
+    clusterStub.cluster.status = ClusterStatus.Stopped;
```
calbach: It would be good to have some kind of assertion that this actually is "Resuming", and not just the boilerplate initialization message

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+    jest.useRealTimers();
+    jest.clearAllMocks();
+  });
+
+  it('should render', () => {
+    const wrapper = mountedComponent();
+    expect(wrapper).toBeTruthy();
+  });
+
+  it('should show redirect display before showing notebook', async() => {
```
calbach: Another case which would be good to capture is when the cluster is already running when the user first hits the page, e.g. what the card looks like after

jmthibault79: added

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: added

ui/src/app/pages/analysis/notebook-redirect.spec.tsx
jmthibault79: added

ui/src/app/pages/analysis/notebook-redirect.tsx
jmthibault79: This retry pattern exists in lots of places, so this would be a good candidate for a tech debt ticket.  I can remove it here.

jmthibault79: I believe I have address this now.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,453 @@
...
+        const resp = await clusterApi().localize(cluster.clusterNamespace, cluster.clusterName,
+          {workspaceNamespace: workspace.namespace, workspaceId: workspace.id,
+            notebookNames: notebookNames, playgroundMode: this.isPlaygroundMode()},
+          {signal: this.aborter.signal});
+        return resp.clusterLocalDirectory;
+      } catch (error) {
+        retryCount += 1;
+        if (retryCount <= 3) {
+          console.error('retrying notebook localization');
+          return this.localizeNotebooksWithRetry(cluster, notebookNames, retryCount);
```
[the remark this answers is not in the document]
jmthibault79: updated to use the retry pattern you suggested

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,453 @@
...
+        } else {
+          console.error(error);
+          this.setState({localizationError: true});
```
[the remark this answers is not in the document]
jmthibault79: now throws properly

ui/src/app/pages/analysis/notebook-redirect.tsx
jmthibault79: removed "initialized" from state - it now initializes in `initializeClusterStatusChecking()` which is called before this

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,497 @@
...
+      const resp = await this.clusterRetry(() => clusterApi().listClusters(
+        billingProjectId, this.props.workspace.id, {signal: this.aborter.signal}));
+      return resp.defaultCluster;
+    }
+
+    private scheduleClusterPoll(billingProjectId) {
+      this.clearAndSetInterval(() => this.pollForRunningCluster(billingProjectId), clusterPollingIntervalMillis);
+    }
+
+    private async clusterRetry<T>(f: () => Promise<T>): Promise<T> {
```
jmthibault79: Just a wrapper around retry with particular parameters.  I can remove or rename this.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -0,0 +1,497 @@
...
+      return await retry(f, clusterApiRetryTimeoutMillis, clusterApiRetryAttempts);
+    }
+
+    private async pollForRunningCluster(billingProjectId) {
+      try {
+        const cluster = await this.getDefaultCluster(billingProjectId);
```
jmthibault79: This call and startCluster below have retries.  This may be unwise, because this function is repeatedly called via setInterval()

ui/src/app/services/swagger-fetch-clients.ts
```diff
@@ -31,13 +31,11 @@ import {
   CohortsApi,
   ConceptsApi,
   ConceptSetsApi,
-  ConfigApi,  // not currently used
```
jmthibault79: not related to the primary purpose of this PR

ui/src/app/utils/retry.tsx
```diff
@@ -0,0 +1,19 @@
+import {isAbortError} from 'app/utils/errors';
+
+export async function retry<T>(f: () => Promise<T>, timeoutMillis: number, maxRetries: number): Promise<T> {
```
jmthibault79: Based on CH's suggestion

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -251,58 +272,86 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
           } else {
             this.incrementProgress(Progress.Initializing);
           }
-          this.setState({initialized: true});
+
+          this.scheduleClusterPoll(billingProjectId);
         }
+      } catch (e) {
+        if (!isAbortError(e)) {
+          this.scheduleClusterPoll(billingProjectId);
```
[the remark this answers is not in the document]
jmthibault79: So we can try again later - though if we're failing after several retries it may make sense to give up here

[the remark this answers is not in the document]
jmthibault79: I'll remove it

ui/src/app/utils/retry.tsx
```diff
@@ -0,0 +1,19 @@
...
+    try {
+      return await f();
+    } catch (e) {
+      retries++;
+      if (isAbortError(e) || retries >= maxRetries) {
+        clearTimeout(timeoutReference);
+        throw e;
+      }
+      clearTimeout(timeoutReference);
+      timeoutReference = setTimeout(f, timeoutMillis);
```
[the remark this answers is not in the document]
jmthibault79: fixed

ui/src/app/utils/retry.tsx
```diff
@@ -0,0 +1,19 @@
+import {isAbortError} from 'app/utils/errors';
+
+export async function retry<T>(f: () => Promise<T>, timeoutMillis: number, maxRetries: number): Promise<T> {
+  let retries = 0;
+  let timeoutReference;
```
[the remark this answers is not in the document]
jmthibault79: removed

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -251,58 +272,86 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
+          await this.connectToRunningCluster(cluster);
+        } else {
+          if (this.isClusterInProgress(cluster)) {
             this.incrementProgress(Progress.Resuming);
           } else {
             this.incrementProgress(Progress.Initializing);
           }
-          this.setState({initialized: true});
+
+          this.scheduleClusterPoll(billingProjectId);
```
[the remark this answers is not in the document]
jmthibault79: `initializeClusterStatusChecking()` is separated out because I think that makes more sense than a single-use "initialized" state variable that requires a check each time.  `connectToRunningCluster()` is split out because there are two paths to get there: via the initial status check on an already-running cluster, and and via polling.

I reordered the methods to make this hopefully easier to follow.

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -251,58 +272,86 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
+    private async getDefaultCluster(billingProjectId) {
+      const resp = await this.clusterRetry(() => clusterApi().listClusters(
+        billingProjectId, this.props.workspace.id, {signal: this.aborter.signal}));
+      return resp.defaultCluster;
+    }
 
-          // give it a second to "redirect"
-          this.clearAndSetTimeout(() => this.incrementProgress(Progress.Loaded), 1000);
+    private scheduleClusterPoll(billingProjectId) {
+      this.clearAndSetInterval(() => this.pollForRunningCluster(billingProjectId), clusterPollingIntervalMillis);
```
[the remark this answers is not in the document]
jmthibault79: Agreed - restored this to setTimeout.

I also think we should eventually move this to retry but that would require setting max-attempts.  That would be a behavior change worthy of some UX discussion, and better handled in RW-3097 IMO.

ui/src/app/utils/retry.tsx
```diff
@@ -1,19 +1,18 @@
...
+      return await fetchFn();
     } catch (e) {
       retries++;
       if (isAbortError(e) || retries >= maxRetries) {
-        clearTimeout(timeoutReference);
         throw e;
       }
-      clearTimeout(timeoutReference);
-      timeoutReference = setTimeout(f, timeoutMillis);
+      await new Promise(resolve => setTimeout(resolve, timeoutMillis));
```
calbach: nit: Would put a comment to describe this as effectively a "sleep" as it may be difficult to understand what's going on here at first

ui/src/app/utils/retry.tsx
```diff
@@ -1,19 +1,18 @@
 import {isAbortError} from 'app/utils/errors';
 
-export async function retry<T>(f: () => Promise<T>, timeoutMillis: number, maxRetries: number): Promise<T> {
+// Retry a fetch `maxRetries` number of times with a `timeoutMillis` wait between retries
+// Respects fetch aborts
+export async function fetchAbortableRetry<T>(fetchFn: () => Promise<T>, timeoutMillis: number, maxRetries: number): Promise<T> {
```
calbach: Nice solution, thanks

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -286,47 +325,34 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
           await this.connectToRunningCluster(cluster);
         } else {
           // re-start cluster if stopped, and try again in the next polling interval
           if (cluster.status === ClusterStatus.Stopped) {
             await this.clusterRetry(() => notebooksClusterApi().startCluster(
               cluster.clusterNamespace, cluster.clusterName, {signal: this.aborter.signal}));
           }
+
+          // TODO(RW-3097): Backoff, or don't retry forever.
+          this.clearAndSetTimeout(() => this.pollForRunningCluster(billingProjectId), clusterPollingIntervalMillis);
```
calbach: nit: clusterPollingTimeoutMillis?

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -286,47 +325,34 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
         }
       } catch (e) {
         if (isAbortError(e)) {
           // stop polling on abort
```
calbach: I'm guessing this was leftover from the interval, but this comment is no longer accurate and calling `clearTimeout` here is no longer necessary

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -286,47 +325,34 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
           // re-start cluster if stopped, and try again in the next polling interval
           if (cluster.status === ClusterStatus.Stopped) {
             await this.clusterRetry(() => notebooksClusterApi().startCluster(
               cluster.clusterNamespace, cluster.clusterName, {signal: this.aborter.signal}));
           }
+
+          // TODO(RW-3097): Backoff, or don't retry forever.
+          this.clearAndSetTimeout(() => this.pollForRunningCluster(billingProjectId), clusterPollingIntervalMillis);
         }
       } catch (e) {
```
calbach: In fact, this entire try catch no longer seems necessary, the level above will catch it just as well

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -352,41 +378,6 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
       this.clearAndSetTimeout(() => this.incrementProgress(Progress.Loaded), 1000);
```
calbach: One comment on this clearAndSetTimeout() pattern: I think this particular usage shows some issues with the pattern. Namely - to use it in this context, you have to know that the timeout is not currently pending on any of the other callsites. I think in this case it would be safer and easier to verify by inspection if you just had two separate timer state trackers:

this.pollTimer
this.redirectTimer

Also to be clear, it is not always necessary to call `clearTimeout`; it's effectively a no-op if the timeout has already fired. I realize I may not have addressed this point when replying to your original proposal. In general if you write your code such that it does not overwrite references to live timeouts (no need to worry about spent ones), then clearing them in the unmount section only is sufficient. I think this is probably the guidance we should use here, though it took me a while to verbalize it properly.

ui/src/app/utils/retry.tsx
jmthibault79: done

ui/src/app/pages/analysis/notebook-redirect.tsx
jmthibault79: done

ui/src/app/pages/analysis/notebook-redirect.tsx
jmthibault79: good points.  updated to use setTimeout() directly for these two handles

ui/src/app/pages/analysis/notebook-redirect.tsx
jmthibault79: as discussed offline - repurposing this to show error messages

ui/src/app/pages/analysis/notebook-redirect.tsx
```diff
@@ -325,11 +323,13 @@ export const NotebookRedirect = fp.flow(withUserProfile(), withCurrentWorkspace(
...
 
-          this.clearAndSetTimeout(() => this.pollForRunningCluster(billingProjectId), clusterPollingIntervalMillis);
+          this.pollTimer = setTimeout(() => this.pollForRunningCluster(billingProjectId), clusterPollingTimeoutMillis);
         }
-      } catch (e) {
-        if (!isAbortError(e)) {
-          throw e;
+      } catch (error) {
+        if (!isAbortError(error)) {
+          console.error(error);
```
calbach: opt: maybe update this to use reportError instead, would be nice to have this logged as a UI error as well I think, since we don't have visibility into all of the API calls server side: https://github.com/all-of-us/workbench/blob/9473ca282c63015fe8d8c7f97258a3302aded80f/ui/src/app/utils/errors.tsx#L16

PR #2828, opened 2019-11-07
[RISK=LOW][RW-3822] Default selection of values

NehaBroad:
Issue: https://precisionmedicineinitiative.atlassian.net/browse/RW-3822
**Solution**: If the selected concept set introduces a new domain add values to valueSet as well as selected domain list 
**Testing**:
1) Add a concept set with Domain say 'Condition'
    a) the values tab should be populated with list of Condition domain values
    b) All of the values should be selected by default
    c) Values tab header should have the selected Checkbox with label Deselect All 
2) De-select some of the values
    a) Values tab header should have the un-selected Checkbox with label Select All 
3) Select another Concept set with the same domain 'Condition'
   a) There should be no change in values, not only the values remain the same also the header should still say 'Select All' with un-selected checkbox

### DIFF
--- a/ui/src/app/pages/data/data-set/dataset-page.spec.tsx
+++ b/ui/src/app/pages/data/data-set/dataset-page.spec.tsx
@@ -90,27 +90,23 @@ describe('DataSetPage', () => {
     async() => {
       const wrapper = mount(<DataSetPage/>);
       await waitOneTickAndUpdate(wrapper);
-      await waitOneTickAndUpdate(wrapper);
 
       // Select Condition Concept set
-      const condition_concept = wrapper.find('[data-test-id="concept-set-list-item"]').first()
+      const conditionConceptSet = wrapper.find('[data-test-id="concept-set-list-item"]').first()
           .find('input').first();
-      condition_concept.simulate('change');
+      conditionConceptSet.simulate('change');
       await waitOneTickAndUpdate(wrapper);
       let valueListItems = wrapper.find('[data-test-id="value-list-items"]');
-      const conditionValueList = valueListItems.first().find('input').first();
       expect(valueListItems.length).toBe(2);
       let checkedValuesList = valueListItems.filterWhere(value => value.props().checked);
-
 
       // All values should be selected by default
       expect(checkedValuesList.length).toBe(2);
 
       // Select second concept set which is Measurement domain
-      const measurement_concept = wrapper.find('[data-test-id="concept-set-list-item"]').at(1)
+      const measurementConceptSet = wrapper.find('[data-test-id="concept-set-list-item"]').at(1)
           .find('input').first();
-      measurement_concept.simulate('change');
-      await waitOneTickAndUpdate(wrapper);
+      measurementConceptSet.simulate('change');
       await waitOneTickAndUpdate(wrapper);
       valueListItems = wrapper.find('[data-test-id="value-list-items"]');
       checkedValuesList = valueListItems.filterWhere(value => value.props().checked)
@@ -118,18 +114,17 @@ describe('DataSetPage', () => {
       expect(valueListItems.length).toBe(5);
       expect(checkedValuesList.length).toBe(5);
 
-      // Unselect first Conidition value
+      // Unselect first Condition value
       valueListItems.first().find('input').first().simulate('change');
       valueListItems = wrapper.find('[data-test-id="value-list-items"]');
       checkedValuesList = valueListItems.filterWhere(value => value.props().checked)
       expect(checkedValuesList.length).toBe(4);
 
       // Select another condition concept set
-      const second_condition_concept =
+      const secondConditionConceptSet =
           wrapper.find('[data-test-id="concept-set-list-item"]').at(2)
           .find('input').first();
-      second_condition_concept.simulate('change');
-      await waitOneTickAndUpdate(wrapper);
+      secondConditionConceptSet.simulate('change');
       await waitOneTickAndUpdate(wrapper);
       valueListItems = wrapper.find('[data-test-id="value-list-items"]');
       checkedValuesList = valueListItems.filterWhere(value => value.props().checked);
@@ -144,7 +139,6 @@ describe('DataSetPage', () => {
   it('should enable save button and preview button once cohorts, concepts and values are selected',
     async() => {
       const wrapper = mount(<DataSetPage />);
-      await waitOneTickAndUpdate(wrapper);
       await waitOneTickAndUpdate(wrapper);
 
       // Preview Button and Save Button should be disabled by default
@@ -180,7 +174,6 @@ describe('DataSetPage', () => {
   it('should display preview data table once preview button is clicked', async() => {
     const spy = jest.spyOn(dataSetApi(), 'previewDataSetByDomain');
     const wrapper = mount(<DataSetPage />);
-    await waitOneTickAndUpdate(wrapper);
     await waitOneTickAndUpdate(wrapper);
 
     // Select one cohort , concept and value
@@ -254,7 +247,6 @@ describe('DataSetPage', () => {
   it('should call load data dictionary when caret is expanded', async() => {
     const spy = jest.spyOn(dataSetApi(), 'getDataDictionaryEntry');
     const wrapper = mount(<DataSetPage />);
-    await waitOneTickAndUpdate(wrapper);
     await waitOneTickAndUpdate(wrapper);
 
     // Select one cohort , concept and value
--- a/ui/src/app/pages/data/data-set/dataset-page.spec.tsx
+++ b/ui/src/app/pages/data/data-set/dataset-page.spec.tsx
@@ -69,11 +69,10 @@ describe('DataSetPage', () => {
     await waitOneTickAndUpdate(wrapper);
     let valueListItems = wrapper.find('[data-test-id="value-list-items"]');
     expect(valueListItems.length).toBe(2);
-    let selectedValue = valueListItems.forEach(value =>
-        value.props().checked)
+    let checkedValuesList = valueListItems.filterWhere(value => value.props().checked);
 
     // All values should be selected by default
-    expect(selectedValue.length).toBe(2);
+    expect(checkedValuesList.length).toBe(2);
 
     // Second Concept set in concept set list has domain "Measurement"
     const measurement_concept = wrapper.find('[data-test-id="concept-set-list-item"]').at(1)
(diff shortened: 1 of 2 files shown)


jaycarlton:
It would be good to put a description on the problem, its root cause, and the solution you put in.

gjuggler:
> It would be good to put a description on the problem, its root cause, and the solution you put in.

+1 -- In general, a good PR description should be a preamble to guide a reviewer (or, equally importantly, someone who encounters this PR via 'git blame' months or years in the future). In this case, both the PR and the Jira ticket are quite terse... a bit more detail would be helpful, perhaps with a screenshot if visual aid is helpful.

I've also started including a "TESTED=" line on my PRs, indicating how and when I've run through manual testing in a local env.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,8 +602,32 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
       this.setState({selectedDomainValuePairs: valuesSelected, dataSetTouched: true});
     }
 
+    // Add the values to the valueSet as well as the selectedDomain list so that they are
+    // all selected by default
+    updateValueSets(updatedValueSets, newValueSets) {
```
gjuggler: Typescript nitpick: please specify the types of each parameter.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,8 +602,32 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+      this.setState({
+        valueSets: updatedValueSets.concat(newValueSets),
+        valuesLoading: false
+      });
+      const allValuesSelected = this.state.selectedDomainValuePairs;
```
gjuggler: (opt / nitpick) The name choice here feels a little off. Is this variable meant to hold "all currently-selected values", or is it mean to indicate that "all values should be selected"?

I think, in terms of communicating the intent to readers, a more generic choice would be to simply destructure the state variable and include a comment:

// Newly-fetched values should be selected by default, so we add them to the selection-tracking state property.
const {selectedDomainValuePairs} = this.state;
...
this.setState({selectedDomainValuePairs: selectedDomainValuePairs});

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -67,15 +67,24 @@ describe('DataSetPage', () => {
...
         .find('input').first();
     measurement_concept.simulate('change');
     await waitOneTickAndUpdate(wrapper);
     await waitOneTickAndUpdate(wrapper);
-    expect(wrapper.find('[data-test-id="value-list-items"]').length).toBe(5);
+    valueListItems = wrapper.find('[data-test-id="value-list-items"]');
+    selectedValue = valueListItems.forEach(value =>
+        value.props().checked);
+    expect(valueListItems.length).toBe(5);
+    expect(selectedValue.length).toBe(5);
```
gjuggler: (no action required) Just for my own understanding: is this the critical assertion that would have failed before the changes introduced in this PR?

And, thought experiment: are there any other edge cases related to the issue addressed by this PR that aren't covered by unit tests here?

It's definitely easier to protect nuanced, complex behavior with a unit test specifically targeted at the given user flow. It might be interesting to consider that for this case, if there's risk that this core test case isn't comprehensive enough.

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -67,15 +67,78 @@ describe('DataSetPage', () => {
...
+    checkedValuesList = valueListItems.filterWhere(value => value.props().checked);
+    expect(valueListItems.length).toBe(5);
+    expect(checkedValuesList.length).toBe(5);
+  });
+
+  it('should select all values by default on selection on concept set only if the new domain is unique',
+    async() => {
+      const wrapper = mount(<DataSetPage/>);
+      await waitOneTickAndUpdate(wrapper);
+      await waitOneTickAndUpdate(wrapper);
```
gjuggler: Is there a reason for the double-waitOneTick call here? I see this pattern is used elsewhere. If it's important to the test, I'd leave a comment to help the reader understand why. Have you tried removing it? If so, what broke?

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -67,15 +67,78 @@ describe('DataSetPage', () => {
...
+      const condition_concept = wrapper.find('[data-test-id="concept-set-list-item"]').first()
+          .find('input').first();
+      condition_concept.simulate('change');
+      await waitOneTickAndUpdate(wrapper);
+      let valueListItems = wrapper.find('[data-test-id="value-list-items"]');
+      const conditionValueList = valueListItems.first().find('input').first();
+      expect(valueListItems.length).toBe(2);
+      let checkedValuesList = valueListItems.filterWhere(value => value.props().checked);
+
+
```
gjuggler: style nit: remove extra newline

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -67,15 +67,78 @@ describe('DataSetPage', () => {
...
+      await waitOneTickAndUpdate(wrapper);
+      await waitOneTickAndUpdate(wrapper);
+
+      // Select Condition Concept set
+      const condition_concept = wrapper.find('[data-test-id="concept-set-list-item"]').first()
+          .find('input').first();
+      condition_concept.simulate('change');
+      await waitOneTickAndUpdate(wrapper);
+      let valueListItems = wrapper.find('[data-test-id="value-list-items"]');
+      const conditionValueList = valueListItems.first().find('input').first();
```
gjuggler: This local variable is never used – please remove.

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
```diff
@@ -67,15 +67,78 @@ describe('DataSetPage', () => {
...
+      measurement_concept.simulate('change');
+      await waitOneTickAndUpdate(wrapper);
+      await waitOneTickAndUpdate(wrapper);
+      valueListItems = wrapper.find('[data-test-id="value-list-items"]');
+      checkedValuesList = valueListItems.filterWhere(value => value.props().checked)
+      // All values condition + measurement will be selected
+      expect(valueListItems.length).toBe(5);
+      expect(checkedValuesList.length).toBe(5);
+
+      // Unselect first Conidition value
```
gjuggler: Nitpick: spelling on "Condition"

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,12 +604,37 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+      });
+      this.setState({
+        valueSets: valueSets.concat(newValueSets),
+        selectedDomainValuePairs: selectedDomainValuePairs,
+        valuesLoading: false
+      });
+    }
+
+    // Returns true if selected values set is empty or is not equal to the total values displayed
+    get partialOrNoneValuesSelected() {
```
gjuggler: Grammar nitpick: partialOrNoValuesSelected reads more correctly. Alternatively, partialOrNoneSelected reads well too – just the phrase "noneValuesSelected" sounds not quite right.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,12 +604,37 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+
+    get valuesCount() {
+      let count = 0 ;
+      this.state.valueSets.map(value => count += value.values.items.length);
+      return count;
     }
 
     selectAllValues() {
-      if (!this.selectAll) {
+      if (!this.partialOrNoneValuesSelected) {
```
gjuggler: (Opt) A common piece of coding advice I've heard is that it's usually better to have if-else statements based on a positive conditional rather than a negated one. In this case, it may actually be simpler to have a method called .isAllSelected, which would let you write something like:

if (this.isAllSelected) {
  this.setState({...});
}

If that matches the intent of the code, it would be much easier to read than the current somewhat-complicated condition.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,12 +604,37 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
       this.setState({selectedDomainValuePairs: valuesSelected, dataSetTouched: true});
     }
 
-    get selectAll() {
-      return fp.isEmpty(this.state.selectedDomainValuePairs);
+    // Append newValueSets (values from API) to state's valueSets and to selectedDomainValuePairs.
+    // Set valuesLoading to false once the state is updated
+    updateValueSets(newValueSets) {
```
gjuggler: (Non-opt) When possible, prefer to annotate the type on any method parameters. Even if TypeScript can infer this, it's helpful for the reader to know which type to keep in mind when reading through this function.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -605,8 +602,32 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+
+      newValueSets.map(valueSet => {
+        valueSet.values.items.map(value => {
+          allValuesSelected.push({domain: valueSet.domain, value: value.value});
+        });
+      });
+      this.setState({selectedDomainValuePairs: allValuesSelected});
+    }
+
     get selectAll() {
```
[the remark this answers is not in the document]
gjuggler: (See my comment in the latest review round for a suggestion to invert the meaning of this method. If this method returns true iff all items are selected, the conditionals within the render method will be simpler.)

ui/src/app/pages/data/data-set/dataset-page.spec.tsx
NehaBroad: I will confirm this but this is because there are 2 promise calls (one for cohort and another for concept set list)

NehaBroad: I tested this and one waitOneTick works so have updated the test

PR #2830, opened 2019-11-08
[RW-3949][risk=low] Bump default cluster size to highmem

calbach:
Only really necessary in prod, bump in staging/stable as well for QA purposes.


calbach:
I also manually created a cluster in Leo dev using this machine type to make sure Leo didn't have any issues with it.

PR #2831, opened 2019-11-08
[RW-3911][risk=no] Add limit to dataset builder query

ericsong:
Adds a limit to our big query sql strings from Dataset Builder.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -660,6 +660,11 @@ public class DataSetServiceImpl implements DataSetService {
             .append(domainAsString)
             .append("\"")
             .toString();
+    String sqlLimitSection =
+        new StringBuilder("# The ‘max_number_of_rows’ parameter limits the number of rows in the query so that the result set can fit in memory.\n")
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -660,14 +660,25 @@ private static String generateNotebookUserCode(
             .append(domainAsString)
             .append("\"")
             .toString();
+    String sqlLimitSection =
+        new StringBuilder(
+                "# The ‘max_number_of_rows’ parameter limits the number of rows in the query so that the result set can fit in memory.\n")
+            .append(
+                "# If you increase the limit and run into responsiveness issues, please request a VM size upgrade.\n")
+            .append("max_number_of_rows = '1000000'")
```
ericsong: I went with a string instead of an integer since it simplifies the string concatenation code. It's especially a problem in R because the default toString() on integers uses scientific notation, requiring the use of a format function.

jaycarlton: BTW, not saying this would be applicable in a bug time frame, but I have some thoughts around making the SQL generation and Python/R generation more robust and flexible. I'd like to get your thoughts.

See https://precisionmedicineinitiative.atlassian.net/browse/RW-3340 and https://precisionmedicineinitiative.atlassian.net/browse/RW-3466

PR #2835, opened 2019-11-08
Draft pattern idea to prevent leaky timeouts: clearAndSetTimeout

jmthibault79:
In a variety of places, we're setting timeout handles which we later clear which may already have handles to existing timeouts.  Thus, we can't clear the original timeout on unmount.  Here is a suggestion for a new pattern to help account for these.

### DIFF
--- a/ui/src/app/pages/analysis/interactive-notebook.tsx
+++ b/ui/src/app/pages/analysis/interactive-notebook.tsx
@@ -122,9 +122,14 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
       this.aborter.abort();
     }
 
+    clearAndSetTimeout(f, millis) {
+      clearTimeout(this.runClusterTimer);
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/analysis/interactive-notebook.tsx
```diff
@@ -122,9 +122,14 @@ export const InteractiveNotebook = fp.flow(withUrlParams(), withCurrentWorkspace
       this.aborter.abort();
     }
 
+    clearAndSetTimeout(f, millis) {
```
jaycarlton: I like it. I'd also pass in the timer to be set and cleared so it's more generally usable.

Or make this something you can install on a timer class/object of some sort, so you can do `MyTimer.reset(fun, callback, duration)`. If there's a way to map functions to timers, that would cut things down even more (unless you needed simultaneous `runCluster` timers.

Whatever patterns you decide on, it'd be worth building a little library or set of components for doing all our timed events so we can tune it in one place.

PR #2836, opened 2019-11-08
[RW-3801][Risk=no] Fix jitteriness with share and delete, and improve loading spinners

blrubenstein:
So, this fixes a few things. First, it addresses the ticket, which is to remove the jitteriness/movement of the page when you share or delete a recent workspace. Then, from feedback from Lou, it hides the recent resources header when there is no content there. Third, it reduces the amount of spinner jitteriness while the data is loading.

There is one todo that I am going to pick up, where if there are no recent resources, it currently hides the header and the area for the content, Lou wants it to have some text stating "you have no recent resources" instead, to reduce content moving. Doing that as a follow up task/PR since it is technically a new functional change.

### DIFF
--- a/ui/src/app/pages/homepage/recent-resources.tsx
+++ b/ui/src/app/pages/homepage/recent-resources.tsx
@@ -10,6 +10,7 @@ import {ResourceCard} from 'app/components/resource-card';
 import {NotebookResourceCard} from 'app/pages/analysis/notebook-resource-card';
 import {CohortResourceCard} from 'app/pages/data/cohort/cohort-resource-card';
 import {RecentResource} from 'generated/fetch';
+import {SmallHeader} from "../../components/headers";
 
 export const RecentResources = (fp.flow as any)(
   withContentRect('client'),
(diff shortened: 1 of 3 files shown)


ui/src/app/pages/homepage/recent-resources.tsx
```diff
@@ -86,24 +87,27 @@ export const RecentResources = (fp.flow as any)(
...
-          dir='left'
-          onClick={() => this.setState({offset: offset - 1})}
-          style={{position: 'absolute', left: 0, paddingBottom: '0.5rem'}}
-        />}
-        {offset + limit < resources.length && <Scroll
-          dir='right'
-          onClick={() => this.setState({offset: offset + 1})}
-          style={{position: 'absolute', right: 0, paddingBottom: '0.5rem'}}
-        />}
+    return (resources !== null && resources.length > 0) || loading ? <React.Fragment>
```
andy7i: nit: I think we may want to start formatting our multi-line ternary expressions like this:
The above can be rewritten as the following to improve readability and more clearly delineate the parts:
```
var foo = bar > baz ?
    value1 :
    value2;
```
https://eslint.org/docs/rules/multiline-ternary

I'd be happy to add this to the linter as well if this is something we want to be consistent about.

ui/src/app/pages/homepage/recent-resources.tsx
```diff
@@ -86,24 +87,27 @@ export const RecentResources = (fp.flow as any)(
...
+      <SmallHeader>Recently Accessed Items</SmallHeader>
```
andy7i: I like the move here!

ui/src/app/pages/homepage/recent-resources.tsx
```diff
@@ -86,24 +87,27 @@ export const RecentResources = (fp.flow as any)(
...
+      <div ref={measureRef} style={{display: 'flex', position: 'relative', minHeight: 247}}>
+        <div style={{display: 'flex', position: 'relative', alignItems: 'center',
```
andy7i: minor: this could be a FlexRow

ui/src/app/pages/homepage/recent-resources.tsx
```diff
@@ -86,24 +87,27 @@ export const RecentResources = (fp.flow as any)(
...
+            dir='right'
+            onClick={() => this.setState({offset: offset + 1})}
+            style={{position: 'absolute', right: 0, paddingBottom: '0.5rem'}}
+          />}
+        </div>
+        {loading && <SpinnerOverlay dark={this.props.dark} />}
       </div>
-      {loading && <SpinnerOverlay dark={this.props.dark} />}
-    </div>;
+    </React.Fragment> : null;
```
andy7i: I didn't know you could do this. Found this info (for future readers):
```
 instead of blank, simply return null if you don't want to render anything from component
```
https://stackoverflow.com/questions/42083181/is-it-possible-to-return-empty-in-react-render-function

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,35 +42,29 @@ export const RecentWorkspaces = withUserProfile()
...
-                      reload={() => this.loadWorkspaces()}
-                    />;
-                  })
-                }
-              </FlexRow>
-            </div>
-      }
-    </React.Fragment>;
+    return <div style={{position: 'relative'}}>
+      {this.state.loading && <SpinnerOverlay dark={true} />}
```
andy7i: Is this more useful than the original ternary because the page may have already loaded but the component may reload and we want the items in the component to be covered by the spinner?

andy7i: In my testing, I didn't see the dark overlay when the spinner was running. I thought it would have had a dark overlay on the recent workspaces section. 

Screen Shot 2019-11-12 at 2 52 34 PM

ui/src/app/pages/homepage/recent-workspaces.tsx
```diff
@@ -42,35 +42,29 @@ export const RecentWorkspaces = withUserProfile()
...
+      {this.state.recentWorkspaces.length === 0 && !this.state.loading ?
+        <div style={{color: colors.primary, margin: '.5em 2em'}}>
+          <h2 style={{fontWeight: 600, lineHeight: 1.5}}>Create your first Workspace</h2>
+          <div>As you create your workspaces, this area will store your most recent workspaces.</div>
+          <div>To see all workspaces created, click on <b>See all Workspaces</b> to the right.</div>
+        </div> :
```
andy7i: ternary formatting 💯

ui/src/app/pages/homepage/recent-workspaces.tsx
blrubenstein: Talked offline, this appears to be a pretty common thing.

ui/src/app/pages/homepage/recent-resources.tsx
blrubenstein: If we are adding to linter, let's do as a separate PR. Meanwhile, style change applied :D

PR #2838, opened 2019-11-08
[RW-3950][risk=no] Archive R2019Q2R1

calbach:
For program reasons, we no longer want this CDR version to be usable for new analysis.


jaycarlton:
Is there any follow-on work to notify users of workspaces pointing to this CDR or do any other kinds of migration?

calbach:
Archival does not directly affect existing workspaces, it just means you can no longer create workspaces (via create or clone) which reference it. I filed another ticket to track full turn down of this version, which is the eventual goal based on discussion with Karhtik: https://precisionmedicineinitiative.atlassian.net/browse/RW-3951

PR #2840, opened 2019-11-08
[risk=no] Add featured workspaces in staging for QA purposes

calbach:
Minimally, this will be useful for manual QA purposes. This is a gap I identified when pushing out new featured workspaces content a few weeks back. 

Before:
image

After:
image

### DIFF
--- a/api/config/featured_workspaces_staging.json
+++ b/api/config/featured_workspaces_staging.json
@@ -1,3 +1,13 @@
 {
-  "featuredWorkspaces": []
+  "featuredWorkspaces": [{
+    "name": "Featured1",
+    "namespace": "aou-rw-staging-99bc96d6",
+    "id": "featured1",
+    "category": "TUTORIAL_WORKSPACES"
+  }, {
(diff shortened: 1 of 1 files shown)


calbach:
Since it's Friday and no one is likely to care, I pushed it up manually to staging so you can see - attached screenshots to PR desc.

jaycarlton:
Thanks. I need to find a staging invite code.

jaycarlton:
Is the plan eventually to allow featuring and un-featuring workspaces without redeploying a json file? Not a priority obviously.

calbach:
No, I'm also not sure we would want to - this is a small set which will change infrequently. Here is the original design: https://docs.google.com/document/d/1lqfXpgupxSNCuwiJT4MOp8PBWTcU6k-2eryWnX6ZxIA/edit

"redeploying a json file": to be clear, these JSON configs can all be pushed out of band of a release if needed. By default, they are pushed with the release and the version in source control is the source of truth.

PR #2841, opened 2019-11-09
[RW-3897][risk=no] Update to latest snippets

calbach:
Corresponds to https://github.com/all-of-us/workbench-snippets/commit/1d31d845d4f4d31a359f6d3d37280efc0a287d3e

Generated per https://github.com/all-of-us/workbench/blob/9f6d83ab3b3bd4a5b92b388890a05b59aa83420a/api/cluster-resources/README.md#updating-snippets

Tested manually by copy-pasting snippets JSON into the snippets extension per https://github.com/all-of-us/workbench/blob/master/api/cluster-resources/README.md#local-testing-1 . Did not execute snippets since there were very few code changes included since previous release.

py-snippets:
image

py-datasets-snippets:
image

r-snippets:
image

r-datasets-snippets:
image


jaycarlton:
For traceability, would it make sense eventually to consider the snippets package as its own deliverable and give it a version number? Grabbing the git hash was a good move, but those can be a mouthful.

api/cluster-resources/py-dataset-snippets-menu.json
```diff
@@ -1,17 +1,17 @@
...
-    },
     {
       "name": "Documentation",
       "external-link": "https://github.com/all-of-us/workbench-snippets"
     },
     "---",
     {
-      "name": "Basic operations",
+      "name": "(1) Setup",
+      "snippet": "!pip3 install --upgrade --user pandas_profiling statsmodels\n\nimport os\nimport numpy as np\nimport pandas as pd\nimport pandas_profiling\nimport plotnine\nfrom plotnine import *  # Provides a ggplot-like interface to matplotlib.\n\n## Plot setup.\ntheme_set(theme_bw(base_size = 11)) # Default theme for plots.\n\ndef get_boxplot_fun_data(df):\n  \"\"\"Returns a data frame with a y position and a label, for use annotating ggplot boxplots.\n\n  Args:\n    d: A data frame.\n  Returns:\n    A data frame with column y as max and column label as length.\n  \"\"\"\n  d = {'y': max(df), 'label': f'N = {len(df)}'}\n  return(pd.DataFrame(data=d, index=[0]))\n\n# NOTE: if you get any errors from this cell, restart your kernel and run it again.\n"
```
jaycarlton: Can you please post a screenshot of the menu as rendered? Also, I wonder if we couldn't set things up such that the snippet field had a reference to a markdown file in the resources directory instead of doing it this way. 

Or if there is such a markdown file that's feeding into this, maybe that should be something we configure and this json file is an output generated at build time. Maybe overkill for right now, but looking at this it's really hard to say much intelligently about it. And yes, the time to say something was probably when it was merged into workbench-snippets anyway.

api/cluster-resources/py-dataset-snippets-menu.json
```diff
@@ -1,17 +1,17 @@
...
-      "snippet": "!pip3 install --upgrade --user pandas_profiling statsmodels\n\nimport os\nimport numpy as np\nimport pandas as pd\nimport pandas_profiling\nimport plotnine\nfrom plotnine import *  # Provides a ggplot-like interface to matplotlib.\n\n## Plot setup.\ntheme_set(theme_bw(base_size = 11)) # Default theme for plots.\n\ndef get_boxplot_fun_data(df):\n  \"\"\"Returns a data frame with a y position and a label, for use annotating ggplot boxplots.\n\n  Args:\n    d: A data frame.\n  Returns:\n    A data frame with column y as max and column label as length.\n  \"\"\"\n  d = {'y': max(df), 'label': f'N = {len(df)}'}\n  return(pd.DataFrame(data=d, index=[0]))\n\n# NOTE: if you get any errors from this cell, restart your kernel and run it again.\n"
-    },
     {
       "name": "Documentation",
       "external-link": "https://github.com/all-of-us/workbench-snippets"
     },
     "---",
     {
-      "name": "Basic operations",
+      "name": "(1) Setup",
```
jaycarlton: nit: Is there a way to avoid encoding the ordinals directly in the name field and having those applied at render time? That would make it easier to reorder things.

api/cluster-resources/py-dataset-snippets-menu.json
jaycarlton: nit: Content-wise, it would be nice if the comments for each package included a link to read more about them, e.g. to the matplotlib or pandas homepage.

api/cluster-resources/py-dataset-snippets-menu.json
jaycarlton: nit: There's probably a more descriptive name than "Setup". What's being set up here?

PR #2842, opened 2019-11-11
[RW-3829][risk=low] Revert ACL check logic

calbach:
This change was originally made as a performance optimization: https://github.com/all-of-us/workbench/pull/2522 . Now that we have partial field responses on `getWorkspace()`, this approach is probably the same as or faster than the `getWorkspaceAcl()` approach, and requires less custom logic on our side.

From looking at the code and surrounding context for the comment about "not using this for access checks", I'm comfortable using this as we are in our application (but have an email out to clarify), and at worst it looks equivalently authoritative to getWorkspaceAcl: https://github.com/broadinstitute/rawls/blob/e4983e78748dffcc444924ec2b3c96a9bcd1e06a/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L317-L330

Manual testing: published a workspace locally, verified a second user could view workspace resources and clone the workspace.


calbach:
Manually testing now

calbach:
Verified fix locally

PR #2844, opened 2019-11-11
[risk=no] Improve cleanBuffer CREATING -> ERROR transition logic

calbach:
Per slack discussion, comparing to the current time may not be correct given:
- large sets of billing projects (syncBillingProjectBuffer may not visit all projects within the deadline)
- Firecloud API errors, or extended outages

Transitions in either of these cases would result in leaked billing projects.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -1,13 +1,17 @@
 package org.pmiops.workbench.billing;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
 import com.google.common.hash.Hashing;
 import java.sql.Timestamp;
 import java.time.Clock;
-import java.time.temporal.ChronoUnit;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
 import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.inject.Provider;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.BillingProjectBufferEntryDao;
@@ -29,8 +33,8 @@ public class BillingProjectBufferService {
 
   private static final int SYNCS_PER_INVOCATION = 5;
   private static final int PROJECT_BILLING_ID_SIZE = 8;
-  private static final int CREATING_TIMEOUT_MINUTES = 60;
-  private static final int ASSIGNING_TIMEOUT_MINUTES = 10;
+  @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
+  @VisibleForTesting static final Duration ASSIGNING_TIMEOUT = Duration.ofMinutes(10);
 
   private final BillingProjectBufferEntryDao billingProjectBufferEntryDao;
   private final Clock clock;
@@ -132,22 +136,14 @@ public class BillingProjectBufferService {
   }
 
   public void cleanBillingBuffer() {
+    Instant now = clock.instant();
     Iterables.concat(
+            findExpiredCreatingEntries(now),
+            // For the ASSIGNING status monitor, we can simply filter by the current time as this
+            // status tracks internal state rather than the Firecloud status.
             billingProjectBufferEntryDao.findAllByStatusAndLastStatusChangedTimeLessThan(
-                DbStorageEnums.billingProjectBufferEntryStatusToStorage(BufferEntryStatus.CREATING),
-                new Timestamp(
-                    clock
-                        .instant()
-                        .minus(CREATING_TIMEOUT_MINUTES, ChronoUnit.MINUTES)
-                        .toEpochMilli())),
-            billingProjectBufferEntryDao.findAllByStatusAndLastStatusChangedTimeLessThan(
-                DbStorageEnums.billingProjectBufferEntryStatusToStorage(
-                    BufferEntryStatus.ASSIGNING),
-                new Timestamp(
-                    clock
-                        .instant()
-                        .minus(ASSIGNING_TIMEOUT_MINUTES, ChronoUnit.MINUTES)
-                        .toEpochMilli())))
+                DbStorageEnums.billingProjectBufferEntryStatusToStorage(BufferEntryStatus.ASSIGNING),
+                new Timestamp(now.minus(ASSIGNING_TIMEOUT).toEpochMilli())))
         .forEach(
             entry -> {
               log.warning(
@@ -158,6 +154,30 @@ public class BillingProjectBufferService {
               entry.setStatusEnum(BufferEntryStatus.ERROR, this::getCurrentTimestamp);
               billingProjectBufferEntryDao.save(entry);
             });
+  }
+
+  private List<DbBillingProjectBufferEntry> findExpiredCreatingEntries(Instant now) {
+    return billingProjectBufferEntryDao
+        .findAllByStatusAndLastStatusChangedTimeLessThan(
+            DbStorageEnums.billingProjectBufferEntryStatusToStorage(BufferEntryStatus.CREATING),
+            new Timestamp(now.minus(CREATING_TIMEOUT).toEpochMilli()))
+        .stream()
+        // Ensure that we've also synchronized the status since the deadline elapsed. This
+        // covers degenerate cases where syncBillingProjectStatus might be backlogged and
+        // hasn't caught up to check all the CREATING projects, or covers periods where
+        // the Firecloud API is unresponsive. Ideally we'd push this logic directly down
+        // into the DAO, but JPQL has poor support for date comparisons. Keep the above
+        // filter as well as a first pass filter to limit what we pull into memory.
+        .filter(entry -> entry.getLastStatusChangedTime() != null)
+        .filter(entry -> entry.getLastSyncRequestTime() != null)
+        .filter(
+            entry ->
+                Duration.between(
+                            entry.getLastStatusChangedTime().toInstant(),
+                            entry.getLastSyncRequestTime().toInstant())
+                        .toMillis()
+                    >= CREATING_TIMEOUT.toMillis())
+        .collect(Collectors.toList());
   }
 
   public DbBillingProjectBufferEntry assignBillingProject(DbUser user) {
--- a/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
@@ -142,7 +142,8 @@ public class BillingProjectBufferService {
             // For the ASSIGNING status monitor, we can simply filter by the current time as this
             // status tracks internal state rather than the Firecloud status.
             billingProjectBufferEntryDao.findAllByStatusAndLastStatusChangedTimeLessThan(
-                DbStorageEnums.billingProjectBufferEntryStatusToStorage(BufferEntryStatus.ASSIGNING),
+                DbStorageEnums.billingProjectBufferEntryStatusToStorage(
+                    BufferEntryStatus.ASSIGNING),
                 new Timestamp(now.minus(ASSIGNING_TIMEOUT).toEpochMilli())))
         .forEach(
             entry -> {
--- a/api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
+++ b/api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
@@ -24,6 +24,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import javax.inject.Provider;
 import javax.persistence.EntityManager;
 import org.junit.Before;
@@ -66,6 +68,14 @@ import org.springframework.transaction.annotation.Transactional;
 public class BillingProjectBufferServiceTest {
 
   private static final Instant NOW = Instant.now();
+  private static final Instant BEFORE_CREATING_TIMEOUT_ELAPSES =
+      NOW.plus(BillingProjectBufferService.CREATING_TIMEOUT.minusMinutes(1));
+  private static final Instant AFTER_CREATING_TIMEOUT_ELAPSED =
+      NOW.plus(BillingProjectBufferService.CREATING_TIMEOUT.plusMinutes(1));
+  private static final Instant BEFORE_ASSIGNING_TIMEOUT_ELAPSES =
+      NOW.plus(BillingProjectBufferService.ASSIGNING_TIMEOUT.minusMinutes(1));
+  private static final Instant AFTER_ASSIGNING_TIMEOUT_ELAPSED =
+      NOW.plus(BillingProjectBufferService.ASSIGNING_TIMEOUT.plusMinutes(1));
   private static final FakeClock CLOCK = new FakeClock(NOW, ZoneId.systemDefault());
 
   @TestConfiguration
@@ -103,6 +113,8 @@ public class BillingProjectBufferServiceTest {
     workbenchConfig.billing.projectNamePrefix = "test-prefix";
     workbenchConfig.billing.bufferCapacity = (int) BUFFER_CAPACITY;
     workbenchConfig.billing.bufferRefillProjectsPerTask = 1;
+
+    CLOCK.setInstant(NOW);
 
     billingProjectBufferEntryDao = spy(billingProjectBufferEntryDao);
     TestLock lock = new TestLock();
@@ -354,9 +366,6 @@ public class BillingProjectBufferServiceTest {
         .getBillingProjectStatus(capturedProjectNames.get(2));
 
     billingProjectBufferService.syncBillingProjectStatus();
-    billingProjectBufferService.syncBillingProjectStatus();
-    billingProjectBufferService.syncBillingProjectStatus();
-    billingProjectBufferService.syncBillingProjectStatus();
 
     assertThat(
             billingProjectBufferEntryDao.countByStatus(
@@ -478,13 +487,65 @@ public class BillingProjectBufferServiceTest {
   @Test
   public void cleanBillingBuffer_creating() {
     DbBillingProjectBufferEntry entry = new DbBillingProjectBufferEntry();
+    entry.setFireCloudProjectName("foo");
     entry.setStatusEnum(BufferEntryStatus.CREATING, this::getCurrentTimestamp);
     entry.setCreationTime(getCurrentTimestamp());
     billingProjectBufferEntryDao.save(entry);
 
-    CLOCK.setInstant(NOW.plus(61, ChronoUnit.MINUTES));
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT_ELAPSED);
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.cleanBillingBuffer();
+    assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
+        .isEqualTo(BufferEntryStatus.ERROR);
+  }
+
+  @Test
+  public void cleanBillingBuffer_creatingNeverSynced() {
+    DbBillingProjectBufferEntry entry = new DbBillingProjectBufferEntry();
+    entry.setFireCloudProjectName("foo");
+    entry.setStatusEnum(BufferEntryStatus.CREATING, this::getCurrentTimestamp);
+    entry.setCreationTime(getCurrentTimestamp());
+    billingProjectBufferEntryDao.save(entry);
+
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT_ELAPSED);
     billingProjectBufferService.cleanBillingBuffer();
 
+    // Time has elapsed but no status sync has occurred, should not transition.
+    assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
+        .isEqualTo(BufferEntryStatus.CREATING);
+  }
+
+  @Test
+  public void cleanBillingBuffer_creatingDelayedSync() {
+    DbBillingProjectBufferEntry entry = new DbBillingProjectBufferEntry();
+    entry.setFireCloudProjectName("foo");
+    entry.setStatusEnum(BufferEntryStatus.CREATING, this::getCurrentTimestamp);
+    entry.setCreationTime(getCurrentTimestamp());
+    billingProjectBufferEntryDao.save(entry);
+
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+    billingProjectBufferService.syncBillingProjectStatus();
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT_ELAPSED);
+    billingProjectBufferService.cleanBillingBuffer();
+
+    // Time has elapsed but no status sync has occurred, should not transition.
+    assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
+        .isEqualTo(BufferEntryStatus.CREATING);
+
+    // Sync again, now after the timeout - should transition.
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.cleanBillingBuffer();
     assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
         .isEqualTo(BufferEntryStatus.ERROR);
   }
@@ -496,7 +557,7 @@ public class BillingProjectBufferServiceTest {
     entry.setCreationTime(getCurrentTimestamp());
     billingProjectBufferEntryDao.save(entry);
 
-    CLOCK.setInstant(NOW.plus(11, ChronoUnit.MINUTES));
+    CLOCK.setInstant(AFTER_ASSIGNING_TIMEOUT_ELAPSED);
     billingProjectBufferService.cleanBillingBuffer();
 
     assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -137,13 +140,34 @@ public void syncBillingProjectStatus() {
...
+                    false)
+                // Ensure that we've also synchronized the status since the deadline elapsed. This
+                // covers degenerate cases where syncBillingProjectStatus might be backlogged and
+                // hasn't caught up to check all the CREATING projects, or covers periods where
+                // the Firecloud API is unresponsive. Ideally we'd push this logic directly down
+                // into the DAO, but JPQL has poor support for date comparisons. Keep the above
+                // filter as well as a first pass filter to limit what we pull into memory.
+                .filter(
+                    entry ->
+                        entry.getLastSyncRequestTime() != null
```
jaycarlton: nit: I'd do `Optional.ofNullable(entry.getLastSyncRequestTime()).map(...).flatMap(Streams::stream)` (to squash out the empty optionals). I suppose this is primarily a style preference though, but I feel like null checks are a bit of a smell in stream logic.

Better yet, I'd likely just make a helper function like `isBufferEntryInTimespan()` and filter on that using a method reference. This would also make the filter function(s) separately testable.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -137,13 +140,34 @@ public void syncBillingProjectStatus() {
 
   public void cleanBillingBuffer() {
     Iterables.concat(
```
jaycarlton: nit: I'm finding this hard to follow, since the arguments to concat() are so involved and far apart. Why not declare two lists separately and concat them at the end? Seems like that's a win for debuggability too.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +479,26 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+    billingProjectBufferService.syncBillingProjectStatus();
+
     CLOCK.setInstant(NOW.plus(61, ChronoUnit.MINUTES));
     billingProjectBufferService.cleanBillingBuffer();
 
+    // Time has elapsed but no status sync has occurred, should not transition.
+    assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
```
jaycarlton: I'm not sure how much of your logic is really getting exercised here. Do you only have a single entry with CREATING status? I think you'd want cases where (1) they're all before the time range starts, (2) some are on each side (3) one is exactly on the border, etc. That is, simulate a stream of them and operate on that.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +479,26 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
+        .isEqualTo(CREATING);
+
+    // Sync again, now @ +61 minutes - should transition.
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.cleanBillingBuffer();
     assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
```
jaycarlton: Are you asserting that the first one you find is an error? Should they not all be in the error state?

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: Thanks, I think the boundary handling is the easiest place to write a bug, so that's good to test.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: So I think part of the issue is that I trained myself to try to read stream statements in one mental "bite" but I can't really do that here. Which may be fine. In other words, if most code is stream operations, this will become more natural.

The biggest issue reading it for me was that I couldn't see the whole statement at once on my laptop, so I have to remember context (like the iterables.concat() around everything), try to find the comma, etc.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -136,21 +141,15 @@ public void syncBillingProjectStatus() {
...
-                DbStorageEnums.billingProjectBufferStatusToStorage(CREATING),
-                new Timestamp(
-                    clock
-                        .instant()
-                        .minus(CREATING_TIMEOUT_MINUTES, ChronoUnit.MINUTES)
-                        .toEpochMilli())),
+            findExpiredCreatingEntries(now),
+            // For the ASSIGNING status monitor, we can simply filter by the current time as this
+            // status tracks internal state rather than the Firecloud status.
             billingProjectBufferEntryDao.findAllByStatusAndLastStatusChangedTimeLessThan(
```
jaycarlton: nit: Aesthetically I might change "TimeLessThan" to "EarlierThan" in this name. Just because I don't think of wall clock time as something comparable in that way.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -163,6 +162,32 @@ public void cleanBillingBuffer() {
...
   }
 
+  private List<DbBillingProjectBufferEntry> findExpiredCreatingEntries(Instant now) {
+    return StreamSupport.stream(
+            billingProjectBufferEntryDao
+                .findAllByStatusAndLastStatusChangedTimeLessThan(
+                    DbStorageEnums.billingProjectBufferStatusToStorage(CREATING),
+                    new Timestamp(
+                        now.minus(CREATING_TIMEOUT_MINUTES, ChronoUnit.MINUTES).toEpochMilli()))
+                .spliterator(),
```
jaycarlton: IntelliJ tells me you can just do 
```
  private List<DbBillingProjectBufferEntry> findExpiredCreatingEntries(Instant now) {
    return billingProjectBufferEntryDao
        .findAllByStatusAndLastStatusChangedTimeLessThan(
            DbStorageEnums.billingProjectBufferStatusToStorage(CREATING),
            new Timestamp(
                now.minus(CREATING_TIMEOUT_MINUTES, ChronoUnit.MINUTES).toEpochMilli()))
        .stream()
// ...
```

jaycarlton: also nit: I'd probably declare the cutoff timestamp separately, since it's a  slightly involved construction and calculation.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +481,26 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
     entry.setStatusEnum(CREATING, this::getCurrentTimestamp);
     entry.setCreationTime(getCurrentTimestamp());
     billingProjectBufferEntryDao.save(entry);
 
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+    billingProjectBufferService.syncBillingProjectStatus();
+
     CLOCK.setInstant(NOW.plus(61, ChronoUnit.MINUTES));
```
jaycarlton: I'd define the number of minutes in terms of the  grace period instead of just saying 61. It would be easy to change the setting and have the test get out of date (though in most scenarios hopefully the test would start breaking).

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +481,26 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
     billingProjectBufferService.cleanBillingBuffer();
 
+    // Time has elapsed but no status sync has occurred, should not transition.
+    assertThat(billingProjectBufferEntryDao.findOne(entry.getId()).getStatusEnum())
+        .isEqualTo(CREATING);
+
+    // Sync again, now @ +61 minutes - should transition.
+    billingProjectBufferService.syncBillingProjectStatus();
```
jaycarlton: Should we always call sync before clean? If so, why not have the latter call the former for us? 

I'm starting to think this service may be exhibiting Temporal Coupling. It's hard to understand what the invariants are w/r/t the billing buffer.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -163,6 +162,32 @@ public void cleanBillingBuffer() {
...
+        // into the DAO, but JPQL has poor support for date comparisons. Keep the above
+        // filter as well as a first pass filter to limit what we pull into memory.
+        .filter(entry -> entry.getLastSyncRequestTime() != null)
+        .filter(
+            entry ->
+                Duration.between(
+                            entry.getLastStatusChangedTime().toInstant(),
+                            entry.getLastSyncRequestTime().toInstant())
+                        .toMinutes()
+                    >= CREATING_TIMEOUT_MINUTES)
```
jaycarlton: Yeah, if your specified timeout is itself a Duration, then you don't have to get into units here. I don't  know if the config system supports durations yet though.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -163,6 +162,32 @@ public void cleanBillingBuffer() {
...
+                        now.minus(CREATING_TIMEOUT_MINUTES, ChronoUnit.MINUTES).toEpochMilli()))
+                .spliterator(),
+            false)
+        // Ensure that we've also synchronized the status since the deadline elapsed. This
+        // covers degenerate cases where syncBillingProjectStatus might be backlogged and
+        // hasn't caught up to check all the CREATING projects, or covers periods where
+        // the Firecloud API is unresponsive. Ideally we'd push this logic directly down
+        // into the DAO, but JPQL has poor support for date comparisons. Keep the above
+        // filter as well as a first pass filter to limit what we pull into memory.
+        .filter(entry -> entry.getLastSyncRequestTime() != null)
```
jaycarlton: do we not also need to check `getLastStatusChangedTime()` for nullity?

Also, could you put a predicate method on `DbBillingProjectBufferEntry` that does this computation  as a derived property? Then it's just `.filter(DbBillingProjectBufferEntry::isInExpiredCreatingState)`. (May want to pass in the timestamp for testing though.)

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +481,26 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
   @Test
   public void cleanBillingBuffer_creating() {
     DbBillingProjectBufferEntry entry = new DbBillingProjectBufferEntry();
+    entry.setFireCloudProjectName("foo");
     entry.setStatusEnum(CREATING, this::getCurrentTimestamp);
     entry.setCreationTime(getCurrentTimestamp());
     billingProjectBufferEntryDao.save(entry);
 
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
```
jaycarlton: +1 for `doReturn()`. It's especially important if you need to return `CompletableFuture`s.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: I don't see where this entry was ever in a state other than CREATING.

If the point is that it stays in creating, then it would be good to see a test with a different timeout such that the opposite is true. I'm still getting context on the overall mission here, so I may be confused.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -529,6 +544,55 @@ public void cleanBillingBuffer_ignoreValidEntries() {
...
+    billingProjectBufferService.bufferBillingProjects();
+
+    CLOCK.setInstant(NOW.plus(30, ChronoUnit.MINUTES));
+    billingProjectBufferService.bufferBillingProjects();
+    billingProjectBufferService.bufferBillingProjects();
+
+    final ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
+    verify(fireCloudService, times(3)).createAllOfUsBillingProject(captor.capture());
+    final List<String> capturedProjectNames = captor.getAllValues();
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
```
jaycarlton: The need to have a mocked return paired with each input may indicate that we'd benefit from a stand-along test fixture that could take in a map (for example). If there were such a fixture for the billing buffer, I might have a small use for it in the billing status PR test.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -529,6 +544,55 @@ public void cleanBillingBuffer_ignoreValidEntries() {
...
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(0));
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(1));
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.ERROR))
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(2));
+
+    Supplier<List<BillingProjectBufferStatus>> statuses =
```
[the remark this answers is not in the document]
jaycarlton: I like it.

I'd probably name  this something like `capturedProjectsStatusSupplier` so it sticks out.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: sync processes a limited number of entries on every run. Yes, it was exhibiting Temporal Coupling, though probably not intentionally. That is the bug this PR is fixing. I'm not aware of any more instances of that pattern in the current buffer system.

Here is the original design doc: http://broad.io/1ppw, it doesn't contain any details about the different cronjobs. Brief summary which we should probably writeup as a documentation task:

Crons:
bufferBillingProjects (every minute): create new entries in the buffer, always in the CREATING state
syncBillingProjectStatus (every minute): poll for state changes on the 5 least recently synced entries in a relevant state, namely tracking transitions from CREATING --> ACTIVE or ERROR
cleanBillingBuffer (every hour): clean bad entries out of the buffer, e.g. things that have been in a particular state for too long; this is necessary as the buffer has a fixed capacity

Methods (buffer consumer):
create/clone workspace: mark an entry as ASSIGNING, then ASSIGNED when used for a workspace. If we fail in the middle, cleanBillingBuffer cleans out entries stuck in ASSIGNING

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -136,21 +141,15 @@ public void syncBillingProjectStatus() {
...
+            // For the ASSIGNING status monitor, we can simply filter by the current time as this
+            // status tracks internal state rather than the Firecloud status.
             billingProjectBufferEntryDao.findAllByStatusAndLastStatusChangedTimeLessThan(
                 DbStorageEnums.billingProjectBufferStatusToStorage(ASSIGNING),
                 new Timestamp(
-                    clock
-                        .instant()
-                        .minus(ASSIGNING_TIMEOUT_MINUTES, ChronoUnit.MINUTES)
-                        .toEpochMilli())))
+                    now.minus(ASSIGNING_TIMEOUT_MINUTES, ChronoUnit.MINUTES).toEpochMilli())))
```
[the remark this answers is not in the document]
calbach: Yes - good suggestion.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: It starts in CREATING and then it moves to ERROR when the time limit expires. That second status check happens later at L505. I split this into two tests to cover the normal path, and the delayed sync path separately

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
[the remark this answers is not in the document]
calbach: Personal preference, but I don't think it's necessary to go full Java verbosity mode for a tightly scoped local variable. Will at least call it `statusSupplier` to make it a little clearer though.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: An in-memory fake for the Firecloud API would probably be useful.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: Always need to balance this against test complexity, but in this situation I think it's an improvement - done

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +489,45 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
+    entry.setCreationTime(getCurrentTimestamp());
+    billingProjectBufferEntryDao.save(entry);
+
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+    billingProjectBufferService.syncBillingProjectStatus();
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT);
+    billingProjectBufferService.cleanBillingBuffer();
```
ericsong: do we have a test case that verifies entries are not removed if `sync` has never been called? It feels like a slightly different test case.

calbach: True, worth covering - added.

calbach:
FYI @jaycarlton I'll wait for reapproval from you here since there were a fair number of changes/replies

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: OK, I guess lots of folks prefer smart services and dumb entities.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +489,65 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
+    entry.setFireCloudProjectName("foo");
+    entry.setStatusEnum(CREATING, this::getCurrentTimestamp);
+    entry.setCreationTime(getCurrentTimestamp());
+    billingProjectBufferEntryDao.save(entry);
+
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT);
```
jaycarlton: nit: "after creating timeout" can be parsed two ways. I initially read it as after (creating timeout), meaning after it was started, rather than after it was expired. Maybe `AFTER_CREATING_TIMEOUT_ELAPSED` would be clearer.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: Awesome, thanks. That's helpful. One thing that jumps out is, why don't we consolidate sync with clean, since both are concerned with updating status based on elapsed time.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: Let's start making a list of such things... It'd be good to settle on some fake design principles potentially.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -506,27 +568,76 @@ public void cleanBillingBuffer_assigning() {
...
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(0));
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(1));
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.ERROR))
+        .when(fireCloudService)
+        .getBillingProjectStatus(capturedProjectNames.get(2));
+
+    Supplier<List<BillingProjectBufferStatus>> statusSupplier =
```
jaycarlton: Since this is a fairly new pattern, a little comment here would be helpful I think.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: That's probably good enough (since you are familiar with the implementation), but I'd want to make sure it's operating on all and only the entries it's supposed to. It's easy to imagine a bug that happened only when more than 1 (or 5) entries were present.

Not a task for this PR though.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: In theory you could combine all three of these cron jobs. They all have different roles, and I think it is a better design to allow these pieces to operate independently both for failure isolation, Firecloud load, and to allow entries to move quickly through the state machine. cleanBillingBuffer is also not scoped exclusively to Firecloud project status changes

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: That's the other test case that I added in response to your other comment: `cleanBillingBuffer_multipleCreates`. I'm not sure what more you're looking for in this test case.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: Done

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: Done

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: Oh, I'm sorry. I lost track of what was new.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -33,8 +37,8 @@
 
   private static final int SYNCS_PER_INVOCATION = 5;
   private static final int PROJECT_BILLING_ID_SIZE = 8;
-  private static final int CREATING_TIMEOUT_MINUTES = 60;
-  private static final int ASSIGNING_TIMEOUT_MINUTES = 10;
+  @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
```
jaycarlton: 👏

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: It occurs to me that if we need to add anything else to that enum, we should probably break out the OK/ERROR status into a separate one, and last known good status in the known transition of states. As it stands, we don't have a way to indicate more than one kind of error (e.g. FAILED_TO_GARBAGE_COLLECT). Right now everything in ERROR is assumed to be a creation failure from what I can tell. I also tend to add a catch-all UNKNOWN status for things being migrated or batch processed to indicated unexpected exceptions, interruptions, etc.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -479,13 +489,65 @@ public void assignBillingProject_locking() throws InterruptedException, Executio
...
+    entry.setCreationTime(getCurrentTimestamp());
+    billingProjectBufferEntryDao.save(entry);
+
+    doReturn(new BillingProjectStatus().creationStatus(CreationStatusEnum.CREATING))
+        .when(fireCloudService)
+        .getBillingProjectStatus(entry.getFireCloudProjectName());
+
+    CLOCK.setInstant(AFTER_CREATING_TIMEOUT_ELAPSED);
+    billingProjectBufferService.syncBillingProjectStatus();
+    billingProjectBufferService.cleanBillingBuffer();
```
jaycarlton: If we expect, in reality, that some more time elapses between the sync and the clean, then it wouldn't hurt to advance the clock some in between these two events. It could potentially catch a bug if somehow things were better when the two events were nearly simultaneous. (Though I imagine it's more common for race conditions to happen when events like this are happening near each other.)

Bonus question: is it safe for more than one of these crons to be operating on an entry at the same time? I haven't looked at how they're scheduled to see if that's possible.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -160,6 +156,30 @@ public void cleanBillingBuffer() {
...
+        // into the DAO, but JPQL has poor support for date comparisons. Keep the above
+        // filter as well as a first pass filter to limit what we pull into memory.
+        .filter(entry -> entry.getLastStatusChangedTime() != null)
+        .filter(entry -> entry.getLastSyncRequestTime() != null)
+        .filter(
+            entry ->
+                Duration.between(
+                            entry.getLastStatusChangedTime().toInstant(),
+                            entry.getLastSyncRequestTime().toInstant())
+                        .toMillis()
```
jaycarlton: nit: if you prefer, [there's also `Duration.compareTo()`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#compareTo-java.time.Duration-), which I think can be woven directly into the filter somehow without converting both to millis.

calbach: I had this initially, but decided it was unreadable; directly comparing these values with a >= sign seems much better to me (would be nice if there were another Duration method or util to do this directly).

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
calbach: I'd rather keep these tests lean and on-topic. These crons are intended to be safe to operate in parallel (and yes, it is possible). In my view this is sufficient to cover the regression as written, though clearly you could go much further here and add coverage of all permutations of clean/sync ordering with varying time between each.

PR #2845, opened 2019-11-11
[RW-3963][RISK=NO] Billing Buffer Status Endpoint

jaycarlton:
To assist with perf testing, I am adding a simple endpoint to retrieve statistics about the billing buffer. Currently, only billing buffer size is supported, but it should be really easy to add fields here). Added a trivial unit test, but then again the function isn't that complicated to begin with.
To test this, use `curl -X GET http://localhost:8081/v1/cron/billingProjectBufferStatus | jq`.

I added a new type named `BillingProjectBufferStatus`, which was named identically to an existing enum. However, since that enum was scoped inside a similarly-named class, and since it didn't indicate it was describing entries instead of buffers, I decided to rename it. (On top of the confusion I added here, it's possible to read it two ways, i.e. "the status of the billing project buffer" or "the buffer status of the billing project". It's possible both these names could be better.

That accounts for most of the noise below. I also qualified statically imported enum values, which I think is a fairly confusing practice in most cases.


api/src/main/resources/workbench.yaml
```diff
@@ -1165,6 +1165,25 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
+  /v1/cron/billingProjectBufferStatus:
+    get:
+      security: []
```
calbach: This makes it so you can call the endpoint unauthenticated. I don't think we want that. It's not particularly sensitive, but I'm not sure I want internal state details to be publicly readable either. Requiring authentication seems like a reasonable roadblock to me.

api/src/main/resources/workbench.yaml
```diff
@@ -1165,6 +1165,25 @@ paths:
...
+      tags:
+        - offlineBilling
+      description: If the AoU Billing Project buffer is not full, refill with one or more billing projects.
```
calbach: Update

api/src/main/resources/workbench.yaml
```diff
@@ -1165,6 +1165,25 @@ paths:
...
+      operationId: getBillingProjectBufferStatus
+      responses:
+        200:
+          description: No Error
+          schema:
+            $ref: '#/definitions/BillingProjectBufferStatus'
+        204:
```
calbach: Seems n/a

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -224,4 +221,13 @@ private long getCurrentBufferSize() {
   private int getBufferMaxCapacity() {
     return workbenchConfigProvider.get().billing.bufferCapacity;
   }
+
+  public BillingProjectBufferStatus getStatus() {
+    final long bufferSize =
+        billingProjectBufferEntryDao.countByStatus(
+            DbStorageEnums.billingProjectBufferEntryStatusToStorage(BufferEntryStatus.AVAILABLE));
+    final BillingProjectBufferStatus result = new BillingProjectBufferStatus();
```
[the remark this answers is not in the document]
jaycarlton: Oh cool, i forget which ones daisy chain.

api/src/main/java/org/pmiops/workbench/billing/OfflineBillingController.java
```diff
@@ -51,4 +52,10 @@
     freeTierBillingService.checkFreeTierBillingUsage();
     return ResponseEntity.noContent().build();
   }
+
+  @Override
+  public ResponseEntity<BillingProjectBufferStatus> getBillingProjectBufferStatus() {
+    final BillingProjectBufferStatus response = billingProjectBufferService.getStatus();
```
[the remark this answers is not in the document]
jaycarlton: Yeah, since there's no arguments to `getStatus()` I think you're right.

api/src/main/resources/workbench.yaml
```diff
@@ -1165,6 +1165,25 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
+  /v1/cron/billingProjectBufferStatus:
```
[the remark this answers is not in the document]
jaycarlton: This sounds like a can of worms for another day. I'm thinking of possibly having ops- or monitoring-specific roots soon under one of Scout's projects, which may be a better home for this one.

api/src/main/resources/workbench.yaml
jaycarlton: Oh good catch. That's not what I thought that did. 

So if I want regular auth (but not admin-only), then I just drop the `security` and it inherits `aou_oauth`. That's what you meant by inheriting the default?

jaycarlton: After removing the `security: []` line, I started seeing [this ClassNotFound exception](https://gist.github.com/jaycarlton/c5552ecd77ffca86370a764479f93e9f). Restoring it was the only thing that fixed it. I'm wondering if maybe the security settings aren't at path-level granularity but rather that I need a new controller if I want a different auth setup. It's easy enough to make a new `OpsController` or whatever.

calbach: Correct.

RE ClassNotFound: That's a result of this nasty piece of code that assumes all controllers live in the API directory: https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java#L220-L222 , it works for the others here since they are unauthenticated crons (only callable via GAE).

I would just split it into an monitoring/ops controller by changing the tag and putting it in the /api directory - that's probably the easiest fix (though separately we eventually need to improve that auth interceptor code)

jaycarlton: Yes. I'd write it up but I don't quite understand it. I didn't see any references to directories in the snippet you highlighted itself.

Another benefit of that fix would be to allow controllers to live in packages that are vertically integrated where that makes sense.

calbach: Ah sorry - I needed to go one level deeper to find the true nastiness I was referring to:
https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/interceptors/InterceptorUtils.java#L21-L38

You can alternatively add the controller name to this map to have it work:
https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/interceptors/InterceptorUtils.java#L11-L16

jaycarlton: OK, that should work. I'm finding several places where naming conventions are mattering in ways I wouldn't expect them to.

api/src/main/resources/workbench.yaml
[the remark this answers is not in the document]
jaycarlton: Yeah, this is more about monitoring than operations, unless we need to expose a methods to evict projects form the buffer or something.

Also, I'm wondering whether it would make sense to create a new Authority for devops, separate from what support needs. I.e. we don't need to adjust individual users in most cases, but there are things we may need to do that support or other roles needn't have access to. Does the authority system currently assume a strictly totem-pole hierarchy, or can we have mutually exclusive privileges?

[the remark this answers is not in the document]
calbach: The latter: https://github.com/all-of-us/workbench/blob/048a8a0744f7421352f315fc8ff0d0f70e0eecb9/api/src/main/resources/workbench.yaml#L2262-L2271

There's an argument that the UI might want to call this to show some UX on the workspace page, so I wouldn't bother with an authority for now unless a broader logical grouping emerges.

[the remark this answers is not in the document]
jaycarlton: Sounds good for now, thanks. Was just thinking down the road if we start adding more specialized endpoints.

PR #2851, opened 2019-11-13
[risk=no][RW-3843] Refactor out usages of @VisibleForTesting for Spring Providers

ericsong:
We found a better way to dynamically change the value provided by Spring Providers in tests that no longer relies on exposing a test only method on our controllers.

To expand on the point above, we used to use a pattern of exposing a `setXProvider` methods on our service/controller classes for the sole purpose of being able to modify the provided value in tests.
ex. `@VisibleForTesting setUserProvider(Provider<User> userProvider) {}`.

We would use this in our tests by creating a provider to return the value we wanted to test and calling the `setUserProvider` method with our mocked provider.

We believe this was necessary because Spring would essentially treat values returned through the `Provider` mechanism as an immutable singleton, so it would cache the initialized value and return that as needed. The `@Scope("prototype")` annotation tells Spring that we can have multiple values for a given Bean so it should fetch the latest version.

^This might be an oversimplification of what's going on but I think its a valid mental model for how Spring DI is behaving within our tests.

https://docs.spring.io/spring/docs/3.0.0.M3/reference/html/ch04s04.html

Most of the changes are regarding the above but I took some liberty to refactor code smells I found along the way including...
- not using `Autowiring` for object initialization
- other usages of unnecessary setting/VisibleForTest

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/CdrVersionsControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/CdrVersionsControllerTest.java
@@ -16,7 +16,6 @@ import org.pmiops.workbench.exceptions.ForbiddenException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.model.CdrVersionListResponse;
 import org.pmiops.workbench.model.DataAccessLevel;
-import org.pmiops.workbench.test.Providers;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
@@ -25,6 +24,7 @@ import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Scope;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.annotation.DirtiesContext.ClassMode;
 import org.springframework.test.context.junit4.SpringRunner;
@@ -45,17 +45,16 @@ public class CdrVersionsControllerTest {
 
   private CdrVersion defaultCdrVersion;
   private CdrVersion protectedCdrVersion;
-  private DbUser user;
+  private static DbUser user;
 
   @TestConfiguration
   @Import({CdrVersionService.class, CdrVersionsController.class})
   @MockBean({FireCloudService.class})
   static class Configuration {
     @Bean
+    @Scope("prototype")
     public DbUser user() {
-      // Allows for wiring of the initial Provider<User>; actual mocking of the
-      // user is achieved via setUserProvider().
-      return null;
+      return user;
     }
 
     @Bean
@@ -68,7 +67,7 @@ public class CdrVersionsControllerTest {
   public void setUp() {
     user = new DbUser();
     user.setDataAccessLevelEnum(DataAccessLevel.REGISTERED);
-    cdrVersionsController.setUserProvider(Providers.of(user));
+
     defaultCdrVersion =
         makeCdrVersion(
             1L, /* isDefault */ true, "Test Registered CDR", 123L, DataAccessLevel.REGISTERED);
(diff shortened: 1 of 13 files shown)


api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
```diff
@@ -183,18 +181,21 @@ public long getConceptId() {
...
+    @Bean
+    DbUser user() {
+      return user;
+    }
   }
 
   @Before
   public void setUp() {
-    DbUser user = new DbUser();
+    user = new DbUser();
```
jaycarlton: BTW, If you wanted, because the user object is static, you could use a `@BeforeClass` method and only initialize it once. I'm not sure that's a win, though, since you're doing dao things.

gjuggler:
Could you clarify in the PR description which part is the most meaningful change here? I'm just following along out of interest, and AFAICT the key change is the @Scope("prototype"), but I'm not 100% certain since there's a lot of unrelated deleted / modified test code.

api/src/test/java/org/pmiops/workbench/conceptset/ConceptSetServiceTest.java
```diff
@@ -30,22 +31,20 @@
...
-  private DbWorkspace workspace;
-
-  @Before
-  public void setUp() {
-    conceptSetService = new ConceptSetService();
-    conceptSetService.setConceptSetDao(conceptSetDao);
-    workspace = mockWorkspace();
-  }
+  @TestConfiguration
+  @Import({
```
als364: Nit: Can you collapse this to one line

api/src/test/java/org/pmiops/workbench/api/CdrVersionsControllerTest.java
```diff
@@ -45,17 +45,16 @@
...
   private CdrVersion protectedCdrVersion;
-  private DbUser user;
+  private static DbUser user;
 
   @TestConfiguration
   @Import({CdrVersionService.class, CdrVersionsController.class})
   @MockBean({FireCloudService.class})
   static class Configuration {
     @Bean
+    @Scope("prototype")
```
als364: Can you please put an explanation in the PR description or maybe even a small inline comment saying what this does vs the default singleton scope?

jaycarlton: I'm not sure we need a comment everywhere we use a language or framework feature that isn't as common as the others to describe how the feature works if it's well documented publicly. But as for *why* this one should be a prototype, I agree that would make a good comment. 

You can also use the constant `org.springframework.beans.factory.config.ConfigurableBeanFactory#SCOPE_PROTOTYPE` instead of a string literal.

api/src/test/java/org/pmiops/workbench/api/CdrVersionsControllerTest.java
```diff
@@ -45,17 +45,16 @@
...
     public DbUser user() {
-      // Allows for wiring of the initial Provider<User>; actual mocking of the
-      // user is achieved via setUserProvider().
-      return null;
+      return user;
```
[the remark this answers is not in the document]
jaycarlton: Actually I chickened out. It's [still there](https://github.com/all-of-us/workbench/blob/f97fe7154847287ae1fa965ca2c9e727da7f654f/api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java#L125) but probably didn't get picked up here b/c it's not an injected provider. The test is merely overwriting the class member with the same value.

Do you mind killing this one too?

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
ericsong: I generally prefer to put setup stuff in `Before` instead of `BeforeClass` so each method can start with the same clean state.

`BeforeClass` might be a good choice if I'm constructing something really heavy or immutable.

In this case, mock models are meant to be modified as needed so we need to reinitialize the model or modifications from a previous test will bleed into the following.

api/src/test/java/org/pmiops/workbench/conceptset/ConceptSetServiceTest.java
ericsong: 👍

ericsong:
@gjuggler @als364 @jmthibault79 expanded on what I was trying to do in the description

ericsong:
LMK if that makes sense! Will merge if so.

als364:
Yes, makes sense - thank you!

PR #2852, opened 2019-11-14
[no ticket][risk=no] Utility functions for prettier comparisons of Comparable implementers

jaycarlton:
Unfortunately Java has no operator overloading or infix functions, so static methods on a utility class are the best I can do. Continuing discussion from [here](https://github.com/all-of-us/workbench/pull/2844#discussion_r346055803).


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -171,13 +172,11 @@ public void cleanBillingBuffer() {
         // filter as well as a first pass filter to limit what we pull into memory.
         .filter(entry -> entry.getLastStatusChangedTime() != null)
         .filter(entry -> entry.getLastSyncRequestTime() != null)
-        .filter(
-            entry ->
+        .filter(entry -> ComparableUtils.isGreaterThanOrEqualTo(
```
jaycarlton: I thought about switching this so it's a <= comparison, which are easier to read (especially if there are multiple ones).

api/src/main/java/org/pmiops/workbench/utils/ComparableUtils.java
```diff
@@ -0,0 +1,28 @@
+package org.pmiops.workbench.utils;
+
+/**
+ * A few generic utilities to hopefully make Comparators less ugly
+ */
+public class ComparableUtils {
+
+  public static <T extends Comparable<T>> Boolean isLessThan(T one, T other) {
```
jaycarlton: I thought about putting the generic stuff at the class level, but was worried that would mean I have to qualify the type explicitly at the call site.

api/src/main/java/org/pmiops/workbench/utils/ComparableUtils.java
```diff
@@ -0,0 +1,28 @@
+package org.pmiops.workbench.utils;
+
+/**
+ * A few generic utilities to hopefully make Comparators less ugly
+ */
+public class ComparableUtils {
```
jaycarlton: A better name would be better.

api/src/main/java/org/pmiops/workbench/utils/Comparables.java
```diff
@@ -0,0 +1,25 @@
...
+  public static <T extends Comparable<T>> Boolean isLessThan(T one, T other) {
+    return one.compareTo(other) < 0;
+  }
+
+  public static <T extends Comparable<T>> Boolean isLessThanOrEqualTo(T one, T other) {
+    return one.compareTo(other) <= 0;
+  }
+
+  public static <T extends Comparable<T>> Boolean isEqualValueTo(T one, T other) {
+    return 0 <= one.compareTo(other);
```
jaycarlton: woops...

jaycarlton:
Incorporated into #2857

PR #2853, opened 2019-11-14
[risk=low][RW-3982] Create project.rb tool to get workspace details from project ID

ericsong:
This adds a new project.rb tool which can take a Firecloud Project ID or workspace namespace and output workspace details (creator + collaborator emails).

Example Input
`./project.rb fetch-workspace-details --project all-of-us-workbench-test --projectId=aou-rw-test-3f23b78e`

Example output
```
***
***
***
Workspace Name: Should be seen
Workspace Namespace: aou-rw-test-3f23b78e
Creator: ericsong3@fake-research-aou.org
Collaborators:
        ericsong3@fake-research-aou.org (OWNER)
        ericsong@fake-research-aou.org (WRITER)
***
***
***
```

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -627,6 +627,16 @@ task fetchFireCloudUserProfile(type: JavaExec) {
   }
 }
 
+// See project.rb command: fetch-workspace-details
+task fetchWorkspaceDetails(type: JavaExec) {
+  classpath = sourceSets.tools.runtimeClasspath
+  main = "org.pmiops.workbench.tools.FetchWorkspaceDetails"
+  systemProperties = commandLineSpringProperties
+  if (project.hasProperty("appArgs")) {
+    args Eval.me(appArgs)
+  }
+}
+
 // This task is called from a few different places:
 // - devstart.rb > load_config (used by "deploy" and "update_cloud_config" commands)
 // - directly via gradlew (used by "run-local-migrations" command)
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1135,7 +1135,7 @@ def fetch_workspace_details(cmd_name, *args)
       "--projectId=[projectId]",
       String,
       ->(opts, v) { opts.projectId = v},
-      "The Workspace to fetch details for (e.g. 'aou-rw-231823128'")
+      "Fetches details for workspace(s) that match the given project ID / namespace (e.g. 'aou-rw-231823128'")
 
   # Create a cloud context and apply the DB connection variables to the environment.
   # These will be read by Gradle and passed as Spring Boot properties to the command-line.
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1124,6 +1124,46 @@ Common.register_command({
   :fn => ->(*args) {fetch_firecloud_user_profile("fetch-firecloud-user-profile", *args)}
 })
 
+def fetch_workspace_details(cmd_name, *args)
+  common = Common.new
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--projectId=[projectId]",
+      String,
+      ->(opts, v) { opts.projectId = v},
+      "The Workspace to fetch details for (e.g. 'aou-rw-231823128'")
+
+  # Create a cloud context and apply the DB connection variables to the environment.
+  # These will be read by Gradle and passed as Spring Boot properties to the command-line.
+  gcc = GcloudContextV2.new(op)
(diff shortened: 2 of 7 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/FetchWorkspaceDetails.java
```diff
@@ -0,0 +1,88 @@
...
+import org.pmiops.workbench.firecloud.model.WorkspaceAccessEntry;
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+/**
+ * A tool that takes an AoU username (e.g. gjordan) and fetches the FireCloud profile associated
```
jmthibault79: bad copy/paste

api/build.gradle
```diff
@@ -627,6 +627,16 @@ task fetchFireCloudUserProfile(type: JavaExec) {
   }
 }
 
+// See project.rb command: fetch-workspace-details
+task fetchWorkspaceDetails(type: JavaExec) {
+  classpath = sourceSets.tools.runtimeClasspath
+  main = "org.pmiops.workbench.tools.FetchWorkspaceDetails"
+  systemProperties = commandLineSpringProperties
+  if (project.hasProperty("appArgs")) {
```
jaycarlton: I'm surprised you have to do this for every task that takes arguments. Also, why can't the `Eval.me()` step handle a missing property so you don't need a conditional here?

api/libproject/devstart.rb
```diff
@@ -1124,6 +1124,46 @@ def fetch_firecloud_user_profile(cmd_name, *args)
...
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--projectId=[projectId]",
+      String,
+      ->(opts, v) { opts.projectId = v},
+      "The Workspace to fetch details for (e.g. 'aou-rw-231823128'")
```
jaycarlton: I'd avoid putting real identifiers into this public repo if we can. Also, it'd be good to be explicit about whether this is a FC workspace name.

ericsong: its fake! I made it up but I wanted it to be similar enough to what a real workspace looks like so the example has some value.

jmthibault79: Project ID, not Workspace

api/tools/src/main/java/org/pmiops/workbench/tools/ServiceAccountAPIClientFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    return apiClient;
+  }
+
+  public WorkspacesApi workspacesApi() throws IOException {
+    WorkspacesApi api = new WorkspacesApi();
+    api.setApiClient(newApiClient(apiUrl));
+    return api;
+  }
+
+  public BillingApi newBillingApi() throws IOException {
```
gjuggler: Is there a reason for the naming inconsistency betw. this and the workspacesApi() method?

ericsong: nope! fixed

api/libproject/devstart.rb
```diff
@@ -1124,6 +1124,46 @@ def fetch_firecloud_user_profile(cmd_name, *args)
   :fn => ->(*args) {fetch_firecloud_user_profile("fetch-firecloud-user-profile", *args)}
 })
 
+def fetch_workspace_details(cmd_name, *args)
+  common = Common.new
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.project = TEST_PROJECT
```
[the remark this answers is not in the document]
ericsong: TBH, I copied over most of the code from other commands but I looked into it a bit and I believe this line sets the default project to cover the case where one isn't given.

I tried running the command without providing a project and it ran with the test environment.

I then removed the line and tried running the command but got an error saying I needed to specify a project.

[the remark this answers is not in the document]
jaycarlton: @gjuggler I'm going to start a wishlist doc for dev tooling inconsistencies, gaps, and gotchas. Please add this to that.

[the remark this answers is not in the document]
jaycarlton: Created http://broad.io/aou-devops-tools-input for this purpose.

PR #2855, opened 2019-11-15
[RW-3771][risk=no] Move help tips sidebar into workspace wrapper

dolbeew:
- Removes `HelpSidebar` from individual pages and adds it to `WorkspaceWrapperComponent` so it shows on all pages within a workspace (except `InteractiveNotebookComponent`).
- Adds a `helpContent` property to the data for each route with the help sidebar that determines the help tip content.
- Moves handling of participants/annotations for `SidebarContent` component into `review-state.service.ts` instead of using `props`.

@freemabd to deploy and test locally, especially annotations
@NehaBroad for general review, especially changes to workbench code

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -95,7 +95,8 @@ const routes: Routes = [
                 component: WorkspaceAboutComponent,
                 data: {
                   title: 'View Workspace Details',
-                  breadcrumb: BreadcrumbType.Workspace
+                  breadcrumb: BreadcrumbType.Workspace,
+                  helpContent: 'data'
                 }
               }, {
                 path: 'edit',
@@ -103,7 +104,8 @@ const routes: Routes = [
                 data: {
                   title: 'Edit Workspace',
                   mode: WorkspaceEditMode.Edit,
-                  breadcrumb: BreadcrumbType.WorkspaceEdit
+                  breadcrumb: BreadcrumbType.WorkspaceEdit,
+                  helpContent: 'data'
                 }
               }, {
                 path: 'duplicate',
@@ -111,7 +113,8 @@ const routes: Routes = [
                 data: {
                   title: 'Duplicate Workspace',
                   mode: WorkspaceEditMode.Duplicate,
-                  breadcrumb: BreadcrumbType.WorkspaceDuplicate
+                  breadcrumb: BreadcrumbType.WorkspaceDuplicate,
+                  helpContent: 'data'
                 }
               },
               {
@@ -122,7 +125,8 @@ const routes: Routes = [
                     component: NotebookListComponent,
                     data: {
                       title: 'View Notebooks',
-                      breadcrumb: BreadcrumbType.Workspace
+                      breadcrumb: BreadcrumbType.Workspace,
+                      helpContent: 'data'
                     }
                   }, {
                     path: ':nbName',
@@ -131,7 +135,8 @@ const routes: Routes = [
                       // use the (urldecoded) captured value nbName
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
-                      minimizeChrome: true
+                      minimizeChrome: true,
+                      helpContent: 'data'
                     }
                   }, {
                     path: 'preview/:nbName',
@@ -139,7 +144,8 @@ const routes: Routes = [
                     data: {
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
-                      minimizeChrome: true
+                      minimizeChrome: true,
+                      helpContent: 'data'
                     }
                   }
                 ]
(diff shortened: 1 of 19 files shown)


ui/src/app/app-routing.module.ts
```diff
@@ -131,15 +135,16 @@ const routes: Routes = [
...
+                      helpContent: 'data'
                     }
                   }, {
                     path: 'preview/:nbName',
                     component: InteractiveNotebookComponent,
                     data: {
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
-                      minimizeChrome: true
+                      minimizeChrome: true,
```
blrubenstein: I think this is an unrelated change? Should revert unless we're planning to have the helpContent here? (Which may make sense. I don't know what the mocks look like.)

ui/src/app/components/help-sidebar.spec.tsx
```diff
@@ -1,11 +1,26 @@
...
+  beforeEach(() => {
+    registerApiClient(CohortReviewApi, new CohortReviewServiceStub());
+    registerApiClient(CohortAnnotationDefinitionApi, new CohortAnnotationDefinitionServiceStub());
+    currentWorkspaceStore.next(workspaceDataStub);
+    cohortReviewStore.next(cohortReviewStubs[0]);
+  });
+
   it('should render', () => {
-    const wrapper = mount(<HelpSidebar location='data' />);
+    const wrapper = mount(<HelpSidebar helpContent='data' hideSidebar={0} />);
```
blrubenstein: Can we add another test that checks whether the content changes from page to page?

Also, we should test that the sidebar is hidden if hide is set to true.

ui/src/app/components/help-sidebar.tsx
```diff
@@ -124,15 +125,15 @@ const iconStyles = {
 };
 
 interface Props {
-  location: string;
+  helpContent: string;
+  hideSidebar: number;
```
blrubenstein: This feels like it should semantically be a number. It looks like it gets incremented on navigate? Could we instead set it to true on every navigate? That feels more like what we are looking to do here.

ui/src/app/components/help-sidebar.tsx
```diff
@@ -124,15 +125,15 @@ const iconStyles = {
...
   profileState: any;
-  participant?: any;
-  setParticipant?: Function;
 }
 
 interface State {
   activeIcon: string;
   filteredContent: any;
+  participant: any;
```
blrubenstein: Nit: We should type this if possible. Maybe as `ParticipantCohortStatus` if I am understanding correctly? (And the filtered content)
Do we know what data it will have?

ui/src/app/components/help-sidebar.tsx
```diff
@@ -156,6 +158,19 @@ export const HelpSidebar = withUserProfile()(
...
     });
 
+    componentDidMount(): void {
+      participantStore.subscribe(participant => this.setState({participant}));
+    }
+
+    componentDidUpdate(prevProps: Readonly<Props>): void {
+      const {helpContent, hideSidebar} = this.props;
+      // close the sidebar on each navigation excluding navigating between participants in cohort review
+      if (hideSidebar > prevProps.hideSidebar &&
```
blrubenstein: If this is a boolean, we could probably do this as: `hideSidebar && !prevProps.hideSidebar`, if I am understanding this correctly?

ui/src/app/components/help-sidebar.tsx
```diff
@@ -156,6 +158,19 @@ export const HelpSidebar = withUserProfile()(
...
+        !(helpContent === 'reviewParticipantDetail' && prevProps.helpContent === 'reviewParticipantDetail')) {
+        this.setState({activeIcon: undefined, sidebarOpen: false});
```
blrubenstein: If we have `hideSidebar` as a boolean, could we get rid of `sidebarOpen`? It seems redundant? I am a little confused by the interaction of `sidebarOpen` and `hideSidebar` here. It seems to me like `sidebarOpen` is the source of truth, but parent components also need to manage state? Which suggests maybe we should lift the state up and use `sidebarOpen` to trigger a callback?

ui/src/app/pages/data/cohort-review/sidebar-content.component.tsx
```diff
@@ -294,6 +294,7 @@ export const SidebarContent = fp.flow(
       .then(({items}) => {
         this.setState({annotationDefinitions: items});
       });
+    participantStore.subscribe(participant => this.setState({participant: participant || {} as ParticipantCohortStatus}));
```
blrubenstein: We should make sure to unsubscribe at `componentWillUnmount` (Per @calbach's unsubscribe note)

ui/src/app/components/help-sidebar.tsx
dolbeew: Yes, I've changed this after reworking state/props based on your comment below

ui/src/app/components/help-sidebar.tsx
dolbeew: Renamed and changed to boolean along with reworking state/props based on your comment below

ui/src/app/pages/data/cohort-review/sidebar-content.component.tsx
dolbeew: Addressed this on the last commit, forgot to comment.

PR #2857, opened 2019-11-17
Misc. refactorings & more logging for BPBS

jaycarlton:
Added some happy path logging and a few small refactorings:

- doing one query for the top 5 instead of 5 queries for top 1
- replacing input of `dao.save()` with its output
- added a transient method to the entry class to get the interval. I'm tempted to put more of the calculations that depend on the entity's state exclusively (and which are used only to update the entities) into the entity class itself. This is apparently not common in our codebase, but it seems more natural to stuff these kinds of calculations into the value class in question.
- Added a helper class Comparables for prettier comparisons between Comparable objects


api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -31,11 +36,13 @@
 
   private static final Logger log = Logger.getLogger(BillingProjectBufferService.class.getName());
 
-  private static final int SYNCS_PER_INVOCATION = 5;
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
   @VisibleForTesting static final Duration ASSIGNING_TIMEOUT = Duration.ofMinutes(10);
-
+  private static ImmutableMap<BufferEntryStatus, Duration> statusToGracePeriod =
```
blrubenstein: Is the goal of this to clean things up that we think may be failed/errored in FireCloud?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -115,36 +130,38 @@ public void syncBillingProjectStatus() {
...
+                    entry.getFireCloudProjectName(), fireCloudStatus.toString()));
             break;
         }
       } catch (NotFoundException e) {
         log.log(
             Level.WARNING,
-            "Get BillingProjectStatus call failed for "
-                + entry.getFireCloudProjectName()
-                + ". Project not found.",
+            "Get BillingProjectStatus call failed for " + entry.toString() + ". Project not found.",
```
blrubenstein: Do we want the full string here? It seems to me that just the project name would be enough? (same for below)

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -173,22 +203,25 @@ public void cleanBillingBuffer() {
...
   }
 
-  public DbBillingProjectBufferEntry assignBillingProject(DbUser user) {
+  private Duration getStatusChangedToLastSync(DbBillingProjectBufferEntry entry) {
+    return Duration.between(
+        entry.getLastStatusChangedTime().toInstant(), entry.getLastSyncRequestTime().toInstant());
+  }
+
+  public DbBillingProjectBufferEntry assignBillingProject(DbUser dbUser) {
     DbBillingProjectBufferEntry entry = consumeBufferEntryForAssignment();
```
blrubenstein: Opt Nit: Not actually part of this PR, but all these objects named `entry` should probably be fully qualified imo. Just to make it clear what types we're working with?

api/src/test/java/org/pmiops/workbench/actionaudit/ActionAuditServiceTest.kt
```diff
@@ -49,7 +49,7 @@ class ActionAuditServiceTest {
 
     @Test
     fun testSendsSingleEvent() {
-        actionAuditService!!.send(EVENT_1!!)
+        actionAuditService!!.send(EVENT_1)
```
blrubenstein: Why this change, and what does it do?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: It's cheap enough to include the other fields, and there's a nonzero chance they could be helpful. For things that are called by crons, I like to be generous with logging context.

api/src/test/java/org/pmiops/workbench/actionaudit/ActionAuditServiceTest.kt
jaycarlton: It is, and it's also on other branches.

It fixes a Kotlin warning about unnecessary promotion of a nullable type to a non-nullable type. Initially the argument was `event1: ActionAuditEvent?`, in order to allow instantiating it in the `@Before` method. When I changed it to a constant without the question mark, that made the compiler enforce a non-null constraint on it everywhere, which makes the `!!` in this call redundant.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: I keep hearing anecdotally things about leaked projects and the buffer getting in a bad state. I'm still not used to dealing with so many cron-based utilities, but I think the best thing to do in this situation is to over-log. We're definitely not dealing with such huge volumes of entries that space is a concern. Likewise, it's not going to matter in terms of performance since most of our time is spent waiting on Firecloud and/or GCP.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: I'm pretty verbose, but if we name them all `dbBillingProjectBufferEntry`, it's a mouthful. I'll change them to `bufferEntry`. Should be clear enough, since this is the only buffer we really have in play here.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -179,21 +214,26 @@ private void setBufferEntryErrorStatus(DbBillingProjectBufferEntry entry) {
         .filter(entry -> entry.getLastSyncRequestTime() != null)
```
blrubenstein: Also a bunch of `entry` down here.

PR #2859, opened 2019-11-18
[risk=no][RW-3740] FIx recent workspace cards overflow 

ericsong:
old:
1878397f-f0ae-420a-87ea-1d46a2ea9d83

new:
rw_scroll


ericsong:
@jmthibault79 got approval from Lou and Karthik

PR #2861, opened 2019-11-18
[RW-3988][Risk=no] Srubenst/preview data time formatting

blrubenstein:
During the workshop, we realized all the dates in the preview table in dataset builder were very close to the Linux Epoch. When I poked around locally, the values from the API seemed to be similarly incorrect. When looking at the values side by side with something that would be pointing to a date similar to what we might expect, it looked like they might be off by a factor of 1000. Ariel and I determined BQ was likely returning data in seconds rather than milliseconds.

Tested manually and it looks fine

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -15,6 +15,8 @@ import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.time.Clock;
+import java.time.Instant;
+import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -422,7 +422,8 @@ private void formatTimestampValues(List<DataSetPreviewValueList> valuePreviewLis
               value -> {
                 if (!value.equals(EMPTY_CELL_MARKER)) {
                   Double fieldValue = Double.parseDouble(value);
-                  queryValues.add(dateFormat.format(new Date(fieldValue.longValue())));
+                  // Bigquery returns time values in seconds, rather than in milliseconds.
+                  queryValues.add(dateFormat.format(new Date(fieldValue.longValue() * 1000)));
```
jaycarlton: You shouldn't have to do conversions inline here. There are things like [ofEpochSecond](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#ofEpochSecond-long-int-java.time.ZoneOffset-)

jaycarlton: or use [Instant](https://www.tutorialspoint.com/javatime/javatime_instant_ofepochsecond.htm)

blrubenstein:
This is the only place where that enum is checked against.

jaycarlton:
Sounds good. Eventually we should move this closer to a BigQueryUtils or something. Maybe have a method that takes in a field and a DateFormatter to make a string.

PR #2864, opened 2019-11-19
[RW-4007][RISK=NO] Login Auditing

jaycarlton:
The closest thing we have right now to a "login event" is the `getMe()` method on the `v1/profile` endpoint. This works to catch logins, but also captures profile display actions.

As of right now, we think this is better than nothing. The extra events are during a session anyway, and shouldn't lead to wildly inaccurate conclusions.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/adapters/ProfileAuditAdapterImpl.kt
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/adapters/ProfileAuditAdapterImpl.kt
@@ -104,6 +104,26 @@ constructor(
         }
     }
 
+    override fun fireLoginAction(dbUser: DbUser) {
+        try {
+            actionAuditService.send(ActionAuditEvent(
+                    timestamp = clock.millis(),
+                    actionId = actionIdProvider.get(),
+                    actionType = ActionType.LOGIN,
+                    agentType = AgentType.USER,
+                    agentId = dbUser.userId,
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/actionaudit/adapters/ProfileAuditAdapterImpl.kt
```diff
@@ -104,6 +104,26 @@ constructor(
...
+                    actionId = actionIdProvider.get(),
+                    actionType = ActionType.LOGIN,
+                    agentType = AgentType.USER,
+                    agentId = dbUser.userId,
+                    agentEmailMaybe = dbUser.email,
+                    targetType = TargetType.WORKBENCH,
+                    targetIdMaybe = null,
+                    targetPropertyMaybe = null,
+                    previousValueMaybe = null,
+                    newValueMaybe = null
```
NehaBroad: just curious, whats newValueMaybe used for? can you give an example?

NehaBroad: nevermind, i just saw the example for fireCreateAction

jaycarlton: I'm going to set up a presentation on the system for after Thanksgiving.

PR #2867, opened 2019-11-20
[RW-3870][Risk=no] Add DUA tests

blrubenstein:
I decided there were two major tests missing. One to verify the saving of DUA information, and the other test to verify initials all need to be correct on the front end to save.

### DIFF
--- a/ui/src/app/pages/profile/data-use-agreement.spec.tsx
+++ b/ui/src/app/pages/profile/data-use-agreement.spec.tsx
@@ -18,6 +18,25 @@ describe('DataUseAgreement', () => {
     expect(wrapper).toBeTruthy();
   });
 
+  it('should not allow DataUseAgreement without identical initials', () => {
+    const wrapper = component();
+    expect(wrapper.find('[data-test-id="submit-dua-button"]').prop('disabled')).toBeTruthy();
+
+    // fill required fields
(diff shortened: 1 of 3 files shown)


ui/src/app/pages/profile/data-use-agreement.spec.tsx
```diff
@@ -18,6 +18,25 @@ describe('DataUseAgreement', () => {
...
 
+  it('should not allow DataUseAgreement without identical initials', () => {
+    const wrapper = component();
+    expect(wrapper.find('[data-test-id="submit-dua-button"]').prop('disabled')).toBeTruthy();
+
+    // fill required fields
+    wrapper.find('[data-test-id="dua-name-input"]')
+      .simulate('change', {target: {value: 'Fake Name'}});
+    // add initials to just one initials input field.
+    wrapper.find('[data-test-id="dua-initials-input"]').first().simulate('change', {target: {value: 'XX'}});
```
NehaBroad: If i understand this correctly this is testing that the submit button should be disabled if only one initials is added? If yes this feels like a different test case. For test case 'without identical initials' populating each input within forEach is apt

NehaBroad:
It would be better to add test to check if the name and username inputs are populated too

PR #2868, opened 2019-11-20
[risk=no][no ticket] Add ListClusterResponse to notebooks.yaml

jmthibault79:
To match changes coming in https://github.com/DataBiosphere/leonardo/pull/1124

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
@@ -84,12 +84,15 @@ public class OfflineClusterController implements OfflineClusterApiDelegate {
     int activeDeletes = 0;
     int unusedDeletes = 0;
     for (ListClusterResponse listClusterResponse : clusters) {
-      String clusterId = listClusterResponse.getGoogleProject() + "/" + listClusterResponse.getClusterName();
+      String clusterId =
+          listClusterResponse.getGoogleProject() + "/" + listClusterResponse.getClusterName();
       Cluster c;
       try {
         // Refetch the cluster to ensure freshness as this iteration may take
         // some time.
-        c = clusterApi.getCluster(listClusterResponse.getGoogleProject(), listClusterResponse.getClusterName());
+        c =
+            clusterApi.getCluster(
+                listClusterResponse.getGoogleProject(), listClusterResponse.getClusterName());
       } catch (ApiException e) {
         log.log(Level.WARNING, String.format("failed to refetch cluster '%s'", clusterId), e);
         errors++;
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
@@ -16,6 +16,7 @@ import org.pmiops.workbench.notebooks.NotebooksConfig;
 import org.pmiops.workbench.notebooks.api.ClusterApi;
 import org.pmiops.workbench.notebooks.model.Cluster;
 import org.pmiops.workbench.notebooks.model.ClusterStatus;
+import org.pmiops.workbench.notebooks.model.ListClusterResponse;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.http.ResponseEntity;
@@ -71,7 +72,7 @@ public class OfflineClusterController implements OfflineClusterApiDelegate {
     Duration idleMaxAge = Duration.ofDays(config.firecloud.clusterIdleMaxAgeDays);
 
     ClusterApi clusterApi = clusterApiProvider.get();
-    List<Cluster> clusters;
+    List<ListClusterResponse> clusters;
     try {
       clusters = clusterApi.listClusters(null, false);
     } catch (ApiException e) {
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClient.java
```diff
@@ -25,9 +25,6 @@ Cluster createCluster(String googleProject, String clusterName, String workspace
   /** Deletes a notebook cluster */
   void deleteCluster(String googleProject, String clusterName) throws WorkbenchException;
 
-  /** Lists all existing clusters */
-  List<Cluster> listClusters(String labels, boolean includeDeleted) throws WorkbenchException;
```
jmthibault79: this was removed because it was not used, and it's easy to confuse this with ClusterController.listClusters()

api/src/main/resources/notebooks.yaml
```diff
@@ -1183,7 +1183,6 @@ definitions:
   Cluster:
     description: ''
     required:
```
calbach: Would instead update this to reference the updated YAML per Qi's PR, these fields still exist on GetCluster

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -145,12 +145,6 @@ public void deleteCluster(String googleProject, String clusterName) {
         });
   }
 
-  @Override
-  public List<Cluster> listClusters(String labels, boolean includeDeleted) {
```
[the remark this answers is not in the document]
calbach: My mistake, this wrapper library is not actually needed there - but the underlying method gets used there

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -138,14 +138,7 @@
     if (org.pmiops.workbench.notebooks.model.ClusterStatus.ERROR.equals(fcCluster.getStatus())) {
       if (retries <= 2) {
         this.userService.setClusterRetryCount(retries + 1);
-        log.warning("Cluster has errored with logs: ");
```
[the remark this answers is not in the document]
jmthibault79: Oops - you're right.  They're splitting the data model into `Cluster` and `ListClusterResponse` so we should do the same.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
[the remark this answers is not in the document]
jmthibault79: Thanks for the ManageClusters catch

api/src/main/java/org/pmiops/workbench/api/OfflineClusterController.java
```diff
@@ -82,12 +83,16 @@
...
+    for (ListClusterResponse listClusterResponse : clusters) {
+      String clusterId =
+          listClusterResponse.getGoogleProject() + "/" + listClusterResponse.getClusterName();
+      Cluster c;
       try {
         // Refetch the cluster to ensure freshness as this iteration may take
         // some time.
-        c = clusterApi.getCluster(c.getGoogleProject(), c.getClusterName());
+        c =
+            clusterApi.getCluster(
```
jmthibault79: lint is insisting on this extra line, oh well

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -20,8 +20,8 @@
 import org.pmiops.workbench.notebooks.ApiClient;
 import org.pmiops.workbench.notebooks.ApiException;
 import org.pmiops.workbench.notebooks.api.ClusterApi;
-import org.pmiops.workbench.notebooks.model.Cluster;
 import org.pmiops.workbench.notebooks.model.ClusterStatus;
+import org.pmiops.workbench.notebooks.model.ListClusterResponse;
```
jmthibault79: @ericsong this is what fixed the project.rb tool today

api/src/test/java/org/pmiops/workbench/api/OfflineClusterControllerTest.java
```diff
@@ -81,47 +82,59 @@ public void setUp() {
...
+  private Cluster toCluster(ListClusterResponse listClusterResponse) {
+    return new Cluster()
+        .clusterName(listClusterResponse.getClusterName())
+        .googleProject(listClusterResponse.getGoogleProject())
+        .status(listClusterResponse.getStatus())
+        .createdDate(listClusterResponse.getCreatedDate())
+        .dateAccessed(listClusterResponse.getDateAccessed());
+  }
+
+  private void stubClusterListResponse(List<ListClusterResponse> clusters) throws Exception {
```
ericsong: nit: I don't think this has to change but I found it a little more intuitive to keep passing in `Cluster` objects and then used a `Cluster -> ClusterListResponse` method to modify line 111.

api/src/main/resources/notebooks.yaml
```diff
@@ -1258,6 +1258,72 @@ definitions:
...
+      - status
+      - createdDate
+      - labels
+      - dateAccessed
+      - autopauseThreshold
+      - defaultClientId
+    properties:
+      internalId:
+        type: string
+        description: Internal resource ID of the cluster
```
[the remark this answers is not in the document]
jmthibault79: This is a direct copy/paste from Leonardo.  I'd rather keep in their style, to make it easier to stay in sync

PR #2869, opened 2019-11-20
[risk=no] renaming db model objects.

freemabd:
Removed any fully qualified name references in codebase as well as the Db prefix to the following model classes:

**/api**
- org.pmiops.workbench.cdr.model.CBCriteriaAttribute

**/common-api**
- org.pmiops.workbench.cdr.model.Concept
- org.pmiops.workbench.cdr.model.ConceptRelationship
- org.pmiops.workbench.cdr.model.ConceptRelationshipId
- org.pmiops.workbench.cdr.model.ConceptSynonym
- org.pmiops.workbench.cdr.model.CBCriteria
- org.pmiops.workbench.cdr.model.DomainInfo
- org.pmiops.workbench.cdr.model.MenuOption
- org.pmiops.workbench.cdr.model.SurveyModule
- org.pmiops.workbench.db.model.CdrVersion
- org.pmiops.workbench.db.model.Config


api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
```diff
@@ -23,12 +23,12 @@
 import org.pmiops.workbench.cdr.CdrVersionService;
 import org.pmiops.workbench.cdr.dao.CBCriteriaAttributeDao;
 import org.pmiops.workbench.cdr.dao.CBCriteriaDao;
-import org.pmiops.workbench.cdr.model.CBCriteria;
+import org.pmiops.workbench.cdr.model.DbCriteria;
```
jaycarlton: Why isn't this `DbCBCriteria`? Did you decide it's not cohort-builder-specific, or is that just too much of a mouthful?

api/src/main/java/org/pmiops/workbench/api/CohortBuilderController.java
```diff
@@ -25,17 +25,19 @@
 import org.pmiops.workbench.cdr.CdrVersionService;
 import org.pmiops.workbench.cdr.dao.CBCriteriaAttributeDao;
 import org.pmiops.workbench.cdr.dao.CBCriteriaDao;
-import org.pmiops.workbench.cdr.model.CBCriteria;
-import org.pmiops.workbench.cdr.model.CBCriteriaAttribute;
-import org.pmiops.workbench.cdr.model.MenuOption;
+import org.pmiops.workbench.cdr.model.DbCriteria;
```
jaycarlton: Oh, so these are in `cdr.model` and not `db.model`. That explains why I missed them.Do you expect other DB models to be stored with their most closely related packages?

common-api/src/main/java/org/pmiops/workbench/cdr/dao/CBCriteriaDao.java
```diff
@@ -83,16 +83,16 @@
   /** This query returns all parents matching the parentConceptIds. */
   @Query(
       value =
-          "select c from CBCriteria c where conceptId in (:parentConceptIds) and domainId = :domain and type = :type and standard = :standard and "
+          "select c from DbCriteria c where conceptId in (:parentConceptIds) and domainId = :domain and type = :type and standard = :standard and "
```
jaycarlton: Double check that this is what's wanted.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
```diff
@@ -42,16 +42,16 @@
    */
   @Query(
       value =
-          "select new org.pmiops.workbench.cdr.model.DomainInfo(\n"
+          "select new org.pmiops.workbench.cdr.model.DbDomainInfo(\n"
               + "d.domain, d.domainId, d.name, d.description,\n"
               + "d.conceptId, COUNT(*), 0L, d.participantCount)\n"
-              + "from DomainInfo d\n"
-              + "join Concept c ON d.domainId = c.domainId\n"
+              + "from DbDomainInfo d\n"
```
jaycarlton: I remember when I went through this I had to manually fix up one spot where there was a query. But that may only be true because it happened to be a legacy SQL query.

common-api/src/main/java/org/pmiops/workbench/cdr/model/DbSurveyModule.java
```diff
@@ -10,9 +10,9 @@
 
 @Entity
 @Table(name = "survey_module")
-public class SurveyModule {
+public class DbSurveyModule {
 
-  public static final Function<SurveyModule, org.pmiops.workbench.model.SurveyModule>
+  public static final Function<DbSurveyModule, org.pmiops.workbench.model.SurveyModule>
```
jaycarlton: Can we kill this qualification?

api/src/main/java/org/pmiops/workbench/api/CohortBuilderController.java
freemabd: Yes. Some background on the split between db.model and cdr.model. Anything that lives in db.model is in the cloud mysql workbench db schema(1 schema for all cdrs) and anything in cdr.model is in the cloud mysql cdr(multiple cdr versions) schemas. We keep this separation  because when querying any model with the cdr.model package, we have to switch cdr db schemas on the fly per client request.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/CBCriteriaDao.java
freemabd: Yes, intended.
You probably already know this but will toss this out there just in case. 
This `@Query` annotation is using JPQL, so the it uses the object name instead of the table name. 
JPQL Example:
`@Query("select obj from DbMyObject obj where obj.status = 1")`
Native Example:
`@Query("select * from my_table where status = 1", nativeQuery = true)`

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
freemabd: I had to manually change DbConcept below. For some reason Intellij did not refactor Concept to DbConcept in this specific query string when using it's refactor functionality(Shift+F6).

common-api/src/main/java/org/pmiops/workbench/cdr/model/DbSurveyModule.java
freemabd: Done.

api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
jaycarlton: You should go to Hollywood and shop the rights to this story for a feature film. I think Matt Damon should play CBCriteria.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/CBCriteriaDao.java
jaycarlton: Cool. I wish they'd made the native one a separate annotation instead of an option on this one, but it works.

common-api/src/main/java/org/pmiops/workbench/cdr/dao/DomainInfoDao.java
jaycarlton: I think it treats it like a regular string literal. I.e. it doesn't take into consideration that it's supposed to be a query, or that the class name maps to a table name. You have to be pretty careful with the checkboxes when renaming, since there are some imperfect heuristics or logic going on.

PR #2870, opened 2019-11-20
[RW-3602][RISK=NO] Replace uses of AssertJ with Truth

jaycarlton:
The decision was to go with Truth, which means moving out the assertions from AssertJ.

The only place where there might be pain is around the use of `isWithinPercentage()` which has no direct analog. I tried to find a reasonable workaround there.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
+++ b/api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
@@ -56,23 +56,17 @@ public class FreeTierBillingServiceTest {
 
   private static final double HALF_TOLERANCE_PERCENTAGE = 0.001;
   private static final double DEFAULT_FREE_CREDITS_LIMIT = 1.0;
-  @Autowired
-  BigQueryService bigQueryService;
+  @Autowired BigQueryService bigQueryService;
 
-  @Autowired
-  FreeTierBillingService freeTierBillingService;
+  @Autowired FreeTierBillingService freeTierBillingService;
 
-  @Autowired
-  UserDao userDao;
+  @Autowired UserDao userDao;
 
-  @Autowired
-  NotificationService notificationService;
+  @Autowired NotificationService notificationService;
 
-  @Autowired
-  WorkspaceDao workspaceDao;
+  @Autowired WorkspaceDao workspaceDao;
 
-  @Autowired
-  WorkspaceFreeTierUsageDao workspaceFreeTierUsageDao;
+  @Autowired WorkspaceFreeTierUsageDao workspaceFreeTierUsageDao;
 
   private static WorkbenchConfig workbenchConfig;
 
@@ -269,9 +263,7 @@ public class FreeTierBillingServiceTest {
       DbWorkspaceFreeTierUsage usage = workspaceFreeTierUsageDao.findOneByWorkspace(ws);
       assertThat(usage.getUser()).isEqualTo(ws.getCreator());
       assertThat(usage.getWorkspace()).isEqualTo(ws);
-      assertWithinPercentage(usage.getCost(),
-          expectedCosts.get(ws.getWorkspaceId()),
-          0.001);
+      assertWithinPercentage(usage.getCost(), expectedCosts.get(ws.getWorkspaceId()), 0.001);
     }
   }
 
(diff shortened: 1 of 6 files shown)


jmthibault79:
I wrote those `isWithinPercentage()` tests.  `assertCloseEnough()` fits the semantics of what I was trying to do, so you can convert all of them.  👍

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -261,8 +263,7 @@ public void checkFreeTierBillingUsage_twoUsers() {
       DbWorkspaceFreeTierUsage usage = workspaceFreeTierUsageDao.findOneByWorkspace(ws);
       assertThat(usage.getUser()).isEqualTo(ws.getCreator());
       assertThat(usage.getWorkspace()).isEqualTo(ws);
-      assertThat(usage.getCost())
-          .isCloseTo(expectedCosts.get(ws.getWorkspaceId()), withinPercentage(0.001));
+      assertWithinPercentage(usage.getCost(), expectedCosts.get(ws.getWorkspaceId()), 0.001);
```
blrubenstein: Could this use the `assertCloseEnough` method?

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -229,8 +231,8 @@ public void checkFreeTierBillingUsage_combinedProjectsExceedsLimit() {
       DbWorkspaceFreeTierUsage usage = workspaceFreeTierUsageDao.findOneByWorkspace(ws);
       assertThat(usage.getUser()).isEqualTo(user);
       assertThat(usage.getWorkspace()).isEqualTo(ws);
-      assertThat(usage.getCost())
-          .isCloseTo(expectedCosts.get(ws.getWorkspaceId()), withinPercentage(0.001));
+
+      assertCloseEnough(usage.getCost(), expectedCosts.get(ws.getWorkspaceId()));
```
blrubenstein: Does this automatically pickup the percentage value? Otherwise this is a change of behavior.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -383,13 +385,12 @@ public void getUserCachedFreeTierUsage() {
...
-    assertThat(freeTierBillingService.getUserCachedFreeTierUsage(user))
-        .isCloseTo(SINGLE_WORKSPACE_TEST_COST, withinPercentage(0.001));
+    assertCloseEnough(
+        freeTierBillingService.getUserCachedFreeTierUsage(user), SINGLE_WORKSPACE_TEST_COST);
 
     freeTierBillingService.checkFreeTierBillingUsage();
 
-    assertThat(freeTierBillingService.getUserCachedFreeTierUsage(user))
-        .isCloseTo(1100.0, withinPercentage(0.001));
+    assertCloseEnough(freeTierBillingService.getUserCachedFreeTierUsage(user), 1100.0);
```
blrubenstein: Should this be pulled out into a variable, like the ones above?

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
blrubenstein: Edit: I learned later in the PR that this is not a provided method, but a helper we have written. We should make that explicit somehow, or just use the `assertWithinPercentage` method. (I don't know if the abstraction gives us much, cause we've already abstracted the variable so changes only need to happen in one place regardless)

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -455,4 +454,14 @@ private DbWorkspace createWorkspace(
     workspace.setBillingMigrationStatusEnum(billingMigrationStatus);
     return workspaceDao.save(workspace);
   }
+
+  private void assertCloseEnough(double actualValue, double expectedValue) {
```
blrubenstein: I thought that these were provided methods at first, because they had the same `assert` prefix. I think it would be good to make it clear this is code that we own.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
jaycarlton: `assertCloseEnough()` is written in terms of `assertWithinPercentage`.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
jaycarlton: The default value is a class constant, so we should be free to raise the abstraction level from "within CLASS_CONSTANT" to "close enough for the purposes of this class". It's very cheap to make private methods, and pretty much any time you can shrink an argument list and maintain an appropriate abstraction, that's always a win in my book.

There's also still the possibility of someone passing in the wrong value in the 3-arg form.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
jaycarlton: I hear what you're saying... there are certainly a huge number of functions starting with assert.

In the present case, though, it's a private method, so there's no risk of using it outside this class. Anyone interested can go to its definition. But I renamed it `assertWithinBillingTolerance()` for now so it's obvious at a glance.

PR #2876, opened 2019-11-22
[risk=low] CommandLineToolConfig restructuring for CLI bean isolation

calbach:
- `@SpringBootApplication` is an alias for 3 annotations, one of which was problematic:
  - `@Configuration` (effectively: this class can declare beans)
  - `@EnableAutoconfiguration` (import basic Spring modules, e.g. JPA, REST stuff, Liquibase)
  -  :warning: `@ComponentScan` (start grabbing all beans indiscriminately from the current source tree rooted at this package)
- ComponentScan had the side effect of muddling up all tool bean configs (including everything in every CLI tool) in the source directory into the same context, this was unexpected and undesirable.
- For a given CLI tool, in addition to that tool's `@Configuration`, the only thing we actually wanted to scan was the `CommandLineToolConfig`, which now needs to be explicitly brought in
- Also switch to a parent/child relationship, which may allow for easier overrides of particular bean declarations. `@Import`ing `CommandLineToolConfig` would have also worked.
  - **Note:** Parent/child contexts exhibit what I'd call "early binding" behavior. This means once you're injecting a bean that is sourced from the parent context, all dependencies for that bean can only be produced by the parent context. For example, if I got a `WorkspaceService` from the parent context (which depends on `WorkspacesApi`), as a child I'm not allowed to patch in my own custom `WorkspacesApi` override - to do that I'd need to provide my on `WorkspaceService` as well. If this proves to be undesirable, we can simply union the contexts as was done previously.

In particular, this article was highly relevant: https://www.blackpepper.co.uk/blog/a-modular-architecture-with-spring-boot

Example code in this PR shows a failing Bean in tool B (loadDataDictionary) coexisting with tool A (setAuthority). On master, this would cause tool A to fail.

---
**PR checklist**

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
@@ -20,12 +20,13 @@ import org.pmiops.workbench.monitoring.MonitoringServiceImpl;
 import org.pmiops.workbench.monitoring.MonitoringSpringConfiguration;
 import org.pmiops.workbench.monitoring.StackdriverStatsExporterService;
 import org.pmiops.workbench.notebooks.NotebooksServiceImpl;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.domain.EntityScan;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Lazy;
-import org.springframework.context.annotation.Scope;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 import org.springframework.retry.backoff.BackOffPolicy;
 import org.springframework.retry.backoff.ExponentialRandomBackOffPolicy;
@@ -38,9 +39,13 @@ import org.springframework.retry.backoff.ThreadWaitSleeper;
  *
  * <p>The main difference is that request-scoped dependencies cannot be used from a command-line
  * context.
+ *
+ * <p>IMPORTANT: This config should only be used in a database context. In order to implement a tool
+ * which uses this config, that tool must be run with an active SQL server available, e.g. via Cloud
+ * SQL Proxy.
  */
 @Configuration
-@EnableJpaRepositories({"org.pmiops.workbench.db.dao"})
+@EnableAutoConfiguration
 @Import({
   CommonConfig.class,
   MonitoringServiceImpl.class,
@@ -50,6 +55,8 @@ import org.springframework.retry.backoff.ThreadWaitSleeper;
   StackdriverStatsExporterService.class,
   UserRecentResourceServiceImpl.class
 })
+@EnableJpaRepositories({"org.pmiops.workbench.db.dao"})
+@EntityScan("org.pmiops.workbench.db.model")
 // Scan the google module, for CloudStorageService and DirectoryService beans.
 @ComponentScan("org.pmiops.workbench.google")
 // Scan the FireCloud module, for FireCloudService bean.
@@ -76,8 +83,6 @@ public class CommandLineToolConfig {
    *
    * <p>Any command-line tool which loads this bean needs to be called from a project.rb command
    * which is preceded with "get_gsuite_admin_key" to ensure the local key file si populated.
-   *
-   * @return
    */
   @Lazy
   @Bean(name = Constants.GSUITE_ADMIN_CREDS)
@@ -112,19 +117,11 @@ public class CommandLineToolConfig {
   /**
    * Instead of using the CacheSpringConfiguration class (which has a request-scoped bean to return
    * a cached workbench config), we load the workbench config once from the database and use that.
-   *
-   * @param configDao
-   * @return
    */
   @Bean
   @Lazy
-  @Scope("prototype")
   WorkbenchConfig workbenchConfig(ConfigDao configDao) {
     DbConfig config = configDao.findOne(DbConfig.MAIN_CONFIG_ID);
-
-    if (config == null) {
-      return null;
-    }
 
     Gson gson = new Gson();
     return gson.fromJson(config.getConfiguration(), WorkbenchConfig.class);
@@ -133,8 +130,6 @@ public class CommandLineToolConfig {
   /**
    * Returns the Apache HTTP transport. Compare to CommonConfig which returns the App Engine HTTP
    * transport.
-   *
-   * @return
    */
   @Bean
   HttpTransport httpTransport() {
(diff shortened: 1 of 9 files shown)


ericsong:
Cool. I think this would be a good change. What do you need to move this from POC to ready to merge?

api/tools/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+debug=true
```
ericsong: Is there a way to hook up the intelliJ debugger? I've been making do this print statements... being able to use the debugger would be awesome.

calbach: I haven't personally done this or tried to do it, may be possible though. This line really just tells Spring to generate a lot of log spam (including dependency injection details)

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
```diff
@@ -34,6 +34,10 @@
  *
  * <p>The main difference is that request-scoped dependencies cannot be used from a command-line
  * context.
+ *
+ * <p>IMPORTANT: This config should only be used in a database context. In order to implement a tool
```
jaycarlton: Can we set it up so it fails to inject when that condition isn't met?

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
```diff
@@ -63,6 +58,9 @@ public CommandLineRunner run(UserDao userDao, FireCloudService fireCloudService)
   }
 
   public static void main(String[] args) throws Exception {
-    new SpringApplicationBuilder(FetchFireCloudUserProfile.class).web(false).run(args);
+    new SpringApplicationBuilder(CommandLineToolConfig.class)
```
jaycarlton: Since only the child class ever changes, you could make a helper method that just takes in that class object.

But I think the simplest would be `runAouCommandLineApp(getClass(), args);`. You might be able to genericize out the first argument, but type erasure might get in the way.

The biggest win would be localizing all the Spring incantations to a generic main/wrapper class.

calbach:
Putting this on hold for https://github.com/all-of-us/workbench/pull/2879

api/tools/src/main/java/org/pmiops/workbench/tools/ConfigLoader.java
```diff
@@ -103,6 +99,9 @@ public CommandLineRunner run(ConfigDao configDao) {
   }
 
   public static void main(String[] args) throws Exception {
-    new SpringApplicationBuilder(ConfigLoader.class).web(false).run(args);
+    new SpringApplicationBuilder(CommandLineToolConfig.class)
+        .child(ConfigLoader.class)
```
ericsong: what does the `child` param do?

calbach: The child config can override beans defined in the parent: https://www.baeldung.com/spring-boot-context-hierarchy

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
calbach: It's not a clear win to me. Adding a layer of indirection has a price in terms of reader comprehension. This case is borderline, but since it's still a fairly simple builder invocation with only a few LOC, I'd rather leave it for now.

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
freemabd: @jaycarlton I'm not aware of any addons that do this

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -84,20 +78,9 @@ WorkspaceService workspaceService() {
...
-            new ServiceAccountAPIClientFactory(configProvider.get().firecloud.baseUrl)
-                .workspacesApi();
-      } catch (IOException e) {
-        throw new RuntimeException(e);
-      }
-    }
-
-    return workspacesApi;
+  WorkspacesApi workspaceApi(WorkbenchConfig config) throws IOException {
+    return new ServiceAccountAPIClientFactory(config.firecloud.baseUrl).workspacesApi();
```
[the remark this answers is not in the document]
calbach: Done

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
```diff
@@ -24,10 +22,7 @@
  * <p>This is intended mostly for demonstration / testing purposes, to show how we leverage
  * domain-wide delegation to make FireCloud API calls impersonating other users.
  */
-@SpringBootApplication
-// Load the DBA and DB model classes required for UserDao.
-@EnableJpaRepositories({"org.pmiops.workbench.db.dao"})
-@EntityScan("org.pmiops.workbench.db.model")
+@Configuration
 public class FetchFireCloudUserProfile {
```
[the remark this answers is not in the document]
calbach: Every CLI here follows a convention of being a `@Configuration`. I'd prefer to give them semantically meaningful names and not bloat them with `Configuration`; in my view that's secondary and a means to and end of being a CLI in this framework.

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
calbach: I get the rationale. Like I said, it's borderline to me, so since you seem to feel strongly I'll make the change.

api/tools/src/main/java/org/pmiops/workbench/tools/ConfigLoader.java
jaycarlton: Very interesting. Thanks.

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
jaycarlton: Thanks.

I thought about it some more, and I realized I should've asked a different question first: How did we decide each CLI tool would be a separate spring application as opposed to having one app with different sub-commands that shared a Spring context.  Is Spring smart enough not to instantiate branches of the dependency tree it doesn't need. If not, that would be one argument for separating them.

api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java
jaycarlton: OK. I know emacs can do it. I'll find one.

Even better if we could tweak spotless to do it for us.

calbach:
OK, I manually verified `export-workspace-data`, `fetch-firecloud-profile`, and everything hit by `dev-up`. I will plan to merge this once tests go green.

PR #2877, opened 2019-11-22
generic setter for collection and count

jaycarlton:
I like looking for excuses to play with generics. This would've only taken 10 minutes if I knew what I was doing.

Couldn't get spotless to run because of an issue in an unrelated class. Something about command line tool config.

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -13,6 +13,8 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


ericsong:
This is pretty cool but I think the prior version without generics is easier to understand for someone reading the code.

In general, I try to limit my usage of Generics except for the most clear cut cases since its not an easy concept to understand and its a common source of bugs and confusion for people that did not write the implementation.

This case seems a bit on the edge of Generic's use cases. Another indication of this to me is the _increase_ in LOC after applying this refactoring tool.

jaycarlton:
It raises the level of abstraction, so there are fewer concepts to understand, even if there's slightly more code. You've also got to consider the possibility of one of the original call sites getting out of sync semantically with how the others work. It's confusing for readers to try to guess if it's an intentional pattern or not.

One tradeoff you could make is to pull the string formatting into the generic function so it doesn't have to be on your row class.

Another (better) one I thought of on my way in this morning is to make the count columns set by the same method on the row object that takes in the values. I.e. the caller need never do any of the counting.

PR #2878, opened 2019-11-22
Working on prototype for monitoring

blrubenstein:
Description:
This PR outlines a proof of concept (POC) of stackdriver monitoring integration in our API server. This is not intended to merge, just prove out the connectivity proof of concept. This PR connection generates a report like this:

Screen Shot 2019-11-22 at 12 47 13 PM

There's a task after this to start speccing out a design for how we are going to send signals from code from a code structure perspective, but that is probably a separate task
---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
@@ -1,13 +1,25 @@
 package org.pmiops.workbench.interceptors;
 
+import com.google.common.collect.Lists;
 import io.opencensus.common.Scope;
+import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
 import io.opencensus.exporter.trace.stackdriver.StackdriverTraceConfiguration;
 import io.opencensus.exporter.trace.stackdriver.StackdriverTraceExporter;
+import io.opencensus.stats.Aggregation;
+import io.opencensus.stats.BucketBoundaries;
+import io.opencensus.stats.Measure;
+import io.opencensus.stats.Stats;
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.stats.View;
+import io.opencensus.stats.ViewManager;
 import io.opencensus.trace.SpanBuilder;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
 import io.opencensus.trace.samplers.Samplers;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -25,17 +37,36 @@
   private static final Tracer tracer = Tracing.getTracer();
   private static final Logger log = Logger.getLogger(FirecloudApiClientTracer.class.getName());
   private static final String TRACE_ATTRIBUTE_KEY = "Tracing Span";
+  private static final StatsRecorder STATS_RECORDER = Stats.getStatsRecorder();
```
jaycarlton: nit: if we go this route, let's wrap this in a `Provider<StatsRecorder>`. That will allow us to mock it and capture arguments.

api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -68,5 +99,11 @@ public void afterCompletion(
       HttpServletRequest request, HttpServletResponse response, Object handler, Exception ex)
       throws Exception {
     ((Scope) request.getAttribute(TRACE_ATTRIBUTE_KEY)).close();
+    Random rand = new Random();
+    for (int i = 0; i < 100; i++) {
```
jaycarlton: Why do you need to create a random latency value? If you wanted to do that, you needn't use the interceptor.

jaycarlton: What you could do is make a test endpoint that accepts options for different distributions on different signals...

blrubenstein:
@jaycarlton I'll add a description. To be clear, this is not going to merge (ever). Its mainly a POC that we _can_ do this, so that when we have our lists of what we want to capture, we can signal on that. The code is kinda sloppy on purpose, cause I didn't want to spend time in making the choice _what_ of alerting service depend on a design of _how_ we implement.

jaycarlton:
Maybe our next trick is a prototype that you _can_ check in, that tracks a couple of useful, AoU-only signals. I don't think we need to spend time verifying all the documented functionality of the tool(s), but rather how well they fit our project.

blrubenstein:
I didn't see everything I expected in the plot, based on the description on the site that I was following. So may be worth digging into that. I am going to put together one place that details all the prototyping work and share it out, because I think at this point we've verified that Stackdriver Alerting has everything we need, so we can make the go/no-go decision on it. (And then close the ticket for deciding on an alert system)

Then I'd want your work on what alerts we want to be done before trying to actually start checking in the code on it I think?

PR #2879, opened 2019-11-22
[risk=no][RW-3995] Export Workspace Data - nitpick code review feedback

ericsong:
Description:
I addressed some lingering feedback from RW-3995 but I forgot to upload it before merging.

---
**PR checklist**

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -5,7 +5,6 @@ import static org.pmiops.workbench.tools.BackfillBillingProjectUsers.extractAclR
 import com.opencsv.bean.BeanField;
 import com.opencsv.bean.ColumnPositionMappingStrategy;
 import com.opencsv.bean.CsvBindByName;
-import com.opencsv.bean.StatefulBeanToCsv;
 import com.opencsv.bean.StatefulBeanToCsvBuilder;
 import java.io.FileWriter;
 import java.io.IOException;
@@ -144,10 +143,10 @@ public class ExportWorkspaceData {
 
       try (FileWriter writer =
           new FileWriter(opts.getOptionValue(exportFilenameOpt.getLongOpt()))) {
-          new StatefulBeanToCsvBuilder(writer)
-              .withMappingStrategy(mappingStrategy)
-              .build()
-              .write(rows);
+        new StatefulBeanToCsvBuilder(writer)
+            .withMappingStrategy(mappingStrategy)
+            .build()
+            .write(rows);
       }
     };
   }
@@ -175,16 +174,13 @@ public class ExportWorkspaceData {
     }
 
     Collection<DbCohort> cohorts = cohortDao.findByWorkspaceId(workspace.getWorkspaceId());
-    row.setCohortNames(
-        cohorts.stream().map(DbCohort::getName).collect(Collectors.joining(",\n")));
+    row.setCohortNames(cohorts.stream().map(DbCohort::getName).collect(Collectors.joining(",\n")));
     row.setCohortCount(String.valueOf(cohorts.size()));
 
     Collection<DbConceptSet> conceptSets =
         conceptSetDao.findByWorkspaceId(workspace.getWorkspaceId());
     row.setConceptSetNames(
-        conceptSets.stream()
-            .map(DbConceptSet::getName)
-            .collect(Collectors.joining(",\n")));
+        conceptSets.stream().map(DbConceptSet::getName).collect(Collectors.joining(",\n")));
     row.setConceptSetCount(String.valueOf(conceptSets.size()));
 
     Collection<DbDataset> datasets = dataSetDao.findByWorkspaceId(workspace.getWorkspaceId());
@@ -197,9 +193,7 @@ public class ExportWorkspaceData {
           notebooksService.getNotebooks(
               workspace.getWorkspaceNamespace(), workspace.getFirecloudName());
       row.setNotebookNames(
-          notebooks.stream()
-              .map(FileDetail::getName)
-              .collect(Collectors.joining(",\n")));
+          notebooks.stream().map(FileDetail::getName).collect(Collectors.joining(",\n")));
       row.setNotebooksCount(String.valueOf(notebooks.size()));
     } catch (NotFoundException e) {
       row.setNotebookNames("Error: Not Found");
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -21,7 +21,6 @@ import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.lang3.reflect.FieldUtils;
-import org.jetbrains.annotations.NotNull;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.CohortDao;
 import org.pmiops.workbench.db.dao.ConceptSetDao;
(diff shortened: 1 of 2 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -97,86 +98,37 @@ WorkspacesApi workspaceApi(Provider<WorkbenchConfig> configProvider) {
...
+
   @Bean
   public CommandLineRunner run(
-      WorkspaceDao workspaceDao,
-      CohortDao cohortDao,
-      ConceptSetDao conceptSetDao,
-      DataSetDao dataSetDao,
-      NotebooksService notebooksService,
-      WorkspaceFreeTierUsageDao workspaceFreeTierUsageDao,
+      WorkspaceDao workspaceDaoArg,
```
jaycarlton: Is the "Arg" suffix meant to indicate that these are passed-in and not injected? Or can you not do `this.workspaceDao = workspaceDao` because you're not in a constructor?

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
 
-      beanWriter.write(rows);
-
-      writer.close();
+        beanWriter.write(rows);
+      }
     };
   }
 
+  private WorkspaceExportRow toWorkspaceExportRow(DbWorkspace workspace) {
```
jaycarlton: This is a small point, but worth considering if you want to generalize to other representations. I feel like the target class WorkspaceExportRow should contain the knowledge of how the columns are formatted, (or indeed if there are columns at all), and this method should pass in raw data in its most natural form (like integers instead of strings).

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+
+    try {
+      row.setCollaborators(
+          extractAclResponse(
+                  workspacesApi.getWorkspaceAcl(
+                      workspace.getWorkspaceNamespace(), workspace.getFirecloudName()))
+              .keySet().stream()
+              .collect(Collectors.joining(",\n")));
+    } catch (ApiException e) {
+      row.setCollaborators("Error: Not Found");
```
jaycarlton: I wouldn't put error messages inside individual cells in the CSV file, especially if you're not going to log an error. For large data sets (i.e. more than one screenful of rows), it's just way too easy to miss things, and then you'll see the error messages show up in visualizations, which is embarrassing.

 Rather I'd leave an empty cell and log any problems.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+    }
+
+    Collection<DbCohort> cohorts = cohortDao.findByWorkspaceId(workspace.getWorkspaceId());
+    row.setCohortNames(
```
jaycarlton: I'd simply have a `row.setCohorts(Collection<DbCohort> cohorts)` and let the export row class be in charge of this formatting convention. It shortens this file a bit and allows you to support things like a Json field with a list of names, for example, without changing the logic here.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+        cohorts.stream().map(cohort -> cohort.getName()).collect(Collectors.joining(",\n")));
```
jaycarlton: nit: I think `.map(DbCohort::getName)` is no less expressive and is cleaner. When I see a named lambda argument, I slow down because I'm looking for something more complicated than a single-valued function.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+              .map(notebook -> notebook.getName())
+              .collect(Collectors.joining(",\n")));
+      row.setNotebooksCount(String.valueOf(notebooks.size()));
+    } catch (NotFoundException e) {
+      row.setNotebookNames("Error: Not Found");
+      row.setNotebooksCount("N/A");
+    }
+
+    DbWorkspaceFreeTierUsage usage = workspaceFreeTierUsageDao.findOneByWorkspace(workspace);
+    row.setWorkspaceSpending(usage == null ? "0" : String.valueOf(usage.getCost()));
```
jaycarlton: Please document (and present with the spreadsheet) any special cases like this, where we are promoting missing values to defaults. It can make a difference for analytics. For example, when computing averages, null values should be left out of the set, whereas zeros are typically valid data points.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -97,86 +98,37 @@ WorkspacesApi workspaceApi(Provider<WorkbenchConfig> configProvider) {
...
-        row.setWorkspaceSpending(usage == null ? "0" : String.valueOf(usage.getCost()));
-
-        row.setReviewForStigmatizingResearch(toYesNo(workspace.getReviewRequested()));
-        row.setWorkspaceLastUpdatedDate(dateFormat.format(workspace.getLastModifiedTime()));
-        row.setActive(toYesNo(workspace.isActive()));
-
-        rows.add(row);
+        rows.add(toWorkspaceExportRow(workspace));
 
         if (rows.size() % 10 == 0) {
```
jaycarlton: I'd make a constant and/or parameter for this 10 value. Maybe `PROGRESS_UPDATE_INTERVAL` or some such.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
       final CustomMappingStrategy mappingStrategy = new CustomMappingStrategy();
       mappingStrategy.setType(WorkspaceExportRow.class);
 
-      StatefulBeanToCsv beanWriter =
-          new StatefulBeanToCsvBuilder(writer).withMappingStrategy(mappingStrategy).build();
+      try (FileWriter writer =
```
jaycarlton: 👏 

nit: I'd probably name this `fileWriter` though.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+          new FileWriter(opts.getOptionValue(exportFilenameOpt.getLongOpt()))) {
+        StatefulBeanToCsv beanWriter =
```
jaycarlton: nit: the convention I like for builders is to break after the `.builder()` or constructor call in this case, then after every argument, and finally have `build()` on one line.

In this case, you could even keep going and call `.write()` on the next line write after `.build()`, since I don't think there's that much to see in the debugger for this kind of class. But I'd probably do it the way you have it here.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -190,15 +142,79 @@ public CommandLineRunner run(
...
+    row.setDatasetCount(String.valueOf(datasets.size()));
+
+    try {
+      Collection<FileDetail> notebooks =
+          notebooksService.getNotebooks(
+              workspace.getWorkspaceNamespace(), workspace.getFirecloudName());
+      row.setNotebookNames(
+          notebooks.stream()
+              .map(notebook -> notebook.getName())
+              .collect(Collectors.joining(",\n")));
```
jaycarlton: nit: define a constant or parameter or this joining string. Also, I'm really nervous about trying to embed sub-cell formatting this way. I'm not sure how well different tools will handle embedded newlines (and we'd want to test Windows as well I think). You just don't want to ask too much of CSV files, because lots of tools think they're compliant but I don't think there's a universal standard. But my opinion is based on experience a while back, so hopefully things are better now.

calbach:
Also this is not entirely consistent, but most of the tools use a space instead of an `=` to specify the argument value

PR #2884, opened 2019-11-23
[RW-4074][RISK=NO] Avoid null booleans when generating API model classes

jaycarlton:
By default, swagger-generated classes initialize `Boolean` values to null. This is very painful to work with, as we have to check for null and then interpret it as false.

If there are any cases where a null value is desired, then I propose that field should really be an enum. Optional booleans and null booleans do nothing but confuse and cause NPEs.

I'm still figuring out how to default `array` types to `[]` in swagger. I'm seeing mixed indications of support at the moment.


api/src/main/resources/cb_review_api.yaml
```diff
@@ -692,6 +692,7 @@ definitions:
         description: The Value of the annotation if the AnnotationType is DATE
       annotationValueBoolean:
         type: boolean
+        default: null
```
jaycarlton: @freemabd When I defaulted this to false, it broke a test case [here](https://github.com/all-of-us/workbench/blob/d2e1064dfbb793036a1e146613abbf00478122b8/api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java#L599). I didn't quite follow, but the test assumed something not to be set that was getting set when I made `annotationValueBoolean` default to false.

jaycarlton:
@jmthibault79 looks like CircleCI is happy now.

api/src/main/resources/cb_review_api.yaml
freemabd: null is the preferred default in this case, because this change will cause `annotationValueBoolean` to persist as false instead of null if it isn't provided by the client.

api/src/main/resources/cb_review_api.yaml
```diff
@@ -737,6 +738,7 @@ definitions:
         description: The Value of the annotation if the AnnotationType is DATE
       annotationValueBoolean:
         type: boolean
+        default: false
```
freemabd: This should also default to null for the same reasons as the above comment.

api/src/main/resources/cb_review_api.yaml
jaycarlton: Could we change it to `annotationValueNullableBoolean`?

Let me think how to write a unit test for those cases.

api/src/main/resources/cb_review_api.yaml
jaycarlton: ok cool.

api/src/main/resources/cb_review_api.yaml
freemabd: We can if needed.

jaycarlton:
@jmthibault79 looks like checks are passed here. Are we good?

api/src/main/resources/cb_review_api.yaml
jaycarlton: OK. I'll defer to you if you think that's desirable. Ideally you'd change the `AnnotationType` to `NULLABLE_BOOOLEAN` at the same time if possible.

api/src/main/resources/cb_search_api.yaml
```diff
@@ -447,13 +452,15 @@ definitions:
       hasHierarchy:
         description: Whether criteria is linked to a tree.
         type: boolean
+        default: false
       hasAncestorData:
         description: Whether criteria needs lookup in the criteria_ancestor table.
         type: boolean
+        default: false
       isStandard:
         description: Reveals if this criteria is standard or source.
```
jaycarlton: @freemabd I think this would make a better enum {standard, source} instead of a boolean.

PR #2886, opened 2019-11-25
[RW-4008][risk=no] Audit Data Access and Admin User Enable/Disable Actions

jaycarlton:
Instrument Access control state changes.

- Added a new SimpleTargetProperty interface for non-model-backed target property definitions.
- Added Administrative action audit around enable/disable access
- Added event for system changes to user registration status
- Simplified `AuthDomainControllerTest` by mocking `UserService`. Setting up that object took half the test class, and we weren't really exercising it much in this test (and really shouldn't be)


api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
```diff
@@ -38,11 +44,17 @@
...
-    DbUser updatedUser = userService.setDisabledStatus(user.getUserId(), request.getDisabled());
+    final DbUser targetDbUser = userDao.findUserByEmail(request.getEmail());
+    final Boolean previousDisabled = targetDbUser.getDisabled();
+    final DbUser updatedTargetUser =
+        userService.setDisabledStatus(targetDbUser.getUserId(), request.getDisabled());
+    authDomainAuditAdapter.fireSetAccountEnabled(
+        updatedTargetUser.getUserId(), !request.getDisabled(), !previousDisabled);
     userService.logAdminUserAction(
-        user.getUserId(), "updated user disabled state", previousDisabled, request.getDisabled());
+        updatedTargetUser.getUserId(),
```
jaycarlton: Do we know yet if we can phase out the legacy admin action log once we verify the action audit logs do all we need here?

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -341,74 +368,85 @@ public void setDataUseAgreementNameOutOfDate(String newGivenName, String newFami
...
+      long userId,
+      Timestamp bypassTime,
+      BypassTimeTargetProperty targetProperty,
+      BiFunction<DbUser, Timestamp, DbUser> setter) {
+    return setBypassTimeWithRetries(
+        userDao.findUserByUserId(userId), bypassTime, targetProperty, setter);
+  }
+
+  public void setDataUseAgreementBypassTime(Long userId, Timestamp bypassTime) {
+    setBypassTimeWithRetries(
```
jaycarlton: I was pleased with how this came out. I shrunk these methods to one statement each, and added the audit instrumentation with a single parameter.

jaycarlton:
@ericsong if you could take a look at the test injection changes in UserServiceTest especially. That one gave me fits, but I think I got it sorted.

I also realized some fields weren't private in the data entry test, so I made them private, and then IJ told me a bunch of them weren't necessary, so I axed them.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    this.clock = clock;
+    this.adminDbUserProvider = adminDbUserProvider;
+    this.actionIdProvider = actionIdProvider;
+  }
+
+  @Override
+  public void fireSetAccountEnabled(
+      long targetUserId, boolean newEnabledValue, boolean previousEnabledValue) {
+    final ActionAuditEvent event =
+        new ActionAuditEvent(
```
calbach: This constructor seems ripe for ordering mistakes. Is this defined in Kotlin? Does Kotlin not provide an alternate builder interface perhaps?

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
```diff
@@ -0,0 +1,88 @@
...
+              dbUserProvider.get().getEmail(),
+              actionIdProvider.get(),
+              ActionType.EDIT,
+              TargetType.ACCOUNT,
+              AccountTargetProperty.REGISTRATION_STATUS.getPropertyName(),
+              targetUser.getUserId(),
+              previousDataAccessLevel.toString(),
+              dataAccessLevel.toString());
+      actionAuditService.send(event);
+    } catch (RuntimeException e) {
```
calbach: It's a bit abnormal to catch a RuntimeException. What is the need for catching this specific type, and why is it necessary to log it at this level?

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  @Override
+  public void fireAdministrativeBypassTime(
+      long userId,
+      BypassTimeTargetProperty bypassTimeTargetProperty,
+      Optional<Instant> bypassTime) {
+    try {
+      actionAuditService.send(
```
calbach: nit: consistency with above code - should use a variable or not in both locations

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
calbach: I'd file a ticket - does the new stuff cover everything and is there a need for a backfill?

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
```diff
@@ -38,11 +44,17 @@
...
   @AuthorityRequired({Authority.ACCESS_CONTROL_ADMIN})
   public ResponseEntity<Void> updateUserDisabledStatus(UpdateUserDisabledRequest request) {
-    DbUser user = userDao.findUserByEmail(request.getEmail());
-    Boolean previousDisabled = user.getDisabled();
-    DbUser updatedUser = userService.setDisabledStatus(user.getUserId(), request.getDisabled());
+    final DbUser targetDbUser = userDao.findUserByEmail(request.getEmail());
+    final Boolean previousDisabled = targetDbUser.getDisabled();
+    final DbUser updatedTargetUser =
+        userService.setDisabledStatus(targetDbUser.getUserId(), request.getDisabled());
+    authDomainAuditAdapter.fireSetAccountEnabled(
```
calbach: Is there a plan to reconcile this? I don't like this middle-ground where most of the codebase talks about "disabling" users, and then one part of the code switches to talking about "enabling" them. They are not conceptually exact inverses and I think there could have been a renaming or switch to an enum could have improved things instead without the inconsistency.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -151,7 +158,45 @@ public DbUser updateUserWithRetries(Function<DbUser, DbUser> userModifier, DbUse
     }
   }
 
-  private void updateDataAccessLevel(DbUser user) {
+  private void updateDataAccessLevel(DbUser dbUser) {
+    final DataAccessLevel previousDataAccessLevel = dbUser.getDataAccessLevelEnum();
```
calbach: Could you please split the audit changes and this into separate PRs? There's a lot of content already in the PR without these changes and I don't see how this relates to the rest.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapter.java
```diff
@@ -0,0 +1,5 @@
+package org.pmiops.workbench.actionaudit.adapters;
+
+public interface AuthDomainAuditAdapter {
```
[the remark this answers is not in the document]
jaycarlton: I've been looking for a good place to document these. There are several such adapters, but they don't inherit from anything. I even made an empty interface just to have somewhere to put a comment, but pulled that out after I couldn't think of anything useful to hang on the interface.

It's easier to see the intent of the pattern with the `WorkspaceAuditAdapter`. These little services provide a way to instrument controllers and services with minimal footprint. You just call `fireCreate(Workapace w, long id)` or something and the adapter is in charge of building the ActionAuditEvents and passing them along. In some cases, this removes a good deal of unwieldy code that would have to go in the controller or service.

Additionally, these services are insulated from the real Cloud Logging machinery, as they just pass opaque data class instances down to that service.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    this.actionAuditService = actionAuditService;
+    this.clock = clock;
+    this.adminDbUserProvider = adminDbUserProvider;
+    this.actionIdProvider = actionIdProvider;
+  }
+
+  @Override
+  public void fireSetAccountEnabled(
+      long targetUserId, boolean newEnabledValue, boolean previousEnabledValue) {
+    final ActionAuditEvent event =
```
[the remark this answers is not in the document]
jaycarlton: It's a holdover from my C++ days, where if I wanted to see something in the debugger easily, I'd better declare it. I think I started inlining these, so I don't mind doing that here.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
jaycarlton: Yes, it's abnormal. One overarching goal for this project is that if any code being instrumented fails or throws something, it won't be because of a problem with the auditing system. So the idea is to eat the exception and log an error. That way I'll still see a spike in errors and respond quickly, but I won't take down the UserService or WorkspacesController.

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
jaycarlton: I just thought it was a best practice to have any enum be "active high", and be named after a positive rather than a negative.

Is it worth doing a database migration and backfill to get the whole code base consistent? Definitely a tough sell. It's the classic tension between consistency and incremental improvement.

On the other hand, the auditing schema is flexible here, and I can certainly use an enum on my side. If you feel strongly that consistency trumps clarity (as measured by number of mental inversions to interpret the variable), then it's easy enough to take in an `isDisabled` and `previousWasDisabled` value, and rename the function.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: This refactoring gives me a single place to call `fireUpdateDataAccessActon()`, as now there's only one place we call `setDataAccessLevelEnum()`.

Additionally, I found the existing function a bit difficult to follow and tried to squeeze out some duplication so I could have smaller methods and understand what I was working with.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapter.java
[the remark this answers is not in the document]
calbach: Perhaps a README.md in this directory, or some other readme

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
calbach: It looks like the next level down already catches runtime exceptions. The only thing you'd feasibly be catching here is a programming error - is that the goal?

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
calbach: I see the connection, I just think it could be split up into multiple PRs, e.g. UserService and the corresponding tests. If I'm missing something and they are too deeply intertwined - fine. I will take a look through the rest of this PR later today given the size.

jaycarlton:
Filed [RW-4211](https://precisionmedicineinitiative.atlassian.net/browse/RW-4121) for replacing the `admin_action_history` table.

jaycarlton:
_This constructor seems ripe for ordering mistakes. Is this defined in Kotlin? Does Kotlin not provide an alternate builder interface perhaps?_

I looked for a builder, but it doesn't appear to be a thing. The easy solution is to translate whatever calling code you have into Kotlin so you can take advantage of the named parameters, but so far I haven't found anything in Java. I'll write and out-of-class builder for this spot and a couple of others.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -341,92 +367,88 @@ public void setDataUseAgreementNameOutOfDate(String newGivenName, String newFami
...
+  private void setBypassTimeWithRetries(
+      DbUser dbUser,
+      Timestamp bypassTime,
+      BypassTimeTargetProperty targetProperty,
+      BiConsumer<DbUser, Timestamp> setter) {
+    updateUserWithRetries(
         (u) -> {
-          u.setTwoFactorAuthBypassTime(bypassTime);
+          setter.accept(u, bypassTime);
+          userServiceAuditAdapter.fireAdministrativeBypassTime(
```
ericsong: Curious why you decided to tie the audit event to the retry loop instead of firing once when the user initiates the action.

Do we really need to know if an update was tried several times? I thought the idea was that an audit event should indicate a user's attempt to do something. With this retry loop, we technically can't distinguish between multiple attempts at bypassing versus a chain of failing calls.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+            adminDbUserProvider.get().getUserId(),
+            adminDbUserProvider.get().getEmail(),
+            actionIdProvider.get(),
+            ActionType.EDIT,
+            TargetType.ACCOUNT,
+            AccountTargetProperty.IS_ENABLED.getPropertyName(),
+            targetUserId,
+            Boolean.toString(previousEnabledValue),
+            Boolean.toString(newEnabledValue));
+    actionAuditService.send(event);
```
calbach: I'd honestly prefer you not catch here, but should be consistent with the other adapters

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
ericsong: I looked over the changes to `updateDataAccessLevel` and I think they're correct. I like the changes and I think its more readable than what we had before but to echo @calbach 's sentiments, it can be hard to give a good review when there's a lot to cover.

Splitting up the necessary refactorings into its own PR can make it easier on us to give a thorough review, especially for critical code paths like this one.

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/ModelBackedTargetProperty.kt
```diff
@@ -0,0 +1,9 @@
+package org.pmiops.workbench.actionaudit.targetproperties
+
+// This interface is a contract to allow any model type T to support
+// arbitrary String properties without any changes to T. Intended
+// to be implemented by enum classes.
+interface ModelBackedTargetProperty<T> : SimpleTargetProperty {
+    override val propertyName: String
```
calbach: Does this string value have to exactly match a property name? Or is this just for display purposes in the audit log? Would be good to specify here

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapter.java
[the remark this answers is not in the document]
ericsong: Looking at it more, also wondering if `Adapter` is the right name for these interfaces. Not clear to me what its "adapting". I feel like I get just as much context from a shorter name like `WorkspaceAuditer`.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -341,92 +367,88 @@ public void setDataUseAgreementNameOutOfDate(String newGivenName, String newFami
...
+   */
+  private void setBypassTimeWithRetries(
+      long userId,
+      Timestamp bypassTime,
+      BiConsumer<DbUser, Timestamp> setter,
+      BypassTimeTargetProperty targetProperty) {
+    setBypassTimeWithRetries(userDao.findUserByUserId(userId), bypassTime, targetProperty, setter);
+  }
+
+  private void setBypassTimeWithRetries(
```
calbach: Nice

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+  private ActionAuditService actionAuditService;
+  private Clock clock;
+  private Provider<DbUser> adminDbUserProvider;
+  private Provider<String> actionIdProvider;
+
+  @Autowired
+  public AuthDomainAuditAdapterImpl(
+      ActionAuditService actionAuditService,
+      Clock clock,
+      Provider<DbUser> adminDbUserProvider,
```
ericsong: what makes this an `admin` user provider? I'm assuming the provider being given is the same as the other ones since there isn't a qualifier.

api/src/test/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterTest.java
```diff
@@ -0,0 +1,89 @@
...
+  @Captor private ArgumentCaptor<ActionAuditEvent> eventCaptor;
+
+  @Autowired private AuthDomainAuditAdapter authDomainAuditAdapter;
+
+  @TestConfiguration
+  @Import({AuthDomainAuditAdapterImpl.class})
+  @MockBean({ActionAuditService.class})
+  static class Config {
+    @Bean(name = "ACTION_ID")
+    @Scope(value = ConfigurableBeanFactory.SCOPE_PROTOTYPE)
```
ericsong: Scope technically isn't necessary since a single constant value is being returned. You could also just inline the value here instead of referencing a static variable.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -151,7 +158,45 @@ public DbUser updateUserWithRetries(Function<DbUser, DbUser> userModifier, DbUse
     }
   }
 
-  private void updateDataAccessLevel(DbUser user) {
+  private void updateDataAccessLevel(DbUser dbUser) {
```
calbach: This looks good for the bypass (administrative) methods. I don't think this is handling the "normal" path anymore though, e.g. `submitDataUseAgreement`, `syncComplianceStatus`, `syncEraCommons` etc

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -41,52 +49,73 @@
...
+  @MockBean({
+    AdminActionHistoryDao.class,
+    FireCloudService.class,
+    ComplianceService.class,
+    DirectoryService.class,
+    UserServiceAuditAdapter.class
+  })
+  static class Configuration {
+    @Bean
+    @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
```
calbach: singleton should be the default, the convention is generally to not include that

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -41,52 +49,73 @@
...
+      return PROVIDED_CLOCK;
+    }
+
+    @Bean
+    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+    WorkbenchConfig getWorkbenchConfig() {
+      return providedWorkbenchConfig;
+    }
+
+    @Bean
```
ericsong: Without prototype, this will return the same value on every call to randomProvider.get() within the same test run.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -41,52 +49,73 @@
...
+    providedDbUser = user;
+
+    providedWorkbenchConfig = WorkbenchConfig.createEmptyConfig();
+
+    // Since we're injecting the same static instance of this FakeClock,
+    // increments and other mutations will carry across tests if we don't reset it here.
+    // I tried easier ways of ensuring this, like creating new instances for every test run,
+    // but it was injecting null clocks giving NPEs long after construction, and so far this
+    // is the only working approach I've seen.
+    PROVIDED_CLOCK.setInstant(START_INSTANT);
```
ericsong: if you make it a prototype and remove final, you should be able to reinitialize it here. Basically what we do for user and workbenchConfig above.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -213,4 +254,73 @@ public void testSyncTwoFactorAuthStatus() throws Exception {
...
     assertThat(user.getTwoFactorAuthCompletionTime()).isNull();
   }
+
+  @Test
+  public void testSetBypassTimes() {
+    DbUser dbUser = userDao.findUserByEmail(EMAIL_ADDRESS);
+    assertThat(dbUser.getDataUseAgreementBypassTime()).isNull();
+    assertThat(dbUser.getComplianceTrainingBypassTime()).isNull();
+    assertThat(dbUser.getBetaAccessBypassTime()).isNull();
+    assertThat(dbUser.getEraCommonsBypassTime()).isNull();
```
ericsong: are these asserts necessary? This is more testing our test code instead of our prod code.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapter.java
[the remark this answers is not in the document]
jaycarlton: Yeah, I'm not married to that name. Typically an adapter connects something we write to something somebody else wrote. Maybe `AuthDomainAuditManager` or something? With the understanding that a manager is below the service level. Or event `AuthDomainAuditService`.

I'd prefer to do that rename on a follow-on PR since this one is already bloated.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
jaycarlton: 💯 I looked but couldn't find one, so I wrote my own.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
jaycarlton: Good catch.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
jaycarlton: uses a declared builder now, so it can conditionally add optional value.

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/ModelBackedTargetProperty.kt
jaycarlton: This string is as close as we get to a formal schema for the audit log, so these values shouldn't change. They may often match field names on the model object, but by design I'm trying to avoid relying on those staying the same forever. Most of the schema is supposed to use user-friendly names and concepts, without being coupled to our DB (with the exception of ID columns).

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
jaycarlton: I'll buy your argument about the signature here matching the existing sense of the disabled columns and fields. I'm still thinking bout the best way to represent it in the audit schema, but I think your idea of an enum instead of a boolean fits well here.

The conversation around disabled & enabled not being 100% opposites makes me think we shouldn't be using booleans at all for this state on the DB side.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: Yeah, this one got away from me a bit, and it wasn't helped when I merged RW-4009 into this branch. Lots of folks were out so it sat for a while. Plus I was on the home stretch and ready to have all this behind me. I'll try to keep them shorter in the future.

I'll try to split this out, but if I can't get it in 30 minutes I don't know if it's worth it.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: I was thinking a failure in the retryer would skip firing the audit event. Let me verify that's what's happening.

I'm really confused why this particular code path needs its own custom retry logic. I'd have though we'd just want one retry wrapper for all DB actions. I.e. why is the user table special? 

Also, we don't have that many users. I don't understand how this situation was frequent enough to warrant writing all this retry stuff.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
jaycarlton: Right. But `random.nextInt()` or whatever would still be different each time, right? I'm not actually sure what we were doing before.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
jaycarlton: Is that preferred? I kind of like this mental model, just from a general aversion to reassigning references where possible.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/UserServiceAuditAdapterImpl.java
jaycarlton: Yeah, it's a bit belt-and-suspenders. I wanted a hermetic exception boundary at the level of the adapters, but you make a good point. Since the level below catches everything, we can drop this. Programming errors will show up soon enough.

jaycarlton: After removing the try/catch, I remembered one place where it's doing something for me. In the WorkspaceAuditAdapter I'm injecting a user provider, and there are situations [where it returns null](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/actionaudit/adapters/WorkspaceAuditAdapterImpl.kt#L206), because we're not in a user-initiated query, or we're just starting up, or something.

It's better to return gracefully in that situation anyway than it is to throw an NPE.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: PR is up at #2911

I haven't reverted any of the changes on this branch yet though; waiting on that to merge first.

api/src/main/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterImpl.java
jaycarlton: It's just to remind me which user is which, since in this case the administrative user is the agent and the main user is the target. It's not explicitly guaranteed that the user coming out of that provider has an admin authority, but if I understand the code I'm instrumenting, that should be the case if we land in one of these methods.

api/src/test/java/org/pmiops/workbench/actionaudit/adapters/AuthDomainAuditAdapterTest.java
jaycarlton: I wanted to match the scope on the real Bean, which delivers a different one each time. From the perspective of this test as it exists now, you're right, but I wanted to reserve the right to change the value in between calls in the same test.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
jaycarlton: It is. I wanted to remind myself it was deliberate since we've been struggling with these lately.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: Are you saying those should all funnel through the retry stuff? Or that they're broken now (in which case we likely have a unit test gap). I think I was thinking this was orthogonal to that.

calbach: Sorry, nevermind - I misunderstood the goal of the PR. I see that updateUser() is still calling into this, so those should still function.

I mistakenly thought this change was also resolving RW-3830

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
calbach: I'd prefer a consistent codebase, I don't think this is going to be useful for the next reader who comes along - they will probably wonder why this line specifies singleton, and then 10 lines later another leaves the default scope (also singleton)

jaycarlton: 👍

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: I see what you mean. Pulling the audit line out of the retrier and doing it last is the most consistent with the intent of the audit system and the least noisy as well. Good catch.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
jaycarlton: These are for future-proofing the assertions that come later. I want to make sure that there's only a time assigned because of the code under test, and not as a side effect of some earlier initialization logic. I'll put a little comment to that effect.

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/ModelBackedTargetProperty.kt
calbach: Please capture this in the interface comment here

PR #2888, opened 2019-11-25
Jaycarlton/factory initializer

jaycarlton:
This an attempt at a uniform, external initialization system for the Swagger-generated classes. In particular, it allows updating default-null collections and objects into reasonable defaults. This is helpful for testing and hopefully for simplifying API implementation as well.


jmthibault79:
In this plan, should we prevent instantiation of the Swagger model classes directly?

jaycarlton:
@jmthibault79 I wrote this stuff before understanding that we actually have the ability to tweak how the swagger models are generated. 

The problem with making the constructor private is that the factory method wouldn't be able to see it, so I'd need to research some more. It's very possible this approach goes nowhere.

jmthibault79:
Oh yeah, fixing this at generation time would be strongly preferred 👍

PR #2889, opened 2019-11-25
[RW-3914][Risk=moderate] Add billing alerting

blrubenstein:
Description:
This adds a signal for billing buffer from our code to stackdriver monitoring. This is to prove out the Stackdriver Monitoring integration, as well as ensure we have a proof of concept that has something meaningful attacked to it.

---
**PR checklist**


api/build.gradle
```diff
@@ -393,6 +393,7 @@ dependencies {
   compile "io.opencensus:opencensus-api:${OPENCENSUS_VERSION}"
   compile "io.opencensus:opencensus-exporter-trace-stackdriver:${OPENCENSUS_VERSION}"
   compile "io.opencensus:opencensus-impl:${OPENCENSUS_VERSION}"
+  compile "io.opencensus:opencensus-exporter-stats-stackdriver:${OPENCENSUS_VERSION}"
```
jaycarlton: nit: alphabetize these

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -16,6 +19,15 @@
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
+
+import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
```
jaycarlton: I'm assuming once we get the wrapper service going (i.e. the Instrumentation and Metrics Abstraction Layer), we won't have any of these third-party imports in individual classes.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -40,6 +52,13 @@
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
   @VisibleForTesting static final Duration ASSIGNING_TIMEOUT = Duration.ofMinutes(10);
+  private static final Measure.MeasureLong BUFFER_ENTRY_MEASUREMENT = Measure.MeasureLong.create(
+      "buffer_entries",
+      "The number of billing projects in the buffer",
+      "projects");
+  private static final BucketBoundaries PROJECT_BOUNDARIES = BucketBoundaries.create(
+      Lists.newArrayList(0d, 10d, 20d, 50d, 200d));
+  private static final StatsRecorder STATS_RECORDER = Stats.getStatsRecorder();
```
jaycarlton: Depending on how you want to do things, we could set up a Bean that would give you a `Provider<Stats>`, or maybe we could inject a StatsService directly.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -40,6 +52,13 @@
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
   @VisibleForTesting static final Duration ASSIGNING_TIMEOUT = Duration.ofMinutes(10);
+  private static final Measure.MeasureLong BUFFER_ENTRY_MEASUREMENT = Measure.MeasureLong.create(
+      "buffer_entries",
```
jaycarlton: We have an opportunity to get a bit clever here and have a hierarchy or some other convention. For example you could have "billing.billing_project_buffer_entries".

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -59,6 +78,20 @@ public BillingProjectBufferService(
     this.clock = clock;
     this.fireCloudService = fireCloudService;
     this.workbenchConfigProvider = workbenchConfigProvider;
+    View view = View.create(
```
jaycarlton: Can we do all this in one place, so we're not cluttering up individual classes being instrumented? If not, can I bet we can still use adapter services.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -59,6 +78,20 @@ public BillingProjectBufferService(
...
+        "The number of billing project buffer entries.",
+        BUFFER_ENTRY_MEASUREMENT,
+        Aggregation.Distribution.create(PROJECT_BOUNDARIES),
+        Collections.emptyList());
+    ViewManager viewManager = Stats.getViewManager();
+    viewManager.registerView(view);
+    try {
+      StackdriverStatsExporter.createAndRegister();
+    } catch (IOException e) {
+      log.log(Level.WARNING, "Failure to initialize stackdriver stats exporter");
```
jaycarlton: This is going to need to be ERROR or SEVERE so that our log-based alerts will let us know the primary alert channel is down.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -59,6 +78,20 @@ public BillingProjectBufferService(
     this.clock = clock;
     this.fireCloudService = fireCloudService;
     this.workbenchConfigProvider = workbenchConfigProvider;
+    View view = View.create(
+        View.Name.create("billing_project_buffer_entries"),
+        "The number of billing project buffer entries.",
```
jaycarlton: Does StackDriver distinguish between a counter and a gauge? We'd want this to be a gauge.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
blrubenstein: I think we would have a `registerView` method or something? (`registerSignal` maybe? To be system agnostic?)

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -40,25 +53,38 @@
...
   private final FireCloudService fireCloudService;
+  private final MonitoringService monitoringService;
   private final Provider<WorkbenchConfig> workbenchConfigProvider;
 
   @Autowired
   public BillingProjectBufferService(
       BillingProjectBufferEntryDao billingProjectBufferEntryDao,
       Clock clock,
       FireCloudService fireCloudService,
+      MonitoringService monitoringService,
```
jaycarlton: A pattern that has worked well in the auditing system was to introduce site-specific adapter services to handle the event construction logic for each class being instrumented. This minimized the intrusiveness of the auditing system and also insulated the code being monitored from details of the actual system.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: See https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#MetricKind

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -67,6 +85,8 @@ private Timestamp getCurrentTimestamp() {
 
   /** Makes a configurable number of project creation attempts. */
   public void bufferBillingProjects() {
+
+    monitoringService.sendLongSignal(BUFFER_ENTRY_MEASUREMENT, getCurrentBufferSize());
```
jaycarlton: For example, with an adapter service injected, you'd just have `billingProjectBufferMonitoringAdapter.logBufferSize(getBufferSize());`

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,16 @@
...
+import io.opencensus.stats.Measure;
+
+public interface MonitoringService {
+  public void registerSignal(
+      String signalName,
+      String signalDescription,
+      Measure measurementInformation,
+      Aggregation aggregation);
+
+  public void sendLongSignal(Measure.MeasureLong measurementInformation, Long longValue);
```
jaycarlton: nit: use `value` instead of `longValue`, since we're statically typed in Java.

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,16 @@
...
+
+  public void sendDoubleSignal(Measure.MeasureDouble measurementInformation, Double doubleValue);
```
jaycarlton: If we go this route, we can define a little helper value class like `DoubleSignal` that has members for the `MeasureDouble` and `Dobule` value, and then just call a method like `sendSignal(DoubleSignal ds)` overloaded with `sendSignal(LongValue lv)`.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverMonitoringService.java
```diff
@@ -0,0 +1,53 @@
...
+  StackdriverMonitoringService() {
+    try {
+      StackdriverStatsExporter.createAndRegister();
+    } catch (IOException e) {
+      log.log(Level.SEVERE, "Failure to initialize stackdriver stats exporter");
+    }
+  }
+
+  @Override
+  public void registerSignal(
```
jaycarlton: In the audit code, I made sure that each method was hermetic in terms of exceptions. Thus, I catch all `RuntimeException`s from StackDriver and just log errors. We don't ever want a stackdriver outage to take down the system (though that scenario isn't common).

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverMonitoringService.java
```diff
@@ -0,0 +1,53 @@
...
+
+  StackdriverMonitoringService() {
+    try {
+      StackdriverStatsExporter.createAndRegister();
+    } catch (IOException e) {
+      log.log(Level.SEVERE, "Failure to initialize stackdriver stats exporter");
+    }
+  }
+
+  @Override
```
jaycarlton: If we had an adapter service, we'd initialize this once in the constructor.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverMonitoringService.java
```diff
@@ -0,0 +1,53 @@
...
+      Measure measurementInformation,
+      Aggregation aggregation) {
+    View view =
+        View.create(
+            View.Name.create(signalName),
+            signalDescription,
+            measurementInformation,
+            aggregation,
+            Collections.emptyList());
+    ViewManager viewManager = Stats.getViewManager();
```
jaycarlton: We'll probably want to set up a Singleton Bean to provide the ViewManager so we can mock it in testing.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverMonitoringService.java
```diff
@@ -0,0 +1,53 @@
...
+    viewManager.registerView(view);
+  }
+
+  @Override
+  public void sendLongSignal(Measure.MeasureLong measurementInformation, Long longValue) {
+    STATS_RECORDER.newMeasureMap().put(measurementInformation, longValue).record();
```
jaycarlton: OK, so we could potentially have a number of data points we want to send at once, then put them all into a measure map and record at once. I bet we find a use for that (such as the counts of all the billing project states like CREATING, AVAILABLE, etc).

jaycarlton:
Also, please add a PR description.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
blrubenstein: Let's sync up on this tomorrow. I remember seeing that but didn't understand it as well as I could have, so pairing on the impl sounds good.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
blrubenstein: Because we're using opencensus, this is probably the link that matters:
https://cloud.google.com/monitoring/custom-metrics/open-census (found it though.)

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -3,13 +3,7 @@
...
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doReturn;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
blrubenstein: I think it is so we are clear what imports are getting used. * doesn't make it clear.

PR #2896, opened 2019-12-01
[risk=no] test cleanup

freemabd:
Removing the following from tests using `@DataJpaTest`

`@Import(LiquibaseAutoConfiguration.class)`
We don't use liquibase in dao test and this import keeps the test from throwing -> java.lang.IllegalStateException: Cannot find changelog location: class path resource [db/changelog/db.changelog-master.yaml]. So this auto configuration swallows this error and doesn't load any liquibase migrations. Replacing this with a liquibase property in the application.properties file liquibase.enabled=false just turns it off. The dao test have always used the create-drop configuration -> [Here](https://github.com/all-of-us/workbench/blob/master/api/src/test/resources/application.properties#L5). Basically `@DataJpaTest` loads all classes annotated with `@Entity` and uses Hibernate ddl-auto mode to create and drop tables.

`@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)`
By default, tests annotated with `@DataJpaTest` will use an embedded in-memory
database (replacing any explicit or usually auto-configured DataSource). The
'@AutoConfigureTestDatabase' annotation can be used to override these settings. So we don't need to replace it with NONE, since we don't want to override.

`@Transactional`
Removing this because you inherit it from `@DataJpaTest` -> [Here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/orm/jpa/DataJpaTest.java#L79)

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/OfflineAuditControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/OfflineAuditControllerTest.java
@@ -26,8 +26,6 @@ import org.pmiops.workbench.db.model.DbCdrVersion;
 import org.pmiops.workbench.db.model.DbUser;
 import org.pmiops.workbench.test.FakeClock;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration;
-import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
 import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.boot.test.mock.mockito.MockBean;
@@ -41,8 +39,6 @@ import org.springframework.transaction.annotation.Transactional;
 
 @RunWith(SpringRunner.class)
 @DataJpaTest
-@Import(LiquibaseAutoConfiguration.class)
-@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
 @DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD)
 @Transactional(propagation = Propagation.NOT_SUPPORTED)
 public class OfflineAuditControllerTest {
--- a/api/src/test/java/org/pmiops/workbench/cdr/dao/CBCriteriaAttributeDaoTest.java
+++ b/api/src/test/java/org/pmiops/workbench/cdr/dao/CBCriteriaAttributeDaoTest.java
@@ -8,20 +8,13 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.pmiops.workbench.cdr.model.DbCriteriaAttribute;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration;
-import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
-import org.springframework.context.annotation.Import;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.junit4.SpringRunner;
-import org.springframework.transaction.annotation.Transactional;
 
 @RunWith(SpringRunner.class)
 @DataJpaTest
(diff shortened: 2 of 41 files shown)


api/src/test/java/org/pmiops/workbench/cdr/dao/CBCriteriaAttributeDaoTest.java
```diff
@@ -8,20 +8,13 @@
...
-import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
-import org.springframework.context.annotation.Import;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.junit4.SpringRunner;
-import org.springframework.transaction.annotation.Transactional;
 
 @RunWith(SpringRunner.class)
 @DataJpaTest
-@Import(LiquibaseAutoConfiguration.class)
```
jaycarlton: Can you please add to your description why these are no longer needed (or were never needed)? Should these come out from all the tests in the backend?

aside: I've noticed that a couple of my tests are happier with this import (or at least IntelliJ warnings go away with it present), even if they don't interact with any DAOs at all. I'm wondering if there's some side effect of this configuration that I should be pulling in.

api/src/test/resources/application.properties
```diff
@@ -4,7 +4,7 @@ spring.datasource.username=sa
 spring.datasource.password=sa
 spring.jpa.hibernate.ddl-auto=create-drop
 spring.jpa.properties.hibernate.dialect=org.pmiops.workbench.cdr.CommonTestDialect
-
+liquibase.enabled=false
```
jaycarlton: Is this for speed or correctness? Are we still running the liquibase migrations regularly?

api/src/test/java/org/pmiops/workbench/cdr/dao/CBCriteriaAttributeDaoTest.java
freemabd: Updated the description. I don't think they are hurting anything by having them in the dao test but they can be removed. I don't see any problem in Intellij.

api/src/test/resources/application.properties
freemabd: We have never loaded the liquibase migrations in the dao test. We could probably write a different test that validates the liquibase migrations. Currently the way to validate migrations if by running `./project.rb dev-up` or a sub command of this.

jaycarlton: OK, thanks. There might be a slight risk of an entity change or misunderstood design masking a schema issue, but I think we're OK.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -165,8 +163,6 @@
 
 @RunWith(SpringRunner.class)
 @DataJpaTest
-@Import(LiquibaseAutoConfiguration.class)
-@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
 @DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD)
 @Transactional(propagation = Propagation.NOT_SUPPORTED)
```
jaycarlton: Is this `@Transactional` not needed any other places?

api/src/test/java/org/pmiops/workbench/api/OfflineAuditControllerTest.java
```diff
@@ -41,8 +39,6 @@
 
 @RunWith(SpringRunner.class)
```
jaycarlton: If we need these 4 annotations for all our tests, I wonder if we can't either make a new annotation that combines them or a parent class that sets these for us. Not for this PR, though.

freemabd: Agreed.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
freemabd: So any test that has `@Transactional(propagation = Propagation.NOT_SUPPORTED)` i left alone since they are overriding the behavior of Transactional. So I only removed `@Transactional` since it's inherited from `@DataJpaTest`

PR #2898, opened 2019-12-04
[RW-3689][risk=no] Google Analytics Events - Workspaces

ericsong:
Dev Notes
- I tried to trigger the event as close to the onClick as much as possible
- I wanted to centralize the logic so there's a single place to look at what events we're tracking
- The `getCurrentPageLabel()` bit might be controversial but I felt like it was the cleanest way to infer where we currently are on the page. The traditional way passing the info down through props/args would have resulted in a *lot* of new info being shuttled around. 

Things I learned
- One form of Google Analytics tracking is firing “Events”.  “Events” have three levels specificity of Category, Action, and Label. See https://support.google.com/analytics/answer/1033068#Anatomy
- As it applies to our application, I view “Category” as a product area, “Action” as a verb that the user would do, “Label” as miscellaneous info. For my Events, I used “Label” as a way to distinguish between different ways a user can get perform the same action.  
- One way to test Google Analytics Events is to view the Network tab in the developer console. Look for a request going out to Google and view its query params to verify that the Category/Action/Label you wanted is there.
- Request access from Greg Jordan and/or file a PD ticket
- There is a Google Analytics project per environment.
- The project that it’s going to is specified by the `gaId` in the front end environment variables.
- The Google Analytics console will show you which GA IDs map to which project/environment.
- **Developing locally will send events to the TEST environment, not local. This is probably due to the fact that local development uses the `test-env-base` for most of its config.
- Data Studio is report generator that uses data from Google Analytics. They are two separate products with separate access controls.
- Google Analytics reports take 24-48 hours to update
- They offer a real time report that is useful for testing but only shows Event Category, and Event Action. Event Label is missing
- **Google Analytics will not record events unless you create a filter to include your requests. The default is to ignore all incoming requests unless there’s a filter to allow it through. I had to change the test environment’s filter to include localhost hostnames. 
- We have two versions of the homepage right now. Make sure changes go to both.

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/workspace/workspace-edit.tsx
+++ b/ui/src/app/pages/workspace/workspace-edit.tsx
@@ -533,6 +533,18 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
       return fp.includes(populationEnum, this.state.workspace.researchPurpose.populationDetails);
     }
 
+    onSaveClick() {
+      if (this.isMode(WorkspaceEditMode.Create)) {
+        AnalyticsTracker.Workspaces.Create();
+      } else if (this.isMode(WorkspaceEditMode.Duplicate)) {
+        AnalyticsTracker.Workspaces.Duplicate();
+      } else if (this.isMode(WorkspaceEditMode.Edit)) {
+        AnalyticsTracker.Workspaces.Edit();
+      }
+
+      this.saveWorkspace();
+    }
+
     async saveWorkspace() {
       try {
         this.setState({loading: true});
@@ -907,16 +919,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
               </ul>
             } disabled={!errors}>
               <Button type='primary'
-                      onClick={() => {
-                        if (this.isMode(WorkspaceEditMode.Create)) {
-                          AnalyticsTracker.Workspaces.Create();
-                        } else if (this.isMode(WorkspaceEditMode.Duplicate)) {
-                          AnalyticsTracker.Workspaces.Duplicate();
-                        } else if (this.isMode(WorkspaceEditMode.Edit)) {
-                          AnalyticsTracker.Workspaces.Edit();
-                        }
-                        this.saveWorkspace();
-                      }}
+                      onClick={() => this.onSaveClick()}
                       disabled={errors || this.state.loading}
                       data-test-id='workspace-save-btn'>
                 {this.renderButtonText()}
--- a/ui/src/app/pages/workspace/workspace-edit.tsx
+++ b/ui/src/app/pages/workspace/workspace-edit.tsx
@@ -908,7 +908,13 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
             } disabled={!errors}>
               <Button type='primary'
                       onClick={() => {
-                        AnalyticsTracker.Workspaces.Create();
+                        if (this.isMode(WorkspaceEditMode.Create)) {
+                          AnalyticsTracker.Workspaces.Create();
+                        } else if (this.isMode(WorkspaceEditMode.Duplicate)) {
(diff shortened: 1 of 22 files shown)


jaycarlton:
Can we make a lick config file to either separate local events or apply a
label so they can be filtered out?

jaycarlton:
*local

gjuggler:
> Can we make a lick config file to either separate local events or apply a label so they can be filtered out?

I think the confusion here is that we have two local configs: environment.local.ts, and environment.localtest.ts. We do have a separate Analytics property ID for local reporting; if we add the following line to localtest, then all events from local UI server will go to the "local" property which may be more ideal.

  gaId: 'UA-112406425-5',

jaycarlton:
Oh awesome. So the front end has a totally separate config system from the backend? I guess I would've assumed they'd all be per-environment but commingled across tiers/ends.

Thanks.

jaycarlton:
@ericsong your GA notes are valuable. Would be good to have them in a markdown doc in the repo or other handy place.

ericsong:
~~I tried that out but I don't think it's working. I think there may be a bug with the build script where the `environment.localtest.ts` file doesn't get swapped in for the `dev` configuration.~~

~~`yarn dev-up --configuration=dev` will not swap in `environment.localtest.ts`~~

~~Interestingly, I found out that `yarn dev-up` is not an alias for `yarn dev-up --configuration=dev`. After changing the `"dev": { "browserTarget": "aou-workbench-ui:build:test"}` to `"dev": { "browserTarget": "aou-workbench-ui:build:local"}`, we see the divergence in behavior.~~

~~`yarn dev-up --configuration=dev` will grab `environment.local.ts`~~
~~`yarn dev-up` will not swap in any environment file (same behavior in the bug we see above)~~

~~Targeting the local environment does work but that requires spinning up a local API server.~~

I think I figured it out. There were two bugs with our build script.
1. The `dev` configuration was pointing to `test`. Fixed by changing `test` to `dev`.
2. The default configuration was invalid so nothing would get swapped in. Fixed by setting it to `dev`. 

This answers the weird behavior I was seeing earlier.

Things I learned
- The environments specified by `build.configuration` are independent from the ones specified by `serve.configuration`
- It looks like `serve.configuration` is determined by the configuration flag that we pass in with the command line ex. `yarn dev-up --configuration=local`
- The `browserTarget` in `serve.configuration` will specify which `build.configuration` environment to use.
  - ex. `dev: { browserTarget: 'aou-workbench-ui:build:test'}` will use the `build.configuration.test` config

ericsong:
@jaycarlton thanks. My plan is to keep creating these lists with every PR I put up and compiling them together and organizing it a bit at some cadence.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -910,7 +906,17 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                 {errors.diseaseOfFocus && <div>You must specify a disease of focus</div>}
               </ul>
             } disabled={!errors}>
-              <Button type='primary' onClick={() => this.saveWorkspace()}
+              <Button type='primary'
+                      onClick={() => {
+                        if (this.isMode(WorkspaceEditMode.Create)) {
```
NehaBroad: The if else logic can be moved out of render onClick and into its own function

jaycarlton:
@gjuggler How expensive is it to have a separate local config for each developer? E.g. `environment.local.jaycarlton.ts` or some such? Or is it better to tag the values in some other way?

I guess maybe we don't need to do much individualized debugging on these.

gjuggler:
> @gjuggler How expensive is it to have a separate local config for each developer? E.g. `environment.local.jaycarlton.ts` or some such? Or is it better to tag the values in some other way?
> 
> I guess maybe we don't need to do much individualized debugging on these.

I don't think we've ever had a use case for individual checked-in environment files. I'm sure it's possible, but it may be extra complexity. You can always just locally modify your local file if you need to test some changes.

PR #2899, opened 2019-12-04
[no ticket][risk=no] Improve trace method name & labels for inbound API spans

gjuggler:
Dev notes

* This PR makes some small changes / additions to our tracing code to improve the readability of spans representing incoming AoU API calls.
* The changes mostly improve the way we extract information from Swagger-generated API methods and whatnot.
* One potentially controversial change: the "Test" prefix is gone from the span name. I realized that treating it as a label probably made more sense.
* I coded this one rainy Friday afternoon several weeks back... it's been sitting in a local branch since then and I finally had a chance to wrap it into shape. It's super low-priority.

Before:
Screen Shot 2019-12-03 at 9 58 26 PM
Screen Shot 2019-12-03 at 9 58 31 PM

After:
Screen Shot 2019-12-03 at 9 57 41 PM
Screen Shot 2019-12-03 at 9 57 52 PM

(Note: the "Recv" prefix is auto-added by Stackdriver when the span type is set to Span.Kind.SERVER, see [code](https://github.com/census-instrumentation/opencensus-go/commit/b8f3b913f034c5682f7255d5d51efb6f151fb09b))

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
@@ -10,14 +10,13 @@ import io.opencensus.trace.SpanBuilder;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
 import io.opencensus.trace.samplers.Samplers;
+import io.swagger.annotations.ApiOperation;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.inject.Provider;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-
-import io.swagger.annotations.ApiOperation;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.firecloud.FirecloudApiClientTracer;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


jaycarlton:
Neat. I wonder if we can pull stack driver metrics straight off these a
spans to track individual API times or if we would need a separate
interceptor for that.

gjuggler:
> Neat. I wonder if we can pull stack driver metrics straight off these a spans to track individual API times or if we would need a separate interceptor for that.
> […](#)

That would be great follow-up – it looks like GAE doesn't send per-path latency metrics to Stackdriver by default, so this might be a reasonable place to hook into that. (It would be a good way, for instance, to alert specifically if workspace creation latency is above a certain expected value.)

We could also do the same for tracking & alerting on latencies in our outgoing API calls (more on that in a second, likely more controversial but also more interesting PR).

api/src/main/java/org/pmiops/workbench/interceptors/TracingInterceptor.java
```diff
@@ -45,28 +51,57 @@ public TracingInterceptor(Provider<WorkbenchConfig> workbenchConfigProvider) {
...
       requestSpanBuilder.setSampler(Samplers.alwaysSample());
     }
+    Scope requestScopedSpan = requestSpanBuilder.startScopedSpan();
 
-    Scope requestSpan = requestSpanBuilder.startScopedSpan();
-    request.setAttribute(TRACE_ATTRIBUTE_KEY, requestSpan);
+    // Log some additional key-value attributes, which will be visible in the details pane on
+    // Stackdriver.
+    final Span currentSpan = tracer.getCurrentSpan();
+    currentSpan.putAttribute(
```
blrubenstein: I think the env is actually implicit in the stackdriver project we see it in. Although this could help with screenshots. (I have no objections with this being here, just noting)

gjuggler: Ack – good point. I'll leave it here though, because for local testing, traces go to Test project but we want to see the "local" indication somewhere at least.

gjuggler:
> I think tracing also batches, and submits at once, so we may not get the latency we want for monitoring API times if we use tracing to drive it. (Not 100% sure on that though)

Indeed – I interpreted Jay's comment as suggesting that we could use the same interceptor approach to log Stackdriver metrics tagged with API names.

PR #2901, opened 2019-12-04
[no ticket][risk=no] Improve labeling and metadata on downstream API requests

gjuggler:
Dev notes

* This is the sister PR to #2899 (which improved trace metadata for Workbench API spans). This focuses on downstream API calls, e.g. FireCloud, Moodle, etc.
* My solution involved a slightly controversial decision: to create a locally-modified copy of the Swagger-codegen template we've been using to generate our API client classes. There is quite a lot of power to this approach, but it also risks our API client code becoming out of sync with mainline Swagger-codegen. (This is potentially mitigated by the fact that this file hasn't been updated for the past ~2.5 years in Swagger-codegen.)

I think the improvements are probably worth the potential risks. The resulting traces are much clearer, and we may be interested in hooking into the API client template code in more ways in the future (e.g. for exporting monitoring metrics).

Before:

Screen Shot 2019-12-04 at 10 01 30 AM

After:

Screen Shot 2019-12-04 at 2 13 09 PM

**PR checklist**


api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+import javax.validation.executable.ExecutableValidator;
+import java.util.Set;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+{{/performBeanValidation}}
+
+{{#imports}}import {{import}};
+{{/imports}}
+
+{{! Begin AoU-specific additions }}
```
jaycarlton: Can this not be an open/close pair like the others? E.g. `#allOfUsAddition`?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+public class {{classname}} {
+    private static final Tracer tracer = Tracing.getTracer();
+    private ApiClient {{localVariablePrefix}}apiClient;
+
+    public {{classname}}() {
+        this(Configuration.getDefaultApiClient());
+    }
+
+    public {{classname}}(ApiClient apiClient) {
+        this.{{localVariablePrefix}}apiClient = apiClient;
```
jaycarlton: Can you declare meta-variables as you go? So that you only have to do the string operation `{{localVariablePrefix}}apiClient` once?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    public ApiClient getApiClient() {
```
jaycarlton: If this were Kotlin, you'd be able to skip all this 😉

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
jaycarlton: Java doesn't need a local variable prefix, because it has `this.` I thought things like `m_apiClient` were a relic of older IDEs that can't style locals differently. Is there some external requirement to support these prefixes? 

Or do you want something like `s3apiClient`? If so, `apiPrefix` would be better than `localVariablePrefix` I think.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+import io.opencensus.common.Scope;
+import io.opencensus.trace.AttributeValue;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Status;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.lang.reflect.Type;
+{{! End AoU-specific additions }}
+{{^fullJavaUtil}}
+import java.util.ArrayList;
```
jaycarlton: I think this is fine, but was just observing that it looks like these imports won't be in the canonical order after generation. Totally not worth fixing.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+     * @param progressListener Progress listener
+     * @param progressRequestListener Progress request listener
+     * @return Call to execute
+     * @throws ApiException If fail to serialize the request body object
+     */
+    public com.squareup.okhttp.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}final ProgressResponseBody.ProgressListener progressListener, final ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
+        Object {{localVariablePrefix}}localVarPostBody = {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
+
+        // create path and map variables
+        String {{localVariablePrefix}}localVarPath = "{{{path}}}"{{#pathParams}}
```
jaycarlton: Do all these string acrobatics need to be generated, or could you write a utility function that gets called from here?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+            .replaceAll("\\{" + "{{baseName}}" + "\\}", {{localVariablePrefix}}apiClient.escapeString({{{paramName}}}.toString())){{/pathParams}};
+
+        {{javaUtilPrefix}}List<Pair> {{localVariablePrefix}}localVarQueryParams = new {{javaUtilPrefix}}ArrayList<Pair>();{{#queryParams}}
+        if ({{paramName}} != null)
+        {{localVariablePrefix}}localVarQueryParams.addAll({{localVariablePrefix}}apiClient.parameterToPairs("{{#collectionFormat}}{{{collectionFormat}}}{{/collectionFormat}}", "{{baseName}}", {{paramName}}));{{/queryParams}}
```
jaycarlton: Do `{{#queryParams}}` and `{{/queryParams}}` need to be on the same line, or could you add line breaks after the end of the statement?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+        final String {{localVariablePrefix}}localVarAccept = {{localVariablePrefix}}apiClient.selectHeaderAccept({{localVariablePrefix}}localVarAccepts);
+        if ({{localVariablePrefix}}localVarAccept != null) {{localVariablePrefix}}localVarHeaderParams.put("Accept", {{localVariablePrefix}}localVarAccept);
+
+        final String[] {{localVariablePrefix}}localVarContentTypes = {
+            {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}}
+        };
+        final String {{localVariablePrefix}}localVarContentType = {{localVariablePrefix}}apiClient.selectHeaderContentType({{localVariablePrefix}}localVarContentTypes);
+        {{localVariablePrefix}}localVarHeaderParams.put("Content-Type", {{localVariablePrefix}}localVarContentType);
+
+        if(progressListener != null) {
```
jaycarlton: Can we use an `Optional<ProgressListener` instead of a nullable?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+            apiClient.getHttpClient().networkInterceptors().add(new com.squareup.okhttp.Interceptor() {
```
jaycarlton: Can you use a lambda instead of an anonymous class here?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+     */
+    public com.squareup.okhttp.Call {{operationId}}Call({{#allParams}}{{{dataType}}} {{paramName}}, {{/allParams}}final ProgressResponseBody.ProgressListener progressListener, final ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
+        Object {{localVariablePrefix}}localVarPostBody = {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
+
+        // create path and map variables
+        String {{localVariablePrefix}}localVarPath = "{{{path}}}"{{#pathParams}}
+            .replaceAll("\\{" + "{{baseName}}" + "\\}", {{localVariablePrefix}}apiClient.escapeString({{{paramName}}}.toString())){{/pathParams}};
+
+        {{javaUtilPrefix}}List<Pair> {{localVariablePrefix}}localVarQueryParams = new {{javaUtilPrefix}}ArrayList<Pair>();{{#queryParams}}
+        if ({{paramName}} != null)
```
jaycarlton: It might be nicer to do this with an optional stream if possible, something like

```
        final Map<String, String> localVarHeaderParams = Optional.ofNullable(
apiClient.selectHeaderAccept(localVarAccepts))
            .map(val -> ImmutableMap.of("Accept", val))
            .orElse(ImmutableMap.of());
```

You could even have a little helper method like `buildParamMap(headerAccepts, "Accept")`

gjuggler:
@jaycarlton – maybe I should have been clearer in the PR or file comments, but I didn't write the entire api.mustache file from scratch; I only made 20-30 lines of changes to add Stackdriver tracing. Before this PR, the original api.mustache file would be invoked via the swagger-codegen library; the change here is that we're overriding the main api client template with a forked copy and some local changes. (I think this approach is laid out in https://blog.arnoldgalovics.com/using-a-custom-template-for-swagger-codegen-with-gradle/, and the template comments link to the original file in swagger-codegen repo.)

Many of your comments are aimed at bringing the entire file into consistency with our house style. IMO, making more changes to this file than necessary would be onerous, and would increase the risk of regressions.

If you have high-level concerns with the approach of having this code replicated inside our repo that's a reasonable objection to this PR as a whole, but distinct from issues w/ the style of the surrounding code.

I'm happy to add some clarifying text in the template top-of-file comments if you think it would help clarify how I'm proposing we handle this file.

jaycarlton:
Can we see side-by-side your changes to the master mustache file?

One ulterior motive for simplifying and cleaning up the generated code is so I can understand better what it's doing and look for further consolidation and abstractions. I'm not 100% convinced we need codegen for this, but I still don't have my bearings. The other motive is that if we expect humans to write code compliant with our standards, we should expect nothing less from codegen tools. If that's riskier than a developer making the same changes to their own code, then I'm not sure this is the toolchain for us. But I need to sit down with you and go over the whole flow.

gjuggler:
> Can we see side-by-side your changes to the master mustache file?
> 
> One ulterior motive for simplifying and cleaning up the generated code is so I can understand better what it's doing and look for further consolidation and abstractions. I'm not 100% convinced we need codegen for this, but I still don't have my bearings. The other motive is that if we expect humans to write code compliant with our standards, we should expect nothing less from codegen tools. If that's riskier than a developer making the same changes to their own code, then I'm not sure this is the toolchain for us. But I need to sit down with you and go over the whole flow.

I surrounded all changes with comments (as noted in the top of the file). I can create a gist with a file-level diff against the master template later this afternoon, to ensure I didn't miss anything.

jaycarlton:
@gjuggler if someone makes a PR against swagger-codegen and they approve the changes, then will you rebase against a Java 8 version of the template?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
jaycarlton: Also, I'm wondering if putting the aou-specific code into separate files and including them into the main template would make sense. Then it would be more obvious what's ours and what's theres.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+        span.putAttribute("http.host", AttributeValue.stringAttributeValue(apiClient.getBasePath()));
+        span.putAttribute("http.path", AttributeValue.stringAttributeValue("{{path}}"));
+        span.putAttribute("http.method", AttributeValue.stringAttributeValue("{{httpMethod}}"));
+        span.putAttribute("description", AttributeValue.stringAttributeValue("{{summary}}"));
+        {{#returnType}}span.putAttribute("responseType", AttributeValue.stringAttributeValue("{{{returnType}}}"));{{/returnType}}
+        span.putAttribute("connectionTimeout", AttributeValue.longAttributeValue(apiClient.getConnectTimeout()));
+
+        try {
+            // Create and execute the API request.
+            {{#returnType}}ApiResponse<{{{returnType}}}> {{localVariablePrefix}}resp = {{/returnType}}{{operationId}}WithHttpInfo({{#allParams}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}});{{#returnType}}
```
jaycarlton: Putting directives like `{{#returnType}}` on their own lines helps readability here. I think it's a readability win even if the generated code has extra blank lines.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+        // https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md
+        span.putAttribute("package", AttributeValue.stringAttributeValue("{{package}}"));
+        span.putAttribute("http.host", AttributeValue.stringAttributeValue(apiClient.getBasePath()));
+        span.putAttribute("http.path", AttributeValue.stringAttributeValue("{{path}}"));
+        span.putAttribute("http.method", AttributeValue.stringAttributeValue("{{httpMethod}}"));
+        span.putAttribute("description", AttributeValue.stringAttributeValue("{{summary}}"));
+        {{#returnType}}span.putAttribute("responseType", AttributeValue.stringAttributeValue("{{{returnType}}}"));{{/returnType}}
+        span.putAttribute("connectionTimeout", AttributeValue.longAttributeValue(apiClient.getConnectTimeout()));
+
+        try {
```
jaycarlton: Can you do a try-with-resources block here on `scope`?

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+            scope.close();
+        }
+    {{! End AoU-specific additions }}
+    }
+
+    /**
+     * {{summary}}
+     * {{notes}}{{#allParams}}
+     * @param {{paramName}} {{description}}{{#required}} (required){{/required}}{{^required}} (optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}{{/allParams}}
+     * @return ApiResponse&lt;{{#returnType}}{{returnType}}{{/returnType}}{{^returnType}}Void{{/returnType}}&gt;
```
jaycarlton: double check these escape sequences. I expect it's right.

api/build.gradle
```diff
@@ -427,9 +429,10 @@ dependencies {
   testCompile 'org.liquibase:liquibase-core:3.5.3'
   testCompile 'org.bitbucket.radistao.test:before-after-spring-test-runner:0.1.0'
   testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:${MOCKITO_KOTLIN_VERSION}"
-  // For Swagger generation. These should include dependencies found in the Swagger codegen
-  // example here:
-  // https://github.com/swagger-api/swagger-codegen/blob/v2.2.3/samples/client/petstore/spring-stubs/pom.xml
+
+  // Dependencies for Swagger codegen. This should include all dependencies required by Swagger's
```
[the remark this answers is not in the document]
gjuggler: Sure, that would be nice. It hasn't felt like too urgent of a task though, since there are some nice things about having this stuff all in one place.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -138,7 +138,7 @@ public StatusApi statusApi(WorkbenchConfig workbenchConfig) {
   }
 
   public static ApiClient buildApiClient(WorkbenchConfig workbenchConfig) {
-    ApiClient apiClient = new FirecloudApiClientTracer();
+    ApiClient apiClient = new ApiClient();
```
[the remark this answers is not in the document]
gjuggler: This code, and the removed files, is removing the custom instrumentation that's now replaced by centrally adding traces into each swagger-generated api client.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
+{{!
```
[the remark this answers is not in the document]
gjuggler: See https://gist.github.com/gjuggler/68342857c3fb39983a37625b5ad42410 for an example generated API, before and after this PR, with a diff.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: Acknowledged. As discussed offline, I'm inclined to leave as-is for now, until we get a better feel for exactly how to treat overridden Swagger templates such as this.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: /noideadog.

I'll resolve most of this and the following comment threads, as I don't want to dig into too many details around the swagger-codegen project's code style or design decisions. I've tried to limit the scope of the template changes made in this PR – here's the old and new template contents, with a diff: https://gist.github.com/gjuggler/0a485915da5e68e26150e9decf4a7d67

As I dug into this, I did realize / remember that the actual version of api.mustache that I'd copied over is the version from the 2.2.3 release, which is the version we import in Gradle and was released in ~July 2017. Here is the file link: https://github.com/swagger-api/swagger-codegen/blob/v2.2.3/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/api.mustache

I've added some more comments to the top of the file, as well as to where we import swagger-codegen, about this gnarly implicit dependency. Unfortunately I don't think there's a clean & nice way to make it easier to override a specific part of this codegen template. If we upgrade our version of swagger-codegen, it's almost certainly going to require us to update our template overrides to be compatible with the newer version. An interesting question is whether this add'l friction to upgrade would be worth the benefit of having our ability to modify how Swagger is generating things.

api/build.gradle
```diff
@@ -4,7 +4,9 @@ import org.gradle.api.tasks.JavaExec
 import org.pmiops.workbench.tooling.GenerateAPIListingTask
 import org.pmiops.workbench.tooling.IncrementalHotSwapTask
 
-def swaggerTemplateDir = 'src/main/resources'
+// This path is the Java sources root for finding any overridden Swagger-codegen
+// templates we may be using.
+def swaggerTemplateDir = 'src/main/resources/Java'
```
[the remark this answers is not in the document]
gjuggler: Good idea – I renamed to "swaggercodegen". I think at some point I'd thought the exact name of the template directory affected whether Swagger would pick it up, but I confirmed otherwise.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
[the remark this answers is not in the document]
jaycarlton: Thanks. I see there's a ` System.out.println("asdf")` statement in there. Is that intentional?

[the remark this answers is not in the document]
gjuggler: Leftover from debugging – I wanted some positive proof that the new template was getting picked up when I was renaming the directory. I removed before checking it in.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,277 @@
...
+     */
+    public {{#returnType}}{{{returnType}}} {{/returnType}}{{^returnType}}void {{/returnType}}{{operationId}}({{#allParams}}{{{dataType}}} {{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) throws ApiException {
+        {{! Begin AoU-specific additions }}
+        // Generate an OpenCensus tracing span for this API method call.
+        Scope scope = tracer.spanBuilderWithExplicitParent("{{classname}}.{{operationId}}", tracer.getCurrentSpan())
+            .setSpanKind(Span.Kind.CLIENT).startScopedSpan();
+        Span span = tracer.getCurrentSpan();
+        // Some of these attribute keys use naming from the OpenCensus standard. See
+        // https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md
+        span.putAttribute("package", AttributeValue.stringAttributeValue("{{package}}"));
```
jaycarlton: Would it make sense to have a utility class to build and populate the span and its attributes and just call that form here? This is a pretty repetitive style. Another approach would be to populate a map here and pass it in.

Even fancier might be a system where you instantiate a span-aware api caller and pass it the code to execute via a lambda. That class could simply be handwritten java, so we don't need to tweak the template as much.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: Wow – that worked great. This helps mitigate the concern that upgrading to future Swagger versions would be difficult; the AoU-specific material is now much more isolated. 👍

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: I'd originally used a map, but moved back to this style. It's not significantly different, and this ended up being more consistent with the direct span.putAttribute calls in the request-handling and exception-handling cases.

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: It's tempting, but I don't think it would significantly add to the clarity here. There were also some pretty scary warnings about try-with-resources on these scoped spans, which may affect the error-handling case: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/SpanBuilder.java#L236

api/src/main/resources/Java/libraries/okhttp-gson/api.mustache
gjuggler: Done.

gjuggler:
I can't reply to the original comment, but your suggestion re: importing the AoU-specific stuff via mustache templating was great – this worked like a charm, and definitely clarifies what's "ours" vs. "theirs".

gjuggler:
Thanks Jay for the helpful comments & suggestions. I'll be interested to see if you find other ways to apply this approach in the future!

api/src/main/resources/swaggercodegen/libraries/okhttp-gson/aouTraceImports.mustache
```diff
@@ -0,0 +1,7 @@
+import io.opencensus.common.Scope;
+import io.opencensus.trace.AttributeValue;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Status;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.lang.reflect.Type;
```
jaycarlton: nit: you probably want a newline at the end of this file.

api/src/main/resources/swaggercodegen/libraries/okhttp-gson/api.mustache
```diff
@@ -0,0 +1,239 @@
...
+import javax.validation.executable.ExecutableValidator;
+import java.util.Set;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+{{/performBeanValidation}}
+
+{{#imports}}import {{import}};
+{{/imports}}
+
+{{>aouTraceImports}}
```
jaycarlton: You _might_ want to have a switch to generate the vanilla template (without the aou calls) for future testing/debugging.  So something like
```
{{#aouTracingEabled}}
{{>aouTraceImports}}
{{/aouTracingEabled}}
``` 
That would let you quickly determine that your changes are scoped to the places you expect. This would become more helpful if there were more than one layer of customizations in play.

PR #2900, opened 2019-12-04
[RW-4083][risk=no] Google Analytics - Dataset Builder

ericsong:
Description:

Building off of #2898 

- [ ] This PR meets the Acceptance Criteria in the JIRA story
- [x] The JIRA story has been moved to Dev Review
- [ ] This PR includes appropriate unit tests
- [x] I have run and tested this change locally
- [ ] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers
- [ ] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later

PR #2902, opened 2019-12-04
[RW-4084][risk=no] Google Analytics - Notebooks

ericsong:
Description:

Building off of #2898 and #2900 

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/analysis/notebook-resource-card.tsx
+++ b/ui/src/app/pages/analysis/notebook-resource-card.tsx
@@ -8,12 +8,12 @@ import {dropNotebookFileSuffix} from 'app/pages/analysis/util';
 import {workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
 import {formatRecentResourceDisplayDate} from 'app/utils';
+import {AnalyticsTracker} from 'app/utils/analytics';
 import {encodeURIComponentStrict} from 'app/utils/navigation';
 import {ResourceType} from 'app/utils/resourceActions';
 import {CopyRequest, RecentResource} from 'generated/fetch';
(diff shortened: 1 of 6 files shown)


ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -146,6 +160,8 @@ export const NotebookResourceCard = fp.flow(
   }
 
   copyNotebook(copyRequest: CopyRequest) {
+    AnalyticsTracker.Notebooks.Copy();
+
```
NehaBroad: nit: extra line

ui/src/app/pages/analysis/notebook-resource-card.tsx
```diff
@@ -146,6 +160,8 @@ export const NotebookResourceCard = fp.flow(
   }
 
   copyNotebook(copyRequest: CopyRequest) {
+    AnalyticsTracker.Notebooks.Copy();
```
NehaBroad: Question why only in this case the AnalyticsTracker is mentioned inside the copyNotebook  function rather than in render

ericsong: It's because the render function is in a more generic `CopyModal` component. If I added the tracker there, I would have to add some logic so that the generic `CopyModal` knows what its parent component is.

I thought the cleaner way forward would be to add the tracker in this function since its what gets called `onClick`.

ui/src/app/pages/analysis/notebook-resource-card.tsx
ericsong: I was hoping to keep this line to space out different blocks of logic.

PR #2903, opened 2019-12-05
[RW-3769][RW-3770][risk=no] Update help sidebar with FontAwesome icons

dolbeew:
- Install FontAwesome and update sidebar icons based on new design
- Remove Data Dictionary and add Compute Configuration (disabled for now)
- Add tooltips to icons
---
**PR checklist**


ui/.npmrc
```diff
@@ -0,0 +1,2 @@
+@fortawesome:registry=https://npm.fontawesome.com/
+//npm.fontawesome.com/:_authToken=5D54A493-EC53-4B6E-972E-5A200343CFC7
```
blrubenstein: Nit: newline at end of file.

blrubenstein: Is that auth token secret? It looks like a registry key, which we should probably keep secret?

dolbeew: Deleted this file

dolbeew: Good catch. Switched to adding the pro icons individually as svg files

ui/src/app/components/help-sidebar.tsx
```diff
@@ -260,22 +284,18 @@ export const HelpSidebar = withUserProfile()(
...
-            </a>
-          </div>
-          {helpContent === 'reviewParticipantDetail' &&
-            <div style={activeIcon === 'annotations' ? iconStyles.active : styles.icon}>
-              <ClrIcon shape='note' size={32}  title='Participant Status and Annotations'
-                onClick={() => this.onIconClick('annotations')} />
-            </div>
-          }
+          {icons.map((icon, i) => (!icon.page || icon.page === helpContent) && <div key={i} style={{display: 'table'}}>
+            <TooltipTrigger disabled={tooltipId !== i} content={<div>{icon.tooltip}</div>}>
```
blrubenstein: Could you put the tooltip off to the left? (I think `side=left`?) Right now it is off the page

blrubenstein: Opt nit: I would do this as `content={tooltipId !== i && <div>...}` rather than the disabled? I think it reads a little easier, and avoids disabled prop accidentally propagating.

dolbeew: Done

dolbeew: Agreed, done.

PR #2911, opened 2019-12-06
[no ticket][risk=no] User Service refactorings for audit

jaycarlton:
PR #2886 is pretty bloated, so I'm pulling these changes out as requested by @calbach.

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
+++ b/api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
@@ -6,15 +6,15 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import java.sql.Date;
 import java.sql.Timestamp;
+import java.time.Clock;
 import java.time.Instant;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.Random;
-import javax.inject.Provider;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.mockito.Mock;
 import org.pmiops.workbench.compliance.ComplianceService;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.model.DbUser;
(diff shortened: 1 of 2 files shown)


jaycarlton:
@calbach From comment [here](https://github.com/all-of-us/workbench/pull/2886#discussion_r354642074) I'll check on the other methods in addition to the bypass time setters. Was trying to limit footprint, but you know how that goes.

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -107,101 +132,113 @@ public void testSyncComplianceTrainingStatus() throws Exception {
     assertThat(user.getComplianceTrainingExpirationTime()).isEqualTo(new Timestamp(12345));
 
     // Completion timestamp should not change when the method is called again.
-    CLOCK.increment(1000);
+    incrementClock();
```
blrubenstein: I'm not sure this abstraction gives us much other than obfuscating milliseconds. I could see an argument for pulling it out, but I'd like it to either be configurable or clear how much it increments by. (So either `incrementClockOneThousandMillis` or `incrementClock(Integer millis)`

jaycarlton: If this were a public API I'd probably agree with you. Private methods (especially in test classes) don't need to be big reusable abstractions. 1000 ms is an internal detail, since the actual value doesn't really matter much for our purposes, so I hid it. I explicitly don't want the reader to have to care how big the increment is in this case, as it's irrelevant to the test. Anyone interested can dive into its implementation and immediately see it. A better name might be `tick()`.

Finally, there's a big difference between Encapsulation and Obfuscation. Encapsulation makes things simpler by hiding private details, whereas Obfuscation makes things more complex or difficult to understand by introducing convoluting details, constraints, and redirections.

blrubenstein: Talked offline. We agreed that `tick()` might be better. My problem was it added enough of a redirection while not hiding enough private details for me. So it felt like it added a redirect without actually giving us much. (It hides the 1000, but for me, because it was `increment`, I thought the number was relevant)

PR #2912, opened 2019-12-09
[RW-3998][risk=low] Reject previews of large notebooks, add preview page error handling

calbach:
5MB is chosen as the cutoff here. From manual testing, I found that with our current GAE task configuration, previews would start failing around ~10MB. I couldn't find any good tooling for analyzing memory use on GAE Java servers, there is probably room for improvement here. It is possible that we are still susceptible to OOMs from concurrent preview requests on a single server with this approach.

Too large error:
image

Unexpected error:
image

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/google/CloudStorageService.java
+++ b/api/src/main/java/org/pmiops/workbench/google/CloudStorageService.java
@@ -43,9 +43,11 @@ public interface CloudStorageService {
   GoogleCredential getGarbageCollectionServiceAccountCredentials(String garbageCollectionEmail)
       throws IOException;
 
-  JSONObject getFileAsJson(String bucketName, String fileName);
-
   Map<String, String> getMetadata(String bucketName, String objectPath);
+
+  Blob getBlob(String bucketName, String objectPath);
+
+  JSONObject readBlobAsJson(Blob blob);
 
   void deleteBlob(BlobId blobId);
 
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/pmiops/workbench/google/CloudStorageService.java
```diff
@@ -43,10 +43,12 @@
   GoogleCredential getGarbageCollectionServiceAccountCredentials(String garbageCollectionEmail)
       throws IOException;
 
-  JSONObject getFileAsJson(String bucketName, String fileName);
```
jmthibault79: This is still used by `NotebooksServiceImplTest`

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -28,6 +29,7 @@
 @Service
 public class NotebooksServiceImpl implements NotebooksService {
 
+  private static final long MAX_NOTEBOOK_READ_SIZE_BYTES = 5 * 1000 * 1000; // 5MB
```
jmthibault79: can this be a config value instead?

calbach: Yes, but I prefer it to not be in this case. My policy on when one should use a flag vs constant:

- Is the value likely to change?
- Is there likely to be a need to change the value dynamically / out of band of a release?

In this case, I'd say no and no, which makes me prefer a const. The disadvantages of using a flag are:
- Introduces an additional moving part which your tests/devs need to cover
- Obfuscates the runtime value for a reader
- Adds additional configuration code

This case is definitely borderline, but I will default to a constant unless given a reason to do otherwise.

calbach:
Tested manually be pushing my own GAE version

PR #2897, opened 2019-12-02
[RW-3907][risk=moderate] "non-breaking" updates for UI packages/dependencies

andy7i:
Description:

---
**PR checklist**

- [ ] This PR meets the Acceptance Criteria in the JIRA story
- [x] The JIRA story has been moved to Dev Review
- [x] This PR includes appropriate unit tests
- [x] I have run and tested this change locally
- [ ] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers
- [ ] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later

After getting my brain jostled around, this was interesting to come back to. Skip to the bottom for a summary. Risk level was set to moderate just in case, since this includes a lot (~700) of updated dependencies.

Backstory:
The ticket was created after seeing this list of vulnerabilities from `yarn audit` :

```
594 vulnerabilities found - Packages audited: 66320
Severity: 65 Low | 3 Moderate | 526 High
✨  Done in 2.71s.
```

Not all of them were updated, but ~500 "high severity" vulnerabilities were updated. Latest `yarn audit`:
```
83 vulnerabilities found - Packages audited: 69645
Severity: 68 Low | 2 Moderate | 13 High
✨  Done in 2.10s.
```

--------

Summary:
Of the 594 vulnerabilities (Severity: 65 Low | 3 Moderate | 526 High) from `yarn audit`, ~500 were updated and 83 remain (Severity: 68 Low | 2 Moderate | 13 High).

The changes here reflect those updates as well as a few test fixes for these "non-breaking" changes. For example, highcharts needed a specific version in order to not break functionality. There are some other updates to remove test warnings and clean up a couple tests.

There is more in-depth info in the commit messages (and in my notes) which I can add to the PR description if that would be helpful.

There is one warning introduced to the tests which is not resolved in this PR. This will require removing `componentWillReceiveProps` from `ConceptTable`. [Ticket for this issue is here](https://precisionmedicineinitiative.atlassian.net/browse/RW-4106)

PR #2914, opened 2019-12-09
[risk=low][RW-4136] Remove buildOptimization from our angular builds

ericsong:
Description:
This fixes a bug where our workspace lists page was broken on staging, perf, stable, and prod. The bug is caused by a combination of issues

1. angular optimization/buildOptimizer can create breaking optimizations. This is likely due to a particularly aggressive setting for uglify which is non configurable through angular, https://github.com/angular/angular-cli/issues/11439.
2. We updated some of our libraries to a version which is affected by the issue above. https://github.com/all-of-us/workbench/pull/2897

There were a few possible solutions
1. Revert the version upgrade PR. We didn't want to do this because it would also remove other changes which we felt like were good to keep.
2. Remove `buildOptimizer` from our angular builds. The con here is that the build step may take produce larger output files.
3. Upgrade our angular-cli tools to version 8, where they changed the default options for uglify to be less aggressive. https://github.com/angular/angular-cli/pull/14187. This was ruled out due to the possibility of introducing other bugs by making a major version upgrade.

This PR implements solution 2 as the loss in compression is fairly small. 3.03 MB with buildOptimizer to 3.37 MB without it.

**Things I learned**

- We have different front end build settings for our various environments. This is controlled by `architect.configurations` in `angular.json`
- We use angular build tools to assemble our assets, optimize, create source maps, etc. 
- The front end build stuff I learned about in https://github.com/all-of-us/workbench/pull/2898 was pretty useful here. 

---
**PR checklist**

### DIFF
--- a/ui/angular.json
+++ b/ui/angular.json
@@ -76,8 +76,7 @@
               "namedChunks": false,
               "aot": true,
               "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true
+              "vendorChunk": false
             },
             "staging": {
               "fileReplacements": [
@@ -93,8 +92,7 @@
               "namedChunks": false,
               "aot": true,
               "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true
+              "vendorChunk": false
             },
             "stable": {
               "fileReplacements": [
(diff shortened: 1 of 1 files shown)


jaycarlton:
Double-check that we don't need a certain uglify level for security. I'd think not since we're open source...

ericsong:
@jaycarlton Good point but I'll use your counterpoint as a reason for why I think its a minor issue.

ericsong:
sounds good @gjuggler

PR #2915, opened 2019-12-09
[RW-4128][risk=no] Extending the metrics framework

jaycarlton:
Added some testing and ability to fire multiple signals at once.


api/src/main/java/org/pmiops/workbench/monitoring/MonitoringServiceOpenCensusImpl.java
```diff
@@ -0,0 +1,159 @@
...
+  private void registerSignals() {
+    Arrays.stream(MonitoringViews.values())
+        .map(MonitoringViews::toStatsView)
+        .forEach(viewManager::registerView);
+  }
+
+  @Override
+  public void recordValue(OpenCensusStatsViewInfo viewProperties, Number value) {
+    try {
+      initStatsConfigurationIdempotent();
```
jaycarlton: @ericsong here's the hack I was talking about. I put it in a separate service so we can safely unit test this monitoring service.

jaycarlton:
Removing reviewers for a while as I've got Spring problems apparently. Not sure how it found my bean in one place but not others.

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -42,15 +48,22 @@
   private static final int PROJECT_BILLING_ID_SIZE = 8;
   @VisibleForTesting static final Duration CREATING_TIMEOUT = Duration.ofMinutes(60);
   @VisibleForTesting static final Duration ASSIGNING_TIMEOUT = Duration.ofMinutes(10);
-  private static ImmutableMap<BufferEntryStatus, Duration> statusToGracePeriod =
+  private static final ImmutableMap<BufferEntryStatus, Duration> statusToGracePeriod =
```
jmthibault79: nit: could you make these maps ALL_CAPS

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
```diff
@@ -74,14 +87,43 @@ private Timestamp getCurrentTimestamp() {
...
   }
 
+  // TODO(jaycarlton): Move most (all) of this to its own cron method on a controller dedicated to
+  // polling gauge metric values.
+  private void updateBillingProjectBufferMetrics() {
+    ImmutableMap.Builder<OpenCensusStatsViewInfo, Number> signalToValueBuilder =
+        ImmutableMap.builder();
+    signalToValueBuilder.put(MonitoringViews.BILLING_BUFFER_SIZE, getCurrentBufferSize());
+    signalToValueBuilder.put(MonitoringViews.DEBUG_MILLISECONDS_SINCE_EPOCH, clock.millis());
+    signalToValueBuilder.put(MonitoringViews.DEBUG_RANDOM_DOUBLE, random.nextDouble());
```
jmthibault79: what is this for?

api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceTest.java
```diff
@@ -116,14 +150,36 @@ public void testGetReadOnlyHtml_allowsDataImage() {
...
-    when(firecloudService.staticNotebooksConvert(any()))
+    when(mockCloudStorageService.getFileAsJson(any(), any())).thenReturn(new JSONObject());
+    when(mockFirecloudService.staticNotebooksConvert(any()))
         .thenReturn("<img src=\"https://eviltrackingpixel.com\" />\n");
 
     String html = new String(notebooksService.getReadOnlyHtml("", "", "").getBytes());
     assertThat(html).doesNotContain("eviltrackingpixel.com");
   }
+
+  @Test
```
jmthibault79: love how clean these are

api/src/test/java/org/pmiops/workbench/monitoring/MonitoringServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+  // We require that the measure map exists when created, and returns itself after any entry
+  // is added, but we won't be using the result directly.
+  private void initMockMeasureMap() {
+    doReturn(mockMeasureMap).when(mockStatsRecorder).newMeasureMap();
+    doReturn(mockMeasureMap).when(mockMeasureMap).put(any(MeasureLong.class), anyLong());
+    doReturn(mockMeasureMap).when(mockMeasureMap).put(any(MeasureDouble.class), anyDouble());
+  }
+
+  @Test
+  public void testRecordIncrement() {
```
jmthibault79: nit: could you add `testRecordValue` for some other N != 1, for completion?

api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java
jaycarlton: They're just toy timeseries so we can experiment even with low traffic pre-launch (or on non-production environments). Metrics backends tend to flake when there are very low-traffic signals. I want a couple of curves that always have something to say (particularly in nearly dead environments).

Also, I didn't have a double-valued time series yet. I will soon, but this lets us exercise both sides of that coin right away.

blrubenstein: Do you mind throwing a comment to that effect above it? Just so we have it documented here?

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterInitializationService.java
```diff
@@ -0,0 +1,41 @@
...
+      try {
+        StackdriverStatsExporter.createAndRegister(stackdriverStatsConfiguration);
+        initialized = true;
+        return true;
+      } catch (IOException e) {
+        logger.log(Level.WARNING, "Failed to initialize global StackdriverStatsExporter.", e);
+        return false;
+      }
+    } else {
+      return false;
```
blrubenstein: Opt nit: TOL if it is already initialized, should we return true? Because it is live and ready to go?

api/src/test/java/org/pmiops/workbench/notebooks/NotebooksServiceTest.java
jaycarlton: Thanks. I'm a complete slob in every area except code.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterInitializationService.java
jaycarlton: oh good catch.

I'm not using that return value yet, and maybe it should be an enum 🤔

gjuggler:
Dropping a note here after chatting offline with Jay about CircleCI failures (e.g. https://circleci.com/gh/all-of-us/workbench/62647).

The key signal here is that CircleCI is failing on the "loadConfig" step, which is driven by a command-line Gradle task. These tasks have some subtle differences in their Spring configuration, since they're never run within the context of a web request. (See https://github.com/all-of-us/workbench/pull/1803 for some context & gory details on how we got to this setup.)

I suspect what's going on is that adding the MonitoringService as a dependency to NotebooksService is making command-line tools fail, because NotebooksService is loaded by one or more of our command-line tools. The place to look for this is the CommandLineToolConfig, https://github.com/all-of-us/workbench/blob/2ad7ff14bbe3023f65e007937a037904a5388062/api/tools/src/main/java/org/pmiops/workbench/tools/CommandLineToolConfig.java#L44-L44

There are a couple of ways to add package-level and class-level component scanning to that config... the first thing I'd try would be to add the monitoring package as a whole and see if that fixes the issue. To re-test, either push the change and let CircleCI do its thing (probably faster for now), or run a local dev-up command.

jaycarlton:
@gjuggler thanks! The other amusing thing about this issue is that I can replicate it in CircleCI and dev-up, but not appEngineRun (at least on my machine). But that's likely because the latter doesn't do the config step.

PR #2918, opened 2019-12-11
[RW-1911][Risk=low] Srubenst/reactify error handler

blrubenstein:
Description:
This replaces the angular global error handler with a react component to handle errors. This also tears out some old angular services and rolls them into the component. It does not handle automatically adding this global handling to all api calls. That is to be addressed in a follow up PR. (I think this still provides value and that will require more research)

---
**PR checklist**


ui/src/app/pages/data/data-page.tsx
```diff
@@ -114,7 +115,7 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
         isLoading: true
       });
       const [cohorts, cohortReviews, conceptSets, dataSets] = await Promise.all([
-        cohortsApi().getCohortsInWorkspace(namespace, id),
+        fetchWithGlobalErrorHandler(() => cohortsApi().getCohortsInWorkspace(namespace, id)),
```
blrubenstein: This is an example of how you would manually call it. If you wanted error handling that was not done via the global handler, you would add it after the api call within the `fetchWithGlobalErrorHandler` method.

ui/src/app/components/error-handler.tsx
```diff
@@ -0,0 +1,132 @@
...
+  closeError() {
+    globalErrorStore.next(undefined);
+  }
+
+  render() {
+    const {globalError} = this.props;
+    const {apiDown, firecloudDown, notebooksDown} = this.state.serverStatus;
+
+    return globalError !== undefined && <React.Fragment>
+      {globalError.statusCode === 500 && <div style={styles.errorHandler}>
```
NehaBroad: This can be combined statusCode 500 and 503

ui/src/app/components/error-handler.tsx
```diff
@@ -0,0 +1,132 @@
...
+  notebooksDown: boolean;
+}
+
+interface Props {
+  globalError: ErrorResponse;
+}
+
+interface State {
+  serverStatus: ServerStatus;
+  serverStatusAcknowledged: boolean;
```
NehaBroad: Does this mean serverStatus change acknowledged?

ui/src/app/components/error-handler.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// We need to maintain this outside of the component, because new errors could come in at any time and we don't want to
+// have it reconstruct/reinitialize with new errors.
+const shouldCheckStatus = new BehaviorSubject<any>(true);
+
+export const ErrorHandler = withGlobalError()(class extends React.Component<Props, State> {
+  obs;
```
NehaBroad: Whats this used for?

ui/src/app/utils/retry.tsx
```diff
@@ -17,3 +20,70 @@ export async function fetchAbortableRetry<T>(fetchFn: () => Promise<T>, timeoutM
...
 }
+
+
+
+export async function fetchWithGlobalErrorHandler<T>(fetchFn: () => Promise<T>, maxRetries: number = 3): Promise<T> {
+  let retries = 0;
+  while (true) {
+    try {
+      const returnVal = await fetchFn();
+      console.log(returnVal);
```
NehaBroad: So this is going to log the returnVal if the api is successful , i wonder if we do want to print the success result of each API thats going to use this

ui/src/app/utils/retry.tsx
```diff
@@ -17,3 +20,70 @@ export async function fetchAbortableRetry<T>(fetchFn: () => Promise<T>, timeoutM
     }
   }
 }
+
+
+
+export async function fetchWithGlobalErrorHandler<T>(fetchFn: () => Promise<T>, maxRetries: number = 3): Promise<T> {
```
NehaBroad: It will be great to add description of what this method is doing,

ui/src/app/components/error-handler.tsx
```diff
@@ -0,0 +1,132 @@
...
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      serverStatus: {apiDown: false, firecloudDown: false, notebooksDown: false},
```
NehaBroad: Wondering if this should be serverStatus or serverDownStatus

ui/src/app/components/error-handler.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.serverStatus !== prevState.serverStatus) {
+      this.setState({serverStatusAcknowledged: false});
+    }
+    const {globalError} = this.props;
+    if (globalError !== prevProps.globalError && globalError !== undefined) {
+      if ((globalError.statusCode === 500
+          || globalError.statusCode === 503)
+          && shouldCheckStatus.getValue()) {
```
NehaBroad: i am not sure i understand the what shouldCheckStatus is doing, if there is a new globalError with the either code 500 and 503 we set this to false and in pingServer set it to True. Is it like a static variable that is like yes we know there is a service down and we are taking care of it so do not send a new request again to pingServer?

blrubenstein: Exactly. Otherwise if the api is down, it would get a 500, and decide it should check status, ping again, and infinite loop. This ensures that it only pings once, and doesn't try again while a server check is already in progress.

PR #2920, opened 2019-12-11
[risk=no]Split free_tier_credits_limit_override in 2: dollars and days

jmthibault79:
I prefer small PRs so I split off this independent piece of RW-3660.  It's also a prerequisite to other billing stories, such as the Override API.

Description:

---
**PR checklist**

### DIFF
--- a/api/config/config_perf.json
+++ b/api/config/config_perf.json
@@ -21,7 +21,8 @@
     "retryCount": 4,
     "bufferCapacity": 300,
     "bufferRefillProjectsPerTask": 5,
-    "defaultFreeCreditsLimit": 100.0,
+    "defaultFreeCreditsDollarLimit": 100.0,
+    "defaultFreeCreditsDaysLimit": 180,
     "garbageCollectionUserCapacity": 1000,
     "garbageCollectionUsers": [
       "aou-rw-gc-perf-1@all-of-us-rw-perf.iam.gserviceaccount.com",
--- a/api/config/config_perf.json
+++ b/api/config/config_perf.json
@@ -21,6 +21,7 @@
     "retryCount": 4,
     "bufferCapacity": 300,
     "bufferRefillProjectsPerTask": 5,
+    "defaultFreeCreditsLimit": 100.0,
     "defaultFreeCreditsDollarLimit": 100.0,
     "defaultFreeCreditsDaysLimit": 180,
     "garbageCollectionUserCapacity": 1000,
--- a/api/db/changelog/db.changelog-110-free-credits-days-override.xml
+++ b/api/db/changelog/db.changelog-110-free-credits-days-override.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="db.changelog-110-free-credits-days-override">
+    <renameColumn
+      newColumnName="free_tier_credits_limit_dollars_override"
+      oldColumnName="free_tier_credits_limit_override"
+      columnDataType="double"
+      tableName="user"/>
+    <addColumn tableName="user">
+      <column name="free_tier_credits_limit_days_override" type="tinyint">
+        <constraints nullable="true"/>
+      </column>
+    </addColumn>
+
+  </changeSet>
+</databaseChangeLog>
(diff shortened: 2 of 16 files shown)


api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -199,7 +199,7 @@ public Profile getProfile(DbUser user) {
     profile.setEmailVerificationStatus(user.getEmailVerificationStatusEnum());
 
     profile.setFreeTierUsage(freeTierBillingService.getUserCachedFreeTierUsage(user));
```
jmthibault79: Could also make this `setFreeTierDollarUsage()` to match.

api/config/config_perf.json
```diff
@@ -21,7 +21,8 @@
     "retryCount": 4,
     "bufferCapacity": 300,
     "bufferRefillProjectsPerTask": 5,
-    "defaultFreeCreditsLimit": 100.0,
```
gjuggler: Ooh – interesting to see a PR that removes an existing WorkbenchConfig field. I'm not aware of any clear operational guidelines on how to do this safely (it's arguably somewhat similar to removing a database field, w.r.t. rollout and rollback safety). @calbach I'm curious to hear if there is some existing rule we've been following here – in which case I'd love to find a good place to document it.

Taking a quick dip into the details: when we deploy the Workbench, our scripts push the config (https://github.com/all-of-us/workbench/blob/1ab8006207ebc5e01dfa3059b18bfd63863b4167/api/libproject/devstart.rb#L1692) before deploying the App Engine app (https://github.com/all-of-us/workbench/blob/1ab8006207ebc5e01dfa3059b18bfd63863b4167/api/libproject/devstart.rb#L1709). Depending on how long the config is cached in-memory by tasks (and/or whether any new tasks are kicked off in-between these steps), and how the config-parsing machinery works, I think we may risk some transient app failures during this time window.

But I can't speak with confidence one way or another. I haven't heard of this happening in the past ~year, but also our user base on prod has been very low up until November.

calbach: Hmm, yep I don't think we have any guidance on this yet. The usual pattern for flag removals is:

1. Send PR to eliminate all code references to the flag and document it as deprecated / pending removal
2. Wait for a release to go out (and wait beyond a reasonable doubt of needing to roll that back)
3. Send PR to remove the flag.

(1) is also a bit more nuanced than usual because our config update and code release do not happen atomically. So I think additionally, (1) should leave the prior config value in place s.t. it will still work in combination with the previous binary.

jmthibault79: 👍 will restore the old value and make a comment

api/db/changelog/db.changelog-110-free-credits-days-override.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="thibault" id="db.changelog-110-free-credits-days-override">
+    <renameColumn
+      newColumnName="free_tier_credits_limit_dollars_override"
```
calbach: Bonus: this now says the units which I was noticing before was a gap.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -73,8 +73,12 @@ public static WorkbenchConfig createEmptyConfig() {
     // The full table name for the BigQuery billing export, which is read from by the free-tier
     // usage tracking cron endpoint.
     public String exportBigQueryTable;
-    // The default limit to apply to free-credit usage in this environment.
-    public Double defaultFreeCreditsLimit;
+    // DEPRECATED.  Renamed to defaultFreeCreditsDollarLimit.
```
calbach: nit: as a convention would be nice to also include in the comment: "Remove after `https://github.com/all-of-us/workbench/pull/2920` reaches production."

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -63,6 +63,7 @@ public void fireAdministrativeBypassTime(
       Optional<Instant> bypassTime) {
     ActionAuditEvent.Builder eventBuilder =
         ActionAuditEvent.builder()
+            .timestamp(clock.millis())
```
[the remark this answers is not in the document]
jmthibault79: This is #2921 so it will go away after rebase

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
jmthibault79: sure

PR #2921, opened 2019-12-11
[risk=no]Fix self bypass

jmthibault79:
Description:

Alerting threw an exception due to missing fields.  The offender turned out to be the timestamp.

---
**PR checklist**


blrubenstein:
If we could follow up by adding a test that would have caught this, that would be awesome.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -63,6 +63,7 @@ public void fireAdministrativeBypassTime(
       Optional<Instant> bypassTime) {
     ActionAuditEvent.Builder eventBuilder =
         ActionAuditEvent.builder()
+            .timestamp(clock.millis())
```
jaycarlton: Hmm. This means I don't have unit test coverage here. I'll add a test for this function. 

I'm glad my validator did its job though.

PR #2922, opened 2019-12-11
[risk=no] scope and dao save fixes

freemabd:
A series of "improvements" and "fixes" to this test case started by @jaycarlton 

- Removing transient properties from DbParticipantCohortStatus(gender, race and ethnicity).
- Adding a new method getGenderRaceEthnicityMap() that returns a Map of all gender, race and ethnicity values used in the db to client object conversion.
- removed ComponentScan annotation and explicitly imported CohortReviewServiceImpl.class, CohortQueryBuilder.class and ReviewQueryBuilder.class
- Adding `@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)` to `DbUser` Bean
- Assigning reference of `DbParticipantCohortStatus` after calling dao `save()`
- Moving `verify(userRecentResourceService, atLeastOnce()).updateCohortEntry(anyLong(), anyLong(), anyLong())` into `assertParticipantCohortStatuses` method


api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -126,19 +125,20 @@
    * Converter function from backend representation (used with Hibernate) to client representation
    * (generated by Swagger).
    */
-  private static final Function<DbParticipantCohortStatus, ParticipantCohortStatus>
+  private static final BiFunction<
+          DbParticipantCohortStatus, Map<Long, String>, ParticipantCohortStatus>
```
jaycarlton: Do you mind just making this a vanilla function like `ParticipantCohortStatus toClientParticipant(participant, genderRaceEthMap)`? The hyper-functional stuff makes me think this is going to be more complicated than it needs to be. 

I can bring this up in Eng Sync.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
```diff
@@ -800,29 +801,10 @@ private DbCohortReview createNewCohortReview(
...
-  }
-
-  private String getName(List<DbCriteria> criteriaList, Long conceptId) {
-    List<DbCriteria> returnList =
-        criteriaList.stream()
-            .filter(c -> c.getConceptId().equals(conceptId.toString()))
-            .collect(Collectors.toList());
-    return returnList.isEmpty() ? null : returnList.get(0).getName();
+  /** Build a map that contains all gender/race/ethnicity names with the concept id as the key. */
+  private Map<Long, String> getGenderRaceEthnicityMap() {
```
jaycarlton: So could you still have a transient method to get the gender directly(that uses this map)? I.e. if you don't know or care about the conceptId?

api/src/main/java/org/pmiops/workbench/db/model/DbParticipantCohortStatus.java
```diff
@@ -192,25 +147,14 @@ public boolean equals(Object o) {
...
-    return Objects.hash(
-        status,
-        genderConceptId,
-        gender,
-        birthDate,
-        raceConceptId,
-        race,
-        ethnicityConceptId,
-        ethnicity);
+    return Objects.hash(status, genderConceptId, birthDate, raceConceptId, ethnicityConceptId);
```
jaycarlton: I wish we had a linter to find places where we leave out members in this method...

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
```diff
@@ -108,6 +115,7 @@
   private DbWorkspace workspace;
   private static final Instant NOW = Instant.now();
   private static final FakeClock CLOCK = new FakeClock(NOW, ZoneId.systemDefault());
+  private Map<Long, String> genderRaceEthnicityMap = new HashMap<>();
```
jaycarlton: you don't need to initialize it here if you're going to blow it away in setup.

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
```diff
@@ -169,7 +177,13 @@ public long getConceptId() {
   private static DbUser user;
 
   @TestConfiguration
-  @Import({CdrVersionService.class, CohortReviewController.class})
+  @Import({
+    CdrVersionService.class,
```
jaycarlton: do you have a macro for alphabetizing lines like this?

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
```diff
@@ -233,6 +248,10 @@ public void setUp() {
             .conceptId(String.valueOf(TestDemo.WHITE.conceptId))
             .name(TestDemo.WHITE.name));
 
+    genderRaceEthnicityMap =
```
jaycarlton: You can make a static `asMap` method inside your enum that returns this map.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
freemabd: I'm not a huge fan of the Function/BiFunction and can change these to regular methods. I can do a follow up PR to convert all of these Function/BiFunctions instead of just changing the one in this PR. Assuming that everyone approves in Eng Sync.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
freemabd: I suppose we could, but not sure how easy this would be to implement with a Hibernate Entity since i each entity is back by 1 table. This gender/race/eth map is backed by a different table/sql statement and is part of the cdr and can change per cdr version.

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
freemabd: Opted for you suggestion of static `asMap` method on enum.

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
freemabd: I don't.

api/src/test/java/org/pmiops/workbench/api/CohortReviewControllerTest.java
freemabd: Done.

api/src/main/java/org/pmiops/workbench/api/CohortReviewController.java
jaycarlton: Thanks. 

To be clear, there's definitely a role for those classes as function parameters and map values, but we don't need to declare local ones in most cases (unless you need lambda-style capturing, which isn't super common, but I certainly don't mean to proscribe.

PR #2925, opened 2019-12-13
[risk=no][RW-3932] Alert banner display

als364:
Screen Shot 2019-12-13 at 1 57 13 PM
I'm going to do the admin UI in a separate PR.
---
**PR checklist**


api/src/main/resources/workbench.yaml
```diff
@@ -3799,3 +3811,28 @@ definitions:
...
+          Title of the status alert, e.g. 'Service Incident: December 6, 2019'
+        type: string
+      message:
+        description: >
+          The actual message of the status alert, e.g. 'AoU RW is down because GCP is down'
+        type: string
+      link:
+        description: >
+          A URL linking to an incident report where users can read more
+        type: string
```
blrubenstein: Nit: Newline at end of file.

ui/src/app/components/status-alert-banner.tsx
```diff
@@ -0,0 +1,134 @@
...
+      message: '',
+      link: '',
+      isVisible: false
+    };
+  }
+
+  async componentDidMount() {
+    const statusAlerts = await statusAlertApi().getStatusAlert();
+    if (statusAlerts.length > 0) {
+      // There should be either one or zero alerts.
```
blrubenstein: The API suggests that there could be N status alerts. TOL: If we are expecting only one or zero alerts, should it be returning a list? (I'd expect a nullable object)

TOL: I think it is fine to have it return as a list, I'd just remove this comment? Because I could see us future proofing against multiple alerts. (I think Lou has comps of multiple with tabs in the tooltip for future work)

ui/src/app/components/status-alert-banner.tsx
```diff
@@ -0,0 +1,134 @@
...
+      const statusAlert = statusAlerts[0];
+      this.setState({
+        statusAlertId: statusAlert.statusAlertId,
+        title: statusAlert.title,
+        message: statusAlert.message,
+        link: statusAlert.link
+      });
+      // This needs to happen separately so that the other state is already set in isVisible()
+      this.setState({
```
blrubenstein: I believe setState is asynchronous. So we can't depend on the previous setState having resolved. https://css-tricks.com/understanding-react-setstate/ 

If we are depending on state existing, we should use the updater function. Alternatively, we could do the styling in a ternary on the banner object, and then it will pull whatever is right at render time, as opposed to state.

Basically removing this state variable, and changing the style from
`style={this.getStyles(isVisible)}` => `style={{...styles.alertBanner, height: this.isVisible() ? '200px' : '0px'}}`, and removing the `getStyles` function.

(Has the added benefit of removing state, which is always good)

ui/src/app/components/status-alert-banner.tsx
```diff
@@ -0,0 +1,134 @@
...
+      const cookie = localStorage.getItem(cookieKey);
+      return (cookie && cookie !== `${this.state.statusAlertId}`) || !!this.state.message;
+    } else {
+      return !!this.state.message;
+    }
+  }
+
+  render() {
+    const {title, message, link, isVisible} = this.state;
+    return isVisible && <FlexColumn style={this.getStyles(isVisible)}>
```
blrubenstein: See comment on setState logic.

ui/src/app/components/status-alert-banner.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (cookiesEnabled()) {
+      localStorage.setItem(cookieKey, `${this.state.statusAlertId}`);
+    }
+    this.setState({isVisible: false});
+  }
+
+  isVisible() {
+    if (cookiesEnabled()) {
+      const cookie = localStorage.getItem(cookieKey);
+      return (cookie && cookie !== `${this.state.statusAlertId}`) || !!this.state.message;
```
jmthibault79: If multiple alerts are possible we'd also have to store multiple cookies (or a list or something)

ui/src/app/components/status-alert-banner.tsx
als364: I'll change it to returning an `Option<Alert>`, I think I was thinking too pythonic or something.

api/src/main/java/org/pmiops/workbench/api/StatusAlertController.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public ResponseEntity<StatusAlertResponse> getStatusAlert() {
+    List<DbStatusAlert> dbStatusAlerts = statusAlertDao.findAll();
+    List<StatusAlert> statusAlerts =
+        dbStatusAlerts.stream()
+            .map(
+                dbStatusAlert ->
+                    new StatusAlert()
```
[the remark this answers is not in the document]
als364: `StatusAlert` is autogenerated so I'd have to make a utils class or another function in this class that does this work. Not worth the fragmentation IMO

api/src/main/java/org/pmiops/workbench/db/dao/StatusAlertDao.java
```diff
@@ -0,0 +1,11 @@
+package org.pmiops.workbench.db.dao;
+
+import java.util.List;
+import org.pmiops.workbench.db.model.DbStatusAlert;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface StatusAlertDao extends CrudRepository<DbStatusAlert, Long> {
+  List<DbStatusAlert> findAll();
```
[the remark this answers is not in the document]
als364: it only kind of does.

ui/src/app/components/status-alert-banner.tsx
als364: The requirement is to allow only one

ui/src/app/components/status-alert-banner.tsx
als364: Removing state is not inherently good in all cases. We shouldn't be forcing react components to rerender or change appearance, it should always be based off state and props. The Reactiest way to do this is probably binding a `hideStatusAlert` to the `NavBar` to pass as `onClick` to the `StatusAlertBanner`.

api/src/main/resources/workbench.yaml
```diff
@@ -3799,3 +3811,28 @@ definitions:
...
         format: int64
+
+  StatusAlertResponse:
+    type: array
+    items:
+      $ref: "#/definitions/StatusAlert"
+
+  StatusAlert:
+    type: object
+    properties:
```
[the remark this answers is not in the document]
als364: It doesn't look like that's a thing in Swagger. It looks like you can require parameters in requests but not in responses.

api/src/main/java/org/pmiops/workbench/api/StatusAlertController.java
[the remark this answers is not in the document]
jmthibault79: Ah, nope

api/src/main/resources/workbench.yaml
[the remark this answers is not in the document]
jmthibault79: huh ok

ui/src/app/components/status-alert-banner.tsx
jmthibault79: yep

PR #2926, opened 2019-12-13
[rw-4157][risk=no] Cron endpoint and service to collect metric views.

jaycarlton:
I define a new interface to allow any controller, service, dao, or bean to be a gauge metric collector. Then there's a GaugeRecorderService that grabs those events and records all at once. This is wired to a cron endpoint on a new controller, and is working locally via curl.

I've got a cron job configured in the yaml file, but I haven't seen it work yet. @calbach I'm trying to figure what I need to make that work.

I'm using `alias gauges='curl -X GET http://localhost:8081/v1/cron/monitoring/updateGaugeMetrics --header "Authorization: Bearer `gcloud auth print-access-token`" --header "X-AppEngine-Cron: true" | jq'`

Log output (temporary) is basically
`
Dec 13, 2019 3:42:36 PM org.pmiops.workbench.monitoring.GaugeRecorderService logValues
INFO: billing_project_buffer_entries = 3
debug_epoch_millis = 1576269755821
debug_random_double = 0.132487327907033
billing_project_buffer_available_project_count = 3
billing_project_buffer_assigning_project_count = 0
billing_project_buffer_creating_project_count = 0
cohort_count = 21
dataset_count = 15
workspace_total_count = 38`

**Things I learned**:
- I can inject a list of all beans implementing an interface very simply by using `List<MyInterface>`
- I can count all entries in a table by using `fooDao.count()`. This is good enough for getting started.
- While for event metrics, the class being instrumented gets a dependency on the MonitoringService (which bit me with `NotebooksServiceImpl` in the ToolConfig on the last PR), in the new gauge design I've inverted the dependency. Any class wishing to provide a gauge value simply exposes the new interface and is collected and relied upon to collect the metrics every minute.
- It takes about 1 Mississippi to collect all the gauge values locally. We can probably do better, but since we only collect once per minute, we're probably ok.

**Things confusing me**
- I haven't found an easy way to configure the logging level specifically for an individual package or Java class name, at least not at runtime. That's a huge advantage, and something we'll likely want or production. I know at one employer we were using log4j with this capability, but I'm not sure what else was required, or if that was even necessary.
- The awesomeness that let me pull in all the data collector beans at once is breaking down in the `CommandLineToolConfig`


calbach:
After you `dev-up` or `start-api-reqs` it should automatically pickup your cron configuration and start executing it.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -61,7 +65,7 @@
  * <p>This needs to implement an interface to support Transactional
  */
 @Service
-public class WorkspaceServiceImpl implements WorkspaceService {
+public class WorkspaceServiceImpl implements WorkspaceService, GaugeDataCollector {
```
jaycarlton: In writing up a test, I'm starting to think this approach is [asking for trouble](https://stackoverflow.com/questions/39498827/spring-service-implementing-multiple-interfaces-good-or-bad). Basically, while Spring is happy to inject the (same instance of the) service as either bean, it's tricky to mock them separately. One way to do it is to add qualifiers, which is probably the safest and sanest thing to do if we want to keep multiple interfaces on the same service.

Adding the `GaugeDataCollector` as a parent interface to `WorkspaceService` would work, but is a nasty hack and doesn't reflect the design intent we have here.

Another option would be to have a separate implementation of `GaugeDataCollector` for each gauge collector, and then distinguish those by qualifier. Pretty bloated, but it fits our patterns better. Those could then depend on the services (or data managers, if we had those) to hit the daos.

jaycarlton:
@calbach I'm still trying to get `dev-up` up. I still haven't seen the cron job fire on its own power. If you get a minute, could you please see if it works on your system?

api/src/main/java/org/pmiops/workbench/monitoring/views/MonitoringViews.java
```diff
@@ -28,7 +28,10 @@
       Aggregation.Sum.create()),
   NOTEBOOK_SAVE("notebook_save", "Save (or create) a notebook"),
   NOTEBOOK_CLONE("notebook_clone", "Clone (duplicate) a notebook"),
-  NOTEBOOK_DELETE("notebook_delete", "Delete a notebook");
+  NOTEBOOK_DELETE("notebook_delete", "Delete a notebook"),
+  WORKSPACE_TOTAL_COUNT("workspace_total_count", "Count of all workspaces (including inactive)"),
```
blrubenstein: Is the goal of the count metrics to tell if a DB has been dropped? Could we add CONCEPT_SET_COUNT to this then? (Just so we have a complete set) And maybe COHORT_REVIEW_COUNT?

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
```diff
@@ -668,6 +669,13 @@ public void testGetStatus() {
     assertThat(bufferStatus.getBufferSize()).isEqualTo(numberAvailable);
   }
 
+  @Test
+  public void testGetGaugeData() {
+    final Map<OpenCensusStatsViewInfo, Number> result = billingProjectBufferService.getGaugeData();
+    assertThat(result).hasSize(6);
```
blrubenstein: Opt: What is the number six here? Does it make sense to capture it in a variable so there's some documentation?

api/src/test/java/org/pmiops/workbench/monitoring/GaugeRecorderServiceTest.java
```diff
@@ -0,0 +1,102 @@
...
+  private static final Map<OpenCensusStatsViewInfo, Number> BILLING_BUFFER_GAUGE_MAP =
+      ImmutableMap.of(
+          MonitoringViews.BILLING_BUFFER_ASSIGNING_PROJECT_COUNT,
+          2L,
+          MonitoringViews.BILLING_BUFFER_AVAILABLE_PROJECT_COUNT,
+          1L,
+          MonitoringViews.BILLING_BUFFER_CREATING_PROJECT_COUNT,
+          0L);
+  private static final long WORKSPACES_COUNT = 101L;
+  private static final Map<OpenCensusStatsViewInfo, Number> WORKSPACE_GAUGE_MAP =
```
blrubenstein: TOL: If we're testing the WORKSPACE_GAUGE_MAP and the BILLING_BUFFER_GAUGE_MAP, does it make sense to test the other maps? Or is the goal here to just test a couple of gauges?

I think it's fine to just test a couple of gauges, so we know it works and don't add too much overhead to adding more?

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Create and register the stats configuration on the stats exporter. This operation should only
+   * happen once, so we have an isInitialized guard for that.
+   */
+  public void createAndRegister() {
+    if (!initialized) {
+      try {
+        final StackdriverStatsConfiguration configuration =
+            StackdriverStatsConfiguration.builder()
+                .setMetricNamePrefix(buildMetricNamePrefix())
```
blrubenstein: Should this be stackdriver specific now? Or leave generic?

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.inject.Provider;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.springframework.stereotype.Service;
+
+/**
+ * Simple wrapper service so we can mock out the initialization step (which depends directly on a
+ * public static method)
+ */
+@Service
+public class StackdriverStatsExporterService {
```
blrubenstein: TOL: Could we give this an interface, so that all ExporterServices implement that interface? None of the methods seem unique to it being Stackdriver.

jaycarlton: It specifically wraps the `StackdriverStatsExporter` class from OpenCensus and constructs a `StackdriverStatsConfiguration`. If anyone is going to make a parent interface to the exporters, OpenCensus would be the ones to do it. So we could add an interface, but we're as likely as not to have to break it when we do the next one. Still, I'll keep this in mind if I do a SignalFX exporter.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
jaycarlton: The `StackdriverStatsConfiguration` object is theirs, not mine. It looks like OpenCensus stuffs all backend-specific things in their exporters. I don't think it's a good candidate for generic-hood until we know all the exporters will be the same shape. (I don't want to make an interface for existing third-party classes to support when I don't control any of those classes). Good idea, though.

api/src/main/java/org/pmiops/workbench/monitoring/views/MonitoringViews.java
jaycarlton: These are just a smattering of gauges to showcase the system and its variety. I think they're fun to show off (for example in the Product Dashboard), as well as to cross-reference when something fishy happens (like the number of workspaces goes up 10x in a day).

I'd like for someone else to add the next few gauges and report back on how easy or difficult it was.

api/src/test/java/org/pmiops/workbench/monitoring/GaugeRecorderServiceTest.java
jaycarlton: The point of multiple maps is to test the map streaming/joining logic and demonstrate how the injection scheme works. It's not mean to be all-inclusive.

For a simple enum, you could declare a local test enum, like `TestGauges.GAUGE_A`, but I think that would be hairy in this case (since all the entries have to satisfy the parent interface).

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
jaycarlton: I'm pretty satisfied for now. I can handle any ambiguity by adding qualifiers if we care.

api/src/test/java/org/pmiops/workbench/billing/BillingProjectBufferServiceTest.java
jaycarlton: Yeah, this is a really fragile test. It's asserting that there are as many as there were the first time I ran the test.

I thought about having a system where each service or whatever could have its own enum for metrics, but that's hard to do cleanly when I want nontrivial metadata on each enum value. But if I had that, I'd make this be the size of the right one.

api/src/main/java/org/pmiops/workbench/monitoring/GaugeDataCollector.java
```diff
@@ -0,0 +1,8 @@
+package org.pmiops.workbench.monitoring;
+
+import java.util.Map;
+import org.pmiops.workbench.monitoring.views.OpenCensusStatsViewInfo;
+
+public interface GaugeDataCollector {
```
calbach: Please add a comment - defining a Gauge would be good.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -0,0 +1,56 @@
...
+import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.inject.Provider;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.springframework.stereotype.Service;
+
+/**
+ * Simple wrapper service so we can mock out the initialization step (which depends directly on a
```
calbach: This description feels a bit too far zoomed in for me to understand. What is this wrapping and what initialization step are you referring to and why is it necessary to do that statically?

api/src/main/java/org/pmiops/workbench/monitoring/GaugeRecorderService.java
```diff
@@ -0,0 +1,44 @@
...
+    this.gaugeDataCollectors = gaugeDataCollectors;
+    this.monitoringService = monitoringService;
+  }
+
+  public Map<OpenCensusStatsViewInfo, Number> record() {
+    ImmutableMap<OpenCensusStatsViewInfo, Number> result =
+        gaugeDataCollectors.stream()
+            .map(GaugeDataCollector::getGaugeData)
+            .flatMap(m -> m.entrySet().stream())
+            .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
```
calbach: I'm not sure what happens with the immutable map collector if you have duplicate keys.

Either way, I would document what the expected behavior here is (probably on GaugeDataCollector) and probably add a test for it. I think ideally it should be one of the following:
- No two GaugeDataCollectors may export the same data key, or
- Values are grouped by key and summed across all GaugeDataCollectors

api/src/main/java/org/pmiops/workbench/cohortreview/CohortReviewServiceImpl.java
```diff
@@ -26,13 +28,16 @@
...
+import org.pmiops.workbench.monitoring.views.MonitoringViews;
+import org.pmiops.workbench.monitoring.views.OpenCensusStatsViewInfo;
 import org.pmiops.workbench.workspaces.WorkspaceService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 @Service
-public class CohortReviewServiceImpl implements CohortReviewService {
+public class CohortReviewServiceImpl implements CohortReviewService, GaugeDataCollector {
```
jaycarlton: @freemabd FYI I sampled a data point from this class to get the total number of cohorts in the system. Please let me know if there's a better way to get this information periodically or if there are other metrics you'd like to see sampled.

freemabd: This service is for things related to CohortReview objects which are part of the review section of the workbench. Users can create a cohort and never use the review tool, so it doesn't seem logical to track cohort counts here. CohortReview is a subset of a Cohort. This would be nice to track, but i would use a different property for this.  `MonitoringViews.COHORT_REVIEW_COUNT`? I believe that COHORT_COUNT is a good metric to sample, but it should be sampled in CohortsController.... not sure it has a CohortsService to follow this pattern?

PR #2928, opened 2019-12-14
[RW-4174][RISK=NO] Unique ID for 500 errors to associate logs to browser

jaycarlton:
For security and privacy reasons, we strip most of hte information from server errors before returning a status to the front end. This PR adds an attribute to the errorResponse object so that it shows up in the JS Network tab. A future UI change could make this visible in the error messaging or main JS console too. For example, it could say "An internal server error has occurred. Please use the unique event code $GUID if you wish to report it to support."

We don't need every user to do this, as we'll have log-based metrics on the backend.

The appearance of the error in curl is

`
jaycarlton~/repos/workbench/api (jaycarlton/RW-4174) $ curl -X GET http://localhost:8081/v1/status -H Content-Type: application/json | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   127  100   127    0     0    414      0 --:--:-- --:--:-- --:--:--   415
{
  "message": null,
  "statusCode": 500,
  "errorClassName": null,
  "errorCode": null,
  "errorUniqueId": "1e65e766-3aa3-449e-8626-131e8a67209d"
}
`

In the Server log, it looks like
`
SEVERE: ErrorId 1e65e766-3aa3-449e-8626-131e8a67209d: java.lang.IllegalArgumentException
java.lang.IllegalArgumentException: Who dat?
        at org.pmiops.workbench.api.StatusController.getStatus(StatusController.java:25)
        at org.pmiops.workbench.api.StatusApiController.getStatus(StatusApiController.java:35)
`

In the Chrome Devtools Networking tab it looks like this:
Screen Shot 2019-12-14 at 11 14 47 AM

It's a few steps to get a user navigated to the network tab, but hopefully the UI can use this attribute directly.


jaycarlton:
@blrubenstein this could be useful in your global UI error handler 🌍 ✋.

api/src/main/java/org/pmiops/workbench/exceptions/ExceptionAdvice.java
```diff
@@ -25,19 +25,21 @@
 
   @ExceptionHandler({Exception.class})
   public ResponseEntity<?> serverError(Exception e) {
-    ErrorResponse errorResponse = new ErrorResponse();
-    Integer statusCode = HttpStatus.INTERNAL_SERVER_ERROR.value();
-
+    ErrorResponse errorResponse = WorkbenchException.errorResponse();
```
freemabd: Why a static method in WorkbenchException, instead of allowing ErrorResponse to generate the random UUID? Would we like this functionality outside of things that extend WorkbenchException?

jaycarlton: `ErrorResponse` is swagger-generated, so no touchies.

freemabd: Gotcha... and noticed the Who dat?... Geaux Saints!

PR #2929, opened 2019-12-16
[rw-4182][RISK=NO] Replace email field with username in DB entity and deprecate on model

jaycarlton:
This is a simple rename of `DbUser.getEmail` and `DbUser.setEmail`, and associated dao methods. For the generated model, I added a `userName` field, but did not yet remove the `email` field. In returned models, both should be populated. I didn't find a case where we took action based on a value a client passed in for that field (not sure if any APIs even use this model for input).

Changing to `getUsername()` based on [OED spelling](https://english.stackexchange.com/questions/43436/username-user-name-or-user-name)


api/src/main/java/org/pmiops/workbench/api/CohortsController.java
```diff
@@ -74,7 +74,7 @@ public Cohort apply(DbCohort cohort) {
                   .name(cohort.getName())
                   .type(cohort.getType());
           if (cohort.getCreator() != null) {
-            result.setCreator(cohort.getCreator().getEmail());
+            result.setCreator(cohort.getCreator().getUsername());
```
jaycarlton: @freemabd FYI the `email` entity field is being replaced but we're keeping the table schema as is. So it's just a replacement for that column. Take care with manual sql queries. I found one place I had to fix it by hand.

freemabd: Thanks for info.

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
```diff
@@ -44,7 +44,7 @@
   @Override
   @AuthorityRequired({Authority.ACCESS_CONTROL_ADMIN})
   public ResponseEntity<Void> updateUserDisabledStatus(UpdateUserDisabledRequest request) {
```
NehaBroad: It would be nice to have a follow up PR to update the field in UpdateUserDisabledRequest as well to username to be consistent

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -284,74 +284,74 @@ public DbUser createUser(
...
     // For existing user that do not have address
     if (address != null) {
-      address.setUser(user);
+      address.setUser(dbUser);
     }
-    if (demographicSurvey != null) demographicSurvey.setUser(user);
+    if (demographicSurvey != null) demographicSurvey.setUser(dbUser);
     if (institutionalAffiliations != null) {
-      final DbUser u = user;
+      final DbUser u = dbUser;
```
NehaBroad: I know this is not part of your change however can we change the variable to a better name than u

api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -100,7 +100,7 @@ DbUser getDbUser() {
   @Before
   public void setUp() {
     DbUser user = new DbUser();
-    user.setEmail(EMAIL_ADDRESS);
+    user.setUsername(EMAIL_ADDRESS);
```
NehaBroad: The constant should be rename  to USERNAME for consistency

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -292,7 +292,7 @@ public void updateRecentWorkspaces_multipleUsers() {
 
     DbUser mockUser = mock(DbUser.class);
     doReturn(mockUser).when(mockUserProvider).get();
-    doReturn(DEFAULT_USER_EMAIL).when(mockUser).getEmail();
+    doReturn(DEFAULT_USER_EMAIL).when(mockUser).getUsername();
```
NehaBroad: Same for this, the constant should be renamed to DEFAULT_USERNAME

api/src/main/java/org/pmiops/workbench/api/AuthDomainController.java
jaycarlton: Good catch. Let's do that with the UI changes.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: I think I stole `dbUser` out of this internal scope. I only use single letter variables if they live 2-3 lines or less (and usually only for a lambda).

But that's not the problem here. I don't understand the line `final DbUser u = dbUser;`. I think the original author expected this to allow them to work with a copy of `dbUser`, but objects don't work that way in Java. I'll cut out `u` entierly.

jaycarlton: The situation is a smell in that we want to capture a variable and later reassign it, which Java doesn't support. So we need a dummy name. I picked a longer one than `u` for now, though for short lambdas I typically am OK with very short variable names.

PR #2934, opened 2019-12-17
[RW-4201][RISK=NO] Support for Attachments.

jaycarlton:
Time series data in most (all?) monitoring systems can have metadata attached to it to allow splitting and grouping graphs, reports, alerts, etc, by these dimensions. Common values used are Region, Environment, server ID, etc.

In OpenCensus, this is done with an AttachmentMap associated with a MeasureMap. All view/value pairs in in the MeasurementMap (aka measurements of the registered View(s)) share the same attachment map.

This necessitated a bit of rethinking my current interface, but the migration wasn't all that bad.

As usually, I renamed a thing or two. I'm still learning what the words mean and how I'm expected to be thinking about them.

It's debatable whether I should've supported this in the first cut, but that first PR was so dense I think I made the right call.

**Things I learned**
1. I can create a test bean that returns a lambda implementing an interface.
2. I can use name/qualifier pairs within a test application context. I thought this would work, but it's neat to see.
3. I can also (apparently) create a mock and return it from a test bean, but this is a relatively bad idea compared to approach `#1.
4. Default methods on Dao interfaces let you do useful things with returned results (like make a map), although you'd think there's something builtin in Spring/JPA for that.
5. It can be handy to have two different enums implementing a common interface, for example when default values for some fields and invariants are different.
6. The Global MonitoredResourceType isn't always/ever what you want.


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -57,7 +64,7 @@
  * ensuring we call a single updateDataAccessLevel method whenever a User entry is saved.
  */
 @Service
-public class UserService {
+public class UserService implements GaugeDataCollector {
```
jaycarlton: This usage of the new interface is causing Spring issues.
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'authDomainController' defined in file [/Users/jaycarlton/repos/workbench/api/build/exploded-api/WEB-INF/classes/org/pmiops/workbench/api/AuthDomainController.class]: Unsatisfied dependency expressed through constructor parameter 1; nested exception is org.springframework.beans.factory.BeanNotOfRequiredTypeException: Bean named 'userService' is expected to be of type 'org.pmiops.workbench.db.dao.UserService' but was actually of type 'com.sun.proxy.$Proxy134'
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:749)
```

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -38,6 +38,14 @@ public void createAndRegister() {
             StackdriverStatsConfiguration.builder()
                 .setMetricNamePrefix(buildMetricNamePrefix())
                 .setProjectId(workbenchConfigProvider.get().server.projectId)
+                .setMonitoredResource(
+                    MonitoredResource.newBuilder()
```
jaycarlton: These are placeholder values. I thought i could implement all of https://precisionmedicineinitiative.atlassian.net/browse/RW-4226 in this PR, but looks like there's going to be a new dependency on an AppEngine service.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
```diff
@@ -0,0 +1,131 @@
...
+      Map<AttachmentKey, String> attachmentKeyStringMap) {
+    return attachmentKeyStringMap.entrySet().stream()
+        .map(
+            e ->
+                new SimpleImmutableEntry<String, AttachmentValue>(
+                    e.getKey().getKeyName(), AttachmentValueString.create(e.getValue())))
+        .collect(ImmutableMap.toImmutableMap(Entry::getKey, SimpleImmutableEntry::getValue));
+  }
+
+  public static Builder builder() {
```
jaycarlton: If we had Immutables, we'd basically be done with this file now. Probably just define a validate() function for the builder to use.

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringService.java
```diff
@@ -1,20 +1,51 @@
...
-   * Record an occurrence of a counted event.
+   * Record an occurrence of a counted (a.k.a. delta or cumulative) time series. These are events
+   * that are typically measured more for frequency than for absolute value.
    *
-   * @param viewInfo
+   * @param eventMetric
    */
-  default void recordIncrement(OpenCensusStatsViewInfo viewInfo) {
-    recordValue(viewInfo, 1);
+  default void recordEvent(EventMetric eventMetric) {
```
jaycarlton: I almost named all of these `record` and let the overloads do their thing, but I thought it would be clearer to be opinionated. E.g. you should never record a gauge metric without a value, or an event metric with a value.

api/src/main/java/org/pmiops/workbench/monitoring/views/EventMetric.java
```diff
@@ -0,0 +1,58 @@
...
+    return MeasureLong.class;
+  }
+
+  @Override
+  public Aggregation getAggregation() {
+    return Aggregation.Count.create();
+  }
+
+  @Override
+  public List<TagKey> getColumns() {
```
jaycarlton: I'm still trying to understand why I need these `TagKey`s, and if they have anything to do with attachments.

api/src/main/java/org/pmiops/workbench/utils/DaoUtils.java
```diff
@@ -0,0 +1,27 @@
...
+   * Helper for mapping queries. For example, to count all workspaces by active status.
+   *
+   * @param queryResults - Generally dao.findAll()
+   * @param indexer - method to return a value. Frequently a method reference like
+   *     DbUser::getDisabled
+   * @param <ENTITY_T> - Type of entity
+   * @param <ATTRIBUTE_T> - Type of attribute
+   * @return
+   */
+  public static <ENTITY_T, ATTRIBUTE_T> Map<ATTRIBUTE_T, Long> getAttributeToCountMap(
```
jaycarlton: This generic parameter naming is controversial. I think traditionally this would just be `<T, U>`, but I thought this was was less scary for a reader.

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringServiceImpl.java
```diff
@@ -42,13 +43,13 @@ private void initStatsConfigurationIdempotent() {
...
-    Arrays.stream(MonitoringViews.values())
-        .map(MonitoringViews::toOpenCensusView)
+    Arrays.stream(GaugeMetric.values())
+        .map(OpenCensusView::toView)
         .forEach(viewManager::registerView);
   }
 
   @Override
-  public void recordValue(OpenCensusStatsViewInfo viewInfo, Number value) {
+  public void recordValue(OpenCensusView viewInfo, Number value) {
```
gjuggler: I'm a confused by the method override here. Why is a default method implementation provided in the interface if the only concrete class overrides it? I suspect this was lost in refactoring and should be removed.

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringServiceImpl.java
```diff
@@ -73,22 +74,31 @@ public void recordValue(OpenCensusStatsViewInfo viewInfo, Number value) {
   }
 
   /**
-   * Record multiple values at once, indexed by keys in the StatsViewProperties enum. This should
-   * save calls to Stackdriver.
+   * Record multiple values at once. An attachment map allows associating these measurements with
```
gjuggler: nit: bump the method description up to the interface level, for consistency with the nice docs added for recordEvent.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/AttachmentKey.java
```diff
@@ -0,0 +1,23 @@
+package org.pmiops.workbench.monitoring.attachments;
+
+public enum AttachmentKey implements AttachmentKeyBase {
```
gjuggler: Why did you separate out the AttachmentKeyBase interface? This pattern isn't used elsewhere in our codebase AFAICT, and it's non-obvious to me why a trivial interface was pulled out. Worth a comment somewhere at least.

api/src/main/java/org/pmiops/workbench/monitoring/views/GaugeMetric.java
```diff
@@ -19,19 +21,25 @@
...
-  NOTEBOOK_CLONE("notebook_clone", "Clone (duplicate) a notebook"),
-  NOTEBOOK_DELETE("notebook_delete", "Delete a notebook"),
   WORKSPACE_TOTAL_COUNT("workspace_total_count", "Count of all workspaces (including inactive)"),
+  WORKSPACE_COUNT_BY_ACTIVE_STATUS(
+      "workspace_count_by_active_status", "Count of workspaces by active status"),
+  WORKSPACE_COUNT_BY_DATA_ACCESS_LEVEL(
+      "workspace_count_by_data_access_level", "Count of workspaces for each data access level"),
   DATASET_COUNT("dataset_count", "Count of all datasets in existence"),
-  COHORT_COUNT("cohort_count", "Count of all cohorts in existence");
+  DATASET_COUNT_BY_INVALID("dataset_count_by_invalid", "Count of all datasets by invalid status"),
```
gjuggler: [nit] Can you keep this list of enums alphabetized? Otherwise it's going to get pretty intractable.

api/src/main/java/org/pmiops/workbench/monitoring/views/OpenCensusView.java
```diff
@@ -63,9 +63,10 @@ default MeasureDouble getMeasureDouble() {
 
   Aggregation getAggregation();
 
+  // TODO(jaycarlton) What is this for?
   List<TagKey> getColumns();
 
-  default View toOpenCensusView() {
+  default View toView() {
```
gjuggler: nitpick: given the conversions going on, I think the prior method name is worth the extra chars.

api/src/main/java/org/pmiops/workbench/monitoring/views/OpenCensusView.java
```diff
@@ -22,12 +22,12 @@
  * we may wish to support other metrics backends, and don't want to depend on Stackdriver concepts
  * or implementation details.
  */
-public interface OpenCensusStatsViewInfo {
+public interface OpenCensusView {
```
gjuggler: Naming nitpick: I'd slightly lean towards this being name more in alignment with the classes that implement the interface, e.g. MetricBase or Metric. Now that we're using some non-opencensus terminology for classes in this package (which I think is a good thing – GaugeMetric and EventMetric are exceedingly clear), it may be best to align this interface with the concrete classes.

(Related note, it may be worth updating the javadoc comment re: naming, since it seems slightly out of date now.)

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
```diff
@@ -158,16 +160,21 @@ public FileDetail copyNotebook(
...
             fromNotebookName,
             workspaceNamespace,
             workspaceName,
             newName);
-    monitoringService.recordIncrement(MonitoringViews.NOTEBOOK_CLONE);
-    return result;
+    monitoringService.recordBundle(
+        MeasurementBundle.builder()
+            .addValue(EventMetric.NOTEBOOK_CLONE, 1L)
+            .attach(AttachmentKey.NOTEBOOK_NAME, copiedNotebookFileDetail.getName())
```
gjuggler: [side question] I could be wrong, but isn't it best practice not to set attachments / facets on fields with an unbounded space of possible values?

There's also a potential concern here around user data privacy, where the file name of a researcher's notebook may include details about the nature of their research. Permanently attaching this to metrics in Stackdriver, which is otherwise generally used to track operational metrics without an expectation of housing private data, may be a slight risk.

(Musings from past experience: Google is probably exceptionally conservative in this regard, but my past team was required to treat *cloud project names* as PII, subject to extreme auditing & access requirements and wipeout compliance. We're likely a solid step less strict than that, but it's a constant thought in my head.)

gjuggler: Quick amendment: based on some unrelated digging into Stackdriver metrics, I'm reasonably convinced that my internalized guidance of "don't use unbounded values for attachment fields" is outdated. They seem pretty happy to store arbitrary strings here; presumably there's some fairly high limit re: the number of independent strings that they like to track during any one time window, but maybe that's not likely something we'll run up against.

gjuggler: Further amendment to the privacy implications: due to similarly-unrelated Stackdriver digging, I think I'm seeing that metrics tend to be kept for only 6 weeks by default, unless you actively do something to permanently store them. This makes me less concerned on the privacy front, though this is the kind of thing where I'd want to be sure we have an anticipated use case in mind before we just start tracking it because we can.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/AttachmentKey.java
```diff
@@ -0,0 +1,23 @@
...
+  BUFFER_ENTRY_STATUS("buffer_entry_status"),
+  DEBUG_COLOR("debug_color"),
+  NOTEBOOK_NAME("notebook_name"),
+  NOTEBOOK_WORKSPACE_NAMESPACE("notebook_workspace_namespace"),
+  USER_DISABLED("user_disabled"),
+  DATASET_INVALID("dataset_invalid"),
+  WORKSPACE_ACTIVE_STATUS("workspace_active_status"),
+  WORKSPACE_DATA_ACCESS_LEVEL("workspace_data_access_level");
```
ericsong: Not sure how we can make it happen but I think it would be ideal to have a way to associate the metadata keys and the metrics that they are describing. Through code references would be the most ideal, code locality can also work.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Simple bundle class to avoid dealing with lists of pairs of maps. This situation can occur in
+ * classes that have multiple kinds of signals with different attribute maps among subsets. In order
+ * to avoid excessive proliferation of (confusing) argument list types, we try to consolidate here.
+ */
+public class MeasurementBundle {
+  private final Map<OpenCensusView, Number> monitoringViews;
+  private final Map<AttachmentKey, String> attachmentKeyToString;
+
+  private MeasurementBundle(
```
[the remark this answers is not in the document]
ericsong: OK. I think I get it now. It looks like the attachmentKey's get sent with every OpenCensusView?

I think part of what was confusing me is that the names are not ones that I am used to when working with metrics.

My current understanding is that `OpenCensusView` is basically a `Measurement` and `AttachmentKey` is `Metadata` when looking at it with terms that I am familiar with.

Does that align with how you're thinking about these classes?

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
jaycarlton: It's really extensible. For instnace with a `check` method: https://immutables.github.io/immutable.html#precondition-check-method

It's really not an either-or situation, and as long as we have any Java it still makes sense to do this. Yes it's way too heavy to bring in Kotlin if this is all you wanted.

api/src/main/java/org/pmiops/workbench/monitoring/views/EventMetric.java
jaycarlton: These are used for [filter and group metrics](https://opencensus.io/tag/key/#2), of which we currently have none. They're for aggregation rather than attachments (and don't match attachment keys).

For this interface, I wanted to span all the inputs for the `View` create method. I'm thinking a bit ahead about pulling this all out into stand-alone library for doing Spring with OpenCensus, and somebody might need it.

I'd like to leave it here for now since it costs us nothing and allows us to use group metrics more easily when we're ready. I could try to make a demo metric using a group if that helps.

api/src/main/java/org/pmiops/workbench/utils/DaoUtils.java
jaycarlton: The generic problem is orthogonal to the perf problem.

I'd love to do the right query directly, but I couldn't find a good way to do it in our setup (I.e. to return a map from a repository method). Not sure where I found this trick, but I thought it was OK for small tables for now, as we only do this once a minute (and they can hopefully be cached).

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -62,8 +64,18 @@
   private static final int DATA_SET_VERSION = 1;
 
   @Override
-  public Map<OpenCensusStatsViewInfo, Number> getGaugeData() {
-    return ImmutableMap.of(MonitoringViews.DATASET_COUNT, dataSetDao.count());
+  public Collection<MeasurementBundle> getGaugeData() {
+    Map<Boolean, Long> invalidToCount = dataSetDao.getInvalidToCountMap();
+    return ImmutableSet.of(
+        MeasurementBundle.builder().addValue(GaugeMetric.DATASET_COUNT, dataSetDao.count()).build(),
```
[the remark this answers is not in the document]
jaycarlton: The main reason is that the ones without the attachment are already established and I didn't want to break anyone's dashboards until we prove the new ones. I plan to clean things up soon, but in general, we don't want to change the semantics of a metric; we should just waste the name and start with a new one.

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -38,4 +40,8 @@
       @Param("dals") List<Short> dataAccessLevels, @Param("term") String term, Sort sort);
 
   Set<DbUser> findByFirstRegistrationCompletionTimeNotNull();
+
+  default Map<Boolean, Long> getDisabledToCountMap() {
+    return DaoUtils.getAttributeToCountMap(findAll(), DbUser::getDisabled);
```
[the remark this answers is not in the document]
jaycarlton: We can do that. That does sound more interesting.

Thinking out loud, if I want to put multiple dimensions attachments on an object, that's definitely doable, but now I'm in the business of scanning through all of them and making a different bundle for each tuple in the cartesian product of the labels. Since we're talking about single digit thousands of users, I think we can do this fast enough.

api/src/main/java/org/pmiops/workbench/monitoring/GaugeDataCollector.java
```diff
@@ -1,8 +1,27 @@
 package org.pmiops.workbench.monitoring;
 
-import java.util.Map;
-import org.pmiops.workbench.monitoring.views.OpenCensusStatsViewInfo;
+import java.util.Collection;
 
+/**
+ * Each class wishing to supply gauge data to the gauge cron job can simply implement this
+ * interface. By returning a well-formed MeasurementBundle, including any appropriate attachments,
+ * in getGaugeData(), the class is ensured that these values are polled as appropriate.
```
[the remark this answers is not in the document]
jaycarlton: Kidding aside, it wouldn't be hard to allow for a class of gauges that ware sampled less frequently than once a minute. I suppose we could have a `pollExpensiveGauges()` method, and corresponding one on `GaugeDataCollector` if I only wanted to count the left-handed people with missing right pinky toes from Arkansas once a week.

api/src/main/java/org/pmiops/workbench/monitoring/views/GaugeMetric.java
jaycarlton: I've been thinking about this a bunch. Java doesn't let me split an enum into multiple files, and it also doesn't allow enums to extend abstract classes or other enums (or any classes). Any of those would give me an easy way to break things down.

As it is, having more than one GaugeMetric enum would involve some duplication of code between the files. For example, if the enum is going to have constructor calls, the constructor needs to be defined in that enum class.

But more to the point, if I have aspirations to make this Swagger/Spring adapter a separate library, I need a way to pull out our aou-specific metrics into an input file, like JSON or XML or the other one I can't think of that ain't some other thing. I'll be bummed to lose the strongly-typed metric identifiers, though.

Or a .properties file or something. Basically some way to make the metric definitions part of the configuration instead of the code and safely use them at runtime.

api/src/main/java/org/pmiops/workbench/monitoring/views/OpenCensusView.java
jaycarlton: No one will ever need to call this directly, so it's just a question of clarity. The real problem here is file name. I may rename it `Metric`.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
jaycarlton: All very good points. 

I really just needed an event metric with some kind of attribute. It definitely makes sense to only do this when you really want it.

I should definitely think about our retention policy. I was hoping it would save older data at a lower resolution.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
```diff
@@ -0,0 +1,131 @@
...
+  // enum directly and String values. For some reason we have to wrap strings
+  // in AttachmentValueStrings now (though naked Strings are still supported but
+  // deprecated).
+  private static Map<String, AttachmentValue> fromAttachmentKeyToString(
+      Map<AttachmentKey, String> attachmentKeyStringMap) {
+    return attachmentKeyStringMap.entrySet().stream()
+        .map(
+            e ->
+                new SimpleImmutableEntry<String, AttachmentValue>(
+                    e.getKey().getKeyName(), AttachmentValueString.create(e.getValue())))
```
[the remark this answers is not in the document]
jaycarlton: `e` is for `Map.Entry<AttachmentKey, String>`, so I have to break down the key and build up the value. It's kind of an awkward flip, but I think it's easiest on the collection side to use a regular string as the value.

One thing that's misleading here is that `AttachmentKey` is my name for an enum, and `AttachmentValue` is an OpenCensus concept. It feels a little dirty but makes sense.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
```diff
@@ -0,0 +1,131 @@
...
+      measurementsBuilder.put(viewInfo, value);
+      return this;
+    }
+
+    public Builder addAll(Map<OpenCensusView, Number> map) {
+      measurementsBuilder.putAll(ImmutableMap.copyOf(map));
+      return this;
+    }
+
+    public Builder attach(AttachmentKey attachmentKey, String value) {
```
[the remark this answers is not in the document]
jaycarlton: Yeah, it's non-standard enough to be less clear than I'd hoped.

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringService.java
```diff
@@ -1,20 +1,51 @@
...
+import java.util.Collections;
 import java.util.Map;
-import org.pmiops.workbench.monitoring.views.OpenCensusStatsViewInfo;
+import org.pmiops.workbench.monitoring.views.EventMetric;
+import org.pmiops.workbench.monitoring.views.OpenCensusView;
 
 public interface MonitoringService {
 
+  int DELTA_VALUE = 1;
+  Map<String, AttachmentValue> DEFAULT_ATTACHMENTS = Collections.emptyMap();
```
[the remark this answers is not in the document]
jaycarlton: The key reason for doing it this way is to remove the Stackdriver dependence from this contract.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/AttachmentKey.java
jaycarlton: Interesting. I don't know if anonymous enums are a thing, but if they were, then I could declare them scoped to each curve.

The real problem is that they're many:many w/r/t metrics. Say I had some machines running Linux and others running Solaris or some randomness. Lots of my gauge metrics might want to indicate the OS I was running, and we'd use the same string key for that. OpenCensus doesn't enforce that, but it's part of our model. The attachments refer to the whole MeasureMap (i.e. the MeasurementBundle), so it's key to keep things consistent there.

Some of these I don't want to share (like `user_disabled`), which I think is what you're getting at. We could have a `Set<Metric>` associated with each AttachmentKey if we wanted (or vice-versa). I don't know the rules on enums referring to other enums in their constructors, but I remember from C++ Thou Shalt Not rely on instantiation ordering of static objects.

api/src/main/java/org/pmiops/workbench/monitoring/views/OpenCensusView.java
jaycarlton: you read my mind.

api/src/main/java/org/pmiops/workbench/monitoring/views/GaugeMetric.java
jaycarlton: Alphabetizing for now.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksServiceImpl.java
jaycarlton: Dropped the notebook name & workspace properties for now, since they were just placeholders.

I don't yet have a continuous-valued attachment use case, but it should work.

api/src/main/java/org/pmiops/workbench/utils/DaoUtils.java
jaycarlton: Spoke with @gjuggler and @freemabd. It sounds like we need to use a `JdbcTemplate` connection or otherwise break out of our existing pattern in order to do the queries necessary. Which seems odd to me: all I need to do is select a bunch of columns and a count column, then bind the results to some kind of value class.

Proposal for now is to check this in and monitor performance of the database on Test and Prod.  I'm also open to adding feature flags, but that seems like overkill. We could simply disable the cron job if it got noticeably bad.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/AttachmentKey.java
jaycarlton: New version supports allowed Metrics and Discrete Values, enforced by the MeasurmentBundle builder.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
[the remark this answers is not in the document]
jaycarlton: Almost. If I had a dollar for every time I renamed one of these classes, I'd pay someone to explain it to me. I'm definitely going to need to write some docs. I still have an aspiration to support more than just Stackdriver, and going through that exercise should help sort the names again.

In OpenCensus, a `MeasureMap` is the thing that gets recorded at a particular time. It contains a map of `View`s (That is, a time series, or what Stackdriver calls a Metric) to numerical values (i.e. the measurements), as well as an optional map of attachment key strings to `AttachmentValue`s. It's fair to call that metadata, but I think introducing the term attachment gave some more precision around the specific measurement's metadata, distinct from the View's metadata or other context.

Note that all measurements in a MeasureMap (and conveniently this is true for a MeasurementBundle as well) are annotated by the same attachment map, if present.
 When you call `record()`, the active exporter will cache all the values until the configured reporting interval is up. For Stackdriver, that's always a minimum of 1 minute, for both gauges and events.

In order to extrapolate across reporting intervals or gaps, the aggregation type is used. For gauges, it's valid to use the last value, or possibly the mean. For events, you're typically counting things.

api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java
```diff
@@ -0,0 +1,131 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import io.opencensus.metrics.data.AttachmentValue;
+import io.opencensus.metrics.data.AttachmentValue.AttachmentValueString;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.pmiops.workbench.monitoring.attachments.AttachmentKey;
```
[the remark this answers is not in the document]
jaycarlton: `AttachmentKey` is now Attachment, which kicks the can down the road a bit. As discussed below, there's probably another round of violent renaming that needs to happen to clean all this up.

api/src/main/java/org/pmiops/workbench/monitoring/GaugeDataCollector.java
```diff
@@ -1,8 +1,27 @@
...
 public interface GaugeDataCollector {
-  Map<OpenCensusStatsViewInfo, Number> getGaugeData();
+
+  /**
+   * Return a collection of @link {MeasurementBundle} objects. Each bundle can have one or more
+   * View/Value pairs and zero or more metadata AttachmentKey/String pairs. There a 1:1
+   * correspondents between MeasurementBundles returned and instantiation of MeasureMaps (and calls
+   * to record).
+   *
+   * <p>If a class has no attachments to provide, then it's fine to include all samples in a single
```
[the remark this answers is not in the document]
jaycarlton: good catch. Do we have a way to render our Javadoc currently?

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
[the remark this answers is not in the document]
jaycarlton: I have a prototype of a multidimensional categorization of users.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/Attachment.java
```diff
@@ -0,0 +1,63 @@
...
+  USER_DISABLED(
+      "user_disabled", Collections.singleton(GaugeMetric.USER_COUNT), Booleans.getValueStrings()),
+  DATA_ACCESS_LEVEL(
+      "user_data_access_level",
+      ImmutableSet.of(GaugeMetric.USER_COUNT, GaugeMetric.WORKSPACE_COUNT),
+      Enums.getValueStrings(DataAccessLevel.class)),
+  WORKSPACE_ACTIVE_STATUS("workspace_active_status");
+
+  private String keyName;
+  private Set<Metric> allowedMetrics;
```
jaycarlton: After thinking about this some more, I don't like having the attachments know about metrics. It seems the list should be the other way around, with Metrics having a list of allowed Attachments, and Attachments only knowing about valid values.

The setup here isn't technically circular, but I don't want both enums to be "smart". So smart metrics with dumb attachments make more sense (particularly when most MeasurementBundles only have a single Metric).

api/src/main/java/org/pmiops/workbench/monitoring/GaugeDataCollector.java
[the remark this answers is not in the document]
gjuggler: I don't think so. I also realized in my own PRs that spotlessApply auto-creates these not-closed paragraph tags, so I don't think it will be super easy to change.

api/src/main/java/org/pmiops/workbench/monitoring/attachments/AttachmentBase.java
```diff
@@ -0,0 +1,46 @@
+package org.pmiops.workbench.monitoring.attachments;
+
+import java.util.Set;
+import org.pmiops.workbench.monitoring.views.Metric;
+
+/**
+ * Top-level interface for dealing with attachments. Currently the only supported feature is
```
gjuggler: Nitpick: update out-of-date docs.

jaycarlton:
> I'm still a little worried about the full table scans we're doing every minute but I agree its probably not going to be an issue right now given we don't have much data.

Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4265

PR #2942, opened 2019-12-20
[no ticket][risk=low] Fix js error with 204 from status alert

als364:
When there was no status alert in the DB, we were getting a JS error because the Swagger autogen client side code was trying to parse json out of the response (which doesn't exist in a 204). I guess 204s are Bad. We send back an empty StatusAlert object now.


jaycarlton:
Why not make a 204 respomse class in swagger that's just empty? Should get
you some curly braces. I worry changing the response code is an api change
that could conceivably break clients. Also we would want the sane behavior
for every no content response.

I'm on mobile so I apologize if I missed obvious context.

-Jay

als364:
Unless I'm misunderstanding you, that's what I had before, and it was causing this error.

jaycarlton:
I was saying see if we could get the server to generate an empty json
object instead of an empty/null payload.

als364:
I tried that, it converted it to an empty payload under the hood somewhere...

PR #2945, opened 2019-12-20
[no ticket][risk=no] Clickable checkbox label

als364:
Haven't fixed this everywhere because we do a bunch of inane DOM choices like sticking the checkbox in one flexbox and the label at the top of the flexbox next to it... but here is at least one clickable checkbox label.

### DIFF
--- a/ui/src/app/components/inputs.tsx
+++ b/ui/src/app/components/inputs.tsx
@@ -162,7 +162,13 @@ export class LabeledCheckbox extends React.Component<LabeledCheckboxProps, Label
           style={{...this.props.checkboxStyle, verticalAlign: 'middle'}}
           checked={this.state.value}
           disabled={this.props.disabled}
-          onChange={e => this.props.onChange(e)}
+          onChange={
+            e => {
+              if (this.props.onChange) {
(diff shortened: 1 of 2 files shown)


ui/src/app/components/inputs.tsx
```diff
@@ -127,6 +128,52 @@ export const CheckBox = ({onChange, ...props}) => {
...
+    }
+  }
+
+  render() {
+    return <div style={this.props.style}>
+      <CheckBox
+          style={{...this.props.checkboxStyle, verticalAlign: 'middle'}}
+          checked={this.state.value}
+          disabled={this.props.disabled}
+          onChange={e => this.props.onChange(e)}
```
gjuggler: onChange is nullable in the props typedef, so I think this needs to check first whether this.props.onChange exists.

als364: Good catch, thank you.

als364:
No visual change.

jaycarlton:
I filed this a while back if you want to use it for this and related PRs https://precisionmedicineinitiative.atlassian.net/browse/RW-3650

PR #2948, opened 2019-12-26
[RISK=MEDIUM][RW-4091] Batch Job to send researcher data to RDR Api

NehaBroad:
This PR, 

- Creates a queue in cloud task
- Create cloud task that contains ids that needs to be exported
- Call RDR Api and send user and workspace object

NOTE: This PR will not have unit test, the purpose of this PR is to test end to end functionality(create task and task calling workbench API) on TEST environment since this cannot be test totally on local environment.

**PR checklist**

- [x ] The JIRA story has been moved to Dev Review
- [ Partially] I have run and tested this change locally
- [ X] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later


api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+paths:
+  /workbench/directory/workspaces:
+    post:
+      description: Submits one or more workspace snapshots for storage in the RDR.
+      tags:
+        - RDR
+      consumes:
+        - application/json
+      operationId: sendWorkspace
```
gjuggler: [Non-opt] For operation name here, I'd consider something like "saveWorkspaces" or "updateWorkspaces". Think about if this API were published as part of the Workbench system, we'd probably use words like "save" or "update" rather than "send".

I would also be sure to make the operation use plural wording, since it accepts one or more workspaces.

This comment applies equally to the researchers endpoint.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+definitions:
+  RDRWorkspace:
+    type: object
+    properties:
+      workspaceId:
+        type: integer
+      name:
+        type: string
+      creationTime:
```
gjuggler: Would it be too late to work with the RDR team to rename this field to "createdTime"? See my comments in the Liquibase changelog for why I think it would be preferred. It would be nice to have the RDR API align with our internal schema where possible.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+          - WRITER
+          - OWNER
+      status:
+        type: string
+        description: Workspace User Status
+        enum:
+          - ACTIVE
+          - INACTIVE
+
+  RDRResearcher:
```
gjuggler: Maybe I missed it in the design doc, but we will also need an "excludeFromPublicDirectory" field on the researcher level, to allow admins to prevent non-end-user objects from being published to the directory. That may require some follow-up with the RDR team, so it's worth tracking as a separate small ticket.

api/src/main/resources/workbench.yaml
```diff
@@ -1141,6 +1141,22 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
+  /v1/cron/exportToRDR:
+    get:
+      security: []
+      tags:
+        - offlineRDRExport
```
gjuggler: Side note: don't forget to rename "RDR" --> "Rdr" for proper camel-casing in these YAML files as well. Sorry for the busywork, but experience shows it's super helpful to have consistency in how we handle naming issues like this.

api/src/main/resources/workbench.yaml
```diff
@@ -2159,10 +2175,58 @@ paths:
           schema:
             $ref: "#/definitions/DataSetPreviewResponse"
 
+  /v1/offline/syncResearcherData:
```
gjuggler: [API design thought] I know we sort of had a pattern with "offline" in the URL path, but in this context it feels a little awkward. It might fit better with two changes:

1. Replace "offline" with "rdrExport" here, so the URL path here would become /v1/rdrExport/syncResearcherData
2. Remove "offline" from the controller, so it becomes RdrExportController.

api/src/main/resources/workbench.yaml
```diff
@@ -2159,10 +2175,58 @@ paths:
...
+      description: >
+        Send Researcher data
+      operationId: "syncResearcherData"
+      parameters:
+        - in: body
+          name: researchIds
+          required: true
+          description: The IDs of concepts to be added to the concept set.
+          schema:
+            type: object
```
gjuggler: Is there a reason this needs to be an object rather than an array of numbers? You could follow the pattern from CreateConceptSetRequest:

        type: array
        description: The IDs of concepts to be added to the concept set.
        items:
          type: integer
          format: int64

api/src/main/resources/workbench.yaml
gjuggler: [Non-opt] Naming nitpick: For consistency let's align on a single verb. Prefer using "export" rather than "sync" or "send", e.g. /v1/rdrExport/exportResearcherData, or even /v1/rdrExport/exportResearchers which makes it obvious that the input is a list.

This applies equally to the workspaces method name / URL path.

api/src/main/resources/workbench.yaml
```diff
@@ -2159,10 +2175,58 @@ paths:
...
+        200:
+          description: >
+            A SQL query for each domain in the Data Set
+
+
 ##########################################################################################
 ## DEFINITIONS
 ##########################################################################################
 definitions:
+  ResearcherId:
```
gjuggler: It looks like this definition is unused (unless it's used by the code but not referenced from within the YAML). Remove?

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -24,6 +24,7 @@
   public FeatureFlagsConfig featureFlags;
   public BillingConfig billing;
   public ActionAuditConfig actionAudit;
+  public RDRConfig rdrServer;
```
gjuggler: [Non-opt] Consistency note: please follow the pattern used for other fields, where the class name is "FooConfig" and the field name is "foo". I think the base name here could be "RdrExport", so you'd have

public RdrExportConfig rdrExport;

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -191,9 +193,17 @@ public static WorkbenchConfig createEmptyConfig() {
...
     public boolean enableNewAccountCreation;
+    // Flag to indicate if USER/WORSPACE data is exported to RDR
+    public boolean enableRDRExport;
   }
 
   public static class ActionAuditConfig {
     public String logName;
   }
+
+  // Defines the rdr host per environment
```
gjuggler: Please expand on the comments on these fields – at a minimum it's helpful to provide an example value and some sort of pointer to who "owns" these values or where else in our system they're configured.

api/src/main/java/org/pmiops/workbench/rdr/RDRConfig.java
```diff
@@ -0,0 +1,44 @@
+package org.pmiops.workbench.rdr;
```
gjuggler: Package naming thought: in line with my other comments strewn about, I think we should treat this package as the "RDR export" package. This might motivate us to rename the package to "rdrexport", and use this as a base name for all other classes: RdrExportService, RdrExportConfig, etc.

I don't have a super-strong feeling here, but I suspect there will be less overall cognitive overhead if we stick to one thing, and I think "RDR export" will capture the gist of it pretty well.

api/db/changelog/db.changelog-112-addColumn-user-create-modified.xml
```diff
@@ -0,0 +1,17 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="nsaxena" id="db.changelog-112-addColumn-user-create-modified">
+    <addColumn tableName="user">
+       <column name="last_modified_time" type="datetime" defaultValueDate="current_timestamp">
+        <constraints nullable="true"/>
+      </column>
+      <column name="create_time" type="datetime" defaultValueDate="current_timestamp">
+        <constraints nullable="false" />
```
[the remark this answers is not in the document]
NehaBroad: For #2: I started out thinking that the last_modified_time will be updated only if user has updated/made some changes their Profile page . However i changed the logic later after looking how last_modified is used at other places. I forgot to update liquibase will update this by removing non-nullable constraint.

api/src/main/java/org/pmiops/workbench/db/dao/RDRExportDao.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.data.repository.CrudRepository;
+
+public interface RDRExportDao extends CrudRepository<DbRDRExport, Long> {
+
+  @Query(
+      nativeQuery = true,
+      value =
+          "select u.user_id from user u LEFT JOIN rdr_export rdr on"
+              + " u.user_id = rdr.export_id and rdr.entity = 1 where u.last_modified_time > rdr.last_Export_date or rdr.export_id is null")
+  List<BigInteger> findDbUserToExport();
```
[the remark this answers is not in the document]
NehaBroad: This gives an error if i use Long as the return type, i suspect it is because i am using native query rather than mapping it to DBUser. i'll add comments around it

api/src/main/resources/workbench.yaml
NehaBroad: I will be updating the comment, since this is not integer but a comma separated list of ids, since we can send up task body can just contain ByteStrings

api/build.gradle
```diff
@@ -160,6 +160,30 @@ task generateMoodleClient {
   }
 }
 
+def rdrApiFile = 'src/main/resources/rdr.yaml'
```
jaycarlton: Do we need to keep their swagger file locally, or can we pull it in from a URL so we always have the latest?

api/build.gradle
```diff
@@ -322,6 +346,8 @@ ideaModule.dependsOn generateMandrillClient
 compileGeneratedJava.dependsOn generateMandrillClient
 ideaModule.dependsOn generateMoodleClient
 compileGeneratedJava.dependsOn generateMoodleClient
+ideaModule.dependsOn generateRdrClient
```
jaycarlton: Make sure you can run tests in IntelliJ without doing external steps in gradle, if possible.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+        type: string
+        description: Workspace Status
+        enum:
+          - ACTIVE
+          - INACTIVE
+      workspaceUsers:
+        type: array
+        items:
+          $ref: '#/definitions/RdrWorkspaceUser'
+      excludeFromPublicDirectory:
```
jaycarlton: Is it too late to replace all these named booleans with an array of enum type? for example you could have `[diseaseFocusedResearch, ancestry, methodsDevelopment]` if only those three are true.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+  ResearcherAffiliation:
+    type: object
+    properties:
+      institution:
+        type: string
+      role:
+        type: string
+#        Need to talk to Yu about this to be an ENUM
+      nonAcademicAffiliation:
+        type: boolean
```
jaycarlton: check editor settings, as we shouldn't end the file without a newline.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+  - googleoauth:
+    - openid
+    - email
+    - profile
+    - https://www.googleapis.com/auth/devstorage.full_control
+
+produces:
+  - application/json
+
+##########################################################################################
```
jaycarlton: I'd prefer to avoid comments in swagger files, as they don't stick around when the files are processed and transformed. It's easy to find the paths section using the Structure tab in IJ.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -0,0 +1,265 @@
...
+   * @param entity
+   * @param idList
+   */
+  private void updateDBRdrExport(RdrEntity entity, List<Long> idList) {
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+
+    List<DbRdrExport> exportList =
+        idList.stream()
+            .map(
+                id -> {
```
jaycarlton: It would be more readable to extract a function here. It's very rare that you want an if statement inside a lambda.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -0,0 +1,265 @@
...
+                  DbRdrExport rd =
+                      rdrExportDao.findByEntityTypeAndEntityId(
+                          RdrEntityEnums.entityToStorage(entity), id);
+                  // If Entry doesn't exist in rdr_export create an object else just update the
+                  // export Date
+                  // to right now
+                  if (rd == null) {
```
jaycarlton: You can use `DbRdrExport rd =Optional.ofNullable(dao.findBy...).orElse(exportFromEntity(entity))` to lose the `if` if you want.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+        type: string
+      zipCode:
+        type: string
+      country:
+        type: string
+      ethnicity:
+        type: string
+#      Need to talk to Yu about this to be a list
+      gender:
+#      Need to talk to yu about this to be a list
```
jaycarlton: Or an object. I'd probably make it an object that contains a list so you have more room to add stuff later.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+produces:
+  - application/json
+
+##########################################################################################
+## PATHS
+##########################################################################################
+
+paths:
+  /workbench/directory/workspaces:
```
jaycarlton: Do we want a version indicator in the path like `v1`?

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+      openid: open id authorization
+      email: email authorization
+      profile: profile authorization
+
+security:
+  - googleoauth:
+    - openid
+    - email
+    - profile
+    - https://www.googleapis.com/auth/devstorage.full_control
```
jaycarlton: FWIW I got the error 

> Semantic error at security.0.3
> Security scope definition https://www.googleapis.com/auth/devstorage.full_control could not be resolved

 when validating this at https://editor.swagger.io/.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
+#This file provide the end points for interaction with RDR
+
+
+swagger: '2.0'
```
jaycarlton: Since this is a standalone, short API spec, it might be a good first experiment in OpenApi 3...

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+info:
+  title: RDR
+  description: |
+    Service for RDR Interactions.
+  version: "0.1"
+#The host will get updated as per the environment it will be pmi-drc-api-test for dev and test
```
jaycarlton: Please put yaml comments in descriptions instead of using `#`. The latter aren't well-supported by tools, since they're not a real part of the object model for the doc.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+##########################################################################################
+## PATHS
+##########################################################################################
+
+paths:
+  /workbench/directory/workspaces:
+    post:
+      description: Submits one or more workspace snapshots for storage in the RDR.
+      tags:
+        - Rdr
```
jaycarlton: Why is this an initial case identifier? Our tags in other swagger files are all lowercase.

api/src/main/resources/rdr.yaml
```diff
@@ -0,0 +1,202 @@
...
+      parameters:
+        - in: body
+          required: true
+          name: workspaces
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/RdrWorkspace"
+      responses:
+        200:
```
jaycarlton: nit/opinion: I don't think we really need to call out empty 200 and 500 responses explicitly unless there's a good reason. We're basically just documenting the http spec.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -0,0 +1,263 @@
...
+      log.severe("Error while sending workspace data to RDR");
+    }
+  }
+
+  // Convert workbench DBUser to RDR Model
+  private RdrResearcher toRdrResearcher(DbUser dbUser) {
+    RdrResearcher researcher = new RdrResearcher();
+    researcher.setUserId((int) dbUser.getUserId());
+    researcher.setCreationTime(dbUser.getCreationTime().toLocalDateTime().atOffset(offset));
+    if (dbUser.getLastModifiedTime() != null)
```
gjuggler: Is there a reason why we check for null values for lastModifiedTime but not for creationTime? It seems slightly inconsistent that we don't handle these timestamp fields in the same way.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -0,0 +1,263 @@
...
+                  return new ResearcherAffiliation()
+                      .institution(inst.getInstitution())
+                      .role(inst.getRole());
+                })
+            .collect(Collectors.toList()));
+    return researcher;
+  }
+
+  private RdrWorkspace toRdrWorkspace(DbWorkspace dbWorkspace) {
+    RdrWorkspace RdrWorkspace = new RdrWorkspace();
```
gjuggler: Same issue with uppercased local variable name.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -0,0 +1,263 @@
...
+   * @param idList
+   */
+  private void updateDBRdrExport(RdrEntity entity, List<Long> idList) {
+    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
+
+    List<DbRdrExport> exportList =
+        idList.stream()
+            .map(
+                id -> {
+                  DbRdrExport rd =
```
gjuggler: What does "rd" mean as a variable name? I'd lean towards using non-abbreviated variable names where possible. This could be "rdrExport".

api/src/main/resources/rdr.yaml
gjuggler: +1 to this – it looks like we insert a "/rdr/v1/" to the base path of our API client. If it comes out the same, it may be clearer to include that sub-path in each of the API method paths here.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -143,6 +145,7 @@ public static WorkbenchConfig createEmptyConfig() {
     // Controls whether all api requests are traced and sent to Stackdriver tracing, or
     // whether we only trace at the default frequency.
     public boolean traceAllRequests;
+    public String location;
```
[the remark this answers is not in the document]
gjuggler: Follow-up note: @jaycarlton is introducing a very similar config value in https://github.com/all-of-us/workbench/pull/2978/files, but that PR uses "us-central" instead of "us-central1". I have a feeling "us-central1" is more correct since that's the GCE-aligned location name, but there may be reasons we'd need both.

You guys should sync up and make sure we don't end up with competing config values here!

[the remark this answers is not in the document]
NehaBroad: Will talk to @jaycarlton  about this, For gcoud queues we will be using us-central1 
_This location will serve as the LOCATION_ID parameter for your Cloud Tasks requests, so make a note of it. Note that two locations, called europe-west and us-central in App Engine commands, are called, respectively, europe-west1 and us-central1 in Cloud Tasks commands_
https://cloud.google.com/tasks/docs/creating-queues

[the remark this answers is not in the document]
NehaBroad: Hmm i will be looking at jay's PR and if there does seem to be a need to maintain 2 locations i would prefer to name it something more specific to cloud task queue since it _might_ just be used for that

PR #2952, opened 2019-12-26
generated api files on master

jaycarlton:
Baseline generated files for validating swagger changes


jaycarlton:
We don't need a draft PR here; we can still do comparisons as long as the branch is open.

PR #2953, opened 2019-12-26
Jaycarlton/after rw 4183

jaycarlton:
swagger codegen output after changes from RW-4183. Hopefully I haven't abused the repo too much.


jaycarlton:
no longer needed after merging #2949

PR #2962, opened 2020-01-02
[RW-2840][risk=low] Migrate away from the deprecated GoogleCredential library.

gjuggler:
This is step 1 of 2 in our path towards moving away from GCS-stored private service account keys, which has been flagged as a lingering security issue (not urgent, but important to fix before too long). I decided to take a stab at this over the quiet holiday period. See the ticket and/or [this design sketch](https://docs.google.com/document/d/16Y1-bI8KqjuI7Fjv2XL-iPSpBzpL7PJWY5gxgHQbe-s/edit#) for more context.

This PR should be largely a no-op, except for one functional change: instead of reading credentials for our gsuite-admin service account from the App Engine package, we will read them from GCS (to be more consistent with our other GCS-stored service account keys).

The bulk of code changes here are split into two threads:

1) Remove all instances of com.google.api.client.googleapis.auth.oauth2.GoogleCredential (which is now deprecated) into use of com.google.auth.oauth2.GoogleCredentials and related subclasses (including ServiceAccountCredentials).
2) Clean up and simplify our integration test configs for clarity and to reduce duplication of credential-loading beans.

The second part (simplifying integration tests) might have been doable as a separate PR, but I found it hard to reason about the integration tests (and especially which credentials were being used for which test scenario) with all the inconsistency and boilerplate. If as a reviewer it's extremely distracting I'd be willing to try and split it out.

Things I learned along the way:
- [Credentials] By adopting the more modern credentials library, our creation of user-delegated credentials was dramatically simplified (see ServiceAccounts.java). To the extent where I felt that a unit test of that class became largely unnecessary.
- [Credentials] We'd been injecting an instance of the ServiceAccounts class, even though it was largely populated with utils-type methods. Pulling this out of our service constructors felt better.
- [Integration tests] It's possible to allow the use of request-scoped beans by including the @WebAppConfiguration annotation on a test class. This allowed us to remove the re-definition of a handful of beans.
- [Integration tests] Despite spending a fair bit of time trying to learn the ins & outs of Spring testing, I still find myself struggling a bit to predict whether a given change in boilerplate / infrastructure will work well or not. The one thing I did find helpful was to create a single base test class, which helps enforce that all test classes are using a similar base pattern of spring annotations.
- [Spring] I read up on some Spring fundamentals while working through some of the test refactors, and finally realized what Provider<X> is really all about: when you have a singleton bean that needs to interact with beans that have a shorter lifetime (e.g. prototype- or request-scoped), Provider<X> is the conventional pattern to allow the singleton bean to continuously re-request the more narrowly-scoped bean. Interesting! I'm not sure if we follow that pattern in a clear / crisp way.

---
**PR checklist**

- [Partially fulfills criteria] This PR meets the Acceptance Criteria in the JIRA story
- [Leaving as in-progress] The JIRA story has been moved to Dev Review
- [Mostly integration tests] This PR includes appropriate unit tests
- [Tested locally & tested a GAE-deployed version] I have run and tested this change locally


api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
```diff
@@ -66,56 +60,6 @@ public DbUser user(UserAuthentication userAuthentication) {
     return userAuthentication.getUser();
   }
 
-  @Bean
-  public WorkbenchEnvironment workbenchEnvironment() {
-    return new WorkbenchEnvironment();
-  }
-
-  /**
-   * Service account credentials for Gsuite administration. These are derived from a key JSON file
```
gjuggler: The change to this method (now located in StoredCredentialsConfig) should be the only functional change in this PR. Instead of loading the gsuite-admin SA creds from the App Engine resource package, I decided to follow-up on my prior TODO and switch to loading them from GCS. What this effectively means is that each App Engine task will talk to GCS when the first API call is made that requires G Suite interactions (e.g. creating a user). This bean is singleton-scoped, so the credentials should be loaded and cached for the lifetime of each GAE task.

gjuggler:
Tagging @jaycarlton and @calbach on this review.

CH, I'd like a second set of eyes on the change to load gsuite-admin creds from GCS instead of from the appengine package. I can't think of any obvious breakage this may cause, but please let me know if I'm missing something.

Jay – I thought you'd be good to review / critique the integration test refactor / cleanup. I got my hands a bit dirtier than I'd originally intended, but I think the end result is vastly cleaner to the original state of things, so possibly worth the effort to review.

api/src/integration/java/org/pmiops/workbench/NotebooksIntegrationTest.java
```diff
@@ -3,47 +3,23 @@
...
-          new NotebooksRetryHandler(new NoBackOffPolicy()),
-          workspaceService);
+  @TestConfiguration
+  @ComponentScan("org.pmiops.workbench.notebooks")
+  @Import({LeonardoNotebooksClientImpl.class})
+  static class Configuration {}
 
   @Test
   public void testStatus() {
     assertThat(leonardoNotebooksClient.getNotebooksStatus()).isTrue();
```
jaycarlton: I'd rename `getNotebooksStatus` to `isNotebooksServiceUp()`, as it better matches what it does and corresponds to a boolean.

Again, I'd rename this to `NotebooksSmokeTest`. I expected a lot more actual testing considering the boilerplate.

api/src/integration/java/org/pmiops/workbench/NotebooksIntegrationTest.java
```diff
@@ -3,47 +3,23 @@
...
-  private final LeonardoNotebooksClient leonardoNotebooksClient =
-      new LeonardoNotebooksClientImpl(
-          Providers.of(clusterApi),
-          Providers.of(notebooksApi),
-          Providers.of(createConfig()),
-          Providers.of(null),
-          new NotebooksRetryHandler(new NoBackOffPolicy()),
-          workspaceService);
+  @TestConfiguration
+  @ComponentScan("org.pmiops.workbench.notebooks")
```
jaycarlton: please use a typed basePackages ref if possible, as above.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -2,37 +2,23 @@
...
-import org.springframework.retry.backoff.NoBackOffPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
 
-public class DirectoryServiceImplIntegrationTest {
-  private DirectoryServiceImpl service;
-  private final GoogleCredential googleCredential = getGoogleCredential();
-  private final WorkbenchConfig workbenchConfig = createConfig();
-  private final ApacheHttpTransport httpTransport = new ApacheHttpTransport();
+public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
```
jaycarlton: Do we need the `Impl` in the middle of the test class name (here and above)?

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -2,37 +2,23 @@
...
-            Providers.of(workbenchConfig),
-            httpTransport,
-            new GoogleRetryHandler(new NoBackOffPolicy()));
-  }
+  @TestConfiguration
+  @Import(DirectoryServiceImpl.class)
+  static class Configuration {}
 
   @Test
   public void testDummyUsernameIsNotTaken() {
```
jaycarlton: Wait, so this test is explicitly making sure usernames are recyclable? I didn't know we allowed that. What's the benefit of allowing that?

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -1,52 +1,26 @@
...
-import org.pmiops.workbench.config.WorkbenchEnvironment;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
 
 /**
  * Handles functionality related to loading service account credentials and generating derived /
  * impersonated credentials.
  */
-@Component
 public class ServiceAccounts {
```
jaycarlton: I was very pro-utility class until I realized just how much Spring and Mockito aren't excited about working with them. If there are unit tests you want to write that depend on any of these functions, you can't inject a mock.

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -1,52 +1,26 @@
...
-    if (workbenchEnvironment.isDevelopment()) {
-      GoogleCredential credential = GoogleCredential.getApplicationDefault().createScoped(scopes);
-      credential.refreshToken();
-      return credential.getAccessToken();
-    }
-    AppIdentityService appIdentity = AppIdentityServiceFactory.getAppIdentityService();
-    final AppIdentityService.GetAccessTokenResult accessTokenResult =
-        appIdentity.getAccessToken(scopes);
-    return accessTokenResult.getAccessToken();
+  public static String workbenchAccessToken(List<String> scopes) throws IOException {
```
jaycarlton: I'd probably have a verb in the function name, like `getRefreshedWorkbenchAccessToken()`

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -1,52 +1,26 @@
...
+    GoogleCredentials scopedCreds = GoogleCredentials.getApplicationDefault().createScoped(scopes);
+    scopedCreds.refresh();
```
jaycarlton: nit: can this just be chained with the statement prior?

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -56,30 +30,18 @@ public String workbenchAccessToken(WorkbenchEnvironment workbenchEnvironment, Li
...
+   * @param originalCredentials
    * @param userEmail Email address of the user to impersonate.
    * @param scopes The list of Google / OAuth API scopes to be authorized for.
    * @return
    * @throws IOException
    */
-  public GoogleCredential getImpersonatedCredential(
-      GoogleCredential serviceAccountCredential, String userEmail, List<String> scopes)
+  public static GoogleCredentials getImpersonatedCredentials(
+      GoogleCredentials originalCredentials, String userEmail, List<String> scopes)
```
jaycarlton: is this the `username` property (i.e. not contact email) in the new usage? If so, I recommend `username`. as a parameter name.

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -35,7 +35,7 @@
   private final Provider<WorkbenchConfig> workbenchConfigProvider;
   private final Provider<CloudStorageService> cloudStorageServiceProvider;
   private final Clock clock;
-  private final LoadingCache<String, GoogleCredential> garbageCollectionSACredentials;
+  private final LoadingCache<String, GoogleCredentials> garbageCollectionSACredentials;
```
jaycarlton: nit: I'd probably abreviate Garbage Collection as gc and spell out the rest, like `gcServiceAccountCredentials`. Not used to seeing abbreviations in the middle of identifiers.

api/src/main/java/org/pmiops/workbench/billing/BillingGarbageCollectionService.java
```diff
@@ -137,13 +137,12 @@ private void transferOwnership(final String projectName) {
     fireCloudService.addOwnerToBillingProject(garbageCollectionSA, projectName);
 
     try {
-      final GoogleCredential gcsaCredential =
+      final GoogleCredentials gcsaCredential =
```
jaycarlton: Can we factor it out a bit more and inject a `Provider<GoogleCredentials>` that always gives you refreshed ones? Having to call `refresh()` like this is kinda temporal coupling, and in any case, we shouldn't expect a programmer to remember to do something we're not willing to ask the computer to remember. So if the provider doesn't work, we can at least have a helper method that fetches and refreshes in one shot and returns one ready to go, RAII-style.

api/src/main/java/org/pmiops/workbench/config/StoredCredentialsConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import org.pmiops.workbench.google.CloudStorageService;
+import org.pmiops.workbench.google.CloudStorageServiceImpl;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Lazy;
+
+@Configuration
+@Import(CloudStorageServiceImpl.class)
+public class StoredCredentialsConfig {
```
jaycarlton: nit: recommend `StoredCredentialsSpringConfiguration`, as above.

api/src/main/java/org/pmiops/workbench/config/StoredCredentialsConfig.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Service account credentials for Gsuite administration, corresponding to the "gsuite-admin"
+   * service account in each environment. Enabled for domain-wide delegation of authority.
+   */
+  @Lazy
+  @Bean(name = Constants.GSUITE_ADMIN_CREDS)
```
jaycarlton: nit: After fiddling with this, I remember why this has to be a constant string rather than an enum name. Still, I'd likely rename `Constants` to `AuthBeanNames`.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
jaycarlton: But it still needs to be refreshed during the task lifetime?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudConfig.java
```diff
@@ -55,14 +54,10 @@ public ApiClient endUserApiClient(
...
-      WorkbenchEnvironment workbenchEnvironment,
-      WorkbenchConfig workbenchConfig,
-      ServiceAccounts serviceAccounts) {
+  public ApiClient allOfUsApiClient(WorkbenchConfig workbenchConfig) {
     ApiClient apiClient = buildApiClient(workbenchConfig);
     try {
-      apiClient.setAccessToken(
-          serviceAccounts.workbenchAccessToken(workbenchEnvironment, BILLING_SCOPES));
+      apiClient.setAccessToken(ServiceAccounts.workbenchAccessToken(BILLING_SCOPES));
     } catch (IOException e) {
```
jaycarlton: nit: is there any additional context you want to add before rethrowing?

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudService.java
```diff
@@ -24,6 +24,9 @@
 
   String WORKSPACE_DELIMITER = "__";
 
+  /** Returns the base path for the FireCloud API. Exposed for integration testing purposes only. */
+  String getBasePath();
```
jaycarlton: Should you add `@VisibleForTesting` at the interface level too? I don't believe there's any differentiation between unit and other types of testing.

Would it make sense to define this one level up, at some common parent interface for all REST services we depend on? Probably not worth it right now.

api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -152,6 +150,12 @@ private void checkAndAddRegistered(FirecloudWorkspaceIngest workspaceIngest) {
     workspaceIngest.setAuthorizationDomain(ImmutableList.of(registeredDomain));
   }
 
+  @Override
+  @VisibleForTesting
+  public String getBasePath() {
```
jaycarlton: nit: maybe rename this to `getFirecloudApiBasePath()` so it's more obviously distinct from the service's object model.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -130,36 +130,36 @@ private String readCredentialsBucketString(String objectPath) {
...
     final String json = readCredentialsBucketString(objectPath);
-    return GoogleCredential.fromStream(new ByteArrayInputStream(json.getBytes()));
+    return ServiceAccountCredentials.fromStream(new ByteArrayInputStream(json.getBytes()));
   }
 
   @Override
-  public GoogleCredential getGSuiteAdminCredentials() throws IOException {
-    return getCredential("gsuite-admin-sa.json");
+  public ServiceAccountCredentials getGSuiteAdminCredentials() throws IOException {
+    return getCredentials("gsuite-admin-sa.json");
```
jaycarlton: I'd make these json strings constants or configurable.

api/src/main/java/org/pmiops/workbench/google/DirectoryService.java
```diff
@@ -6,8 +6,21 @@
 public interface DirectoryService {
```
jaycarlton: Would renaming this to `GSuiteDirectoryService` or similar be clearer at a glance? If you're looking in the tree view it's probably clear, but I like to jump around.

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -65,42 +69,35 @@
...
-        .setServiceAccountPrivateKey(googleCredential.getServiceAccountPrivateKey())
-        .setServiceAccountPrivateKeyId(googleCredential.getServiceAccountPrivateKeyId())
-        .setTokenServerEncodedUrl(googleCredential.getTokenServerEncodedUrl())
-        .build();
+    return ServiceAccounts.getImpersonatedCredentials(
+        googleCredentialsProvider.get(), "directory-service@" + gSuiteDomain, SCOPES);
   }
 
-  private Directory getGoogleDirectoryService() {
+  private Directory getGoogleDirectoryService() throws IOException {
```
jaycarlton: maybe document when you should expect an IOException.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -171,25 +167,7 @@ public void setUp() throws MessagingException {
             fireCloudService,
             directoryService,
             cloudStorageService,
-            leonardoNotebooksClient,
```
jaycarlton: nit: if you still have appetite for refactoring, consider autowiring this. I believe we shouldn't need to use `Providers.of()` if we're setting up the test configuration properly.

api/src/test/java/org/pmiops/workbench/firecloud/FireCloudServiceImplTest.java
```diff
@@ -197,21 +190,6 @@ public void testNihCallbackServerError() throws Exception {
     service.postNihCallback(any());
   }
 
-  @Test
```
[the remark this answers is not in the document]
jaycarlton: For me, the value in this kind of test is just live documentation rather than coverage. Also, if there are any failure modes (like when failing to refresh), it might be good to demonstrate what is expected to happen.

api/tools/src/main/java/org/pmiops/workbench/tools/ServiceAccountAPIClientFactory.java
```diff
@@ -24,10 +24,10 @@ public ServiceAccountAPIClientFactory(String apiUrl) {
   private ApiClient newApiClient(String apiUrl) throws IOException {
     ApiClient apiClient = new ApiClient();
     apiClient.setBasePath(apiUrl);
-    GoogleCredential credential =
-        GoogleCredential.getApplicationDefault().createScoped(Arrays.asList(FC_SCOPES));
-    credential.refreshToken();
-    apiClient.setAccessToken(credential.getAccessToken());
+    GoogleCredentials credentials =
```
jaycarlton: So is "credentials" singular now? Or both?

api/tools/src/main/java/org/pmiops/workbench/tools/ServiceAccountAPIClientFactory.java
```diff
@@ -1,6 +1,6 @@
 package org.pmiops.workbench.tools;
 
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.auth.oauth2.GoogleCredentials;
 import java.io.IOException;
 import java.util.Arrays;
 import org.pmiops.workbench.firecloud.ApiClient;
```
jaycarlton: Why do we need to use a factory class instead of defining a bean in a configuration somewhere? If we had an ApiUrlProvider it could be beans all the way down...

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
calbach: This is not making any statement about recycling usernames. It's testing that a user which has never been created is not taken. The test would break if someone create such a user, but hopefully they won't as indicated by the username.

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -1,52 +1,26 @@
...
-      GoogleCredential credential = GoogleCredential.getApplicationDefault().createScoped(scopes);
-      credential.refreshToken();
-      return credential.getAccessToken();
-    }
-    AppIdentityService appIdentity = AppIdentityServiceFactory.getAppIdentityService();
-    final AppIdentityService.GetAccessTokenResult accessTokenResult =
-        appIdentity.getAccessToken(scopes);
-    return accessTokenResult.getAccessToken();
+  public static String workbenchAccessToken(List<String> scopes) throws IOException {
+    GoogleCredentials scopedCreds = GoogleCredentials.getApplicationDefault().createScoped(scopes);
```
calbach: Per Jay's comment above, in this case I would either change this function to accept a GoogleCredentials as a parameter (the client would pass ADC), or revert the static change and go back to a component/service

api/src/main/java/org/pmiops/workbench/config/StoredCredentialsConfig.java
```diff
@@ -0,0 +1,48 @@
...
+  public ServiceAccountCredentials gsuiteAdminCredential(CloudStorageService cloudStorageService)
+      throws IOException {
+    return cloudStorageService.getGSuiteAdminCredentials();
```
calbach: Probably just a TODO: this method (and friends) should not exist on cloudStorageService. Now that you have this bean, I would remove it there, and just have this call a basic readBlob() method instead; knowledge of that file's name etc would be better encoded in this provider.

api/src/main/java/org/pmiops/workbench/config/RetryConfig.java
```diff
@@ -14,6 +18,7 @@
 import org.springframework.retry.policy.SimpleRetryPolicy;
 
 @Configuration
+@Import({NotebooksRetryHandler.class, GoogleRetryHandler.class, FirecloudRetryHandler.class})
```
calbach: Hmmm... this is importing all of the subclasses of `RetryConfig.ResponseCodeRetryPolicy`. That seems a bit circular to me. What was the rationale for this change? Please also add a comment here explaining the criteria for adding another handler here.

api/src/integration/java/org/pmiops/workbench/BaseIntegrationTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.context.web.WebAppConfiguration;
+
+@RunWith(SpringRunner.class)
+@WebAppConfiguration
+@Import({IntegrationTestConfig.class})
+public abstract class BaseIntegrationTest {
+
+  static WorkbenchConfig config;
```
calbach: package private doesn't seem correct here - is the intent that subclasses can modify this? If so, probably should be public or protected (likewise for `loadConfig`)

calbach: I'd be a bit worried about state carrying over between test cases. Perhaps a setup/tearDown base method integrated approach would be safer. i.e. `setUp()` calls an optional method `getWorkbenchConfig()` (used when there's a consistent config used for the whole test class), `tearDown()` nulls it out. Still allow direct access to `config` in case a subclass needs to customize the config per individual test cases.

api/src/main/java/org/pmiops/workbench/config/RetryConfig.java
jaycarlton: Aside: If the intent is to import a bunch of classes that have something in common (like a parent interface), I bet there's a pattern for that, such that another new subclass would automatically get picked up.

api/src/integration/java/org/pmiops/workbench/BaseIntegrationTest.java
jaycarlton: So the first question for me is whether a subclass is expected to grab this config object and mutate it, or supply its own config bean and import it in with a qualifier. I feel like the latter approach is more idiomatic in Spring. Maybe have a default builder or something for an IntegrationTestConfig for convenience.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -80,20 +66,4 @@ private static RetryTemplate retryTemplate() {
     tmpl.setThrowLastExceptionOnExhausted(true);
     return tmpl;
   }
-
-  private static GoogleCredential getGoogleCredential() {
-    try {
-      String saKeyPath = "src/main/webapp/WEB-INF/gsuite-admin-sa.json";
```
calbach: Sorry, I missed your original request from me. Not sure why we were doing things this way - reading from GCS seems totally fine. Either as part of this, or as a follow-up cleanup ticket, I would also teardown the code that bundles the keyfile into the application itself: https://github.com/all-of-us/workbench/blob/dee66cf03b19209b35d087c35adeb8f4ce893ec8/api/libproject/devstart.rb#L344

api/src/main/java/org/pmiops/workbench/config/RetryConfig.java
calbach: Thanks, I find this to be much clearer.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
gjuggler: Thanks for the sanity-check. I'll flag that task as part of the overall private-key-storage cleanup (just filed as RW-4250).

PR #2968, opened 2020-01-03
[risk=no][RW-4177] Add compute tags to endpoints

ericsong:
Description:
Adding compute tag to endpoints that can incur compute costs. There will be a follow up PR that disables these endpoints when workspaces are marked as billing inactive.

---
**PR checklist**


api/src/main/resources/workbench.yaml
```diff
@@ -550,6 +550,7 @@ paths:
       operationId: listClusters
       tags:
         - cluster
+        - compute
```
calbach: may want to make this tag name more specific to billing lockout, something like billingRequired; we'll likely have compute-related endpoints in other contexts (e.g. administrative) which should not have the lockout restriction.

api/src/main/resources/workbench.yaml
```diff
@@ -1335,6 +1338,7 @@ paths:
     post:
       tags:
         - workspaces
+        - compute
```
calbach: I don't know whether this should be tagged. This is different because lockout should be applied to the destination (found in the request payload), and is unrelated to the source workspace in the URL. Therefore any kind of interceptor-based approach will not really be able to use this tag to apply any standard URL-based filtering.

api/src/main/resources/workbench.yaml
jaycarlton: +1 Also, are tags the only way to have custom attributes on paths? Or can we define something with more structure if we need it? It's somewhat worrying that we have to remember the first tag is the controller name, then this new tag pertains to billing, etc.

Also, please let me know if getting to OpenAPI 3 earlier would make your job easier here.

api/src/main/resources/workbench.yaml
jaycarlton: I'm also nervous because our API structure wasn't originally designed to be partitioned along these lines. I'd almost rather see a v2 API that separates billable and non-billable endpoints by path. I imagine that's a nonstarter for the first launch though.

ericsong: Which workspace gets billed after a copy operation? I did some googling but was unable to find a clear answer.

I'll ask Karthik what he thinks the right product behavior is. From a technical perspective, we can move the billing check into the controller method if we want to check the destination workspace's status.

calbach: The billing costs involved here are purely coming from GCS. There might be a small egress charge to the source workspace during a copy, but if so I think it's negligible enough to ignore. If we copy a file into a destination workspace bucket, that workspace will be billed monthly for any storage costs.

The goal with billing lockout would be to prevent increases in workspace cost, namely due to the user creating a cluster, or (less significantly) adding GCS objects to the workspace (increasing the storage cost per month).

als364: If it makes it easier (/ possible) to use this approach for billing lockout, given our time constraints, I would be ok with disabling copying the notebook to another workspace - but ultimately the behavior of the app is a product decision.

jaycarlton: What would be the advantage of copying the notebook in this situation? The other workspace could still have a billing acct/free credits?

als364: Correct.

PR #2969, opened 2020-01-04
[RW-4235][risk=no] Log when a user observes a held notebook lock

calbach:
Description:
This was requested to establish a rough estimate of the effect of how often users are affect by notebook locking.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
+++ b/api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
@@ -703,6 +703,22 @@ public class WorkspacesController implements WorkspacesApiDelegate {
       }
     }
 
+    // If a lock is held by another user, log this to establish a rough estimate of how often
+    // locked notebooks are encountered. Note that this only covers locks encountered from the
+    // Workbench - any Jupyter UI-based lock detection does not touch this code path.
+    String currentUsername = userProvider.get().getUsername();
+    if (response.getLockExpirationTime() == null
+        || response.getLastLockedBy() == null
+        || response.getLockExpirationTime() < clock.millis()
+        || response.getLastLockedBy().equals(currentUsername)) {
+      log.info(String.format("user '%s' observed notebook available for editing", currentUsername));
+    } else {
+      log.info(
+          String.format(
+              "user '%s' observed notebook locked by '%s'",
+              currentUsername, response.getLastLockedBy()));
+    }
+
     return ResponseEntity.ok(response);
   }
 


api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -703,6 +703,20 @@ private Boolean copyBlob(String bucketName, Blob b) {
...
     }
 
+    // If a lock is held by another user, log this to establish a rough estimate of how often
+    // locked notebooks are encountered. Note that this only covers locks encountered from the
+    // Workbench - any Jupyter UI-based lock detection does not touch this code path.
+    if (response.getLockExpirationTime() != null && response.getLastLockedBy() != null) {
+      String currentUsername = userProvider.get().getUsername();
+      if (clock.millis() < response.getLockExpirationTime()
+          && !response.getLastLockedBy().equals(currentUsername)) {
+        log.info(
```
jaycarlton: If it's useful, we can either fire an event metric for this to StackDriver, or add a log-based metric.

calbach: Nice suggestion, added - PTAL

calbach:
@jaycarlton please review

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -703,6 +708,23 @@ private Boolean copyBlob(String bucketName, Blob b) {
       }
     }
 
+    // If a lock is held by another user, log this to establish a rough estimate of how often
+    // locked notebooks are encountered. Note that this only covers locks encountered from the
+    // Workbench - any Jupyter UI-based lock detection does not touch this code path.
+    MonitoringViews lockMetric = MonitoringViews.NOTEBOOK_CHECK_AVAILABLE;
```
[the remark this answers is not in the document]
calbach: I expected you would leave a final-related comment here. I'm aware that Java supports doing it the way you describe, but my preference is to initialize to a happy-path default value, and then override in exceptional cases.

api/src/main/java/org/pmiops/workbench/monitoring/views/MonitoringViews.java
```diff
@@ -29,6 +29,11 @@
   NOTEBOOK_SAVE("notebook_save", "Save (or create) a notebook"),
   NOTEBOOK_CLONE("notebook_clone", "Clone (duplicate) a notebook"),
   NOTEBOOK_DELETE("notebook_delete", "Delete a notebook"),
+  NOTEBOOK_CHECK_LOCKED(
```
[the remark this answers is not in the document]
calbach: I saw that go by - sure, that would be preferred, I will wait for it since this isn't particularly urgent.

calbach:
@jaycarlton please take another look, I've updated the PR to utilize the new attachment functionality.

api/src/main/java/org/pmiops/workbench/monitoring/views/GaugeMetric.java
```diff
@@ -19,6 +19,10 @@
       "dataset_count_2",
       "Count of all datasets in existence",
       ImmutableList.of(MetricLabel.DATASET_INVALID)),
+  NOTEBOOK_LOCK_CHECK(
+      "notebook_lock_check",
+      "Counts of notebook availability due to locking, as observed by the caller",
```
jaycarlton: Looks good.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -703,6 +710,27 @@ private Boolean copyBlob(String bucketName, Blob b) {
...
+    if (notebookAvailable) {
+      log.info(
+          String.format(
+              "user '%s' observed notebook locked by '%s'",
+              currentUsername, response.getLastLockedBy()));
+    }
+    monitoringService.recordBundle(
+        MeasurementBundle.builder()
+            .addEvent(GaugeMetric.NOTEBOOK_LOCK_CHECK)
+            .addTag(MetricLabel.NOTEBOOK_AVAILABLE_TO_USER, Boolean.toString(notebookAvailable))
```
jaycarlton: Just a heads up: I'm about 75% sure that Stackdriver considers the Labels (tags) list to be part of the metric descriptor, so if you want to add another tag in the future, we have to either drop this one and replace it, or add a new one. I'll put this in the docs.

calbach:
I tried to validate this locally and I don't see any metrics making to to Stackdriver. I confirmed that the initial Stackdriver setup was logged correctly on my local API server, and that my cronjob is running. However, I don't see any metrics logged from my server - including existing metrics like workspace_count.

jaycarlton:
Hmm. I know I've seen gauge metrics at least from other users. I haven't done very much testing with the event metrics since it became clear they were second priority. Doing that now.

calbach:
Per offline discussions, reverted to normal logging since this type of gauge metric currently doesn't work.

calbach:
Confirmed via slack that Joel's approval still stands (effectively we're back to a similar solution as in the initial PR).

PR #2970, opened 2020-01-05
[RW-3726][RISK=NO] Single Swagger File

jaycarlton:
Remove merging code and check in the merged file as the input file (renamed workbench-api.yaml).

Verified outputs correspond using Beyond Compare 3 with unimportant text regex `@javax\.annotation\.Generated.*$` to mask out timestamp changes. I'll put up a separate draft PR to show the diff in github, but I don't know of a way to do the masking there.

Comparison PR is https://github.com/all-of-us/workbench/pull/2971.
---
**PR checklist**

Screen Shot 2020-01-05 at 1 58 50 PM

Screen Shot 2020-01-05 at 2 35 53 PM


api/build.gradle
```diff
@@ -7,22 +7,15 @@ import org.pmiops.workbench.tooling.IncrementalHotSwapTask
 // This path is the Java sources root for finding any overridden Swagger-codegen
 // templates we may be using.
 def swaggerTemplateDir = 'src/main/resources/swaggercodegen'
-def mergedApiSourceFile = 'src/main/resources/merged.yaml'
-def clientApiSourceFile = 'src/main/resources/client_api.yaml'
-def cbReviewApiSourceFile = 'src/main/resources/cb_review_api.yaml'
```
calbach: These other ones are still around, presumably should be deleted.

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+        name: term
+        description: String to find in user's name or email address. Search is a case-insensitive
+          substring match.
+        required: true
+        type: string
+      - in: query
+        name: pageToken
+        description: 'Pagination token retrieved from a previous call to user; used
+          for retrieving additional pages of results.
+
```
calbach: rm extra newline

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+        name: body
+        description: Cluster config update request.
+        schema:
+          "$ref": "#/definitions/UpdateClusterConfigRequest"
+      responses:
+        200:
+          description: success
+          schema:
+            "$ref": "#/definitions/EmptyResponse"
+  "/v1/cron/auditBigQuery":
```
calbach: An important comment was dropped here:

```
  # Note: all requests tagged as "cron" must have the header X-Appengine-Cron:
  # true, which app engine itself only sets when invoking as a cronjob.
  # See https://cloud.google.com/appengine/docs/standard/java/config/cron#securing_urls_for_cron
  # and o.p.w.interceptors.CronInterceptor which implements the header check.
```

Please preserve this

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+        schema:
+          "$ref": "#/definitions/DataTableSpecification"
+      responses:
+        200:
+          description: the query to run against the CDR to retrieve the data
+          schema:
+            "$ref": "#/definitions/CdrQuery"
+  "/v1/workspaces/{workspaceNamespace}/{workspaceId}/getCohortAnnotations":
+    parameters:
+    - "$ref": "#/parameters
```
calbach: Important doc dropped:

```
  # Authorities are a tool used to add discrete permissions to users. We currently
  # define the following authorities:
  #   1) REVIEW_RESEARCH_PURPOSE: This authority allows the user to review workspaces where
  #      the user has requested review to approve or reject.
  #   2) DEVELOPER: *Grants all other authorities* as well as some internal-only
  #      administrative endpoints, including group management and cluster
  #      management. Intended to be granted for devops needs, e.g. for oncall.
  #   3) ACCESS_CONTROL_ADMIN: This is actually basically a user admin authority, for people
  #      to perform actions on a user's enabled status and manual verification.
  #   4) FEATURED_WORKSPACE_ADMIN: Allows a user to publish workspaces
```

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+  version: 0.1.0
+  title: AllOfUs Workbench API
+  description: The API for the AllOfUs workbench.
+  termsOfService: http://www.pmi-ops.org/terms_of_service.html
+  contact:
+    name: developer_help@pmi-ops.org
+  license:
+    name: BSD
+host: api.pmi-ops.org
+securityDefinitions:
```
calbach: A comment explaining the significance of this security definition seems important to carry over, namely that it triggers the auth interceptor.

api/src/main/resources/workbench-api.yaml
calbach: This comment was dropped, it seems meaningful:

`      # These property names should match what is in ModifyParticipantCohortAnnotationRequest`

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+    required: true
+    description: specifies which cdr version
+  participantId:
+    in: path
+    name: participantId
+    type: integer
+    format: int64
+    required: true
+    description: specifies which participant
+paths:
```
calbach: Dropped a comment that seems important:

```
# Throughout, we use integer/int64 in preference to string/date-time because Swagger's
# date formatting is inconsistent between server and client. Time values are stored as
# milliseconds since the UNIX epoch.
```

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
+---
```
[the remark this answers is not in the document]
jaycarlton: It's an artifact of the Ruby Yaml writer I believe. The hyphens [separate a Yaml directive](https://stackoverflow.com/questions/50788277/why-3-dashes-hyphen-in-yaml-file) from the main body, and there's no main body here.

Yes, making it workbench.yaml makes sense to me. However, if I'm just checking in the merge output, you're going to lose git history and comments (and likely again, if I use a tool transition to OpenAPI 3).

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+  "/v1/status-alert":
+    get:
+      tags:
+      - status-alert
+      description: Returns status alert information if any exists.
+      operationId: getStatusAlert
+      responses:
+        200:
+          description: 'A status alert information object.
+
```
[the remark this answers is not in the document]
jaycarlton: I'll be glad to get rid of it. I thought it was intentional for some reason, which made me sad.

It looks like the `>` symbol in the original yaml files gives us this.

As for quote conventions, I think we should just pick an editor/linter and use that. It's a very big file to be cleaning up by hand.

api/src/main/resources/workbench-api.yaml
jaycarlton: How about making this a [tag description](https://swagger.io/docs/specification/grouping-operations-with-tags/)? Otherwise it'll get lost again when we translate to OpenAPI 3. There was a brief [conversation](https://pmi-engteam.slack.com/archives/CHRN2R51N/p1576941635031300) about this, and we plan to re-add all the comments at the end.

api/src/main/resources/workbench-api.yaml
jaycarlton: Let's put this on the top level `info` [object](https://swagger.io/docs/specification/api-general-info/) so we don't lose it again.

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,6369 @@
...
+    required: true
+    description: Concept set ID
+  dataSetId:
+    in: path
+    name: dataSetId
+    type: integer
+    format: int64
+    required: true
+    description: Data set ID
+  workspaceNamespace:
```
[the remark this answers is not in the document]
jaycarlton: My current thinking (not implemented yet) is to find or define a canonical Swagger ordering and always save the file to that. This would make it easier to keep the format consistent and avoid conflicts, by standardizing the quote style and always ordering the keys alphabetically. Probably drop all comments and encourage folks to use description fields or other mechanisms built into the structure instead.

On the other hand, this makes the file feel less like source code and more like a serialization format. In reality it's both, but treating it like the latter makes it easier to process for us, and keeps folks from expecting untethered text to stick to following nodes.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
calbach: The original file had newlines in certain places, which were clearly not intended as part of the description strings, that was the cause for this.

Sure that's fine, I'm not sure how you planned to resolve the issue (editing the original source vs modifying this file) so I flagged it.

api/src/main/resources/workbench-api.yaml
calbach: This one would be reasonable as a description tag

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
calbach: I'm not sure I follow, are you trying to build a custom autoformatter for this file? Why would we not treat this as a source file going forward? Why the concern about having things "untethered" to nodes? I assumed this was a one-time issue while you are attempting to automatically merge files and migrate us to OpenAPI 3.

Comments have a role in the source document that is separate from the API being documented, e.g. to establish conventions, track TODOs, explain Swagger-related hacks, noting to keep things in sync.

jaycarlton:
One solution for the internal-facing comments would be to define a vendor extension like `x-aou-dev-comment:` per https://swagger.io/docs/specification/2-0/swagger-extensions/

jaycarlton:
See also this idea for re-homing internal comments: https://github.com/all-of-us/workbench/pull/2981

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
calbach: I buy that we could move some of these comments into a special yaml properties, but I don't think my above question has been answered yet - i.e. why is this actually a problem?

Also note the scale here is 3-4 comments, so I don't think we should spend too much time automating something here.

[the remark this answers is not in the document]
jaycarlton: Sorry, missed these.

It's a subtle difference, but by letting go of comments, we can tighten up semantics.

Basically, I wanted license to use automatic, yaml- and swagger-processing tools to operate on this file for a few purposes, then save it into a canonical ordering (so manual edits continue working). The requirement to keep comments around makes both of those impossible, as they're not supported in Swagger per se, and only supported insofar as they're dropped on the floor in YAML. 

Examples of processes/tools requiring round tripping I was thinking about in addition to OpenAPI 3:
-  A structural linter that would lay out sections in the order we want and alphabetize the rest (in cases where ordering is arbitrary). I'd have to see if any string formatters exist that could apply readability rules to string definitions.
- A script/tool to split the file into subdirectories and reusable YAML files according to rules we decide on
- Custom tooling for us to validate that our APIs conform to our own standards in addition to OpenAPIs, that we don't omit critical tags or defaults, etc. (This would only need round-tripping if it had an "autofix" feature).

Comments absolutely have a role, but the issue is that in true code files, because relative order of lines is important in most places, and you only ever go in one direction, you don't have to worry about losing them. File types like XML, JSON, and JSON don't have as much directionality to them, and can be transformed in interesting ways by processors that only care about the semantics of the file. Comments like these live in the text file describing the document, not the document itself. Line numbers, text wrapping, quote styles, and inline comments seemed like a reasonable price to pay to me.

calbach:
Per offline discussion, fine moving forward with either approach (copy comments, or move some into custom attributes). Of course will need a heavy rebase first

jaycarlton:
Thanks. I'm planning to do an initial PR that re-homes the comments we want to keep, then do this. I'll open a new PR for that with a cleaner, fresher history.

jaycarlton:
New work on this is happening on a new branch, incorporating changes from https://github.com/all-of-us/workbench/pull/3206. Since the process is mostly automatic, and rebasing would be slightly precarious with this kind of operation, I think that's the safest thing to do in this case.

PR #2972, opened 2020-01-06
[risk=no][RW-4238]Update yarn.lock for Node 12 compatibility on Mac

jmthibault79:
Description:
`yarn install` failed on my Mac after upgrading to Node 12.

Problem and Solution:
The transitive dependency `fsevents` added Node 12 compatibility in version 1.2.9, and we call for an earlier version in yarn.lock.  I edited `yarn.lock` to remove the entry for the older fsevents and ran `yarn install` to upgrade.  That resolved the issue.

Previous attempts:
TIL about the "resolutions" section of packages.json which helps force version numbers of transitive dependencies.  Ensuring that fsevents would not attempt to install a version before 1.2.8 resolved my problem.   However!  resolutions has an unfortunate bug: it can override the "optional" flag in some cases.  Since fsevents is a Mac-only optional build dependency, this had the side effect of preventing builds on Linux, which is what CircleCI runs.  

Steps to the solution:
https://github.com/fsevents/fsevents/issues/278
https://github.com/yarnpkg/yarn/issues/2051
https://gitmemory.com/issue/yarnpkg/yarn/6040/492049907
https://github.com/paulmillr/chokidar/issues/836

---
**PR checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ See https://docs.docker.com/docker-for-mac/#advanced for screenshots and instruc
 For local development, also install:
 
   * [yarn](https://yarnpkg.com/lang/en/docs/install/#mac-stable)
-  * [Node.js](https://nodejs.org/en/)
+  * [Node.js](https://nodejs.org/en/) >= 8.  Currently known to work up to 12.16.
   * [Java 8](https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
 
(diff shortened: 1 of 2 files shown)


calbach:
In general this file shouldn't be edited by hand so I want to run this myself to verify your change isn't going to get clobbered (at least on a different machine).

jmthibault79:
Good plan.  My own testing (single machine - Mac) showed that the current Node 8, 10, and 12 installs don't modify the file after this change.

PR #2973, opened 2020-01-06
[RW-3650][risk=no] Add label support to our core CheckBox class and add type-safety and unit tests.

gjuggler:
I ended up digging into the checkbox issue a bit more on Friday. I think this is the most balanced approach to fixing the issue. The crux of the matter is this: should a checkbox manage its own checked state, or should it always blindly render the value passed in via props?

After taking stock of existing usage in the Workbench, I realized that in most scenarios, it's fine to allow checkboxes manage their own state. Wrapping components can pass in the initial value, and listen for changes by providing an .onChange(checked) prop.

In one or two use cases, though, we seem to be expecting the checkbox to always render the value passed in via props. The wrapping component implements the toggling behavior by listening to onChange events, updating its own state, and using that state variable to pass into the CheckBox's 'props'. This pattern should be disfavored, since it causes wrapping components to effectively re-implement the checkbox toggling logic.

I also took the opportunity to add type-safe props and state to the CheckBox component, and a couple unit tests. It's a bit scary that, across our whole codebase, our core UI components have the least type-safety and unit testing. This makes it hard to safely improve or refactor our UI over time, so I'd like to (very incrementally) add these over time.

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/inputs.spec.tsx
+++ b/ui/src/app/components/inputs.spec.tsx
@@ -1,33 +1,44 @@
-import {EnzymeAdapter, mount, shallow, ShallowWrapper} from 'enzyme';
+import {
+  EnzymeAdapter,
+  mount,
+  ReactWrapper,
+  shallow,
+  ShallowWrapper,
+} from 'enzyme';
 import * as React from 'react';
 
 import {CheckBox} from './inputs';
 
-function findInput(wrapper: ShallowWrapper): ShallowWrapper {
+function findInput(wrapper: (ShallowWrapper|ReactWrapper)): (ShallowWrapper|ReactWrapper) {
   return wrapper.find('input[type="checkbox"]').first();
 }
 
-function clickCheckbox(wrapper: ShallowWrapper) {
+function clickCheckbox(wrapper: (ShallowWrapper|ReactWrapper)) {
   const currentChecked = findInput(wrapper).prop('checked');
   findInput(wrapper).simulate('change', {target: {checked: !currentChecked}});
 }
 
 describe('inputs', () => {
-  it('handles click with manageOwnState', () => {
+  it('click causes DOM checked state to change', () => {
     // When the checkbox manages its own state, a click should cause the input
     // to change.
-    const wrapper = shallow(<CheckBox label='asdf' manageOwnState={true} />);
+    const wrapper = shallow(<CheckBox label='asdf' />);
     expect(findInput(wrapper).prop('checked')).toEqual(false);
     clickCheckbox(wrapper);
     wrapper.update();
     expect(findInput(wrapper).prop('checked')).toEqual(true);
   });
 
+  it('uses "checked" to set initial state', () => {
+    const wrapper = shallow(<CheckBox checked={true} />);
+    expect(findInput(wrapper).prop('checked')).toEqual(true);
+  });
+
   it('uses props only when manageOwnState=false', () => {
-    // This is the more common pattern used throughout the RW codebase, where
+    // This pattern is disfavored but still exists in the RW codebase, where
     // the caller of CheckBox provides "checked" as a prop and handles updates
     // by reacting to onChange callbacks.
-    const wrapper = shallow(<CheckBox checked={true}/>);
+    const wrapper = mount(<CheckBox checked={true} manageOwnState={false}/>);
     expect(findInput(wrapper).prop('checked')).toEqual(true);
     // A user click shouldn't directly affect the checked state.
     clickCheckbox(wrapper);
--- a/ui/src/app/components/inputs.spec.tsx
+++ b/ui/src/app/components/inputs.spec.tsx
@@ -1,0 +1,56 @@
+import {EnzymeAdapter, mount, shallow, ShallowWrapper} from 'enzyme';
+import * as React from 'react';
+
+import {CheckBox} from './inputs';
+
+function findInput(wrapper: ShallowWrapper): ShallowWrapper {
+  return wrapper.find('input[type="checkbox"]').first();
+}
+
+function clickCheckbox(wrapper: ShallowWrapper) {
+  const currentChecked = findInput(wrapper).prop('checked');
+  findInput(wrapper).simulate('change', {target: {checked: !currentChecked}});
+}
+
+describe('inputs', () => {
+  it('handles click with manageOwnState', () => {
+    // When the checkbox manages its own state, a click should cause the input
+    // to change.
+    const wrapper = shallow(<CheckBox label='asdf' manageOwnState={true} />);
+    expect(findInput(wrapper).prop('checked')).toEqual(false);
+    clickCheckbox(wrapper);
+    wrapper.update();
+    expect(findInput(wrapper).prop('checked')).toEqual(true);
+  });
+
+  it('uses props only when manageOwnState=false', () => {
+    // This is the more common pattern used throughout the RW codebase, where
+    // the caller of CheckBox provides "checked" as a prop and handles updates
+    // by reacting to onChange callbacks.
+    const wrapper = shallow(<CheckBox checked={true}/>);
+    expect(findInput(wrapper).prop('checked')).toEqual(true);
+    // A user click shouldn't directly affect the checked state.
+    clickCheckbox(wrapper);
+    expect(findInput(wrapper).prop('checked')).toEqual(true);
+    // Changing the props will change the checked state.
+    wrapper.setProps({checked: false} as any);
+    expect(findInput(wrapper).prop('checked')).toEqual(false);
+  });
+
+  it('calls onChange with target checked state', () => {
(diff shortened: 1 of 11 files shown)


ui/src/app/components/inputs.tsx
```diff
@@ -119,64 +133,87 @@ export const RadioButton = ({ onChange, ...props }) => {
...
+  id?: string;
+  // An optional label to show alongside the input. Can be a plain string or
+  // any React node.
+  label?: React.ReactNode;
+  // Styles to apply to the label element. Only relevant when label is non-null.
+  labelStyle?: React.CSSProperties;
+  // Indicates whether the CheckBox should be responsible for managing its own
+  // state. When false, the HTML input will always be rendered with the value of
+  // the passed 'checked' prop.
+  manageOwnState: boolean;
```
als364: Should we file a ticket to deprecate this?

ui/src/app/components/inputs.tsx
```diff
@@ -119,64 +133,87 @@ export const RadioButton = ({ onChange, ...props }) => {
...
+    const {
+      checked, disabled, label, labelStyle, onChange, manageOwnState, style, wrapperStyle,
+      ...otherProps
+    } = this.props;
+    const uniqueId = fp.uniqueId('checkbox');
+
+    const maybeDisabledOverrides = disabled ? styles.disabledStyle : {};
+
+    const input = <input
+      id={uniqueId}
```
als364: not `props.id ? props.id : uniqueId`?

ui/src/app/components/tables.tsx
```diff
@@ -20,14 +20,14 @@ export const TwoColPaddedTable = ({style = {}, header = false, headerLeft = '',
   headerRight = '', cellWidth = {left: '50%', right: '50%'}, contentLeft, contentRight}) => {
   return <FlexColumn style={{...style}}>
     {header &&
-      <FlexRow style={{height: '100%'}}>
-        <PaddedTableCell left={true} leftWidth={cellWidth.left}
+      <FlexRow key='header' style={{height: '100%'}}>
```
als364: this seems unrelated. Did this miss a linter somewhere?

ui/src/app/pages/analysis/confirm-playground-mode-modal.tsx
```diff
@@ -27,10 +27,9 @@ export class ConfirmPlaygroundModeModal extends React.Component<Props, State> {
     };
   }
 
-  toggleChecked() {
-    const newState = !this.state.checked;
-    this.setState({checked: newState});
-    Cookies.set(ConfirmPlaygroundModeModal.DO_NOT_SHOW_AGAIN, String(newState));
+  toggleChecked(checked: boolean) {
+    this.setState({checked: checked});
```
als364: shouldn't this be `checked: !checked`?

als364: or does this happen after the actual checked state changes? if that's the case we should change the name of this function since it is no longer toggling checked

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -110,6 +102,7 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
     // Toggle Includes removes the element if it already exist and adds if not
     const attributeList = toggleIncludes(value, this.state.profile.demographicSurvey[attribute]);
     this.setState(fp.set(['profile', 'demographicSurvey', attribute], attributeList));
+    console.log(this.state);
```
als364: extraneous

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -521,7 +532,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
     }
 
     specificPopulationCheckboxSelected(populationEnum): boolean {
-      return fp.includes(populationEnum, this.state.workspace.researchPurpose.populationDetails);
+      return !!fp.includes(populationEnum, this.state.workspace.researchPurpose.populationDetails);
```
als364: uh oh, what probably hilarious bug made the boolifier necessary?

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -707,13 +718,14 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+          >
           <CheckBox
-                 style={{height: '.66667rem', marginRight: '.31667rem', marginTop: '1.2rem'}}
-          onChange={v => this.setState({cloneUserRole: v})}/>
-          <WorkspaceEditSection header='Copy Original workspace Collaborators'
-            description='Share cloned workspace with same collaborators'/>
-        </FlexRow>
+            style={styles.checkboxStyle}
+            label='Share workspace with the same set of collaborators'
+            labelStyle={styles.text}
```
als364: this results in the same style as the original styling? even with no margins set?

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -707,13 +718,14 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
+        <WorkspaceEditSection header='Options for duplicate workspace'
+          >
           <CheckBox
-                 style={{height: '.66667rem', marginRight: '.31667rem', marginTop: '1.2rem'}}
-          onChange={v => this.setState({cloneUserRole: v})}/>
-          <WorkspaceEditSection header='Copy Original workspace Collaborators'
-            description='Share cloned workspace with same collaborators'/>
-        </FlexRow>
+            style={styles.checkboxStyle}
+            label='Share workspace with the same set of collaborators'
```
als364: This is better wording, thank you

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -744,6 +757,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                 .map((rp, i) =>
                   <WorkspaceCategory shortDescription={rp.shortDescription}
                     longDescription={rp.longDescription} key={i}
+                                     uniqueId={rp.uniqueId}
```
als364: indentation is off here.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -734,6 +746,7 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
                 .map((rp, i) =>
                   <WorkspaceCategory shortDescription={rp.shortDescription} key={i}
                     longDescription={rp.longDescription}
+                     uniqueId={rp.uniqueId}
```
als364: indentation is off here.

ui/src/app/components/inputs.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import {CheckBox} from './inputs';
+
+function findInput(wrapper: (ShallowWrapper|ReactWrapper)): (ShallowWrapper|ReactWrapper) {
+  return wrapper.find('input[type="checkbox"]').first();
+}
+
+function clickCheckbox(wrapper: (ShallowWrapper|ReactWrapper)) {
+  const currentChecked = findInput(wrapper).prop('checked');
+  findInput(wrapper).simulate('change', {target: {checked: !currentChecked}});
```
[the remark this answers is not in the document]
als364: Boooo

This code feels like we're just testing that 1=1.

ui/src/app/pages/workspace/workspace-edit.tsx
als364: I do like the appearance of this better than the old appearance, including the larger checkbox.

ui/src/app/components/inputs.spec.tsx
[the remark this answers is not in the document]
gjuggler: Yeah... I'm not thrilled about it, but (1) it demonstrates expected usage and does test a modicum of the state-management flow, and (2) stupid tests are still better than null tests, esp. in a core component library like this :)

I'll probably poke at this a bit more in the future... stuff like this bugs me, as I pretty strongly believe UI stuff should be tested as realistically as possible, e.g. simulating true clicks on the DOM. But that does seem to go somewhat against the grain of typical React unit test patterns. Maybe the real answer is, we have defense in depth and this kind of stuff will get caught at an end-to-end test stage.

gjuggler:
Hi @dolbeew – I'd like to get your eyes on these changes, in case I'm missing something in how the checkboxes are working in the sidebar-component. And also more generally as a heads up re: the change to how checkboxes work by default.

PR #2978, opened 2020-01-07
[RW-4226][RISK=no] Use descriptive, unique-per-instance MonitoredResource specification

jaycarlton:
We need to use the `generic_node` [MonitoredResource](https://cloud.google.com/logging/docs/api/v2/resource-list) type since we run on AppEngine and can have multiple nodes writing at once.

Since we don't have a namespace, I chose the short name of the environment. For the instanceId, sometimes we get an exception saying it's not available, and I just use a GUID for that. This hopefully only happens on local, but we're more concerned with avoiding hitting the rate limits when we have escalated instance counts than with debugging performance of particular instances at the moment.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
+++ b/api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
@@ -3,12 +3,11 @@ package org.pmiops.workbench.monitoring;
 import com.google.api.MonitoredResource;
 import com.google.appengine.api.modules.ModulesException;
 import com.google.appengine.api.modules.ModulesService;
+import com.google.common.annotations.VisibleForTesting;
 import io.opencensus.exporter.stats.stackdriver.StackdriverStatsConfiguration;
 import io.opencensus.exporter.stats.stackdriver.StackdriverStatsExporter;
 import java.io.IOException;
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -34,23 +49,68 @@ public StackdriverStatsExporterService(Provider<WorkbenchConfig> workbenchConfig
...
+
+    return resultBuilder.build();
+  }
+
+  private String getProjectId() {
+    return workbenchConfigProvider.get().server.projectId;
+  }
+
+  private String getLocation() {
+    return workbenchConfigProvider.get().server.appEngineLocationId;
```
jmthibault79: Is there a plan to have this return the actual location in the future?  Please add a comment.

jaycarlton: Plan would be to have this set in the config at deploy time based on a `gcloud` command. It's app-wide.

Can't add comments to the JSON file tho.

jmthibault79: OK

PR #2981, opened 2020-01-08
quick demo of Swagger extension to support internal anchored comments

jaycarlton:
Looking for ways to better support free-form comments in YAML when processing by various tools. Since the `#` comments are always dropped unceremoniously in the first phase, what we really want IMO is a syntactic comment that doesn't show up in Swagger UI or affect any of the processing, but that will still hang onto its roots.

The Swagger Extension syntax supports this kind of use case. Here, I've added an `x-aou-note` extension for these type of internal comments for other API developers (but not consumers).

Some of our comments are really for API consumers as well, and can go in the public description where available.

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -26,7 +26,7 @@ info:
     name: "BSD"
 host: "api.pmi-ops.org"
 securityDefinitions:
-#  x-aou-note: |
+# (Can't put an extension here unfortunately)
 #    Establish the fact that *some endpoints* are OAuth protected
 #    by defining an `aou_oauth` security mode, which we'll assing
 #    to any protected endpoints below.
@@ -3844,6 +3844,7 @@ definitions:
     enum: [ FREE_TIER, USER_PROVIDED ]
 
   BillingProjectBufferStatus:
+    x-aou-note: This is the status of a billing project in a buffer.
     type: object
     properties:
       bufferSize:
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -1,14 +1,20 @@
-# This file is for endpoints we do NOT intend to generate client libraries for
-# and advertise to notebook developers. Endpoints we DO want notebook developers to use should
-# go in client_api.yaml.
-
-# This file references parameters, paths, and definitions specified in
-# client_api.yaml; the two files get merged together into a merged.yaml file at build time.
-
-# If validation fails, gradle:generateApi fails claiming this file does not exist.
-# For separate validation (with some false positives), do:
-#     ./project.rb validate-swagger
 swagger: '2.0'
+x-aou-note: >
+  This file is for endpoints we do NOT intend to generate client libraries for
+  and advertise to notebook developers. Endpoints we DO want notebook developers to use should
+  go in client_api.yaml.
+
+  This file references parameters, paths, and definitions specified in
+  client_api.yaml; the two files get merged together into a merged.yaml file at build time.
+
+  If validation fails, gradle:generateApi fails claiming this file does not exist.
+  For separate validation (with some false positives), do:
+     ./project.rb validate-swagger
+
+  Throughout, we use integer/int64 in preference to string/date-time because Swagger's
+  date formatting is inconsistent between server and client. Time values are stored as
+  milliseconds since the UNIX epoch.
+
 info:
   version: "0.1.0"
   title: "AllOfUs Workbench API"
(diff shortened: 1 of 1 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -20,30 +26,36 @@ info:
     name: "BSD"
 host: "api.pmi-ops.org"
 securityDefinitions:
-  # Establish the fact that *some endpoints* are OAuth protected
-  # by defining an `aou_oauth` security mode, which we'll assing
-  # to any protected endpoints below.
+#  x-aou-note: |
```
jaycarlton: You can't put the extensions just anywhere, unfortunately. See https://swagger.io/docs/specification/2-0/swagger-extensions/ for details

api/src/main/resources/workbench.yaml
```diff
@@ -20,30 +26,36 @@ info:
...
   - "application/json"
-# Establish the fact that all endpoints are protected: this annotation
-# ensures that client libraries know to send bearer tokens when calling
 security:
   - aou_oauth: []
-
-# Throughout, we use integer/int64 in preference to string/date-time because Swagger's
-# date formatting is inconsistent between server and client. Time values are stored as
-# milliseconds since the UNIX epoch.
+tags:
```
jaycarlton: This is a global tags definition, which lets you put limited metadata about your tags.

We could do this for all the controller tag names too, but I'm not sure of the advantage, since I don't believe swagger-codegen would use the info.

api/src/main/resources/workbench.yaml
```diff
@@ -20,30 +26,36 @@ info:
...
+  - name: cron
+    description: |
```
jaycarlton: Arguably this is an internal note and should move to the `x-aou-note` section or be reworded for the benefit of API consumers.

jaycarlton:
@gjuggler
> I don't think the goal of preserving repeated round-trip integrity through automated tools needs to be a critical requirement for our OpenAPI v3 transition. Personally, I haven't found editing these YAMLs within IntelliJ to be a huge drain on velocity. If we're planning to require the dev team to use a new tool when making any API modifications or additions, there's some burden of proof to demonstrate that the end result is worth the investment or potential friction, compared to manual editing + strong linting & validation, as we do with code.

I don't think we need to dictate a particular editor, but was looking for an example to break a layer of abstraction off between the text file that describes a yaml document (a "yaml file") and the document itself, and to warn folks that certain choices that give equivalent documents (like text styles and comments) wouldn't survive a round trip. You can replace "editor" with "linter" or "translator" or "splitter/joiner" and the same caveat applies. I don't want to require the use of any particular editor, but to enable us to use a different one if it helps us with OpenAPI syntax and semantics (instead of just yaml syntax), but to enable them by use of structured metadata in place of comments and hopefully a linter/canonicalization program that helps keep diffs in source control to a minimum.

One open question, which is already coming up in the single-yaml-file transition, is how hard we should try to preserve the existing history/blame view. I could probably do that for the workbench.yaml file (but not the others) if I want to switch to merging by hand instead of using the Ruby script.

For OpenAPI 3, I'm just looking for the option to use some kind of automatic tool, whether it be a third party translator or something I write. In either case, I would want to work at a YAML-level of abstraction at least (if not a Swagger-level), which means text-level things are a casualty.

I like the desire for "strong linting", though I think any swagger linter we use will be based on a yaml processor (or swagger processor that's based on a yaml processor).  If we want to ensure structural consistency like alphabetized paths, or sections in a certain order, then the best way I know to do that would be with an automatic tool that loads, validates, and rewrites the entries in that prescribed order.

> I strongly suspect we'll want to break up the merged YAML file again once we've moved to OpenAPI v3. This will help preserve straightforward organization, navigation, & editing of logically-related pieces of our API over time. If you're working towards an explicit goal to keep everything in one file long-term, let's discuss further in a team meeting.

I'm definitely supportive of that. The only caveat here is that we might need a pipeline that re-merges things via their external references in order to take advantage of some prebuilt tools. The inclusion syntax is a Swagger feature and not a Yaml feature, so if we split a Swagger definition (for a _single API_) across multiple files, we need to keep that in mind.

If you are talking about splitting into multiple API definitions, with one file each, then that's not an issue. 

Ultimately I want to strike a good balance between good (even improved) readability and automatic process-ability.

> Is it possible to add extension fields to object definitions? It seemed like maybe not from my reading of the extension docs, but I'm not confident on that.

Looks like it is, at least for a definition object. Though it seems like a design smell in most cases if you have something important to put in a note that can't go in the dedicated `description` field.

PR #2984, opened 2020-01-08
[RW-4228][Risk=no] Add cluster resources to static hosted files

blrubenstein:
Description:
This adds the cluster resources to statically hosted files, to get around gcs enforcement. To test, go to srubenst-dot-api-dot-all-of-us-workbench-test.appspot.com/static/activity-checker-extension.js

---
**PR checklist**


api/src/main/webapp/static/activity-checker-extension.js
```diff
@@ -0,0 +1,45 @@
+// Sends a message to the parent frame every second that user activity is detected
```
calbach: We should either remove the files from /cluster-resources, or copy the files here as part of the build step and .gitignore these. Currently the file is duplicated in source.

api/cluster-resources/build.rb
```diff
@@ -27,6 +27,9 @@ def build_snippets_menu()
 
   FileUtils.mkdir_p "generated"
   File.write("generated/aou-snippets-menu.js", tmpl)
```
calbach: I don't think this one will be needed anymore.

api/src/main/webapp/WEB-INF/appengine-web.xml.template
```diff
@@ -26,6 +25,15 @@
     <handler file="server_unavailable.html" />
   </static-error-handlers>
 
+  <static-files>
+    <include path="static/activity-checker-extension.js" />
```
calbach: I just remembered we should probably include this set of http headers: https://github.com/all-of-us/workbench/blob/e70ef354fe09414f273058d6c805e1bcc3212ab5/ui/app.yaml#L11-L15

This would probably push us towards pulling in the whole static/ directory to avoid repeating these on every include element.

api/src/main/webapp/static/activity-checker-extension.js
blrubenstein: I can do the copy. There was a requirement to keep them served to GCS, so I didn't want to mess with that code.

api/cluster-resources/build.rb
blrubenstein: The ticket states:
```
Note: cluster-resources must continue to be pushed to GCS following this change. Teardown of this approach will follow later.
```

I wanted to do minimal changes to that deploy for now, and refactor when we removed that as a requirement. I think we still need this if we want to push that file to GCS.

api/libproject/devstart.rb
```diff
@@ -1503,6 +1503,7 @@ def deploy_gcs_artifacts(cmd_name, args)
 
   Dir.chdir("cluster-resources") do
     common.run_inline(%W{./build.rb build-snippets-menu})
+    common.run_inline(%W{./build.rb generate-static-files})
```
calbach: I would move this into `deploy_app` instead, somewhere around L1579 (before appengineStage). This copy needs to happen every time before we build the GAE app for deployment, so it should be more closely coupled to that step

api/src/main/webapp/static/README.md
```diff
@@ -0,0 +1,8 @@
+#Static Files in AppEngine
```
calbach: Thanks for adding the README

api/src/main/webapp/WEB-INF/appengine-web.xml.template
```diff
@@ -26,6 +25,20 @@
     <handler file="server_unavailable.html" />
   </static-error-handlers>
 
+  <static-files>
+    <include path="static/*">
```
calbach: Can you verify that this is picking up aou-snippets-menu.js? Pretty sure you need `**` to pickup subdirs. Alternative fix would be to just to not mirror the generated/ dir there, just copy it straight into /static. That seems fine, if not correct to me, since all of these static files are copied in during the build time anyways - the fact that this one file is generated is no longer relevant.

api/src/main/webapp/static/README.md
```diff
@@ -0,0 +1,8 @@
...
+We host a number of public static files in GAE for cluster creation. These files are copied at build time from our cluster resources directory. Eventually (when we no longer support bucket extensions), these files should move into this directory and we should no longer gitignore them from here.
+
```
calbach: Since we're globbig the directory, won't this be served as a static resource? If so, I might suggest moving this to either api/doc or cluster-resources/README.md

.gitignore
```diff
@@ -20,6 +20,11 @@ cron-emulator-err.log
 /api/src/main/webapp/WEB-INF/cron.yaml
 /api/src/main/webapp/WEB-INF/sa-key.json
 /api/src/main/webapp/WEB-INF/gsuite-admin-sa.json
+/api/src/main/webapp/static/generated/*
+/api/src/main/webapp/static/activity-checker-extension.js
```
calbach: nit: if you move the README out, you can just ignore the whole directory

api/src/main/webapp/WEB-INF/appengine-web.xml.template
blrubenstein: You're right, it is not making it. I will ** it anyway, since we should allow subdirectories, but also get rid of the generated subdir

api/src/main/webapp/static/README.md
blrubenstein: This will be served as a static resource, but I think that's probably okay? I'd prefer to have it where the files are. Unless you have strong feelings against it being served? (I can also add it as a specific exclude?)

calbach: Good idea - exclude sounds reasonable

PR #2990, opened 2020-01-09
[RW-2840][risk=no] Use AppIdentityService to generate service-backed credentials

gjuggler:
After #2962 merged, we started seeing 401 errors in test when creating a new workspace.

I tracked this down to an issue where the newer GoogleCredentials library doesn't do what I was expecting when calling GoogleCredentials.getApplicationDefault(). See the new comments on the ServiceAccounts method which describes the issue in some detail.

This PR essentially brings back the logic to switch how we generate scoped service credentials based on the operating environment: if App Engine, use the AppIdentityService; if development, use ADCs. This method is called by a couple config classes for APIs which use our service identity to call downstream APIs.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
+++ b/api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
@@ -1,5 +1,9 @@
 package org.pmiops.workbench.auth;
 
+import com.google.appengine.api.appidentity.AppIdentityService;
+import com.google.appengine.api.appidentity.AppIdentityServiceFactory;
+import com.google.appengine.api.utils.SystemProperty;
+import com.google.auth.appengine.AppEngineCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -13,14 +17,46 @@
...
+      AppIdentityService appIdentityService = AppIdentityServiceFactory.getAppIdentityService();
+      return AppEngineCredentials.newBuilder()
+          .setScopes(scopes)
+          .setAppIdentityService(appIdentityService)
+          .build();
+    }
+  }
+
   /**
    * Retrieves an access token with the specified set of scopes derived from a given set of Google
```
jmthibault79: comment is no longer accurate

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
```diff
@@ -13,14 +17,46 @@
...
+   * https://github.com/googleapis/google-auth-library-java/issues/172 for reference; this seems to
+   * be a common pain point for users of the com.google.auth.oauth2 library.
+   *
+   * @param scopes
+   * @return
+   * @throws IOException
+   */
+  private static GoogleCredentials getScopedServiceCredentials(List<String> scopes)
+      throws IOException {
+    if (SystemProperty.environment.value().equals(SystemProperty.Environment.Value.Development)) {
```
jmthibault79: Could use the just-removed `WorkbenchEnvironment` here if it still existed.  Would it make sense to bring it back?

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
gjuggler: Thanks – updated.

api/src/main/java/org/pmiops/workbench/auth/ServiceAccounts.java
gjuggler: Yeah, I thought about that... I'm not convinced that a separate dependency-injected class is worth it for the single usage instance here. If there's no other obvious advantage that I'm missing, I'm inclined to keep the "am I in dev" check local here.

[Update] as a last-ditch effort, I tried seeing if the AppIdentityService branch would work correctly even when running locally. I tested this out locally and it does work fine (I suspect that somewhere under the hood there is a fallback call to GoogleCredentials.getApplicationDefault), but I'm not confident enough to push out that change. I'd rather leave as-is for now and possibly follow-up later with that simplification.

PR #2992, opened 2020-01-10
[RW-2253] [risk=low] Add inbound API to handle SumoLogic egress alerts

gjuggler:
This PR implements a simple version of the API described in the [design doc](https://docs.google.com/document/d/133jYAlQIJMT_TqbR49IXFEerMgWjd2Lly2-BuUu5_ZM/edit#), with the primary goal of logging an action audit event for every inbound egress alert.

The inbound API is controlled by a per-environment feature flag. For now it is only enabled in test.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditor.kt
```diff
@@ -0,0 +1,18 @@
+package org.pmiops.workbench.actionaudit.auditors
+
+import org.pmiops.workbench.model.EgressEvent
+import org.pmiops.workbench.model.EgressEventRequest
+
+interface SumoLogicAuditor {
```
jaycarlton: FWIW, this interface and implementation don't need to be in Kotlin. Java classes can implement Kotlin interfaces and vice-versa. See `UserServiceAuditor.java`.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+        actionAuditService.send(ActionAuditEvent(
+                timestamp = clock.millis(),
+                actionId = actionIdProvider.get(),
+                actionType = ActionType.HIGH_EGRESS,
+                agentType = AgentType.USER,
+                agentId = agentId,
+                agentEmailMaybe = agentEmail,
+                targetType = TargetType.WORKSPACE,
+                targetIdMaybe = dbWorkspace.workspaceId,
+                // TODO: Maybe what would fit best here would be a "comment" field on the AcionAuditEvent and
```
jaycarlton: There are a couple of ways to do this, depending on what level of abstraction you want.

If you want commenting to be an action that someone can take, then we can make an ActionType of COMMENT, and apply it to any of the target types. We could even have a target type of ACTION and use the action's ID as the TARGET_ID.

I'm hesitant to make a new column for comment, as it would mean different rows under the same Action could have different comments. That's not a problem per se, but it's difficult to enforce a consistent usage pattern when lots of folks are using the system, and querying might be icky.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -272,7 +272,13 @@
...
   }
 
-  private static String clusterNameForUser(DbUser user) {
+  /**
+   * Returns a name for the VM / cluster to be created for a given user in the workspace.
+   *
+   * @param user
+   * @return
+   */
+  public static String clusterNameForUser(DbUser user) {
```
jaycarlton: I'd externalize the prefix `all-of-us-`. I don't know from this statement what its scope is.

Also, one pet peeve I have is that if we're using hyphens to delineate subparts of an identifier, then it should be `allofus-`. Likely way too late.

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Autowired
+  SumoLogicController(SumoLogicAuditor sumoLogicAuditor, CloudStorageService cloudStorageService) {
+    this.sumoLogicAuditor = sumoLogicAuditor;
+    this.cloudStorageService = cloudStorageService;
+  }
+
+  @Override
+  public ResponseEntity<EmptyResponse> logEgressEvent(
+      String X_API_KEY, EgressEventRequest request) {
```
jaycarlton: why are we screaming the parameter name?

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditor.kt
gjuggler: Woops – let's keep this between you and me :)

I took it as a chance to try and get a bit more familiar with Kotlin before we need to make a team decision on whether or not to adopt it more widely. That's good to know it's not as infectious as I might have worried.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorImpl.kt
gjuggler: Thanks – I guess the most pragmatic solution will be to call it an ActionType and tie two rows together by a single Action ID: one for the high-egress event, and one for the comment.

I'd feel a little awkward to have to send two events for this, even though it's definitely just one event-plus-a-log-message. But I can wrap it in a helper function on the ActionAuditService, so it probably won't affect usage too much.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
gjuggler: What do you mean by externalize? Bump to a constant in this file, out to a separate class, or all the way out to WorkbenchConfig?

I'm willing to find a separate class to home it in, to at least protect against circular dependencies.

Anything more probably isn't worth the effort. We haven't changed this value for the ~lifetime of the project, so I don't think it's a particularly sensitive constant.

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
gjuggler: Swagger-codegen is choosing the parameter naming here – it's based on what I used in the YAML file. It felt more appropriate to use the traditional UPPERCASE_HEADER casing over there, so... meh.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorImpl.kt
gjuggler: PTAL at the [latest commit](https://github.com/all-of-us/workbench/pull/2992/commits/c5ccc982bdc8c4bb7b790470420fbda199e29007) – I took a stab at something like what I described above.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditService.kt
```diff
@@ -3,9 +3,18 @@ package org.pmiops.workbench.actionaudit
...
+    fun send(event: ActionAuditEvent) {
+        send(setOf(event))
+    }
 
     fun send(events: Collection<ActionAuditEvent>)
 
+    /**
+     * Sends a collection of events plus an event with ActionType.COMMENT and the given comment
+     * string as the newValueMaybe. The COMMENT action derives its other properties from the
+     * first event from the original collection.
```
jaycarlton: I see where you're going, and how something like this could be useful, but there are a couple of things that are of some concern:

1. The model doesn't really support multiple action types per actionId.
2. A utility like this would naturally belong to the `AuditActionEvent.Builder` class, not this service.
3. It's ambiguous to me when this would be used. If we're capturing administrator comments, then that makes sense to me as a dedicated `TargetProperty` that can be added to whatever the administrator (agent) is doing (action). So from that standpoint, we don't need a global comment.
4. If it's for developers to add arbitrary strings, then things can start to get messy in a hurry. It will be tempting to just stuff things into a comment when you don't know where they would logically go in the schema. That's better than not having the information, but one of my original requirements was for a high degree of relational query-ability.
5. If we want to support stand-alone comment actions (independent of other events), then that's another action type (as you have here) with a target of whatever context the admin is looking at (user, workspace, etc).
6. If we decide to migrate some date that's in comments into a more structured form, that's really tough to do compared to more structured data.
7. It's not stable to take the first element from a `Collection`, since they're not ordered.

gjuggler: Chatted offline – check out the latest commit for a different approach to this.

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/EgressEventTargetProperty.kt
```diff
@@ -0,0 +1,27 @@
...
+enum class EgressEventTargetProperty
+constructor(
+    override val propertyName: String,
+    override val extractor: (EgressEvent) -> String?
+) : ModelBackedTargetProperty<EgressEvent> {
+    EGRESS_MIB("egress_mib",
+            { it.egressMib?.toString() }),
+    TIME_WINDOW_START("time_window_start",
+            {
+                if (it.timeWindowStart != null)
```
jaycarlton: nit: IYI You can do this wiht a `.?` and `?:` chain. https://medium.com/tompee/idiomatic-kotlin-elvis-operator-d6f5639a5130

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/EgressEventTargetProperty.kt
```diff
@@ -0,0 +1,27 @@
...
+            }),
+    TIME_WINDOW_DURATION("time_window_duration",
+            { it.timeWindowDuration?.toString() }),
+    VM_NAME("vm_name", { it.vmName });
+}
+
+enum class EgressEventCommentTargetProperty
+constructor(override val propertyName: String) : SimpleTargetProperty {
+    COMMENT("comment");
+}
```
jaycarlton: nit EOF newline (spotless will catch I think)

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/EgressEventTargetProperty.kt
```diff
@@ -0,0 +1,27 @@
...
+                else "[no time]"
+            }),
+    TIME_WINDOW_DURATION("time_window_duration",
+            { it.timeWindowDuration?.toString() }),
+    VM_NAME("vm_name", { it.vmName });
+}
+
+enum class EgressEventCommentTargetProperty
+constructor(override val propertyName: String) : SimpleTargetProperty {
+    COMMENT("comment");
```
jaycarlton: Thanks for structuring it this way.

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
```diff
@@ -0,0 +1,113 @@
...
+import org.pmiops.workbench.model.EgressEventRequest;
+import org.pmiops.workbench.model.EmptyResponse;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class SumoLogicController implements SumoLogicApiDelegate {
+
+  public static final String SUMOLOGIC_KEY_FILENAME = "inbound-sumologic-keys.txt";
```
jaycarlton: I'd make this a per-env property most likely. Can we simulate the events on Test and Local?

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
```diff
@@ -0,0 +1,113 @@
...
+      Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.sumoLogicAuditor = sumoLogicAuditor;
+    this.cloudStorageService = cloudStorageService;
+    this.workbenchConfigProvider = workbenchConfigProvider;
+  }
+
+  @Override
+  public ResponseEntity<EmptyResponse> logEgressEvent(
+      String X_API_KEY, EgressEventRequest request) {
+    if (!workbenchConfigProvider.get().featureFlags.enableSumoLogicEventHandling) {
```
jaycarlton: Should this just be a no-op with a warning?

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
```diff
@@ -0,0 +1,113 @@
...
+      throw new BadRequestException("SumoLogic event handling is disabled in this environment.");
+    }
+
+    authorizeRequest(X_API_KEY, request);
+
+    try {
+      // The "eventsJsonArray" field is a JSON-formatted array of EgressEvent JSON objects. Parse
+      // this
```
jaycarlton: nit: check wrapping

api/src/main/java/org/pmiops/workbench/api/SumoLogicController.java
jaycarlton: Nvm, it's a controller method, so this makes sense.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/AuthDomainAuditorTest.java
```diff
@@ -16,67 +14,36 @@
...
 import org.springframework.boot.test.mock.mockito.MockBean;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
-import org.springframework.context.annotation.Scope;
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
 public class AuthDomainAuditorTest {
 
   private static final long USER_ID = 101L;
```
jaycarlton: 👏 Love the consolidation!

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorImpl.kt
```diff
@@ -0,0 +1,164 @@
...
+        if (egressEvent != null) {
+            val propertyValues = TargetPropertyExtractor.getPropertyValuesByName(
+                    EgressEventTargetProperty.values(), egressEvent)
+            events.addAll(propertyValues.map {
+                baseEvent.copy(
+                        targetPropertyMaybe = it.key,
+                        newValueMaybe = it.value)
+            })
+        }
+        if (comment != null) {
```
jaycarlton: Kotlin trick: If we had multiple threads, then in some cases `comment` could change value in the block when using `if` but not when using the operator `?.let()`.

So I believe you can do `comment?.let {
            events.add(baseEvent.copy(
                    targetPropertyMaybe = EgressEventCommentTargetProperty.COMMENT.propertyName,
                    newValueMaybe = comment
            ))
        }
`

https://discuss.kotlinlang.org/t/let-vs-if-not-null/3542/6

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -36,7 +36,7 @@ public CloudStorageServiceImpl(Provider<WorkbenchConfig> configProvider) {
 
   @Override
   public String readInvitationKey() {
-    return readCredentialsBucketString("invitation-key.txt");
+    return getCredentialsBucketString("invitation-key.txt");
```
jaycarlton: I'd probably make an enum with these values or store them in per-env configs. In principle we could have separate Moodles, etc.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -27,6 +27,10 @@
 
   DbWorkspace get(String ns, String firecloudName);
 
+  // Returns the requested workspace looked up by workspace namespace (aka billing project name).
+  // Only active workspaces are searched. Returns null if no active workspace is found.
+  DbWorkspace getByNamespace(String workspaceNamespace);
```
jaycarlton: Please return `Optional<DbWorkspace>` instead of using null.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -235,6 +235,12 @@ public DbWorkspace getRequired(String ns, String firecloudName) {
     return workspace;
   }
 
+  @Override
+  public DbWorkspace getByNamespace(String ns) {
```
jaycarlton: I believe you can return an optional here, if you do so in the Dao.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -500,8 +506,10 @@ public DbWorkspace getWorkspaceEnforceAccessLevelAndSetCdrVersion(
   }
 
   @Override
-  public List<UserRole> convertWorkspaceAclsToUserRoles(
-      Map<String, FirecloudWorkspaceAccessEntry> rolesMap) {
+  public List<UserRole> getFirecloudUserRoles(String workspaceNamespace, String firecloudName) {
+    Map<String, FirecloudWorkspaceAccessEntry> rolesMap =
+        getFirecloudWorkspaceAcls(workspaceNamespace, firecloudName);
+
```
jaycarlton: nit: UserRole could just define its own internal Comparator and implement Comparable so you don't need to do it below.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -500,8 +506,10 @@ public DbWorkspace getWorkspaceEnforceAccessLevelAndSetCdrVersion(
   }
 
   @Override
-  public List<UserRole> convertWorkspaceAclsToUserRoles(
-      Map<String, FirecloudWorkspaceAccessEntry> rolesMap) {
+  public List<UserRole> getFirecloudUserRoles(String workspaceNamespace, String firecloudName) {
+    Map<String, FirecloudWorkspaceAccessEntry> rolesMap =
```
jaycarlton: nit: it helps to name maps like `keyToValue`

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/ActionAuditTestConfig.java
```diff
@@ -0,0 +1,39 @@
...
+@Configuration
+public class ActionAuditTestConfig {
+
+  public static final Instant INSTANT = Instant.parse("2000-01-01T00:00:00.00Z");
+  public static final String ACTION_ID = "9095d2f9-8db2-46c3-8f8e-4f90a62b457f";
+  public static final long ADMINISTRATOR_USER_ID = 222L;
+  public static final String ADMINISTRATOR_EMAIL = "admin@aou.biz";
+
+  @Bean
+  public static DbUser getUser() {
```
jaycarlton: Might make sense to give this a `Name`, and have another one for a regular user.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/AuthDomainAuditorTest.java
```diff
@@ -16,67 +14,36 @@
...
-      administrator.setUsername(ADMINISTRATOR_EMAIL);
-      return administrator;
-    }
-
-    @Bean
-    public Clock getClock() {
-      return new FakeClock(INSTANT);
-    }
-  }
+  @Import({AuthDomainAuditorImpl.class, ActionAuditTestConfig.class})
```
jaycarlton: Oh, so that's how you do it. Neat.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/AuthDomainAuditorTest.java
```diff
@@ -16,67 +14,36 @@
...
-  private static final String ADMINISTRATOR_EMAIL = "admin@aou.biz";
-  private static final Instant INSTANT = Instant.parse("2000-01-01T00:00:00.00Z");
-  private static final String ACTION_ID = "9095d2f9-8db2-46c3-8f8e-4f90a62b457f";
-
-  @Autowired private ActionAuditService mockActionAuditService;
-
-  @Captor private ArgumentCaptor<ActionAuditEvent> eventCaptor;
 
   @Autowired private AuthDomainAuditor authDomainAuditAdapter;
+  @MockBean private ActionAuditService mockActionAuditService;
```
jaycarlton: Cool. I didn't know about this version. I'd been using `@Autowired` even for the mocks and putting a `@MockBean` on the config class.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+            .vmName(EGRESS_EVENT_VM_NAME)
+            .timeWindowStart(0l)
+            .egressMib(12.3));
+    verify(mockActionAuditService).send(eventsCaptor.capture());
+    Collection<ActionAuditEvent> events = eventsCaptor.getValue();
+
+    // Ensure all events have the expected set of constant fields.
+    assertThat(events.stream().map(event -> event.getAgentType()).collect(Collectors.toSet()))
+        .containsExactly(AgentType.USER);
+    assertThat(events.stream().map(event -> event.getActionType()).collect(Collectors.toSet()))
```
jaycarlton: I was going to say, you can make these a bit shorter and avoid collecting by using `allMatch()`. But then your assertion failure message wouldn't be as clear. So keep it this way.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        .containsExactly(ActionType.HIGH_EGRESS);
+    assertThat(events.stream().map(event -> event.getAgentId()).collect(Collectors.toSet()))
+        .containsExactly(USER_ID);
+    assertThat(events.stream().map(event -> event.getAgentEmailMaybe()).collect(Collectors.toSet()))
+        .containsExactly(USER_EMAIL);
+    assertThat(events.stream().map(event -> event.getTargetIdMaybe()).collect(Collectors.toSet()))
```
jaycarlton: You can use method references like `ActionAuditEvent::getEmailMaybe`.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+            .egressMib(12.3));
+    verify(mockActionAuditService).send(eventsCaptor.capture());
+    Collection<ActionAuditEvent> events = eventsCaptor.getValue();
+
+    // Ensure all events have the expected set of constant fields.
+    assertThat(events.stream().map(event -> event.getAgentType()).collect(Collectors.toSet()))
+        .containsExactly(AgentType.USER);
+    assertThat(events.stream().map(event -> event.getActionType()).collect(Collectors.toSet()))
+        .containsExactly(ActionType.HIGH_EGRESS);
+    assertThat(events.stream().map(event -> event.getAgentId()).collect(Collectors.toSet()))
```
jaycarlton: You're doing enough of this that it could be helpful to have an assertion helper funciton in the parent class like `assertContainsUniqueValue(Collection<T> objects, Function<T, U> extractor, U value);` I Could use that in a few places as well.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                            == EgressEventTargetProperty.EGRESS_MIB.getPropertyName())
+                .map(event -> event.getNewValueMaybe())
+                .collect(Collectors.toSet()))
+        .containsExactly("12.3");
+    assertThat(
+            events.stream()
+                .filter(
+                    event ->
+                        event.getTargetPropertyMaybe()
+                            == EgressEventTargetProperty.VM_NAME.getPropertyName())
```
jaycarlton: nit: for Java string comparisons, it's nice to put the known non-null string on the left so you can use `.equals()`

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        new EgressEvent().projectName(EGRESS_EVENT_PROJECT_NAME).vmName(EGRESS_EVENT_VM_NAME));
+    verify(mockActionAuditService).send(eventsCaptor.capture());
+    Collection<ActionAuditEvent> events = eventsCaptor.getValue();
+
+    // Some of the properties should be nulled out, since we can't identify the target workspace
+    // for the egress event.
+    assertThat(events.stream().map(event -> event.getAgentId()).collect(Collectors.toSet()))
+        .containsExactly(0);
+    assertThat(events.stream().map(event -> event.getAgentEmailMaybe()).collect(Collectors.toSet()))
+        .containsExactly((Object[]) null);
```
jaycarlton: This is kind of scary. I'd instead do a  `.filter(Objects::nonNull)` and assert that the set is empty.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                .findFirst()
+                .get())
+        .contains("Failed to find workspace");
+  }
+
+  @Test
+  public void testFailedParsing() {
+    // When the inbound request parsing fails, an event is logged at the system agent.
+    when(mockWorkspaceService.getByNamespace(WORKSPACE_NAMESPACE)).thenReturn(null);
+    sumoLogicAuditor.fireFailedToParseEgressEvent(new EgressEventRequest().eventsJsonArray("asdf"));
```
jaycarlton: nit: If the array isn't valid JSON, I would've expected a lower-level exception to be thrown first.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                    event ->
+                        event.getTargetPropertyMaybe()
+                            == EgressEventCommentTargetProperty.COMMENT.getPropertyName())
+                .map(event -> event.getNewValueMaybe())
+                .findFirst()
+                .get())
+        .contains("Received bad API key");
+  }
+
+  @TestConfiguration
```
jaycarlton: nit: Oh, I'm used to seeing these at the top.

api/src/test/java/org/pmiops/workbench/api/SumoLogicControllerTest.java
```diff
@@ -0,0 +1,95 @@
...
+  static class Configuration {}
+
+  @Before
+  public void setUp() throws JsonProcessingException {
+    event = new EgressEvent();
+    event.setProjectName("aou-rw-test-c7dec260");
+    event.setEgressMibThreshold(100.0);
+    event.setEgressMib(123.0);
+    event.setEnvironment(EgressEvent.EnvironmentEnum.TEST);
+    event.setTimeWindowDuration(300L);
```
jaycarlton: nit: I'd use a Duration object and then call `toSecionds()` or w/e.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/ActionAuditTestConfig.java
```diff
@@ -0,0 +1,39 @@
...
+    final DbUser administrator = new DbUser();
+    administrator.setUserId(ADMINISTRATOR_USER_ID);
+    administrator.setUsername(ADMINISTRATOR_EMAIL);
+    return administrator;
+  }
+
+  @Bean(name = "ACTION_ID")
+  @Scope(value = ConfigurableBeanFactory.SCOPE_PROTOTYPE)
```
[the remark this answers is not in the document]
gjuggler: Done.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/AuthDomainAuditorTest.java
gjuggler: Yeah, I think I only discovered this while digging into some Spring examples out on the internets. I think it's marginally cleaner this way.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: Yeah... I almost started doing that, but it only would have saved a "bit" of code, and would add indirection which has a cost on readability. I'll leave as-is, but I agree some helpers may be warranted if this grows in size.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: In this case, the lambda actually saves characters :)

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: Ah, I always forget about that. Done.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: Good idea – done.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: 🤷‍♂

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
gjuggler: Yeah I don't know why these ended up down here... moved back up.

api/src/test/java/org/pmiops/workbench/api/SumoLogicControllerTest.java
gjuggler: Heh... it reads as a bit repetitive, but it does add clarity. Done.

api/src/main/resources/workbench.yaml
```diff
@@ -4011,3 +4039,50 @@ definitions:
...
           A URL linking to an incident report where users can read more
         type: string
+
+  EgressEventRequest:
+    type: object
+    properties:
+      eventsJsonArray:
+        type: string
+        description: >
+          A JSON string containing an array of JSON-encoded egress events. See the EgressEvent
```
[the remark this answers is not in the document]
jaycarlton: Pretty much what I figured. Thanks.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/AuthDomainAuditorTest.java
jaycarlton: Yep, there's less duplication and you don't rely only on the name convention.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
jaycarlton: Except that your controller is passing it in internally, so this makes sense.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/SumoLogicAuditorTest.java
jaycarlton: But you gain an operator, and a scope, and lose at least half an abstraction level :/

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
```diff
@@ -10,5 +10,6 @@ enum class ActionType {
     DUPLICATE_FROM,
     DUPLICATE_TO,
     COLLABORATE,
-    DELETE
+    DELETE,
+    HIGH_EGRESS
```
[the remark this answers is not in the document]
jaycarlton: In reviewing this, I'm also reviewing hypothetical PRs in the future, and some of those are what's irking me. Not really fair to you I guess. I don't have the luxury of a real schema, so conventions are enforced by nagging.

What about `DETECT_EGRESS_VIOLATION`? It aligns narrowly with this anomaly type and is still a present tense verb that makes sense with the target. (For purity, `DETECT` would be the action and `VPC_EGRESS_VIOLATION` would be the target, but that won't work well because the workspace needs to be the target.)

The other consideration is how we treat any future system-detected security or policy flags against a user or workspace. If it's a separate action for each one, that's OK, but might mean more bespoke querying. In other words, by choosing an action that matches your classes, you're tipping the scale such that there are going to be more actions to manage, which many not be a net loss in cognitive overhead.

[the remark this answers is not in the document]
gjuggler: I'm willing to bite on the idea that consistency around verbs vs. nouns is important. How about DETECT_HIGH_EGRESS_EVENT?

This is edging towards too many word-tokens for comfort, but I think we'll survive it.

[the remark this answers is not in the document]
jaycarlton: That works, and balances out consistency with your other names.

PR #2994, opened 2020-01-10
Add admin delete clusters in project endpoint

blrubenstein:
Description:
This adds an endpoint which allows security admins to delete all clusters in a project. We went with delete because Leonardo doesn't currently allow for stopping clusters as a project owner, just deleting them.

Manual test on local included:
Deleting current user's cluster as that user.
Deleting another user's cluster as one user, in a workspace the user does not have access to. 

I would love some thoughts on the unit tests here, because I am unsure where the cost/benefit lands for amount of unit tests. (The one I wrote just basically verifies that it hits Leo)

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
@@ -1,0 +1,47 @@
+package org.pmiops.workbench.actionaudit.auditors;
+
+import org.pmiops.workbench.actionaudit.ActionAuditEvent;
+import org.pmiops.workbench.actionaudit.ActionAuditService;
+import org.pmiops.workbench.actionaudit.ActionType;
+import org.pmiops.workbench.actionaudit.AgentType;
+import org.pmiops.workbench.actionaudit.TargetType;
+import org.pmiops.workbench.actionaudit.targetproperties.AccountTargetProperty;
+import org.pmiops.workbench.actionaudit.targetproperties.values.AccountDisabledStatus;
+import org.pmiops.workbench.db.model.DbUser;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Service;
+
+import javax.inject.Provider;
+import java.time.Clock;
+
+@Service
+public class ClusterAuditorImpl implements ClusterAuditor {
+  private ActionAuditService actionAuditService;
+  private Clock clock;
+  private Provider<DbUser> adminDbUserProvider;
+
+  @Autowired
+  public ClusterAuditorImpl(
+      ActionAuditService actionAuditService,
+      Clock clock,
+      Provider<DbUser> adminDbUserProvider) {
+    this.actionAuditService = actionAuditService;
+    this.clock = clock;
+    this.adminDbUserProvider = adminDbUserProvider;
+  }
+
+  @Override
+  public void fireDeleteClustersInProject(String projectId) {
+    actionAuditService.send(
+        ActionAuditEvent.builder()
+            .timestamp(clock.millis())
+            .agentType(AgentType.ADMINISTRATOR)
+            .agentId(adminDbUserProvider.get().getUserId())
+            .agentEmailMaybe(adminDbUserProvider.get().getUsername())
+            .actionType(ActionType.DELETE)
+            .targetType(TargetType.NOTEBOOK_SERVER)
+            .targetPropertyMaybe(projectId)
+            .build());
+  }
+}
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
@@ -1,47 +1,53 @@
 package org.pmiops.workbench.actionaudit.auditors;
 
+import java.time.Clock;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.inject.Provider;
 import org.pmiops.workbench.actionaudit.ActionAuditEvent;
 import org.pmiops.workbench.actionaudit.ActionAuditService;
 import org.pmiops.workbench.actionaudit.ActionType;
 import org.pmiops.workbench.actionaudit.AgentType;
 import org.pmiops.workbench.actionaudit.TargetType;
-import org.pmiops.workbench.actionaudit.targetproperties.AccountTargetProperty;
-import org.pmiops.workbench.actionaudit.targetproperties.values.AccountDisabledStatus;
 import org.pmiops.workbench.db.model.DbUser;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
-
-import javax.inject.Provider;
-import java.time.Clock;
 
 @Service
 public class ClusterAuditorImpl implements ClusterAuditor {
+  private Provider<String> actionIdProvider;
   private ActionAuditService actionAuditService;
   private Clock clock;
   private Provider<DbUser> adminDbUserProvider;
 
   @Autowired
   public ClusterAuditorImpl(
+      Provider<String> actionIdProvider,
       ActionAuditService actionAuditService,
       Clock clock,
       Provider<DbUser> adminDbUserProvider) {
+    this.actionIdProvider = actionIdProvider;
     this.actionAuditService = actionAuditService;
     this.clock = clock;
     this.adminDbUserProvider = adminDbUserProvider;
   }
 
   @Override
-  public void fireDeleteClustersInProject(String projectId) {
+  public void fireDeleteClustersInProject(String projectId, List<String> clusterNames) {
     actionAuditService.send(
-        ActionAuditEvent.builder()
-            .timestamp(clock.millis())
-            .agentType(AgentType.ADMINISTRATOR)
-            .agentId(adminDbUserProvider.get().getUserId())
-            .agentEmailMaybe(adminDbUserProvider.get().getUsername())
-            .actionType(ActionType.DELETE)
-            .targetType(TargetType.NOTEBOOK_SERVER)
-            .targetPropertyMaybe(projectId)
-            .build());
+        clusterNames.stream()
+            .map(
+                clusterName ->
+                    ActionAuditEvent.builder()
+                        .timestamp(clock.millis())
+                        .agentType(AgentType.ADMINISTRATOR)
+                        .agentId(adminDbUserProvider.get().getUserId())
+                        .agentEmailMaybe(adminDbUserProvider.get().getUsername())
+                        .actionType(ActionType.DELETE)
+                        .newValueMaybe(clusterName)
+                        .targetType(TargetType.NOTEBOOK_SERVER)
+                        .targetPropertyMaybe(projectId)
+                        .build())
+            .collect(Collectors.toList()));
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
@@ -11,6 +11,7 @@ import org.pmiops.workbench.actionaudit.AgentType;
 import org.pmiops.workbench.actionaudit.TargetType;
 import org.pmiops.workbench.db.model.DbUser;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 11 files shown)


api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorImpl.java
```diff
@@ -0,0 +1,47 @@
...
+public class ClusterAuditorImpl implements ClusterAuditor {
+  private ActionAuditService actionAuditService;
+  private Clock clock;
+  private Provider<DbUser> adminDbUserProvider;
+
+  @Autowired
+  public ClusterAuditorImpl(
+      ActionAuditService actionAuditService,
+      Clock clock,
+      Provider<DbUser> adminDbUserProvider) {
```
gjuggler: Nitpick: I'm not sure the "admin" prefix is too helpful here. We have a pretty strong pattern of calling this member variable "userProvider". I see that locally within this package AuthDomainAuditorImpl also calls it "adminDbUserProvider" (which is probably where this naming came from), but I'd argue both usages are inconsistent with a pretty global pattern across our codebase. It crease more confusion than it resolves IMO, since the "user" is still the "user from the current request context" in all cases.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -116,7 +123,51 @@
   }
 
   @Override
-  public ResponseEntity<ClusterListResponse> listClusters(
+  @AuthorityRequired(Authority.SECURITY_ADMIN)
+  public ResponseEntity<List<ListClusterResponse>> deleteClustersInProject(String billingProjectId) {
+    if (billingProjectId == null) {
+      throw new BadRequestException("Must specify billing project");
+    }
+    clusterAuditor.fireDeleteClustersInProject(billingProjectId);
```
gjuggler: I'd be inclined to move this towards the bottom of this method, so the audit log is only sent after the Firecloud interactions are completed. Just checking how Jay set this up for other controllers, there seems to be a pretty clear pattern of "do the work first, then fire an audit log".

api/src/main/resources/workbench.yaml
```diff
@@ -538,6 +538,32 @@ paths:
 
   # Notebook clusters ####################################################################
 
+  /v1/admin/security/clusters/{billingProjectId}/delete-clusters:
+    post:
+      summary: Pause all clusters in a project
```
gjuggler: Nitpick: update summary s/pause/delete/

api/src/main/resources/workbench.yaml
```diff
@@ -3634,14 +3661,44 @@ definitions:
...
-  ClusterListResponse:
+  DefaultClusterResponse:
     type: object
     required:
       - defaultCluster
     properties:
       defaultCluster:
         $ref: "#/definitions/Cluster"
 
+  ListClusterResponse:
```
gjuggler: Is there a need to return data about the clusters that were just paused? I'd be inclined to call YAGNI (you ain't gonna need this) on this piece. We can easily add more return info if required, but I don't seen an obvious use case for the client to use this data.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -234,6 +249,27 @@ public void testListClusters() throws Exception {
         .isEqualTo(testCluster);
   }
 
+  @Test
+  public void testDeleteClustersInProject() throws Exception {
```
gjuggler: The testing pattern here is a bit awkward. What I might have expected to see in a controller-level test would be to make sure all inbound parameters are correctly mapped to the service-level call, e.g.

1. Set up any required mocks, e.g. when(notebookService).listClustersByProjectAsAdmin(...)
2. Call the controller method.
3. Verify that notebookService.deleteClusterAsAdmin was called with BILLING_PROJECT_ID
4. Verify that clusterAuditor.fireDeleteClustersInProject is called with BILLING_PROJECT_ID

api/src/main/resources/workbench.yaml
```diff
@@ -538,6 +538,32 @@ paths:
 
   # Notebook clusters ####################################################################
 
+  /v1/admin/security/clusters/{billingProjectId}/delete-clusters:
```
gjuggler: [Important] Per discussion on Slack, we'll want to flex the requirements to this ticket a bit and add support for deleting a specific named cluster from within the project. The easiest way to handle this would probably be to add an optional body parameter to this API method, "cluster_names", which allows one or more cluster names to be provided.

The idea is that if / when we implement auto-cluster-delete in response to high-egress events, we'll likely have the specific name of the offending cluster. To reduce the blast radius of the alert response, we'll favor deleting just the known cluster name rather than blasting all clusters from the project.

Do you think it'll be possible to fit that into this PR without too much re-working? It could also be a follow-up – your judgement call to make.

api/src/main/resources/workbench.yaml
blrubenstein: I am fairly confident that this is useful data, because if one cluster didn't delete for some reason, we can display information about that on the client side by filtering on cluster status on the returned values. In general, I think of deleting clusters as an 'update to be deleted asynchronously', and believe that update requests should return the new values

api/src/main/resources/workbench.yaml
blrubenstein: Probably, let me add that.

blrubenstein:
This is how it looks in BQ:
```
logName: projects/all-of-us-workbench-test/logs/workbench-action-audit-local
resource.type: global
resource.labels.project_id: all-of-us-workbench-test
textPayload: null
jsonPayload.target_type: NOTEBOOK_SERVER
jsonPayload.agent_type: ADMINISTRATOR
jsonPayload.agent_id: 1.0
jsonPayload.agent_email: brubenst2@fake-research-aou.org
jsonPayload.action_id: 677a5d8d-9b89-4d63-a922-d9f895035897
jsonPayload.timestamp: 1.578948782668E12
jsonPayload.new_value: all-of-us-1
jsonPayload.action_type: DELETE
jsonPayload.target_property: aou-rw-local1-da0e3514
jsonPayload.target_id: null
jsonPayload.prev_value: null
timestamp: 2020-01-13 20:53:03.113660 UTC
receiveTimestamp: 2020-01-13 20:53:03.113660 UTC
severity: INFO
insertId: 4q9mfrg8l8n5dm```

File: api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java (hunk 5 above)
Comment (blrubenstein) [reply api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java]: Its kinda tricky because all the notebookService stuff is calling to Leo. I will update though.

File: api/src/main/java/org/pmiops/workbench/api/ClusterController.java [hunk 7]
```diff
@@ -116,7 +124,65 @@
...
-  public ResponseEntity<ClusterListResponse> listClusters(
+  @AuthorityRequired(Authority.SECURITY_ADMIN)
+  public ResponseEntity<List<ListClusterResponse>> deleteClustersInProject(
+      String billingProjectId, ListClusterDeleteRequest clusterNamesToDelete) {
+    if (billingProjectId == null) {
+      throw new BadRequestException("Must specify billing project");
+    }
+    List<org.pmiops.workbench.notebooks.model.ListClusterResponse> clustersToDelete =
+        leonardoNotebooksClient.listClustersByProjectAsAdmin(billingProjectId).stream()
+            .filter(
```
jmthibault79: This filter block is repeated below, so we can extract it as a small function

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -234,6 +249,26 @@ public void testListClusters() throws Exception {
...
```
jmthibault79: does not throw Exception - this can actually be removed from every method in this class

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/ClusterAuditorTest.java
```diff
@@ -0,0 +1,89 @@
...
+    doReturn(Y2K_EPOCH_MILLIS).when(mockClock).millis();
+    doReturn(ACTION_ID).when(mockActionIdProvider).get();
+  }
+
+  @Test
+  public void testFireDeleteClustersInProject() {
+    clusterAuditor.fireDeleteClustersInProject(BILLING_PROJECT_ID, CLUSTER_NAMES);
+    verify(mockActionAuditService).send(eventCollectionCaptor.capture());
+    Collection<ActionAuditEvent> eventsSent = eventCollectionCaptor.getValue();
+    assertThat(eventsSent).hasSize(3);
```
jmthibault79: changing this to `CLUSTER_NAMES.length()` would make it explicit

api/src/main/resources/workbench.yaml
```diff
@@ -565,7 +597,7 @@ paths:
         200:
           description: Available clusters
           schema:
-            $ref: '#/definitions/ClusterListResponse'
+            $ref: '#/definitions/DefaultClusterResponse'
```
jmthibault79: good rename

api/src/main/resources/workbench.yaml
```diff
@@ -565,7 +597,7 @@ paths:
         200:
           description: Available clusters
```
jmthibault79: Can we change this to "the user's cluster" or similar?

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
jmthibault79: I would also like to see these test cases:
* `clusterNamesToDelete.getClustersToDelete()` is null
* `clusterNamesToDelete.getClustersToDelete()` is an empty list
* delete some but not all clusters
* request to delete cluster(s) which exist but are in a different project
* returned cluster status is DELETED (I think there's a bug here)
* returned cluster status is in a non-acceptable state

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -116,7 +124,65 @@
...
+                            .getClustersToDelete()
+                            .contains(cluster.getClusterName()))
+            .collect(Collectors.toList());
+
+    clustersToDelete.forEach(
+        cluster ->
+            leonardoNotebooksClient.deleteClusterAsAdmin(
+                cluster.getGoogleProject(), cluster.getClusterName()));
+    List<ListClusterResponse> clustersInProjectAffected =
+        leonardoNotebooksClient.listClustersByProjectAsAdmin(billingProjectId).stream()
```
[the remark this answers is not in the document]
blrubenstein: Hmmm... I don't think we want to list all deleted clusters though, because we reuse names. We may be okay with this, because clusters probably don't delete that quickly? It will affect return value if they do delete quickly, but won't affect auditing.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: Talked offline, we are not adding the last two of these test cases because: 

- With DELETED, we don't want to receive those clusters, because we reuse names, and thus there could be many clusters with the same name in deleted.
- We could do the non-acceptable state, but it is just testing logging, and is going to be incredibly messy to maintain that test. (And is sorta testing 'what happens if Leo fails to follow contract')

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -242,7 +294,96 @@ public void testListClusters() throws Exception {
...
+    List<ListClusterResponse> listClusterResponseList =
+        ImmutableList.of(testFcClusterListResponse, testFcClusterListResponse2);
+    List<String> clustersToDelete =
+        ImmutableList.of(testFcClusterDifferentProject.getClusterName());
+    when(notebookService.listClustersByProjectAsAdmin(BILLING_PROJECT_ID))
+        .thenReturn(listClusterResponseList);
+
+    clusterController.deleteClustersInProject(
+        BILLING_PROJECT_ID, new ListClusterDeleteRequest().clustersToDelete(clustersToDelete));
+    verify(notebookService, times(0))
```
jmthibault79: nit, optional: there's a `verifyZeroCalls()` (or similar) that can be used instead

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -242,7 +294,96 @@ public void testListClusters() throws Exception {
...
+        .deleteClusterAsAdmin(BILLING_PROJECT_ID, testFcCluster.getClusterName());
+    verify(clusterAuditor, times(0))
+        .fireDeleteClustersInProject(BILLING_PROJECT_ID, clustersToDelete);
+  }
+
+  @Test
+  public void testDeleteClustersInProjectNoClusters() {
```
jmthibault79: This is something I was curious about: do we want a missing param (null) to have different semantics from an empty-list param?  We should be explicit about this in the swagger, if so.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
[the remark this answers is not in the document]
jmthibault79: agreed offline - we do not want to list deleted

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: I think we do, because I think not providing a list should give all, but an empty list should behave the same way as any list. I can be more explicit in swagger.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: `verifyZeroInteractions()` is a mock level method, rather than a method level method (from what I could tell. Sticking with this.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
blrubenstein: Basically I think there should be some default that says include all clusters, and the body parameter can't be null. We could also do this with a bool on the object that says 'deleteAll' or something, but that seems like extra that we don't need.

jaycarlton:
Discussed offline. What we want is
- actionType = DELETE
- N events with  target=clusterId, targetType=NOTEBOOK_CLUSTER, and property parent_project = projectId
- possibly other cluster metadata as available.

PR #2995, opened 2020-01-10
[risk=low][RW-4178] Prevent compute cost increasing API calls on billing inactive workspaces

ericsong:
Description:
After many rounds of discussion, we have settled on the following user operations to be prevented when a workspace becomes billing inactive.

- Edit Notebook
- Playground Mode (Notebooks)
- Create Notebook
- Copy Notebook
  - Copying into active billing workspaces is OK, even if the current workspace is billing inactive
  - Copying into inactive billing workspaces is not OK
- Export to Notebook
- Rename Notebook

This PR adds a check such that a 403 is returned to the client if the targeted workspace is in an inactive billing state. This only prevents access to the API. A UI to reflect the changes and improve the user experience will be coming in a future ticket.

I decided to change the implementation from the swagger + AuthInterceptor approach to one that performs the checks at the beginning of the controller methods. Some reasons why I made this change

- There is no generic way to check all of the requests since the identifiers for the workspace are not uniform. Nor can they be since some of the requests are checking against the destination workspace, not the current one.
- The current approach is an existing pattern that we use to check for Firecloud access permissions.
- It's still easy to determine which methods are being gated by finding usages of the `WorkspaceService.requireActiveBilling` method.

---
**PR checklist**


ericsong:
@als364 @calbach @jaycarlton 

Tagging you three for review since you all looked at the previous PR with the compute tags.

ericsong:
NVM - I forgot to add the feature flag. Please come back after I add it. I'll add a comment when I do.

ericsong:
Feature flag added and ready for review

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -438,6 +438,8 @@ private void formatTimestampValues(List<DataSetPreviewValueList> valuePreviewLis
   @Override
   public ResponseEntity<EmptyResponse> exportToNotebook(
       String workspaceNamespace, String workspaceId, DataSetExportRequest dataSetExportRequest) {
+    workspaceService.requireActiveBilling(workspaceNamespace, workspaceId);
```
calbach: These checks should follow the access check, else this reveals some amount of information about existence of the workspace and its billing status, to any caller. In the cases where there is no explicit access check happening in the controller method, it would be probably be *correct* to add one - or else push this logic down into the service which is probably making that check. If this gets too hairy, I could probably be convinced it's not worth the effort for some of these without heavier refactoring.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -243,6 +248,19 @@ public DbWorkspace getRequiredWithCohorts(String ns, String firecloudName) {
...
+  @Override
+  public void requireActiveBilling(String workspaceNamespace, String workspaceId)
+      throws ForbiddenException {
+    if (!workbenchConfigProvider.get().featureFlags.enableBillingLockout) {
+      return;
+    }
+
+    if (BillingStatus.INACTIVE.equals(
+        getRequired(workspaceNamespace, workspaceId).getBillingStatus())) {
+      throw new ForbiddenException("Workspace is in an inactive billing state");
```
calbach: this is potentially ambiguous, should this include the workspace ID in the message? Alternatively this method could return bool and the controller could throw a contextually relevant message.

api/config/config_local.json
```diff
@@ -85,7 +85,8 @@
     "unsafeAllowDeleteUser": true,
     "enableVpcFlowLogs": true,
     "enableVpcServicePerimeter": true,
-    "enableNewAccountCreation": true
+    "enableNewAccountCreation": true,
+    "enableBillingLockout": true
```
calbach: nit: I think we should keep local and test in sync typically. I don't feel strongly about both true or both false

jaycarlton: aside: is it too late to rename `enableNewAccountCreation`? It's really talking about a revised flow, not enabling people to create new accounts altogether?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
jaycarlton: The need to add code to a number methods in just the right place makes me think we've missed an abstraction/opportunity to have something more systematic in place. Maybe a shim in between the ApiDelegate class and the Controller would be a good place to stick these validation methods generically.

What's here is probably enough/appropriate for now to avoid scope creep, but if we need multiple overlapping kinds of checks soon, it could get unwieldy.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -40,6 +41,9 @@
 
   DbWorkspace saveWithLastModified(DbWorkspace workspace);
 
+  void requireActiveBilling(String workspaceNamespace, String workspaceId)
```
jaycarlton: nit: I've been naming these kind of functions starting with `validate`, which I associate with a check that throws. As written, it's active voice, as if we're telling the class to require active billing on these arguments.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -418,6 +424,31 @@ public void testLocalize_noNotebooks() throws Exception {
     assertThat(resp.getClusterLocalDirectory()).isEqualTo("workspaces/wsid");
   }
 
+  @Test(expected = ForbiddenException.class)
+  public void listCluster_requireActiveBilling() {
+    testWorkspace.setBillingStatus(BillingStatus.INACTIVE);
```
jaycarlton: I'm confused why this test case needs you to both set inactive here and then stub the call below.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -418,6 +424,31 @@ public void testLocalize_noNotebooks() throws Exception {
...
+    doThrow(ForbiddenException.class)
+        .when(workspaceService)
+        .requireActiveBilling(WORKSPACE_NS, WORKSPACE_ID);
+
+    clusterController.listClusters(WORKSPACE_NS, WORKSPACE_ID);
+  }
+
+  @Test(expected = ForbiddenException.class)
+  public void localize_requireActiveBilling() {
+    testWorkspace.setBillingStatus(BillingStatus.INACTIVE);
```
jaycarlton: Also, I'd probably just make two test workspaces, one active and one not, to cut down on the per-testcase setup.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -418,6 +424,31 @@ public void testLocalize_noNotebooks() throws Exception {
...
+
+    doThrow(ForbiddenException.class)
+        .when(workspaceService)
+        .requireActiveBilling(WORKSPACE_NS, WORKSPACE_ID);
```
jaycarlton: So the only assertion is that the test throws this exception, and you're only throwing the exception because of Mockito. This feels like you're possibly testing this behavior at the wrong level, or in the wrong class.

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -960,6 +968,20 @@ public void exportToNewNotebook() {
         .saveNotebook(eq(WORKSPACE_BUCKET_NAME), eq(notebookName), any(JSONObject.class));
   }
 
+  @Test(expected = ForbiddenException.class)
```
jaycarlton: I just learned that there's an error code `402 Payment Required`, but it's nonstandard. Would've fit like a glove here though. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402

api/src/test/java/org/pmiops/workbench/workspaces/WorkspaceServiceTest.java
```diff
@@ -55,6 +56,7 @@
   @Mock private DataSetService mockDataSetService;
```
jaycarlton: If you're looking for a rabbit hole, these should all be `MockBean`s

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -84,7 +84,7 @@
   // Importing the real one requires importing a large subtree of dependencies
   @Bean
   public WorkspaceService workspaceService() {
-    return new WorkspaceServiceImpl(null, null, null, null, null, null, null, null, null);
+    return new WorkspaceServiceImpl(null, null, null, null, null, null, null, null, null, null);
```
jaycarlton: I was going to post some reaction gifs to this constructor invocation, but I think there's an easier way to solve this issue.

I'd go ahead and make a `FakeWorkspaceServiceImpl` class, use IntelliJ's default overload implementations that just return null, give it a zero-arg constructor, and tag it with `@Service` and give it a bean name. Then just inject it like a real service.

You'll likely need to mark the real one as primary so that the fake doesn't get pulled into real code.

Even simpler would be to mock it with Mockito, but I think using it in production code is controversial if not frowned upon.

It's still a smell to me that you need to pull in this bean but don't need a working version, but I haven't looked around very much to understand the main issue you're working around.

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -196,6 +196,9 @@ public static WorkbenchConfig createEmptyConfig() {
     // Flag to indicate whether to enable the new Create Account flow
     // https://precisionmedicineinitiative.atlassian.net/browse/RW-3284
     public boolean enableNewAccountCreation;
+    // Setting this to true will prevent users from making compute increasing operations on
```
jaycarlton: If you have a few minutes to write a small markdown file about this mechanism and how it works and when you should make your APIs require billing, you could simply put the URL to that here. I don't think there's enough context here to help folks who come later really understand it.

api/config/config_local.json
jaycarlton: sounds good.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -95,6 +98,7 @@ public WorkspaceServiceImpl(
       UserDao userDao,
       Provider<DbUser> userProvider,
       UserRecentWorkspaceDao userRecentWorkspaceDao,
+      Provider<WorkbenchConfig> workbenchConfigProvider,
```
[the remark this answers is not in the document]
jaycarlton: I meant the last parameter in the list. It's not worth worrying about.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
jaycarlton: I was thinking about this a bit. I think just verifying that `requireActiveBilling()` is called is enough here, since we test that it throws or not correctly elsewhere. That way you could drop the `expected` and the `doThrow` in favor of one statement that's a bit clearer I think.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
jaycarlton: Right, so you need something depends on WorkspaceService, but aren't actually calling any method on it that needs WorkspaceService. That's a symptom of the services getting too large most likely. Not something to chase here if you're trying to keep this small though.

api/build.gradle
```diff
@@ -428,7 +428,7 @@ dependencies {
   toolsCompile 'commons-cli:commons-cli:1.4'
   toolsCompile 'com.opencsv:opencsv:4.6'
 
-  testCompile 'junit:junit:4.12'
+  testCompile 'junit:junit:4.13'
```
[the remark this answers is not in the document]
jaycarlton: Oh nice. I wanted to use that somewhere not long ago.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
ericsong: I was able to correctly order the access checks by pushing down the validate call to the underlying copy notebook function which performs the workspace access check. One side effect of this change, is that Rename Notebook will no longer be allowed since rename works by creating a new file and deleting the original.

Given that we were unsure if Rename should be allowed or not, I thought it would be OK to not allow it if it makes the implementation easier.

Another note to Jay's point, this refactoring went in a direction of checking access at the point where we perform the relevant action instead of checking it at the top of the controller methods. It's a design that he brought up a while ago and I think that may be the way to go if we need to add more access control. It is going to be difficult to have to fully understand what an API call is going to do and declare the necessary permissions.

One downside to that approach though, is that our access control checks can be expensive (if we ask Firecloud) so we do not want to be calling it multiple times in a single client serving API call. One solution to this may be to add some caching mechanism that exists for the duration of a client serving API call so that we can add access control checks at various points without worrying about performance.

PR #3000, opened 2020-01-14
[no ticket][risk=no] Did I say Attachments? I meant Tags.

jaycarlton:
In coming and going on my previous branch, I formed a memory in my head validating that the attachments I was using were being reflected in Stackdriver. It turns out that that's doubly not working: attachments don't support my use case and the Stackdriver exporter doesn't support exporting attachments. 

Luckily, changing to Tags wasn't too bad. It's a bit irksome because you have to commit to your labels (tags) when creating your metrics. I'm researching the best way to establish things, whether separately through the API or in our Java code.

I added some utilities to [workbench-devops](https://github.com/all-of-us/workbench-devops/pull/11) for clearing out old metrics. They die after the retention period ends, but in the meantime they really clutter things up.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/monitoring/views/GaugeMetric.java
```diff
@@ -1,72 +1,60 @@
...
       "total number of users",
-      ImmutableSet.of(
-          Attachment.USER_BYPASSED_BETA, Attachment.USER_DISABLED, Attachment.DATA_ACCESS_LEVEL)),
+      ImmutableList.of(
+          MetricLabel.USER_BYPASSED_BETA,
+          MetricLabel.USER_DISABLED,
+          MetricLabel.DATA_ACCESS_LEVEL)),
   WORKSPACE_COUNT(
-      "workspace_count",
+      "workspace_count_3",
```
gjuggler: Everything else was incremented to _2, but this was incremented to _3. Who should know about why this metric is different from all the others?

jaycarlton: These are all going to get scrapped at least one more time, so I didn't invest in consistency or name conventions yet. I need to circulate a convention where we have a way to give versions to the metric descriptors.

It appears that registering (in OpenCensus) a view corresponding to an existing metric is a no-op if that metric is actually live, but the Google api fails with a "metric exists" error if the metric has been deleted. Still sorting this out a bit.

PR #3005, opened 2020-01-15
[RISK=LOW][No ticket] Removing null constraint from User table column creation and modified time 

NehaBroad:
This will remove the not null constrain on the  columns creation and last_modified_time as well as removing the default values.

The reason behind this is that rather than populating creation_time with first_sign_in (which can be null as well as incorrect) or the current date time, we will spend effort in trying to find ways to populate the column with close to actual creation_date time for existing user (maybe checking when the got the email for registration or other way). In order to do that its better to have the values as null so it will be easier to find users that needs updating.

There is another ticket that is going to take care of populating the correct creation and last_modified_time for existing users https://precisionmedicineinitiative.atlassian.net/browse/RW-4278

### DIFF
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -7,17 +7,11 @@
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified">
     <addColumn tableName="user">
       <column name="creation_time" type="datetime">
-        <constraints nullable="false" />
+        <constraints nullable="true" />
       </column>
       <column name="last_modified_time" type="datetime">
-        <constraints nullable="false"/>
+        <constraints nullable="true"/>
       </column>
     </addColumn>
-  </changeSet>
-  <changeSet id="update creation time" author="nsaxena">
-    <sql>update user set creation_time = first_sign_in_time</sql>
-  </changeSet>
-  <changeSet id="update modified time" author="nsaxena">
-    <sql>update user set last_modified_time = first_sign_in_time</sql>
   </changeSet>
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -9,9 +9,12 @@
       <column name="creation_time" type="datetime">
         <constraints nullable="true" />
       </column>
-      <column name="last_modified_time" type="datetime">
+      <column name="last_modified_time" type="datetime" defaultValueDate="CURRENT_TIMESTAMP">
         <constraints nullable="true"/>
       </column>
     </addColumn>
+  </changeSet>
+  <changeSet id="update modified time" author="nsaxena">
+    <sql>update user set last_modified_time = first_sign_in_time where first_sign_in_time is not null</sql>
   </changeSet>
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -10,7 +10,7 @@
         <constraints nullable="true" />
       </column>
       <column name="last_modified_time" type="datetime" defaultValueDate="CURRENT_TIMESTAMP">
-        <constraints nullable="true"/>
+        <constraints nullable="false"/>
       </column>
     </addColumn>
   </changeSet>
--- a/api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
@@ -12,7 +12,7 @@ public interface RdrExportDao extends CrudRepository<DbRdrExport, Long> {
       nativeQuery = true,
       value =
           "select u.user_id from user u LEFT JOIN rdr_export rdr on"
-              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where u.last_modified_time is null or "
+              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where "
               + "u.last_modified_time > rdr.last_Export_date or rdr.export_id is null")
   List<BigInteger> findDbUserIdsToExport();
 
--- a/api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
@@ -12,7 +12,7 @@ public interface RdrExportDao extends CrudRepository<DbRdrExport, Long> {
       nativeQuery = true,
       value =
           "select u.user_id from user u LEFT JOIN rdr_export rdr on"
-              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where "
+              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where u.last_modified_time is null or "
               + "u.last_modified_time > rdr.last_Export_date or rdr.export_id is null")
   List<BigInteger> findDbUserIdsToExport();
 
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -372,24 +372,28 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
   @Override
   public DbUser submitDataUseAgreement(
       DbUser dbUser, Integer dataUseAgreementSignedVersion, String initials) {
-    // FIXME: this should not be hardcoded
-    if (dataUseAgreementSignedVersion != CURRENT_DATA_USE_AGREEMENT_VERSION) {
-      throw new BadRequestException("Data Use Agreement Version is not up to date");
-    }
-    final Timestamp timestamp = new Timestamp(clock.instant().toEpochMilli());
-    DbUserDataUseAgreement dataUseAgreement = new DbUserDataUseAgreement();
-    dataUseAgreement.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
-    dataUseAgreement.setUserId(dbUser.getUserId());
-    dataUseAgreement.setUserFamilyName(dbUser.getFamilyName());
-    dataUseAgreement.setUserGivenName(dbUser.getGivenName());
-    dataUseAgreement.setUserInitials(initials);
-    dataUseAgreement.setCompletionTime(timestamp);
-    userDataUseAgreementDao.save(dataUseAgreement);
-    // TODO: Teardown/reconcile duplicated state between the user profile and DUA.
-    dbUser.setDataUseAgreementCompletionTime(timestamp);
-    dbUser.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
-    dbUser.setLastModifiedTime(timestamp);
-    return userDao.save(dbUser);
+
+    return updateUserWithRetries(
+            (user) -> {
+              // FIXME: this should not be hardcoded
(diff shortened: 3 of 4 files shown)


api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -259,10 +267,11 @@ private RdrWorkspace toRdrWorkspace(DbWorkspace dbWorkspace) {
...
     aclMap.forEach(
         (email, access) -> {
-          RdrWorkspaceUser workspaceUderMap = new RdrWorkspaceUser();
-          workspaceUderMap.setUserId((int) userDao.findUserByUsername(email).getUserId());
-          workspaceUderMap.setRole(RdrWorkspaceUser.RoleEnum.fromValue(access.getAccessLevel()));
-          rdrWorkspace.addWorkspaceUsersItem(workspaceUderMap);
+          RdrWorkspaceUser workspaceUserMap = new RdrWorkspaceUser();
+          workspaceUserMap.setUserId((int) userDao.findUserByUsername(email).getUserId());
+          workspaceUserMap.setRole(RdrWorkspaceUser.RoleEnum.fromValue(access.getAccessLevel()));
+          workspaceUserMap.setStatus(RdrWorkspaceUser.StatusEnum.fromValue("ACTIVE"));
```
ericsong: why is this being set now?

api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
```diff
@@ -12,7 +12,7 @@
       nativeQuery = true,
       value =
           "select u.user_id from user u LEFT JOIN rdr_export rdr on"
-              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where "
+              + " u.user_id = rdr.export_id and rdr.entity_type = 1 where u.last_modified_time is null or "
```
ericsong: are we ok with hitting the RDR multiple times with the same user account? As is, I believe a user will keep getting sent until they log in and their lat_modified_time gets set.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -7,17 +7,11 @@
...
-        <constraints nullable="false"/>
+        <constraints nullable="true"/>
       </column>
     </addColumn>
   </changeSet>
-  <changeSet id="update creation time" author="nsaxena">
-    <sql>update user set creation_time = first_sign_in_time</sql>
-  </changeSet>
-  <changeSet id="update modified time" author="nsaxena">
-    <sql>update user set last_modified_time = first_sign_in_time</sql>
```
ericsong: I think its OK to leave the last modified time update sql. That value changes often and its not as important to ensure its integrity.

api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
```diff
@@ -12,7 +12,7 @@
       nativeQuery = true,
```
ericsong: unrelated but we should try to avoid native queries wherever possible since our test db and production db are different.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
NehaBroad: This has been made required for user by RDR. this is NOT part of this PR it just got sneaked in. I can revert this and make a new PR for this

api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
NehaBroad: Since rdr_export doesn't have ENTITY_ID as a foreign key (since it can contain either WORKSPACE  or USER id) i was unable to use it the using Db classes.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
ericsong: I think that may be better even if its a small change since I have a lingering comment about its implementation.

That comment being, how do we know this value can be set to "ACTIVE"? IIRC, firecloud can return a value for the NO_ACCESS role.

api/src/main/java/org/pmiops/workbench/db/dao/RdrExportDao.java
NehaBroad: Actually i am going to push change to set the  value of  last_modified_date time as either today's date or first_sign in just so it is picked  only the first time

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -386,6 +386,7 @@ public DbUser submitDataUseAgreement(
     // TODO: Teardown/reconcile duplicated state between the user profile and DUA.
     dbUser.setDataUseAgreementCompletionTime(timestamp);
     dbUser.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
+    dbUser.setLastModifiedTime(timestamp);
```
gjuggler: This feels misplaced. If the issue is that last_modified_time isn't getting updated when the user is saved, I think we should really be changing that value up within the UserServiceImpl.updateUserWithRetries method, which is where all user-modifying changes funnel through. And then this method should probably call updateUserWithRetries instead of userDao.save.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
NehaBroad: Thats because as of now for RDR we are going with assumption that all the USERS are active until we introduce an **official** way of removing users from workbench

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -157,8 +157,16 @@ public void exportWorkspaces(List<Long> workspaceIds) {
   private RdrResearcher toRdrResearcher(DbUser dbUser) {
     RdrResearcher researcher = new RdrResearcher();
     researcher.setUserId((int) dbUser.getUserId());
-    researcher.setCreationTime(dbUser.getCreationTime().toLocalDateTime().atOffset(offset));
-    researcher.setModifiedTime(dbUser.getLastModifiedTime().toLocalDateTime().atOffset(offset));
+    if (null != researcher.getCreationTime()) {
+      researcher.setCreationTime(dbUser.getCreationTime().toLocalDateTime().atOffset(offset));
+    } else {
+      researcher.setCreationTime(OffsetDateTime.now());
```
gjuggler: It's super surprising to see this conversion code take a null value from the DB and set it to the current timestamp in the RDR export call. I'm still kind of scratching my head about this... what's the motivation behind this choice? Is the RDR API unwilling to accept null values? Are they okay with receiving data about researchers whose creation time is constantly changing?

This absolutely needs some comment in the code explaining why we are making a somewhat-surprising choice.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -7,17 +7,14 @@
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified">
     <addColumn tableName="user">
       <column name="creation_time" type="datetime">
-        <constraints nullable="false" />
```
gjuggler: [Prod safety comment] How sure are we that it's safe to modify an existing Liquibase change in-place? Will we have to do something manually to make sure the new version of this change gets applied in our test environment? What happens if someone tries to rollback this PR and/or https://github.com/all-of-us/workbench/pull/2948 on their own? (I suspect something will likely go wrong in our prod DB.) We should think of a clear way to signal to the oncall next week that this PR isn't atomically rollback-safe.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
ericsong: I see. It does seem a bit odd that we would send over "ACTIVE" for disabled users but if that's what they want ¯\_(ツ)_/¯.

Maybe add a comment saying we should set this dynamically when we do start removing users from workbench?

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
NehaBroad: I will add comments about this, right now i wanted to solve the creation_date issue with liquibase just so it unblocks the test build.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
ericsong: I think the issue is that something may go wrong if we perform a rollback to this PR or the previous one since the liquibase commits are split between two files.

I think we might actually be OK given that we don't execute any database rollbacks ATM so it should just be the Java/UI code that switches back.

This made me think of an interesting point though. We may need to mark some commits on master as invalid candidates for rollback because it contains a known bug. Previously, I was only thinking about incompatibilities due to database changes.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -7,17 +7,14 @@
...
+        <constraints nullable="true" />
       </column>
-      <column name="last_modified_time" type="datetime">
-        <constraints nullable="false"/>
+      <column name="last_modified_time" type="datetime" defaultValueDate="CURRENT_TIMESTAMP">
```
ericsong: do we need a default value here?

We are already allowing nullable and we're copying over the first_sign_in_time (where it exists).

The integrity of `last_modified_time` is not as important but I still think its better to avoid setting it to a mix of values (first_sign_in_time and current timestamp).

If we do want to keep the default value, we can get rid of the nullable constraint.

NehaBroad: Lsat_modified_time should have some value so in case of null first_sign it would be current date.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
ericsong: Can we use the updated `updateUserWithRetries` method to avoid setting this here?

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
jaycarlton: Is there a separate notion of `RDR_ACTIVE` from `WORKBENCH_ACTIVE`?

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -370,23 +372,28 @@ public DbUser createUser(
...
-    dataUseAgreement.setUserInitials(initials);
-    dataUseAgreement.setCompletionTime(timestamp);
-    userDataUseAgreementDao.save(dataUseAgreement);
-    // TODO: Teardown/reconcile duplicated state between the user profile and DUA.
-    dbUser.setDataUseAgreementCompletionTime(timestamp);
-    dbUser.setDataUseAgreementSignedVersion(dataUseAgreementSignedVersion);
-    return userDao.save(dbUser);
+
+    return updateUserWithRetries(
+        (user) -> {
```
ericsong: the user modifier function should only contain code to update the user object.

Otherwise, everything else will be run multiple times with the retries. In this case, the data use agreement will be created several times.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
jaycarlton: Liquibase migrations should work the same way every time. If we change the default, we're really changing the schema *every time we run the migration*.

PR #3007, opened 2020-01-16
[RW-4233][risk=no] adding PM and Surveys to concept search

freemabd:
adding PM and Surveys to concept search. This PR includes changes for:

- make-bq-data.sh - update standard concept counts for measurements
- ConceptsController - converting TO_CLIENT_CONCEPT function to a regular method and updating   search to include PM/Survey
- ConceptService - updating service to include PM/Survey calls
- CbCriteriaDao - implemented survey db calls here to allow proper ordering of questions
- ConceptDao - Implemented PM db calls here
- DomainInfoDao - added source/standard checks and implemented call for PM

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
@@ -1,5 +1,7 @@
 package org.pmiops.workbench.api;
 
+import static org.pmiops.workbench.model.StandardConceptFilter.ALL_CONCEPTS;
+
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import java.util.ArrayList;
@@ -102,7 +104,7 @@ public class ConceptsController implements ConceptsApiDelegate {
   private void addDomainCounts(SearchConceptsRequest request, ConceptListResponse response) {
     if (request.getIncludeDomainCounts()) {
       StandardConceptFilter standardConceptFilter = request.getStandardConceptFilter();
-      boolean allConcepts = standardConceptFilter == StandardConceptFilter.ALL_CONCEPTS;
+      boolean allConcepts = standardConceptFilter == ALL_CONCEPTS;
       DbDomainInfo pmDomainInfo = null;
       String matchExp = ConceptService.modifyMultipleMatchKeyword(request.getQuery());
       List<DbDomainInfo> allDbDomainInfos = conceptService.getAllDomainsOrderByDomainId();
@@ -110,11 +112,11 @@ public class ConceptsController implements ConceptsApiDelegate {
           matchExp == null ? allDbDomainInfos : new ArrayList<>();
       if (matchingDbDomainInfos.isEmpty()) {
         matchingDbDomainInfos =
-            allConcepts
-                ? conceptService.getAllConceptCounts(matchExp)
-                : conceptService.getStandardConceptCounts(matchExp);
+            conceptService.getConceptCounts(matchExp, request.getStandardConceptFilter().name());
         if (allConcepts) {
-          pmDomainInfo = conceptService.findPhysicalMeasurementConceptCounts(matchExp);
+          pmDomainInfo =
+              conceptService.findPhysicalMeasurementConceptCounts(
+                  matchExp, request.getStandardConceptFilter().name());
         }
       }
       Map<Domain, DbDomainInfo> domainCountMap =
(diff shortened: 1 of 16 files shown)


ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -31,6 +25,16 @@ const styles = reactStyles({
   }
 });
 
+const filterConcepts = (concepts: any[], domain: Domain) => {
```
freemabd: Will, does PM still have a separate bucket at the bottom of the UI?

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -348,49 +342,51 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
-        selectedDomain: conceptDomainCounts
-          .find(domainCount => domainCount.domain === domain.domain)},
-        this.searchConcepts);
-    }
-
-    browseSurvey(surveyName) {
-      this.setState({browsingSurvey: true,
-        selectedSurvey: surveyName});
+      const selectedDomain = !domain ? {domain: Domain.SURVEY, name: 'Surveys', conceptCount: 0}
+        : conceptDomainCounts.find(domainCount => domainCount.domain === domain.domain);
```
freemabd: Is it easier to separate domain counts api calls from data api calls from a UI perspective? Seems like it would simplify logic/responsibility of the api call.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -411,23 +407,11 @@ export const ConceptHomepage = withCurrentWorkspace()(
       return selectedConceptDomainMap[selectedDomain.domain].length;
     }
 
-    selectedQuestion(selectedQues) {
```
freemabd: Does the api need a call the get survey answers? Or is that obsolete now?

ui/src/app/pages/data/concept/concept-table.tsx
```diff
@@ -107,23 +158,24 @@ export class SynonymsObject extends React.Component<{}, SynonymsObjectState> {
 }
 
 interface Props {
-  concepts: Concept[];
+  concepts: any[];
```
freemabd: Did we change this to any because it can also be PM?

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
```diff
@@ -195,12 +192,34 @@ private void addDomainCounts(SearchConceptsRequest request, ConceptListResponse
...
+        .prevalence(concept.getPrevalence())
+        .standardConcept(ConceptService.STANDARD_CONCEPT_CODE.equals(concept.getStandardConcept()))
+        .vocabularyId(concept.getVocabularyId())
+        .conceptSynonyms(concept.getSynonyms());
+  }
+
+  private SurveyQuestions toClientSurveyQuestions(DbCriteria dbCriteria) {
+    return new SurveyQuestions()
+        .conceptId(dbCriteria.getLongConceptId())
+        .question(dbCriteria.getName());
```
dolbeew: Do surveys have to use this structure or can it be modified to add properties (possibly survey name)? Not needed now, just asking for future reference.

ui/src/app/pages/data/concept/concept-add-modal.tsx
dolbeew: Yes, the code is there but is currently feature flagged. Should only appear in local and test environments

ui/src/app/pages/data/concept/concept-table.tsx
dolbeew: Actually changed so surveys could be included. We used to use a separate component but now we use the same one for all domains

ui/src/app/pages/data/concept/concept-homepage.tsx
dolbeew: Yes, that would actually make it simpler on the UI also.

ui/src/app/pages/data/concept/concept-homepage.tsx
dolbeew: We've removed the answers since they're not selectable. It's possible we may need them in the future though

common-api/src/main/java/org/pmiops/workbench/cdr/dao/ConceptDao.java
```diff
@@ -42,5 +43,59 @@
...
+   * @param matchExp SQL MATCH expression to match concept name or synonym
+   * @param conceptTypes can be 'S', 'C' or ''
+   * @param domainIds domain IDs to use when filtering concepts
+   * @param vocabularyId vocabulary id type to search
+   * @param conceptClassId concept class id type to search
+   * @return per-vocabulary concept counts
+   */
+  @Query(
+      value =
+          "select distinct c from DbConcept c "
```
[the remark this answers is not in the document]
freemabd: It is, but we query from the db_criteria table for surveys. This allows us to get the questions in order, which we cannot do when querying from concept table.

api/src/main/java/org/pmiops/workbench/api/ConceptsController.java
freemabd: Yes, we can definitely add survey name to this model

freemabd: Yes, survey name can definitely be added to the model.

PR #3009, opened 2020-01-16
[risk=no][RW-3175] (Part 1) Track billing account names being used by workspaces

ericsong:
Splitting out a small piece of work to prep for the billing lockout stuff. This adds a field to the workspace that tracks the billing account that is currently being used.

Notes

- I used `Name` instead of `Id` because that is what Google calls it in their API
- I dropped the `BillingAccountType` field because we can infer the type by comparing the billing account name field to the environment's free tier account. This will simplify some of our upcoming update billing account code.

TIL

- MySQL has a function parse JSON

Description:

---
**PR checklist**


api/db/changelog/db.changelog-116-add-billing-account-to-workspaces.xml
```diff
@@ -0,0 +1,18 @@
...
+  <changeSet author="ericsong" id="db.changelog-116-add-billing-account-to-workspaces">
+    <addColumn tableName="workspace">
+      <column name="billing_account_name" type="varchar(100)">
+        <constraints nullable="false"/>
+      </column>
+    </addColumn>
+
+    <sql>update workspace set billing_account_name=(SELECT JSON_UNQUOTE(JSON_EXTRACT(configuration, '$.billing.accountId')) from config where config_id='main');</sql>
+
+    <dropColumn tableName="workspace" columnName="billing_account_type"/>
```
calbach: I misunderstood, I thought the proposal was null (free tier) OR user_billing_account_name. I think this would be preferred as it makes the usage of free tier a more explicit, separate state and avoids reencoding the exact free tier billing name for each workspace.

ericsong: I was planning to add the free tier billing name for all the workspaces with a backfill because it simplifies the update/syncing code. 

Ex. in my feature branch, there is no need for special handling of a null or empty string in the billing_account_name field. The UI is able to match the value from `billing_account_name` to the accounts in `listBillingAccounts` for both the free tier and user supplied case. The same should be true for the update billing account case.

Given that, is it still worth keeping both fields? If so, I can add a constraint to ensure that the type is always set to free tier if the billing account name matches the config account name.

jaycarlton: +1 for keeping both fields. It seems fragile to rely on a naming convention, especially as project names aren't enforced by the DB schema. Also, how is the local vs test env name handled? Are they the same billing acct?

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -62,6 +62,12 @@ public FreeTierBillingService(
     this.clock = clock;
   }
 
+  public boolean isUsingFreeTier(DbWorkspace workspace) {
```
jaycarlton: thinking out loud: Currently, there's no capability or plan for our support to be able to give someone enough free credits to a customer with a billing account to get through a couple of days while they sort out a billing or config issue with Google? I.e. you can only ever get free credits on a free tier account.

calbach: That specific implementation solution is not something that's at all supported by GCP billing or by the billing model we're using.

Setting that aside, the way to resolve that use case in the current model would be to switch the user back to free tier. As currently proposed, the user would need to do this manually; or if they wait too long (and data is at risk of deletion), we will automatically switch them back.

api/db/changelog/db.changelog-116-add-billing-account-to-workspaces.xml
calbach: Having both columns sounds fine. I would not add any constraints that created dependencies between the config table and the backend data - I think we need to allow these to remain independent; the fact that we store our configs in SQL is mostly out of convenience rather than any logical requirement, and keeping them in lock-step may not always be correct (for example, consider projects pending assignment in the buffer).

I'm happy with either:
1. a billingAccountName and a billingType column, or
2. just userBillingAccountName (null if free tier)

@jaycarlton there is no proposed dependency here on naming conventions. local and test are the same account, the value is in the config JSONs if you want to reference it.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: Our implementation plan is to set user-specific overrides (for both cost and time) which will restore users to the free tier when set.

calbach: I don't think that matches my understanding of the plan. I think this is all outlined in the synthesis doc, but let me walk through this example to be clear.

Scenario: Workspace's user-provided billing account becomes inactive -> workspace becomes "billing locked"
Resolutions:
1. (if creator still has free credits) owner can switch the workspace to free tier
2. owner can fix the billing account
3. owner can attach a different billing account
4. owner does not resolve billing issues for 2 weeks: workspace reverts to free tier
4a. If the user still has free tier credits, this is now a normal, functional free tier workspace
4b. If the user has no credits, this is still a "billing locked" workspace; but one which is attached to the free tier billing account (and therefore, data will not be deleted)

(4b) is an identical end-state to the easier-to-follow scenario where a user has a free tier workspace and runs out of credits. The workspace is locked, but still incurs some service-level cost to free tier

calbach: also, overriding their free tier quota (on dollars or time), would not directly effect the free tier status of any of the workspaces, it would just change whether or not they are billing locked.

jmthibault79: Sorry, I was confused.  I was thinking of the free-tier-only flow.  Your response matches my understanding.

api/db/changelog/db.changelog-116-add-billing-account-to-workspaces.xml
jaycarlton: Thanks. I misunderstood. I was thinking one proposal involved inferring something from a name that was outside the database's control. Whether "not null" is a naming convention or not... 😉 

It's still a bit odd to me that we have baked in an assumption that the presence of an account always indicates paid tier and vice versa. It seems like the kind of thing that reality might make less convenient. 

Option #1 seems like it's clearer in the short term and gives more flexibility in the long term (exactly what CH said in the pending case).

api/db/changelog/db.changelog-118-add-billing-account-to-workspaces.xml
```diff
@@ -0,0 +1,16 @@
...
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="ericsong" id="db.changelog-118-add-billing-account-to-workspaces">
+    <addColumn tableName="workspace">
+      <column name="billing_account_name" type="varchar(100)">
+        <constraints nullable="false"/>
+      </column>
+    </addColumn>
+
+    <sql>update workspace set billing_account_name=(SELECT JSON_UNQUOTE(JSON_EXTRACT(configuration, '$.billing.accountId')) from config where config_id='main');</sql>
```
jaycarlton: If you want to be a bit more conservative (e.g. if you think perhaps some config objects might not have what you need), then you can use a temp table to hold {workspaceId, billing_acct} pairs and verify that looks good before committing a new column to the main table. I.e. if we get stuck, it's only this temporary (non-application) table that is in bad shape.

api/src/main/java/org/pmiops/workbench/db/model/DbWorkspace.java
```diff
@@ -569,6 +570,15 @@ public void setBillingStatus(BillingStatus billingStatus) {
     this.billingStatus = DbStorageEnums.billingStatusToStorage(billingStatus);
   }
 
+  @Column(name = "billing_account_name")
+  public String getBillingAccountName() {
```
jaycarlton: So this will always be a gcp project id?

api/src/main/java/org/pmiops/workbench/db/model/DbWorkspace.java
```diff
@@ -569,6 +570,15 @@ public void setBillingStatus(BillingStatus billingStatus) {
...
+    return billingAccountName;
+  }
+
+  public void setBillingAccountName(String billingAccountName) {
+    this.billingAccountName = billingAccountName;
+  }
+
```
jaycarlton: Re: the mapping between account names and project names, is that something guaranteed by the structure of one or the other name, or something we have to maintain with application code? Do all APIs that have access to change the name get checked against this invariant?

ericsong:
@jaycarlton going to try to answer all of your comments at once

The `billing_account_name` field will match the value that is returned by Google from this API call, https://cloud.google.com/billing/reference/rest/v1/projects/getBillingInfo. This is not the project ID, rather a property of a GCP project. The billing account name for a GCP project can change, allowing this through AoU is what I am currently working on.

We will be maintaining this field through application code. The source of truth of this value is with Google but we mirror it here for caching.

The JSON_EXTRACT call should be OK on all environments since they all have the value specified.

jaycarlton:
Thanks. That's reassuring. I misheard something at some point and thought we had this value concatenated with something else in one column.

PR #3012, opened 2020-01-17
[RISK=NO][No ticket]Add default value to creation_Time

NehaBroad:
This PR will add default values 'Jan 1 1970' TO user column creation_time and last_modified_time. 

In order for this to reflect on test i will rollback changes for existing 115 as well as remove the entry from databasechangelog in just this case

### DIFF
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -6,18 +6,12 @@
                     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified">
     <addColumn tableName="user">
-      <column name="creation_time" type="datetime">
+      <column name="creation_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
         <constraints nullable="false" />
       </column>
-      <column name="last_modified_time" type="datetime">
+      <column name="last_modified_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
         <constraints nullable="false"/>
       </column>
     </addColumn>
-  </changeSet>
-  <changeSet id="update creation time" author="nsaxena">
-    <sql>update user set creation_time = first_sign_in_time</sql>
-  </changeSet>
-  <changeSet id="update modified time" author="nsaxena">
-    <sql>update user set last_modified_time = first_sign_in_time</sql>
   </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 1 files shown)


api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -6,18 +6,12 @@
                     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified">
     <addColumn tableName="user">
-      <column name="creation_time" type="datetime">
```
jaycarlton: Has this already been run? If so, do we know it's safe to change it? We don't want to lose the ability to keep all the environments in lockstep.

In general, if I'm looking at the history of a liquibase xml file, I only ever expect to see changes when a new step is added, and the individual step xmls should only have one commit, right?

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -6,15 +6,12 @@
...
-        <constraints nullable="true" />
+      <column name="creation_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
+        <constraints nullable="false" />
       </column>
-      <column name="last_modified_time" type="datetime" defaultValueDate="CURRENT_TIMESTAMP">
-        <constraints nullable="false"/>
+      <column name="last_modified_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
```
ericsong: I believe the tag we want to use is `valueDate` https://www.liquibase.org/documentation/column.html

That will set the column value when the migration is run. I believe `defaultValueDate` will set the default value in MySQL which we shouldn't need since we should be setting it to the current timestamp on object creation (in our Java code).

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
ericsong: See the slack thread on why it's gotten so complicated but the TLDR on our approach is that we are going to manually rollback the 115 changset that ran on test so that this version can successfully execute.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
NehaBroad: i was finally able to resolve the script thing and test this, defaultValue results in 1970-01-01 00:00:00 to existing entries

NehaBroad: Screen Shot 2020-01-17 at 11 41 34 PM

jaycarlton:
I would not recommend using the current time as a default, as it's going to be different depending on when the migration is run. If you _have_ to have a non-null default here, then please pick a fixed date like Jan 1, 1980, which is a valid time and easy to query for when we do a later migration to clean it up.

PR #3013, opened 2020-01-18
[RISK=MEDIUM][no ticket] Add default value to user creation date and last modified time

NehaBroad:
This PR has 3 database script file:

1) 115 : Skip if USER TABLE have creation_time and last_modfied_time COLUMNS already else create the two columns with
- creation_time with not null 
-  last_modified_time with null constraint. and default value as current_timestamp
2) 116: DROP column creation_time and  LAST_MODIFIED_TIME if they exist in USER else skip
3) 117: Add columns creation_time and  LAST_MODIFIED_TIME  again with null constraints and defaultdate as 1970 Jan 1

The reason for three scripts are as follows:

1) Date with null constraint cannot be updated later with another default value as it starts giving liquibase base exception 
_liquibase.exception.LiquibaseException: Unexpected error running Liquibase: Data truncation: Incorrect datetime value: '0000-00-00 00:00:00' for column 'creation_time' at row 1 [Failed SQL: ALTER TABLE workbench.user ADD creation_time datetime NOT NULL, ADD last_modified_time datetime NOT NULL]_
2) On test 115 has already run and we do not have permission to ALTER table hence the rollback script ran partially removing the entry from DATABASECHANGELOG but unable to alter table hence we are rerunning 115 with preCondition to ignore onFail
3) Since test already has the columns we want to Default value too and we cannot add default values tag 116 will drop them (if they exist) and 117 will add proper values

### DIFF
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -6,18 +6,12 @@
                     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified">
     <addColumn tableName="user">
-      <column name="creation_time" type="datetime">
+      <column name="creation_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
         <constraints nullable="false" />
       </column>
-      <column name="last_modified_time" type="datetime">
+      <column name="last_modified_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
         <constraints nullable="false"/>
       </column>
     </addColumn>
-  </changeSet>
-  <changeSet id="update creation time" author="nsaxena">
-    <sql>update user set creation_time = first_sign_in_time</sql>
-  </changeSet>
-  <changeSet id="update modified time" author="nsaxena">
-    <sql>update user set last_modified_time = first_sign_in_time</sql>
   </changeSet>
 </databaseChangeLog>
--- a/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
+++ b/api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
@@ -11,11 +11,11 @@
       </not>
     </preConditions>
     <addColumn tableName="user">
-      <column name="creation_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
-        <constraints nullable="false" />
+      <column name="creation_time" type="datetime">
+        <constraints nullable="true" />
       </column>
-      <column name="last_modified_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
-        <constraints nullable="false" />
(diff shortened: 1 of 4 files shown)


api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -5,12 +5,17 @@
     xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
                     http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified-with-null-constraint-and-default">
+    <preConditions onFail="MARK_RAN" >
```
jaycarlton: No, we can't have this. The entire point of Liquibase is that it's idempotent, and if it's at a state, it got there by playing everything before it successfully.

Can we just restore the database from a backup and run a migration from there?

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
```diff
@@ -5,12 +5,17 @@
...
   <changeSet author="nsaxena" id="db.changelog-115-addColumn-user-create-modified-with-null-constraint-and-default">
+    <preConditions onFail="MARK_RAN" > 
+      <not>         
+        <columnExists tableName="user" columnName="creation_time" />
+      </not>
+    </preConditions>
     <addColumn tableName="user">
-      <column name="creation_time" type="datetime" defaultValueDate="1970-01-01 00:00:00">
-        <constraints nullable="false" />
+      <column name="creation_time" type="datetime">
```
jaycarlton: We should just delete this changeset and start at 120 if it failed. I think that's more correct. 115 should never be in the history for two different databases with different semantics.

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
NehaBroad: This was not the ideal solution we wanted to do however I had to do this just because i cannot run ALTER commands on test DB and we deleted the entry in databasechangelog for changeset

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
NehaBroad: what do you mean by 120?

NehaBroad:
Also @jaycarlton  do you want me to split the PRs as per the changeSets?

api/db/changelog/db.changelog-115-addColumn-user-create-modified.xml
jaycarlton: That's kinda what I figured.

jaycarlton: We will definitely need a way to obtain such access for the oncall engineer when needed.

jaycarlton:
Up to you. In theory they're still rollbackable, so if that's possible then maybe not. I'd do the minimum that gets us deployed again first. The advantage of splitting them would be to verify each change independently, but if those wouldn't deploy separately, then that won't help.

NehaBroad:
I would then go ahead and then push this for test we would need all three to work

PR #3014, opened 2020-01-20
[risk=no] removing warnings about unchecked method invocation

freemabd:
workbench/api/src/main/java/org/pmiops/workbench/concept/ConceptService.java:171: warning: [unchecked] unchecked method invocation: method findConcepts in interface ConceptDao is applied to given types
          ? conceptDao.findConcepts(conceptTypes, domains, VOCAB_ID, CONCEPT_CLASS_ID, pageable)
                                   ^
  required: ImmutableList<String>,ImmutableList<String>,String,String,Pageable
  found: ImmutableList<String>,ImmutableList,String,String,Pageable
/Users/jaycarlton/repos/workbench/api/src/main/java/org/pmiops/workbench/concept/ConceptService.java:171: warning: [unchecked] unchecked conversion
          ? conceptDao.findConcepts(conceptTypes, domains, VOCAB_ID, CONCEPT_CLASS_ID, pageable)

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/concept/ConceptService.java
+++ b/api/src/main/java/org/pmiops/workbench/concept/ConceptService.java
@@ -165,7 +165,7 @@ public class ConceptService {
     Pageable pageable = new PageRequest(page, limit, new Sort(Direction.DESC, "countValue"));
     ImmutableList<String> conceptTypes = getConceptTypes(standardConceptFilter);
     if (domainIds.contains(CommonStorageEnums.domainToDomainId(Domain.PHYSICALMEASUREMENT))) {
-      ImmutableList domains =
+      ImmutableList<String> domains =
           ImmutableList.of(CommonStorageEnums.domainToDomainId(Domain.MEASUREMENT));
       return StringUtils.isBlank(keyword)
           ? conceptDao.findConcepts(conceptTypes, domains, VOCAB_ID, CONCEPT_CLASS_ID, pageable)


api/src/main/java/org/pmiops/workbench/concept/ConceptService.java
```diff
@@ -165,7 +165,7 @@ public long countSurveys() {
     Pageable pageable = new PageRequest(page, limit, new Sort(Direction.DESC, "countValue"));
     ImmutableList<String> conceptTypes = getConceptTypes(standardConceptFilter);
     if (domainIds.contains(CommonStorageEnums.domainToDomainId(Domain.PHYSICALMEASUREMENT))) {
-      ImmutableList domains =
+      ImmutableList<String> domains =
           ImmutableList.of(CommonStorageEnums.domainToDomainId(Domain.MEASUREMENT));
       return StringUtils.isBlank(keyword)
```
jaycarlton: nit: It's a bit difficult to read ternary statements where each branch is a method call with a bunch of parameters (and only one different). Rather than put the ternary operator in the return statement, I'd probably declare `final matchExpression = StringUtils.isBlank(keyword) ? conceptTypes : keyword;` Then you can return the same statement either way.

freemabd: I might be misunderstanding your statement, but conceptTypes and keyword are not mutually exclusive. If keyword is blank then call:
conceptDao.findConcepts(conceptTypes, ...) otherwise call:
conceptDao.findConcepts(keyword, conceptTypes, ...)

PR #3015, opened 2020-01-20
[RW-4091][RISK=NO] Correct exportRDR cron URL

NehaBroad:
While testing i saw the following (silly) errors:
1) Cron yaml url doesnt case match the entry in workbench.yaml
2) String Equal was using == rather than .equal because of which the correct location was not being send.
3) Send Workspace Collaborators Status as ACTIVE to RDR

### DIFF
--- a/common-api/src/main/java/org/pmiops/workbench/config/WorkbenchLocationConfigService.java
+++ b/common-api/src/main/java/org/pmiops/workbench/config/WorkbenchLocationConfigService.java
@@ -23,8 +23,8 @@ public class WorkbenchLocationConfigService {
    */
   public String getCloudTaskLocationId() {
     String appEngineLocationId = workbenchConfig.get().server.appEngineLocationId;
-    if (appEngineLocationId == "us-central") return "us-central1";
-    else if (appEngineLocationId == "europe-west") return "europe-west1";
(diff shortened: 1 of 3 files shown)


common-api/src/main/java/org/pmiops/workbench/config/WorkbenchLocationConfigService.java
```diff
@@ -23,8 +23,8 @@ public String getWorkbenchAppEngineLocationId() {
    */
   public String getCloudTaskLocationId() {
     String appEngineLocationId = workbenchConfig.get().server.appEngineLocationId;
-    if (appEngineLocationId == "us-central") return "us-central1";
-    else if (appEngineLocationId == "europe-west") return "europe-west1";
+    if (appEngineLocationId.equals("us-central")) return "us-central1";
```
gjuggler: Side note: this kind of issue would have been easily caught by a unit test. Be sure to follow-up soon on developing unit tests, which was a condition of approval for https://github.com/all-of-us/workbench/pull/2948. (It might help to track a separate Jira ticket for tests follow-up, so we can include it in sprint pointing). I would recommend including this service as being in scope for unit test coverage.

PR #3018, opened 2020-01-21
[risk=low]Refactoring to make RW-3661 easier

jmthibault79:
Description:

Splitting off some independent work from the (forthcoming) main RW-3661 PR

---
**PR checklist**

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -436,6 +436,7 @@ dependencies {
   compile 'javax.inject:javax.inject:1'
   compile 'io.swagger:swagger-annotations:1.5.16'
   compile 'org.apache.commons:commons-lang3:3.6'
+  compile 'org.apache.commons:commons-text:1.8'
   compile 'com.google.guava:guava:26.0-jre'
   compile 'com.github.rholder:guava-retrying:2.0.0'
   compile 'org.springframework.retry:spring-retry:1.1.5.RELEASE'
--- a/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
@@ -110,8 +110,13 @@ public class FreeTierBillingService {
       deactivateUserWorkspaces(user);
     }
 
-    sendAlertsForCostThresholds(previousUserCosts, userCosts, currentExpiredUsers);
-    sendAlertsForTimeThresholds(userCosts, currentExpiredUsers);
+    final Set<DbUser> usersWithNonNullRegistration =
+        userDao.findByFirstRegistrationCompletionTimeNotNull();
+    final Set<DbUser> usersToThresholdCheck =
+        Sets.difference(usersWithNonNullRegistration, currentExpiredUsers);
+
+    sendAlertsForCostThresholds(usersToThresholdCheck, previousUserCosts, userCosts);
+    sendAlertsForTimeThresholds(usersToThresholdCheck, userCosts);
   }
 
   private boolean expiredByCost(final DbUser user, final double currentCost) {
@@ -133,16 +138,16 @@ public class FreeTierBillingService {
   }
 
   private void sendAlertsForCostThresholds(
+      Set<DbUser> usersToCheck,
       Map<DbUser, Double> previousUserCosts,
-      Map<DbUser, Double> userCosts,
-      Set<DbUser> currentExpiredUsers) {
+      Map<DbUser, Double> userCosts) {
     final List<Double> costThresholdsInDescOrder =
         workbenchConfigProvider.get().billing.freeTierCostAlertThresholds;
     costThresholdsInDescOrder.sort(Comparator.reverseOrder());
 
     userCosts.forEach(
         (user, currentCost) -> {
-          if (!currentExpiredUsers.contains(user)) {
+          if (usersToCheck.contains(user)) {
             final double previousCost = previousUserCosts.getOrDefault(user, 0.0);
             maybeAlertOnCostThresholds(user, currentCost, previousCost, costThresholdsInDescOrder);
           }
@@ -150,21 +155,19 @@ public class FreeTierBillingService {
   }
 
   private void sendAlertsForTimeThresholds(
-      Map<DbUser, Double> userCosts, Set<DbUser> currentExpiredUsers) {
+      Set<DbUser> usersToCheck, Map<DbUser, Double> userCosts) {
     final List<Double> timeThresholdsInDescOrder =
         workbenchConfigProvider.get().billing.freeTierTimeAlertThresholds;
     timeThresholdsInDescOrder.sort(Comparator.reverseOrder());
 
-    for (final DbUser user : userDao.findByFirstRegistrationCompletionTimeNotNull()) {
-      if (!currentExpiredUsers.contains(user)) {
-        final double currentCost = userCosts.getOrDefault(user, 0.0);
-        final double remainingDollarBalance = getUserFreeTierDollarLimit(user) - currentCost;
-        maybeAlertOnTimeThresholds(user, remainingDollarBalance, timeThresholdsInDescOrder);
+    for (final DbUser user : usersToCheck) {
+      final double currentCost = userCosts.getOrDefault(user, 0.0);
+      final double remainingDollarBalance = getUserFreeTierDollarLimit(user) - currentCost;
+      maybeAlertOnTimeThresholds(user, remainingDollarBalance, timeThresholdsInDescOrder);
 
-        // save current check time
-        user.setLastFreeTierCreditsTimeCheck(Timestamp.from(clock.instant()));
-        userDao.save(user);
-      }
+      // save current check time
+      user.setLastFreeTierCreditsTimeCheck(Timestamp.from(clock.instant()));
+      userDao.save(user);
     }
   }
 
--- a/api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
@@ -13,8 +13,8 @@ import javax.inject.Provider;
 import javax.mail.MessagingException;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
-import org.apache.commons.text.StringSubstitutor;
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.text.StringSubstitutor;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.google.CloudStorageService;
 import org.pmiops.workbench.mandrill.api.MandrillApi;
@@ -54,7 +54,7 @@ public class MailServiceImpl implements MailService {
   }
 
   @Override
-  public void sendBetaAccessRequestEmail(String userName) throws MessagingException {
+  public void sendBetaAccessRequestEmail(final String userName) throws MessagingException {
     WorkbenchConfig workbenchConfig = workbenchConfigProvider.get();
 
     MandrillMessage msg = new MandrillMessage();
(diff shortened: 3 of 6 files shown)


api/build.gradle
```diff
@@ -436,6 +436,7 @@ dependencies {
   compile 'javax.inject:javax.inject:1'
   compile 'io.swagger:swagger-annotations:1.5.16'
   compile 'org.apache.commons:commons-lang3:3.6'
+  compile 'org.apache.commons:commons-text:1.8'
```
jmthibault79: Includes `StringSubstitutor` which replaces the deprecated `StrSubstitutor`

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -110,8 +110,13 @@ public void checkFreeTierBillingUsage() {
...
     }
 
-    sendAlertsForCostThresholds(previousUserCosts, userCosts, currentExpiredUsers);
-    sendAlertsForTimeThresholds(userCosts, currentExpiredUsers);
+    final Set<DbUser> usersWithNonNullRegistration =
+        userDao.findByFirstRegistrationCompletionTimeNotNull();
+    final Set<DbUser> usersToThresholdCheck =
+        Sets.difference(usersWithNonNullRegistration, currentExpiredUsers);
+
+    sendAlertsForCostThresholds(usersToThresholdCheck, previousUserCosts, userCosts);
```
jmthibault79: This is a change: previously we would run this cost check for users with null registration times.

However it is not very meaningful because any user who actually logs into the system will not have a null registration time.

I decided to make this change because it makes the logic a little simpler.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -13,8 +13,8 @@
 import javax.mail.MessagingException;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
-import org.apache.commons.lang3.text.StrSubstitutor;
```
jmthibault79: this was deprecated

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
-  public void sendWelcomeEmail(String contactEmail, String password, User user)
+  public void sendWelcomeEmail(final String contactEmail, final String password, final User user)
       throws MessagingException {
-    try {
-      InternetAddress email = new InternetAddress(contactEmail);
-      email.validate();
-    } catch (AddressException e) {
-      throw new MessagingException("Email: " + contactEmail + " is invalid.");
-    }
-    MandrillMessage msg = buildWelcomeMessage(contactEmail, password, user);
```
jmthibault79: There was a lot of duplication in the `buildXMessage` and `buildXHtml` methods, so I made this process more generic.  I'll need to add 3 more of these for RW-3661.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+      final String contactEmail, final String msgHtml, final String subject)
+      throws MessagingException {
+    return new MandrillMessage()
+        .to(Collections.singletonList(validatedRecipient(contactEmail)))
+        .html(msgHtml)
+        .subject(subject)
+        .fromEmail(workbenchConfigProvider.get().mandrill.fromEmail);
+  }
+
+  private RecipientAddress validatedRecipient(final String contactEmail) throws MessagingException {
```
jmthibault79: new method to replace some common validation logic

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -104,6 +104,11 @@ public void setUp() {
     workbenchConfig = WorkbenchConfig.createEmptyConfig();
     workbenchConfig.billing.freeTierCostAlertThresholds = new ArrayList<>(Doubles.asList(.5, .75));
     workbenchConfig.billing.freeTierTimeAlertThresholds = new ArrayList<>(Doubles.asList(.5, .75));
+    workbenchConfig.billing.defaultFreeCreditsDollarLimit = 1000.0;
```
jmthibault79: set some defaults for tests

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -936,6 +933,7 @@ private void assertSingleWorkspaceTestDbState(
   private DbUser createUser(String email) {
     DbUser user = new DbUser();
     user.setUsername(email);
+    user.setFirstRegistrationCompletionTime(Timestamp.from(START_INSTANT));
```
jmthibault79: since we now only check if the user has a valid registration time

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+    final MandrillMessage msg =
+        buildMessage(contactEmail, msgHtml, "All of Us ID Verification Complete");
+    sendWithRetries(msg, String.format("BetaAccess Complete for %s", contactEmail));
+  }
+
+  private ImmutableMap<String, String> welcomeMessageSubstitutionMap(
+      final String password, final User user) {
+    final CloudStorageService cloudStorageService = cloudStorageServiceProvider.get();
+    return new ImmutableMap.Builder<String, String>()
+        .put("USERNAME", user.getPrimaryEmail())
```
jaycarlton: Can you define these string constants in an Enum like `EmailTemplateKey.USERNAME`?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+        .put("PASSWORD", password)
+        .put("URL", workbenchConfigProvider.get().admin.loginUrl)
+        .put("HEADER_IMG", cloudStorageService.getImageUrl("all_of_us_logo.png"))
```
jaycarlton: This is getting close to content management in a Java file. Ideally we'd allow authors and graphic designers a more accessible place to compose content.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+        .put("BULLET_1", cloudStorageService.getImageUrl("bullet_1.png"))
+        .put("BULLET_2", cloudStorageService.getImageUrl("bullet_2.png"))
+        .build();
+  }
+
+  private ImmutableMap<String, String> betaAccessSubstitutionMap(final String username) {
```
jaycarlton: nit: for private methods this is OK, but generally the guidance is to declare your function to return Map (for compatibility), even when using an ImmutableMap under the hood. Which seems scary.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+    final CloudStorageService cloudStorageService = cloudStorageServiceProvider.get();
+
+    final String action =
+        "login to the workbench via <a class=\"link\" href=\""
+            + workbenchConfigProvider.get().admin.loginUrl
+            + "\">"
+            + workbenchConfigProvider.get().admin.loginUrl
```
jaycarlton: You can just close the `<a/>` if you don't want to give separate anchor text from the href.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+        .to(Collections.singletonList(validatedRecipient(contactEmail)))
+        .html(msgHtml)
+        .subject(subject)
+        .fromEmail(workbenchConfigProvider.get().mandrill.fromEmail);
+  }
+
+  private RecipientAddress validatedRecipient(final String contactEmail) throws MessagingException {
     try {
-      InternetAddress email = new InternetAddress(contactEmail);
+      final InternetAddress email = new InternetAddress(contactEmail);
```
jaycarlton: I'd name this `contactInternetAddress`

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
jaycarlton: aside: We ought to revisit the user data model. We could have a separate registration table that holds . the start, end, repeat count, etc. This is a couple of degrees removed from the actual user's core state.

jmthibault79:
Basic contents of this PR:
* remove deprecatd `StrSubstitutor`
* reuse duplicated code in `MailServiceImpl`
* simplify the logic of `FreeTierBillingService`
* minor test cleanup for `FreeTierBillingServiceTest`

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
     sendWithRetries(msg, String.format("Welcome for %s", user.getName()));
   }
 
   @Override
-  public void sendBetaAccessCompleteEmail(String contactEmail, String username)
+  public void sendBetaAccessCompleteEmail(final String contactEmail, final String username)
       throws MessagingException {
+    final String msgHtml = buildHtml(BETA_ACCESS_RESOURCE, betaAccessSubstitutionMap(username));
+    final MandrillMessage msg =
+        buildMessage(contactEmail, msgHtml, "All of Us ID Verification Complete");
```
ericsong: mega nit: I feel like subject, body parameter order is more intuitive than body, subject

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+    sendWithRetries(msg, String.format("BetaAccess Complete for %s", contactEmail));
```
ericsong: `Beta Access`?

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+
+    return new ImmutableMap.Builder<String, String>()
+        .put("ACTION", action)
+        .put("BETA_ACCESS_REPORT", "approved for use")
+        .put("HEADER_IMG", cloudStorageService.getImageUrl("all_of_us_logo.png"))
+        .put("USERNAME", username)
+        .build();
+  }
+
+  private String buildHtml(final String resource, final ImmutableMap<String, String> replacementMap)
```
ericsong: nit: I like to relax the type for arguments down to the superclass that has the interface that I need. In this case, I'm guessing a `Map<String, String>` will do.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+      throws MessagingException {
+    final URL emailContent = Resources.getResource(resource);
```
ericsong: nit: my first thought on reading `emailContent` as a name was that it contained the email data.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+    final StringBuilder contentBuilder = new StringBuilder();
+
+    try {
+      Resources.readLines(emailContent, StandardCharsets.UTF_8)
```
ericsong: nit: I think we can 1 line this with something like

```
    return new StringSubstitutor(replacementMap)
        .replace(Resources.readLines(null, StandardCharsets.UTF_8)
            .stream()
            .collect(Collectors.joining("\n")));
```

TBD if that's actually more readable but maybe there's something in the middle that you like.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
```diff
@@ -69,29 +69,89 @@ public void sendBetaAccessRequestEmail(String userName) throws MessagingExceptio
...
+          .forEach(s -> contentBuilder.append(s).append("\n"));
+    } catch (IOException e) {
+      throw new MessagingException("Error reading in email");
+    }
+
+    return new StringSubstitutor(replacementMap).replace(contentBuilder.toString());
+  }
+
+  private MandrillMessage buildMessage(
```
ericsong: nit: I actually find it easier to just read the implementation of this function. There isn't too much abstraction here and if anything, I need to jump to a definition to find out what the arguments of `buildMessage` do. The implementation has it declared in a fairly succinct and readable way.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
jmthibault79: This is not mine - I'm not going to modify this.

api/src/main/java/org/pmiops/workbench/mail/MailServiceImpl.java
jmthibault79: This is not mine so I'm not going to change it

PR #3019, opened 2020-01-21
[RW-4279][RW-4293][Risk=low] Add db fields for demographic survey

blrubenstein:
Description:
This adds `professionalUrl` to our DB, and changes gender identity and sexual orientation options to free text with a boolean identifier, per new caps requirements. This does not address the styling on the account creation survey page, which will be done in a follow up task. 

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
@@ -193,11 +193,6 @@ public class RdrExportServiceImpl implements RdrExportService {
             RdrExportEnums.sexAtBirthToRdrSexAtBirth(
                 dbDemographicSurvey.getSexAtBirthEnum().get(0)));
       }
-      if (dbDemographicSurvey.getSexualOrientationEnum() != null) {
-        researcher.setSexualOrientation(
-            RdrExportEnums.sexualOrientationToRdrSexualOrientation(
-                dbDemographicSurvey.getSexualOrientationEnum().get(0)));
-      }
       researcher.setDisability(
           RdrExportEnums.disabilityToRdrDisability(dbDemographicSurvey.getDisabilityEnum()));
     }
@@ -209,14 +204,6 @@ public class RdrExportServiceImpl implements RdrExportService {
               .collect(Collectors.toList()));
     } else {
       researcher.setRace(new ArrayList<>());
-    }
-    if (null != dbDemographicSurvey && dbDemographicSurvey.getGenderEnum() != null) {
-      researcher.setGender(
-          dbDemographicSurvey.getGenderEnum().stream()
-              .map(dbUserGender -> RdrExportEnums.genderToRdrGender(dbUserGender))
-              .collect(Collectors.toList()));
-    } else {
-      researcher.setGender(new ArrayList<>());
     }
     researcher.setAffiliations(
         dbUser.getInstitutionalAffiliations().stream()
--- a/ui/src/app/pages/login/account-creation/account-creation-survey.tsx
+++ b/ui/src/app/pages/login/account-creation/account-creation-survey.tsx
@@ -13,7 +13,7 @@ import {
   Race,
   SexAtBirth,
 } from 'generated/fetch';
-import {Section} from './account-creation';
+import {Section, TextInputWithLabel} from './account-creation';
 
 import {signedOutImages} from 'app/pages/login/signed-out-images';
 import {
@@ -109,16 +109,20 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
   }
 
   toggleDisability(value) {
-    this.setState(fp.set(['profile', 'demographicSurvey', 'disability'], value));
+    this.updateDemographicAttribute('disability', value);
+  }
+
+  toggleIdentifiesAsLgbtq(value) {
+    this.updateDemographicAttribute('identifiesAsLgbtq', value);
   }
 
   updateDemographicAttribute(attribute, value) {
     this.setState(fp.set(['profile', 'demographicSurvey', attribute], value));
   }
 
-  createOptionCheckbox(checkboxLabel: string, optionKey: string, optionValue: any) {
+  createOptionCheckbox(checkboxLabel: string, optionKey: string, optionValue: any, key: string) {
     return <CheckBox label={checkboxLabel}
-                     style={styles.checkbox}
+                     style={styles.checkbox} key={key}
                      wrapperStyle={styles.checkboxWrapper} labelStyle={styles.checkboxLabel}
                      onChange={(value) => this.updateList(optionKey, optionValue)}
     />;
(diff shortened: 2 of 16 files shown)


ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -111,16 +109,20 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
   }
 
   toggleDisability(value) {
```
gjuggler: Nit: methods named "toggle" generally mean "switch from one state to another every time I'm called", while this method and the one below simply update a field to a given boolean value. A slightly clearer name may be "updateDisability" or "setDisability".

gjuggler: Typing nit: it's never too late to start incorporating some type-safety here, e.g. value --> value: Boolean

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -138,70 +140,75 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
       </ListPageHeader>
 
       {/*Race section*/}
-      <Section header='1. Race'>
+      <Section header='Race'>
         <div style={{display: 'flex', justifyContent: 'flex-start', flexWrap: 'wrap'}}>
           {AccountCreationOptions.race.map((race) => {
-            return this.createOptionCheckbox(race.label, 'race', race.value);
+            return this.createOptionCheckbox(race.label, 'race', race.value, race.value.toString());
```
gjuggler: What impact does this change have?

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -138,70 +140,75 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
...
-        </FlexRow>
+      <Section header='Do you identify as lesbian, gay, bisexual, transgender, queer (LGBTQ),
+or another sexual and/or gender minority?'>
+        <div style={{paddingTop: '0.5rem'}}>
+          <RadioButton onChange={
+            (e) => this.toggleIdentifiesAsLgbtq(true)}
+                       checked={demographicSurvey.identifiesAsLgbtq}
+                       style={{marginRight: '0.5rem'}}/>
+          <label style={{paddingRight: '3rem', color: colors.primary}}>
+            Yes
```
gjuggler: Nit: can you make the <label> fit on one line to be consistent with the "no" label? (If it got auto-wrapped, that's fine, but by eye it looks like it might fit.

api/db/changelog/db.changelog-118-update-demographics.xml
```diff
@@ -0,0 +1,22 @@
...
+    <addColumn tableName="user">
+      <column name="professional_url" type="varchar(255)">
+        <constraints nullable="true"/>
+      </column>
+    </addColumn>
+    <addColumn tableName="demographic_survey">
+      <column name="identifies_as_lgbtq" type="boolean">
+        <constraints nullable="true"/>
+      </column>
+      <column name="lgbtq_identity" type="varchar(255)">
```
gjuggler: [Just thinking out loud] Have we encountered scenarios where users attempt to put more text into an input form than our database column width allows? Do we have clear DB design guidelines around choosing between the various text types or lengths for free-form text like this? I suspect 255 is probably enough here, I'm just curious if there is a broader pattern.

(Based on a quick dig into our codebase, it seems we use varchar more often by far, with one instance of type="longtext" and a few instances of type="text" in the data dictionary schema. It's a clear pattern, but I'm not sure the motivation behind preferring varchar more often than not.)

api/db/changelog/db.changelog-118-update-demographics.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
+                    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">
+  <changeSet author="srubenst" id="db.changelog-116-update-demographics">
```
gjuggler: It might be worth adding a comment here, and/or in the PR description, referencing https://precisionmedicineinitiative.atlassian.net/browse/RW-4312 (thanks for filing!). I also marked that ticket as blocked by RW-4293 for clarity.

blrubenstein: Threw a comment in there!

api/db/changelog/db.changelog-118-update-demographics.xml
blrubenstein: We haven't. I can add some front-end validation on that though. I asked Lou and it sounded like we thought 255 char was probably enough for this. I've been mainly following the DB schema pattern we have established.

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
blrubenstein: I may actually just get rid of these and use the `updateDemographicAttribute` cause its really the exact same thing.

blrubenstein: Will do, if I don't get rid of these methods.

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
blrubenstein: It removes a console error message. If you don't provide keys to react within a map, it complains with red in the console.

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
blrubenstein: Probably! Let me try.

blrubenstein:
@gjuggler the discussion happened on the invision mocks. I can add the mocks to the PR description. They are linked in the parent ticket. (The tickets attached here are subtasks of the ticket with most of the description)

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -193,11 +193,6 @@ private RdrResearcher toRdrResearcher(DbUser dbUser) {
             RdrExportEnums.sexAtBirthToRdrSexAtBirth(
                 dbDemographicSurvey.getSexAtBirthEnum().get(0)));
       }
-      if (dbDemographicSurvey.getSexualOrientationEnum() != null) {
```
als364: Is it an AC to not export this anymore?

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -138,70 +144,73 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
...
+          </label>
+          <RadioButton onChange={(e) => this.updateDemographicAttribute('identifiesAsLgbtq', false)}
+                       checked={!demographicSurvey.identifiesAsLgbtq}
+                       style={{marginRight: '0.5rem'}}/>
+          <label style={{color: colors.primary}}>No</label>
+        </div>
+        <label></label>
+        <TextInputWithLabel labelText='If yes, please tell us about your LGBTQ+ identity'
```
als364: Given that this is how we're wording it, I agree with Greg's earlier comment that this could be much larger than 255 characters. I'd bump it all the way up to 4000 because storage is cheap and a lot of the time it will be empty anyways.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
blrubenstein: We aren't capturing this data in enum field anymore. I think there's a conversation with the RDR team about importing the new fields that @NehaBroad was leading.

PR #3025, opened 2020-01-23
[RW-4091][RISK=NO]Change content header for cloud task request TO APPLCIATION/JSON

NehaBroad:
The cloud task was sending  application/octect-stream content-type header which was rejected by our application Hence changing the type to application/json. Since this is urgent the changes to yaml file and method file will be taken care in the next PR this PR will block the RDR flow to start sending researcher and workspace data to RDR

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
@@ -34,7 +34,7 @@ public class OfflineRdrExportController implements OfflineRdrExportApiDelegate {
   private RdrExportService rdrExportService;
   private WorkbenchLocationConfigService locationConfigService;
 
-  private final String BASE_PATH = "/v1/queueTask";
+  private final String BASE_PATH = "/v1/cloudTask";
   private final String EXPORT_RESEARCHER_PATH = BASE_PATH + "/exportResearcherData";
   private final String EXPORT_USER_PATH = BASE_PATH + "/exportWorkspaceData";
   private final String IDS_STRING_SPLIT = ", ";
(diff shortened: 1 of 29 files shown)


api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -108,18 +109,20 @@ private void groupIdsAndPushTask(List<Long> idList, String taskUri) {
...
-              .setAppEngineHttpRequest(
-                  AppEngineHttpRequest.newBuilder()
-                      .setRelativeUri(taskUri)
-                      .setBody(ByteString.copyFrom(commaSepList.getBytes()))
-                      .setHttpMethod(HttpMethod.POST)
-                      .build());
+      AppEngineHttpRequest req =
+          AppEngineHttpRequest.newBuilder()
+              .setRelativeUri(taskUri)
+              .setBody(ByteString.copyFromUtf8(daysJson))
```
blrubenstein: Is there a reason to do this ByteString conversion? With us moving to JSON, I would expect this to pass in a `json.toString` style thing, rather than a bytestring conversion? Do we need to do this for the AppEngineHttpRequest? (Or will it implicitly handle the conversion from string to encoded string?)

(Also, do you know where the docs for this are? All I've found is https://developers.google.com/resources/api-libraries/documentation/cloudtasks/v2/java/latest/com/google/api/services/cloudtasks/v2/model/AppEngineHttpRequest.html#setBody-java.lang.String- which doesn't have a `putHeaders` method or a builder method)

NehaBroad: The documentation for appenginetaTask header is pretty bad or not up to date. I reffered to https://cloud.google.com/appengine/docs/standard/python/taskqueue/push/creating-handlers however i could not get a clear way to set header as it is immutable object. After going thru the source code for cloud task i could find this method putHeader

NehaBroad: Regarding the type, cloud task accept the Body type to be ByteString

PR #3032, opened 2020-01-23
[risk=no][RW-2957] Moodle v2 API Integration

als364:
Our Moodle integration is going to release a new API endpoint specific to AoU RW. This PR introduces functionality, behind a feature flag, to call that endpoint, and deprecates the calls to the old endpoint. This functionality has been manually tested against the Moodle test server, including tests for scheduled and unscheduled expiry of a training.

--

**PR checklist**

  - https://precisionmedicineinitiative.atlassian.net/browse/RW-4316
  - https://precisionmedicineinitiative.atlassian.net/browse/RW-4317


api/config/config_local.json
```diff
@@ -74,7 +74,7 @@
     "enableMoodleBackend": true
   },
   "access": {
-    "enableComplianceTraining": false,
+    "enableComplianceTraining": true,
```
als364: Enabled this so I can actually test these changes. 'bypass all' should still work.

api/config/config_test.json
```diff
@@ -84,7 +84,8 @@
     "unsafeAllowDeleteUser": true,
```
jaycarlton: Why not also enable `enableComplianceTraining` on test?

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -47,7 +47,11 @@ public String readMandrillApiKey() {
 
   @Override
   public String getMoodleApiKey() {
-    return readCredentialsBucketString("moodle-key.txt");
+    if (configProvider.get().featureFlags.enableMoodleV2Api) {
```
als364: names up for debate. it's easy enough to change this, but the keys are different and I do have to go push files around in GCS to change the name.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceService.java
```diff
@@ -22,5 +24,16 @@
    * @return list of badges/completed training by user
    * @throws ApiException
    */
-  List<BadgeDetails> getUserBadge(int userMoodleId) throws ApiException;
+  List<BadgeDetailsDeprecated> getUserBadge(int userMoodleId) throws ApiException;
```
jaycarlton: Why did we move to the deprecated class here? Should this method be marked deprecated as well?

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -84,4 +92,33 @@ public Integer getMoodleId(String email) throws ApiException {
     }
     return response.getBadges();
   }
+
+  @Override
+  public Map<String, BadgeDetails> getUserBadgesByName(String email) throws ApiException {
+    if (!enableMoodleCalls()) {
+      return null;
```
jaycarlton: Can you please return an empty map instead of null? Otherwise, the next best thing is `Optional<Map>`, which is no less awkward. In general, if I call a method returning a collection in Java, I don't expect nor do I check for `null`.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -84,4 +92,33 @@ public Integer getMoodleId(String email) throws ApiException {
...
+    }
+
+    UserBadgeResponse response =
+        moodleApiProvider.get().getMoodleBadge(RESPONSE_FORMAT, getToken(), email);
+    if (response.getException() != null && response.getException().equals(MOODLE_EXCEPTION)) {
+      if (response.getErrorcode().equals(MOODLE_USER_NOT_ALLOWED_ERROR_CODE)) {
+        throw new ApiException(HttpStatus.NOT_FOUND.value(), response.getMessage());
```
jaycarlton: Please log this at a WARNING level (or at least INFO), so that we can track this down.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -555,46 +557,71 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
 
-      List<BadgeDetails> badgeResponse = complianceService.getUserBadge(moodleId);
-      // The assumption here is that the User will always get 1 badge which will be AoU
-      if (badgeResponse != null && badgeResponse.size() > 0) {
-        BadgeDetails badge = badgeResponse.get(0);
-        Timestamp badgeExpiration =
-            badge.getDateexpire() == null
-                ? null
-                : new Timestamp(Long.parseLong(badge.getDateexpire()));
+      String expiryEpoch = null;
```
jaycarlton: This function is over a hundred lines long now. Can we split it into 4-5 pieces? I generally consider 20-line functions in need of a haircut.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -555,46 +557,71 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
+      } else {
+        List<BadgeDetailsDeprecated> badgeResponse = complianceService.getUserBadge(moodleId);
+        // The assumption here is that the User will always get 1 badge which will be AoU
+        if (badgeResponse != null && badgeResponse.size() > 0) {
+          BadgeDetailsDeprecated badge = badgeResponse.get(0);
+          expiryEpoch = badge.getDateexpire();
+        } else {
+          // Moodle has returned zero badges for the given user -- we should clear the user's
+          // training completion & expiration time.
+          dbUser.setComplianceTrainingCompletionTime(null);
```
jaycarlton: Please add a `clearComplianceTrainingCompletion()` and `clearComplianceTrainingExpiration()` methods on `DbUser` so you don't have to pass `null` as your only argument. This will also help sweep the nullability of the timestamp under the rug in the entity class, so at the service level we don't need to think about it.

api/src/main/generated/org/pmiops/workbench/utils/WorkspaceMapperImpl.java
```diff
@@ -0,0 +1,129 @@
...
+import org.pmiops.workbench.db.model.DbWorkspace;
+import org.pmiops.workbench.firecloud.model.FirecloudWorkspace;
+import org.pmiops.workbench.model.ResearchPurpose;
+import org.pmiops.workbench.model.SpecificPopulationEnum;
+import org.pmiops.workbench.model.UserRole;
+import org.pmiops.workbench.model.Workspace;
+import org.pmiops.workbench.utils.mappers.CommonMappers;
+import org.springframework.stereotype.Component;
+
+@Generated(
```
jaycarlton: The mapper impl files are generated at compile time, and should not be checked into source control.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -582,46 +584,71 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
+              badge.getGlobalexpiration() == null
+                  ? badge.getDateexpire()
+                  : badge.getGlobalexpiration();
+        } else {
+          // Moodle has not returned research ethics training information for the given user --
+          // we should clear the user's training completion and expiration time.
+          dbUser.setComplianceTrainingCompletionTime(null);
+          dbUser.setComplianceTrainingExpirationTime(null);
+        }
+      } else {
```
jaycarlton: There's lots of nesting and mutable state going on here, which makes it easy to write bugs.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
```diff
@@ -47,7 +47,12 @@ public String readMandrillApiKey() {
 
   @Override
   public String getMoodleApiKey() {
-    return getCredentialsBucketString("moodle-key.txt");
+    if(configProvider.get().featureFlags.enableMoodleV2Api) {
+      return getCredentialsBucketString("moodle.txt");
```
jaycarlton: Please put these constants into a config file, and also rename them to indicate v1 vs v2 somehow.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -582,46 +584,71 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
-        BadgeDetails badge = badgeResponse.get(0);
-        Timestamp badgeExpiration =
-            badge.getDateexpire() == null
-                ? null
-                : new Timestamp(Long.parseLong(badge.getDateexpire()));
+      String expiryEpoch = null;
+      if (configProvider.get().featureFlags.enableMoodleV2Api) {
+        Map<String, BadgeDetails> userBadgesByName = complianceService.getUserBadgesByName(email);
+        if (userBadgesByName.containsKey(complianceService.getResearchEthicsTrainingField())) {
+          BadgeDetails badge =
```
jaycarlton: I'm wondering why the badge- and moodle-specific logic needs to live in the UserService.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -582,46 +584,71 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
+          // we should clear the user's training completion and expiration time.
+          dbUser.setComplianceTrainingCompletionTime(null);
+          dbUser.setComplianceTrainingExpirationTime(null);
+        }
+      } else {
+        List<BadgeDetailsDeprecated> badgeResponse = complianceService.getUserBadge(moodleId);
+        // The assumption here is that the User will always get 1 badge which will be AoU
+        if (badgeResponse != null && badgeResponse.size() > 0) {
+          BadgeDetailsDeprecated badge = badgeResponse.get(0);
+          expiryEpoch = badge.getDateexpire();
```
jaycarlton: Please make `expiryEpoch` an `Optional`, since it's not guaranteed to be there. You can do that with `badgeResponse.stream().findFirst()`, so you get rid of the scary `get(0)`.

api/src/main/generated/org/pmiops/workbench/utils/WorkspaceMapperImpl.java
als364: something is clearly broken with gitignore.

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
```diff
@@ -1,14 +1,20 @@
 package org.pmiops.workbench.compliance;
 
+import java.util.HashMap;
```
jaycarlton: Hmm. Somehow, the workspace mapper impl file lives in a `genenrated` dir, which explians why it doesn't get picked up: `workbench/api/build/generated/sources/annotationProcessor/java/main/org/pmiops/workbench/utils/WorkspaceMapperImpl.java`

I don't believe we want to . mask `*Impl.java`, as it would match lots fo our implementation services.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -646,6 +648,78 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
+      }
+      else {
+        // There is no record of this person having taken the training.
+        dbUser.clearComplianceTrainingCompletionTime();
+        dbUser.clearComplianceTrainingExpirationTime();
+      }
+
+      return updateUserWithRetries(
+          u -> {
+            u.setComplianceTrainingExpirationTime(u.getComplianceTrainingExpirationTime());
```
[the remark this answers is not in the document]
als364: `dbUser` is `u` here. I just kind of cargo culted this block over from the last function.

Looking at `updateUserWithRetries` now, it applies the function passed to it to the user passed to it and also sets `lastModifiedTime` and enforces data access level changes as a side effect of aforementioned function.

These Moodle functions are the only place we do something like this - everywhere else we pass in a value unattached to the `dbUser` we're changing.  I think that, as originally intended, this function was mean to apply changes calculated elsewhere and we're just mutating `dbUser` so as to not have to think about nullity semantics

What do you think about something like:
```
Timestamp newComplianceTrainingCompletionTime = null;
Timestamp newComplianceTrainingExpirationTime = null;
{...}
if (badge valid) {
    if (badge was previously invalid and is now valid) {
        newComplianceTrainingCompletionTime = now
    } else if (old expiry time is not new expiry time) {
        newComplianceTrainingCompletionTime = now
    }
    newComplianceTrainingExpirationTime = badgeExpiryDate
}
// get rid of all the lines nulling out timestamps and get rid of those 'clear' functions
{...}
return updateUserWithRetries(u -> {
    u.setComplianceTrainingCompletionTime(newComplianceTrainingCompletionTime);
    u.setComplianceTrainingExpirationTime(newComplianceTrainingExpirationTime);
}, dbUser);

File: api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java [hunk 16]
```diff
@@ -646,6 +648,78 @@ public DbUser syncComplianceTrainingStatus(DbUser dbUser)
...
+            u.setComplianceTrainingCompletionTime(u.getComplianceTrainingCompletionTime());
+            return u;
+          },
+          dbUser);
+    } catch (NumberFormatException e) {
+      log.severe("Incorrect date expire format from Moodle");
+      throw e;
+    } catch (org.pmiops.workbench.moodle.ApiException ex) {
+      if (ex.getCode() == HttpStatus.NOT_FOUND.value()) {
+        log.severe(
```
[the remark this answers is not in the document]
als364: I'll put a different log but this one is for a specific reason, the message is just unclear

api/src/main/java/org/pmiops/workbench/compliance/ComplianceServiceImpl.java
als364: What is this comment in reference to? I don't think it is actually about importing `HashMap`

api/src/main/java/org/pmiops/workbench/compliance/ComplianceService.java
```diff
@@ -22,5 +24,16 @@
...
+  List<BadgeDetailsDeprecated> getUserBadgeDeprecated(int userMoodleId) throws ApiException;
+
+  /**
+   * Get details about the Research Ethics Training and the Data Use Agreement badges for a user
+   *
+   * @param email
+   * @return map of badge name to badge details
+   * @throws ApiException
+   */
+  Map<String, BadgeDetails> getUserBadgesByName(String email) throws ApiException;
```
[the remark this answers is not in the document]
als364: I'm happy being verbose as long as I can tab-complete names

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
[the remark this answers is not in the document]
gjuggler: 👍 That approach seems much better, especially if it's more consistent with how we do it elsewhere.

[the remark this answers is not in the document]
jaycarlton: nit: This still feels more badgey than usery, and like the kind of thing that could go in its own little service. While we have a very wide data model for the user table, we don't have to make its service know everything about how to calculate those columns.

FWIW, there are a handful of things whose recalculation is triggered on every user "change", and that's one of the main motivations of the `withRetries()` stuff.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
jaycarlton: We don't yet have application config patterns, which is where these might need to go, or a parent JSON file that individual environments could override. I would still rename the credentials files `moodle-key-v1.txt` and `moodle-key-v2.txt` (or leave the current one alone).

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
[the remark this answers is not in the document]
als364: I'd be ok with moving it, but I'd rather do so after deprecating the old call so that I don't have to either have the v1 and v2 in separate places or move the deprecated function.

api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java
als364: Because we push these creds manually to google buckets I have to leave the current one alone for the meantime. I can rename the new one v2 though.

api/config/config_local.json
```diff
@@ -73,10 +73,12 @@
   },
   "moodle": {
     "host": "aoudev.nnlm.gov",
-    "enableMoodleBackend": true
+    "enableMoodleBackend": true,
+    "credentialsKeyDeprecated": "moodle-key.txt",
```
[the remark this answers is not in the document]
gjuggler: Coming back to this: post-renaming, I don't feel as strongly about this. Given we'll rip most of this stuff out once the v2 has launched to prod, that gives us a chance to reassess this pattern in a ~month's time anyway.

PR #3035, opened 2020-01-23
[RW-3175][risk=low] User provided billing accounts

ericsong:
I may have gone a little overboard in exception checking but I wanted to make it highly unlikely for a user to unknowingly have their billing account associated with a dead billing project in the case of failed calls.

Note - this PR does not include the ability to switch a workspace back to the free tier. That will come in a follow up.

TIL
- We are able to use the User client's credentials from the backend. This is done by grabbing their OAuth tokens from the request headers and creating new API clients that use it. We typically set this up through Spring dependency injection

- doAnswer(invocation -> ()) is a powerful way to mock external dependencies. I use this for the Cloudbilling client to emulate API calls with dynamic responses.

- it is possible to verify multiples calls to Mockito mocks as well the arguments passed into each call using ArgumentCaptor

- **@SpyBean annotation can be used to wrap an autowired dependency with a Mockito spy. This is super useful is verifying calls or overriding behavior of classes that are difficult to mock like the Daos.

- modules imported into UI tests can be mocked using jest

- using a new API from a React component will typically involve adding a Stub into its test file

- setting the property on a nested state object will require a speial setter function. We leverage fp.set() to do this

image

---
**PR checklist**


ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -450,6 +465,22 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
         this.setState(fp.set(['workspace', 'cdrVersionId'], cdrResp.defaultCdrVersionId));
       }
+
+      if (serverConfigStore.getValue().enableBillingLockout) {
+        userApi().listBillingAccounts().then(response =>
+          this.setState({billingAccounts: response.billingAccounts})
+        );
+
+        if (this.isMode(WorkspaceEditMode.Edit)) {
+          getBillingAccountName(this.props.workspace.namespace).then(billingAccountName =>
```
calbach: Some cases to make sure are handled here:

1. The user is an editor (not an owner). We expect this call to 403.
2. The user is an owner, but doesn't have access to the billing account (i.e. the workspace billing account is not present in the dropdown).

ui/src/app/utils/workbench-gapi-client.ts
```diff
@@ -0,0 +1,13 @@
+declare const gapi: any;
+
+export async function getBillingAccountName(workspaceNamespace: string) {
```
calbach: This seems slightly overfit to me. I'd probably have it return the full JSON response structure, then have the client pick out `billingAccountName`

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -201,6 +209,10 @@ export const specificPopulations = [
   }
 ];
 
+const billingDescription = <div>The <i>All of Us</i> Program provides $200 in free credits per user.
```
calbach: Noting we'll need to keep this in sync with the actual value. Probably fine for now

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -244,12 +252,76 @@ private FirecloudWorkspace attemptFirecloudWorkspaceCreation(FirecloudWorkspaceI
 
     dbWorkspace.setBillingMigrationStatusEnum(BillingMigrationStatus.NEW);
 
-    dbWorkspace = workspaceService.getDao().save(dbWorkspace);
+    updateWorkspaceBillingAccount(dbWorkspace, workspace.getBillingAccountName());
+
+    try {
+      dbWorkspace = workspaceService.getDao().save(dbWorkspace);
+    } catch (Exception e) {
+      // Tell Google to set the billing account back to the free tier if the workspace creation
```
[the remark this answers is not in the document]
ericsong: If I am saving the workspace with the updates billing account, I believe I still need to make a try/catch call to update the db if the Google call fails.

Proposed
1. save db w/ new billing account (succeeds)
2. call google (fails)
3. save db w/ old billing account to keep state in sync OR delete the workspace b/c the create call didn't fully complete successfully.

Current
1. call google w/ new billing account (succeeds)
2. save db (fails)
3. call google w/ old billing account to keep state in sync

My reasoning for going with the current workflow was that I felt like a db save call is less likely to fail than an API call. There's also the effect of having a partially successful workspace entry in our database. I could see a case for not failing the entire call if the updateBillingInfo call fails since the entire call is quite expensive but I view that more of a feature add on that we can talk to product about.

calbach:
I can't find your comment on the PR, but I think your rationale sounds good. My only concern is of the worst case situation where:
1. attach billing account
2. server explodes

This results in a project which has the user's billing account attached, but which they cannot access from the Workbench UI. I think if the ASSIGNING buffer cleanup job also forced the billing account back to free tier - that would address this.

ui/src/app/pages/workspace/workspace-edit.tsx
ericsong: It looks like the user is able to call `getBillingInfo` on the project when they're a writer but they are NOT able to update the billing info. Is this OK? Fixing the permissions so that they can't call `getBillingInfo` might be non trivial since I think that is managed on FC's side based on how they interpret the writer role.

From AoU's standpoint, they will not be able to update the workspace since only owners can edit Workspace details. We can also just remove the dropdown for writers if we do not want them to be able to see the billing info.

With regards to #2, the logic here got a bit tricky but I think I have something reasonable and there's enough comments that it should be readable. LMK what you think.

calbach:
Nevermind, this actually wouldn't cover it since the project moves to assigned before the workspace gets created. I think this case should be quite rare, perhaps just add a fair amount of severe logging when the revert fails, and we can ensure our oncall process involves following up on those.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -416,40 +464,15 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
-            ...this.props.workspace.researchPurpose,
-            populationDetails: !this.props.workspace.researchPurpose.populationDetails ?
-              [] : this.props.workspace.researchPurpose.populationDetails,
-            diseaseOfFocus: !this.props.workspace.researchPurpose.diseaseOfFocus ?
-              '' : this.props.workspace.researchPurpose.diseaseOfFocus}
-        }});
+        };
+      }
 
-        if (this.isMode(WorkspaceEditMode.Duplicate)) {
```
calbach: Where did this go?

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -377,9 +373,61 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
...
         showStigmatizationDetails: false,
+        billingAccounts: []
       };
     }
 
+    async componentDidMount() {
+      if (serverConfigStore.getValue().enableBillingLockout) {
+        const billingAccounts = (await userApi().listBillingAccounts()).billingAccounts;
+
+        if (this.isMode(WorkspaceEditMode.Edit)) {
```
calbach: In the clone case, we should just default to the free tier billing account.

ui/src/app/pages/workspace/workspace-edit.tsx
calbach: Yes, default makes sense. It would also be very confusing to explain which user this is referring to (since it depends on the creator, not necesarily the current user's free tier credits).

ui/src/app/pages/workspace/workspace-edit.tsx
calbach: Only owners should be able to change the billing information, so the dropdown should ideally just be disabled for non-owners.

ui/src/app/pages/workspace/workspace-edit.tsx
```diff
@@ -377,9 +373,61 @@ export const WorkspaceEdit = fp.flow(withRouteConfigData(), withCurrentWorkspace
         loading: false,
         showUnderservedPopulationDetails: false,
         showStigmatizationDetails: false,
+        billingAccounts: []
       };
     }
 
+    async componentDidMount() {
+      if (serverConfigStore.getValue().enableBillingLockout) {
```
calbach: nit: pretty heavy nesting. I might at least break out this first case into an inverted flag check + early return to eliminate one level.

ericsong: I pulled it out into a function and I think that helped a little. I tried shuffling around the stuff inside but couldn't find a format that I like more than what I already have.

api/src/main/resources/workbench.yaml
```diff
@@ -2789,6 +2811,33 @@ definitions:
...
+      - displayName
+      - isOpen
+    properties:
+      isFreeTier:
+        type: boolean
+      name:
+        type: string
+      displayName:
+        type: string
+      isOpen:
```
jmthibault79: Could you add a comment about what it means for a Billing Account to be "open" ?

api/src/main/java/org/pmiops/workbench/billing/GoogleApisConfig.java
```diff
@@ -0,0 +1,39 @@
...
+public class GoogleApisConfig {
+
+  @Bean
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
+  public Cloudbilling googleCloudBillingApi(
+      UserAuthentication userAuthentication,
+      JsonFactory jsonFactory,
+      Provider<WorkbenchConfig> workbenchConfigProvider)
+      throws GeneralSecurityException, IOException {
+    GoogleCredential credential =
```
jmthibault79: `GoogleCredential` is deprecated.  Can this be rewritten with `GoogleCredentials` ?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -244,12 +252,80 @@ private FirecloudWorkspace attemptFirecloudWorkspaceCreation(FirecloudWorkspaceI
...
+                  "projects/" + workspace.getWorkspaceNamespace(),
+                  new ProjectBillingInfo().setBillingAccountName(newBillingAccountName));
+
+      ProjectBillingInfo response;
+
+      try {
+        // this is necessary because the grant ownership call in create/clone
+        // may not have propagated. Adding a few retries drastically reduces
+        // the likely of failing due to slow propagation
+        response = cloudBillingRetryer.call(() -> request.execute());
```
jmthibault79: can be `.call(request::execute)` instead

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -460,9 +545,18 @@ private void setDbWorkspaceFields(
...
+    try {
+      dbWorkspace =
+          workspaceService.saveAndCloneCohortsConceptSetsAndDataSets(fromWorkspace, dbWorkspace);
+    } catch (Exception e) {
+      // Tell Google to set the billing account back to the free tier if our clone fails
+      updateWorkspaceBillingAccount(dbWorkspace, workbenchConfigProvider.get().billing.accountId);
+      throw e;
+    }
 
+    // TODO: File a bug about this. Can create a workspace
```
jmthibault79: Could you go into more detail here?

ericsong: oops. Yeah, I expanded my comment. I'm not quite sure if it warrants a ticket so I think I'm going to leave it at that

ericsong:
@calbach @jmthibault79  I believe it should be ready for another round.

PR #3036, opened 2020-01-24
[RW-4322][risk=low] Implement describe-cluster debug tool

calbach:
Usage:

```
./project.rb list-clusters --project all-of-us-rw-prod
... # find cluster of interest

./project.rb describe-cluster --id aou-rw-xxxxxx/all-of-us-xx
```

Prints command line instructions for inspecting the logs directory for the cluster.

Sample output: https://pmi-engteam.slack.com/files/U5KGPEEPK/FSS4UKYF5/describe-cluster_output

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1329,7 +1329,7 @@ def describe_cluster(cmd_name, *args)
     end
   end
   common.warning "unable to determine project by cluster ID" unless project_from_cluster
-  if not op.opts.project
+  unless op.opts.project
     op.opts.project = project_from_cluster
   end
 
@@ -1377,7 +1377,6 @@ Common.register_command({
   :description => "List all clusters in this environment",
   :fn => ->(*args) { list_clusters("list-clusters", *args) }
 })
-
 
 def load_es_index(cmd_name, *args)
   op = WbOptionsParser.new(cmd_name, args)
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1317,16 +1317,22 @@ def describe_cluster(cmd_name, *args)
 
   # Infer the project from the cluster ID project ID. If for some reason, the
   # target cluster ID does not conform to the current billing prefix (e.g. if we
-  # changed the prefix), --project will override this.
+  # changed the prefix), --project can be used to override this.
   common = Common.new
+  matching_prefix = ""
   project_from_cluster = nil
   ENVIRONMENTS.each_key do |env|
-    if op.opts.cluster_id.start_with?(get_billing_project_prefix(env))
+    env_prefix = get_billing_project_prefix(env)
+    if op.opts.cluster_id.start_with?(env_prefix)
       # Take the most specific prefix match, since prod is a substring of the others.
-      if not project_from_cluster or project_from_cluster.length < env.length
+      if matching_prefix.length < env_prefix.length
         project_from_cluster = env
+        matching_prefix = env_prefix
       end
     end
+  end
+  if project_from_cluster == "local"
+    project_from_cluster = TEST_PROJECT
   end
   common.warning "unable to determine project by cluster ID" unless project_from_cluster
   unless op.opts.project
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1340,12 +1340,11 @@ def describe_cluster(cmd_name, *args)
   gcc.validate
 
   api_url = get_leo_api_url(gcc.project)
-  sa_ctx = ServiceAccountContext.new(gcc.project)
-  sa_ctx.run do
+  ServiceAccountContext.new(gcc.project).run do |ctx|
     common = Common.new
     common.run_inline %W{
        gradle manageClusters
-      -PappArgs=['describe','#{api_url}','#{gcc.project}','#{sa_ctx.service_account}','#{op.opts.cluster_id}']}
+      -PappArgs=['describe','#{api_url}','#{gcc.project}','#{ctx.service_account}','#{op.opts.cluster_id}']}
   end
 end
 
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1301,6 +1301,61 @@ Common.register_command({
   :fn => ->(*args) { delete_clusters("delete-clusters", *args) }
 })
 
+def describe_cluster(cmd_name, *args)
+  ensure_docker cmd_name, args
+  op = WbOptionsParser.new(cmd_name, args)
+  op.add_option(
+      "--id [CLUSTER_ID]",
+      ->(opts, v) { opts.cluster_id = v},
+      "Required cluster ID to describe, e.g. 'aou-test-f1-1/all-of-us'")
+  op.add_option(
+      "--project [project]",
+      ->(opts, v) { opts.project = v},
+      "Optional project ID; by default will infer the project form the cluster ID")
+  op.add_validator ->(opts) { raise ArgumentError unless opts.cluster_id }
+  op.parse.validate
+
+  # Infer the project from the cluster ID project ID. If for some reason, the
+  # target cluster ID does not conform to the current billing prefix (e.g. if we
+  # changed the prefix), --project will override this.
+  common = Common.new
+  project_from_cluster = nil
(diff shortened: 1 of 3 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -91,6 +101,42 @@ private static void listClusters(String apiUrl) throws IOException, ApiException
...
+      System.err.println(
+          String.format(
+              "given cluster ID '%s' is invalid, wanted format 'project/clusterName'", clusterId));
+      return;
+    }
+
+    // Leo's getCluster API swagger tends to be outdated; issue a raw getCluster request to ensure
+    // we get all available information for debugging.
+    ClusterApi client = newApiClient(apiUrl);
+    com.squareup.okhttp.Call call = client.getClusterCall(parts[0], parts[1], null, null);
```
jaycarlton: nit: I might name these arguments, even the null ones, for readability.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -91,6 +101,42 @@ private static void listClusters(String apiUrl) throws IOException, ApiException
...
+    ApiResponse<Object> resp = client.getApiClient().execute(call, Object.class);
+
+    // Parse the response as well so we can log specific structured fields.
+    Cluster cluster = PRETTY_GSON.fromJson(PRETTY_GSON.toJson(resp.getData()), Cluster.class);
+
+    System.out.println(PRETTY_GSON.toJson(resp.getData()));
```
jaycarlton: nit: why not allow the caller to specify an output stream?

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -91,6 +101,42 @@ private static void listClusters(String apiUrl) throws IOException, ApiException
...
+    System.out.printf("\n\nTo inspect logs in cloud storage, run the following:\n\n");
```
jaycarlton: should this specify that you should run this on your cluster?

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -140,13 +186,23 @@ private static void deleteClusters(
   public CommandLineRunner run() {
     return (args) -> {
```
jaycarlton: aside: Do we not yet have an argument parser for Java like we do in Ruby?

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
calbach: Done for parts[0], parts[1]. For the nulls, I prefer to comment rather than having placeholder variables.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
calbach: Why? Use of this function is pretty tightly scoped to just this CLI in this file, probably indefinitely - I don't see an advantage to generalizing this.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
calbach: No - you aren't supposed to do that (and permissions wise, this should probably be impossible to run on a workbench Leo cluster). This is intended for execution on a dev workstation.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
calbach: We do, but it hasn't been backfilled to these older CLIs yet. I've filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4340 for that. This situation could also use some more thought - having a ruby layer of flag parsing feeding into a duplicate Java layer of flag parsing is a sad state of affairs.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
jaycarlton: Oh, then I missed the point of why we're printing out stuff to be copied and executed later. The idea is that the logs are protected at a higher level than you necessarily have when you're running this, and/or they're too voluminous to dump to the terminal.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
jaycarlton: I'd be interested to chat. we're looking at the same issue with ops tooling.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
calbach: I've added a bit more instruction to this output. If you decide to run them, these commands get you to effectively a debug session. Human decisions are required because there are multiple log files you might want to look at, and we don't want to depend on the Leo-specific directory structure / naming.

After typing this out, it would good to automate more of this. Once I have https://precisionmedicineinitiative.atlassian.net/browse/PD-4740, I'll revisit this and glob out all the files in the directory, then give a simple commandline someone could use to copy/cat the logs if desired.

api/tools/src/main/java/org/pmiops/workbench/tools/ManageClusters.java
```diff
@@ -91,6 +101,50 @@ private static void listClusters(String apiUrl) throws IOException, ApiException
...
+
+    // TODO(PD-4740): Use impersonation here instead.
+    String keyPath = String.format("/tmp/%s-key.json", workbenchProjectId);
+    System.out.printf(
+        "    gcloud iam service-accounts keys create %s --iam-account %s\n",
+        keyPath, workbenchServiceAccount);
+    System.out.printf("    gcloud auth activate-service-account --key-file %s\n\n", keyPath);
+    System.out.printf("    gsutil ls gs://%s/**\n", cluster.getStagingBucket());
+    System.out.printf("    gsutil cat ... # inspect or copy logs\n\n");
+    System.out.printf("    # Delete the key when done\n");
```
jaycarlton: If I'm [reading this right](https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys), you can set an explicit expiration time by passing `validBeforeTime`, but I think it's only for the REST API and not the gcloud command.

api/libproject/devstart.rb
```diff
@@ -1301,6 +1301,60 @@ def delete_clusters(cmd_name, *args)
...
+  # Infer the project from the cluster ID project ID. If for some reason, the
+  # target cluster ID does not conform to the current billing prefix (e.g. if we
+  # changed the prefix), --project will override this.
+  common = Common.new
+  project_from_cluster = nil
+  ENVIRONMENTS.each_key do |env|
+    if op.opts.cluster_id.start_with?(get_billing_project_prefix(env))
+      # Take the most specific prefix match, since prod is a substring of the others.
+      if not project_from_cluster or project_from_cluster.length < env.length
+        project_from_cluster = env
```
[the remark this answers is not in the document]
calbach: Good catch, worth handling here. Added coverage, though there's a tech debt ticket to make sure we have better uniform handling of local vs test commands.

api/libproject/devstart.rb
```diff
@@ -1301,6 +1301,60 @@ def delete_clusters(cmd_name, *args)
...
+  op = WbOptionsParser.new(cmd_name, args)
+  op.add_option(
+      "--id [CLUSTER_ID]",
+      ->(opts, v) { opts.cluster_id = v},
+      "Required cluster ID to describe, e.g. 'aou-test-f1-1/all-of-us'")
+  op.add_option(
+      "--project [project]",
+      ->(opts, v) { opts.project = v},
+      "Optional project ID; by default will infer the project form the cluster ID")
+  op.add_validator ->(opts) { raise ArgumentError unless opts.cluster_id }
```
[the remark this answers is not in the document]
calbach: Agreed. But not in scope for this PR as there are ~10 unrelated usages of this same pattern.

PR #3040, opened 2020-01-24
[RW-4277][risk=low] Users can set workspace billing account back to free tier

ericsong:
Description:
Follow up to RW-3175.

The additional complexity around setting workspaces to the free tier billing account is that the user does not have sufficient permissions to make the call. Instead, we use the service account's credentials. Most of the logic is around choosing the correct Cloudbilling API depending on the incoming request.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -57,7 +60,7 @@ public UserController(
       Provider<WorkbenchConfig> configProvider,
       FireCloudService fireCloudService,
       UserService userService,
-      Provider<Cloudbilling> cloudBillingProvider) {
+      @Qualifier(USER_PROXY_CLOUD_BILLING) Provider<Cloudbilling> cloudBillingProvider) {
```
calbach: Why the `PROXY` language here? Elsewhere we have `END_USER_FOO` as a token which seems accurate to me. Is there a semantic difference?

api/src/main/java/org/pmiops/workbench/billing/GoogleApisConfig.java
```diff
@@ -18,22 +19,42 @@
...
-        new GoogleCredential()
-            .setAccessToken(userAuthentication.getCredentials())
-            .createScoped(
-                Collections.singletonList("https://www.googleapis.com/auth/cloud-platform"));
+    return createCloudbillingClient(
+        userAuthentication.getCredentials(), jsonFactory, workbenchConfigProvider.get());
+  }
+
+  @Bean(SERVICE_ACCOUNT_CLOUD_BILLING)
+  @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
```
calbach: nit: existing, but why DEFAULT here? Why not just leave blank?

api/src/main/java/org/pmiops/workbench/billing/GoogleApisConfig.java
```diff
@@ -18,22 +19,42 @@
...
+  public Cloudbilling serviceAccountGoogleCloudbillingApi(
+      JsonFactory jsonFactory, Provider<WorkbenchConfig> workbenchConfigProvider) {
+    String accessToken;
+    try {
+      accessToken =
+          ServiceAccounts.getScopedServiceAccessToken(
+              Collections.singletonList("https://www.googleapis.com/auth/cloud-platform"));
+    } catch (IOException e) {
+      throw new RuntimeException("Could not create service account access token for cloud billing");
```
calbach: nit: Per other  PR comments, would just let the exception propagate

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -645,6 +668,61 @@ public boolean maybeDeleteRecentWorkspace(long workspaceId) {
...
+    UpdateBillingInfo request;
+    try {
+      request =
+          cloudbilling
+              .projects()
+              .updateBillingInfo(
+                  "projects/" + workspace.getWorkspaceNamespace(),
+                  new ProjectBillingInfo().setBillingAccountName(newBillingAccountName));
+    } catch (IOException e) {
+      throw new RuntimeException("Could not create Google Cloud updateBillingInfo request", e);
```
calbach: nit: I'd prefer ServerException to a RuntimeException, here and below

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -645,6 +668,61 @@ public boolean maybeDeleteRecentWorkspace(long workspaceId) {
...
+    }
+
+    ProjectBillingInfo response;
+    try {
+      // this is necessary because the grant ownership call in create/clone
```
calbach: Ah... unfortunate.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -233,6 +229,11 @@ private FirecloudWorkspace attemptFirecloudWorkspaceCreation(FirecloudWorkspaceI
 
     Timestamp now = new Timestamp(clock.instant().toEpochMilli());
     DbWorkspace dbWorkspace = new DbWorkspace();
+    // A little unintuitive but setting this here reflects the current state of the workspace
```
calbach: Makes sense to me

calbach: It would probably be ideal here if the value were carried through on the billing project buffer table, rather than pulled from the config, since the billing account could change while projects are still in the buffer.

I think that's probably fine to just note this distinction and account for it when we're doing the free tier backfill. I'll make a note there.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -252,7 +253,13 @@ private FirecloudWorkspace attemptFirecloudWorkspaceCreation(FirecloudWorkspaceI
 
     dbWorkspace.setBillingMigrationStatusEnum(BillingMigrationStatus.NEW);
 
-    updateWorkspaceBillingAccount(dbWorkspace, workspace.getBillingAccountName());
+    try {
+      workspaceService.updateWorkspaceBillingAccount(
```
calbach: Yeah we should definitely have some cron cleanup for orphans here. Do you mind filing a ticket for later?

For example, if this request times out (but actually succeeds on the backend), we wind up with a defunct project attached to the user's billing account.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -425,15 +379,22 @@ private void setDbWorkspaceFields(
       dbWorkspace.setReviewRequested(researchPurpose.getReviewRequested());
     }
 
-    updateWorkspaceBillingAccount(dbWorkspace, request.getWorkspace().getBillingAccountName());
+    try {
+      workspaceService.updateWorkspaceBillingAccount(
```
calbach: The PATCH API is only supposed to include fields that have actually been modified. See all the other conditional checks above. This should mirror that behvaior

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -545,14 +510,20 @@ private void setDbWorkspaceFields(
 
     dbWorkspace.setBillingMigrationStatusEnum(BillingMigrationStatus.NEW);
 
-    updateWorkspaceBillingAccount(dbWorkspace, body.getWorkspace().getBillingAccountName());
+    try {
+      workspaceService.updateWorkspaceBillingAccount(
```
calbach: Eventually we'll also want some validation here that the creator still has free tier credit available. Do you know whether this is tracked?

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -136,7 +139,7 @@ private BillingAccount freeTierBillingAccount() {
     return new BillingAccount()
         .isFreeTier(true)
         .displayName("Use All of Us free credits")
-        .name("billingAccounts/" + configProvider.get().billing.accountId)
+        .name(configProvider.get().billing.freeTierBillingAccountName())
```
jmthibault79: nice

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: I thought we were going to allow users to continue creating workspaces in the free tier so I left it as is. I can ask Karthik if we want some UI treatment around how we handle the expired free tier case.

ericsong:
@calbach @jmthibault79 should be good for round 2

api/config/config_test.json
```diff
@@ -90,7 +90,7 @@
     "enableVpcServicePerimeter": true,
     "enableNewAccountCreation": true,
     "enableRdrExport": true,
-    "enableBillingLockout": false,
```
jmthibault79: is this intentional?

ericsong: yeah, I figure we can let it go on test now. It seems to work from my testing.

PR #3041, opened 2020-01-24
[RW-4247][risk=no] Reworking Concept Set Builder to include PM and Surveys

freemabd:
- Added the ability to create/edit/delete PM and Surveys inside the Concept Set Builder.
- Updated/added test coverage for each new bucket type
- Updated routing after deleting concept sets
- In test/staging/stable/prod PM will be behind feature flag because it is causing problems in Data Set Builder

image

image

### DIFF
--- a/api/db-cdr/generate-cdr/make-bq-data.sh
+++ b/api/db-cdr/generate-cdr/make-bq-data.sh
@@ -117,7 +117,7 @@ VALUES
 (1586134,'The Basics','Survey includes participant demographic information.',0,0,1),
 (43529712,'Personal Medical History','This survey includes information about past medical history, including medical conditions and approximate age of diagnosis.',0,0,4),
 (43528895,'Healthcare Access & Utilization','Survey includes information about a participants access to and use of health care.',0,0,5),
-(43528698,'Family Medical History','Survey includes information about the medical history of a participants immediate biological family members.',0,0,6)"
+(43528698,'Family History','Survey includes information about the medical history of a participants immediate biological family members.',0,0,6)"
 
 # Populate some tables from cdr data
 ###############
--- a/api/src/main/resources/client_api.yaml
+++ b/api/src/main/resources/client_api.yaml
@@ -126,6 +126,31 @@ paths:
           schema:
             $ref: "#/definitions/ConceptListResponse"
 
+  /v1/workspaces/{workspaceNamespace}/{workspaceId}/domainCounts:
+    post:
+      tags:
+        - concepts
+      consumes:
+        - application/json
+      description: >
+        Searches for concepts in concept table by name, and optionally filter on
+        domain, vocabulary IDs, or standard concept status. Uses the CDR version affiliated
+        with the workspace specified in the path.
+      operationId: "domainCounts"
+      parameters:
+        - $ref: '#/parameters/workspaceNamespace'
+        - $ref: '#/parameters/workspaceId'
+        - in: body
+          name: request
+          description: domain counts search request
+          schema:
+            $ref: "#/definitions/DomainCountsRequest"
+      responses:
+        200:
+          description: A collection of concepts
+          schema:
+            $ref: "#/definitions/DomainCountsListResponse"
+
 definitions:
 
   DataTableSpecification:
@@ -501,7 +526,6 @@ definitions:
     required:
       - standardConceptFilter
       - domain
-      - includeDomainCounts
     properties:
       query:
         type: string
@@ -528,14 +552,32 @@ definitions:
         type: integer
         format: int32
         description: The maximum number of results returned. Defaults to 20.
-      includeDomainCounts:
-        type: boolean
-        default: false
-        description: True if per-domain counts of concepts matching the criteria should be included in the response
       pageNumber:
         type: integer
         default: 0
         description: By default it returns the first page and then its next pages from that on.
+
+  DomainCountsRequest:
+    type: object
+    required:
+      - standardConceptFilter
+    properties:
+      query:
+        type: string
+        description: >
+          A query string that can be used to match a subset of the name (case-insensitively),
+          the entire code value (case-insensitively), or the concept ID. If not specified, all
+          concepts are returned.
+      standardConceptFilter:
+        description: >
+          STANDARD_CONCEPTS if only standard concepts should be returned,
+          NON_STANDARD_CONCEPTS if only non-standard
+          concepts should be returned; defaults to ALL_CONCEPTS, meaning both
+          standard and non-standard concepts will be returned.
+        $ref: "#/definitions/StandardConceptFilter"
+      surveyName:
+        description: The name of the survey clicked
+        type: string
 
   ConceptListResponse:
     type: object
@@ -550,6 +592,12 @@ definitions:
         type: "array"
         items:
           $ref: "#/definitions/SurveyQuestions"
+
+  DomainCountsListResponse:
+    type: object
+    required:
+      - domainCounts
+    properties:
       domainCounts:
         type: "array"
         items:
--- a/common-api/src/main/java/org/pmiops/workbench/cdr/model/DbConcept.java
+++ b/common-api/src/main/java/org/pmiops/workbench/cdr/model/DbConcept.java
@@ -49,6 +49,33 @@ public class DbConcept {
         .synonymsStr(a.getSynonymsStr());
   }
 
+  // Used from JQL queries in DomainInfoDao
+  public DbConcept(
+      long conceptId,
+      String conceptName,
+      String standardConcept,
+      String conceptCode,
+      String conceptClassId,
+      String vocabularyId,
+      String domainId,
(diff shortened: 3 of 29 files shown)


api/db-cdr/generate-cdr/make-bq-data.sh
```diff
@@ -117,7 +117,7 @@ VALUES
 (1586134,'The Basics','Survey includes participant demographic information.',0,0,1),
 (43529712,'Personal Medical History','This survey includes information about past medical history, including medical conditions and approximate age of diagnosis.',0,0,4),
 (43528895,'Healthcare Access & Utilization','Survey includes information about a participants access to and use of health care.',0,0,5),
-(43528698,'Family Medical History','Survey includes information about the medical history of a participants immediate biological family members.',0,0,6)"
+(43528698,'Family History','Survey includes information about the medical history of a participants immediate biological family members.',0,0,6)"
```
freemabd: @dolbeew need to confirm that all references to this survey use 'Family History'

api/src/main/resources/client_api.yaml
```diff
@@ -126,6 +126,31 @@ paths:
...
+        with the workspace specified in the path.
+      operationId: "domainCounts"
+      parameters:
+        - $ref: '#/parameters/workspaceNamespace'
+        - $ref: '#/parameters/workspaceId'
+        - in: body
+          name: request
+          description: domain counts search request
+          schema:
+            $ref: "#/definitions/DomainCountsRequest"
```
freemabd: @dolbeew Should this api call be lumped in the the domain data api call, since we will only we showing 1 domain per bucket click?

common-api/src/main/java/org/pmiops/workbench/cdr/model/DbConcept.java
```diff
@@ -49,6 +47,33 @@ public DbConcept(DbConcept a) {
         .synonymsStr(a.getSynonymsStr());
   }
 
+  // Used from JQL queries in DbConcept
```
freemabd: Eventually i think we will be replacing DbConcept with DbCriteria

ui/src/app/pages/data/concept/concept-add-modal.tsx
```diff
@@ -26,13 +26,10 @@ const styles = reactStyles({
...
 const filterConcepts = (concepts: any[], domain: Domain) => {
-  if (domain === Domain.PHYSICALMEASUREMENT) {
-    return concepts.filter(concept => concept.domainId === fp.capitalize(Domain[Domain.MEASUREMENT])
-      && concept.vocabularyId === CriteriaType[CriteriaType.PPI]);
-  } else if (domain === Domain.SURVEY) {
+  if (domain === Domain.SURVEY) {
     return concepts.filter(concept => !!concept.question);
   } else {
-    return concepts.filter(concept => concept.domainId === fp.capitalize(Domain[domain]));
+    return concepts.filter(concept => concept.domainId.replace(' ', '').toLowerCase() === Domain[domain].toLowerCase());
```
freemabd: At some point we should probably look at using Domain or DomainType but not both.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -262,25 +241,24 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
         let conceptDomainCounts: DomainCount[] = conceptDomainInfo.items.map((domain) => ({
           domain: domain.domain,
           name: domain.name,
           conceptCount: 0
         }));
-        if (environment.enableNewConceptTabs) {
-          conceptsCache = [...conceptsCache, ...conceptCacheStub];
-          conceptDomainCounts = [...conceptDomainCounts];
-        } else {
+        if (!environment.enableNewConceptTabs) {
```
freemabd: @dolbeew we should probably only shot this in local for the time being.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -388,24 +366,28 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
     browseSurvey(surveyName) {
-      if (environment.enableNewConceptTabs) {
-        this.browseDomain();
-      } else {
-        this.setState({browsingSurvey: true, selectedSurvey: surveyName});
-      }
+      this.setState({
+        currentInputString: '',
+        currentSearchString: '',
+        selectedDomain: {domain: Domain.SURVEY, name: 'Surveys', conceptCount: 0},
```
freemabd: Should conceptCount be dynamic instead of 0?

ui/src/app/pages/data/concept/survey-details.tsx
```diff
@@ -82,11 +82,11 @@ export const SurveyDetails = withCurrentWorkspace()(
 
     async loadSurveyDetails() {
       try {
-        const {workspace, surveyName} = this.props;
-        const surveys = await conceptsApi().getSurveyQuestions(
-          workspace.namespace, workspace.id, surveyName);
+        const {surveyName, workspace} = this.props;
+        const request = {standardConceptFilter: StandardConceptFilter.ALLCONCEPTS, surveyName};
+        const surveys = await conceptsApi().searchSurveys(workspace.namespace, workspace.id, request);
```
freemabd: This new api call brings questions back in same order as CB.

api/src/main/resources/client_api.yaml
dolbeew: SGTM. And when searching from the homepage (all tabs), we would call that combined endpoint once for each tab?

api/db-cdr/generate-cdr/make-bq-data.sh
dolbeew: AKAIK we only use the value returned by the api. I don't believe the names are hardcoded anywhere, will double-check though.

common-api/src/main/java/org/pmiops/workbench/cdr/model/DbConcept.java
dolbeew: Good to know

ui/src/app/pages/data/concept/concept-add-modal.tsx
dolbeew: Agreed. Looks like they both have at least one value that the other doesn't. Would be good to see where those are being used, if at all, and consolidate them.

ui/src/app/pages/data/concept/concept-homepage.tsx
dolbeew: Yes, currently this flag is only enabled in the local environment

ui/src/app/pages/data/concept/concept-homepage.tsx
dolbeew: It's basically a placeholder at this point. It will get replaced with the real count when `searchConcepts()` runs below.

ui/src/app/pages/data/concept/survey-details.tsx
dolbeew: Ok, cool. That helps when we're ready to add numbers to the rows to match data browser.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -336,33 +314,36 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
-          query: currentSearchString,
-          standardConceptFilter: standardConceptFilter,
-          domain: cacheItem.domain,
-          includeDomainCounts: activeTabSearch,
-          maxResults: this.MAX_CONCEPT_FETCH
-        });
+        if (cacheItem.domain === Domain.SURVEY) {
+          await conceptsApi().searchSurveys(namespace, id, request).then(resp => cacheItem.items = resp);
+        } else {
+          await conceptsApi().searchConcepts(namespace, id, {...request, domain: cacheItem.domain})
```
dolbeew: I noticed that the `SearchConceptsRequest` accepted by this call has an optional `surveyName` param. Will that change anything in the response?

api/db-cdr/generate-cdr/make-bq-data.sh
freemabd: Confirmed, we should be good to go.

api/src/main/resources/client_api.yaml
freemabd: Maybe, we can discuss this more after public launch.

ui/src/app/pages/data/concept/concept-homepage.tsx
freemabd: Response will remain the same. SurveyName is only relevant when clicking a specific survey bucket.

ui/src/app/pages/data/concept/concept-homepage.tsx
freemabd: Oh.. gotcha. I think this is so, because surveys doesn't exist in the the domain_info table.

PR #3043, opened 2020-01-27
[RW-4091][RISK=NO]Change the return object for cloud task to array of long

NehaBroad:
The request body for cloud task was set to Object just so RDR is unblocked and we can start export data. This PR is replacing generic Object to Array of Long

---
**PR checklist**

- [ X] The JIRA story has been moved to Dev Review
- [ X] I have run and tested this change locally
- [ X] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -2270,14 +2270,15 @@ paths:
       description: >
         Task handler which handle request from cloud task queue rdrQueueTest to send Researcher data
       operationId: "exportResearcherData"
+      consumes:
+        - application/json
       parameters:
         - in: body
           name: researcherIds
           required: true
-          description: The IDs of concepts to be added to the concept set.
+          description: The IDs of Users to be exported to Rdr.
           schema:
-            type: object
-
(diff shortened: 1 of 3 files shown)


api/src/main/resources/workbench.yaml
```diff
@@ -4094,9 +4095,8 @@ definitions:
         description: The name of the Jupyter VM which generated egress, e.g. all-of-us-260-m.
         type: string
 
-  RdrExportId:
-    type: object
-    properties:
-      exportId:
+  ArrayOfLong:
```
jmthibault79: Please add a note here that this is only necessary because of a Swagger bug and we can remove it after we update Swagger.
https://github.com/swagger-api/swagger-codegen/issues/6745

NehaBroad:
OfflineRdrExportController will work as is

NehaBroad:
> Will you need to update `OfflineRdrExportController` or will it work as-is?
OfflineRdrExportController will work as is, its still sending List<Long> as JSON

PR #3051, opened 2020-01-28
[risk=no][no ticket] Add STRUCTURE.md

als364:
This file describes, broadly, the code structure of the Workbench app.


als364:
Feel free to unassign yourself for review if you're busy. I want more than one pair of eyes on this.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
+# All of Us Workbench API Structure & Technologies
```
jaycarlton: Please add an inbound link from the top level README (one directory above api). Actually, this could be merged with (or replace) some of the preamble material there.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+TODO: do we have a preference of which of those to do?
+
+## DB Models
+
+DB Models represent a single row of a single table in the database, possibly with extender tables attached. We denote DB models by prefixing `Db` to the name of the database table. DB models heavily use `javax.persistence` annotations to describe where in the database table each part of the model lives.
+
+In order to automatically gather data from extender tables, we use the `@JoinColumn`, `@OneToOne`, `@OneToMany`, `@ManyToOne`, and `@MappedBy` [annotations](https://www.baeldung.com/jpa-join-column). When using these annotations, be sure to carefully consider which table should 'own' the relationship. With a `@ManyToOne` relationship, the `@ManyToOne` side should generally own the relationship. For example. if we have an extender table containing a user's addresses, `DbAddress` should have a `@ManyToOne` `DbUser` member and `DbUser` should have a `@OneToMany` `DbAddress` member, with the annotation containing a `mappedBy` attribute pointing to `DbAddress`'s `DbUser` member as the member defining the relationship.
+
+TODO: wording on the above is Bad. pls help
```
jaycarlton: You're getting into the weeds a bit here. Maybe breaking out a separate doc on spring data will help.

Links to usage in the code (though fragile) could help too. Might be better to put snippets of canonical usage here.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
+# All of Us Workbench API Structure & Technologies
+
+The _All of Us_ Workbench API is largely built on [Spring Boot](https://spring.io/projects/spring-boot). It also uses [Swagger](https://swagger.io/) for autogenerating REST APIs.
+
+## Database
+
+The database is a [Google Cloud SQL](https://cloud.google.com/sql/) instance running MySQL. We modify its schema using [Liquibase](https://www.liquibase.org/) migrations, which live in [db/changelog/](https://github.com/all-of-us/workbench/tree/master/api/db/changelog).
```
jaycarlton: I'd probably say "The primary application database is..." so as not to hurt BigQuery's feelings.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+TODO: wording on the above is Bad. pls help
+
+Much of the structure of DB models can be autogenerated in IntelliJ by defining the class members and then doing Generate -> Getter and Setter and Generate -> `equals()` and `hashCode()`. This will not automatically add the required annotations - those must be added manually.
+
+## Services and ServiceImpls
+
+Services are interfaces that declare methods of business logic. ServiceImpls are classes implementing their correspondingly named Service and the business logic methods declared therein.
+
+TODO: Why do we do this this way?
```
jaycarlton: It's 75% philosophy, 25% requirement. Basically, a Java class can implement any number of Spring beans for one or more services, and a given Spring bean can have any number of implementations.

Declaring a `@Service` directly on the implementation class is stepping on both of those degrees of freedom. A dependent service shouldn't need to know (or have any way to detect) the implementation class chosen.

I actually got bit by this in the case of the UserService/UserServiceImpl pair. Before the latter was created, there was no (direct) way for implementation context to be reused to support "sidecar services" like the `GaugeDataCollector` bean.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+ServiceImpls can call into other services, but we try to keep the number of dependencies down. We try to avoid having ServiceImpls call into a large number of DAOs, especially DAOs not directly related to the functional area for which the ServiceImpl is responsible. For instance, the `UserServiceImpl` calls into the `UserDao`, the `AdminActionHistoryDao`
+
+ServiceImpls should handle only application logic. They should not need explicit knowledge of the database, the REST API, or the UI.
+
+## Controllers
+
+Controllers take Request objects from the API, call into the Service layer, and return Response objects back to the API. Controllers implement autogenerated [APIControllers](#apicontrollers).
+
+Controller methods should generally be small and should not contain any actual logic. The most common conditional in the Controller layer should be checking a feature flag. Most of the work done by Controllers should be munging data provided by the API or a Service.
```
jaycarlton: > and should not contain any actual logic
💯 

> munging data provided by the API or a Service
I'd say "packaging results into a ResponseEntity" and possibly something about translating errors (occasionally).

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Mappers
+
+There are four types of models in the system - [API models](#swagger), [DB models](#db-models), and [API Request and Response](#requests-and-responses) models. Mappers are utility classes that own methods that convert between these types of models.
```
jaycarlton: I'd mention/link to MapStruct specifically.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+## API
+
+The actual REST API is defined in a series of YAML files according to the [Swagger OpenAPI spec](https://swagger.io/specification/). Swagger takes these YAML files, merges them, and generates all the classes defined by the YAMLs. The command to do so is `./project.rb compile-generated-java`.
+
+One non-Swagger-standard thing about these YAML files is that the `tag` param is used to indicate which Controller / APIController an endpoint should be associated with.
```
jaycarlton: Might want to mention swagger-codegen, which is technically a separate project.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+Each REST endpoint is defined under `PATHS` in the YAML file.
+
+### API Models
+
+API models are defined under `DEFINITIONS` in the YAML file. Swagger autogenerates classes from these. Both API models and DB models are currently used in the Controller and Service layer. There is not currently consensus on whether the standard should switch to only using one or the other.
+
+TODO: is the latter sentence true?
```
jaycarlton: The major advantage of separating these is allowing a more layered architecture and stricter separation of business logic from database details. Granted, Spring is already giving us some abstraction.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Requests and Responses
+
+Swagger interprets the data sent to a given endpoint according to the Request model definition associated with that endpoint and throws an error if the data is improperly formatted. Likewise, Swagger validates that the data being returned from a given endpoint conforms to the Response model definition associated with that endpoint and throws an error if it does not.
+
+Technically, Requests and Responses are just more API model classes, similarly autogenerated by Swagger, but they are only used when communicating directly with the API layer. API models that represent business logic concepts are more freely used throughout the system.
```
jaycarlton: The missing bit here is a business-logic-level class. I think a major answer here will be Immutables.org objects, which let you define ad hoc value objects with builders that aren't subject to mutation.

api/STRUCTURE.md
```diff
@@ -0,0 +1,83 @@
...
+
+### APIControllers
```
jaycarlton: I don't know if i'd even go into this here. It's a bit of an implementation detail that you don't need to worry about. Worth discussing in a spring-speicfic document.

api/STRUCTURE.md
als364: I don't actually have a good sense of what parts of our API generation are swagger and which are swagger-codegen.

api/STRUCTURE.md
als364: I understand this. I'm just asking whether we as a team actually agree on what types of models we should be using where. We certainly haven't prioritized conforming to any particular standard.

api/STRUCTURE.md
als364: This document is about the current state of the system.

api/STRUCTURE.md
als364: This comment is here so that people understand what these random classes are that their Controllers talk to and inherit from. I should expand this section.

api/STRUCTURE.md
jaycarlton: That's fair. I might rename this new one to something like `BACKEND_OVERVIEW.md` or something. (There are structures everywhere.)

That one is a train wreck. I have a couple of tickets stashed away on reorganizing it and breaking it into useful sub-docs.

api/STRUCTURE.md
jaycarlton: Swagger (now OpenAPI) is the standard governing how we describe REST APIs in YAML or JSON. Swagger-codegen is a less-standard set of tools around building code from a specification in Swagger.

The bit where we use a tag to name the controller is a hack somewhere in Swagger-codegen. (There are many output languages and frameworks from what I understand; ours is specifically Java/Spring)

api/STRUCTURE.md
jaycarlton: There's a lot of value in what you're doing, i.e. to describe in some systematic way the general understanding of how things work today. It makes it possible to lay out a workable plan for improving things.

api/STRUCTURE.md
jaycarlton: 🤷‍♂ I'm not sure I'd get into it then. It's kind of a detour.

api/STRUCTURE.md
jaycarlton: I'd talk about BigQuery before getting into these details. It's just middleware really.

api/STRUCTURE.md
```diff
@@ -0,0 +1,84 @@
+# All of Us Workbench API Structure & Technologies
```
gjuggler: A couple notes / thoughts on file naming. Feel free to ignore (though I think it does help a lot when we get these things right, since it makes it obvious where the next person who comes along should put content):

1. We have an existing directory, api/doc/, which contains some similar "here's how parts of the system work" Markdown files. It might be cleaner for this file to live over there, so there's one place where all of our backend-related team docs live.
2. In terms of file naming, I think all-caps should probably be reserved for truly standard filenames, e.g. README or INDEX. Structure as a name isn't bad, but I'm not sure it perfectly captures the intent. Spitballing a few other ideas in case any of them jump out to you as an improvement:
  - api/doc/overview.md
  - api/doc/index.md (this could eventually become a summary page of sorts, which links out to the more topic-specific docs we already have)
  - api/doc/tech-stack.md
  - api/doc/getting-started.md

api/STRUCTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Database
+
+The primary application database is a [Google Cloud SQL](https://cloud.google.com/sql/) instance running MySQL. We modify its schema using [Liquibase](https://www.liquibase.org/) migrations, which live in [db/changelog/](https://github.com/all-of-us/workbench/tree/master/api/db/changelog).
+
+When creating a new database migration, add an XML file to the changelog directory that describes the migration to perform. This new file should be named `db.changelog-###-description-of-migration.xml`, where ### is the number of the newest migration + 1. Then, add the file to the end of the [db.changelog-master.xml](https://github.com/all-of-us/workbench/blob/master/api/db/changelog/db.changelog-master.xml) file.
+
+We generally name database objects in lower_snake_case.
+
+Liquibase migrations are applied (indirectly) via a bash script called [db/run-migrations.sh](https://github.com/all-of-us/workbench/blob/master/api/db/run-migrations.sh). Do not run this script directly; instead, call one of the following `project.rb` commands:
```
gjuggler: Not necessarily worth noting in here, but TIL that for our GAE environments, the Liquibase migrations are actually executed using the Liquibase Gradle plugin: https://github.com/all-of-us/workbench/blob/331b1228bb7c7b8d392f2acf393c64d82e24a2ef/api/db/build.gradle#L16-L16

api/STRUCTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+CrudRepository also allows queries to be [automatically derived](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details) from the name of a method defined on the interface. As such, you will see methods like `findByUserIdAndWorkspaceId` with no apparent implementation. The implementation is generated by Spring.
+
+To define a query that uses a foreign key, there are a couple of options. First, you could use `@JoinColumn`/`@[One|Many]ToOne` annotation on the corresponding DB Model, as described [below](#db-models). Second, you could use an [@Query](https://www.baeldung.com/spring-data-jpa-query) annotation to just describe the query in stringy SQL.
+
+TODO: do we have a preference of which of those to do?
```
gjuggler: @ericsong  and I had a pretty good discussion about the various options here this week. I'd love if he could follow-up with some recommendations and/or pointers to models & DAOs where we use healthy patterns.

api/STRUCTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+## DB Models
+
+DB Models represent a single row of a single table in the database, possibly with extender tables attached. We denote DB models by prefixing `Db` to the name of the database table. DB models heavily use `javax.persistence` annotations to describe where in the database table each part of the model lives.
+
+In order to automatically gather data from extender tables, we use the `@JoinColumn`, `@OneToOne`, `@OneToMany`, `@ManyToOne`, and `@MappedBy` annotations. Documentation and blogs on the usage of these annotations can be found in these links:
+
+- https://www.baeldung.com/jpa-join-column
```
jmthibault79: These links are very helpful.  Thank you!

api/STRUCTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Services and ServiceImpls
+
+Services are interfaces that declare methods of business logic. ServiceImpls are classes implementing their correspondingly named Service and the business logic methods declared therein.
+
+Practically, this allows one ServiceImpl to implement multiple interfaces. This is used in several `UserServiceImpl` and others to mixin [GaugeDataCollector](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/monitoring/GaugeDataCollector.java), which needs a different implementation per-service.
+
+Services / ServiceImpls are generally named after a feature or functional area, like 'UserService', 'CohortService', or 'MonitoringService'.
+
+ServiceImpls can call into other services, but we try to keep the number of dependencies down. We try to avoid having ServiceImpls call into a large number of DAOs, especially DAOs not directly related to the functional area for which the ServiceImpl is responsible. For instance, the `UserServiceImpl` calls into the `UserDao`, the `AdminActionHistoryDao`
```
jmthibault79: The last sentence looks incomplete.  Also, is this intended as a positive example because it only talks to two DAOs or a negative one because we'd prefer it to talk to only one?

api/STRUCTURE.md
als364: That directory was not on my radar at all. I agree that it makes sense to put this document there, but I'm a bit concerned that I'd just be round filing it. Maybe I'll point out this directory and this file in both the top level and API level READMEs.

jaycarlton: @gjuggler there's a ticket somewhere (or at least in my head) where we break down the main README into other sections and files. We need a navigation convention where everything there is linked from the top and/or intermediate sections, and link back in turn as a courtesy.

One annoyance that I think actually stops people from updating that file is that it's not in the `api` project. So maybe a top-of-the-backend readme makes sense.

api/STRUCTURE.md
als364: Positive example because everything it talks to is directly related to the purpose of the service. Something massive like `WorkspaceServiceImpl` should probably 1) get broken up and 2) stop talking to the `UserDao` directly.

als364: Ideally, each DAO would roll up to only one Service.

jaycarlton: No, you could certainly imagine a service that needs more than one table (dao), or a table that's useful to more than one service. For example, the CohortDao is injected into a couple of services.

But yes, it makes lots of sense to break up some of these mega-services, even if they share DAOs.

PR #3054, opened 2020-01-29
[RW-4159][risk=no] Terms of service data model and API

gjuggler:
This PR introduces a new `user_terms_of_service` table modeled loosely on the `user_data_use_agreement` table. Although the intent here is likely to implement TOS as a full access module, the initial requirement for Beta launch is to record TOS acknowledgement events for new users, but not require it for existing users. (See comments in [RW-4159](https://precisionmedicineinitiative.atlassian.net/browse/RW-4159) for more detail.)

The goal here is to be as backwards-compatible as possible. There should be no impact of an API server rollback on the database (the new table simply won't be used by a prior version) or on the API. Feature-flagging isn't necessary here, since the new API parameter is optional.

Things I learned:
 - It's often considered best practice to use bidirectional relationship mappings in JPA to define parent-child relationships like this. When the number of children is known to be small, this works well.
 - The application is responsible for maintaining the bidirectional link when modifying data entity classes directly. See the comment in the DbUserTermsOfService class – I was pulling my hair out trying to get the controller tests working, when I realized that dbUser.getTermsOfServiceRows was still empty because I hadn't been correctly updating the in-memory model classes.
 - Our codebase is very inconsistent with how DB relations are modeled and used. For example, look at Authorities which is done *very* differently. There's definitely some room for cleanup over time here.

---
**PR checklist**

- [NO] This PR meets the Acceptance Criteria in the JIRA story
- [NO] I have run and tested this change locally
- [X] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers


api/src/main/java/org/pmiops/workbench/db/model/DbUserTermsOfService.java
```diff
@@ -0,0 +1,75 @@
...
+  public long getUserTermsOfServiceId() {
+    return userTermsOfServiceId;
+  }
+
+  public void setUserTermsOfServiceId(long userTermsOfServiceId) {
+    this.userTermsOfServiceId = userTermsOfServiceId;
+  }
+
+  @Column(name = "user_id", nullable = false)
+  public long getUserId() {
```
ericsong: Was it necessary to change the User model to a long? At some point, we should decide which access pattern we want to standardize on. I think there's good arguments for both approaches.

api/src/main/java/org/pmiops/workbench/db/model/DbUserTermsOfService.java
```diff
@@ -0,0 +1,75 @@
...
+  @Column(name = "user_terms_of_service_id")
+  public long getUserTermsOfServiceId() {
+    return userTermsOfServiceId;
+  }
+
+  public void setUserTermsOfServiceId(long userTermsOfServiceId) {
+    this.userTermsOfServiceId = userTermsOfServiceId;
+  }
+
+  @Column(name = "user_id", nullable = false)
```
ericsong: Also curious about the nullable annotation. Is it for documentation or does it affect our code somehow?

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -76,4 +76,21 @@ public void fireAdministrativeBypassTime(
...
+  public void fireAcknowledgeTermsOfService(DbUser targetUser, Integer termsOfServiceVersion) {
+    actionAuditService.send(
+        ActionAuditEvent.builder()
+            .timestamp(clock.millis())
+            .agentType(AgentType.USER)
+            .agentId(dbUserProvider.get().getUserId())
+            .agentEmailMaybe(dbUserProvider.get().getUsername())
+            .actionId(actionIdProvider.get())
+            .actionType(ActionType.EDIT)
+            .targetType(TargetType.ACCOUNT)
```
jaycarlton: If you don't mind, please tag me on PRs with new audit log entries, since we don't have a great migration story there.

The main thing I see here is that the action and target don't feel quite right: at the moment compliance docs and compliance acceptance aren't first-class concepts in the audit system. It would be slightly more idiomatic (and potentially help with audit queries), if you had something like 
```
AgentType.USER
ActionType.ACCEPT
TargetType.COMPLIANCE_DOCUMENT

// 2 property fields:
DocumentTargetProperty.VERSION,
DocumentTargetProperty.NAME,
```

Then the log entry for DUA acceptance would be almost identical to this code (and in fact we could bust out a separate `ComplianceDocumentAuditor` when we're ready to do that.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -76,4 +76,21 @@ public void fireAdministrativeBypassTime(
...
     actionAuditService.send(eventBuilder.build());
   }
+
+  @Override
+  public void fireAcknowledgeTermsOfService(DbUser targetUser, Integer termsOfServiceVersion) {
+    actionAuditService.send(
+        ActionAuditEvent.builder()
+            .timestamp(clock.millis())
+            .agentType(AgentType.USER)
+            .agentId(dbUserProvider.get().getUserId())
```
jaycarlton: Please add a comment about why you use the provider here and the passed-in user below if that's intentional.

api/src/main/java/org/pmiops/workbench/auth/ProfileService.java
```diff
@@ -207,6 +215,14 @@ public Profile getProfile(DbUser user) {
     profile.setFreeTierUsage(freeTierBillingService.getUserCachedFreeTierUsage(user));
     profile.setFreeTierDollarQuota(freeTierBillingService.getUserFreeTierDollarLimit(user));
 
+    Optional<DbUserTermsOfService> latestTermsOfServiceMaybe =
+        userTermsOfServiceDao.findFirstByUserIdOrderByTosVersionDesc(user.getUserId());
+    if (latestTermsOfServiceMaybe.isPresent()) {
```
jaycarlton: nit: one pattern that's kinda nice is to do:
```
latestTermsOfServiceMaybe.ifPresent(profile::setTermsOfServiceFields)
```
where `setTermsOfServiceFields` is a helper method that sets both these things.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -43,6 +43,9 @@ DbUser createUser(
   DbUser submitDataUseAgreement(
       DbUser user, Integer dataUseAgreementSignedVersion, String initials);
 
+  // Registers that a user has agreed to a given version of the Terms of Service.
+  void submitTermsOfService(DbUser dbUser, Integer tosVersion);
```
jaycarlton: nit: I'd name this `recordTermsOfServiceAcceptance()`

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -412,6 +417,21 @@ public void setDataUseAgreementNameOutOfDate(String newGivenName, String newFami
     userDataUseAgreementDao.save(dataUseAgreements);
   }
 
+  @Override
+  @Transactional
+  public void submitTermsOfService(DbUser dbUser, Integer tosVersion) {
+    if (tosVersion != CURRENT_TERMS_OF_SERVICE_VERSION) {
+      throw new BadRequestException("Terms of Service version is not up to date");
```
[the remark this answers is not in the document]
jaycarlton: +1 the lifecycles on all these compliance checks are really adding up, and we have to deal with combinatoric effects of some of them not being done, with overlapping grace periods, etc. It's worth thinking about the migration story if we want to keep things super simple up front.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -414,6 +419,21 @@ public void setDataUseAgreementNameOutOfDate(String newGivenName, String newFami
...
+    }
+
+    DbUserTermsOfService userTermsOfService = new DbUserTermsOfService();
```
jaycarlton: This would be a good spot to make a builder and an immutable class I think.

api/src/main/java/org/pmiops/workbench/db/model/DbUserTermsOfService.java
jaycarlton: Spring can use this to talk to the database in a more intelligent way. +1 for readable annotations instead of comments; -1 that we have to specify this stuff multiple places.

api/src/main/java/org/pmiops/workbench/db/model/DbUserTermsOfService.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import org.hibernate.annotations.CreationTimestamp;
+
+@Entity
+@Table(name = "user_terms_of_service")
+public class DbUserTermsOfService {
```
jaycarlton: aside: how hard would it be to merge the data model and services, etc, for the TOS and DUA. I think only a little business logic separates them, right?

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
jaycarlton: Filed [this story](https://precisionmedicineinitiative.atlassian.net/browse/RW-4372) to add constraints to these object types to make the intent more clear (and assist in query-building tools and services for this dataset).

api/src/main/java/org/pmiops/workbench/db/model/DbUserTermsOfService.java
gjuggler: My understanding is that these impact how the H2 database is generated for unit tests. It's slightly more realistic for those tables to match the nullability constraints in our actual Liquibase-defined SQL schema.

(Although, that being said, see a long comment I added to the "agreement_time" column. Due to some weirdness with how the CreationTimestamp annotation is used by Hibernate, I wasn't able to mark this column as non-nullable in the @Column annotation, even though it is non-nullable in SQL.)

PR #3058, opened 2020-01-29
[RW-4348][risk=low] Disable some notebook related actions for inactive billing workspaces

ericsong:
These are the actions we are preventing users from carrying out if their workspace is billing inactive.

- Creating notebooks
- Duplicating notebook
- Exporting datasets to notebooks (from a resource card)
- Exporting dataset to notebooks (from dataset builder page)
- Starting clusters
   - Playground Mode
   - Edit Mode

This PR will also fixes
- RW-4351, a bug with a checkbox in the new dataset modal 
- RW-4350, Project Owners have their actions disabled on recent resource cards

I went with a fairly crude approach for handling the feature flag by sending back "ACTIVE" billing if the locked billing feature is disabled. I did this because it involved adding 2 switches on the backend versus 10+ on the frontend.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -174,7 +181,7 @@ public void setDistinctWorkspaceLimit(int limit) {
             .limit(distinctWorkspacelimit)
             .collect(Collectors.toList());
 
-    final ImmutableMap<Long, FirecloudWorkspaceResponse> idToLiveWorkspace =
+    final Map<Long, DbWorkspace> idToDbWorkspace =
```
ericsong: I'm not opposed to using immutables. This was in the interest of using the least strict type which in this case is Map. IMO, immutability is an implementation detail.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -247,12 +260,21 @@ public boolean hasValidBlobIdIfNotebookNamePresent(DbUserRecentResource dbUserRe
...
+      Map<Long, DbWorkspace> idToDbWorkspace,
+      Map<Long, FirecloudWorkspaceResponse> idToFcWorkspaceResponse,
       DbUserRecentResource dbUserRecentResource) {
     RecentResource resource = TO_CLIENT.apply(dbUserRecentResource);
     FirecloudWorkspaceResponse workspaceDetails =
         idToFcWorkspaceResponse.get(dbUserRecentResource.getWorkspaceId());
-    resource.setPermission(workspaceDetails.getAccessLevel());
+    if (workbenchConfigProvider.get().featureFlags.enableBillingLockout) {
+      resource.setWorkspaceBillingStatus(
+          idToDbWorkspace.get(dbUserRecentResource.getWorkspaceId()).getBillingStatus());
```
ericsong: I decided to return this value as part of the RecentResource API model because the response object is used to construct UI resource cards which needs this info. The alternative was to make a getWorkspaces() call for each workspace in the response list and that seemed too heavy handed.

api/src/main/java/org/pmiops/workbench/workspaces/ManualWorkspaceMapper.java
```diff
@@ -6,22 +6,34 @@
...
 import org.pmiops.workbench.model.SpecificPopulationEnum;
 import org.pmiops.workbench.model.UserRole;
 import org.pmiops.workbench.model.Workspace;
 import org.pmiops.workbench.model.WorkspaceAccessLevel;
+import org.springframework.stereotype.Service;
 
-public class WorkspaceConversionUtils {
+// We should migrate over to the Mapstruct supported WorkspaceMapper
+@Service
+public class ManualWorkspaceMapper {
```
ericsong: I made this injectable so I can pull in the WorkbenchConfig provider. 

Good example of why static utility classes can be hard to refactor.

I am guilty of writing this one heh.

Also, the name is intentionally a bit crude to draw attention to the need to migrate this class to the Mapstruct supported WorkspaceMapper.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -75,16 +75,15 @@
   protected static final int RECENT_WORKSPACE_COUNT = 4;
   private static final Logger log = Logger.getLogger(WorkspaceService.class.getName());
 
-  // Note: Cannot use an @Autowired constructor with this version of Spring
```
ericsong: AFAICS, we're using the `@Autowired` constructor so I removed this comment.

ui/src/app/utils/strings.ts
```diff
@@ -0,0 +1 @@
+export const ACTION_DISABLED_INVALID_BILLING = 'A valid billing account is required for this action';
```
ericsong: I can see this being controversial but I felt like we don't really have a good way to organize shared code on the front end. Ex. no `BillingService` to shove this into.

I've used this kind of approach with a global string file when I worked with Android and felt like it worked pretty well.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
calbach: In the case of the Immutable variants, the general preferred style is actually to use the more specific `Immutable` type where possible. The reason is that it has a semantically meaningful distinction in its interface - it's not just an implementation detail. Code using this variable might need to know it's immutable, i.e. that certain methods will just fail.

Here's the guidance from the Guava library (see: "interfaces", not implementations):
https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableCollection.html

Note that since this is just a local variable, this guidance is not super meaningful, and arguably Map<> is also acceptable in this case (that's probably how I would have written it, TBH).

calbach: Sorry, I should reword "where possible" to "where relevant". In this particular instance, I could see an argument either way, but this is at least to point out that there are legitimate cases where you should be using the `ImmutableMap` type over `Map`.

ericsong: Got it, that makes sense. I've definitely run into cases where I had collection methods fail on me.

ui/src/app/utils/strings.ts
als364: I'm ok with this, I've seen it work before when the number of strings was pretty small. If you end up with more than ten lines in here you want to start thinking about why the string can't be part of a tiny component or whatever but this is totally fine.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
als364: Another alternative would be to change `RecentResource` to include the entire workspace, but then I think you're still calling out to firecloud on each request... just not returning as much.

api/src/main/java/org/pmiops/workbench/workspaces/ManualWorkspaceMapper.java
ericsong: Done - https://precisionmedicineinitiative.atlassian.net/browse/RW-4367

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -185,6 +192,12 @@ public void setDistinctWorkspaceLimit(int limit) {
                                 new AbstractMap.SimpleImmutableEntry<>(
                                     dbWorkspace.getWorkspaceId(), dbWorkspace)))
             .flatMap(Streams::stream)
+            .collect(
+                ImmutableMap.toImmutableMap(
+                    SimpleImmutableEntry::getKey, SimpleImmutableEntry::getValue));
+
+    final Map<Long, FirecloudWorkspaceResponse> idToLiveWorkspace =
```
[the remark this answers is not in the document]
ericsong: Agreed, I felt the same. I'll change it

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
ericsong: yeah... adding the entire workspace feels cleaner but it would defeat the purpose of this change which is to improve performance

PR #3059, opened 2020-01-30
[RW-4132][risk=low] Disallow editing of workspace "request review"

calbach:
This simplifies the flow for the new Zendesk ticket integration. In the future, we might add this capability back in.


NehaBroad:
Question: how will this simplify the zendesk flow?

calbach:
Sorry, that context was on the ticket. It simplifies the flow because we don't need to track:

1. Dismissing review requests after they are initially filed
2. New requests being created for reviews after initial workspace creation
3. Handling of edge cases such as a user turning "review requested" on/off multiple times.

PR #3060, opened 2020-01-30
[RW-3062][risk=no] Fix "Aou" and "All of Us" text across workbench

dolbeew:
Make sure all instances of 'AoU' and 'All of Us' are spelled out and italicized

*Note: The 'All of Us' in `account-creation-options.tsx` is left unitalicized since it needs to be a string as the label for the dropdown option

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/homepage/registration-dashboard.tsx
+++ b/ui/src/app/pages/homepage/registration-dashboard.tsx
@@ -77,7 +77,7 @@ interface RegistrationTask {
   key: string;
   completionPropsKey: string;
   title: string;
-  description: JSX.Element;
+  description: React.ReactNode;
   buttonText: string;
   completedText: string;
   isRefreshable?: boolean;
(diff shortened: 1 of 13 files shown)


ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -77,7 +77,7 @@ interface RegistrationTask {
   key: string;
   completionPropsKey: string;
   title: string;
-  description: string;
+  description: JSX.Element;
```
gjuggler: [no action needed] I've used React.ReactNode for type-annotating in similar scenarios ([code](https://github.com/all-of-us/workbench/blob/331b1228bb7c7b8d392f2acf393c64d82e24a2ef/ui/src/app/pages/workspace/workspace-edit.tsx#L52-L52)). This seems maybe more specific / appropriate, but React.ReactNode may make it possible to avoid wrapping plain-text descriptions in <span> elements to satisfy the type-checker. Either way, we should probably aim to eventually make our type usage consistent here, and/or clarify when to use one vs. the other.

jaycarlton:
Might want to check Rainforest scripts.

ui/src/app/pages/homepage/registration-dashboard.tsx
dolbeew: Nice idea. I went ahead and switched to React.ReactNode for the reason you mentioned. Now the description can be a string or element.

dolbeew:
@gjuggler I probably didn't test 100% of the cases, but I at least hit the less straightforward instances, like the registration tasks.

@jaycarlton Good call on checking Rainforest. This is exactly the type of change that doesn't affect any functionality but would cause some tests to fail.

PR #3061, opened 2020-01-30
[risk=no][RW-2957] fix logger name in ComplianceServiceImpl

als364:
gave it the wrong class. copy paste has failed me again.


als364:
totally not going to jump down that rabbit hole right now.

PR #3065, opened 2020-01-30
[RW-4091][RISK=NO]Fix exportWorkspace error

NehaBroad:
exportToWorkspace call was throwing a lot of 404 error while collecting collaborator list, as part of this PR we now use workspaceService to collect the collaborators list as well as enclose the call in try catch to handle exception 

---
**PR checklist**

- [ X] I have run and tested this change locally

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
@@ -41,7 +41,7 @@ public class RdrExportServiceImpl implements RdrExportService {
   private RdrExportDao rdrExportDao;
   private WorkspaceDao workspaceDao;
   private UserDao userDao;
-  private WorkspaceService workspacesApi;
+  private WorkspaceService workspaceService;
 
   private static final Logger log = Logger.getLogger(RdrExportService.class.getName());
   ZoneOffset offset = OffsetDateTime.now().getOffset();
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -42,6 +41,7 @@
   private RdrExportDao rdrExportDao;
   private WorkspaceDao workspaceDao;
   private UserDao userDao;
+  private WorkspaceService workspacesApi;
```
gjuggler: Naming nit: for consistency w/ the class name, prefer 'workspaceService' as the member variable name.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -245,22 +247,28 @@ private RdrWorkspace toRdrWorkspace(DbWorkspace dbWorkspace) {
...
+            workspaceUserMap.setUserId(
+                (int) userDao.findUserByUsername(userRole.getEmail()).getUserId());
+            workspaceUserMap.setRole(
+                RdrWorkspaceUser.RoleEnum.fromValue(userRole.getRole().toString()));
+            workspaceUserMap.setStatus(RdrWorkspaceUser.StatusEnum.ACTIVE);
+            rdrWorkspace.addWorkspaceUsersItem(workspaceUserMap);
+          });
+    } catch (Exception ex) {
+      log.warning("Exception while retrieving workspace Collaborators");
+      rdrWorkspace.addWorkspaceUsersItem(new RdrWorkspaceUser());
```
gjuggler: Will the RDR really be happy to receive a totally-empty entry for workspace users? I might have expected here to either: (1) skip exporting this workspace entirely, since the collaborators list is pretty important to the state of the workspace, or (2) leave the collaborators list empty.

NehaBroad: As discussed offline ,in case of error we will not log the workspace id for which there was an error and will skip it

PR #3066, opened 2020-01-30
[RW-4159][risk=no] UI environment flags & feature flag cleanup

gjuggler:
This PR contains a bit of prep work and cleanup that's loosely related to the Terms of Service signature UI:

1) Remove our redundant environment.enableAccountPages flag, and instead propagate the WorkbenchConfig.enableNewAccountCreation flag and use that. This will help reduce bugs as we roll out the new-account functionality in the backend & UI.

2) Remove the unused enableHomepageRestyle flag. It's set to true in all environments, so is overdue for removal.

### DIFF
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -5,7 +5,6 @@ import * as fp from 'lodash/fp';
 import * as React from 'react';
 
 import {
-  CardButton,
   Clickable,
 } from 'app/components/buttons';
 import {FadeBox} from 'app/components/containers';
@@ -23,7 +22,6 @@ import {profileApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors, {addOpacity} from 'app/styles/colors';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {AnalyticsTracker} from 'app/utils/analytics';
-import {environment} from 'environments/environment';
 import {
   Profile,
 } from 'generated/fetch';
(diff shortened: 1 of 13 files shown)


ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -277,156 +241,32 @@ export const Homepage = withUserProfile()(class extends React.Component<
       }
     ];
 
-    if (environment.enableHomepageRestyle) {
```
gjuggler: GitHub is making this diff look more "interesting" than it really is. In this if-else statement here, I kept everything within the "if environment.enableHomepageRestyle" and removed everything in the "else".

ui/src/environments/test-env-base.ts
```diff
@@ -13,11 +13,9 @@ export const testEnvironmentBase = {
   zendeskHelpCenterUrl: 'http://aousupporthelp.zendesk.com/hc',
   shibbolethUrl: 'http://mock-nih.dev.test.firecloud.org',
   shouldShowDisplayTag: true,
-  enablePublishedWorkspaces: true,
   inactivityTimeoutSeconds: 99999999999,
   inactivityWarningBeforeSeconds: 5 * 60,
-  enableHomepageRestyle: true,
+  enablePublishedWorkspaces: true,
```
calbach: While you're touching this, I'd propose turning this off everywhere. We're very far away from flipping this flag on, and certainly the current incarnation is not something we're planning to launch.

jaycarlton: We could write a script to look periodically for stale flags that haven't changed in X days and are the same in all environments. You really have to stay on these or they gang up on you.

PR #3068, opened 2020-01-30
[RW-4284][risk=no] Inactive Billing status banner

ericsong:
This adds a banner for users viewing a workspace with inactive billing.

The banner
Screen Shot 2020-01-30 at 4 55 46 PM

I had to refactor the existing status banner a bit to reuse it. Some styling changes were made along the way.

#### Original
Screen Shot 2020-01-30 at 1 27 34 PM

#### New
Screen Shot 2020-01-30 at 5 00 40 PM

Another big change was using `ReactDOM.createPortal` to allow the banner to be rendered from any component. Previously it would render to the top right corner of its parent component so it wouldn't go to the correct position unless the parent fit the entire page.

TIL
- ReactDOM.createPortal allow you to render components anywhere in the DOM. https://reactjs.org/docs/portals.html
- Our method of passing in props through `withCurrentWorkspace` `withUserProfile` etc, pass in the props AFTER the component has already been constructed and rendered. Thus, any state that relies on it must be checked in `componentDidUpdate` and not in the initial constructor.

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/status-alert-banner.tsx
+++ b/ui/src/app/components/status-alert-banner.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 
-import {Button} from 'app/components/buttons';
 import {FlexColumn, FlexRow} from 'app/components/flex';
 import {ClrIcon} from 'app/components/icons';
 import colors, {colorWithWhiteness} from 'app/styles/colors';
@@ -10,6 +10,7 @@ const styles = reactStyles({
   alertBanner: {
     backgroundColor: colorWithWhiteness(colors.highlight, .5),
     color: colors.primary,
+    fontSize: '12px',
     marginRight: '1rem',
     marginTop: '1rem',
     padding: '0.5rem',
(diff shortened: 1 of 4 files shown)


ui/src/app/components/status-alert-banner.tsx
```diff
@@ -18,17 +19,14 @@ const styles = reactStyles({
...
 });
 
-
-
 export interface Props {
-  statusAlertId: number;
   title: string;
   message: string;
-  link: string;
+  footer: JSX.Element;
```
als364: ooh neat. Is the ability to pass elements around built into React or is this something new you brought in?

ericsong: it's built in! Another common way react elements are passed around is through `props.children` which is automatically added to every react component. `props.children` will be set to whatever is inside of a JSX tag.

Ex.
```
<Container>
  <div>
    ... w/e JSX content you want
  <div>
</Container>
```

props.children within the `Container` component will be equal to the `div` JSX element.


https://reactjs.org/docs/jsx-in-depth.html#children-in-jsx

als364: I'd seen that and figured it was a special case (which, I guess, it is.)

PR #3069, opened 2020-01-31
[no ticket][risk=no] Singleton-scoped version of WorkbenchConfig Bean

jaycarlton:
Currently, our WorkbenchConfig is provided by a Request-scoped bean. This is the best option in most cases as it's responsive to changes in the config cache, and also ensures the config retrieved is are the same throughout a request's execution.

The disadvantage of this arrangement is that this configuration information is unavailable in Autowired constructors or Bean function parameters that depend on such constructors, as in most cases these are run outside of any request scope. This means we can't, for example, have one implementation of a bean provided in the local environment and a different one in the deployed environments, since the information about which environment we're running in is only available via the config. Currently I have some code that waits until the first request to call an ugly "init" function in at least one place.

Another nice thing about having this available in constructors is that we can inject objects created at a higher level of abstraction (and depend on the config) instead of the config itself. This provides better encapsulation, as you can argue that very few methods need the whole config object.

The `MonitoredResource` provider is an example of a "thing" that's a singleton but configuration-dependent. If I have more than one service that needs one of these, now I can just inject this provider. 

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
+++ b/api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
@@ -9,7 +9,6 @@ import io.opencensus.stats.ViewManager;
 import io.opencensus.tags.Tagger;
 import io.opencensus.tags.Tags;
 import java.util.UUID;
-import java.util.logging.Level;
 import javax.inject.Provider;
 import org.pmiops.workbench.config.CacheSpringConfiguration;
 import org.pmiops.workbench.config.WorkbenchConfig;
@@ -53,9 +52,10 @@ public class MonitoringSpringConfiguration {
     }
   }
 
-  @Bean()
+  @Bean
   public MonitoredResource getMonitoredResource(
-      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON) Provider<WorkbenchConfig> workbenchConfigProvider,
+      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON)
+          Provider<WorkbenchConfig> workbenchConfigProvider,
       @Qualifier(APP_ENGINE_NODE_ID) Provider<String> appEngineNodeIdProvider) {
     return MonitoredResource.newBuilder()
         .setType(MONITORED_RESOURCE_TYPE)
@@ -76,5 +76,4 @@ public class MonitoringSpringConfiguration {
   private String makeRandomNodeId() {
     return UNKNOWN_INSTANCE_PREFIX + UUID.randomUUID().toString();
   }
-
 }
--- a/api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
+++ b/api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
@@ -1,15 +1,33 @@
 package org.pmiops.workbench.monitoring;
 
+import com.google.api.MonitoredResource;
+import com.google.appengine.api.modules.ModulesException;
+import com.google.appengine.api.modules.ModulesService;
 import io.opencensus.stats.Stats;
 import io.opencensus.stats.StatsRecorder;
 import io.opencensus.stats.ViewManager;
 import io.opencensus.tags.Tagger;
 import io.opencensus.tags.Tags;
+import java.util.UUID;
+import java.util.logging.Level;
+import javax.inject.Provider;
+import org.pmiops.workbench.config.CacheSpringConfiguration;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 public class MonitoringSpringConfiguration {
+
+  private static final String UNKNOWN_INSTANCE_PREFIX = "unknown-";
+  private static final String MONITORED_RESOURCE_TYPE = "generic_node";
+  private static final String PROJECT_ID_LABEL = "project_id";
+  private static final String LOCATION_LABEL = "location";
+  private static final String NAMESPACE_LABEL = "namespace";
+  private static final String NODE_ID_LABEL = "node_id";
+
+  public static final String APP_ENGINE_NODE_ID = "appEngineNodeId";
 
   @Bean
   public ViewManager getViewManager() {
@@ -25,4 +43,38 @@ public class MonitoringSpringConfiguration {
   public Tagger getTagger() {
     return Tags.getTagger();
   }
+
+  @Bean(name = APP_ENGINE_NODE_ID)
+  public String getAppEngineNodeId(ModulesService modulesService) {
+    try {
+      return modulesService.getCurrentInstanceId();
+    } catch (ModulesException e) {
+      return makeRandomNodeId();
+    }
+  }
+
+  @Bean()
+  public MonitoredResource getMonitoredResource(
+      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON) Provider<WorkbenchConfig> workbenchConfigProvider,
+      @Qualifier(APP_ENGINE_NODE_ID) Provider<String> appEngineNodeIdProvider) {
+    return MonitoredResource.newBuilder()
+        .setType(MONITORED_RESOURCE_TYPE)
+        .putLabels(PROJECT_ID_LABEL, workbenchConfigProvider.get().server.projectId)
+        .putLabels(LOCATION_LABEL, workbenchConfigProvider.get().server.appEngineLocationId)
+        .putLabels(NAMESPACE_LABEL, workbenchConfigProvider.get().server.shortName.toLowerCase())
+        .putLabels(NODE_ID_LABEL, appEngineNodeIdProvider.get())
+        .build();
+  }
+
+  /**
+   * Stackdriver instances have very long, random ID strings. When running locally, however, the
+   * ModulesService throws an exception, so we need to assign non-conflicting and non-repeating ID
+   * strings.
+   *
+   * @return
+   */
+  private String makeRandomNodeId() {
+    return UNKNOWN_INSTANCE_PREFIX + UUID.randomUUID().toString();
+  }
+
 }
--- a/api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
+++ b/api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
@@ -8,9 +8,7 @@ import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.inject.Provider;
-import org.pmiops.workbench.config.CacheSpringConfiguration;
 import org.pmiops.workbench.config.WorkbenchConfig;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
 /**
@@ -24,15 +22,12 @@ public class StackdriverStatsExporterService {
       Logger.getLogger(StackdriverStatsExporterService.class.getName());
   private static final String STACKDRIVER_CUSTOM_METRICS_PREFIX = "custom.googleapis.com/";
 
+  private MonitoredResource monitoredResource;
   private Provider<WorkbenchConfig> workbenchConfigProvider;
-  private Provider<MonitoredResource> monitoredResourceProvider;
 
   public StackdriverStatsExporterService(
-      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON)
-          Provider<WorkbenchConfig> workbenchConfigProvider,
-      Provider<MonitoredResource> monitoredResourceProvider) {
-    this.monitoredResourceProvider = monitoredResourceProvider;
-    logger.warning("Making StackdriverStatsExporterService");
+      MonitoredResource monitoredResource, Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.monitoredResource = monitoredResource;
     this.workbenchConfigProvider = workbenchConfigProvider;
   }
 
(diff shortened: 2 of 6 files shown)


api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -25,93 +23,41 @@
...
-  private boolean initialized;
   private Provider<WorkbenchConfig> workbenchConfigProvider;
-  private ModulesService modulesService;
+  private Provider<MonitoredResource> monitoredResourceProvider;
 
   public StackdriverStatsExporterService(
-      Provider<WorkbenchConfig> workbenchConfigProvider, ModulesService modulesService) {
+      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON)
+          Provider<WorkbenchConfig> workbenchConfigProvider,
+      Provider<MonitoredResource> monitoredResourceProvider) {
```
freemabd: Since monitoredResourceProvider is created using Singleton workbenchConfigProvider above in MonitoringSpringConfiguration, why do we still need a reference to workbenchConfigProvider here?

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
```diff
@@ -25,93 +23,41 @@
...
     }
   }
 
   @VisibleForTesting
   public StackdriverStatsConfiguration makeStackdriverStatsConfiguration() {
     return StackdriverStatsConfiguration.builder()
         .setMetricNamePrefix(STACKDRIVER_CUSTOM_METRICS_PREFIX)
-        .setProjectId(getProjectId())
-        .setMonitoredResource(makeMonitoredResource())
+        .setProjectId(workbenchConfigProvider.get().server.projectId)
```
freemabd: Seems like the projectId can be fetched from monitoredResourceProvider

common-api/src/main/java/org/pmiops/workbench/config/CacheSpringConfiguration.java
```diff
@@ -66,14 +70,30 @@ public static FeaturedWorkspacesConfig lookupFeaturedWorkspacesConfig(
...
       @Qualifier("configCache") LoadingCache<String, Object> configCache)
       throws ExecutionException {
     return lookupWorkbenchConfig(configCache);
   }
 
+  /**
+   * Singleton version for use in Service constructors and Bean function arguments. Not updated
+   * dynamically
+   */
+  @Bean(name = WORKBENCH_CONFIG_SINGLETON)
```
freemabd: Not sure how often workbench config changes happen off-schedule or without an app deploy. If this does happen off-schedule, seems like this could bite us.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
jaycarlton: yep.

common-api/src/main/java/org/pmiops/workbench/config/CacheSpringConfiguration.java
jaycarlton: The primary bean is still the request-scoped one. How do you mean "bite"?

jaycarlton:
Seeing some interesting errors in the `ApplicationTest` integration test, even though this loads when I launch the actual application. It instantiates beans that aren't actually in the type to be retrieved, and the thread/context in which things are being evaluated is somehow not aligned with what needs to happen there 🤔 .

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
```diff
@@ -25,4 +42,38 @@ public StatsRecorder getStatsRecorder() {
...
+      return modulesService.getCurrentInstanceId();
+    } catch (ModulesException e) {
+      return makeRandomNodeId();
+    }
+  }
+
+  @Bean
+  public MonitoredResource getMonitoredResource(
+      @Qualifier(CacheSpringConfiguration.WORKBENCH_CONFIG_SINGLETON)
+          Provider<WorkbenchConfig> workbenchConfigProvider,
```
calbach: Why inject providers here instead of just injecting the thing itself?

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
calbach: +1, I don't think there are any situations where we'd want to inject the SINGULAR qualified provider - only rationale I could see would be for testing purposes

api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java
jaycarlton: 😮 Didn't know I could do that. I was thinking value objects were different from services somehow.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
jaycarlton: Very good eye. We don't, really.

api/src/main/java/org/pmiops/workbench/monitoring/StackdriverStatsExporterService.java
jaycarlton: One caveat is that not every `MonitoredResource` type will have a label for `PROJECT_ID`. If I make this change, that makes this code a bit less reusable, since this service has no business knowing that we're in a GCP project.

common-api/src/main/java/org/pmiops/workbench/config/CacheSpringConfiguration.java
freemabd: Is there a chance that MonitoredResource could be used places besides a constructor and gets stale if the workbench config changes without a restart/deploy?

jaycarlton: Yes, but that's by design. There are things in the config that are dynamically scoped but that should never be changed at runtime. for instance the project names, environment name, etc. (I'm working on a config document now, because I'm seeing both low-hanging fruit and issues on the horizon. The one-size-fits-all nature of how we provide it is one issue.)

I intend the MonitoredResource to be created once per process, since its job is to describe the "resource being monitored", i.e. this node of this project, with this namespace and location. It has two purposes: (1) to allow multiple processes to write to the same time series without trigging rate limit exceptions, and (2) to provide context to all the graphs for the thing being monitored.

The "scope" of it should be the entire lifetime of the process. Since we can't change regions or projectIds during execution, it would be an error to pick them up here.

But, and here's the sad thing, if I did move all that to a request scope, I wouldn't be able to use it in constructors or when building singleton-scoped things. But it's effectively a constant for the lifetime of the process.

freemabd: Thanks for the info, that helps a lot.

jaycarlton:
@freemabd, do you know why might fail on this [test case](https://circleci.com/gh/all-of-us/workbench/67133?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)? I believe you wrote it. Trying to think if that's a false positive or if I need to rethink this slightly.

freemabd:
Seems that this test no longer runs in intellij(making this harder to debug), getting the following error: 

Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

When running the test with project.rb like so:
./project.rb integration --tests "org.pmiops.workbench.ApplicationTest.contextLoads" --project-prop verboseTestLogging=yes
I'm able to reproduce the error from your link

Seems that the main config needs to be updated to reflect the latest workbenchConfig -> https://github.com/all-of-us/workbench/blob/2dc27b2082433308c0610a35c820e9adef5081f2/api/src/integration/resources/data.sql#L3

I tried updating the main config and we get passed that null pointer and on to a new exception:
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [java.lang.String]: Factory method 'getAppEngineNodeId' threw exception; nested exception is java.lang.IllegalStateException: Operation not allowed in a thread that is neither the original request thread nor a thread created by ThreadManager
Caused by: java.lang.IllegalStateException: Operation not allowed in a thread that is neither the original request thread nor a thread created by ThreadManager

So my guess is that moduleService is not being initialize properly(not sure if we can mock this) and throwing this exception: https://github.com/all-of-us/workbench/blob/2dc27b2082433308c0610a35c820e9adef5081f2/api/src/main/java/org/pmiops/workbench/monitoring/MonitoringSpringConfiguration.java#L49-L49

Maybe this test is more trouble than it's worth. The intention was to catch any dependency injection problems before they get merged.

gjuggler:
> Seems that this test no longer runs in intellij(making this harder to debug), getting the following error:
> 
> Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
> 

Ooh, this is mostly unrelated to the singleton bean breakage, but I might be responsible for breaking ApplicationTest running on IntelliJ. My PR from a couple weeks back https://github.com/all-of-us/workbench/pull/3016 had caused a similar error message on the same test, in api-local-test on CircleCI, due to the new IamCredentialsClient bean which is loaded as a Spring bean using ADCs during application startup. I fixed this for Circle, but hadn't been paying attention to whether / how this runs in IntelliJ. See https://github.com/all-of-us/workbench/pull/3016#issuecomment-578859763 for more discussion on that issue.

@freemabd I see that for our bigquery tests, there's code that explicitly looks to sa-key.json for credentials instead of allowing ADCs to be loaded (https://github.com/all-of-us/workbench/blob/aedd1db6ea286126be26bdb891525619bb806872/api/src/bigquerytest/java/org/pmiops/workbench/testconfig/TestBigQueryConfig.java#L23-L23). Should we adopt that pattern for the integration tests as well?

gjuggler:
> Maybe this test is more trouble than it's worth. The intention was to catch any dependency injection problems before they get merged.

Personally I've found this test to be very helpful – it's caught a number of DI errors during my development flow and pre-PR-merging that I might have otherwise missed. I'm not sure if others have found the same, though.

freemabd:
> > Seems that this test no longer runs in intellij(making this harder to debug), getting the following error:
> > Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
> 
> Ooh, this is mostly unrelated to the singleton bean breakage, but I might be responsible for breaking ApplicationTest running on IntelliJ. My PR from a couple weeks back #3016 had caused a similar error message on the same test, in api-local-test on CircleCI, due to the new IamCredentialsClient bean which is loaded as a Spring bean using ADCs during application startup. I fixed this for Circle, but hadn't been paying attention to whether / how this runs in IntelliJ. See [#3016 (comment)](https://github.com/all-of-us/workbench/pull/3016#issuecomment-578859763) for more discussion on that issue.
> 
> @freemabd I see that for our bigquery tests, there's code that explicitly looks to sa-key.json for credentials instead of allowing ADCs to be loaded (
> 
> https://github.com/all-of-us/workbench/blob/aedd1db6ea286126be26bdb891525619bb806872/api/src/bigquerytest/java/org/pmiops/workbench/testconfig/TestBigQueryConfig.java#L23-L23
> 
> ). Should we adopt that pattern for the integration tests as well?



> > Seems that this test no longer runs in intellij(making this harder to debug), getting the following error:
> > Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
> 
> Ooh, this is mostly unrelated to the singleton bean breakage, but I might be responsible for breaking ApplicationTest running on IntelliJ. My PR from a couple weeks back #3016 had caused a similar error message on the same test, in api-local-test on CircleCI, due to the new IamCredentialsClient bean which is loaded as a Spring bean using ADCs during application startup. I fixed this for Circle, but hadn't been paying attention to whether / how this runs in IntelliJ. See [#3016 (comment)](https://github.com/all-of-us/workbench/pull/3016#issuecomment-578859763) for more discussion on that issue.
> 
> @freemabd I see that for our bigquery tests, there's code that explicitly looks to sa-key.json for credentials instead of allowing ADCs to be loaded (
> 
> https://github.com/all-of-us/workbench/blob/aedd1db6ea286126be26bdb891525619bb806872/api/src/bigquerytest/java/org/pmiops/workbench/testconfig/TestBigQueryConfig.java#L23-L23
> 
> ). Should we adopt that pattern for the integration tests as well?

Sounds reasonable to me.

freemabd:
>  Should we adopt that pattern for the integration tests as well?

Seems like a good place to start

jaycarlton:
Oh good, so it sounds like a test problem and not something fundamentally broken with what I'm doing.

Did you want me to try that change on this branch, or would it be better separately? This PR is not in the critical path.

gjuggler:
> Oh good, so it sounds like a test problem and not something fundamentally broken with what I'm doing.
> 
> Did you want me to try that change on this branch, or would it be better separately? This PR is not in the critical path.

To be clear: the change I referred to would handle the first part of Brian's comment (IntelliJ breakage), not the second or third. The second and third parts seem to be the real error, while IntelliJ breakage just makes it harder to run the tests from IntelliJ instead of project.rb.

jaycarlton:
closing this for now, as I no longer have the immediate use case as a requirement. I think the basic idea is going to be important though.

PR #3070, opened 2020-01-31
[risk=no][RW-2957] seconds/millis issue

als364:
Moodle sends us epochs in seconds and everything we use for time in Java thinks everything should be in milliseconds.


api/src/test/java/org/pmiops/workbench/db/dao/UserServiceTest.java
```diff
@@ -53,7 +53,9 @@
 
   // An arbitrary timestamp to use as the anchor time for access module test cases.
   private static final Instant START_INSTANT = Instant.parse("2000-01-01T00:00:00.00Z");
+  @Deprecated
   private static final long TIMESTAMP_MSECS = START_INSTANT.toEpochMilli();
```
jmthibault79: There is still one non-deprecated usage here: please change line 343 so it doesn't use TIMESTAMP_MSECS.  `Timestamp.from(START_INSTANT)` would fine in this context.

PR #3072, opened 2020-02-01
[RW-4132][risk=moderate] Integrate with Zendesk for workspace review requests

calbach:
- Note: this pulls in a new 3p Java library (ZD unofficial client library)
- Integrates using an unauthenticated Zendesk client via client lib. See Readme.md for details.
- On create/clone, if review is requested files a Zendesk request
- Zendesk request ticket contains both formatted workspace description (similar to edit/about pages) as well as a raw JSON view of the research purpose, in the event that we fail to update the ticket formatter when making changes to the workspace model.

Manually tested this in the sandbox env.

Future work:
- The Zendesk comment API supports HTML comments. A future version could integrate an HTML templating library to render a more readable view.
- A path for setting up an authenticated client is described in the readme, if we wanted to do subsequent ticket updates later.

Sample ticket output:
```
[Workspace name]
v5 Legit Workspace

[Workspace ID]
aou-rw-local1-94faadfb/v5legitworkspace

[Review requester]
CH Albach (calbach@fake-research-aou.org, null)

[Primary purpose of project]
- Disease focused research: cancerophobia
- Genetic Research
- Drug/Therapeutics Development Research
- For-Profit Purpose
- Educational Purpose
- Other Purpose: QA

[Provide the reason for choosing All of Us data for your investigation]
I would like to leverage the diversity of the data.

[What are the specific scientific question(s) you intend to study]
Whether X can affect Y.

[What are your anticipated findings from this study]
Huge scientific breakthroughs.

[Population area(s) of focus]
ACCESS_TO_CARE, RACE_ETHNICITY, SEX, GEOGRAPHY

[Raw research purpose JSON]
{
...
}
```

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/zendesk/ZendeskRequestsTest.java
+++ b/api/src/test/java/org/pmiops/workbench/zendesk/ZendeskRequestsTest.java
@@ -1,0 +1,115 @@
+package org.pmiops.workbench.zendesk;
+
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assertWithMessage;
+
+import com.google.common.collect.ImmutableList;
+import org.junit.Before;
+import org.junit.Test;
+import org.pmiops.workbench.db.model.DbUser;
+import org.pmiops.workbench.model.ResearchPurpose;
+import org.pmiops.workbench.model.SpecificPopulationEnum;
+import org.pmiops.workbench.model.Workspace;
+
+public class ZendeskRequestsTest {
+  private DbUser user;
+
+  @Before
+  public void setUp() {
+    user = new DbUser();
+    user.setGivenName("Falco");
+    user.setFamilyName("Lombardi");
+    user.setUsername("flaco@fake-research-aou.org");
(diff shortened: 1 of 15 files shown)


calbach:
Snyk doesn't like this library - I will take a look

api/src/test/java/org/pmiops/workbench/zendesk/ZendeskRequestsTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class ZendeskRequestsTest {
+  private DbUser user;
+
+  @Before
+  public void setUp() {
+    user = new DbUser();
+    user.setGivenName("Falco");
+    user.setFamilyName("Lombardi");
+    user.setUsername("flaco@fake-research-aou.org");
```
als364: Typo here. Seems like it would be difficult to type with one's wings.

calbach: Hah - this was actually an intentional typo, but I've fixed it here to avoid confusion.

PR #3073, opened 2020-02-03
[RW-4160][risk=no] DRAFT Institution API+CRUD

jmthibault79:
Description:
Institution Swagger entity, API, Controller, Service, DAOs, DB classes, and tests

TODO: Liquibase, gate on a new Authority

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
@@ -1,6 +1,6 @@
 package org.pmiops.workbench.db.dao;
 
-import java.util.List;
+import java.util.Set;
 import org.pmiops.workbench.db.model.DbInstitution;
 import org.pmiops.workbench.db.model.DbInstitutionEmailAddress;
 import org.springframework.data.repository.CrudRepository;
@@ -8,7 +8,7 @@ import org.springframework.data.repository.CrudRepository;
 public interface InstitutionEmailAddressDao
     extends CrudRepository<DbInstitutionEmailAddress, Long> {
 
-  List<DbInstitutionEmailAddress> findAllByInstitution(DbInstitution institution);
+  Set<DbInstitutionEmailAddress> findAllByInstitution(DbInstitution institution);
 
   default void deleteAllByInstitution(DbInstitution institution) {
     delete(findAllByInstitution(institution));
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
@@ -1,0 +1,16 @@
+package org.pmiops.workbench.db.dao;
+
+import java.util.List;
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailAddress;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailAddressDao
+    extends CrudRepository<DbInstitutionEmailAddress, Long> {
+
+  List<DbInstitutionEmailAddress> findAllByInstitution(DbInstitution institution);
+
+  default void deleteAllByInstitution(DbInstitution institution) {
+    delete(findAllByInstitution(institution));
+  }
+}
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
@@ -1,0 +1,15 @@
+package org.pmiops.workbench.db.dao;
+
+import java.util.List;
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
+
+  List<DbInstitutionEmailDomain> findAllByInstitution(DbInstitution institution);
+
+  default void deleteAllByInstitution(DbInstitution institution) {
+    delete(findAllByInstitution(institution));
+  }
+}
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
@@ -1,13 +1,13 @@
 package org.pmiops.workbench.db.dao;
 
-import java.util.List;
+import java.util.Set;
 import org.pmiops.workbench.db.model.DbInstitution;
 import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
 import org.springframework.data.repository.CrudRepository;
 
 public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
 
-  List<DbInstitutionEmailDomain> findAllByInstitution(DbInstitution institution);
+  Set<DbInstitutionEmailDomain> findAllByInstitution(DbInstitution institution);
 
   default void deleteAllByInstitution(DbInstitution institution) {
     delete(findAllByInstitution(institution));
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -1,0 +1,95 @@
+package org.pmiops.workbench.db.model;
+
+import java.util.List;
+import javax.persistence.*;
+
+@Entity
+@Table(name = "institution")
+public class DbInstitution {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @Column(name = "institution_id")
+  private long institutionId; // primary opaque key for DB use only
+
+  @Column(name = "api_id", nullable = false, unique = true)
+  private String apiId; // unique key exposed to API
+
+  @Column(name = "long_name", nullable = false)
+  private String longName;
+
+  @Column(name = "organization_type_enum")
+  private Short organizationTypeEnum;
+
+  @Column(name = "organization_type_other_text")
+  private String organizationTypeOtherText;
+
+  @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
+  private List<DbInstitutionEmailDomain> emailDomains;
+
+  @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
+  private List<DbInstitutionEmailAddress> emailAddresses;
+
+  public DbInstitution() {}
+
+  public DbInstitution(final String apiId, final String longName) {
+    setApiId(apiId);
+    setLongName(longName);
+  }
+
+  public long getInstitutionId() {
+    return institutionId;
+  }
+
+  public void setInstitutionId(long institutionId) {
+    this.institutionId = institutionId;
+  }
+
+  public String getApiId() {
+    return apiId;
+  }
+
+  public void setApiId(String apiId) {
+    this.apiId = apiId;
+  }
+
+  public String getLongName() {
+    return longName;
+  }
+
+  public void setLongName(String longName) {
+    this.longName = longName;
+  }
+
+  public Short getOrganizationTypeEnum() {
+    return organizationTypeEnum;
+  }
+
+  public void setOrganizationTypeEnum(Short organizationTypeEnum) {
+    this.organizationTypeEnum = organizationTypeEnum;
+  }
+
+  public String getOrganizationTypeOtherText() {
+    return organizationTypeOtherText;
+  }
+
+  public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
+    this.organizationTypeOtherText = organizationTypeOtherText;
+  }
+
+  public List<DbInstitutionEmailDomain> getEmailDomains() {
+    return emailDomains;
+  }
+
+  public void setEmailDomains(List<DbInstitutionEmailDomain> emailDomains) {
+    this.emailDomains = emailDomains;
+  }
+
+  public List<DbInstitutionEmailAddress> getEmailAddresses() {
+    return emailAddresses;
+  }
+
+  public void setEmailAddresses(List<DbInstitutionEmailAddress> emailAddresses) {
+    this.emailAddresses = emailAddresses;
+  }
+}
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -1,7 +1,14 @@
 package org.pmiops.workbench.db.model;
 
-import java.util.List;
-import javax.persistence.*;
+import java.util.Set;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
 
 @Entity
 @Table(name = "institution")
@@ -10,13 +17,13 @@ public class DbInstitution {
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
   @Column(name = "institution_id")
-  private long institutionId; // primary opaque key for DB use only
+  private long dbId; // primary opaque key for DB use only
 
-  @Column(name = "api_id", nullable = false, unique = true)
-  private String apiId; // unique key exposed to API
+  @Column(name = "short_name", nullable = false, unique = true)
+  private String shortName; // unique key exposed to API
 
-  @Column(name = "long_name", nullable = false)
-  private String longName;
+  @Column(name = "display_name", nullable = false)
+  private String displayName;
 
   @Column(name = "organization_type_enum")
   private Short organizationTypeEnum;
@@ -25,40 +32,32 @@ public class DbInstitution {
   private String organizationTypeOtherText;
 
   @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
-  private List<DbInstitutionEmailDomain> emailDomains;
+  private Set<DbInstitutionEmailDomain> emailDomains;
 
   @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
-  private List<DbInstitutionEmailAddress> emailAddresses;
+  private Set<DbInstitutionEmailAddress> emailAddresses;
 
   public DbInstitution() {}
 
-  public DbInstitution(final String apiId, final String longName) {
-    setApiId(apiId);
-    setLongName(longName);
+  public DbInstitution(final String shortName, final String displayName) {
+    setShortName(shortName);
+    setDisplayName(displayName);
   }
 
-  public long getInstitutionId() {
-    return institutionId;
+  public String getShortName() {
+    return shortName;
   }
 
-  public void setInstitutionId(long institutionId) {
-    this.institutionId = institutionId;
+  public void setShortName(String shortName) {
+    this.shortName = shortName;
   }
 
-  public String getApiId() {
-    return apiId;
+  public String getDisplayName() {
+    return displayName;
   }
 
-  public void setApiId(String apiId) {
-    this.apiId = apiId;
-  }
-
-  public String getLongName() {
-    return longName;
-  }
-
-  public void setLongName(String longName) {
-    this.longName = longName;
+  public void setDisplayName(String displayName) {
+    this.displayName = displayName;
   }
 
   public Short getOrganizationTypeEnum() {
@@ -77,19 +76,19 @@ public class DbInstitution {
     this.organizationTypeOtherText = organizationTypeOtherText;
   }
 
-  public List<DbInstitutionEmailDomain> getEmailDomains() {
+  public Set<DbInstitutionEmailDomain> getEmailDomains() {
     return emailDomains;
   }
 
-  public void setEmailDomains(List<DbInstitutionEmailDomain> emailDomains) {
+  public void setEmailDomains(Set<DbInstitutionEmailDomain> emailDomains) {
     this.emailDomains = emailDomains;
   }
 
-  public List<DbInstitutionEmailAddress> getEmailAddresses() {
+  public Set<DbInstitutionEmailAddress> getEmailAddresses() {
     return emailAddresses;
   }
 
-  public void setEmailAddresses(List<DbInstitutionEmailAddress> emailAddresses) {
+  public void setEmailAddresses(Set<DbInstitutionEmailAddress> emailAddresses) {
     this.emailAddresses = emailAddresses;
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
@@ -1,6 +1,14 @@
 package org.pmiops.workbench.db.model;
 
-import javax.persistence.*;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
 
 @Entity
 @Table(name = "institution_email_address")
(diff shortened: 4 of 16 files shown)


api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.persistence.*;
+
+@Entity
+@Table(name = "institution")
+public class DbInstitution {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @Column(name = "institution_id")
+  private long institutionId; // primary opaque key for DB use only
```
jaycarlton: nit: you can just name this `dbId` or `id`, since it's a private member of the Institution entity object.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Column(name = "api_id", nullable = false, unique = true)
```
jaycarlton: Which API? There are a number of them running around.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+  private String apiId; // unique key exposed to API
+
+  @Column(name = "long_name", nullable = false)
+  private String longName;
+
+  @Column(name = "organization_type_enum")
```
jaycarlton: I wouldn't put the type in the column name. SQL will let you reflect that. Plus, if you change the type (like to a real SQL enum) later, you'd have to rename at the same time, which is a headache.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+  private Short organizationTypeEnum;
+
+  @Column(name = "organization_type_other_text")
+  private String organizationTypeOtherText;
+
+  @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
+  private List<DbInstitutionEmailDomain> emailDomains;
+
+  @OneToMany(mappedBy = "institution", fetch = FetchType.LAZY)
+  private List<DbInstitutionEmailAddress> emailAddresses;
```
jaycarlton: aside: do we need/want a mapping between email addresses and domains? Like would you want to do a report with all the institutions, then their domains, then the addresses? Not needed yet I guess.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public DbInstitution() {}
+
+  public DbInstitution(final String apiId, final String longName) {
+    setApiId(apiId);
```
jaycarlton: I don't mind overloaded constructors in Java, and use them frequently. One thing to watch out for though, is if different versions of the constructor don't produce the same object when given consistent results.

For example, if I have a 1-arg constructor, and a 2-arg constructor, I generally expect the first one to call the second with `Constructor(myOnlyArg, DEFAULT_SECOND_ARG)` so that both give consistent results.

All that to say, I'd either drop your no-arg constructor, or have it call the 2-arg one with some reasonable default values. If defaults can't be chosen safely, then that's a smell you don't want a no-arg ctor.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public void setApiId(String apiId) {
+    this.apiId = apiId;
+  }
+
+  public String getLongName() {
+    return longName;
+  }
+
+  public void setLongName(String longName) {
```
jaycarlton: nit: "longName" isn't super clear to me.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
+    this.organizationTypeOtherText = organizationTypeOtherText;
+  }
+
+  public List<DbInstitutionEmailDomain> getEmailDomains() {
+    return emailDomains;
+  }
+
+  public void setEmailDomains(List<DbInstitutionEmailDomain> emailDomains) {
```
jaycarlton: I'd use a `Set` here, as we don't want duplicates if I understand correctly. Similarly when returning them in the get() method.

Also make sure you have hash codes and equals() overrides in the entity classes.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
...
+    this.emailDomains = emailDomains;
+  }
+
+  public List<DbInstitutionEmailAddress> getEmailAddresses() {
+    return emailAddresses;
+  }
+
+  public void setEmailAddresses(List<DbInstitutionEmailAddress> emailAddresses) {
```
jaycarlton: Use a Set here, for uniqueness and to indicate order doesn't matter.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -0,0 +1,49 @@
+package org.pmiops.workbench.db.model;
+
+import javax.persistence.*;
```
jaycarlton: spotless might ding you here, but there's an IJ setting somewhere to let you turn this off for everything except a whitelist, e.g. Mockito.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Entity
+@Table(name = "institution_email_address")
+public class DbInstitutionEmailAddress {
+
+  @Id
```
jaycarlton: I like how this gives you the flexibility to have an email belong to more than one institution someday. (This should be in your test planning too).

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -0,0 +1,49 @@
...
+  private long institutionEmailAddressId;
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "institution_id")
+  private DbInstitution institution;
+
+  @Column(name = "email_address", nullable = false)
+  private String emailAddress;
+
+  public DbInstitutionEmailAddress(DbInstitution institution, String emailAddress) {
```
jaycarlton: nit: I'd probably make this class an immutable object and just give it a builder, then take away the set() functions. Unless we expect to update the fields separately, I'm guessing we just add & remove whole rows from this table. I wouldn't expose more degrees of freedom than you need to support.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
gjuggler: I'm pretty sure we don't have much choice here, see https://stackoverflow.com/questions/32467520/hibernate-and-no-arg-constructor/32467567 and https://docs.jboss.org/hibernate/core/3.5/reference/en/html/persistent-classes.html#persistent-classes-pojo-constructor

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
```diff
@@ -0,0 +1,15 @@
+package org.pmiops.workbench.db.dao;
+
+import java.util.List;
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
+
+  List<DbInstitutionEmailDomain> findAllByInstitution(DbInstitution institution);
```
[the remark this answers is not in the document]
jmthibault79: No.  The user explicitly chooses an institution, and we confirm whether their email matches one of its patterns.  We have no uniqueness constraint here, either.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -0,0 +1,95 @@
+package org.pmiops.workbench.db.model;
+
+import java.util.List;
+import javax.persistence.*;
```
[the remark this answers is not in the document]
jmthibault79: Linter did it

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: the workbench's swagger api - do you have a different name preference?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: Because the string+enum construction is a bit wacky, I want to make this explicit.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: Does that help, though?  We can't merge if spotless doesn't like it.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: Is that possible with our DB paradigm?

api/src/main/java/org/pmiops/workbench/db/model/DbStorageEnums.java
```diff
@@ -262,6 +252,23 @@ public static Short degreeToStorage(Degree degree) {
...
+  private static final BiMap<OrganizationType, Short> CLIENT_TO_STORAGE_ORGANIZATION_TYPE =
+      ImmutableBiMap.<OrganizationType, Short>builder()
+          .put(OrganizationType.ACADEMIC_RESEARCH_INSTITUTION, (short) 0)
+          .put(OrganizationType.INDUSTRY, (short) 1)
+          .put(OrganizationType.EDUCATIONAL_INSTITUTION, (short) 2)
+          .put(OrganizationType.HEALTH_CENTER_NON_PROFIT, (short) 3)
+          .put(OrganizationType.OTHER, (short) 4)
+          .build();
+
+  public static OrganizationType organizationTypeFromStorage(Short organizationType) {
```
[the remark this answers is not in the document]
jaycarlton: [Found it](https://github.com/all-of-us/workbench/pull/2718). LMK if you want help. Not saying you have to do it, but it's super fun.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: Never mind.  Was able to fix.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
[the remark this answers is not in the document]
jmthibault79: Actually that appears to be inaccurate.  The delineated version passed the linter.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: Dunno, I'm not seeing it.  Certainly not for existing stories.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test
+  public void testDao() {
+    final DbInstitution testInst = new DbInstitution("Broad", "The Broad Institute");
+    institutionDao.save(testInst);
+    assertThat(institutionDao.findOneByApiId("Broad")).isEqualTo(testInst);
+    assertThat(institutionDao.findAll()).hasSize(1);
+
+    // update existing entity, don't change size
+
+    testInst.setApiId("Verily");
```
jaycarlton: "ApiId" is hard to read. I'd just name this the `shortName` if I'm understanding this.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -0,0 +1,69 @@
...
+    institutionDao.save(testInst);
+    assertThat(institutionDao.findAll()).hasSize(1);
+    assertThat(institutionDao.findOneByApiId("Verily")).isEqualTo(testInst);
+
+    testInst.setLongName("Yea, Verily");
```
jaycarlton: Cool, but might be a more robust test if the short name isn't a substring of the long name.

api/src/test/java/org/pmiops/workbench/institution/InstitutionServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void test_InstitutionNotFoundDelete() {
+    service.deleteInstitution("missing");
+  }
+
+  @Test(expected = DataIntegrityViolationException.class)
+  public void test_nonUniqueIds() {
+    service.createInstitution(new Institution().id("test").longName("We are all individuals"));
```
jaycarlton: So if I have "UA" (University of Alabama), and "UA" (University of Alberta), then one institution is going to be unhappy. In what situation do we see the short name and long name? Should the long name be unique too (or instead)?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: ok.  How about displayName?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: tests now show this

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: I removed a few sets we don't need - but yeah looks like we are obligate mutators

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailDomain.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "institution_email_domain")
+public class DbInstitutionEmailDomain {
```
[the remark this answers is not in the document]
jmthibault79: Yes 1:many

Not clear how this table differs from what you're looking for.

api/src/main/java/org/pmiops/workbench/institution/InstitutionController.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class InstitutionController implements InstitutionApiDelegate {
+
+  private final Provider<InstitutionService> institutionServiceProvider;
+
+  @Autowired
+  InstitutionController(Provider<InstitutionService> institutionServiceProvider) {
```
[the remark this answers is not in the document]
jmthibault79: ok I'll update

api/src/main/java/org/pmiops/workbench/institution/InstitutionService.java
```diff
@@ -0,0 +1,16 @@
+package org.pmiops.workbench.institution;
+
+import java.util.List;
+import org.pmiops.workbench.model.Institution;
+
+public interface InstitutionService {
+  List<Institution> getInstitutions();
+
+  Institution getInstitution(final String id);
```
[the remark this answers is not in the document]
jmthibault79: Optional makes sense here - updated

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    institutionDaoProvider.get().delete(getDbInstitution(id));
+  }
+
+  @Override
+  public Institution updateInstitution(final String id, final Institution institutionToUpdate) {
+    return toModelClass(saveInstitution(institutionToUpdate, getDbInstitution(id)));
+  }
+
+  @NotNull
+  private DbInstitution getDbInstitution(String id) {
```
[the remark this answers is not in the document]
jmthibault79: pushed null/optional checking up a level

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    dbClass.setOrganizationTypeEnum(
+        DbStorageEnums.organizationTypeToStorage(modelClass.getOrganizationTypeEnum()));
+    dbClass.setOrganizationTypeOtherText(modelClass.getOrganizationTypeOtherText());
+
+    // save so the domain and address DAOs have something to reference
+    institutionDaoProvider.get().save(dbClass);
+
+    final InstitutionEmailDomainDao domainDao = institutionEmailDomainDaoProvider.get();
+    domainDao.deleteAllByInstitution(dbClass);
+    final List<String> domains = modelClass.getEmailDomains();
```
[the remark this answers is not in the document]
jmthibault79: This is the Swagger model, which only gives Lists

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    if (domains != null) {
+      domainDao.save(
+          domains.stream()
+              .map(domain -> new DbInstitutionEmailDomain(dbClass, domain))
+              .collect(Collectors.toSet()));
+    }
+
+    final InstitutionEmailAddressDao addrDao = institutionEmailAddressDaoProvider.get();
+    addrDao.deleteAllByInstitution(dbClass);
+    final List<String> addrs = modelClass.getEmailAddresses();
```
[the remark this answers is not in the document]
jmthibault79: cool, I like this pattern

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: shortName works for me

api/src/test/java/org/pmiops/workbench/institution/InstitutionServiceTest.java
jmthibault79: The short name is the key for API operations.  "UA2" is just going to have to be unhappy.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+  private DbInstitution getDbInstitution(String id) {
+    final DbInstitution institution = institutionDaoProvider.get().findOneByApiId(id);
+    if (institution == null) {
+      throw new NotFoundException(String.format("Could not find Institution with ID %s", id));
+    }
+    return institution;
+  }
+
+  private DbInstitution saveInstitution(final Institution modelClass, final DbInstitution dbClass) {
+    dbClass.setApiId(modelClass.getId());
```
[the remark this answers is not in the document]
jmthibault79: I'm not really seeing a good way to separate these.

api/src/main/resources/workbench.yaml
```diff
@@ -4172,3 +4259,45 @@ definitions:
...
+      id:
+        type: string
+      longName:
+        type: string
+      organizationTypeEnum:
+        $ref: '#/definitions/OrganizationType'
+        description: The Organization Type of this institution, if it is one of the enumerated OrganizationTypes
+      organizationTypeOtherText:
+        type: string
+        description: The Organization Type of this institution, as text, if it is not one of the enumerated OrganizationTypes
```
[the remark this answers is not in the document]
jmthibault79: updated text

api/src/main/java/org/pmiops/workbench/db/model/DbStorageEnums.java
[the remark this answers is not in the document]
jmthibault79: Can we apply that technique to Swagger enums?

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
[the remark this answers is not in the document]
jmthibault79: Specifically, attempting to collect everything together to save all at once results in:
`TransientObjectException: object references an unsaved transient instance - save the transient instance beforeQuery flushing: org.pmiops.workbench.db.model.DbInstitution;`

api/src/main/java/org/pmiops/workbench/db/model/DbStorageEnums.java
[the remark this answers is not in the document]
jaycarlton: Yeah, your Java enum has two fields on it, one for the API enum and one for the Short DB enum.

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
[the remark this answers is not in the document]
jaycarlton: I guess if users have a direct pointer to their institution, that's all we need. We can look it up from a log with just username or whatever if the need arises.

api/src/test/java/org/pmiops/workbench/institution/InstitutionServiceTest.java
jaycarlton: So ordinarily I would most likely just use a numeric ID and not document that it's a database ID. When you create a new institution, we can put that in the object we return. Is short text ID called out in the UI somewhere?

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
```diff
@@ -0,0 +1,16 @@
...
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailAddress;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailAddressDao
+    extends CrudRepository<DbInstitutionEmailAddress, Long> {
+
+  List<DbInstitutionEmailAddress> findAllByInstitution(DbInstitution institution);
+
+  default void deleteAllByInstitution(DbInstitution institution) {
```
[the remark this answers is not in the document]
jaycarlton: So does this operation leave users with those email addresses in a bad state? Do/should they still have institutions or not?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jaycarlton: +1 to explicitly differentiating it, though I still think it's redundant to put the type in the column name. Maybe name the enum `organization_type` and the text field `other_organization_type_description`?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailDomain.java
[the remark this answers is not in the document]
jaycarlton: I was confused by the JoinColumn. This looks good.

api/src/test/java/org/pmiops/workbench/institution/InstitutionServiceTest.java
jmthibault79: The short text ID is in our design: https://docs.google.com/document/d/1--wtRB14DK1qAMiz1mlEx12nT4HoUfYRjNkwQJwsycA/edit#

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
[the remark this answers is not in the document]
jmthibault79: Existing users will never be affected by any changes to these tables.  The Launch MVP version of Verification is strictly a one-time-only event at registration.  The Institution could go away entirely and the user would still be enabled.

Obviously there's a whole world of verification *maintenance* that is not addressed here.  We are nowhere close to having an agreed-upon set of requirements for those stories.  It may not even be the RW that does this.

jmthibault79:
Closing this draft to give the eventual merge-ready PR a clean slate for comments

PR #3078, opened 2020-02-03
[risk=no] Reduce api-local-test memory limit

calbach:
I'm not sure if we're just using more virtual memory now, or if our machine size was somehow reduced with the CircleCI plan change. 

See inline comment for explanation. Also https://circleci.com/docs/2.0/java-oom/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,7 +68,7 @@ jobs:
           - MYSQL_PASSWORD=ubuntu
     working_directory: ~/workbench
     environment:
-      JAVA_TOOL_OPTIONS: -Xmx2g
+      JAVA_TOOL_OPTIONS: -Xmx1536m
       TERM: dumb
       MYSQL_ROOT_PASSWORD: ubuntu
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,13 @@ jobs:
           - MYSQL_PASSWORD=ubuntu
     working_directory: ~/workbench
     environment:
+      # Drop to ~1.5G; the local API server starts two Java processes (the API
+      # server and the cron emulator). The default machine has 4GB, so we want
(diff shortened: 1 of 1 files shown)


calbach:
My other PR is failing 100% from api-local-test for some reason: https://github.com/all-of-us/workbench/pull/3072. When I patched this commit over there, it passed: https://circleci.com/gh/all-of-us/workbench/67364

.circleci/config.yml
```diff
@@ -68,7 +68,14 @@ jobs:
...
     environment:
-      JAVA_TOOL_OPTIONS: -Xmx2g
+      # Drop to ~1.5G; the local API server starts two Java processes (the API
+      # server and the cron emulator). The default machine has 4GB, so we want
+      # the Java memory limit to be below half of that: other things on the
+      # machine need memory as well. Using 2G here manifested in 137 errors.
+      # If this limit proves to be too low, an alternative would be to use a
+      # larger resource class:
+      #   https://circleci.com/docs/2.0/configuration-reference/#resource_class
+      JAVA_TOOL_OPTIONS: -Xmx1536m
```
jaycarlton: SGTM. I wonder if we could write a test suite that allocates more and more memory so we can figure out what the practical limit is.

calbach:
I think an ideal state would be:
- Cron emulator does start (maybe with lower memory - though I'm not sure it should fundamentally need less than the API server)
- Emulator is given a no-op cron.yaml so it doesn't actually run billing buffer creation, for example.
- If emulator crashes on startup, it fails Circle

.circleci/config.yml
calbach: This doesn't require a plan upgrade, we can use these larger machines today if we wanted. It just uses more credits / cost more if we use a larger machine. Given what our server is doing, even 1G seems like it should be fine, but we can increase if there's an actual need to do so.

PR #3088, opened 2020-02-04
[RW-3881][risk=no] Fix small typo in DbUserDataUseAgreement annotation.

gjuggler:
I was taking a look at RW-3881 (API failure when updating a user profile), and this was the first error I ran into. There may be more reasons it's failing, but this one looks like a super simple fix.

I decided to update the entity annotation rather than write a Liquibase migration to rename the column, to avoid adding a Liquibase change that might affect our rollback-ability in next week's release.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbUserDataUseAgreement.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbUserDataUseAgreement.java
@@ -23,7 +23,7 @@ public class DbUserDataUseAgreement {
 
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
-  // TODO: rename this column with a Liquibase change once we're post-beta launch!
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/db/model/DbUserDataUseAgreement.java
```diff
@@ -23,7 +23,8 @@
 
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
-  @Column(name = "user_data_use_agreement_id")
+  // TODO: rename this column with a Liquibase change once we're post-beta launch!
```
als364: Can you create a ticket for this if it doesn't already exist and link to it in the TODO?

gjuggler: Good idea – done

PR #3090, opened 2020-02-05
[RW-4359][risk=no] Not sure if this is actually a good idea: Deprecate by renaming

jmthibault79:
Description:

---
**PR checklist**

### DIFF
--- a/api/src/main/resources/workbench.yaml
+++ b/api/src/main/resources/workbench.yaml
@@ -3629,10 +3629,11 @@ definitions:
         type: string
       areaOfResearch:
         type: string
-      institutionalAffiliations:
+      deprecatedInstitutionalAffiliations:
         type: "array"
         items:
-          $ref: "#/definitions/InstitutionalAffiliation"
+          $ref: "#/definitions/DeprecatedInstitutionalAffiliation"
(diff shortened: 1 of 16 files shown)


jmthibault79:
This may be a sledgehammer: I'm also trying (elsewhere) the route of thorough commenting with no functional code changes.

api/src/main/resources/workbench.yaml
```diff
@@ -3629,10 +3629,11 @@ definitions:
         type: string
       areaOfResearch:
         type: string
-      institutionalAffiliations:
+      deprecatedInstitutionalAffiliations:
```
calbach: If we do a process like this, we shouldn't change wire format (model names are fine, but not property names like this)

jaycarlton: A couple of things can be marked deprecated on the swagger side as well https://swagger.io/docs/specification/describing-parameters/

I'm kinda thinking the path of least resistance is to come up with a different, shorter name for the new stuff,  and then just kill this one all at once. Maybe just Affiliation for the object, etc.

jmthibault79:
I don't think this will be necessary.

PR #3091, opened 2020-02-05
[noTicket][RISK=NO]Send user demographic survey attribute Gender identity to RDR

NehaBroad:
**PR checklist**

- [X ] I have run and tested this change locally
- [ X] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportEnums.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportEnums.java
@@ -72,8 +72,8 @@ public class RdrExportEnums {
   private static final BiMap<Disability, org.pmiops.workbench.rdr.model.Disability>
       CLIENT_TO_RDR_DISABILITY =
           ImmutableBiMap.<Disability, org.pmiops.workbench.rdr.model.Disability>builder()
-              .put(Disability.TRUE, org.pmiops.workbench.rdr.model.Disability.TRUE)
-              .put(Disability.FALSE, org.pmiops.workbench.rdr.model.Disability.FALSE)
+              .put(Disability.TRUE, org.pmiops.workbench.rdr.model.Disability.YES)
+              .put(Disability.FALSE, org.pmiops.workbench.rdr.model.Disability.NO)
               .build();
 
   public static org.pmiops.workbench.rdr.model.Race raceToRdrRace(Race race) {
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportEnums.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportEnums.java
@@ -30,11 +30,11 @@ public class RdrExportEnums {
   private static final BiMap<GenderIdentity, org.pmiops.workbench.rdr.model.Gender>
       CLIENT_TO_RDR_GENDER =
           ImmutableBiMap.<GenderIdentity, org.pmiops.workbench.rdr.model.Gender>builder()
-              .put(GenderIdentity.MAN, org.pmiops.workbench.rdr.model.Gender.MALE)
-              .put(GenderIdentity.WOMAN, org.pmiops.workbench.rdr.model.Gender.FEMALE)
+              .put(GenderIdentity.MAN, org.pmiops.workbench.rdr.model.Gender.MAN)
+              .put(GenderIdentity.WOMAN, org.pmiops.workbench.rdr.model.Gender.WOMAN)
               .put(GenderIdentity.NON_BINARY, org.pmiops.workbench.rdr.model.Gender.NON_BINARY)
               .put(GenderIdentity.TRANSGENDER, org.pmiops.workbench.rdr.model.Gender.TRANSGENDER)
-              .put(GenderIdentity.NONE_DESCRIBE_ME, org.pmiops.workbench.rdr.model.Gender.NONE)
+              .put(GenderIdentity.NONE_DESCRIBE_ME, org.pmiops.workbench.rdr.model.Gender.NONE_DESCRIBE_ME)
               .put(
                   GenderIdentity.PREFER_NO_ANSWER,
                   org.pmiops.workbench.rdr.model.Gender.PREFER_NOT_TO_ANSWER)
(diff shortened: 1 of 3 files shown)


api/src/main/resources/rdr.yaml
```diff
@@ -234,12 +234,11 @@ definitions:
...
     enum:
-      - MALE
-      - FEMALE
+      - MAN
+      - WOMAN
       - NON_BINARY
       - TRANSGENDER
-      - INTERSEX
-      - NONE
+      - NONE_DESCRIBE_ME
```
jaycarlton: Yeah, I like this better. It'll be cleaner definitely.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -200,6 +200,12 @@ private RdrResearcher toRdrResearcher(DbUser dbUser) {
                 .map(RdrExportEnums::sexAtBirthToRdrSexAtBirth)
                 .collect(Collectors.toList()));
       }
+      if (null != dbDemographicSurvey.getGenderIdentityEnumList()) {
+        researcher.setGender(
+            dbDemographicSurvey.getGenderIdentityEnumList().stream()
+                .map(RdrExportEnums::genderToRdrGender)
+                .collect(Collectors.toList()));
+      }
```
jaycarlton: nit: It takes a bit of getting used to, but I think it's cleaner to do
```
Optional.ofNullable(dbDemographicSurvey.getGenderIdentityEnumList())
    .map(g -> g.stream().map(RdrExportEnums::genderToRdrGender).collect(Collectors.toList()))
    .ifPresent(researcher::setGender);
```

`if`, `null`, and `!=` should generally be minimized IMO.

NehaBroad: @jaycarlton : Do you mind taking another look?

api/src/main/resources/rdr.yaml
```diff
@@ -281,10 +279,10 @@ definitions:
...
       - MASTER
       - DOCTORATE
-      -
+
   Disability:
     type: string
     enum:
-      - YES
-      - NO
+      - "YES"
```
NehaBroad: There is a bug in Swagger that converts ENUM YES and NO to true and false unless its Explicitly used as string "YES"

api/src/main/java/org/pmiops/workbench/rdr/RdrExportEnums.java
```diff
@@ -70,8 +72,8 @@
   private static final BiMap<Disability, org.pmiops.workbench.rdr.model.Disability>
       CLIENT_TO_RDR_DISABILITY =
           ImmutableBiMap.<Disability, org.pmiops.workbench.rdr.model.Disability>builder()
-              .put(Disability.TRUE, org.pmiops.workbench.rdr.model.Disability.TRUE)
-              .put(Disability.FALSE, org.pmiops.workbench.rdr.model.Disability.FALSE)
+              .put(Disability.TRUE, org.pmiops.workbench.rdr.model.Disability.YES)
```
jaycarlton: This confuses me. Either we have a boolean, in which case we don't need an enum, or we have an enum that describes  the situation like `Disability.NOT_DISABLED`, `Disability.SOME`, etc. But if you want to start an enum, it's hard to imagine what you'd add to yes and no.

NehaBroad: We basically have three options for DISABILITY FOR RDR its YES, NO ,PREFER_NOT_TO_ANSWER

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -194,26 +195,31 @@ private RdrResearcher toRdrResearcher(DbUser dbUser) {
...
-          dbDemographicSurvey.getRaceEnum().stream()
-              .map(RdrExportEnums::raceToRdrRace)
-              .collect(Collectors.toList()));
-    } else {
-      researcher.setRace(new ArrayList<>());
+          Optional.ofNullable(
+                  dbDemographicSurvey.getRaceEnum().stream()
+                      .map(RdrExportEnums::raceToRdrRace)
+                      .collect(Collectors.toList()))
+              .orElse(new ArrayList<Race>()));
```
jaycarlton: Do you want the race field to be written to, or should the set of entries be fixed?

Clients might do `researcher.getRace().add(Race.DAYTONA_500);`, which works, but is counterintuitive.

api/src/main/resources/rdr.yaml
jaycarlton: 😠 

Can you please double-check that when this goes through our script to create `merged.yaml` that the quotes are still present? I'm worried they won't be.

In fact, I'm confused Swagger can distinguish the difference; by the time the YAML is loaded all bets are off w/r/t quotes.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -194,26 +195,31 @@ private RdrResearcher toRdrResearcher(DbUser dbUser) {
...
-            dbDemographicSurvey.getSexAtBirthEnum().stream()
-                .map(RdrExportEnums::sexAtBirthToRdrSexAtBirth)
-                .collect(Collectors.toList()));
-      }
+      researcher.setSexAtBirth(
+          Optional.ofNullable(
+                  dbDemographicSurvey.getSexAtBirthEnum().stream()
+                      .map(RdrExportEnums::sexAtBirthToRdrSexAtBirth)
+                      .collect(Collectors.toList()))
+              .orElse(new ArrayList<SexAtBirth>()));
```
jaycarlton: Looks good. Passing in `.ifPresent(researcher::setSexAtBirth)` is more a matter of taste.

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
NehaBroad: no RDR needs to simple take whatever value we give to them and not manipulate them

api/src/main/resources/rdr.yaml
NehaBroad: Since this is in RDR.yaml, merged.yaml doesnt have these changes

PR #3098, opened 2020-02-07
[RW-4404][risk=low] Bump Jupyter and Welder images

calbach:
Will manually test this tomorrow. Need to pickup a bug fix in Jupyter. Welder also upgraded since it's a month old.


calbach:
Awesome - thanks for doing all the work here :)

PR #3099, opened 2020-02-07
[RW-3685][risk=no] Add some error handling for user registration and profile editing

gjuggler:
Small PR which I'm treating as part of the "get user registration ready for prime-time" ticket RW-3685.

This changes the profile page and user-registration page to do something more than console.log when an Ajax error occurs. In the long term we should aim to align with the overall Workbench user-facing error display story (see [this doc](https://docs.google.com/presentation/d/1vVK0s9AxXKyhcgZJqWAWrqiitsG-RTwh16DOTp47zEY/edit#slide=id.p) and [this ticket](https://precisionmedicineinitiative.atlassian.net/browse/RW-2820)).

Screenshots from local testing:

Screen Shot 2020-02-05 at 10 39 30 AM

Screen Shot 2020-02-05 at 10 48 53 AM

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/error-handler.tsx
+++ b/ui/src/app/components/error-handler.tsx
@@ -127,9 +127,11 @@ export const ErrorHandler = fp.flow(withUserProfile(), withGlobalError())(class 
   render() {
     const {globalError} = this.props;
     const {serverDownStatus: {apiDown, firecloudDown, notebooksDown}} = this.state;
-
+    console.log('Error handler render', this.props);
     return globalError !== undefined && <React.Fragment>
-      {globalError.statusCode === 500 && <div style={styles.errorHandler}>
+      {/* TODO(RW-RW-4392): this component should be revamped to be more generic, allowing display
+      of arbitrary error messages as well as response code based messaging. */}
+      {[400, 500].includes(globalError.statusCode) && <div style={styles.errorHandler}>
         <FlexRow style={styles.errorContent}>
           <FlexColumn>
           Unexpected Error
--- a/ui/src/app/components/error-handler.tsx
+++ b/ui/src/app/components/error-handler.tsx
@@ -127,7 +127,6 @@ export const ErrorHandler = fp.flow(withUserProfile(), withGlobalError())(class 
   render() {
     const {globalError} = this.props;
     const {serverDownStatus: {apiDown, firecloudDown, notebooksDown}} = this.state;
-    console.log('Error handler render', this.props);
     return globalError !== undefined && <React.Fragment>
       {/* TODO(RW-RW-4392): this component should be revamped to be more generic, allowing display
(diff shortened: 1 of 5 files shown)


ui/src/app/components/error-handler.tsx
```diff
@@ -127,9 +127,10 @@ export const ErrorHandler = fp.flow(withUserProfile(), withGlobalError())(class
   render() {
     const {globalError} = this.props;
     const {serverDownStatus: {apiDown, firecloudDown, notebooksDown}} = this.state;
-
     return globalError !== undefined && <React.Fragment>
-      {globalError.statusCode === 500 && <div style={styles.errorHandler}>
+      {/* TODO(RW-RW-4392): this component should be revamped to be more generic, allowing display
+      of arbitrary error messages as well as response code based messaging. */}
+      {[400, 500].includes(globalError.statusCode) && <div style={styles.errorHandler}>
```
blrubenstein: Why are we adding 400s here? I think we would want BadRequests to be handled differently?

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -97,17 +99,25 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
...
-        console.log(error);
-        this.setState({creatingAccount: false});
-      });
+    try {
+      await profileApi().createAccount({profile: this.profileObj, invitationKey: invitationKey})
+        .then((savedProfile) => {
+          this.setState({profile: savedProfile, creatingAccount: false});
+          setProfile(savedProfile, {stepName: 'accountCreationSuccess', backgroundImages: signedOutImages.login});
+        });
+    } catch (e) {
```
blrubenstein: Can we use this here instead of rehandling the exception separately?
https://github.com/all-of-us/workbench/blob/master/ui/src/app/utils/retry.tsx#L34

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -99,7 +100,11 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
       await profileApi().updateProfile(this.state.profileEdits);
       await reload();
     } catch (e) {
-      console.error(e);
+      if (e instanceof Response) {
```
blrubenstein: Similarly, can we use https://github.com/all-of-us/workbench/blob/master/ui/src/app/utils/retry.tsx#L34?

gjuggler:
Thanks for the quick review – after some thought, I'm going to abandon this PR. With enough QA, error display for these specific failed requests shouldn't be too impactful. And there's a larger story around following up on our client-side error messaging infrastructure, which this wasn't really making a dent at anyway.

PR #3103, opened 2020-02-07
[RW-3790][RW-3587][Risk=low] Remove named params and move to bigrquery

blrubenstein:
Description:
This adds cdr env variable support, moves us to bigrquery, and removes the named parameters section. This has been tested locally with cohorts with both list variables, variables with single values, and queries with no named parameters, in both R and Python

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -279,7 +279,7 @@ public class DataSetController implements DataSetApiDelegate {
     // TODO(jaycarlton): return better error information form this function for common validation
     // scenarios
     final Map<String, QueryJobConfiguration> bigQueryJobConfigsByDomain =
-        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest);
+        dataSetService.domainToBigQueryConfig(dataSetRequest);
 
     if (bigQueryJobConfigsByDomain.isEmpty()) {
       log.warning("Empty query map generated for this DataSetRequest");
@@ -323,15 +323,9 @@ public class DataSetController implements DataSetApiDelegate {
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
     DataSetPreviewResponse previewQueryResponse = new DataSetPreviewResponse();
     DataSetRequest dataSetRequest = generateDataSetRequestFromPreviewRequest(dataSetPreviewRequest);
+    // Generate a query for the preview.
     Map<String, QueryJobConfiguration> bigQueryJobConfig =
-        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest).entrySet()
-            .stream()
-            .collect(
-                Collectors.toMap(
-                    Map.Entry::getKey,
-                    stringQueryJobConfigurationEntry ->
-                        bigQueryService.filterBigQueryConfig(
-                            stringQueryJobConfigurationEntry.getValue())));
+        dataSetService.domainToBigQueryConfig(dataSetRequest);
 
     if (bigQueryJobConfig.size() > 1) {
       throw new BadRequestException(
@@ -500,7 +494,7 @@ public class DataSetController implements DataSetApiDelegate {
     }
 
     Map<String, QueryJobConfiguration> queriesByDomain =
-        dataSetService.generateQueryJobConfigurationsByDomainName(
+        dataSetService.domainToBigQueryConfig(
             dataSetExportRequest.getDataSetRequest());
 
     String qualifier = generateRandomEightCharacterQualifier();
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -494,8 +494,7 @@ public class DataSetController implements DataSetApiDelegate {
     }
 
     Map<String, QueryJobConfiguration> queriesByDomain =
-        dataSetService.domainToBigQueryConfig(
-            dataSetExportRequest.getDataSetRequest());
+        dataSetService.domainToBigQueryConfig(dataSetExportRequest.getDataSetRequest());
 
     String qualifier = generateRandomEightCharacterQualifier();
 
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -324,7 +324,14 @@ public class DataSetController implements DataSetApiDelegate {
     DataSetPreviewResponse previewQueryResponse = new DataSetPreviewResponse();
     DataSetRequest dataSetRequest = generateDataSetRequestFromPreviewRequest(dataSetPreviewRequest);
     Map<String, QueryJobConfiguration> bigQueryJobConfig =
-        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest);
+        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest).entrySet()
+            .stream()
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    stringQueryJobConfigurationEntry ->
+                        bigQueryService.filterBigQueryConfig(
+                            stringQueryJobConfigurationEntry.getValue())));
 
     if (bigQueryJobConfig.size() > 1) {
       throw new BadRequestException(
--- a/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
@@ -23,6 +23,7 @@ import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import javax.transaction.Transactional;
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
 import org.pmiops.workbench.api.BigQueryService;
 import org.pmiops.workbench.cdr.ConceptBigQueryService;
 import org.pmiops.workbench.cohortbuilder.CohortQueryBuilder;
@@ -53,6 +54,9 @@ import org.springframework.stereotype.Service;
 
 @Service
 public class DataSetServiceImpl implements DataSetService, GaugeDataCollector {
+  private static final String CDR_STRING = "\\$\\{projectId}.\\$\\{dataSetId}";
+  private static final String PYTHON_CDR_ENV_VARIABLE = "\"\"\" + os.environ[\"WORKSPACE_CDR\"] + \"\"\"";
+  private static final String R_CDR_ENV_VARIABLE = "\", Sys.getenv(\"WORKSPACE_CDR\"), \"";
 
   private static final String SELECT_ALL_FROM_DS_LINKING_WHERE_DOMAIN_MATCHES_LIST =
       "SELECT * FROM `${projectId}.${dataSetId}.ds_linking` "
@@ -503,8 +507,9 @@ public class DataSetServiceImpl implements DataSetService, GaugeDataCollector {
         .build();
   }
 
-  private QueryJobConfiguration buildQueryJobConfigurationWithProjectInformation(
+  private QueryJobConfiguration buildAndFilterQueryJobConfiguration(
       Map<String, QueryParameterValue> namedCohortParameters, String query) {
+    // This runs filterBigQueryConfig as well, to make a runnable query.
     return bigQueryService.filterBigQueryConfig(
         buildQueryJobConfiguration(namedCohortParameters, query));
   }
@@ -636,7 +641,7 @@ public class DataSetServiceImpl implements DataSetService, GaugeDataCollector {
 
     final TableResult valuesLinkingTableResult =
         bigQueryService.executeQuery(
-            buildQueryJobConfigurationWithProjectInformation(
+            buildAndFilterQueryJobConfiguration(
                 queryParameterValuesByDomain,
                 SELECT_ALL_FROM_DS_LINKING_WHERE_DOMAIN_MATCHES_LIST));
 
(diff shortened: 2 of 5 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
+                startingIndex,
+                startingIndex + stringToReplace.length(),
+                "("
+                    .concat(
+                        nullableListToEmpty(value.getArrayValues()).stream()
+                            .map(QueryParameterValue::getValue)
+                            .collect(Collectors.joining(", ")))
+                    .concat(")"));
+          } else {
+            String stringToReplace = "@".concat(key);
```
calbach: Is this going to handle strings vs numbers properly? Wouldn't strings need to be quoted?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
     return StringUtils.capitalize(text.toLowerCase());
   }
 
+  private static String generateSqlWithEnvironmentVariables(
+      String query, KernelTypeEnum kernelTypeEnum) {
+    switch (kernelTypeEnum) {
+      case PYTHON:
+        return query.replaceAll(
+            "\\$\\{projectId}.\\$\\{dataSetId}", "\"\"\" + os.environ[\"WORKSPACE_CDR\"] + \"\"\"");
```
calbach: Did you look at str_glue as Nicole suggested? I think it would be more readable to declare variables then substitute them into the templated SQL as is done in the snippets. Note: it would require setting up an additional package

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
+        return query;
+    }
+  }
+
+  private static String fillInQueryParams(
+      String query, Map<String, QueryParameterValue> queryParameterValueMap) {
+    StringBuilder stringBuilder = new StringBuilder(query);
+    queryParameterValueMap.forEach(
+        (key, value) -> {
+          if (StandardSQLTypeName.ARRAY.equals(value.getType())) {
```
calbach: I don't love inlining these into the SQL itself. I think string substitution within the notebook code would be more reusable / understandable / readable. In R: str_glue, in Py `f""`

Alternatively, the parameterized queries in Python were perhaps an ideal way to show how to properly use BigQuery, but understood that diverging this much between R and Python could be a problem.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -324,7 +324,14 @@ private DataSetRequest generateDataSetRequestFromPreviewRequest(
...
     DataSetRequest dataSetRequest = generateDataSetRequestFromPreviewRequest(dataSetPreviewRequest);
     Map<String, QueryJobConfiguration> bigQueryJobConfig =
-        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest);
+        dataSetService.generateQueryJobConfigurationsByDomainName(dataSetRequest).entrySet()
+            .stream()
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    stringQueryJobConfigurationEntry ->
+                        bigQueryService.filterBigQueryConfig(
```
calbach: This deserves a comment, it seems like you're trying to replace the `{project}.{dataset}` templates within the query parameters at this step. But I'm not very clear on why

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -692,7 +692,7 @@ public void testGetPythonQuery() {
     String prefix = "dataset_00000000_condition_";
     assertThat(response.getCode())
         .isEqualTo(
-            "import pandas\n\n"
+            "import pandas\nimport os\n\n"
```
calbach: There should probably be at least one test which incorporates parameters

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
   }
 
+  private static String generateSqlWithEnvironmentVariables(
+      String query, KernelTypeEnum kernelTypeEnum) {
+    switch (kernelTypeEnum) {
+      case PYTHON:
+        return query.replaceAll(
+            "\\$\\{projectId}.\\$\\{dataSetId}", "\"\"\" + os.environ[\"WORKSPACE_CDR\"] + \"\"\"");
+      case R:
+        return query.replaceAll(
```
jaycarlton: nit: I generally don't want to see three languages in one function.Can we wrap the bodies of these case clauses into their own functions?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
+  }
+
+  private static String fillInQueryParams(
+      String query, Map<String, QueryParameterValue> queryParameterValueMap) {
+    StringBuilder stringBuilder = new StringBuilder(query);
+    queryParameterValueMap.forEach(
+        (key, value) -> {
+          if (StandardSQLTypeName.ARRAY.equals(value.getType())) {
+            String stringToReplace = "unnest(@".concat(key.concat(")"));
+            int startingIndex = stringBuilder.indexOf(stringToReplace);
```
jaycarlton: Too much is happening here to test effectively or to really understand. Please do put comments at the top of this function and split up the string operations.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
+            stringBuilder.replace(
+                startingIndex,
+                startingIndex + stringToReplace.length(),
+                "("
```
jaycarlton: super-nit; Starting with the string literal and calling concat isn't really idiomatic in Java.

I'd make a helper function like  `formatArrayValueStringValuesOrWhatever(QueryParameterValue queryParameterValue)` that gives . you everything you're concatenating here.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -664,14 +664,58 @@ public static String capitalizeFirstCharacterOnly(String text) {
...
+                            .map(QueryParameterValue::getValue)
+                            .collect(Collectors.joining(", ")))
+                    .concat(")"));
+          } else {
+            String stringToReplace = "@".concat(key);
+            int startingIndex = stringBuilder.indexOf(stringToReplace);
+            stringBuilder.replace(
+                startingIndex,
+                startingIndex + stringToReplace.length(),
+                Optional.ofNullable(value.getValue()).orElse(""));
```
jaycarlton: nit: I think this reads better with `Optional.ofNullable(...).ifPresent(v -> stringBuilder.replace(..., v)`

But also, I'm not used to seeing replace used on a stringBuilder. It's generally expected that you build up the correct string from the beginning. Is there a way to do the swapping earlier?

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -771,23 +755,12 @@ public void testGetRQuery() {
...
-                + "\n"
-                + prefix
-                + "query_config <- list(\n"
-                + "  query = list(\n"
-                + "    parameterMode = 'NAMED',\n"
-                + "    queryParameters = list(\n\n"
-                + "    )\n"
-                + "  )\n"
-                + ")\n"
+                + "LIMIT \", max_number_of_rows, sep=\"\")\n"
```
jaycarlton: We need to have test cases for all the edge cases,  like when things that are expected to be there are missing. We should be failing the operation early instead of sending a broken query to the notebook if anything is amiss. (We don't have a great automatic way to find all the notebooks with queries for a given release yet from what I remember.)

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: +1 . I get scared seeing system calls in the middle of a complex string operation. We need to have the cleanest, earliest error code we can if one of those values is missing or malformed.

Also, I'll give you a dollar for every string or regex replacement we can kill in this class. 😉 I believe we can build almost everything up in one direction, and that's much easier to reason about and test. Later PR of course.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: I think the key is always @key_name, so this should work?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Where is the str_glue comment? I didn't see that in the email I got.

I do hate the amount of string replacement we're doing, but we would need to go into the cohort builder sql builder to change it too much. (Unfortunately we're using the same SQL building code that they use.)

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: I am happy using str_glue or `f""`, but honestly, I'm unsure what benefit it gives over the config or query parameters (They're both variables that get substitutions in)

I can revert back to that if we like. BigRQuery does just have it use a parameter option instead of a configuration option, so it does trim a lot of boilerplate and just make it basically a substitution.

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
blrubenstein: I tried checking this and it appears to be nonsense, so I removed it

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Sure, will do.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Will do.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -496,11 +497,16 @@ private boolean supportsConceptSets(Domain domain) {
 
   private QueryJobConfiguration buildQueryJobConfiguration(
       Map<String, QueryParameterValue> namedCohortParameters, String query) {
+    return QueryJobConfiguration.newBuilder(query)
```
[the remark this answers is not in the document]
blrubenstein: I can do the function that returns a builder with the arg set, that seems reasonable to me. Because its in two places, I would prefer having the default args change in both with one change.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
calbach: Sorry I'm not sure if she explicitly suggested this, I was more thinking you were basing the approach on the code she had already written for the snippets menu. Here's an example of str_glue: https://github.com/all-of-us/workbench-snippets/blob/f6e714328ba8c0d3ff11387944fccb0df5ba7ab5/sql-snippets/snippets_setup.R#L30 ; see more complicated variants by inserting from the snippets SQL menu. The relevant thing here is that it lets you interpolate variables into a string.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
calbach: The benefit is that it can work with arrays in R, otherwise I'd prefer config/query parameters.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
calbach: Sorry, I should have highlighted the code block. I meant to refer to the value with this comment, not the key, i.e. line 704

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: Not without changing the Cohort Builder SQL building code, unfortunately. Will change to the Optional.ofNullable method though.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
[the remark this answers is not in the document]
blrubenstein: Talked offline: 
I tried to address the comment here https://github.com/all-of-us/workbench/pull/3103/files#diff-5324f051d4e486d6f33e30b9f6bebd92R500 but it looks like all the builder methods that don't require a query up front are private methods. :cry:
I could have it build each time it is used, but that ends up with a bunch of repeated boilerplate for the defaults.
11:20
(I could maybe build with an empty string literal and then set the query later?) (edited) 
11:20
Do you have a preference?

Jay Carlton  11:26 AM
it's not a huge win to do what I think I was saying





11:26
just pointing out that builders smaller than ~5 lines are all right inline unless there's lots of repeated fields

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
blrubenstein: I will take a look at that.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
```diff
@@ -533,16 +544,10 @@ private String getQualifiedColumnName(Domain currentDomain, String columnName) {
...
     switch (kernelTypeEnum) {
       case R:
-        prerequisites =
-            // RW-4241 workaround: update when the Jupyter image with the reticulate fix is fully
-            // rolled out
-            "require(devtools)\n"
-                + "devtools::install_github(\"rstudio/reticulate\", ref=\"00172079\")\n"
-                + "library(reticulate)\n"
-                + "pd <- reticulate::import(\"pandas\")";
+        prerequisites = "library(bigrquery)";
```
jmthibault79: Do we need to upgrade our notebook docker image to include bigrquery?

jaycarlton: I think the ticket said we don't have to in this case... Don't quote me on that.

blrubenstein: It functions without it, so we should be good. My guess is Leo includes it by default?

jmthibault79: Great.  I'll add it to my sanity-checking notebook.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
jaycarlton: OK, I really want to get to tinkering on a new third party query builder.

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -716,27 +728,16 @@ public void testGetPythonQuery() {
                 + "condition_source_concept_id IN (123)) \n"
                 + "AND (c_occurrence.PERSON_ID IN (SELECT * FROM person_id from `"
                 + TEST_CDR_TABLE
-                + ".person` person)) "
+                + ".person` person WHERE "
+                + NAMED_PARAMETER_VALUE.getValue()
```
calbach: If I'm reading this correctly, this turns into: `WHERE ICD9 IN (2, 5)`. This doesn't seem right

calbach:
Per offline discussion, inlining of variables via str_glue / `f""` would be ideal. Should probably really happen before launch. But I think this is fine as a first-pass to get us onto bigrquery, we'll see if Karthik has feedback on this

api/src/main/java/org/pmiops/workbench/db/dao/DataSetServiceImpl.java
calbach: What was the outcome here?

api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
blrubenstein: In this case, I just threw some random text in. In truth, it is more likely to be something like `WHERE concept_id IN (132543, 2135643...)`

calbach: Could you change it to be a more realistic value? I think future editors will find this confusing

PR #3106, opened 2020-02-07
[RW-4408][risk=no] drop firecloud-tools and its one user

calbach:
I've updated step 35 of the new app guide accordingly: https://broad.io/aou-new-environment


jaycarlton:
👏

PR #3107, opened 2020-02-08
[RW-3554][risk=low] Shore up inactivity detection logic

calbach:
Test this out @ https://calbach-dot-all-of-us-workbench-test.appspot.com/
-> currently configured for 30s modal, 60s timeout.

Two additions in combination should cover the known issues with setTimeout:
- Before updating the last active timestamp - check that they haven't already surpassed the time limit. This covers the case where a user sleeps their machine, comes back, and resumes activity immediately.
- Every 1m, reset the `setTimeouts()` to account for the actual remaining time according to LocalStorage. This covers the case where a user wakes up their machine and then doesn't interact with the tab again (they could still view the content).

Manually verified by sleeping my laptop and reopening the site after 1m.

### DIFF
--- a/ui/src/app/pages/signed-in/component.ts
+++ b/ui/src/app/pages/signed-in/component.ts
@@ -7,7 +7,7 @@ import {ServerConfigService} from 'app/services/server-config.service';
 import {SignInService} from 'app/services/sign-in.service';
 import {cdrVersionsApi} from 'app/services/swagger-fetch-clients';
 
-import {debouncer, hasRegisteredAccessFetch, resettableTimeout} from 'app/utils';
+import {debouncer, hasRegisteredAccessFetch} from 'app/utils';
 import {cdrVersionStore, navigateSignOut, routeConfigDataStore} from 'app/utils/navigation';
 import {initializeZendeskWidget} from 'app/utils/zendesk';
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/signed-in/component.ts
```diff
@@ -93,14 +94,25 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
       this.minimizeChrome = minimizeChrome;
     }));
 
+    this.signOutIfLocalStorageInactivityElapsed();
```
ericsong: Does this get called when a sleeping machine is woken up? I think I got most of it but wondering how the check is enforced the moment the user returns to their machine.

calbach: This line was actually just a refactoring of a check which was already here - no logic changed on this line. I didn't think hard about why it was here; I think this was to catch returning users.

It won't necessarily be enforced the moment they return to their machine, it will be enforced either:
1. When they next interact with the page: L175, or
2. when the interval check resets the timers (within <1m of machine wakeup): L163

ericsong: Got it. L175 was what I was looking for.

PR #3110, opened 2020-02-10
[RW-3685][risk=low] Add a config variable controlling invitation key verification

gjuggler:
In prep for launch, this PR adds a new config variable requireVerificationKey which controls the frontend & backend components which require a verification key to be input by users. When this is off, the frontend will not collect an invitation key & the backend will not verify it.

A few notes:

**Jira ticket**
I'm attaching this to the broader "enable feature flag for registration pages" issue, since it's thematically related and this is one more step on the way towards enabling that flag.

**Environment var vs. feature flag**
I'm modeling this flag as an "access" config variable rather than a feature flag, since we may wish to control access with an invitation key in lower environments (such as preprod) even once this is opened up in prod. Feature flags should have an expectation that they will be removed at some point in the future, which doesn't fit here.

**Controller tests**
I also took the opportunity to try and quickly clean up the ProfileControllerTest a bit. This was initially motivated by a desire to more easily change the WorkbenchConfig for individual test cases. I made two major changes: (1) create a BaseControllerTest class (similar to what I'd done in https://github.com/all-of-us/workbench/blob/master/api/src/integration/java/org/pmiops/workbench/BaseIntegrationTest.java), and (2) remove explicit constructors in favor of @Autowired and @MockBean. I think this came out a little cleaner on the other side, though it does add some extra weight to the PR – apologies!

**UI tests**
On the UI side, I had a bit of "fun" trying to find a way to gain access to the inner React component from Enzyme. First off, with higher-order components and shallow rendering, you need to make a few extra calls to shallow() to ensure the inner component gets rendered. Secondly, the core component was originally an anonymous inner class, which didn't allow me to call its public methods from the Enzyme test. This was fixed by bumping the class up to the top-level and exporting it – a pattern that we may want to use elsewhere.

---
**PR checklist**

- [X] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -293,6 +293,7 @@ sourceSets {
     runtimeClasspath += generated.output
   }
   test {
+    // Allow unit tests to directly load config files.
     resources {
       srcDir "config/"
     }
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -293,6 +293,9 @@ sourceSets {
     runtimeClasspath += generated.output
   }
   test {
+    resources {
+      srcDir "config/"
+    }
     compileClasspath += generated.output
     runtimeClasspath += generated.output
   }
--- a/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
@@ -19,7 +19,7 @@ import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
 @DataJpaTest
-public class BaseControllerTest {
+public abstract class BaseControllerTest {
 
   protected static WorkbenchConfig config;
 
--- a/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
@@ -17,6 +17,22 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Scope;
 import org.springframework.test.context.junit4.SpringRunner;
 
+/**
+ * A base class for functionality shared by many controller-level tests in the Workbench.
+ *
+ * <p>This class currently provides a simple way for test cases to use and modify the
+ * WorkbenchConfig, for testing differential controller behavior depending on the environment
+ * configuration.
+ *
+ * <p>Most or all controller tests should eventually extend this class, but adoption may proceed
+ * incrementally to avoid a large one-time refactoring cost.
+ *
+ * <p>This class is heavily inspired by the BaseIntegrationTest class; maybe there is some way for
+ * these two classes to share a common core, since they both care mostly about WorkbenchConfig bean
+ * management for tests.
+ *
+ * <p>TODO(RW-4443): update all controller tests to extend this class.
+ */
 @RunWith(SpringRunner.class)
 @DataJpaTest
 public abstract class BaseControllerTest {
--- a/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
@@ -1,25 +1,25 @@
-package org.pmiops.workbench;
+package org.pmiops.workbench.api;
 
 import com.google.common.io.Resources;
 import com.google.gson.Gson;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.util.Random;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.runner.RunWith;
 import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.test.FakeLongRandom;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
 import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Scope;
 import org.springframework.test.context.junit4.SpringRunner;
-import org.springframework.test.context.web.WebAppConfiguration;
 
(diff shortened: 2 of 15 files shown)


ui/src/app/pages/login/sign-in.tsx
```diff
@@ -127,127 +127,177 @@ interface SignInState {
   termsOfServiceVersion?: number;
 }
 
-export const SignInReact = fp.flow(withServerConfig(), withWindowSize())(
```
gjuggler: GitHub is kinda mangling the diff here. The key change made was to extract the inner class and turn it into a top-level component called SignInReactImpl. This allows us to refer to that class from the unit test, which this PR uses to verify that the set of sign-in steps is correct when invitation key verification is turned off.

gjuggler: The diff looks much cleaner if you go "gear menu --> hide whitespace changes", or this link: https://github.com/all-of-us/workbench/pull/3110/files?utf8=%E2%9C%93&diff=unified&w=1

api/build.gradle
```diff
@@ -293,6 +293,10 @@ sourceSets {
     runtimeClasspath += generated.output
   }
   test {
+    // Allow unit tests to directly load config files.
+    resources {
+      srcDir "config/"
```
als364: Ooh. Will this mean that we'll no longer have to create a `WorkbenchConfigProvider` per test?

api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Scope;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@DataJpaTest
+public abstract class BaseControllerTest {
```
als364: With a name like this, I'd expect _everything_ to extend it. Can you call this something that makes it clearer this class is intended for composition? Maybe something like `WithWorkbenchConfig`?

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -644,24 +660,8 @@ private void assertUser(
     assertThat(user.getFamilyName()).isEqualTo(familyName);
     assertThat(user.getGivenName()).isEqualTo(givenName);
     assertThat(user.getDataAccessLevelEnum()).isEqualTo(dataAccessLevel);
-    assertThat(user.getFirstSignInTime()).isEqualTo(firstSignInTime);
+    assertThat(user.getFirstSignInTime().getTime()).isEqualTo(firstSignInTime.getTime());
+    // assertThat(user.getFirstSignInTime()).isEqualTo(firstSignInTime);
```
als364: extraneous

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -184,16 +184,21 @@ public String freeTierBillingAccountName() {
...
     // Allows a user to bypass their own access modules. This is used for testing purposes so that
     // We can give control over 3rd party access modules
     public boolean unsafeAllowSelfBypass;
+    // These booleans control whether each of our core access modules are enabled per environment.
     public boolean enableComplianceTraining;
     public boolean enableEraCommons;
     public boolean enableDataUseAgreement;
     public boolean enableBetaAccess;
+    // Controls whether an invitation key is required for user creation. When true, the account
+    // creation UI will show an invitation key form, and the server-side will valide the key before
```
als364: validate

als364: valide -> validate

ui/src/app/pages/login/sign-in.spec.tsx
```diff
@@ -74,7 +80,7 @@ describe('SignInReact', () => {
   it('should handle sign-up flow for legacy account creation', () => {
     // To correctly shallow-render this component wrapped by two HOCs, we need to add two extra
     // .shallow() calls at the end.
```
als364: This should be moved up to the new `shallowComponent` in line 23.

ui/src/app/pages/login/sign-in.spec.tsx
```diff
@@ -97,7 +103,7 @@ describe('SignInReact', () => {
 
     // To correctly shallow-render this component wrapped by two HOCs, we need to add two extra
     // .shallow() calls at the end.
```
als364: And so should this.

ui/src/app/pages/login/sign-in.tsx
als364: it's mangled even with but at least it highlights the differences

ui/src/app/pages/login/sign-in.tsx
```diff
@@ -127,127 +127,177 @@ interface SignInState {
...
+      SignInStep.INVITATION_KEY,
+      SignInStep.TERMS_OF_SERVICE,
+      SignInStep.ACCOUNT_CREATION,
+      SignInStep.DEMOGRAPHIC_SURVEY,
+      SignInStep.SUCCESS_PAGE
+    ];
+
+    const {enableNewAccountCreation, requireInvitationKey} = this.props.serverConfig;
+
+    if (!enableNewAccountCreation) {
```
als364: nit: I'd prefer to just see separate lists returned in each of these cases. In this case, we're not going to change what `enableNewAccountCreation` or `requireInvitationKey` means.

api/src/test/java/org/pmiops/workbench/api/BaseControllerTest.java
als364: Works for me. Can you coordinate with Joel to put this on the list of best practices things?

PR #3112, opened 2020-02-11
[RW-4303][risk=no] Refactor datasets page and fix race conditions

calbach:
Key changes:
- Consolidate domain/value set loading through one pathway
  - The original trigger is setting `state.selectedDomains`
  - Subsequent loading and autoselection of domain values is done in `componentDidUpdate`
- Add a unified loading/lookup layer for domain -> valueSets
- Switched prepackaged concept set representation from `boolean` -> `Set<demograpics|surveys>`

Demo @ http://calbach-dot-all-of-us-workbench-test.appspot.com/

Also identified several areas for further simplification here.

### DIFF
--- a/ui/src/app/pages/data/data-set/dataset-page.tsx
+++ b/ui/src/app/pages/data/data-set/dataset-page.tsx
@@ -554,10 +554,10 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
         newLookup.set(domain, {domain, values});
 
         // Autoselect the newly added domain, except if we're editing an
-        // existing dataset. This avoids having us overwrite the selected pairs
-        // on initial load.
+        // existing dataset which already covers the domain. This avoids having
+        // us overwrite the selected pairs on initial load.
         let morePairs = [];
-        if (!this.getDomainsFromDataSet(dataSet).has(domain)) {
+        if (!this.editing || !this.getDomainsFromDataSet(dataSet).has(domain)) {
           morePairs = values.items.map(v => ({domain, value: v.value}));
         }
 
--- a/ui/src/app/pages/data/data-set/dataset-page.tsx
+++ b/ui/src/app/pages/data/data-set/dataset-page.tsx
@@ -474,9 +474,9 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
         return;
       }
 
-      // TODO: There is a lot of complexity here around loading domain values
-      // which is static data for a given CDR version. Consider refactoring this
-      // page to load all schema data before rendering.
+      // TODO(RW-4426): There is a lot of complexity here around loading domain
+      // values which is static data for a given CDR version. Consider
+      // refactoring this page to load all schema data before rendering.
 
       // If any of these new domains have already been loaded, autoselect all of
       // their value pairs. The desired product behavior is that when a new
--- a/ui/src/app/pages/data/data-set/dataset-page.tsx
+++ b/ui/src/app/pages/data/data-set/dataset-page.tsx
@@ -219,7 +219,8 @@ const DOMAIN_DISPLAY_ORDER = {
   [Domain.SURVEY]: 1
 };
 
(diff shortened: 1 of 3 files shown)


ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -426,47 +444,100 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+    async componentDidUpdate({}, prevState: State) {
+      // After a state update, first check whether any new domains have been added.
+      const newDomains = Array.from(this.state.selectedDomains)
+          .filter(d => !prevState.selectedDomains.has(d));
+      if (!newDomains.length) {
+        return;
+      }
+
+      // TODO: There is a lot of complexity here around loading domain values
+      // which is static data for a given CDR version. Consider refactoring this
```
blrubenstein: Is there a ticket for the refactor here? If not, we probably should file one.

NehaBroad:
I tried testing this on http://calbach-dot-all-of-us-workbench-test.appspot.com/ but i found a bug, After selecting SURVEY, if i select DEMOGRAPHIC values are not updated with PERSON domain, basically the values are not updated with whatever value we select second

NehaBroad:
Ignore the previous comment, maybe it was a timing thing, i can no longer reproduce the bug.

ui/src/app/pages/data/data-set/dataset-page.tsx
```diff
@@ -919,25 +934,27 @@ const DataSetPage = fp.flow(withUserProfile(), withCurrentWorkspace(), withUrlPa
...
+                  <div style={{height: selectedDomains.size > 0 ? '7.625rem' : '9rem', overflowY: 'auto'}}>
+                    {domainValueSetIsLoading.size > 0 && <Spinner style={{position: 'relative',
                       top: '2rem', left: 'calc(50% - 36px)'}}/>}
-                    {valueSets.map(valueSet =>
-                      <div key={valueSet.domain}>
+                    {Array.from(selectedDomains).map(domain =>
+                      domainValueSetLookup.has(domain) &&
+                      <div key={domain}>
                         <Subheader style={{fontWeight: 'bold'}}>
-                          {valueSet.survey ? 'Survey' : formatDomain(valueSet.domain)}
```
calbach: @NehaBroad @blrubenstein  I could not understand what was going on here with this `.survey` value. I didn't find any comments hinting at the rationale for this. I removed it in my PR because it was a potential source of bugs/inconsistency. If this is tracking a real requirement, let me know.

Relatedly, I'm confused by this line: https://github.com/all-of-us/workbench/blob/master/ui/src/app/pages/data/data-set/dataset-page.tsx#L1010

ui/src/app/pages/data/data-set/dataset-page.tsx
calbach: Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4426

ui/src/app/pages/data/data-set/dataset-page.tsx
NehaBroad: It might be because Survey was not initially not a DOMAIN hence we were checking if survey object is populated treat it as SURVEY

NehaBroad: Could the OBSERVATION thing be because of api/src/main/java/org/pmiops/workbench/db/dao/ConceptSetDao.java L26?

.put(Domain.SURVEY, "observation")

calbach:
All my own follow-ups have been addressed, just needs review/approval now.

calbach:
I caught one more regression my PR caused, which was around initial loading of dataset value pairs.

ui/src/app/pages/data/data-set/dataset-page.tsx
calbach: Surveys used to be in the Observation domain, I've removed the special handling of this for now, since presumably we could have a concept set in the observation domain as well, and we'd want it to be labeled accordingly.

PR #3118, opened 2020-02-11
[risk=no]Institution DB/Model updates on the road toward RW-4259

jmthibault79:
Description:

Does NOT resolve RW-4259.  This is just refactoring, partially to help make that easier.  Thanks to @freemabd who did the majority of the work.

TIL a lot:
* RW best practices are to annotate the accessors, not the fields.  This lets us add extra logic which Hibernate will use.
* LAZY vs EAGER fetch types affect functionality, not just resource usage.  (TODO: look into this more)
* Hibernate does not work well when you reassign collections it's managing.  Always use add/remove/set/clear functions on the collection in-place.
* It may be necessary to assign bidirectional links in both directions, but it is not always necessary to sever the link in both directions.
  * TODO: look into this much more.  There's a lot of subtlety here.

---
**PR checklist**


api/db/changelog/db.changelog-126-institution-foreign-key.xml
```diff
@@ -0,0 +1,32 @@
...
+  <changeSet author="brianfreeman" id="changelog-126-institution-foreign-key">
+
+    <dropForeignKeyConstraint
+      baseTableName="institution_email_domain"
+      constraintName="fk_institution_email_domain"/>
+
+    <addForeignKeyConstraint baseColumnNames="institution_id"
+      baseTableName="institution_email_domain"
+      constraintName="fk_institution_email_domain"
+      onDelete="CASCADE"
```
jmthibault79: adds delete cascade

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailAddressDao.java
```diff
@@ -1,16 +0,0 @@
-package org.pmiops.workbench.db.dao;
```
jmthibault79: These are now fully managed by DbInstitution

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,41 +12,35 @@
...
-  private Set<DbInstitutionEmailAddress> emailAddresses;
+  @NotNull private Set<DbInstitutionEmailDomain> emailDomains = Sets.newHashSet();
+  @NotNull private Set<DbInstitutionEmailAddress> emailAddresses = Sets.newHashSet();
 
   public DbInstitution() {}
 
-  public DbInstitution(final String shortName, final String displayName) {
-    setShortName(shortName);
-    setDisplayName(displayName);
+  @Id
```
jmthibault79: RW best practices are to annotate the accessors, not the fields.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -78,21 +94,69 @@ public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
     this.organizationTypeOtherText = organizationTypeOtherText;
   }
 
+  /** Builder method to help streamline the building of a DbInstitution. */
+  public DbInstitution organizationTypeOtherText(String organizationTypeOtherText) {
+    this.organizationTypeOtherText = organizationTypeOtherText;
+    return this;
+  }
+
+  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
```
jmthibault79: EAGER fetching is key to making this work correctly.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -78,21 +94,69 @@ public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
...
+   * call addAll() to add the diff(new - old) rows.
+   *
+   * <p>https://stackoverflow.com/questions/5587482/hibernate-a-collection-with-cascade-all-delete-orphan-was-no-longer-referenc
+   *
+   * @param emailDomains the new set of domains for this Institution
+   */
+  public void setEmailDomains(final Set<DbInstitutionEmailDomain> emailDomains) {
+    final Set<DbInstitutionEmailDomain> attachedDomains =
+        emailDomains.stream()
+            .map(address -> address.setInstitution(this))
```
jmthibault79: Need to make the connection in both directions, but we only need to break it in one.  It is NOT necessary to make this null when removing.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -64,74 +55,50 @@ public boolean deleteInstitution(final String shortName) {
...
 
-  private DbInstitution saveInstitution(final Institution modelClass, final DbInstitution dbClass) {
-    dbClass.setShortName(modelClass.getShortName());
-    dbClass.setDisplayName(modelClass.getDisplayName());
-    dbClass.setOrganizationTypeEnum(
-        DbStorageEnums.organizationTypeToStorage(modelClass.getOrganizationTypeEnum()));
-    dbClass.setOrganizationTypeOtherText(modelClass.getOrganizationTypeOtherText());
-
-    return saveInstitutionEmailPatterns(modelClass, institutionDao.save(dbClass));
+  private DbInstitution newDbObject(final Institution modelObject) {
```
jmthibault79: Model <-> API conversions are now fully independent of DB state!

TODO (not in this PR) move conversions out of here into a MapStruct mapper.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jaycarlton: Huh, you don't have to give it the key column names. Neat.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jaycarlton: nit: why are you calling the map variable `address` instead of `ed` or similar? I didn't think domains were addresses.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -78,21 +94,69 @@ public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
...
+  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
   public Set<DbInstitutionEmailAddress> getEmailAddresses() {
     return emailAddresses;
   }
 
-  public void setEmailAddresses(Iterable<DbInstitutionEmailAddress> emailAddresses) {
-    this.emailAddresses =
-        StreamSupport.stream(emailAddresses.spliterator(), false).collect(Collectors.toSet());
+  /**
+   * Hibernate doesn't like it when you reassign collections. Instead, modify in-place. First, call
```
jaycarlton: Can you explain at a higher level what this function is trying to accomplish? That is, why are we doing multiple set intersections? And why do we need a set difference if emailAddresses is a set to begin with? I thought adding an existing item was a no-op.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -78,21 +94,69 @@ public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
...
+    final Set<DbInstitutionEmailAddress> attachedAddresses =
+        emailAddresses.stream()
+            .map(address -> address.setInstitution(this))
+            .collect(Collectors.toSet());
+    // modifies this set so that its value is the intersection of the two sets
+    this.emailAddresses.retainAll(attachedAddresses);
+    this.emailAddresses.addAll(Sets.difference(attachedAddresses, this.emailAddresses));
+  }
+
+  /** Builder method to help streamline the building of a DbInstitution. */
```
jaycarlton: For clarity, I would make a public nested Builder class and put this there? An example is [here](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/monitoring/MeasurementBundle.java#L70). Though I guess you still need your no-arg ctor 🤔 .

Anyway, I wouldn't make an alias function (a non-overload with the same body) unless there's a compelling reason.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -15,51 +15,57 @@
...
 
   public DbInstitutionEmailAddress() {}
 
-  public DbInstitutionEmailAddress(DbInstitution institution, String emailAddress) {
-    this.institution = institution;
-    this.emailAddress = emailAddress;
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @Column(name = "institution_email_address_id")
+  public long getInstitutionEmailAddressId() {
```
jaycarlton: If this is nullable, should we return `Optional<Long>`?

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -19,34 +17,27 @@
...
-      InstitutionEmailAddressDao institutionEmailAddressDao) {
+  InstitutionServiceImpl(InstitutionDao institutionDao) {
     this.institutionDao = institutionDao;
-    this.institutionEmailDomainDao = institutionEmailDomainDao;
-    this.institutionEmailAddressDao = institutionEmailAddressDao;
   }
 
   @Override
   public List<Institution> getInstitutions() {
     return StreamSupport.stream(institutionDao.findAll().spliterator(), false)
```
jaycarlton: I'd probably put this on the dao itself as a `default List<Institution> getAllInstitutions()` so all services using it get the benefit.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -64,74 +55,50 @@ public boolean deleteInstitution(final String shortName) {
...
+        .shortName(dbObject.getShortName())
+        .displayName(dbObject.getDisplayName())
+        .organizationTypeEnum(dbObject.getOrganizationTypeEnum())
+        .organizationTypeOtherText(dbObject.getOrganizationTypeOtherText())
+        .emailDomains(
+            Optional.ofNullable(dbObject.getEmailDomains()).orElse(Collections.emptySet()).stream()
+                .map(DbInstitutionEmailDomain::getEmailDomain)
+                .collect(Collectors.toList()))
+        .emailAddresses(
+            Optional.ofNullable(dbObject.getEmailAddresses()).orElse(Collections.emptySet())
```
jaycarlton: With the db objects, we can (and I think should) avoid returning null for collections. We don't have a choice with the generated classes though.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,175 @@
...
+                .emailAddresses(emailAddresses)
+                .emailDomains(emailDomains));
   }
 
   @Test
   public void test_save() {
-    assertThat(institutionDao.findAll()).hasSize(1);
-    institutionDao.save(new DbInstitution("VUMC", "Vanderbilt"));
-    assertThat(institutionDao.findAll()).hasSize(2);
+    // no need to call save since we have a setup method.
```
jaycarlton: code under test shouldn't really be executed in setup() if we can avoid it. It's just for mocks and fixtures.  You'll still fail your tests if it blows up, but you should ideally have assertions after important steps, especially in a test_save() method.

api/src/test/java/org/pmiops/workbench/institution/InstitutionServiceTest.java
```diff
@@ -49,25 +59,32 @@ public void test_deleteInstitution() {
...
+    assertThat(service.getInstitution(testInst.getShortName())).hasValue(testInstAfterRT);
     assertThat(service.getInstitution("otherInst")).isEmpty();
 
     final Institution otherInst =
-        new Institution().shortName("otherInst").displayName("The Institution of testing");
+        new Institution()
+            .shortName("otherInst")
+            .displayName("The Institution of testing")
+            .emailAddresses(new ArrayList<>())
+            .emailDomains(new ArrayList<>());
```
jaycarlton: So if you want to use the builder paradigm, you could make a `billed()` method that fills in missing array values and other defaultable things.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
freemabd: Actually dbObject.getEmailAddresses() should never be null, since emailAddresses is initialized -> https://github.com/all-of-us/workbench/blob/joel/hibernate-institution/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java#L28-L28 Hibernate also calls the setEmailAddresses(final Set<DbInstitutionEmailAddress> emailAddresses) with an empty collection if no emailAddresses exist. This test verifies this: https://github.com/all-of-us/workbench/blob/joel/hibernate-institution/api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java#L64-L64

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: copypasta.  good catch.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: Should our DAO level be aware of our API?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: Added a comment.  Effectively it is doing a simple set assignment.

For the last step, we could add the difference or the whole set again.  Should be identical.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: Merged these to avoid confusion

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: It is not nullable.  Thanks for the catch.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: Good point - updated

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: updated

jmthibault79:
Addressed comments and ready for re-review @jaycarlton @freemabd

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -69,16 +69,16 @@ private DbInstitution newDbObject(final Institution modelObject) {
...
+        .setOrganizationTypeEnum(modelObject.getOrganizationTypeEnum())
+        .setOrganizationTypeOtherText(modelObject.getOrganizationTypeOtherText())
+        .setEmailDomains(
             Optional.ofNullable(modelObject.getEmailDomains()).orElse(Collections.emptyList())
                 .stream()
                 .map(domain -> new DbInstitutionEmailDomain().setEmailDomain(domain))
                 .collect(Collectors.toSet()))
-        .emailAddresses(
+        .setEmailAddresses(
             Optional.ofNullable(modelObject.getEmailAddresses()).orElse(Collections.emptyList())
```
freemabd: This will also not be null but an empty collection.

jmthibault79: I don't think we can guarantee this because it's generated by Swagger with a null default:
`  private List<String> emailAddresses = null;`

freemabd: gotcha

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jaycarlton: Oh I thought it was boxed. nvm.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
+    return this;
   }
 
+  @Column(name = "organization_type_other_text")
   public String getOrganizationTypeOtherText() {
     return organizationTypeOtherText;
   }
 
-  public void setOrganizationTypeOtherText(String organizationTypeOtherText) {
+  public DbInstitution setOrganizationTypeOtherText(String organizationTypeOtherText) {
```
jaycarlton: So should we assert that the enum value is actually other when setting this? Should we just go ahead and update it if we're setting this? They're a package deal.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
-  public Short getOrganizationTypeEnum() {
-    return organizationTypeEnum;
+  @Column(name = "organization_type_enum")
+  public OrganizationType getOrganizationTypeEnum() {
+    return DbStorageEnums.organizationTypeFromStorage(organizationTypeEnum);
   }
 
-  public void setOrganizationTypeEnum(Short organizationTypeEnum) {
-    this.organizationTypeEnum = organizationTypeEnum;
+  public DbInstitution setOrganizationTypeEnum(OrganizationType type) {
```
jaycarlton: Yay for fluent methods.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
+   * <p>https://stackoverflow.com/questions/5587482/hibernate-a-collection-with-cascade-all-delete-orphan-was-no-longer-referenc
+   *
+   * @param emailDomains the new set of domains for this Institution
+   */
+  public DbInstitution setEmailDomains(final Set<DbInstitutionEmailDomain> emailDomains) {
+    final Set<DbInstitutionEmailDomain> attachedDomains =
+        emailDomains.stream()
+            .map(domain -> domain.setInstitution(this))
+            .collect(Collectors.toSet());
+    // modifies this set so that its value is the intersection of the two sets
```
jaycarlton: This looks like you're manually doing Hibernate's job. It seems really odd to me for an entity to take in a set of things it owns, then modify them and save them. I'd much rather see immutable email domain objects being pointed to by this set, unless you have a really urgent reason to work with domains but not the institution. Like where in the UI would you even have such an issue?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
+    this.emailDomains.retainAll(attachedDomains);
+    this.emailDomains.addAll(Sets.difference(attachedDomains, this.emailDomains));
+    return this;
   }
 
+  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+  @NotNull
```
jaycarlton: Are these all addresses belonging to the institution's users from anywhere in the system? I wouldn't have thought to look for that here; it's more of a service method thing in our current pattern. Could you equivalently expose a set of DbUser objects? I don't really know when we do one versus the other.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
   public Set<DbInstitutionEmailAddress> getEmailAddresses() {
     return emailAddresses;
   }
 
-  public void setEmailAddresses(Iterable<DbInstitutionEmailAddress> emailAddresses) {
-    this.emailAddresses =
-        StreamSupport.stream(emailAddresses.spliterator(), false).collect(Collectors.toSet());
+  /**
+   * Effectively: do an in-place this.emailAddresses = emailAddresses
```
jaycarlton: So the intention of this function is to replace all matching email addresses with updated ones (not sure what other fields that would involve). If this isn't bound directly to the table(s), then I'm confused about the use case.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
+   * @param emailAddresses the new set of addresses for this Institution
+   */
+  public DbInstitution setEmailAddresses(final Set<DbInstitutionEmailAddress> emailAddresses) {
+    final Set<DbInstitutionEmailAddress> attachedAddresses =
+        emailAddresses.stream()
+            .map(address -> address.setInstitution(this))
+            .collect(Collectors.toSet());
+    // modifies this set so that its value is the intersection of the two sets
+    this.emailAddresses.retainAll(attachedAddresses);
+    this.emailAddresses.addAll(Sets.difference(attachedAddresses, this.emailAddresses));
```
jaycarlton: does this save new ones to the database?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -15,51 +15,58 @@
...
-
-  @Column(name = "email_address", nullable = false)
   private String emailAddress;
 
   public DbInstitutionEmailAddress() {}
 
-  public DbInstitutionEmailAddress(DbInstitution institution, String emailAddress) {
-    this.institution = institution;
-    this.emailAddress = emailAddress;
+  @Id
```
jaycarlton: I think you should only define entity objects where it makes sense to your application to use as a standalone object. As in something you think of. If it's just a table row (with no more real abstraction), then you may be able to absorb it into the containing entity.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -15,51 +15,58 @@
...
-        && institution.equals(that.institution)
-        && emailAddress.equals(that.emailAddress);
+    return Objects.equals(institution, that.institution)
+        && Objects.equals(emailAddress, that.emailAddress);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(institutionEmailAddressId, institution, emailAddress);
+    return Objects.hash(institution, emailAddress);
```
jaycarlton: So this class allows the same email address to belong to more than one institution?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -11,88 +12,128 @@
...
+   * Then, call addAll() to add the diff(new - old) rows.
+   *
+   * <p>https://stackoverflow.com/questions/5587482/hibernate-a-collection-with-cascade-all-delete-orphan-was-no-longer-referenc
+   *
+   * @param emailAddresses the new set of addresses for this Institution
+   */
+  public DbInstitution setEmailAddresses(final Set<DbInstitutionEmailAddress> emailAddresses) {
+    final Set<DbInstitutionEmailAddress> attachedAddresses =
+        emailAddresses.stream()
+            .map(address -> address.setInstitution(this))
```
jaycarlton: I think we should validate that the incoming emailAddress doesn't already have a different institution. Otherwise this is destructive to that institution's map of the world.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: You mean of the type? Good point. In practice there's not much cyclical dependence added but you're right.

This is why I think we need a DataManager layer.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: I had a proposal for API object factories to help with that, but it didn't make it far.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -64,74 +55,49 @@ public boolean deleteInstitution(final String shortName) {
...
-
-    return institutionDao.save(dbClass);
+  private DbInstitution updateDbObject(
+      final DbInstitution dbObject, final Institution modelObject) {
+    return dbObject
+        .setShortName(modelObject.getShortName())
+        .setDisplayName(modelObject.getDisplayName())
+        .setOrganizationTypeEnum(modelObject.getOrganizationTypeEnum())
+        .setOrganizationTypeOtherText(modelObject.getOrganizationTypeOtherText())
+        .setEmailDomains(
```
jaycarlton: I like chained stream statements, but I have trouble reading nested streams. What would you think about making little helper methods to process the email domains and addresses?

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -64,74 +55,49 @@ public boolean deleteInstitution(final String shortName) {
...
+            Optional.ofNullable(modelObject.getEmailDomains()).orElse(Collections.emptyList())
+                .stream()
+                .map(domain -> new DbInstitutionEmailDomain().setEmailDomain(domain))
+                .collect(Collectors.toSet()))
+        .setEmailAddresses(
+            Optional.ofNullable(modelObject.getEmailAddresses()).orElse(Collections.emptyList())
+                .stream()
+                .map(address -> new DbInstitutionEmailAddress().setEmailAddress(address))
```
jaycarlton: If the `dbObject` is already known, then why not fill in both the address and institution here? It wold simplify your `setEmailAddresses()` method as well.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
-    assertThat(institutionDao.findAll()).hasSize(1);
-    institutionDao.save(new DbInstitution("VUMC", "Vanderbilt"));
-    assertThat(institutionDao.findAll()).hasSize(2);
+    final DbInstitution toSaveWith =
+        new DbInstitution()
+            .setShortName("Vanderbilt")
+            .setDisplayName("Vanderbilt University")
+            .setEmailDomains(emailDomains)
+            .setEmailAddresses(emailAddresses);
+    final DbInstitution savedWith = institutionDao.save(toSaveWith);
```
jaycarlton: nit: maybe `savedWithEmail`?

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
   }
 
   @Test
   public void test_delete() {
-    assertThat(institutionDao.findAll()).hasSize(1);
-    institutionDao.delete(institutionDao.findOneByShortName("Broad").get());
-    assertThat(institutionDao.findAll()).hasSize(0);
+    institutionDao.delete(institutionNoEmailAddressesOrDomains.getInstitutionId());
+    DbInstitution dbInstitution =
+        institutionDao.findOne(institutionNoEmailAddressesOrDomains.getInstitutionId());
```
jaycarlton: nit: if findOne wold return Optional, that'd be great.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
 @RunWith(SpringRunner.class)
 @DataJpaTest
-@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD)
 public class InstitutionDaoTest {
+
   @Autowired InstitutionDao institutionDao;
 
-  private DbInstitution testInst = new DbInstitution("Broad", "The Broad Institute");
+  private DbInstitution institutionNoEmailAddressesOrDomains;
```
jaycarlton: these are a mouthful and also very hard to distinguish (b/c the different letters are in the middle). Maybe `simpleInstitution` and `emailPopulatedInstition` or something?

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
     assertThat(institutionDao.findOneByShortName("Verily")).isEmpty();
+  }
+
+  @Test
+  public void test_updateAllNewEmailAddresses() {
+    Set<DbInstitutionEmailAddress> newEmailAddresses =
+        Sets.newHashSet(
+            new DbInstitutionEmailAddress().setEmailAddress("emailAddress1"),
+            new DbInstitutionEmailAddress().setEmailAddress("emailAddress2"));
+    institutionWithEmailAddressesAndDomains.setEmailAddresses(newEmailAddresses);
```
jaycarlton: Why can't this `setEmailAddresses` simply take in a collection of Strings? It could do the construction and population of the dbInstitution reference in one place so the caller doesn't have to.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
+  }
 
-    DbInstitution otherInst = new DbInstitution("Verily", "An Alphabet Company");
-    otherInst = institutionDao.save(otherInst);
-    assertThat(institutionDao.findOneByShortName("Verily")).hasValue(otherInst);
+  @Test
+  public void test_updateRemoveAllEmailAddresses() {
+    Set<DbInstitutionEmailAddress> newEmailAddresses = Sets.newHashSet();
+    institutionWithEmailAddressesAndDomains.setEmailAddresses(newEmailAddresses);
+    institutionDao.save(institutionWithEmailAddressesAndDomains);
```
jaycarlton: please overwrite the arg passed to save().

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
+        institutionDao.findOne(institutionWithEmailAddressesAndDomains.getInstitutionId());
+    assertThat(dbInstitution).isEqualTo(institutionWithEmailAddressesAndDomains);
+    assertThat(dbInstitution.getEmailDomains()).containsExactlyElementsIn(newEmailDomains);
+  }
+
+  @Test
+  public void test_updateWithExistingEmailDomains() {
+    Set<DbInstitutionEmailDomain> newEmailDomains =
+        Sets.newHashSet(
+            new DbInstitutionEmailDomain().setEmailDomain("emailDomain"),
```
jaycarlton: please make the test data more representative, e.g., `auburn.edu`.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -3,65 +3,192 @@
...
+        institutionDao.findOne(institutionWithEmailAddressesAndDomains.getInstitutionId());
+    assertThat(dbInstitution).isEqualTo(institutionWithEmailAddressesAndDomains);
+    assertThat(dbInstitution.getEmailAddresses()).containsExactlyElementsIn(newEmailAddresses);
+  }
+
+  @Test
+  public void test_updateWithExistingEmailAddresses() {
+    Set<DbInstitutionEmailAddress> newEmailAddresses =
+        Sets.newHashSet(
+            new DbInstitutionEmailAddress().setEmailAddress("emailAddress"),
```
jaycarlton: nit: make args look like email addresses

Also, we need some testing of the sharing (if allowed) of addresses across institutions.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: I will start one

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: We have no real requirement for this, so I'm keeping it loose to avoid added complexity.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: No.  These are the "allowed" emails from https://docs.google.com/document/d/1--wtRB14DK1qAMiz1mlEx12nT4HoUfYRjNkwQJwsycA/edit#heading=h.b5vt06vb56bp

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: yes

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: yes.  We have no requirement to exclude this.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: That is not how this works.  See the design doc.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: That is coming tomorrow when I convert to MapStruct

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: How would we specify which data table to use for that?

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: I'm not following

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: ok

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: will update

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: findOne() is built-in to CrudRepository.
the main accessor for client code findOneByShortName() does return an optional.

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: good catch

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
jmthibault79: Now that is an interesting idea.  I'll look into it.

jmthibault79: This does not appear to be possible with accessor-based Hibernate.

jaycarlton: ☹️ I thought the new hotness was field-based https://pmi-engteam.slack.com/archives/CPC84FQRF/p1580305523002500?thread_ts=1580305523.002500. I guess the jury is still out, but this is one more reason to do it on the properties I would think.

jmthibault79: I also came to that incorrect conclusion based on that conversation.  I have updated the conversation hopefully to avoid further confusion.

PR #3120, opened 2020-02-12
[RW-4437][risk=no] Track sidebar state in local storage

dolbeew:
Creates a local storage item to indicate whether the sidebar is open or closed. This way, the sidebar is open automatically only the first time a user logs in and enters a workspace. If the user re-enters a workspace or refreshes the page, it will get the sidebar state from local storage.

Also added a box-shadow in the sidebar so it's clearer that there is content being covered up.

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/workspace/workspace-wrapper/component.ts
+++ b/ui/src/app/pages/workspace/workspace-wrapper/component.ts
@@ -16,6 +16,8 @@ import {
 import {AnalyticsTracker} from 'app/utils/analytics';
 import {ResourceType, UserRole, Workspace, WorkspaceAccessLevel} from 'generated/fetch';
 
+const LOCAL_STORAGE_KEY_SIDEBAR_STATE = 'WORKSPACE_SIDEBAR_STATE';
+
 @Component({
   styleUrls: ['../../../styles/buttons.css',
     '../../../styles/headers.css'],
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/workspace/workspace-wrapper/component.ts
```diff
@@ -58,6 +58,12 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit(): void {
+    if (!!localStorage.getItem('sidebarState')) {
```
ericsong: nit: pull out the `sidebarState` value into a constant. I think we're (loosely) following the convention of using capital letters for local storage keys. 

Also, prefixing the key value with `WORKSPACE` may be helpful as we're planning to add the side bar component to the notebooks page.

ui/src/app/pages/workspace/workspace-wrapper/component.ts
```diff
@@ -58,6 +58,12 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
...
+      this.sidebarOpen = localStorage.getItem('sidebarState') === 'open';
+    } else {
+      this.sidebarOpen = true;
```
ericsong: can we use `setSidebarState` here?

ui/src/app/pages/workspace/workspace-wrapper/component.ts
```diff
@@ -67,6 +73,7 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
           this.setHelpContent();
           if (this.helpContent !== 'reviewParticipantDetail') {
             this.sidebarOpen = false;
+            localStorage.setItem('sidebarState', 'closed');
```
ericsong: also here.

nit - IMO, it's easier to maintain the syncing to localStorage if its done in a one place (the setter) and the rest of the file just uses the setter and doesn't have to worry about localStorage.

ui/src/app/pages/workspace/workspace-wrapper/component.ts
dolbeew: Okay, now I see this convention used in the `signed-in` component also. Fixed.

ui/src/app/pages/workspace/workspace-wrapper/component.ts
dolbeew: Done.

ui/src/app/pages/workspace/workspace-wrapper/component.ts
dolbeew: Agreed, much cleaner this way.

PR #3123, opened 2020-02-12
[RW-4386][risk=low] Set correct billing status after updating a workspace's billing account

ericsong:
Correctly sets the billing account status after a new billing account is set on a workspace.

Effectively, it does the following
-  Setting a workspace to an open billing account will reactive its billing status
-  Users cannot set a workspace to a closed billing account
-  Users that set a workspace to free tier
    - With free credits remaining - workspace is set to billing active
    - With no credits remaining - workspace is set to billing inactive

TIL
- Javascript exceptions thrown by API requests need to be deserialized
- `ExceptionAdvice.java` has the logic for our catch all exception -> API mapping. There's a bug where the top level exception is ignored if there are nested exceptions.

---
**PR checklist**


api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -787,22 +793,156 @@ public void testUpdateWorkspace() throws Exception {
...
     workspacesController.updateWorkspace(ws.getNamespace(), ws.getId(), request);
 
-    verifyNoMoreInteractions(endUserCloudbillingProvider.get());
+    verifyZeroInteractions(endUserCloudbillingProvider.get());
     verify(serviceAccountCloudbillingProvider.get()).projects();
   }
 
+  @Test
+  public void testUpdateWorkspace_freeTierBilling_noCreditsRemaining() throws Exception {
+    Workspace ws = createWorkspace();
```
gjuggler: [opt] I tend to try to avoid abbreviations for var names in Java, though they're prevalent enough in our codebase I wouldn't block on this.

ui/src/app/components/breadcrumb.tsx
```diff
@@ -167,10 +167,13 @@ export const Breadcrumb = fp.flow(
       if (!prevProps.workspace && this.props.workspace &&
         this.props.workspace.billingStatus === BillingStatus.INACTIVE) {
         this.setState({showInvalidBillingBanner: true});
-      }
-
-      if (prevProps.workspace && !this.props.workspace) {
+      } else if (prevProps.workspace && !this.props.workspace) {
```
gjuggler: [No action] I haven't come up against the breadcrumb code before, so it's a little surprising to see some very workspace & billing specific logic showing up in a component that otherwise seems to be very generic. Could you imagine (not for this PR, or not anytime soon) a better home for this? Or is there a a strong reason that this logic belongs in this component?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -694,6 +698,16 @@ public void updateWorkspaceBillingAccount(DbWorkspace workspace, String newBilli
       cloudbilling = serviceAccountCloudbillingProvider.get();
     } else {
       cloudbilling = endUserCloudbillingProvider.get();
+      try {
+        if (!Optional.ofNullable(
```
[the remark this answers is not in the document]
ericsong: That's actually a different statement since `isPresent()` would return true if `isOpen` is set to false.

But probably not a good sign if that's not clear so I refactored it a little bit by pulling out a variable.

ui/src/app/components/breadcrumb.tsx
ericsong: Definitely a bit of a hack. It's mainly being used because it was the best React component I could find that is loaded on every workspace page.

I had initially tried to put it in `workspace-wrapper` but that is still in angular.

PR #3124, opened 2020-02-13
[RW-4307][risk=low] Enable keyless delegation in perf & staging

gjuggler:
See Jira comments for some added context. The idea is to roll this to all lower envs next week, then to prod the week after.

### DIFF
--- a/api/config/config_perf.json
+++ b/api/config/config_perf.json
@@ -98,7 +98,7 @@
     "enableRdrExport": false,
     "enableBillingLockout": false,
     "enableSumoLogicEventHandling": false,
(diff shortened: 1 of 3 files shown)


gjuggler:
> Given our current QA patterns, it's somewhat ideal if staging tracks prod's features. However, if that's not feasible, I'd probably at least hold back stable and do stable/prod together.
> 
> This at least maintains the security-stated goal of stable matching prod as closely as possible.

Sounds good – I think having Rainforest QA run on staging will be a useful stress test here, so I'll keep staging enabled here, and rollout stable with prod the next release.

PR #3125, opened 2020-02-13
[risk=no]Proposal: move mappers to common package org.pmiops.workbench.mappers

jmthibault79:
Description:

I thought it would be nice for all the mappers to be in one place.  LMK if you agree.

---
**PR checklist**


ericsong:
My personal preference would be to organize by feature/domains instead of by function.

Ex. workspace package contains WorkspaceController, Workspace models, WorkspaceDao, WorkspaceMapper

Mainly because I think it scales better long term but I think we can go either way for a small-ish project like workbench. Like most of our other technical decisions, consistency is what mostly matters.

gjuggler:
> My personal preference would be to organize by feature/domains instead of by function.
> 
> Ex. workspace package contains WorkspaceController, Workspace models, WorkspaceDao, WorkspaceMapper
> 
> Mainly because I think it scales better long term but I think we can go either way for a small-ish project like workbench. Like most of our other technical decisions, consistency is what mostly matters.

I agree we should aim to eventually have functionally-separated packages, each containing the full suite of supporting classes.

For the mappers in this PR, it looks like there are reasonably-named functional packages they could go into. And CommonMappers could stay where it is, or go directly into workbench/utils. My vote would be for that.

(Side note: I don't know why it bothers me so, but the workbench/workspaces package should really be named workbench/workspace, preferring the singular form [[I agree with this reasoning](https://softwareengineering.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural)]. And we're almost perfectly inconsistent on this naming rule: cohorts, notebooks, and statusalerts use plural, while conceptset, dataset, and institution use singular.)

jaycarlton:
> : I don't know why it bothers me so, but the workbench/workspaces package should really be named workbench/workspace, preferring the singular form

I got unofficial team approval at one point to make that change and similar ones at some point in the future.

jmthibault79:
Did not realize we were moving to that convention - works for me.

PR #3126, opened 2020-02-13
[risk=no][RW-4438] Use actual query instead of Java stream over all table rows

jaycarlton:
Replace a number of slow queries with sensible HQL versions now that I know how to use projection interfaces. I'm sad to see my nifty generic function go, but it had a good run.

TIL:
- Projection interfaces are a thing and work more easily with HQL
- you can't use an `@Transient` entity property in your queries 😢 
- dao tests are really handy for these kinds of things (and it'd be nice to set up some more soon)

**PR checklist**


api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -40,6 +44,29 @@
     return DaoUtils.getAttributeToCountMap(findAll(), DbBillingProjectBufferEntry::getStatusEnum);
   }
 
+  @Query(value = "select status, count(billing_project_buffer_entry.billing_project_buffer_entry_id) as num_projects\n"
+      + "    from billing_project_buffer_entry\n"
+      + "group by status\n"
+      + "order by status;", nativeQuery = true)
+//  List<BillingProjectBufferEntryStatusToCountResult> computeProjectCountByStatus();
+   BillingProjectBufferEntryStatusToCountResult[] computeProjectCountByStatus();
```
ericsong: Driving by but is it necessary to add this intermediate object? I think the interface we had, `getCountByStatusMap()`, works well. We just need to change the implementation to speed it up.

The `computeProjectCountByStatus2()` call below should work and that gets us most of the way there. All `getCountByStatusMap()` would have to do is convert the result from `computeProjectCountByStatus2()` to match the existing interface.

jaycarlton: I don't know how to capture the query result safely without a new POJO or entity. Like, that's what I'm supposed to get from. packages. like this I thought. If I'm having to. cast from. object and then. manually call a constructor with parameters the same names as my columns, then. I want to shop for another framework.  Ideally I could just make an Immutable for it.

But if that works, it should help.

jmthibault79:
Does this example help?  It's in BillingProjectBufferEntryDao.  Maybe the trick is that this is NOT a native query and it uses JPA/Hibernate's class names.
```
  interface UserCostPair {
    DbUser getUser();

    Double getCost();
  }

  @Query(
      "SELECT u.user, SUM(u.cost) "
          + "FROM DbWorkspaceFreeTierUsage u "
          + "WHERE u.cost IS NOT NULL "
          + "GROUP BY u.user")
  Collection<UserCostPair> totalCostByUser();

  default Map<DbUser, Double> getUserCostMap() {
    return totalCostByUser().stream()
        .collect(Collectors.toMap(UserCostPair::getUser, UserCostPair::getCost));
  }
```

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
freemabd: @jaycarlton If you would like to catch the query output in an object but don't want to have to annotate another entity... you can use a Projection interface. It allows you to project a subset of properties from your dao into another object that isn't your actual entity.
Projection interface example here: https://github.com/all-of-us/workbench/blob/freemabd/address-comments/common-api/src/main/java/org/pmiops/workbench/cdr/model/DbMenuOption.java#L3-L3 
Usage example here: https://github.com/all-of-us/workbench/blob/freemabd/address-comments/common-api/src/main/java/org/pmiops/workbench/cdr/dao/CBCriteriaDao.java#L238-L238

ericsong: oo, this is neat! thanks @freemabd 

Also, another comment, I think its best to avoid using native queries where possible. Native queries will run the query as is against the underlying database. This would be OK if we used the same database everywhere but I believe our tests use some in memory database (H2?).

I have had issues where native queries work in production and fail in tests and vice versa.

Writing HQL will let hibernate figure out the right way to generate the query based on what database is currently in use. Caveat is that you don't have access to all the SQL tools that MySQL provides but HQL should suffice for most cases.

jaycarlton: Great points. Was going to try going up a level. in abstraction after getting something working.

jaycarlton: Projection interface works! Much nicer. No annotatitons even.

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -48,7 +46,21 @@
 
   Set<DbUser> findByFirstRegistrationCompletionTimeNotNull();
 
-  default Map<Boolean, Long> getDisabledToCountMap() {
-    return DaoUtils.getAttributeToCountMap(findAll(), DbUser::getDisabled);
+  @Query(
+      "SELECT dataAccessLevel, disabled, CASE WHEN betaAccessBypassTime IS NOT NULL THEN TRUE ELSE FALSE END AS betaIsBypassed, COUNT(userId) AS userCount "
```
jaycarlton: One way around this ugliness is a persisted, computed column. Not worth it here, as much as I hate seeing this comparison to null everywhere to essentially build a boolean.

api/src/main/java/org/pmiops/workbench/db/dao/WorkspaceDao.java
```diff
@@ -73,11 +69,16 @@ default void updateBillingStatus(long workspaceId, BillingStatus status) {
   @Query("SELECT w.creator FROM DbWorkspace w WHERE w.billingStatus = (:status)")
   Set<DbUser> findAllCreatorsByBillingStatus(@Param("status") BillingStatus status);
 
-  default Map<WorkspaceActiveStatus, Long> getActiveStatusToCountMap() {
-    return DaoUtils.getAttributeToCountMap(findAll(), DbWorkspace::getWorkspaceActiveStatusEnum);
-  }
+  @Query(
+      "SELECT activeStatus, dataAccessLevel, COUNT(workspaceId) FROM DbWorkspace "
+          + "GROUP BY activeStatus, dataAccessLevel ORDER BY activeStatus, dataAccessLevel")
+  List<ActiveStatusAndDataAccessLevelToCountResult> getActiveStatusAndDataAccessLevelToCount();
```
freemabd: Can we write a test case for this in `WorkspaceDaoTest`?

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
```diff
@@ -48,7 +46,21 @@
...
+          + "FROM DbUser "
+          + "GROUP BY dataAccessLevel, disabled, CASE WHEN betaAccessBypassTime IS NOT NULL THEN TRUE ELSE FALSE END "
+          + "ORDER BY NULL")
+  List<DataAccessLevelDisabledAndBetaBypassedToCountRow>
+      getDataAccessLevelDisabledAndBetaBypassedToCount();
```
freemabd: Can we write a test class for `UserDaoTest` and test all methods?

api/src/main/java/org/pmiops/workbench/db/dao/DataSetDao.java
```diff
@@ -16,6 +17,20 @@
...
+    final List<InvalidToCountResult> rows = getInvalidToCount();
+    return rows.stream()
+        .collect(
+            ImmutableMap.toImmutableMap(
+                InvalidToCountResult::getIsInvalid, InvalidToCountResult::getInvalidCount));
+  }
+
+  @Query(
+      "SELECT invalid, count(dataSetId) AS invalidCount FROM DbDataset GROUP BY invalid ORDER BY invalid")
+  List<InvalidToCountResult> getInvalidToCount();
```
freemabd: Can we write a test class for this in `DataSetDaoTest` and test all the methods?

api/src/main/java/org/pmiops/workbench/db/dao/BillingProjectBufferEntryDao.java
```diff
@@ -28,18 +28,27 @@
...
+                StatusToCountResult::getStatusEnum, StatusToCountResult::getNumProjects));
   }
 
+  @Query(
+      value =
+          "select status, count(billing_project_buffer_entry_id) as numpNrojects\n"
+              + "    from DbBillingProjectBufferEntry \n"
+              + "group by status\n"
+              + "order by status")
+  List<StatusToCountResult> computeProjectCountByStatus();
```
freemabd: Can we write a test class for this in `BillingProjectBufferEntryDaoTest` and test all the methods?

api/src/main/java/org/pmiops/workbench/db/dao/UserDao.java
jaycarlton: Yep. This is my first Dao Test, and the only thing that's tricky is understanding when my autowired dao is cleared between tests.

api/src/main/java/org/pmiops/workbench/db/dao/DataSetDao.java
jaycarlton: For scope reasons,  I'm going to just. cover the methods added here for now.

PR #3131, opened 2020-02-14
[risk=no] A grab bag of small refactorings while doing other work

jmthibault79:
Description:
1. About half of `ProfileController.updateProfile()` deals with old-style Institutional Affiliations, so I moved that into its own method.
2. The shorter version of `UserService.createUser()` exists solely for AuthInterceptor's use, and the parameter list now matches what it knows about.
3. CommonStorageEnumsTest did not comprehensively check for missing map values like StorageEnumsTest does.
4. project.rb connect-to-db now starts the local DB if it's not running

---
**PR checklist**

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -505,7 +505,8 @@ Common.register_command({
 
 def connect_to_db()
   common = Common.new
-
+  common.status "Starting database if necessary..."
+  common.run_inline %W{docker-compose up -d db}
   cmd = "MYSQL_PWD=root-notasecret mysql --database=workbench"
   common.run_inline %W{docker-compose exec db sh -c #{cmd}}
 end
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -580,6 +580,18 @@ public class ProfileController implements ProfileApiDelegate {
       // See RW-1488.
       throw new BadRequestException("Changing email is not currently supported");
     }
+    updateInstitutionalAffiliations(updatedProfile, user);
+
+    // This does not update the name in Google.
+    saveUserWithConflictHandling(user);
+
+    final Profile appliedUpdatedProfile = profileService.getProfile(user);
+    profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
+
+    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+  }
+
+  private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user) {
     List<DbInstitutionalAffiliation> newAffiliations =
         updatedProfile.getInstitutionalAffiliations().stream()
             .map(FROM_CLIENT_INSTITUTIONAL_AFFILIATION)
@@ -614,14 +626,6 @@ public class ProfileController implements ProfileApiDelegate {
         user.addInstitutionalAffiliation(affiliation);
       }
     }
-
-    // This does not update the name in Google.
-    saveUserWithConflictHandling(user);
-
-    final Profile appliedUpdatedProfile = profileService.getProfile(user);
-    profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
-
-    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
   @Override
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
@@ -1,5 +1,6 @@
 package org.pmiops.workbench.db.dao;
 
+import com.google.api.services.oauth2.model.Userinfoplus;
 import java.io.IOException;
 import java.sql.Timestamp;
 import java.util.List;
@@ -17,8 +18,8 @@ public interface UserService {
 
   DbUser createServiceAccountUser(String email);
 
-  // minimal version used by AuthInterceptor
-  DbUser createUser(String givenName, String familyName, String userName);
+  // version used by AuthInterceptor
+  DbUser createUser(Userinfoplus oAuth2Userinfo);
 
   DbUser createUser(
       String givenName,
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
@@ -17,19 +17,13 @@ public interface UserService {
 
   DbUser createServiceAccountUser(String email);
 
-  DbUser createUser(
-      String givenName,
-      String familyName,
-      String email,
-      String contactEmail,
-      String currentPosition,
-      String organization,
-      String areaOfResearch);
+  // minimal version used by AuthInterceptor
+  DbUser createUser(String givenName, String familyName, String userName);
 
   DbUser createUser(
       String givenName,
       String familyName,
-      String email,
+      String userName,
       String contactEmail,
       String currentPosition,
       String organization,
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -1,6 +1,7 @@
 package org.pmiops.workbench.db.dao;
 
 import com.google.api.client.http.HttpStatusCodes;
+import com.google.api.services.oauth2.model.Userinfoplus;
 import com.google.common.collect.ImmutableSet;
 import java.io.IOException;
 import java.sql.Timestamp;
@@ -290,9 +291,20 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
   }
 
   @Override
-  public DbUser createUser(String givenName, String familyName, String userName) {
+  public DbUser createUser(final Userinfoplus oAuth2Userinfo) {
     return createUser(
-        givenName, familyName, userName, null, null, null, null, null, null, null, null, null);
+        oAuth2Userinfo.getGivenName(),
+        oAuth2Userinfo.getFamilyName(),
+        oAuth2Userinfo.getEmail(),
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null);
   }
 
   @Override
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -290,27 +290,9 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
   }
 
   @Override
-  public DbUser createUser(
-      String givenName,
-      String familyName,
-      String userName,
-      String contactEmail,
-      String currentPosition,
-      String organization,
-      String areaOfResearch) {
+  public DbUser createUser(String givenName, String familyName, String userName) {
     return createUser(
-        givenName,
-        familyName,
-        userName,
-        contactEmail,
-        currentPosition,
-        organization,
-        areaOfResearch,
-        null,
-        null,
-        null,
-        null,
-        null);
+        givenName, familyName, userName, null, null, null, null, null, null, null, null, null);
   }
 
   @Override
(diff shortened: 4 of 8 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -17,19 +17,13 @@
...
-      String organization,
-      String areaOfResearch);
+  // minimal version used by AuthInterceptor
+  DbUser createUser(String givenName, String familyName, String userName);
 
   DbUser createUser(
       String givenName,
       String familyName,
-      String email,
+      String userName,
```
jmthibault79: we must have missed this while standardizing on the new name

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -290,27 +290,9 @@ public DbUser createServiceAccountUser(String username) {
   }
 
   @Override
-  public DbUser createUser(
-      String givenName,
-      String familyName,
-      String userName,
-      String contactEmail,
-      String currentPosition,
```
jmthibault79: The last 3 were always null

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -155,18 +155,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
       // TODO(danrodney): start populating contact email in Google account, use it here.
       user =
           userService.createUser(
-              userInfo.getGivenName(),
-              userInfo.getFamilyName(),
-              userInfo.getEmail(),
```
jmthibault79: `userInfo` is a Google OAuth2 `Userinfoplus` object so "email" is correct from that perspective, though it's a "userName" to Workbench.

common-api/src/main/java/org/pmiops/workbench/db/model/CommonStorageEnums.java
```diff
@@ -108,14 +108,6 @@ public static Short surveysToStorage(Surveys survey) {
     return CLIENT_TO_STORAGE_SURVEY.get(survey);
   }
 
-  public static String surveyToSurveyId(Surveys survey) {
```
jmthibault79: These are not used

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -51,4 +55,52 @@ public void testBijectiveStorageMapping() {
...
+          enumToShort = m;
+        }
+
+        if (m.getReturnType().equals(enumClass)) {
+          shortToEnum = m;
+        }
+      }
+
+      // stringly typed map - test with testDomainIdBijectiveStorageMapping instead
+      if (enumToShort.getName().equals("domainIdToDomain")) {
```
jmthibault79: Is there a more elegant way?  I want to be sure to exclude only this one, for future-proofing

jmthibault79: Not worth spending much time on.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -580,6 +580,18 @@ private boolean userHasEverLoggedIn(
       // See RW-1488.
       throw new BadRequestException("Changing email is not currently supported");
     }
+    updateInstitutionalAffiliations(updatedProfile, user);
```
jaycarlton: Do we have a name for the old affiliations? Or did we give the new ones a different name?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -580,6 +580,18 @@ private boolean userHasEverLoggedIn(
...
+
+    // This does not update the name in Google.
```
jaycarlton: Can you comment on the semantics/lifecycle keeping these in sync, and why we don't need to here?

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
jaycarlton: Yes. Sometimes the refactoring is smart enough to do what you want with these, but if one of the names is a substring of other identifiers, all heck breaks loose. Safer to risk leaving some local identifiers behind like this.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -290,27 +290,9 @@ public DbUser createServiceAccountUser(String username) {
...
-      String organization,
-      String areaOfResearch) {
+  public DbUser createUser(String givenName, String familyName, String userName) {
     return createUser(
```
jaycarlton: I'm declaring shenanigans on this constructor. We need a builder. 

There's no reason in Java 8+ I should ever have to see the word `null`, let alone 9 in a row. But that may be another ticket. 

If you don't want a builder, then your shorter constructors need to call the longer ones with default methods.

```
User createUser(A a) {
  return createUser(a, defaultB);
}

User createUser(A a, B b) {
  return createUser(a, b, defaultC);
}
// etc 
```
But a builder is still much preferred.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
jaycarlton: Can you put that comment in the code and replace `userInfo` with something like `googleOAuth2id`?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -51,4 +55,52 @@ public void testBijectiveStorageMapping() {
       assertThat(v).isEqualTo(fromStorage.apply(storageValue));
     }
   }
+
+  // domain ID is stringly-typed so special-case this
+
+  @Test
+  public void testDomainIdBijectiveStorageMapping() {
+    for (Domain v : Domain.values()) {
+      String storageValue = CommonStorageEnums.domainToDomainId(v);
```
jaycarlton: CommonStorageEnums should throw IllegalArgumentException if it's not matched IMO. This assertion seems redundant; such checking is that method's job.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: This looks like a fun exercise, but it's way too abstract/reflexive for test code (or application code). Plus, I want to kill dbStorageEnums soon in favor of Lebowski Enums.

Can we shelve this for another day? I'm happy to pair on the enum stuff. It's a bit funny looking at first, but I think you'll like it.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -51,4 +55,52 @@ public void testBijectiveStorageMapping() {
...
+          shortToEnum = m;
+        }
+      }
+
+      // stringly typed map - test with testDomainIdBijectiveStorageMapping instead
+      if (enumToShort.getName().equals("domainIdToDomain")) {
+        continue;
+      }
+
+      for (Object e : enumClass.getEnumConstants()) {
```
jaycarlton: You're just testing BiMap at this point I think.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jmthibault79: I know nothing about this.  I just moved it.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jmthibault79: New ones are Verified Institutional Affiliations

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
jmthibault79: This is a good idea, but a larger refactoring than I am comfortable doing right now

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
jmthibault79: done

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: I don't believe this is fixable without a wholesale rewrite of our enum system.

Which would be a good idea!  Just not in this tiny drive-by PR

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: This actually does catch bugs!  Its StorageEnumsTest counterpart caught one of mine today - that's what prompted me to do this.  Remove an entry from one of the BiMaps to observe this.

jaycarlton: No, I get what it's doing, and it's awesome that it catches bugs.

Is there a reason you have to copy this instead of making a generic utility function to use in both enum classes?

jmthibault79: We have a one-way dependency from /api -> /common-api so any common code would have to live in /common-api.  I don't want to add any new /common-api dependencies.  We'll be removing /common-api entirely, soon after Dover.

api/libproject/devstart.rb
```diff
@@ -505,7 +505,8 @@ def run_gradle(cmd_name, args)
 
 def connect_to_db()
   common = Common.new
-
+  common.status "Starting database if necessary..."
```
blrubenstein: Nice

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jaycarlton: 👍

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jaycarlton: We need to be careful moving the audit stuff like `fireUpdateAction()`. This looks right though. Can you please verify we still get the events written to the audit stream?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: Some more English comments on what each chunk is doing would help. Most of us don't use these kinds of functions every day.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: But why did your new enum maps need to go in common at all?

jmthibault79: They did not.  I saw that the /api test saved my butt and was curious if the /common-api would have as well.  It would not have, so I fixed that here.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jmthibault79: What I moved was the `updateInstitutionalAffiliations()` block into its own method.  Everything else remained in the same location.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: I'll do that but in a follow up along with some other issues I've found

jmthibault79: Comments and fix are at #3159

PR #3133, opened 2020-02-14
[risk=no][RW-4384] Disable closed billing accounts in dropdown

ericsong:
Description:

Note - I wanted to add a tooltip to the disabled options but it's difficult to do since we're using a library to handle our dropdowns. I think we should look into a more flexible library or roll our own. Even finding out if they support `disabled` was a struggle since it's not documented anywhere. I had to read their source to find out.

image

---
**PR checklist**

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -93,7 +93,7 @@
     "enableVpcServicePerimeter": true,
     "enableNewAccountCreation": true,
     "enableRdrExport": true,
-    "enableBillingLockout": false,
+    "enableBillingLockout": true,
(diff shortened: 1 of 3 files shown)


jaycarlton:
UX-wise, I think I'd be really confused/irritated having to scroll over disabled entries in a dropdown list like this. Imagine if I had 9 disabled and 1 enabled. I understand the need to surface them, but this seems like maybe not the most intuitive way to do it.

Come to think of it, I can't remember ever scrolling over disabled entries in a dropdown control. The usual/expected contract is that everything in that kind of control is clickable.

PR #3135, opened 2020-02-14
[RW-2704][risk=low] Split up listCluster API and move initialization logic to the client-side.

gjuggler:
This PR resolves RW-2704 by splitting up the server-side listCluster endpoint into component get/create/delete pieces.

As a result of that server-side simplification, the initialization state management and retry logic had to move somewhere. I opted to implement this on the client side, since we already had a few client-side classes which were effectively repeating the same or similar initialization logic (notebook-redirect, interactive-notebook, and reset-cluster-button).

The resulting controller class, cluster-initializer, is meant to consolidate this logic into one well-tested module. Each React component uses this in a more-or-less simple way (see below for notes on why reset-cluster-button is special) with the Promise based API.

Some notes:
 - Unit testing Promises with time-based delays isn't currently possible in Jest (see https://github.com/facebook/jest/issues/7151). I had originally implemented the polling loop using a while loop within an async function, but I ultimately switched to using a poll() method and setTimeout to make unit testing more feasible.
 - The reset cluster button is a bit of a special case here, since we don't want the button to take any action until the user explicitly chooses to "Reset" (this is the whole point of RW-2704). The solution I took was to create a "no-action" initializer when the button is first loaded, and to create a "action-ready" initializer when the button is clicked. It's a bit awkward, but less awkward than almost all other options I could think of.

Ultimately, while working through this PR it became clear that this functionality should really live somewhere on the server-side, ideally within the analysis application (e.g. Leo) itself. That would allow Terra and AoU, for example, to simplify both of our clients. Something worth thinking about as Framework platform convergence discussions continue.


ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,136 @@
...
+      if (this.isClusterStopped()) {
+        console.log('Cluster is stopped, resuming');
+        await this.resumeCluster();
+      }
+
+      // If cluster is in error state, delete it so it can be re-created at the next poll loop.
+      if (this.isClusterErrored()) {
+        // TODO: report errors to Stackdriver here.
+        console.log('Cluster error -- deleting for retry');
+        await this.deleteCluster();
```
calbach: Per ticket discussion, I might move this above the loop so it only happens on initialization - errors encountered later could always throw and require user intervention. I think this limits the risk of concurrent pollers causing cluster churn - delete is the only really problematic/destructive action here. The rest should be relatively safe for concurrent requests.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,136 @@
...
+        console.log('Getting cluster');
+        this.currentCluster = await this.getCluster();
+        console.log('Cluster status', this.currentCluster.status);
+        this.onStatusUpdate(this.currentCluster.status);
+      } catch (e) {
+        console.log('getCluster error', e);
+        if (isAbortError(e)) {
+          return Promise.reject();
+        }
+        // If we received a NOT_FOUND error, we need to create a cluster for this workspace.
```
calbach: Probably a  pre-emptive comment, but this should actually check for NOT_FOUND

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,136 @@
...
+    console.log('Initializing cluster', this.workspaceNamespace);
+    if (this.isInitializing) {
+      return Promise.reject('Initialization is already in progress');
+    }
+    this.initializeStartTime = new Date().getTime();
+
+    // Overall strategy: continue polling the get-cluster endpoint, with capped exponential backoff,
+    // until we either reach our goal state (a RUNNING cluster) or run up against the overall
+    // timeout threshold.
+    while (!this.isClusterRunning()) {
```
calbach: One thing which isn't covered here yet is error retries. Since this is a very long polling cycle (expected to commonly be ~7-10m), would be ideal if this were resilient to the occasional transient 500. Using this package on the outgoing calls would probably suffice: https://github.com/all-of-us/workbench/blob/d86dead9ffa411924105f85b1116253d7cd53412/ui/src/app/utils/retry.tsx#L8

(e.g. only fail the polling loop if the request fails X times in a row)

gjuggler:
@calbach this is finally worth another look. I'm really not in love with how complex the solution got, but IMO it's probably better to have one gnarly-but-well-tested thing rather than simpler-looking-but-more-fragile logic spread around in a few components. Curious to hear whether you agree.

@jaycarlton as I warned you, here's a client-side PR that feels like a server-side PR (for obvious reasons...). Feel free to just browse, or treat this like a full code review & dig into questionable-looking areas, etc.

Both: I pushed a relatively-recent version of this branch to my test GAE service, accessible here: https://gjordan-dot-all-of-us-workbench-test.appspot.com/ There's one known issue I need to fix (the cluster reset button looks weird while initially fetching cluster status), but please use that to run it through the ringer.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
+  }
+
   @Override
-  public ResponseEntity<DefaultClusterResponse> listClusters(
-      String billingProjectId, String workspaceFirecloudName) {
-    if (billingProjectId == null) {
-      throw new BadRequestException("Must specify billing project");
-    }
+  public ResponseEntity<Cluster> getCluster(String workspaceNamespace) {
+    String firecloudWorkspaceName = lookupWorkspace(workspaceNamespace).getFirecloudName();
```
jaycarlton: nit: can the body of this function be pushed into one of the services? Or do we not want either to know about the other?

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
-        body.getWorkspaceNamespace(), body.getWorkspaceId(), WorkspaceAccessLevel.READER);
-    workspaceService.validateActiveBilling(body.getWorkspaceNamespace(), body.getWorkspaceId());
+        dbWorkspace.getWorkspaceNamespace(),
+        dbWorkspace.getFirecloudName(),
+        WorkspaceAccessLevel.WRITER);
+    workspaceService.validateActiveBilling(
+        dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getFirecloudName());
 
-    FirecloudWorkspace fcWorkspace;
+    FirecloudWorkspace firecloudWorkspace;
```
jaycarlton: I'd make this final so the complier assures it's initialized exactly once.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
     // namespace, simply name the directory as the workspace ID; else we
     // include the namespace in the directory name to avoid possible conflicts
     // in workspace IDs.
-    String gcsNotebooksDir = "gs://" + fcWorkspace.getBucketName() + "/notebooks";
-    Timestamp now = new Timestamp(clock.instant().toEpochMilli());
-    long workspaceId =
-        workspaceService
-            .getRequired(body.getWorkspaceNamespace(), body.getWorkspaceId())
-            .getWorkspaceId();
+    String gcsNotebooksDir = "gs://" + firecloudWorkspace.getBucketName() + "/notebooks";
```
jaycarlton: I'd make a helper method or service method for this bucket path determination. (If `FirecloudWorkspace` weren't generated, I'd put it there.)

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
                     workspaceId, userProvider.get().getUserId(), gcsNotebooksDir + "/" + notebook));
-    String workspacePath = body.getWorkspaceId();
-    if (!projectName.equals(body.getWorkspaceNamespace())) {
-      workspacePath =
-          body.getWorkspaceNamespace()
-              + FireCloudService.WORKSPACE_DELIMITER
-              + body.getWorkspaceId();
-    }
 
+    String workspacePath = dbWorkspace.getFirecloudName();
```
jaycarlton: on ln 263, I was confused that `notebook` was the variable but it was appended to a string. `notebookName` might be better.

Also, it would be nice to have ` userRecentResourceService` expose a batch method for this, so you don't have to loop here. It's a bit odd to see a service getting called in a loop.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
+
+    this.leonardoNotebooksClient.deleteCluster(
+        workspaceNamespace, clusterNameForUser(userProvider.get()));
     return ResponseEntity.ok(new EmptyResponse());
   }
 
   @Override
   public ResponseEntity<ClusterLocalizeResponse> localize(
-      String projectName, String clusterName, ClusterLocalizeRequest body) {
+      String workspaceNamespace, ClusterLocalizeRequest body) {
```
jaycarlton: I don't really know what `localize` means in this context, but I can see this method does a large number of things that are all important (although transaction management between service calls is potentially sloppy). It would be helpful for readability and testability to have a high-level comment and also break down the body into helper methods and service methods. (Can certainly come later)

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -306,7 +285,7 @@
 
     // The Welder extension offers direct links to/from playground mode; write the AoU config file
     // to both locations so notebooks will work in either directory.
-    String aouConfigUri = aouConfigDataUri(fcWorkspace, cdrVersion, projectName);
+    String aouConfigUri = aouConfigDataUri(firecloudWorkspace, cdrVersion, workspaceNamespace);
```
jaycarlton: Above @ line 284, I think we can have a better name than `localizeMap`, e.g. `bucketPathToNotebookPath` if I'm reading things right.

api/src/main/java/org/pmiops/workbench/api/ClusterController.java
```diff
@@ -186,115 +183,97 @@
...
 
+    String workspacePath = dbWorkspace.getFirecloudName();
     String editDir = "workspaces/" + workspacePath;
     String playgroundDir = "workspaces_playground/" + workspacePath;
     String targetDir = body.getPlaygroundMode() ? playgroundDir : editDir;
 
     leonardoNotebooksClient.createStorageLink(
-        projectName,
-        clusterName,
+        workspaceNamespace,
```
jaycarlton: Can we rename `targetDir` to `notebookLocalDir`? Also, collecting lines 269-272 into a helper method would be nice.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -169,7 +171,12 @@ public void deleteCluster(String googleProject, String clusterName) {
   @Override
   public Cluster getCluster(String googleProject, String clusterName) {
     ClusterApi clusterApi = clusterApiProvider.get();
-    return retryHandler.run((context) -> clusterApi.getCluster(googleProject, clusterName));
+    try {
+      return retryHandler.runAndThrowChecked(
+          (context) -> clusterApi.getCluster(googleProject, clusterName));
```
jaycarlton: nit: maybe rename lambda variable to `unusedContext`. You can drop the parenthesis for single-arg lambdas FWIW.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -33,6 +33,8 @@
 
   List<WorkspaceResponse> getWorkspacesAndPublicWorkspaces();
 
+  WorkspaceResponse getWorkspace(String workspaceNamespace);
```
jaycarlton: What do you think about either naming this `getWorkspaceOrThrow()` and/or adding a `throws NotFoundException` to the signature. @ericsong and I came to the compromise that doing this on service methods is a net win, even though you ordinarily don't declare runtime exceptions.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -204,18 +204,32 @@ public DbWorkspace get(String ns, String firecloudName) {
...
     FirecloudWorkspaceResponse fcResponse;
     FirecloudWorkspace fcWorkspace;
 
     WorkspaceResponse workspaceResponse = new WorkspaceResponse();
 
     // This enforces access controls.
-    fcResponse = fireCloudService.getWorkspace(workspaceNamespace, workspaceId);
+    fcResponse =
+        fireCloudService.getWorkspace(
+            dbWorkspace.getWorkspaceNamespace(), dbWorkspace.getFirecloudName());
```
jaycarlton: at line 241 below, I'm surprised to see us throwing a WorkbenchException exception from the service level; in principle code that's not running in a REST stack at all should be able to call this method.

api/src/main/resources/workbench.yaml
```diff
@@ -71,6 +71,21 @@ parameters:
     format: int64
     required: true
     description: Data set ID
+  workspaceNamespace:
+    in: path
+    name: workspaceNamespace
+    type: string
+    required: true
+    description: The workspace namespace, aka Google Cloud project ID.
```
jaycarlton: It would help to indicate if this the firecloud workspace namespace or an AOU one.

api/src/main/resources/workbench.yaml
```diff
@@ -71,6 +71,21 @@ parameters:
...
+  workspaceId:
+    in: path
+    name: workspaceId
+    type: string
+    required: true
+    description: >
+      The workspace ID, aka the Firecloud workspace ID. This is actually an immutable string,
+      which in AoU looks like a lower-cased concatenation of the workspace name when it was first
```
jaycarlton: _looks like_ Is there an actual contract here, or are we in danger of making an assumption about how these are generated that could change on us?

api/src/main/resources/workbench.yaml
```diff
@@ -583,88 +598,77 @@ paths:
           schema:
             $ref: '#/definitions/ErrorResponse'
 
-  /v1/clusters/{billingProjectId}/{workspaceFirecloudName}:
+  /v1/clusters/{workspaceNamespace}:
```
jaycarlton: This may be way too late, but I'm wondering if "cluster" is a concept that should be leaked outside the server. I guess we have enough specialized UI around them that it's fine.

api/src/main/resources/workbench.yaml
```diff
@@ -583,88 +598,77 @@ paths:
...
+            $ref: "#/definitions/ErrorResponse"
 
-  /v1/clusters/{clusterNamespace}/{clusterName}/localize:
+  /v1/clusters/{workspaceNamespace}/localize:
     post:
-      summary: >
-        Localize files from a workspace to notebook cluster. As a side-effect,
+      summary: Localize files to the user's cluster.
+      description: >
+        Localizes files to the cluster for the current user and given workspace. As a side-effect,
```
jaycarlton: I guess I'm still confused on when this should be called in the lifecycle of the cluster. Is it during startup (i.e. is this the thing that gives users their files back?)  Maybe also replace "localizes" with "copies", unless there's some other translation happening.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -83,12 +84,17 @@
...
 
-  private static final String BILLING_PROJECT_ID = "proj";
-  private static final String BILLING_PROJECT_ID_2 = "proj2";
+  private static final String BILLING_PROJECT_ID = "aou-rw-1234";
+  private static final String BILLING_PROJECT_ID_2 = "aou-rw-5678";
   // a workspace's namespace is always its billing project ID
   private static final String WORKSPACE_NS = BILLING_PROJECT_ID;
-  private static final String WORKSPACE_ID = "wsid";
-  private static final String WORKSPACE_NAME = "wsn";
+  // Workspace ID is also known as firecloud_name. This identifier is generated by
```
jaycarlton: This comment would be very useful in the code as well.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -148,14 +144,19 @@ Random random() {
 
   @Captor private ArgumentCaptor<Map<String, String>> mapCaptor;
 
-  @Autowired ClusterAuditor clusterAuditor;
-  @Autowired LeonardoNotebooksClient notebookService;
-  @Autowired FireCloudService fireCloudService;
+  @MockBean AdminActionHistoryDao adminActionHistoryDao;
```
jaycarlton: nit: I like to name mocked members with a `mock` prefix, especially for large test classes.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -148,14 +144,19 @@ Random random() {
...
+  @MockBean ClusterAuditor clusterAuditor;
+  @MockBean ComplianceService complianceService;
+  @MockBean DirectoryService directoryService;
+  @MockBean FireCloudService fireCloudService;
+  @MockBean LeonardoNotebooksClient notebookService;
```
jaycarlton: Is this name intentional? Looks like there's a separate `org.pmiops.workbench.notebooks.NotebooksService`.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -505,96 +490,89 @@ public void testLocalize() {
...
 
     doThrow(ForbiddenException.class)
         .when(workspaceService)
-        .enforceWorkspaceAccessLevel(WORKSPACE_NS, WORKSPACE_ID, WorkspaceAccessLevel.READER);
+        .enforceWorkspaceAccessLevel(WORKSPACE_NS, WORKSPACE_ID, WorkspaceAccessLevel.WRITER);
 
-    assertThrows(
-        ForbiddenException.class, () -> clusterController.listClusters(WORKSPACE_NS, WORKSPACE_ID));
+    assertThrows(ForbiddenException.class, () -> clusterController.getCluster(WORKSPACE_NS));
     verify(workspaceService, never()).validateActiveBilling(anyString(), anyString());
```
jaycarlton: Do we still need the first `doThrow` if we're going to verify that the method that does the throwing isn't called? Here and below.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    ]);
+    const initializer = createInitializer();
+    const cluster = await runInitializerAndTimers(initializer);
+    expect(mockStartCluster).toHaveBeenCalled();
+    expect(cluster.status).toEqual(ClusterStatus.Running);
+  });
+
+  it('should call callback with cluster status', async() => {
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Stopped},
```
jaycarlton: nit: these `mockGetClusterCalls()` invocations might be readable with a helper lambda/method like `baseWith(s: ClusterStatus): Cluster`. So
`{...baseCluster, status: ClusterStatus.Starting}` becomes `baseWith(ClusterStatus.Starting)`.

It's just a lot of `...baseCluster`, but I'm sure some of that distraction is culture shock.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    // which should trigger a creation request...
+    mockCreateCluster.mockImplementationOnce(async(workspaceNamespace) => {
+      return {...baseCluster, status: ClusterStatus.Creating};
+    });
+    // and eventually give us a good cluster.
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Starting},
+      {...baseCluster, status: ClusterStatus.Running}
+    ]);
+
```
jaycarlton: nit: I wonder if there's a way to visually separate the setup from the punchline in these longer tests. The code under test is almost an afterthought in terms of how little space it takes up.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+      throw new Response(null, {status: 503});
+    });
+    mockGetCluster.mockImplementationOnce(async(workspaceNamespace) => {
+      throw new Response(null, {status: 503});
+    });
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Running},
+    ]);
+
+    const initializer = createInitializer();
```
jaycarlton: nit: if we don't need to assert anything on the initializer, might as well inline it.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    const cluster = await runInitializerAndTimers(initializer);
+
+    expect(cluster.status).toEqual(ClusterStatus.Running);
+  });
+
+  it('Should give up after too many server errors', async() => {
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Creating},
+    ]);
+    for (let i = 0; i < 10; i++) {
```
jaycarlton: nit: I looked for a `mockImplementationNTimes()` but didn't see anything. It might not hurt to write your own if you need to do this in more than one place.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+      mockGetCluster.mockImplementationOnce(async(workspaceNamespace) => {
+        throw new Response(null, {status: 503});
+      });
+    }
+
+    const initializer = createInitializer();
+
+    expect.assertions(1);
+    return runInitializerAndTimers(initializer).catch(error => {
```
jaycarlton: How do you detect if the expected error wasn't thrown? Here and below.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    // Mock out getCluster API responses which simulate a cluster in an error state, which is then
+    // reset, but ends up in an error state again. This scenario should warrant two deleteCluster
+    // calls, but the initializer is configured to max out at 1 and should return an error.
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Error},
+      {...baseCluster, status: ClusterStatus.Deleting},
+      {...baseCluster, status: ClusterStatus.Creating},
+      {...baseCluster, status: ClusterStatus.Error},
+    ]);
+    mockDeleteCluster.mockImplementation(async(workspaceNamespace) => {
```
jaycarlton: I think you can use `mockReturnValueOnce` or `mockResolvedValue` or one of the other helpers for these common cases.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    expect(cluster.status).toEqual(ClusterStatus.Running);
+  });
+
+  it('Should give up after too many server errors', async() => {
+    mockGetClusterCalls([
+      {...baseCluster, status: ClusterStatus.Creating},
+    ]);
+    for (let i = 0; i < 10; i++) {
+      mockGetCluster.mockImplementationOnce(async(workspaceNamespace) => {
+        throw new Response(null, {status: 503});
```
jaycarlton: nit: `mockRejectedValueOnce()` looks like a more succinct way to do this.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    // There's some nuance / awkwardness to this test: the ClusterInitializer uses Date.now() to get
+    // the current timestamp, but Jest doesn't support fake clock functionality (see
+    // https://github.com/facebook/jest/issues/2684). So we just set a very quick timeout here to
+    // ensure the threshold is reached after a couple polling loops.
+    const initializer = createInitializer({overallTimeout: 30});
+
+    // Tell Jest that we plan to have 1 assertion. This ensures that the test won't
+    // pass if the promise fails.
+    expect.assertions(1);
+    return runInitializerAndTimers(initializer).catch(error => {
```
jaycarlton: If we can't change the clock, can we at least exert some more control over the timers?

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    const aborter = new AbortController();
+    const initializer = createInitializer({
+      abortSignal: aborter.signal
+    });
+
+    const initializePromise = runInitializerAndTimers(initializer);
+
+    // Wait a reasonably-short amount of time, at least one polling delay period, before sending
+    // an abort signal.
+    await new Promise(resolve => setTimeout(resolve, 20));
```
jaycarlton: Maybe make a helper out of these two lines, like a `doAfter(fn, timeoutMillis)`?

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+  });
+
+  it('Should respect the maxCreateCount option', async() => {
+    // Ensure that the initializer won't take action on a NOT_FOUND cluster if the maxCreateCount
+    // is set to disallow create requests.
+    mockGetCluster.mockImplementationOnce(async(workspaceNamespace) => {
+      throw new Response(null, {status: 404});
+    });
+    const initializer = createInitializer({maxCreateCount: 0});
+    return runInitializerAndTimers(initializer).catch(error => {
```
jaycarlton: Hmm, I don't see where return values of the `it` statements are checked. It actually looks like the return type for `it` is `void`.

ui/src/app/utils/cluster-initializer.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+    mockDeleteCluster = jest.spyOn(clusterApi(), 'deleteCluster');
+    mockStartCluster = jest.spyOn(notebooksClusterApi(), 'startCluster');
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+    jest.clearAllMocks();
+  });
+
+  const mockGetClusterCalls = (responses: Array<Cluster>) => {
```
jaycarlton: It would be simpler to have a version that takes in an array of statuses, since the base cluster props seem to be used everywhere.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
+import {notebooksClusterApi} from 'app/services/notebooks-swagger-fetch-clients';
+import {clusterApi} from 'app/services/swagger-fetch-clients';
+import {isAbortError, reportError} from 'app/utils/errors';
+import {Cluster, ClusterStatus} from 'generated/fetch';
+
+// We're only willing to wait 20 minutes total for a cluster to initialize. After that we return
+// a rejected promise no matter what.
+const DEFAULT_OVERALL_TIMEOUT = 1000 * 60 * 20;
```
jaycarlton: Can you use this `Duration` class? https://nodes.duniter.io/typescript/duniter/typedoc/interfaces/moment.duration.html

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    this.name = 'ExceededErrorCountError';
+  }
+}
+
+export interface ClusterInitializerOptions {
+  // Core options. Most callers should provide these.
+  //
+  // The workspace namespace to initialize a cluster for.
+  workspaceNamespace: string;
+  // Callback which is called every time the cluster updates its status. This callback is *not*
```
jaycarlton: Is it not desirable to call it in that situation, or just not feasible for some reason?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  // called when no cluster is found.
+  onStatusUpdate?: (ClusterStatus) => void;
+  // An optional abort signal which allows the caller to abort the initialization process, including
+  // cancelling any outstanding Ajax requests.
+  abortSignal?: AbortSignal;
+
+  // Override options. These options all have sensible defaults, but may be overridden for testing
+  // or special scenarios (such as an initialization flow which should not take any actions).
+  initialPollingDelay?: number;
```
jaycarlton: nit: these could all be Durations as well

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  // Core options. Most callers should provide these.
+  //
+  // The workspace namespace to initialize a cluster for.
+  workspaceNamespace: string;
+  // Callback which is called every time the cluster updates its status. This callback is *not*
+  // called when no cluster is found.
+  onStatusUpdate?: (ClusterStatus) => void;
+  // An optional abort signal which allows the caller to abort the initialization process, including
+  // cancelling any outstanding Ajax requests.
+  abortSignal?: AbortSignal;
```
jaycarlton: Is this like an event sink?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  maxPollingDelay?: number;
+  overallTimeout?: number;
+  maxCreateCount?: number;
+  maxDeleteCount?: number;
+  maxResumeCount?: number;
+  maxServerErrorCount?: number;
+}
+
+/**
+ * A controller class implementing client-side logic to initialize a Leonardo cluster. This class
```
jaycarlton: question: How much of this class is cluster-specific versus something that could be reused for another long-running operation?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  private resolve: (cluster?: Cluster | PromiseLike<Cluster>) => void;
+  private reject: (error: Error) => void;
+
+  constructor(options: ClusterInitializerOptions) {
+    this.workspaceNamespace = options.workspaceNamespace;
+    this.onStatusUpdate = options.onStatusUpdate ? options.onStatusUpdate : () => {};
+    this.abortSignal = options.abortSignal;
+    this.currentDelay = options.initialPollingDelay != null ? options.initialPollingDelay : DEFAULT_INITIAL_POLLING_DELAY;
+    this.maxDelay = options.maxPollingDelay != null ? options.maxPollingDelay : DEFAULT_MAX_POLLING_DELAY;
+    this.overallTimeout = options.overallTimeout != null ? options.overallTimeout : DEFAULT_OVERALL_TIMEOUT;
```
jaycarlton: FWIW, I think you can just do `this.overallTimeout = this.overallTimeout || DEFAULT_OVERALL_TIMEOUT`. Have to be careful with builtins though.

(nit): Also, I'd probably invert the condition so you're checking  `== null`, just to avoid negation wherever possible.


Finally, there's got to be some kind of builtin notion of a class/struct property with a default value.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  private async getCluster(): Promise<Cluster> {
+    return await clusterApi().getCluster(this.workspaceNamespace, {signal: this.abortSignal});
+  }
+
+  private async createCluster(): Promise<Cluster> {
+    if (this.createCount >= this.maxCreateCount) {
+      throw new ExceededActionCountError('Reached max cluster create count', this.currentCluster);
+    }
+    const cluster = await clusterApi().createCluster(this.workspaceNamespace, {signal: this.abortSignal});
+    this.createCount++;
```
jaycarlton: So if I'm reading this right, we increment before the `await` is complete? Is that what we want? Maybe it's really more of a `createAttemptCount`.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    return cluster;
+  }
+
+  private async resumeCluster(): Promise<void> {
```
jaycarlton: nit: seems odd that we get a Cluster back on create but not on resume.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  private handleUnknownError(e: any) {
+    if (e instanceof Response && e.status >= 500 && e.status < 600) {
+      this.serverErrorCount++;
+    }
+    reportError(e);
+  }
+
+  private tooManyServerErrors(): boolean {
```
jaycarlton: should this have an `is` prefix like the other predicates?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    if (this.isInitializing) {
+      throw new Error('Initialization is already in progress');
+    }
+    this.initializeStartTime = Date.now();
+
+    return new Promise((resolve, reject) => {
+      this.resolve = resolve;
+      this.reject = reject as (error: Error) => {};
+      this.poll();
+    }) as Promise<Cluster>;
```
jaycarlton: So that's interesing. `resolve` and `reject` are magically piped into the promise's callback's arguments, and the output of `resolve` is a `Cluster`?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    // Certain cluster states require active intervention, such as deleting or resuming the cluster;
+    // these are handled within the the polling loop.
+    if (this.abortSignal && this.abortSignal.aborted) {
+      // We'll bail out early if an abort signal was triggered while waiting for the poll cycle.
+      return this.reject(
+        new ClusterInitializationFailedError('Request was aborted.', this.currentCluster));
+    }
+    if (Date.now() - this.initializeStartTime > this.overallTimeout) {
+      return this.reject(
+        new ClusterInitializationFailedError(
```
jaycarlton: nit: maybe indicate the max time and the elapsed time in the message.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    // timeout threshold.
+    //
+    // Certain cluster states require active intervention, such as deleting or resuming the cluster;
+    // these are handled within the the polling loop.
+    if (this.abortSignal && this.abortSignal.aborted) {
+      // We'll bail out early if an abort signal was triggered while waiting for the poll cycle.
+      return this.reject(
+        new ClusterInitializationFailedError('Request was aborted.', this.currentCluster));
+    }
+    if (Date.now() - this.initializeStartTime > this.overallTimeout) {
```
jaycarlton: nit: breaking this chunk out into a `checkTimeout` method could make `poll()` shorter and raise its abstraction level a bit.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    this.initializeStartTime = Date.now();
+
+    return new Promise((resolve, reject) => {
+      this.resolve = resolve;
+      this.reject = reject as (error: Error) => {};
+      this.poll();
+    }) as Promise<Cluster>;
+  }
+
+  private async poll() {
```
jaycarlton: nit: is this a void function?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+      return this.reject(
+        new ClusterInitializationFailedError(
+          'Initialization attempt took longer than the max time allowed.',
+          this.currentCluster));
+    }
+
+    // Fetch the current cluster status, with some graceful error handling for NOT_FOUND response
+    // and abort signals.
+    try {
+      this.currentCluster = await this.getCluster();
```
jaycarlton: Why not have `getCluster()` call `onStatusUpdate()` itself?

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+      this.onStatusUpdate(this.currentCluster.status);
+    } catch (e) {
+      if (isAbortError(e)) {
+        return this.reject(
+          new ClusterInitializationFailedError('Abort signal received during cluster API call',
+            this.currentCluster));
+      } else if (this.isNotFoundError(e)) {
+        // A not-found error is somewhat expected, if a cluster has recently been deleted or
```
jaycarlton: Ideally Leo would expose info on recently deleted clusters (say for 7 days) so that we don't need to guess.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    // Attempt to take the appropriate next action given the current cluster status.
+    try {
+      if (this.currentCluster === null) {
+        await this.createCluster();
+      } else if (this.isClusterStopped()) {
+        await this.resumeCluster();
+      } else if (this.isClusterErrored()) {
+        // If cluster is in error state, delete it so it can be re-created at the next poll loop.
+        reportError(
+          `Cluster ${this.currentCluster.clusterNamespace}/${this.currentCluster.clusterName}` +
```
jaycarlton: nit: exposing a property for the full cluster path/fqn would be cleaner than doing it here I think.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+      } else {
+        this.handleUnknownError(e);
+        if (this.tooManyServerErrors()) {
+          return this.reject(
+            new ExceededErrorCountError('Reached max server error count', this.currentCluster));
+        }
+      }
+    }
+
+    setTimeout(() => this.poll(), this.currentDelay);
```
jaycarlton: nit: do we already have backoff delay programmed someplace? Seems odd for it to live here.

ui/src/app/utils/errors.tsx
```diff
@@ -29,7 +29,7 @@ export function reportError(err: Error) {
 
 /** Returns true if the given error is an AbortError, as used in fetch() aborts. */
 export function isAbortError(e: Error) {
-  return (e instanceof DOMException) && e.name === 'AbortError';
+  return e instanceof DOMException && e.name === 'AbortError';
```
jaycarlton: nit: fwiw I think the parentheses help here.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+ * cluster towards a running state, and will eventually resolve the Promise with a running cluster,
+ * or otherwise reject it with information about the failure mode.
+ *
+ * This is an unusually heavyweight controller class on the client side. It's worth noting a couple
+ * reasons why we ended up with this design:
+ *  - Cluster initialization can take up to 10 minutes, which is beyond the scope of a single
+ *    App Engine server-side request timeout. To reliably implement this control on the server side
+ *    would likely require new database persistence, tasks queues, and additional APIs in order to
+ *    provide the client with status updates.
+ *  - Ultimately, we might expect this type of functionality ("Get me a cluster for workspace X and
```
calbach: I'm not sure I agree with this statement. We've conceptually merged several Leo concepts into one within the AoU product, primarily on account of product decisions to avoid exposing the inner workings of the cluster state. This creates some impedance mismatches which generates complexity here. For example, we don't really expose the fact that a cluster can become paused, or that it can become deleted. If we decided to surface these things (as Terra UI probably does), it would create less business logic here.

-  Polling cannot be removed, unless cluster creation is redesigned and made instantaneous somehow. Even if we pushed the maximum amount of logic into Leo, the operation is fundamentally asynchronous.
-  Logic around when to create/start/stop a cluster at the request of a user needs to live at the product level.
-  "recreate this cluster" could potentially be a Leo operation, but since we customize Leo clusters anyways, we probably want a hand in recreating clusters anyways.

The other avenue we could explore would be to say that Leo should not have an error state, but ultimately some cluster creations will fail regardless of how many retries Leo adds. So I do think a product-level cluster error state is still a necessity.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+// a rejected promise no matter what.
+const DEFAULT_OVERALL_TIMEOUT = 1000 * 60 * 20;
+const DEFAULT_INITIAL_POLLING_DELAY = 2000;
+const DEFAULT_MAX_POLLING_DELAY = 15000;
+// By default, we're willing to retry twice on each of the state-modifying API calls, to allow
+// for transient 500s.
+const DEFAULT_MAX_CREATE_COUNT = 2;
+const DEFAULT_MAX_DELETE_COUNT = 2;
+const DEFAULT_MAX_RESUME_COUNT = 2;
+const DEFAULT_MAX_SERVER_ERROR_COUNT = 5;
```
calbach: 5 seems reasonable for consecutive errors, but might be a bit low for global error count during the polling cycle. May want to make this distinction.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+    // Overall strategy: continue polling the get-cluster endpoint, with capped exponential backoff,
+    // until we either reach our goal state (a RUNNING cluster) or run up against the overall
+    // timeout threshold.
+    //
+    // Certain cluster states require active intervention, such as deleting or resuming the cluster;
+    // these are handled within the the polling loop.
+    if (this.abortSignal && this.abortSignal.aborted) {
+      // We'll bail out early if an abort signal was triggered while waiting for the poll cycle.
+      return this.reject(
+        new ClusterInitializationFailedError('Request was aborted.', this.currentCluster));
```
calbach: opt nit: probably wouldn't put trailing punctuation in my error messages, though I don't know we have any real conventions/guidelines on this.

ui/src/app/utils/cluster-initializer.tsx
```diff
@@ -0,0 +1,301 @@
...
+  private maxDeleteCount: number;
+  private maxResumeCount: number;
+  private maxServerErrorCount: number;
+
+  // Properties to track progress, actions taken, and errors encountered.
+  private createCount = 0;
+  private deleteCount = 0;
+  private resumeCount = 0;
+  private serverErrorCount = 0;
+  private initializeStartTime?: number;
```
calbach: Per my other comment - some of these properties could be made final or eliminated if we just had a single entrypoint for initializing the cluster (rather than the two phase create object-> run once)

gjuggler:
Should be ready for another look now

ui/src/app/utils/cluster-initializer.spec.tsx
jaycarlton: It's not obfuscating if you kill 4 keywords and 2 scopes in a one-liner: `{await, new, promise, setTimeout, fat arrow}`. It's only easy to read if you like reading JS. 😉

ui/src/app/utils/cluster-initializer.tsx
jaycarlton: Yeah, the less we know about the underlying clusters the better. We shouldn't know whether we have a new or existing cluster ideally (provided we can supply config).

ui/src/app/utils/cluster-initializer.tsx
jaycarlton: 🤦‍♂ right.

ui/src/app/utils/cluster-initializer.tsx
jaycarlton: That's fine. I wrote this before I sussed out how the callbacks work, and wondered why we didn't need the result of the poll operation.

ui/src/app/utils/cluster-initializer.tsx
jaycarlton: OK, but there's also a lot of maintenance overhead if we have multiple, slightly inconsistent ways of doing basic things like retrying.

ui/src/app/utils/cluster-initializer.tsx
calbach: It does expose that information via getClusters?includeDeleted=true. However, whether or not it was deleted is irrelevant here. In either situation we're going to create a new cluster.

ui/src/app/utils/cluster-initializer.tsx
calbach: Yes makes sense, I surmised as much from the changes - thanks.

PR #3136, opened 2020-02-14
[RW-4463][RISK=NO] Interceptor-based logging

jaycarlton:
Use a simple interceptor to create a duration and record a distribution metric.

TIL:
- request attributes can hold things that you might need later, and you don't need to worry about threading (e.g. if you were thinking of using an in-class map)
- it's really powerful stuff

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/ElapsedTimeDistributionInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/ElapsedTimeDistributionInterceptor.java
@@ -1,14 +1,10 @@
 package org.pmiops.workbench.interceptors;
 
-import com.google.common.base.Stopwatch;
-import java.lang.reflect.Method;
+import com.google.api.client.http.HttpMethods;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Optional;
-import javax.inject.Provider;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/pmiops/workbench/interceptors/ElapsedTimeDistributionInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+import org.pmiops.workbench.monitoring.MeasurementBundle;
+import org.pmiops.workbench.monitoring.labels.MetricLabel;
+import org.pmiops.workbench.monitoring.views.DistributionMetric;
+import org.springframework.stereotype.Service;
+import org.springframework.web.method.HandlerMethod;
+import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+@Service
+public class ElapsedTimeDistributionInterceptor extends HandlerInterceptorAdapter {
```
blrubenstein: TOL: Since this is about tracking time, it could make sense to fold this and tracing together into a TimeTrackingInterceptor...

jaycarlton:
Yeah @Gregory Jordan <gjordan@verily.com> and I discussed this a bit. Main
reason we went this way as keeping each interceptor as simple as possible.
Ok the other hand, it would be ideal if the numbers always matched between
monitoring and tracing. I'm l kinda leaning toward the unified approach, so
there's one source of truth.

PR #3151, opened 2020-02-19
[RW-4450][Risk=no] Remove limit on SQL query in notebooks

blrubenstein:
Description:
This removes the limit for our generated SQL queries. I tested to make sure large queries fail loudly with both Python and R. R failed loudly, Python failed not as loudly, but its hard to tell if it is because my computer went to sleep or there were memory problems. I did successfully download about 43 million rows though. My test query was `SELECT * FROM condition_occurrence`

---
**PR checklist**


blrubenstein:
Yeah... unfortunately its cause we're generating the SQL, there's just a painful amount of quotes and escaped quotes.

PR #3159, opened 2020-02-19
[risk=no]refactor StorageEnumsTest and CommonStorageEnumsTest

jmthibault79:
Description:

Hopefully better documents and clarifies these tests.   Also fixes a sometimes-bug I introduced in #3131 

I intentionally did not use common code here, so as not to introduce an `/api` -> `/common-api` dependency. LMK if you disagree 

---
**PR checklist**


common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
-  public Enum<?>[] enumValues;
-
-  @Parameter(2)
-  public Function<Short, Enum<?>> fromStorage;
-
-  @Parameter(3)
-  public Function<Enum<?>, Short> toStorage;
-
-  @Test
-  public void testBijectiveStorageMapping() {
```
jmthibault79: required explicitly enumerating the maps in `data()` above

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
-          enumToShort = m;
-        }
-
-        if (m.getReturnType().equals(enumClass)) {
-          shortToEnum = m;
-        }
-      }
-
-      // stringly typed map - test with testDomainIdBijectiveStorageMapping instead
-      if (enumToShort.getName().equals("domainIdToDomain")) {
```
jmthibault79: This was the bug in #3131 - also needed to check for domainToDomainId

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -3,40 +3,68 @@
...
+   *
+   * @return
+   */
+  private Set<Class> getEnumerationClasses() {
+    return Stream.of(DbStorageEnums.class.getDeclaredFields())
+        .filter(f -> f.getType().equals(BiMap.class))
+        .map(
+            f ->
+                (Class)
+                    ((ParameterizedType) f.getAnnotatedType().getType())
```
jmthibault79: This is not my dark magic - it's a copy-paste.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -3,40 +3,68 @@
...
+        methods.stream()
+            .filter(m -> enumClasses.contains(firstMethodParameter(m)))
+            .collect(Collectors.toMap(this::firstMethodParameter, m -> m));
 
-        if (m.getReturnType().equals(enumClass)) {
-          shortToEnum = m;
-        }
-      }
+    // e.g. public static ReviewStatus reviewStatusFromStorage(Short s)
+    final Map<Class, Method> fromStorageMethods =
```
jaycarlton: This is nifty. I might need a little more detail in the comments (especially if my reflection is rusty).

Thinking out loud, I wonder if you could refactor the Enums classes such that each matching pair of to & from methods just has the same name across the board. E.g. they could all inherit from something like
```
interface EnumMappable<T extends Enum<T>> {
  T toEnum(Short short);
  short toStorage(T enum);
}
```

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -3,40 +3,68 @@
...
+        methods.stream()
+            .filter(m -> enumClasses.contains(m.getReturnType()))
+            .collect(Collectors.toMap(Method::getReturnType, m -> m));
 
+    for (Class enumClass : enumClasses) {
+      final Method toStorage = toStorageMethods.get(enumClass);
+      final Method fromStorage = fromStorageMethods.get(enumClass);
       for (Object e : enumClass.getEnumConstants()) {
-        Short shortVal = (Short) enumToShort.invoke(null, e);
+        Short shortVal = (Short) toStorage.invoke(null, e);
```
jaycarlton: Can you make an identifier for `null` in this case, like `STATIC`?

I'm confused about what `e` is here. Maybe `enumEntry` or `enumVal` would help.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -3,40 +3,68 @@
...
 
         assertThat(shortVal).named(enumClass.getName() + ":" + e.toString()).isNotNull();
-        assertThat(shortToEnum.invoke(null, shortVal)).isEqualTo(e);
+        assertThat(fromStorage.invoke(null, shortVal)).isEqualTo(e);
```
jaycarlton: This might be a place where making a local lambda makes things cleaner. E.g. `Function<Short, enumClass> fromStorage = storageValue -> fromStorage.invoke(null, storageValue);` The invoke null takes a while to get used to.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
jaycarlton: deep_magic

Can you generify it so that it takes in either a type parameter or argument for the enum class to be harvested?

I'd like to see some more comments on why it has to be copied and what each chunk is doing.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
-      assertWithMessage("unmapped enum value: " + v).that(storageValue).isNotNull();
-      assertThat(v).isEqualTo(fromStorage.apply(storageValue));
-    }
-  }
 
   // domain ID is stringly-typed so special-case this
 
   @Test
-  public void testDomainIdBijectiveStorageMapping() {
+  public void test_domainId() {
```
jaycarlton: How do we test the reverse direction? I.e. how do we know we don't have shorts written to the db someplace that don't map back to an enum value?

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -3,40 +3,68 @@
...
+        .map(
+            f ->
+                (Class)
+                    ((ParameterizedType) f.getAnnotatedType().getType())
+                        .getActualTypeArguments()[0])
+        .collect(Collectors.toSet());
+  }
+
+  // convenience method describing what is retrieved here
+  private Class firstMethodParameter(Method method) {
```
jaycarlton: Maybe this is really `getFirstParameterType()`?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
+  @Test
+  public void test_noMissingMapEntries() throws Exception {
+    final Set<Class> enumClasses = getEnumerationClasses();
+    final Collection<Method> methods = Arrays.asList(CommonStorageEnums.class.getDeclaredMethods());
+
+    // e.g. public static Short domainToStorage(Domain domain)
+    final Map<Class, Method> toStorageMethods =
+        methods.stream()
+            .filter(m -> enumClasses.contains(firstMethodParameter(m)))
+            // domainToDomainId is stringly typed - test with test_domainId
```
jaycarlton: "strongly"?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
+            .filter(m -> !m.getName().equals("domainToDomainId"))
```
jaycarlton: This test is extremely tightly tied to the implementation of the class. Ideally we capture enough of the functionality that migrating to something cleaner is easier. If we hold on too tightly, this could actually slow us down a bit.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
-      }
+  // this is a copy of StorageEnumsTest.test_noMissingMapEntries()
 
+  @Test
+  public void test_noMissingMapEntries() throws Exception {
+    final Set<Class> enumClasses = getEnumerationClasses();
+    final Collection<Method> methods = Arrays.asList(CommonStorageEnums.class.getDeclaredMethods());
+
+    // e.g. public static Short domainToStorage(Domain domain)
+    final Map<Class, Method> toStorageMethods =
```
jaycarlton: Can we name this `enumClassToToStorageMethod`?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
 import static com.google.common.truth.Truth.assertWithMessage;
```
jaycarlton: Is it worth even fiddling with `common-api` if it's about to get reabsorbed?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
```diff
@@ -4,103 +4,86 @@
...
-        assertThat(shortToEnum.invoke(null, shortVal)).isEqualTo(e);
+        assertThat(fromStorage.invoke(null, shortVal)).isEqualTo(e);
       }
     }
   }
+
+  /**
+   * Retrieve all Enum classes in CommonStorageEnums
+   *
+   * <p>our convention for BiMaps in CommonStorageEnums is <Enum, Short> so we retrieve the first
```
jaycarlton: I really feel like this test knows too much about the implementation. I should ideally be able to go in and change out how we map things and the test cases should still pass if I've done it correctly. If the test expects biMap, that won't be true.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: no, it's typed as a string

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: Perhaps it wasn't when I messed with it last week.  But now that I've done so, I have broken it and hence this fix.  Oh well.

jmthibault79:
@jaycarlton I have updated StorageEnumsTest again.  If you approve, I'll do CommonStorageEnumsTest the same way and merge.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: Can we just move (F6 in IJ) all the public methods from CommonStorageEnums to DbStorageEnums and delete CommonStorageEnums?

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: This is a sign that we're trying too hard I think.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jaycarlton: I guess by virtue of them being in a bimap it's not a problem, but the test shouldn't know that.

common-api/src/test/java/org/pmiops/workbench/db/model/CommonStorageEnumsTest.java
jmthibault79: no.  they are in common-api.

jmthibault79:
These changes are significant and would require some careful thought:
* Update domainId so it is not a special string type
* Migration from common-api to api 
* Re-implementation of enums

That's massive scope creep for a PR which is only a documentation change and a quick fix to unbreak tests.

A wholesale rewrite of the enum tests is also beyond the scope of this PR.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -1,42 +1,117 @@
...
-      for (Object e : enumClass.getEnumConstants()) {
-        Short shortVal = (Short) enumToShort.invoke(null, e);
+  // special-case test for domain ID round trip
+
+  @Test
+  public void test_domainId() {
+    for (Domain v : Domain.values()) {
+      String storageValue = CommonStorageEnums.domainToDomainId(v);
+      assertWithMessage("unmapped enum value: " + v).that(storageValue).isNotNull();
+      assertThat(v).isEqualTo(CommonStorageEnums.domainIdToDomain(storageValue));
```
jaycarlton: nit: I'd put `v` on the right, since it's the known quantity and not the experimental quantity.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -1,42 +1,117 @@
...
+      throws InvocationTargetException, IllegalAccessException {
+    final Object returnValue =
+        enumClassToStorageMethod.get(enumClass).invoke(INDICATES_STATIC_METHOD, enumValue);
+    assertThat(returnValue instanceof Short).isTrue();
+    return (Short) returnValue;
+  }
+
+  private Object storageToEnum(Short shortValue, Class enumClass)
+      throws InvocationTargetException, IllegalAccessException {
+    return storageMethodToEnumClass.get(enumClass).invoke(INDICATES_STATIC_METHOD, shortValue);
```
jaycarlton: nit: might be easier to read if you break it into the method lookup and invocation.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -1,42 +1,117 @@
...
+  }
+
+  /**
+   * Retrieve all Enum classes in DbStorageEnums and CommonStorageEnums
+   *
+   * <p>our convention for BiMaps in DbStorageEnums/CommonStorageEnums is <Enum, Short> so we
+   * retrieve the first parameterized type
```
jaycarlton: nit: you might (later) also look for bimaps that don't have these properties and make noise, but I think it's ok.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -1,42 +1,117 @@
...
+            DbStorageEnums.class.getDeclaredFields(), CommonStorageEnums.class.getDeclaredFields())
+        .flatMap(Stream::of)
+        .filter(field -> field.getType().equals(BiMap.class))
+        .map(
+            field -> {
+              // gets the specific BiMap type including annotations
+              final Type biMapType = field.getAnnotatedType().getType();
+
+              // need a ParameterizedType to access the type arguments
+              assertThat(biMapType instanceof ParameterizedType).isTrue();
```
jaycarlton: nit: ordinarily I don't expect statements with side effects inside a `map` call. It's OK here.

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -1,42 +1,117 @@
...
+  private final Object INDICATES_STATIC_METHOD = null;
 
-      Class enumClass =
-          (Class) ((ParameterizedType) f.getAnnotatedType().getType()).getActualTypeArguments()[0];
+  final Set<Class> enumClasses = getEnumerationClasses();
+  final Set<Method> methods =
+      Stream.of(
+              DbStorageEnums.class.getDeclaredMethods(),
+              CommonStorageEnums.class.getDeclaredMethods())
+          .flatMap(Stream::of)
```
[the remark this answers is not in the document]
jmthibault79: These are old-school [ ] arrays so concat can't handle them

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
jmthibault79: ok

api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
jmthibault79: ok

PR #3160, opened 2020-02-20
[RW-4464][risk=low] Plumb proper agent details to audit system

calbach:
Two-pronged fix:
- Plumb the proper agent details through higher interfaces. After looking more closely, it was clear that the default of "administrator" was incorrect in most cases here.
- Produce null (injectable as Optional) if there is no user. Longer term fix might use a different wrapper object for this. This should make the failure mode clearer and provides a pattern if we need this in the future. Longer term fix is https://precisionmedicineinitiative.atlassian.net/browse/RW-4496

Learnings:

- `Optional` has special magic semantics in Spring, it can be used to express an "optional" injected dependency, as [described here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Autowired.html).
  - This means that providing an Optional typed Bean is problematic, I couldn't get it working properly
  - ~~A `Provider<Optional>` doesn't appear to be a thing that I could get working~~
  - `Provider<Optional> works`
  - Even if your injected bean type is totally bogus, injecting `Optional` works for an empty optional
- A `Provider` is allowed to produce a null value. 
- A Bean cannot be declared as an `Optional` type, as far as I could tell.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditServiceImpl.kt
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditServiceImpl.kt
@@ -5,13 +5,13 @@ import com.google.cloud.logging.LogEntry
 import com.google.cloud.logging.Logging
 import com.google.cloud.logging.Payload.JsonPayload
 import com.google.cloud.logging.Severity
+import org.pmiops.workbench.config.WorkbenchConfig
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
 import java.util.HashMap
 import java.util.logging.Level
 import java.util.logging.Logger
 import javax.inject.Provider
-import org.pmiops.workbench.config.WorkbenchConfig
-import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.stereotype.Service
 
 @Service
 class ActionAuditServiceImpl @Autowired
@@ -47,11 +47,11 @@ constructor(private val configProvider: Provider<WorkbenchConfig>, private val c
         jsonValuesByColumn[AuditColumn.ACTION_ID.name] = auditEvent.actionId
         jsonValuesByColumn[AuditColumn.ACTION_TYPE.name] = auditEvent.actionType
         jsonValuesByColumn[AuditColumn.AGENT_TYPE.name] = auditEvent.agentType
-        jsonValuesByColumn[AuditColumn.AGENT_ID.name] = auditEvent.agentId
+        jsonValuesByColumn[AuditColumn.AGENT_ID.name] = auditEvent.agentIdMaybe
         jsonValuesByColumn[AuditColumn.AGENT_EMAIL.name] = auditEvent.agentEmailMaybe
         jsonValuesByColumn[AuditColumn.TARGET_TYPE.name] = auditEvent.targetType
         jsonValuesByColumn[AuditColumn.TARGET_ID.name] = auditEvent.targetIdMaybe
-        jsonValuesByColumn[AuditColumn.AGENT_ID.name] = auditEvent.agentId
+        jsonValuesByColumn[AuditColumn.AGENT_ID.name] = auditEvent.agentIdMaybe
         jsonValuesByColumn[AuditColumn.TARGET_PROPERTY.name] = auditEvent.targetPropertyMaybe
         jsonValuesByColumn[AuditColumn.PREV_VALUE.name] = auditEvent.previousValueMaybe
         jsonValuesByColumn[AuditColumn.NEW_VALUE.name] = auditEvent.newValueMaybe
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
@@ -154,7 +154,7 @@ constructor(
                 actionId = actionIdProvider.get(),
                 actionType = ActionType.DETECT_HIGH_EGRESS_EVENT,
                 agentType = AgentType.SYSTEM,
-                agentIdMaybe = 0,
+                agentIdMaybe = null,
                 agentEmailMaybe = null,
                 targetType = TargetType.WORKSPACE,
                 targetIdMaybe = null
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
@@ -93,7 +93,7 @@ constructor(
                 actionId = actionId,
                 actionType = ActionType.DETECT_HIGH_EGRESS_EVENT,
                 agentType = AgentType.USER,
-                agentId = agentId,
+                agentIdMaybe = agentId,
                 agentEmailMaybe = agentEmail,
                 targetType = TargetType.WORKSPACE,
                 targetIdMaybe = dbWorkspace.workspaceId
@@ -154,7 +154,7 @@ constructor(
                 actionId = actionIdProvider.get(),
                 actionType = ActionType.DETECT_HIGH_EGRESS_EVENT,
                 agentType = AgentType.SYSTEM,
-                agentId = 0,
+                agentIdMaybe = 0,
                 agentEmailMaybe = null,
                 targetType = TargetType.WORKSPACE,
                 targetIdMaybe = null
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
@@ -26,19 +26,19 @@ public class UserServiceAuditAdapterImpl implements UserServiceAuditor {
 
   private final ActionAuditService actionAuditService;
   private final Clock clock;
-  private Provider<String> actionIdProvider;
   private Provider<DbUser> dbUserProvider;
+  private Provider<String> actionIdProvider;
 
   @Autowired
   public UserServiceAuditAdapterImpl(
       ActionAuditService actionAuditService,
       Clock clock,
-      @Qualifier("ACTION_ID") Provider<String> actionIdProvider,
-      Provider<DbUser> dbUserProvider) {
+      Provider<DbUser> dbUserProvider,
+      @Qualifier("ACTION_ID") Provider<String> actionIdProvider) {
     this.actionAuditService = actionAuditService;
     this.clock = clock;
-    this.actionIdProvider = actionIdProvider;
     this.dbUserProvider = dbUserProvider;
+    this.actionIdProvider = actionIdProvider;
   }
 
   @Override
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
@@ -25,19 +25,19 @@ public class UserServiceAuditAdapterImpl implements UserServiceAuditor {
 
   private final ActionAuditService actionAuditService;
   private final Clock clock;
-  private Provider<DbUser> dbUserProvider;
   private Provider<String> actionIdProvider;
+  private Provider<Optional<DbUser>> dbUserOptionalProvider;
 
   @Autowired
   public UserServiceAuditAdapterImpl(
       ActionAuditService actionAuditService,
       Clock clock,
-      Provider<DbUser> dbUserProvider,
-      @Qualifier("ACTION_ID") Provider<String> actionIdProvider) {
+      @Qualifier("ACTION_ID") Provider<String> actionIdProvider,
+      Provider<Optional<DbUser>> dbUserOptionalProvider) {
     this.actionAuditService = actionAuditService;
     this.clock = clock;
-    this.dbUserProvider = dbUserProvider;
     this.actionIdProvider = actionIdProvider;
+    this.dbUserOptionalProvider = dbUserOptionalProvider;
   }
 
   @Override
@@ -48,7 +48,7 @@ public class UserServiceAuditAdapterImpl implements UserServiceAuditor {
       AgentType agentType) {
     // In the event of an offline cronjob, the request may have been made anonymously and there is
     // no user identity. This should only happen for system agent driven audit events.
-    Optional<DbUser> agent = Optional.ofNullable(dbUserProvider.get());
+    Optional<DbUser> agent = dbUserOptionalProvider.get();
     if (agent.isPresent() && AgentType.SYSTEM == agentType) {
       log.warning(
           String.format(
@@ -80,7 +80,7 @@ public class UserServiceAuditAdapterImpl implements UserServiceAuditor {
       long userId,
       BypassTimeTargetProperty bypassTimeTargetProperty,
       Optional<Instant> bypassTime) {
-    DbUser adminUser = dbUserProvider.get();
+    DbUser adminUser = dbUserOptionalProvider.get().get();
     ActionAuditEvent.Builder eventBuilder =
         ActionAuditEvent.builder()
             .timestamp(clock.millis())
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
@@ -8,6 +8,7 @@ import javax.inject.Provider;
 import org.pmiops.workbench.actionaudit.ActionAuditEvent;
 import org.pmiops.workbench.actionaudit.ActionAuditService;
 import org.pmiops.workbench.actionaudit.ActionType;
+import org.pmiops.workbench.actionaudit.Agent;
 import org.pmiops.workbench.actionaudit.AgentType;
 import org.pmiops.workbench.actionaudit.TargetType;
 import org.pmiops.workbench.actionaudit.targetproperties.AccountTargetProperty;
@@ -26,18 +27,18 @@ public class UserServiceAuditAdapterImpl implements UserServiceAuditor {
   private final ActionAuditService actionAuditService;
   private final Clock clock;
   private Provider<String> actionIdProvider;
-  private Provider<Optional<DbUser>> dbUserOptionalProvider;
+  private Provider<DbUser> dbUserProvider;
 
   @Autowired
   public UserServiceAuditAdapterImpl(
       ActionAuditService actionAuditService,
       Clock clock,
       @Qualifier("ACTION_ID") Provider<String> actionIdProvider,
-      Provider<Optional<DbUser>> dbUserOptionalProvider) {
+      Provider<DbUser> dbUserProvider) {
     this.actionAuditService = actionAuditService;
     this.clock = clock;
     this.actionIdProvider = actionIdProvider;
-    this.dbUserOptionalProvider = dbUserOptionalProvider;
+    this.dbUserProvider = dbUserProvider;
   }
 
   @Override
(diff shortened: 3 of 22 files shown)


api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -39,13 +42,29 @@ public UserServiceAuditAdapterImpl(
 
   @Override
   public void fireUpdateDataAccessAction(
-      DbUser targetUser, DataAccessLevel dataAccessLevel, DataAccessLevel previousDataAccessLevel) {
+      DbUser targetUser,
+      DataAccessLevel dataAccessLevel,
+      DataAccessLevel previousDataAccessLevel,
+      AgentType agentType) {
+    // In the event of an offline cronjob, the request may have been made anonymously and there is
```
jaycarlton: Maybe pull this preamble into its own helper, unless you're sure none of the other methods need to do this.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -39,13 +42,29 @@ public UserServiceAuditAdapterImpl(
...
+
     actionAuditService.send(
         ActionAuditEvent.builder()
             .timestamp(clock.millis())
-            .agentType(AgentType.ADMINISTRATOR)
-            .agentId(dbUserProvider.get().getUserId())
-            .agentEmailMaybe(dbUserProvider.get().getUsername())
+            .agentType(agentType)
+            .agentId(agent.map(DbUser::getUserId).orElse(0L))
+            .agentEmailMaybe(agent.map(DbUser::getUsername).orElse(null))
```
jaycarlton: nit: typically the pattern I use with optionals adn builders is 
`agent.ifPresent(eventBuilder::agentEmailMaybe)`. Downside is you have to declare the builder.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -61,12 +80,13 @@ public void fireAdministrativeBypassTime(
       long userId,
       BypassTimeTargetProperty bypassTimeTargetProperty,
       Optional<Instant> bypassTime) {
+    DbUser adminUser = dbUserProvider.get();
```
jaycarlton: So if we had a cron that happened to set this bypass time for some reason,  we'd. have to fix this one too.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
```diff
@@ -74,9 +75,9 @@ private boolean timestampsEqual(Timestamp a, Timestamp b) {
 
         DbUser updatedUser;
         if (workbenchConfigProvider.get().featureFlags.enableMoodleV2Api) {
-          updatedUser = userService.syncComplianceTrainingStatusV2(user);
+          updatedUser = userService.syncComplianceTrainingStatusV2(user, AgentType.SYSTEM);
```
jaycarlton: Ideally we could keep the footprint smaller, so that callers don't need to know anything about the audit system. Would injecting the class @gjuggler mentioned help here? Could come later.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
```diff
@@ -48,18 +49,28 @@
   @Bean
   @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
   public UserAuthentication userAuthentication() {
-    return (UserAuthentication) SecurityContextHolder.getContext().getAuthentication();
+    Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+    if (auth == null) {
+      return null;
```
jaycarlton: I really think null beans are an anti-pattern. I expect them to be there when they're injected, but maybe i've just. been naive. If we do. need this, then I recommend moving to `OptionalProvider`, so it's. obvious.

Can we not make UserAuthentication. and Userinfoplus a bit smarter so they don't have to be. null?

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
```diff
@@ -48,18 +49,28 @@
...
+    }
     return userAuthentication.getPrincipal();
   }
 
   @Bean
   @RequestScope(proxyMode = ScopedProxyMode.DEFAULT)
   public DbUser user(UserAuthentication userAuthentication) {
+    if (userAuthentication == null) {
+      return null;
```
jaycarlton: Could we make a fake `DbUser` instance to represent the system?

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -842,12 +851,12 @@ public DbUser syncEraCommonsStatusUsingImpersonation(DbUser user)
 
   @Override
   public void syncTwoFactorAuthStatus() {
-    syncTwoFactorAuthStatus(userProvider.get());
+    syncTwoFactorAuthStatus(userProvider.get(), AgentType.USER);
```
jaycarlton: Oh, interesting. So there are up to 3 valid values for agent type.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditorTest.java
```diff
@@ -0,0 +1,79 @@
...
+    user.setUsername("foo@gmail.com");
+    return user;
+  }
+
+  @Test
+  public void testFireUpdateDataAccessAction_userAgent() {
+    USER = createUser();
+    userServiceAuditor.fireUpdateDataAccessAction(
+        USER, DataAccessLevel.UNREGISTERED, DataAccessLevel.REGISTERED, AgentType.USER);
+    verify(mockActionAuditService)
```
jaycarlton: If this isn't working, it might be due to an incompatibility in some of the Mockito stuff. Things like captors return null when the mockito contract requires non-null. There are some examples in the kotlin test files

```
    @Test
    fun testDeleteUserProfile() {
        profileAuditAdapter!!.fireDeleteAction(USER_ID, USER_EMAIL)
        argumentCaptor<ActionAuditEvent>().apply {
            verify(mockActionAuditService).send(capture())
            val eventSent = firstValue

            assertThat(eventSent.targetType).isEqualTo(TargetType.PROFILE)
            assertThat(eventSent.agentType).isEqualTo(AgentType.USER)
            assertThat(eventSent.agentId).isEqualTo(USER_ID)
            assertThat(eventSent.targetIdMaybe).isEqualTo(USER_ID)
            assertThat(eventSent.actionType).isEqualTo(ActionType.DELETE)
//...
```

But this would be the first time we've had this problem in Java AFAIK. Does this work with standard matchers or captors?

 @ericsong thoughts?

api/src/test/java/org/pmiops/workbench/api/OfflineUserControllerTest.java
```diff
@@ -83,69 +83,73 @@ public void testBulkSyncTrainingStatusV1()
       throws org.pmiops.workbench.moodle.ApiException, NotFoundException {
     workbenchConfig.featureFlags.enableMoodleV2Api = false;
     // Mock out the service under test to simply return the passed user argument.
-    doAnswer(i -> i.getArgument(0)).when(userService).syncComplianceTrainingStatusV1(any());
+    doAnswer(i -> i.getArgument(0)).when(userService).syncComplianceTrainingStatusV1(any(), any());
```
jaycarlton: nit: prefer the tightest match, e.g. `any(AgentType.class)`, `anyString()`, etc. This helps avoid false positives when overloads come into play.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
jaycarlton: Also, if we go with this approach, it would help to mark both `userAuthentication` and the method itself with `@Nullable`.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
calbach: That's neat, didn't know about that method. In this case I think I prefer how it's currently written though.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
calbach: Without a bigger refactoring of UserService, the best I can do here is the following:
```
syncComplianceTrainingStatusV2 ->
syncComplianceTrainingStatusAsCurrentUserV2()
syncComplianceTrainingStatusAsSystemV2()
```

I'm not sure this is much better.

Information about the actor needs to originate from the callsite, since UserService is currently overloaded with calls from various actors. I think architecturally we need to draw a line where request scoped information is attenuated in order to clean this up.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
calbach: They don't currently. I think this is a symptom of this method being too heavily overloaded so I'd rather leave the wart here than bless it in a utility class.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditorTest.java
calbach: SOrry for the noise. In reality, my tests were correctly failing - but the error messaging was highly misleading. The arg matchers were working as intended.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
jaycarlton: OK. What would you. think about a small value class with the agent type, agent ID, and (optional) agent email builtin? That might save some lookups later on and ensure we don't accidentally pair a system agent with an email or invalid ID.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
jaycarlton: I agree. My favorite solution is a class like `RequestAuthenticationInfo` because

- The injectee itself is never null or empty
- It's obvious when I declare a `Provider<RequestAuthenticationInfo>` that more than one request context needs to be supported
- It can be adopted incrementally instead of as a Big Bang refactoring.

The point on failed binding is well taken, though I think on balance I much prefer that risk to allowing injected nulls.

api/src/main/java/org/pmiops/workbench/config/WebMvcConfig.java
jaycarlton: I might make more of a case for it if we were just beginning to define the notion of users and sessions, but in any case I think it's too late in the game to shoehorn one in. It would have the nice property of avoiding refactoring any injection sites, though.

api/src/test/java/org/pmiops/workbench/api/OfflineUserControllerTest.java
jaycarlton: I mainly learned it by repeated wrist slaps at my former employer, and I'm perpetuating the cycle of abuse here.

Seriously, though, my thoughts are
- `any()` works everywhere, so why did they bother adding `anyString()`, `anyLong()`, `any(MyClass.class)`?
- it helps with readability if I can see the type. Frequently that helps me understand the method signature a bit better when I read the test so I don't have to look it up.
- a false positive would be rare, but also very difficult to catch. I prefer verifying a tightly scoped & defined method.

Here are a couple of refs that show the strict forms but not the loose ones:
- https://www.journaldev.com/21876/mockito-argument-matchers-any-eq
- https://www.toptal.com/java/a-guide-to-everyday-mockito
- https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/ArgumentMatchers.html

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -20,32 +21,50 @@
...
+    if (agent.isPresent() && AgentType.SYSTEM == agentType) {
+      log.warning(
+          String.format(
+              "found unexpected request scope user '%s' for SYSTEM agent audit action",
+              agent.get().getUsername()));
+    } else if (!agent.isPresent() && AgentType.SYSTEM != agentType) {
+      log.warning(
+          String.format("missing request scope user for audit action agent type of %s", agentType));
+    }
+
```
jaycarlton: Can you please have a block for unhandled agent types? (A `switch` may be better actually)

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
jaycarlton: Would you mind going ahead and filing a spike ticket for this question and the design sketch you mentioned about `RequestAuthenticationInfo`? I agree it's not a short term priority.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditorTest.java
```diff
@@ -0,0 +1,91 @@
...
+                        && a.getAgentType() == AgentType.USER));
+  }
+
+  @Test
+  public void testFireUpdateDataAccessAction_systemAgent() {
+    userServiceAuditor.fireUpdateDataAccessAction(
+        createUser(), DataAccessLevel.UNREGISTERED, DataAccessLevel.REGISTERED, AgentType.SYSTEM);
+    verify(mockActionAuditService)
+        .send(
+            argThat(
```
jaycarlton: This is interesting. It's a bit strange at first, but looks a bit cleaner than using an `ArgumentCaptor`. Though in the latter case you can write separate assertions on the thing passed in, whereas here I suppose it's up to Mockito to format the error information.

api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java
calbach: Switched to passing a small value class. This eliminates the need for the Optional stuff entirely.

Spike is https://precisionmedicineinitiative.atlassian.net/browse/RW-4496

api/src/test/java/org/pmiops/workbench/api/OfflineUserControllerTest.java
calbach: In the case of an ambiguous overload, the compiler will actually just fail. I'm not sure it's difficult to catch at all. Certainly there are scenarios where you'd need typed parameters to distinguish these different calls, which is enough of a reason for them to exist.

The documentation is maybe nice for some, but that seems subjective and comes at the price of needing to keep those types in sync. I'd say that's a wash or net negative in my mind.

The only other argument I can produce right now is that if someone comes along and adds a new method override, that might cause a compilation failure in existing tests which use `any()`.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditorTest.java
calbach: Agreed, I switched to captor since the failure messaging was so poor here.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditServiceImpl.kt
```diff
@@ -5,13 +5,13 @@ import com.google.cloud.logging.LogEntry
 import com.google.cloud.logging.Logging
 import com.google.cloud.logging.Payload.JsonPayload
 import com.google.cloud.logging.Severity
+import org.pmiops.workbench.config.WorkbenchConfig
```
jaycarlton: Yeah, that's odd.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
```diff
@@ -154,7 +154,7 @@ constructor(
                 actionId = actionIdProvider.get(),
                 actionType = ActionType.DETECT_HIGH_EGRESS_EVENT,
                 agentType = AgentType.SYSTEM,
-                agentId = 0,
+                agentIdMaybe = 0,
```
jaycarlton: you should be able to leave this null now. Looks like the local convention is to explicitly set the optionals for clarity. Otherwise I'd generally say to leave out the args that have defaults.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
```diff
@@ -20,32 +22,37 @@
...
   private Provider<String> actionIdProvider;
+  private Provider<DbUser> dbUserProvider;
 
   @Autowired
   public UserServiceAuditAdapterImpl(
       ActionAuditService actionAuditService,
       Clock clock,
-      Provider<DbUser> dbUserProvider,
-      @Qualifier("ACTION_ID") Provider<String> actionIdProvider) {
+      @Qualifier("ACTION_ID") Provider<String> actionIdProvider,
```
jaycarlton: How do you determine the preferred argument order?

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditAdapterImpl.java
jaycarlton: Not likely in the short term I guess, but there's nothing baked in to prevent that, right? Like if we needed to do a batch access change because we added or removed an access level.

Not worth worrying about right now.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
```diff
@@ -71,21 +72,21 @@ DbUser syncComplianceTrainingStatusV1()
...
   @Deprecated
-  DbUser syncComplianceTrainingStatusV1(DbUser user)
+  DbUser syncComplianceTrainingStatusV1(DbUser user, Agent agent)
       throws org.pmiops.workbench.moodle.ApiException, NotFoundException;
 
   DbUser syncComplianceTrainingStatusV2()
       throws org.pmiops.workbench.moodle.ApiException, NotFoundException;
 
-  DbUser syncComplianceTrainingStatusV2(DbUser user)
+  DbUser syncComplianceTrainingStatusV2(DbUser user, Agent agent)
```
jaycarlton: This is much cleaner, thanks. 

Only thing that would be neater would be to inject the agent somehow, but I'm guessing that's not trivial because of threading and auth concerns.

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditorTest.java
```diff
@@ -0,0 +1,65 @@
...
+  public void testFireUpdateDataAccessAction_userAgent() {
+    DbUser user = createUser();
+    userServiceAuditor.fireUpdateDataAccessAction(
+        user, DataAccessLevel.UNREGISTERED, DataAccessLevel.REGISTERED, Agent.asUser(user));
+    verify(mockActionAuditService).send(eventArg.capture());
+
+    ActionAuditEvent event = eventArg.getValue();
+    assertThat(event.getAgentType()).isEqualTo(AgentType.USER);
+    assertThat(event.getAgentIdMaybe()).isEqualTo(user.getUserId());
+    assertThat(event.getAgentEmailMaybe()).isEqualTo(user.getUsername());
```
jaycarlton: aside: I wonder if it's worth changing agentEmail to agentUsername everywhere (except in what's sent to Stackdriver). For another day.

api/src/main/java/org/pmiops/workbench/db/dao/UserService.java
calbach: It's not trivial because it depends on service-level business logic. There is nothing inherent to a user identity which tells us whether they are acting as an admin or a user. Possibly this is something we could bind in an interceptor, based on what method is being called, but being explicit in controller methods is probably fine for now..

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/EgressEventAuditorImpl.kt
calbach: done

calbach:
Ignoring api-bigquery-test, since it's the usual flake:
```
org.pmiops.*********.cdr.ConceptBigQueryServiceTest > testGetConceptCountConceptsSaved FAILED
    expected: 2
    but was : 0
        at org.pmiops.*********.cdr.ConceptBigQueryServiceTest.testGetConceptCountConceptsSaved(ConceptBigQueryServiceTest.java:88)
```

Codacy is complaining about Kotlin. We haven't configured it for Kotline or officially accepted Kotlin as a technology for this project, so I'm also leaving that alone.

calbach:
Did one more manual test to verify both the admin-driven bypass and system driven cron update generate the expected audit logs in Stackdriver.

PR #3162, opened 2020-02-20
[risk=no][RW-4401] Help sidebar in notebooks

als364:
Screen Shot 2020-02-20 at 10 04 07 AM

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/help-sidebar.tsx
+++ b/ui/src/app/components/help-sidebar.tsx
@@ -336,18 +336,15 @@ export const HelpSidebar = fp.flow(withCurrentWorkspace(), withUserProfile())(
     sidebarContainerStyles(activeIcon, notebookStyles) {
       if (notebookStyles) {
         if (activeIcon) {
-          return {...styles.sidebarContainer, ...styles.notebookOverrides, ...styles.sidebarContainerActive}
+          return {...styles.sidebarContainer, ...styles.notebookOverrides, ...styles.sidebarContainerActive};
+        } else {
+          return {...styles.sidebarContainer, ...styles.notebookOverrides};
         }
-        else {
-          return {...styles.sidebarContainer, ...styles.notebookOverrides}
-        }
-      }
-      else {
+      } else {
         if (activeIcon) {
-          return {...styles.sidebarContainer, ...styles.sidebarContainerActive}
-        }
-        else {
-          return {...styles.sidebarContainer}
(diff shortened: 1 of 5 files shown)


ui/src/app/components/help-sidebar.tsx
```diff
@@ -328,6 +333,25 @@ export const HelpSidebar = fp.flow(withCurrentWorkspace(), withUserProfile())(
       }
     }
 
+    sidebarContainerStyles(activeIcon, notebookStyles) {
```
als364: If there's a typescriptier or more concise way to do this, let me know. I gather there is a preference for immutables (and don't know the syntax for pushing all elements onto an existing object anyways)

ui/src/app/components/help-sidebar.tsx
```diff
@@ -206,6 +210,7 @@ interface Props {
   setSidebarState: Function;
   shareFunction: Function;
   sidebarOpen: boolean;
+  notebookStyles: boolean;
```
als364: this could probably be better named - taking suggestions!

ui/src/app/components/help-sidebar.tsx
NehaBroad: how about passing the parameter notebookStyles to the style object sidebarContainer itself and return the height and top based on the parameter ? I think i might  be able to find an example in our code somewhere

ui/src/app/components/help-sidebar.tsx
```diff
@@ -328,6 +333,22 @@ export const HelpSidebar = fp.flow(withCurrentWorkspace(), withUserProfile())(
       }
     }
 
+    sidebarContainerStyles(activeIcon, notebookStyles) {
+      if (notebookStyles) {
+        if (activeIcon) {
+          return {...styles.sidebarContainer, ...styles.notebookOverrides, ...styles.sidebarContainerActive};
```
NehaBroad: Or we can also have notebookOverrides inherit sidebarContainer.

ui/src/app/components/help-sidebar.tsx
NehaBroad: Hmm all i can think of  is overrideForNotebook or isNotebook

ui/src/app/components/help-sidebar.tsx
NehaBroad: Had an offline discussion, we do not have any examples now

PR #3170, opened 2020-02-20
[RW-4418][risk=low] Add project.rb tool to delete workspaces

ericsong:
Example Usage
`./project.rb delete-workspaces --dry-run=false --delete-list-filename delete.txt --project all-of-us-rw-staging`

delete_list.txt
```
aou-rw-test-1, fcname
aou-rw-test-2, fcname2
aou-rw-test-3, fcname3
aou-rw-test-4, fcname4
```

Notes
- fetch-firecloud-user-profile was broken so I fixed that while I was in here
- High level approach (for importing a complex class like WorkspaceService) was to build the Bean directly in the tool class and manually providing the necessary dependencies and `null` for the unnecessary ones.

---
**PR checklist**

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -698,6 +698,16 @@ task exportWorkspaceData(type: JavaExec) {
   }
 }
 
+// See project.rb command: delete-workspaces
+task deleteWorkspaces(type: JavaExec) {
+  classpath = sourceSets.tools.runtimeClasspath
+  main = "org.pmiops.workbench.tools.DeleteWorkspaces"
+  systemProperties = commandLineSpringProperties
+  if (project.hasProperty("appArgs")) {
+    args Eval.me(appArgs)
+  }
+}
+
 // This task is called from a few different places:
 // - devstart.rb > load_config (used by "deploy" and "update_cloud_config" commands)
 // - directly via gradlew (used by "run-local-migrations" command)
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1222,7 +1222,14 @@ def delete_workspaces(cmd_name, *args)
   ensure_docker cmd_name, args
 
   op = WbOptionsParser.new(cmd_name, args)
+  op.opts.dry_run = true
   op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--dry_run=[dry_run]",
+      TrueClass,
+      ->(opts, v) { opts.dry_run = v},
+      "When true, print debug lines instead of performing writes. Defaults to true.")
 
   op.add_typed_option(
       "--delete-list-filename [delete-list-filename]",
@@ -1238,9 +1245,17 @@ def delete_workspaces(cmd_name, *args)
   op.parse.validate
   gcc.validate()
 
+  if op.opts.dry_run
+    common.status "DRY RUN -- CHANGES WILL NOT BE PERSISTED"
+  end
+
   flags = ([
       ["--delete-list-filename", op.opts.deleteListFilename]
   ]).map { |kv| "#{kv[0]}=#{kv[1]}" }
+  if op.opts.dry_run
+    flags += ["--dry-run"]
+  end
+  # Gradle args need to be single-quote wrapped.
   flags.map! { |f| "'#{f}'" }
 
   with_cloud_proxy_and_db(gcc) do
@@ -1250,10 +1265,12 @@ def delete_workspaces(cmd_name, *args)
   end
 end
 
+DELETE_WORKSPACES_CMD = "delete-workspaces"
+
 Common.register_command({
-    :invocation => "delete-workspaces",
+    :invocation => DELETE_WORKSPACES_CMD,
     :description => "Delete workspaces listed in given file.\n",
-    :fn => ->(*args) {delete_workspaces("delete-workspaces", *args)}
+    :fn => ->(*args) {delete_workspaces(DELETE_WORKSPACES_CMD, *args)}
 })
 
 def authority_options(cmd_name, args)
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1217,6 +1217,45 @@ Common.register_command({
     :fn => ->(*args) {export_workspace_data("export-workspace-data", *args)}
 })
 
+def delete_workspaces(cmd_name, *args)
+  common = Common.new
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--delete-list-filename [delete-list-filename]",
+      String,
+      ->(opts, v) { opts.deleteListFilename = v},
+      "File containing list of workspaces to delete.
+      Each line should contain a single workspace's namespace and firecloud name, separated by a comma
+      Example: ws-namespace-1,fc-id-1 \n ws-namespace-2,fc-id-2 \n ws-namespace-3, fc-id-3")
+
+  # Create a cloud context and apply the DB connection variables to the environment.
+  # These will be read by Gradle and passed as Spring Boot properties to the command-line.
+  gcc = GcloudContextV2.new(op)
+  op.parse.validate
+  gcc.validate()
+
+  flags = ([
+      ["--delete-list-filename", op.opts.deleteListFilename]
+  ]).map { |kv| "#{kv[0]}=#{kv[1]}" }
+  flags.map! { |f| "'#{f}'" }
+
+  with_cloud_proxy_and_db(gcc) do
+    common.run_inline %W{
+        gradle deleteWorkspaces
+       -PappArgs=[#{flags.join(',')}]}
+  end
+end
+
+Common.register_command({
+    :invocation => "delete-workspaces",
+    :description => "Delete workspaces listed in given file.\n",
+    :fn => ->(*args) {delete_workspaces("delete-workspaces", *args)}
+})
+
 def authority_options(cmd_name, args)
   op = WbOptionsParser.new(cmd_name, args)
   op.opts.remove = false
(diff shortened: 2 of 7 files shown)


api/build.gradle
```diff
@@ -698,6 +698,16 @@ task exportWorkspaceData(type: JavaExec) {
   }
 }
 
+// See project.rb command: delete-workspaces
```
jaycarlton: Can I also call this with gradle directly?

api/libproject/devstart.rb
```diff
@@ -1217,6 +1217,45 @@ def export_workspace_data(cmd_name, *args)
...
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
+      "--delete-list-filename [delete-list-filename]",
+      String,
+      ->(opts, v) { opts.deleteListFilename = v},
+      "File containing list of workspaces to delete.
+      Each line should contain a single workspace's namespace and firecloud name, separated by a comma
+      Example: ws-namespace-1,fc-id-1 \n ws-namespace-2,fc-id-2 \n ws-namespace-3, fc-id-3")
+
```
jaycarlton: Could you add a dry-run option that verifies the file list and logs whether each workspace exists or not?

api/libproject/devstart.rb
```diff
@@ -1217,6 +1217,45 @@ def export_workspace_data(cmd_name, *args)
...
+  gcc = GcloudContextV2.new(op)
+  op.parse.validate
+  gcc.validate()
+
+  flags = ([
+      ["--delete-list-filename", op.opts.deleteListFilename]
+  ]).map { |kv| "#{kv[0]}=#{kv[1]}" }
+  flags.map! { |f| "'#{f}'" }
+
+  with_cloud_proxy_and_db(gcc) do
```
jaycarlton: aside: I'm surprised this function doesn't pass anything to the block.

api/libproject/devstart.rb
```diff
@@ -1217,6 +1217,45 @@ def export_workspace_data(cmd_name, *args)
...
+    common.run_inline %W{
+        gradle deleteWorkspaces
+       -PappArgs=[#{flags.join(',')}]}
+  end
+end
+
+Common.register_command({
+    :invocation => "delete-workspaces",
```
jaycarlton: Can we define a constant for the string "delete-workspaces"?

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
```diff
@@ -45,6 +45,8 @@
 
   DbWorkspace saveWithLastModified(DbWorkspace workspace);
 
+  void deleteWorkspace(DbWorkspace dbWorkspace);
```
jaycarlton: If it's not too tricky, I'd probably return some kind of status for whether the workspace was successfully deleted, wasn't there in the first place, or was otherwise not deletable. Might want to capture this on both our side and the firecloud side.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -369,24 +369,15 @@ private void setDbWorkspaceFields(
   public ResponseEntity<EmptyResponse> deleteWorkspace(
       String workspaceNamespace, String workspaceId) {
     recordOperationTime(
-        () -> deleteWorkspaceImpl(workspaceNamespace, workspaceId), "deleteWorkspace");
+        () -> {
```
jaycarlton: BTW, most of these are coming out soon in favor of the interceptor approach. I'm keeping them in place for a week or so since they're one release ahead and I need to get automation working on the logs-based stuff.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -369,24 +369,15 @@ private void setDbWorkspaceFields(
...
+          DbWorkspace dbWorkspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
+          workspaceService.deleteWorkspace(dbWorkspace);
+          workspaceAuditor.fireDeleteAction(dbWorkspace);
```
jaycarlton: If you want, you could also inject the auditor into the service and fire from there. This is the only controller that has one I think, and that was due to how it was structured at the time. It's fine here though.

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
```diff
@@ -0,0 +1,175 @@
...
+          .desc(
+              "File containing list of workspaces to delete. Each line should contain a single workspace's namespace and firecloud name, separated by a comma"
+                  + "Example: ws-namespace-1,fc-id-1 \n ws-namespace-2,fc-id-2 \n ws-namespace-3, fc-id-3")
+          .required()
+          .hasArg()
+          .build();
+
+  private static Options options = new Options().addOption(deleteListFilename);
+
+  @Bean
```
jaycarlton: Was this easier than simply implementing a batch delete API endpoint? I'd rather have swagger boilerplate than have to hack the service injection like this.

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
```diff
@@ -0,0 +1,175 @@
...
+  public WorkspaceService workspaceService(
+      FireCloudService fireCloudService,
+      Clock clock,
+      WorkspaceDao workspaceDao,
+      UserRecentWorkspaceDao userRecentWorkspaceDao,
+      Provider<DbUser> dbUserProvider) {
+    return new WorkspaceServiceImpl(
```
jaycarlton: future: if we determine a useful core of the workspaceService we could break that into its own service and inject the whole thing.

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
```diff
@@ -0,0 +1,175 @@
...
+        null,
+        null);
+  }
+
+  static DbUser currentImpersonatedUser;
+
+  @Bean
+  @Primary
+  @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+  DbUser user() {
```
jaycarlton: This user will actually show up in the audit logs. @calbach and I have been [discussing several angles](https://github.com/all-of-us/workbench/pull/3160) to this problem (i.e. attributing the agent correctly with an anonymous user), and we considered but ultimately shied away from a placeholder user.

How would you feel about providing the username of the devops person running the tool on the command line and looking them up so that we have an actual, helpful user ID & email address in the log?

api/tools/src/main/java/org/pmiops/workbench/tools/FetchFireCloudUserProfile.java
```diff
@@ -23,7 +25,7 @@
  * domain-wide delegation to make FireCloud API calls impersonating other users.
  */
 @Configuration
-@ComponentScan("org.pmiops.workbench.firecloud")
+@Import({FireCloudServiceImpl.class, FireCloudConfig.class})
```
jaycarlton: 👏  `ComponentScan` assumes a lot about our package management discipline.

api/build.gradle
ericsong: Wouldn't recommend it. You need the cloud sql database proxy that project.rb will set up for you.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
ericsong: my preference is to handle failures through exceptions

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
ericsong: I left it in the controller because it seemed like that was the level where all of the audit actions were being fired and it made sense to me if the intent is to track user actions

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
ericsong: For me, yes. Adding it as an admin endpoint would involve adding a one-off to our API, adding a new authority role, and writing some front end to fire the endpoint. Writing the calls in ruby/node isn't trivial either since we don't have API client packages for either and we'd have to figure out an authentication workflow to fire them.

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
ericsong: I don't think we're logging the delete action ATM.

api/libproject/devstart.rb
ericsong: good idea

api/build.gradle
jaycarlton: interesting.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceService.java
jaycarlton: But if you're deleting 1000 workspaces and 3 of them fail, an exception at the outer level isn't that helpful. Since we're in an eventually-consistent-ish system, I think you have to consider partial success cases (especially if it's not all in one transaction).

api/tools/src/main/java/org/pmiops/workbench/tools/DeleteWorkspaces.java
jaycarlton: I know it feels heavy, but some of these are one-time things that will pay off.

For example, if we don't already have an administrative/devops authority, that's something I'll be wanting. If you want to use Ruby, I've already got a ticket to start generating a client library for that at https://precisionmedicineinitiative.atlassian.net/browse/RW-4499. But really, couldn't you just use `curl` for this?

The other consideration is that we already have testing patterns for this kind of thing, but with the tools, it's less clear how we make sure they keep working if they're infrequently used.

PR #3176, opened 2020-02-24
[RW-4484][RW-4485][risk=no] Account creation bugfixes

gjuggler:
This PR fixes a couple issues with account creation that Eric uncovered:

1) When switching between steps, the window scroll position remains unchanged. It should scroll to the top of each new step.

The solution here was to add a componentDidUpdate method in sign-in.tsx, which resets the scroll position whenever a new step is reached.

2) Switching back and forth from account-creation to demographic survey appeared to clear out some existing data entry.

We do seem to be correctly persisting the Profile object between account creation steps, but the demographic survey component was missing some logic to auto-populate checkboxes with existing profile data. I added some unit tests to demonstrate the new behavior.

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/login/account-creation/account-creation-survey.tsx
+++ b/ui/src/app/pages/login/account-creation/account-creation-survey.tsx
@@ -62,7 +62,7 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
     super(props);
     this.state = {
       creatingAccount: false,
-      profile: this.props.profile,
+      profile: {...this.props.profile},
     };
   }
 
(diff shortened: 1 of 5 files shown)


ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -131,30 +135,31 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
...
+      <DropDownSection data-test-id='dropdown-ethnicity'
+                       header='Ethnicity' options={AccountCreationOptions.ethnicity}
                        value={demographicSurvey.ethnicity}
                        onChange={(e) => this.updateDemographicAttribute('ethnicity', e)}/>
       <Section header='Do you identify as lesbian, gay, bisexual, transgender, queer (LGBTQ),
 or another sexual and/or gender minority?'>
         <FlexColumn>
           <FlexRow style={{alignItems: 'baseline'}}>
-            <RadioButton onChange={
+            <RadioButton data-test-id='radio-lgbtq-yes' id='identifiesAsLgbtqYes' onChange={
```
ericsong: nit: I believe we use dash? casing for most of our HTML id/class names.

gjuggler: Yep – fixed.

PR #3177, opened 2020-02-24
[no ticket][risk=no] Implemented CriteriaMapper

freemabd:
- Implemented CriteriaMapper and test case
- Refactored DbCriteria to abstract out builder class and updated referencing test cases
- Removed old builder methods from DbCriteria


freemabd:
> Looks good overall. Just curious, the code looks pretty similar where you switched from using `TO_CLIENT_CRITERIA` function to the CriteriaMapper. What's the main benefit of switching?

Well it abstracts this code into a reusable class and automatically maps properties with the same property names, so you only specify names that don't match.

PR #3179, opened 2020-02-24
[risk=low][RW-4423] Include researcher contact email in RDR export

als364:
Added a proper unit test. Cleaned up a typo and some other slight jank in `RdrExportServiceImpl`.

--

**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
@@ -1,5 +1,6 @@
 package org.pmiops.workbench.rdr;
 
+import java.math.BigInteger;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -409,7 +405,7 @@ RdrWorkspaceDemographic toRdrWorkspaceDemographics(
    * @param entity
    * @param idList
    */
-  private void updateDBRdrExport(RdrEntity entity, List<Long> idList) {
+  public void updateDBRdrExport(RdrEntity entity, List<Long> idList) {
```
gjuggler: nit: while you're here, I'd fix the CamelCase in the method name: should be updateDbRdrExport (per Google's guidelines which I always link to, https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)

gjuggler: Is this worth a @VisibleForTesting annotation?

api/src/test/java/org/pmiops/workbench/rdr/RdrExportServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+  private static final Timestamp NOW_TIMESTAMP = Timestamp.from(NOW);
+  private static final FakeClock CLOCK = new FakeClock(NOW, ZoneId.systemDefault());
+
+  private DbUser dbUserWithEmail;
+  private DbUser dbUserWithoutEmail;
+
+  @TestConfiguration
+  @Import({RdrExportServiceImpl.class})
+  @MockBean({
+      RdrApi.class,
```
gjuggler: Some of these (RdrApi, RdrExportDao, and UserDao) should be unnecessary since you use the inline @MockBean annotation above (which I tend to prefer, as it directly associates it with an instance variable if needed). In fact, I tend to use only the inline version, even for mock beans that aren't directly used in the test (since it avoids creating a situation where some mockbeans are defined in one place, and some in the other place).

I'd be happy to hear reasons for preferring a different approach, though... I'm definitely still honing my Spring test aesthetic.

api/src/test/java/org/pmiops/workbench/rdr/RdrExportServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.pmiops.workbench.workspaces.WorkspaceService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+public class RdrExportServiceImplTest {
```
gjuggler: Thanks for adding a test case! I know this was a TODO that @NehaBroad had been tracking, I'm sure we can add to it over time, but usually the biggest hurdle is just creating a test file with some basic assertions :)

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
als364: 👍

als364: completely forgot about this annotation. thank you!

api/src/test/java/org/pmiops/workbench/rdr/RdrExportServiceImplTest.java
als364: No reason - still don't have a good sense of how to make tests work in Spring. I'll remove from this declaration the beans I have a handle for - if I pull all of them out someone will delete them for being unused and then wonder why the test doesn't work...

PR #3181, opened 2020-02-24
Early draft for pattern review: render sub-functions

jmthibault79:
Description:

I'm updating the Profile page to be agnostic to whether the user creation happened under the old (Demographics Survey page) Institutional Affiliation flow or the new Verified Institutional Affiliation flow.  It seems natural to me to break up the huge render() function into sub-functions, but I don't see this pattern a lot in our code.  Is there a reason for this?  LMK what you think about `renderInstitutionalAffiliationComponents()` and friends.

Some discussion on this topic which I found after a quick Google and will read now:
https://medium.com/front-end-weekly/react-quick-tip-split-up-your-render-method-5eb30e432e2
https://www.barrymichaeldoyle.com/sub-rendering/
https://silvenon.com/blog/dont-split-render-into-functions

I'll look at comments on other code here but that's not the primary purpose of this Draft.  Thanks!

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/profile/profile-page.tsx
+++ b/ui/src/app/pages/profile/profile-page.tsx
@@ -12,12 +12,12 @@ import {TooltipTrigger} from 'app/components/popups';
 import {SpinnerOverlay} from 'app/components/spinners';
 import { getRegistrationTasksMap } from 'app/pages/homepage/registration-dashboard';
 import { ProfileRegistrationStepStatus } from 'app/pages/profile/profile-registration-step-status';
-import {profileApi} from 'app/services/swagger-fetch-clients';
+import {institutionApi, profileApi} from 'app/services/swagger-fetch-clients';
 import colors, {colorWithWhiteness} from 'app/styles/colors';
 import {reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {serverConfigStore} from 'app/utils/navigation';
 import {environment} from 'environments/environment';
-import {Profile} from 'generated/fetch';
+import {Institution, InstitutionalAffiliation, InstitutionalRole, Profile} from 'generated/fetch';
 
 const styles = reactStyles({
   h1: {
@@ -43,6 +43,13 @@ const styles = reactStyles({
     backgroundColor: colors.white,
     borderRadius: 8,
     padding: 21
+  },
+  uneditableProfileElement: {
+    paddingLeft: '0.5rem',
+    marginRight: 20,
+    marginBottom: 20,
+    height: '1.5rem',
+    color: colors.primary
   }
 });
 
@@ -64,7 +71,7 @@ const validators = {
 
 export const ProfilePage = withUserProfile()(class extends React.Component<
   { profileState: { profile: Profile, reload: Function } },
-  { profileEdits: Profile, updating: boolean }
+  { profileEdits: Profile, updating: boolean, verifiedInstitution?: Institution }
 > {
   static displayName = 'ProfilePage';
 
@@ -73,13 +80,23 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
 
     this.state = {
       profileEdits: props.profileState.profile || {},
-      updating: false
+      updating: false,
+      verifiedInstitution: undefined
     };
   }
 
   navigateToTraining(): void {
     window.location.assign(
       environment.trainingUrl + '/static/data-researcher.html?saml=on');
+  }
+
+  async componentDidMount() {
+    const {profileState: {profile: {verifiedInstitutionalAffiliation}}} = this.props;
+    if (verifiedInstitutionalAffiliation) {
+      await institutionApi().getInstitution(verifiedInstitutionalAffiliation.institutionShortName).then(institution =>
+        this.setState({verifiedInstitution: institution})
+      );
+    }
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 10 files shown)


ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -157,6 +176,85 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
...
+            {institutionalRoleEnum === InstitutionalRole.OTHER ? institutionalRoleOtherText : institutionalRoleEnum}
+            </div>
+          </FlexColumn>
+        </FlexRow>
+      </React.Fragment>;
+    }
+
+    function renderOldInstitutionalAffiliationComponents(componentHandle: React.Component) {
+      return <React.Fragment>
+        <div style={{...styles.h1, marginBottom: 24}}>Institution Affiliations</div>
```
jmthibault79: Moved from the existing `render()` with `this.setState()` changed to `componentHandle.setState()`

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -82,6 +90,15 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
       environment.trainingUrl + '/static/data-researcher.html?saml=on');
   }
 
+  async componentDidMount() {
+    const {profileState: {profile: {verifiedInstitutionalAffiliation}}} = this.props;
+    if (verifiedInstitutionalAffiliation) {
+      await institutionApi().getInstitution(verifiedInstitutionalAffiliation.institutionShortName).then(institution =>
```
gjuggler: File this under "unrelated nitpicks you reserved the right to ignore", but IMO this might be cleaner if you do the data join on the backend and added a new optional read-only field in the Profile API response.

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -157,6 +176,85 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
       </div>;
     };
 
+    function renderInstitutionalAffiliationComponents(componentHandle: React.Component) {
```
gjuggler: I would definitely annotate each of these functions with the return type to enforce that they need to return some sort of React.Fragment as part of their contract.

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -157,6 +176,85 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
...
+      if (requireInstitutionalVerification) {
```
gjuggler: The logic here could probably use a bit of cleanup. Once we're no longer rendering everything in one big TSX template, I usually tend to favor explicit control flow. You could also defer evaluating this.profile into the helper method, so this method becomes simply:

const {requireInstitutionalVerification} = this.configStore.getValue();
if (requireInstitutionalVerification) {
  return this.renderVerifiedInstitutionalAffiliationComponents();
} else {
  return this.renderOldInstitutionalAffiliationComponents();
}

ui/src/app/pages/profile/profile-page.tsx
jmthibault79: If we include the Institute inline in the profile it would look like this, which feels unwieldy to me:
```
{ 
...
  verifiedInstitutionalAffiliation: {
    shortName: "Broad",
    displayName: "The Broad Institute",
    emailDomains: [ "broad.org", "broad.edu", ... ],
    emailAddress: ["alice@broad-collabs.net", "bob@sanger.uk", "clyde@vumc.org", ..]
  }
}
```

vs:
```
{
...
  verifiedInstitutionalAffiliation: "Broad"
}
```

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -157,6 +177,89 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
...
 
+    function renderInstitutionalAffiliationComponents(componentHandle: React.Component): ReactFragment {
+      if (requireInstitutionalVerification) {
+        return renderVerifiedInstitutionalAffiliationComponents();
+      } else {
+        return renderOldInstitutionalAffiliationComponents(componentHandle);
+      }
+    }
+
+    function renderVerifiedInstitutionalAffiliationComponents(): ReactFragment {
```
calbach: This looks like a React function component - may want to try just formalizing it as such and see how that compares:
https://fettblog.eu/typescript-react/components/

If you do that, for small cases like this I would probably inline the prop type rather than defining an interface as in the above code example.

ui/src/app/pages/profile/profile-page.tsx
gjuggler: I agree, that's probably too much. But if all we're using in the UI is the display name, you might just pluck that field out and copy it into the Profile response (and it may be useful elsewhere in our UI as well).

ui/src/app/pages/profile/profile-page.tsx
jmthibault79: I have been unable to get that to work

calbach: Here's an example of it being used in our codebase: https://github.com/all-of-us/workbench/blob/dfc7f2a507fe5128b992f73079ada7d494ef81ed/ui/src/app/icons/card-menu-icon.tsx#L4

calbach: Hmm, perhaps a bad example since it isn't actually referenced from anywhere.. Presumably it worked at some point..

ui/src/app/pages/profile/profile-page.tsx
jmthibault79: Done.  Easy enough with MapStruct: https://github.com/all-of-us/workbench/pull/3181/commits/3f629454e746693c97f8de17b5c29b8a2d9edc3d

ui/src/app/pages/profile/profile-page.tsx
calbach: I just remembered this discussion - I wound up using a function component in a recent PR while you were out: https://github.com/all-of-us/workbench/blob/master/ui/src/app/pages/workspace/research-purpose.tsx#L57

PR #3184, opened 2020-02-24
[risk=no][RW-4402] remove references to free tier time limits

als364:
Marking the columns deprecated for now, will delete after this goes out.

The unit tests on this seem pretty comprehensive. Will test manually tomorrow.

---

**PR checklist**


api/src/main/resources/emails/dollarthresholdemail/content.html
```diff
@@ -30,8 +30,8 @@
 <img src="${HEADER_IMG}">
 <div>Hello ${FIRST_NAME} ${LAST_NAME},</div>
 <div>This is a reminder that you have used ${USED_CREDITS} in free credits offered by the All of Us Program.
-  The remaining balance on your free credits is ${CREDIT_BALANCE} which will be valid until ${EXPIRATION_DATE}.</div>
-<div>Once your free credits expire, you can request for an extension of free credits by contacting support or
+  The remaining balance on your free credits is ${CREDIT_BALANCE}.</div>
+<div>Once you use all of your free credits, you can request for an extension of free credits by contacting support or
```
jmthibault79: I agree with the semantic change, but check w/ Karthik

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -131,15 +124,11 @@ public void checkFreeTierBillingUsage_exceedsDollarThresholds() throws Messaging
     final DbUser user = createUser(SINGLE_WORKSPACE_TEST_USER);
     createWorkspace(user, SINGLE_WORKSPACE_TEST_PROJECT);
 
-    final short daysLimit = 1000;
-    workbenchConfig.billing.defaultFreeCreditsDaysLimit = daysLimit;
-
     // set an arbitrary registration time
```
jmthibault79: Free tier alerting for cost has (or had?) a dependency on registration time which is no longer appropriate.  Removing this block would be a good test to see if we still have that.  IIRC there are other similar blocks in this file, which we should also remove.

als364: We _do_ also use it to make sure we don't cost-check every single user, only users that are already registered.

PR #3189, opened 2020-02-26
[RW-4488][risk=no] Add workspace-specific routes for WS admin page

gjuggler:
See the Jira ticket for a bit more context. We want to be able to have URLs of the form

http://researchallofus.org/admin/workspaces/aou-rw-0a95bc0e

that will land on the admin page for that given workspace.

It was a pretty straightforward task to split the view into a search & detail component. Down the line, we can expand the search component to actually search through workspaces & show a summary table before someone digs into any of the detailed pages.

Screenshots:

Screen Shot 2020-02-26 at 10 02 25 AM

Screen Shot 2020-02-26 at 10 02 20 AM

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/admin/admin-workspace-search.tsx
+++ b/ui/src/app/pages/admin/admin-workspace-search.tsx
@@ -1,0 +1,62 @@
+import {Component} from '@angular/core';
+import {Button} from 'app/components/buttons';
+import {FlexRow} from 'app/components/flex';
+import {TextInput} from 'app/components/inputs';
+import colors from 'app/styles/colors';
+import {ReactWrapperBase, UrlParamsProps, withUrlParams} from 'app/utils';
+import {navigate} from 'app/utils/navigation';
+import * as React from 'react';
+
+interface State {
+  workspaceNamespace: string;
+}
+
+class AdminWorkspaceSearchImpl extends React.Component<UrlParamsProps, State> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      workspaceNamespace: '',
+    };
+  }
+
+  navigateToWorkspace() {
+    navigate(['/admin/workspaces/' + this.state.workspaceNamespace]);
+  }
+
+  render() {
+    return <FlexRow style={{justifyContent: 'flex-start', alignItems: 'center', marginTop: '1rem'}}>
+        <label style={{color: colors.primary, marginRight: '1rem'}}>Workspace namespace</label>
+        <TextInput
+            style={{
+              width: '10rem',
+              marginRight: '1rem'
+            }}
+            onChange={value => this.setState({workspaceNamespace: value})}
+            onKeyDown={(event: KeyboardEvent) => {
+              if (event.key === 'Enter') {
+                this.navigateToWorkspace();
+              }
+            }}
+        />
+        <Button
+            style={{height: '1.5rem'}}
+            onClick={() => this.navigateToWorkspace()}
+        >
+          Load Workspace
+        </Button>
+      </FlexRow>;
+  }
+}
+
+export const AdminWorkspaceSearch = withUrlParams()(AdminWorkspaceSearchImpl);
+
+@Component({
+  template: '<div #root></div>'
+})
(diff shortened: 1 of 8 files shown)


ui/src/app/pages/admin/admin-workspace-search.tsx
```diff
@@ -0,0 +1,62 @@
...
+class AdminWorkspaceSearchImpl extends React.Component<UrlParamsProps, State> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      workspaceNamespace: '',
+    };
+  }
+
+  navigateToWorkspace() {
```
als364: This I object to more than the above comment. Navigating away from the search bar means that you have to go back in order to search a second thing. I would strongly prefer to keep the search bar and do something with state that forces a fetch of the workspace information when the router is sent to `admin/workspaces/{namespace}`.

ui/src/app/pages/admin/admin-workspace-search.tsx
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    return <FlexRow style={{justifyContent: 'flex-start', alignItems: 'center', marginTop: '1rem'}}>
+        <label style={{color: colors.primary, marginRight: '1rem'}}>Workspace namespace</label>
+        <TextInput
+            style={{
+              width: '10rem',
+              marginRight: '1rem'
+            }}
+            onChange={value => this.setState({workspaceNamespace: value})}
+            onKeyDown={(event: KeyboardEvent) => {
```
als364: ty for keeping the enter-to-search though.

ui/src/app/pages/admin/admin-workspace-search.tsx
gjuggler: You can easily return to the search form via the browser "back" command.

I do agree that this isn't an ideal UX, but it's an intermediate step on the way to a longer-term goal, where the "search" page actually shows search results, and then you click into those results to open up a new tab with each workspace's detail view. I think that setup would address your concerns, but we can chat offline if you're worried it won't.

ui/src/app/pages/admin/admin-workspace-search.tsx
gjuggler: I would die if this went away. Other things that I am unreasonably bothered by: not being able to ctrl-click links to open in a new tab (basically all of our in-app links don't support this), and not being able to cmd-enter to perform primary action on modal dialogs. It may not happen soon, but eventually I will fix all these things :)

gjuggler:
> Is there no way to keep the search bar component and the workspace-specific routes at the same time?

I didn't try super hard, but maybe there's a way to have the search bar component be a "wrapper" component with a persistent UI, and the workspace admin component as a "child". But it wasn't obvious enough of a win to me to dig further into this, especially since that whole setup is heavily Angular-based and may change dramatically when we finally wrap up the migration.

als364:
> that whole setup is heavily Angular-based

I wonder if it's a holdover from a Backbone-ish school of thought... that also had a heirarchy of components that started themselves when dropped into the DOM... 

Anyways, if it would be too much work to implement here, I do prefer having the ability to go directly to the workspace page from whatever opsgenie gives us to having the persistent search bar. As long as you file a spike ticket or something to follow up on this.

ui/src/app/pages/admin/admin-workspace-search.tsx
als364: This proooobably would. Especially if we want to do different kinds of search, like workspaces-by-problem-user or something.

gjuggler:
> > that whole setup is heavily Angular-based
> 
> I wonder if it's a holdover from a Backbone-ish school of thought... that also had a heirarchy of components that started themselves when dropped into the DOM...
> 
> Anyways, if it would be too much work to implement here, I do prefer having the ability to go directly to the workspace page from whatever opsgenie gives us to having the persistent search bar. As long as you file a spike ticket or something to follow up on this.

Thanks – filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4515 for follow-up on this flow.

PR #3190, opened 2020-02-26
[risk=low] Back end updates for RW-4160 which we need for RW-4161

jmthibault79:
Description:

RW-4160 (Inst Affil user model) did not quite match the needs of RW-4161 (Inst Affil UI).  This is why we do API and UI work together!

I'm more confident in this version because my UI PR actually uses it.  This PR is **_included_** in #3194

The current endpoints should be restricted to INSTITUTION_ADMIN users, because they reveal email patterns, which are sensitive.  Non-admin users do not need any direct access, because the relevant details (shortName and displayName) for their Verified Institutional Affiliation are available as part of the V.I.A. model.  Instead, we need a non-authenticated endpoint to list all institutions without emails, to enable institution selection as part of the signup process.  

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
```diff
@@ -34,6 +35,7 @@
   }
 
   @Override
+  @AuthorityRequired({Authority.INSTITUTION_ADMIN})
```
jmthibault79: Regular users should not be able to see the Inst's list of approved individual emails

api/src/main/java/org/pmiops/workbench/db/model/DbStorageEnums.java
```diff
@@ -344,7 +344,7 @@ public static Short organizationTypeToStorage(OrganizationType organizationType)
           .put(InstitutionalRole.LATE_CAREER, (short) 5)
           .put(InstitutionalRole.PRE_DOCTORAL, (short) 6)
           .put(InstitutionalRole.POST_DOCTORAL, (short) 7)
-          .put(InstitutionalRole.PI, (short) 8)
+          .put(InstitutionalRole.SENIOR_RESEARCHER, (short) 8)
```
jmthibault79: OK Greg you were right

api/src/main/java/org/pmiops/workbench/institution/PublicInstitutionDetailsMapper.java
```diff
@@ -0,0 +1,10 @@
+package org.pmiops.workbench.institution;
+
+import org.mapstruct.Mapper;
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.model.PublicInstitutionDetails;
+
+@Mapper(componentModel = "spring")
+public interface PublicInstitutionDetailsMapper {
+  PublicInstitutionDetails dbToModel(DbInstitution dbObject);
```
jmthibault79: Don't even need to specify the mappings!
This is because it's a subset and the names are the same.

api/src/main/resources/workbench.yaml
```diff
@@ -4407,6 +4426,28 @@ definitions:
...
+        type: string
+        description: A short unique name for the Institution used as an identifier, such as 'VUMC'
+      displayName:
+        type: string
+        description: A more descriptive name for the Institution, such as 'Vanderbilt University Medical Center'
+      organizationTypeEnum:
+        $ref: '#/definitions/OrganizationType'
+        description: >
+          The Organization Type of this institution if it is one of the enumerated
+          OrganizationTypes, or OrganizationType.OTHER if it is not
```
jmthibault79: well.  OTHER is still one of these types.  Maybe I should just remove this.

api/src/test/java/org/pmiops/workbench/api/ClusterControllerTest.java
```diff
@@ -108,7 +109,8 @@
     ClusterController.class,
     UserServiceImpl.class,
     InstitutionServiceImpl.class,
-    InstitutionMapperImpl.class
+    InstitutionMapperImpl.class,
+    PublicInstitutionDetailsMapperImpl.class
```
jmthibault79: These are ultimately due to the inclusion of `UserService`

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -446,6 +448,7 @@ public void testMe_verifiedInstitutionalAffiliation() {
     final VerifiedInstitutionalAffiliation verifiedInstitutionalAffiliation =
         new VerifiedInstitutionalAffiliation()
             .institutionShortName(broad.getShortName())
+            .institutionDisplayName(broad.getDisplayName())
```
jmthibault79: yes there are a few actual test changes down here

api/src/main/resources/workbench.yaml
```diff
@@ -4407,6 +4426,28 @@ definitions:
         items:
           type: string
 
+  PublicInstitutionDetails:
+    type: object
+    description: >
+      Represents an institution which has been approved to validate researchers who wish to use the system.
+      Does not include the sensitive email patterns used to validate users.
+    required:
+      - shortName
```
blrubenstein: Could this contain an institution with minimal details? It feels like there's some overlap which implies we could build a hierarchy?

PR #3197, opened 2020-02-27
[RW-4246][RISK=LOW] Captcha requirement at the time of Registration

NehaBroad:
Screen Shot 2020-02-27 at 2 22 28 PM

Adds Captcha Checkbox at the end of demographic survey page and do
client validation and server validation

**PR checklist**

Story for Prod Feature Flag and key: https://precisionmedicineinitiative.atlassian.net/browse/RW-4555


api/src/main/java/org/pmiops/workbench/captcha/CaptchaServerVerificationResponse.java
```diff
@@ -0,0 +1,93 @@
+package org.pmiops.workbench.captcha;
+
+import com.fasterxml.jackson.annotation.*;
```
blrubenstein: Nit: Please use specific imports as opposed to * imports.

api/src/main/java/org/pmiops/workbench/captcha/CaptchaServerVerificationResponse.java
```diff
@@ -0,0 +1,93 @@
+package org.pmiops.workbench.captcha;
```
blrubenstein: Opt: Could either of these files be generated with Swagger? Assuming I understand and this is an api response object and a caller to the captcha API?

This is opt because this is a small number of methods and is unlikely to change, but may be worth considering.

NehaBroad: I started with the swagger approach that did not work for me hence i used this.

api/config/config_local.json
```diff
@@ -105,5 +105,9 @@
     "host": "pmi-drc-api-test.appspot.com",
     "queueName": "rdrQueueTest",
     "exportObjectsPerTask" : 10
+  },
+  "captcha": {
+    "enableCaptcha": true,
+    "serverKeyFileName": "captcha-server-key.txt"
```
gjuggler: Drive-by comment: I think I lost the last round of "battle" over this topic in a PR from a couple months back (@als364 and @jaycarlton were on that one), but I still don't believe that filenames like this really belong in our per-environment config. Our convention has always been to have consistent filenames across environments, with just the bucket name changing (see credentialsBucketName). Most of our key / credential filenames are code-constants in [CloudStorageServiceImpl](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/google/CloudStorageServiceImpl.java#L38)

I'm more than happy to be overruled if folks strongly disagree with that pattern. But it's important to note that we're actively introducing inconsistency in our system by storing key filenames in our config: some filenames will be stored in WorkbenchConfig, while most others continue to be code constants.

If we think there's benefit to migrating to the other approach, we should track a ticket & acknowledge the cost of migrating everything over there by filing a tech debt ticket.

api/src/main/java/org/pmiops/workbench/captcha/CaptchaServerVerificationResponse.java
NehaBroad: Apparently i need to borrow Greg's shame cube, the reason why this was not working in swagger was spelling mistake and security Definitions (we do not need it here)

api/config/config_local.json
als364: I agree we should migrate to the other approach, I don't think creating this new pattern should block this ticket, I do think creating said tech debt ticket should block this ticket.

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationService.java
```diff
@@ -0,0 +1,72 @@
...
+import org.pmiops.workbench.captcha.api.CaptchaApi;
+import org.pmiops.workbench.captcha.model.CaptchaVerificationResponse;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.google.CloudStorageService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+/** Service to verify Captcha */
+@Service
+public class CaptchaVerificationService {
```
als364: This looks more like a ServiceImpl than a service - please pull out the Service interface.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -94,6 +96,8 @@
   private static final String FAMILY_NAME = "Bobberson";
   private static final String CONTACT_EMAIL = "bob@example.com";
   private static final String INVITATION_KEY = "secretpassword";
+  private static final String CAPTCHA_TOKEN = "captchaToken";
+  private static final String WRONG_CAPTCHA_TOKEN = "WrongcaptchaToken";
```
als364: nit: either camelcase this or make it all lowercase

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -13,7 +13,9 @@ import {SpinnerOverlay} from 'app/components/spinners';
 import {profileApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
 import {toggleIncludes} from 'app/utils';
+import {environment} from 'environments/environment';
 import {Profile} from 'generated/fetch';
+import ReCAPTCHA from 'react-google-recaptcha';
```
als364: jeez is this really how they cased it

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -83,11 +91,21 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
         onComplete(savedProfile);
       }).catch(error => {
         console.log(error);
-        this.setState({creatingAccount: false});
+        if (environment.enableCaptcha) {
+          // Reset captcha
+          this.captchaRef.current.reset();
+          this.setState({captcha: false});
+        }
         // TODO: we need to show some user-facing error message when this fails.
```
als364: nit: can you move this TODO up to line 93?

ui/src/environments/environment.prod.ts
```diff
@@ -4,6 +4,7 @@ export const environment: Environment = {
   displayTag: '',
   shouldShowDisplayTag: false,
   allOfUsApiUrl: 'https://api.workbench.researchallofus.org',
+  captchaSiteKey: '',
```
als364: If you're not going to enable this straight away, please file a ticket to enable captcha in prod.

PR #3198, opened 2020-02-27
[RW-4429][risk=no] Add age type options in CB age wizard

dolbeew:
- Adds radio buttons with three age type options to search by: current age, age at consent and age at CDR date.
- Loads counts for all ages and age types so we can calculate the current count as the user changes the age slider
- Both changes are behind the `enableCBAgeTypeOptions` feature flag (enabled only in local env)

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/ConfigController.java
```diff
@@ -34,7 +34,7 @@
             .unsafeAllowSelfBypass(config.access.unsafeAllowSelfBypass)
             .enableNewAccountCreation(config.featureFlags.enableNewAccountCreation)
             .requireInvitationKey(config.access.requireInvitationKey)
-            .requireInstitutionalVerification(
-                config.featureFlags.requireInstitutionalVerification));
+            .requireInstitutionalVerification(config.featureFlags.requireInstitutionalVerification)
+            .enableCBAgeTypeOptions(config.featureFlags.enableCBAgeTypeOptions));
```
freemabd: Why do we need this?

api/src/main/resources/workbench.yaml
```diff
@@ -2483,6 +2483,10 @@ definitions:
         description: >
           Feature flag for verifying the institutional affiliation of new users by validating their
           contact emails.
+      enableCBAgeTypeOptions:
```
freemabd: Why do we need this?

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
```diff
@@ -237,6 +237,8 @@ public String freeTierBillingAccountName() {
     // enforced by pattern-matching the user's contact email against the institution's
     // set of whitelisted email domains or addresses
     public boolean requireInstitutionalVerification;
+    // Flag to indicate whether to use the new age type options in cohort builder age wizard
+    public boolean enableCBAgeTypeOptions;
```
freemabd: Why do we need this?

api/src/main/java/org/pmiops/workbench/api/ConfigController.java
dolbeew: Per [this comment](https://github.com/all-of-us/workbench/blob/master/ui/src/environments/environment-type.ts#L59), that is now the preferred way to add feature flags.

api/src/main/resources/workbench.yaml
dolbeew: See above response

common-api/src/main/java/org/pmiops/workbench/config/WorkbenchConfig.java
dolbeew: See above response

gjuggler:
> Looks good. Few comments about server side flags. We don't need the server side flags. Seems like client side flags are enough.

I may be somewhat responsible for this – we're trying to unify our flag control to be only based on the server-side config. It became too confusing when some flags were in environment.ts, while others were in WorkbenchConfig.

It's also easier to unit test variations of flag configs when they're not hard-coded into environment.ts.

I added this loud warning in environment.ts: https://github.com/all-of-us/workbench/blob/master/ui/src/environments/environment-type.ts#L59

freemabd:
> > Looks good. Few comments about server side flags. We don't need the server side flags. Seems like client side flags are enough.
> 
> I may be somewhat responsible for this – we're trying to unify our flag control to be only based on the server-side config. It became too confusing when some flags were in environment.ts, while others were in WorkbenchConfig.
> 
> It's also easier to unit test variations of flag configs when they're not hard-coded into environment.ts.
> 
> I added this loud warning in environment.ts: https://github.com/all-of-us/workbench/blob/master/ui/src/environments/environment-type.ts#L59

Thanks for the background Greg, this makes much more sense now. Seems like the right thing to do Will.

PR #3199, opened 2020-02-28
[RW-4446][risk=no] Clean up the enableNewAccountCreation flag

gjuggler:
This flag was pushed to prod with last week's release (see https://github.com/all-of-us/workbench/pull/3130). We should clean this up sooner than later, since the verified institutional affiliation stuff will soon add some extra config-driven complexity to the account creation page.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -248,9 +248,7 @@ public class ProfileController implements ProfileApiDelegate {
   }
 
   private void validateAndCleanProfile(Profile profile) throws BadRequestException {
-    if (profile.getDemographicSurvey() == null) {
-      profile.setDemographicSurvey(new DemographicSurvey());
-    }
+    profile.setDemographicSurvey(Optional.ofNullable(profile.getDemographicSurvey()).orElse(new DemographicSurvey()));
     if (profile.getInstitutionalAffiliations() == null) {
       profile.setInstitutionalAffiliations(new ArrayList<>());
     }
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -8,6 +8,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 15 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -247,15 +247,25 @@ private void validateStringLength(String field, String fieldName, int max, int m
     }
   }
 
-  private void validateProfileFields(Profile profile) {
+  private void validateAndCleanProfile(Profile profile) throws BadRequestException {
```
gjuggler: This is one minor functional change I made along the way... I felt we should probably have more of the profile object validation centralized in this method, which is called during both the createProfile and updateProfile flows.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -247,15 +247,25 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
+    if (profile.getDemographicSurvey() == null) {
+      profile.setDemographicSurvey(new DemographicSurvey());
+    }
```
freemabd:
```suggestion
    profile.setDemographicSurvey(Optional.ofNullable(profile.getDemographicSurvey()).orElse(new DemographicSurvey()));
```

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -247,15 +247,25 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
+    if (profile.getInstitutionalAffiliations() == null) {
+      profile.setInstitutionalAffiliations(new ArrayList<>());
+    }
```
freemabd:
```suggestion
    profile.setInstitutionalAffiliations(Optional.ofNullable(profile.getInstitutionalAffiliations()).orElse(new ArrayList()));
```

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -323,28 +333,22 @@ private DbUser initializeUserIfNeeded() {
...
-    }
-    if (request.getProfile().getInstitutionalAffiliations() == null) {
-      request.getProfile().setInstitutionalAffiliations(new ArrayList<>());
+    final Profile profile = request.getProfile();
+
+    // We don't include this check in validateAndCleanProfile since some existing user profiles
+    // may have empty addresses. So we only check this on user creation, not update.
+    if (profile.getAddress() == null) {
+      throw new BadRequestException("Address must not be empty");
     }
```
freemabd:
```suggestion
    Optional.ofNullable(profile.getAddress()).orElseThrow(() -> new BadRequestException("Address must not be empty"));
```

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jmthibault79: This seems reasonable.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -247,15 +248,23 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
 
-  private void validateProfileFields(Profile profile) {
+  private void validateAndCleanProfile(Profile profile) throws BadRequestException {
+    profile.setDemographicSurvey(
+        Optional.ofNullable(profile.getDemographicSurvey()).orElse(new DemographicSurvey()));
+    profile.setInstitutionalAffiliations(
+        Optional.ofNullable(profile.getInstitutionalAffiliations()).orElse(new ArrayList()));
+
+    String userName = profile.getUsername();
+    if (userName == null || userName.length() < 3 || userName.length() > 64) {
```
jmthibault79: Let's make this check first.

gjuggler: I'm not 100% sure what you meant, but I reshuffled things so the "validation" steps and the "cleanup" steps are more clearly delineated.

gjuggler:
> There are a few more methods which can be deleted, at 228 and 372

Thanks! Removed from create-account.tsx

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jmthibault79: thanks - was just a nit, no worries

PR #3201, opened 2020-02-28
[RW-4135][risk=no] Add non null constraint for user.contact_email

ericsong:
I've already run the necessary SQL updates across the environments so that every user has a contact_email.

---
**PR checklist**


jaycarlton:
🎖

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -307,7 +307,7 @@ public DbUser createUser(final Userinfoplus oAuth2Userinfo) {
         oAuth2Userinfo.getGivenName(),
         oAuth2Userinfo.getFamilyName(),
         oAuth2Userinfo.getEmail(),
-        null,
+        oAuth2Userinfo.getEmail(),
```
ericsong: This covers the edge case where the user already exists in Gsuite when accessing the workbench for the first time. 

In this case, I decided to just set it to the workbench username email (what we did for the backfill).

It's not the best solution since this email is non reachable but this is an edge case that should only occur in test/local environments.

PR #3205, opened 2020-02-29
[RW-4544][risk=no]Puppeteer code refactor

aweng98:
The `e2e` Puppeteer automation framework rewrite. Best approach to review PR is load branch in IDE. Read code, run a test or all tests.
writeup is in ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-4544


aweng98:
Screen Shot 2020-03-02 at 9 37 50 AM

e2e/app/FeaturedWorkspacesPage.ts
```diff
@@ -0,0 +1,24 @@
+import {waitUntilFindTexts} from '../driver/waitFuncs';
+import AuthenticatedPage from './page-mixin/AuthenticatedPage';
+
+export const FIELD_LABEL = {
+  TITLE: 'Workspace Library',
+  HEADER: 'Researcher Workbench Workspace Library',
```
aweng98: best practice:
externalize hardcoded strings outside of functions.

aweng98:
standardize each page element with its own types. allows easier test writing.
Screen Shot 2020-03-02 at 9 44 58 AM

e2e/tsconfig.json
```diff
@@ -1,19 +1,30 @@
 {
-  "extends": "../ui/tsconfig.json",
```
aweng98: extends doesn't work when I load `e2` project independently without `ui` project.

e2e/tests/workspace/workspace-ui.spec.ts
```diff
@@ -18,132 +18,171 @@ describe('Edit Workspace page', () => {
...
+  test('Click Create New Workspace link on My Workspaces page', async () => {
+    const workspaces = new WorkspacesPage(page);
+    await workspaces.goToURL();
+    const workspaceEdit = await workspaces.clickCreateNewWorkspace();
+    const workspaceNameTextbox = await workspaceEdit.getWorkspaceNameTextbox();
+    expect(await workspaceNameTextbox.isVisible()).toBe(true);
+  });
+
+
+/*
```
aweng98: commented out broken tests for now. needs fixing.

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -0,0 +1,128 @@
+import TextOptions from './TextOptions';
```
aweng98: best practice: standardized and centralize xpaths selectors in `xpath-defaults.ts`.

e2e/app/aou-elements/WebElement.ts
```diff
@@ -0,0 +1,192 @@
+import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppeteer';
+import {ElementInterface} from './ElementInterface';
+
+export default class WebElement implements ElementInterface {
```
aweng98: goodies (useful wrapper or/custom functions) are in `WebElement.ts`.

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export const FIELD_LABEL = {
+  READ_UNDERSTAND_PRIVACY_STATEMENT: 'I have read and understand the All of Us Research Program Privacy Statement',
+  READ_UNDERSTAND_TERMS_OF_USE: 'I have read and understand the All of Us Research Program Terms of Use described above',
+  INSTITUTION_NAME: 'Institution Name',
+  ARE_YOU_AFFILIATED: 'Are you affiliated with an Academic Research Institution',
+  RESEARCH_BACKGROUND: 'Please describe your research background, experience and research interests',
+  EDUCATION_LEVEL: 'Highest Level of Education Completed',
+  YEAR_OF_BIRTH: 'Year of Birth',
+};
```
aweng98: best practice: externalize hardcoded strings outside of functions.
Same pattern are in many page objects.

aweng98:
test results screenshot
Screen Shot 2020-03-02 at 4 01 26 PM

e2e/app/aou-elements/CheckBox.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (throwErr) {
+        console.error(`FAILED finding CheckBox: "${this.textOptions}".`);
+        throw e;
+      }
+    }
+    return this.element;
+  }
+
+  // find checkbox label by matching label
+  public async withMatchLabel(checkboxLabel: string): Promise<ElementHandle> {
```
aweng98: Unused at this time.

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  public async getNextButton(): Promise<Button> {
+    const button = new Button(this.puppeteerPage);
+    await button.withLabel({text: 'Next'});
+    return button;
+  }
+
+  public async scrollToNthPdfPage(nth: number): Promise<ElementHandle> {
+    const selector = `.react-pdf__Document :nth-child(${nth}).react-pdf__Page.tos-pdf-page`;
```
calbach: These kinds of selectors look really brittle. I think I'd prefer to inject a test ID into the code to make the dependency explicit.

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+    const pdfPage = await this.puppeteerPage.waitForSelector(selector);
+    await this.puppeteerPage.evaluate(el => el.scrollIntoView(), pdfPage);
+    await this.puppeteerPage.waitFor(1000);
```
calbach: By policy, I would either eliminate these or if it's unavoidable - always have a comment which explains why the sleep is needed. Arbitrary sleeps generally end in flakiness, in my experience.

Do we have something more targeted here, e.g. which can just wait for outstanding network requests to complete?

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+    await checkbox.withLabel({normalizeSpace: FIELD_LABEL.READ_UNDERSTAND_TERMS_OF_USE});
+    return checkbox;
+  }
+
+
+  public async getPrivacyStatementLabel(): Promise<Label> {
+    const label = new Label(this.puppeteerPage);
+    await label.withLabel({normalizeSpace: FIELD_LABEL.READ_UNDERSTAND_PRIVACY_STATEMENT});
+    return label;
+    // return this.getCheckboxLabel(FIELD_LABEL.READ_UNDERSTAND_PRIVACY_STATEMENT);
```
calbach: Please remove commented out code

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+    await this.puppeteerPage.waitFor(1000);
+    return newUserName;
+  }
+
+  // Step 4: Enter demographic survey default information (All Survey Fields are optional)
+  public async fillOutDemographicSurvey() {
+    // Find and check on all checkboxes with same label: Prefer not to answer
+    const targetXpath = '//*[normalize-space(text())="Prefer not to answer"]/ancestor::node()[1]/input[@type="checkbox"]';
+    await this.puppeteerPage.waitForXPath(targetXpath, { visible: true });
+    const checkboxs = await this.puppeteerPage.$x(targetXpath);
```
calbach: spelling

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,218 @@
...
+    for (const ck of checkboxs) {
+      await ck.click();
+    }
+    // Select year of birth 1955
```
calbach: nit: I would remove this kind of comment (here and below). This is just restating what the next line of code is doing, which I could have learned just as easily by reading the code. I would use comments for higher level information - or alternatively to explain "why" we are doing something. The comment 10 lines above is fine because the code is pretty hairy and difficult to read (arguably, you could alternatively eliminate the comment and breakup the xpath selection to be easier to read).

e2e/app/page-mixin/PageNavigation.ts
```diff
@@ -0,0 +1,95 @@
...
+  HOME: 'home',
+  ADMIN: 'user',
+  CONTACT_US: 'envelope',
+  USER_SUPPORT: 'help',
+  YOUR_WORKSPACES: 'applications',
+  FEATURED_WORKSPACES: 'star',
+};
+
+
+export default class PageNavigation {
```
calbach: Are these really mixins? Aren't they just base classes? Mixin implies I could include multiple of these into a page - I don't see how you would be able to do that in the current inheritance model.

e2e/app/page-mixin/PageNavigation.ts
```diff
@@ -0,0 +1,95 @@
+import {Page} from 'puppeteer';
+import {iconXpath} from '../aou-elements/xpath-defaults';
+import {findIcon} from '../aou-elements/xpath-finder';
+
+
+export const LINK = {
```
calbach: Are all of these actively used? Or is this just trying to be exhaustive?

If you haven't had it already, I think we should have a discussion about how we should think about keeping these things in sync with the application, especially if the goal is to have 100% UI coverage of the different elements.

For example, if we're depending on text labels like this, it might be more maintainable to export specific values/labels from our React components. This will result in a quick compile-time issue, rather than a runtime integration test failure, which might take a very long time to manifest.

e2e/tests/aou-elements-verify.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const linTexts = await myCard.getPopupLinkTextsArray();
+    expect(linTexts).toEqual(expect.arrayContaining(['Share', 'Edit', 'Duplicate', 'Delete']));
+  });
+
+  /**
+   * This is not a Puppeteer test for AoU application. It is for framework functions testing.
+   * If you make any change in aou-elements classes, you want to run this test to verify changes.
+   */
+  test('Click on Create New Workspace link on Home page', async () => {
```
calbach: This looks like multiple test cases to me. Depending on how expensive a single test case setup is, I would split this out into at least a few targets categories of tests.

e2e/tests/home/homepage.spec.ts
```diff
@@ -1,51 +1,142 @@
...
     await browser.close();
   });
 
-  test('Homepage is the landing page after sign in', async () => {
+  test('Homepage is the landing page after Sign In', async () => {
+    // Enable networks requests inspection
+    await page.setRequestInterception(true);
+
+    // Following network requests expected to happen
+    const targetRequestsUrls = [
```
calbach: What's the rationale for checking this? Isn't this a bit of a blackbox implementation detail? I would think in an integration test we would prefer to just check whether everything got rendered.

e2e/tests/misc/page-navigation.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  test('Sign Out', async () => {
+
+    const home = new HomePage(page);
+    await home.goToURL();
+
+    // Select Sign Out link
+    await PageNavigation.goTo(page, LINK.SIGN_OUT);
+    expect(await page.url()).toContain('login');
```
calbach: Probably at least `/login`. Better is to check the end of the path

e2e/tests/workspace/workspace-ui.spec.ts
calbach: Please fix or remove this code before merging

e2e/app/aou-elements/xpath-defaults.ts
calbach: Most of these functions seem very boilerplate-y, and seem like things that could live in any project. I believe I made this comment in the past but I don't remember the resolution. Can you add a comment here or maybe a README that explains why this layer is necessary in our application, and what goes here?

e2e/app/CreateAccountPage.ts
aweng98: a test-id attribute is desirable.  https://precisionmedicineinitiative.atlassian.net/browse/RW-4532 to keep track.

e2e/tests/workspace/workspace-ui.spec.ts
aweng98: created https://precisionmedicineinitiative.atlassian.net/browse/RW-4531 to keep track of code that need to fix. commented out code removed here.

e2e/tests/home/homepage.spec.ts
aweng98: couple reason. first, I wanted to try out code to learn how to intercept network requests in test. second, I want to grab workspaces found in response json and verify against displayed workspaces loaded in UI. I haven't been able to find time now to implement part two to make this test do more, but I will.

e2e/tests/misc/page-navigation.spec.ts
aweng98: 👍good call.

e2e/app/CreateAccountPage.ts
aweng98: Done. Also, looked elsewhere for commented out and removed them as well.

e2e/app/CreateAccountPage.ts
aweng98: corrected

e2e/app/CreateAccountPage.ts
aweng98: removed.

e2e/tests/aou-elements-verify.spec.ts
aweng98: I didn't pay too much attention to this test's organization and structure because it's used to test new or/and modified functions. I think there should be functions unit test cases to ensure functions don't break. I just haven't put in time to think it through more thoroughly. I need time to make a plan.

aweng98:
@calbach thank you for your feedback. I have addressed easy issues. Two questions still remain for addressing: mixin `PageNavigation.ts` and boilerplate xpath expressions. 
First, implement page-navigation as mixin was my original intention. but I haven't done it correctly due to I'm new to Javascript, I'm learning the language slowly. I will update code in future.
Second, reason for the xpath helper class `xpath-defaults.ts`. I want to keep all xpath selectors in one place when possible and make selectors as generic as possible. I don't know a better way to do it.
I'll set up a quick meeting sometimes this week to go over this in person.

aweng98:
@calbach `SideNav` mixin properly implemented now.

e2e/app/page-mixin/SideNav.ts
```diff
@@ -0,0 +1,83 @@
...
+import {iconXpath} from '../aou-elements/xpath-defaults';
+import {findIcon} from '../aou-elements/xpath-finder';
+
+export interface PuppeteerPage {
+  page: Page;
+}
+
+type Constructor<T> = new(...args: any[]) => T;
+
+// @ts-ignore
```
calbach: I don't think this mixin pattern is going to work well in Typescript as is and I'm not sure it's buying much. I would suggest using a composition based approach instead.

```
class Data extends AuthenticatedPage {
  public sidenav: SideNav;
}

...

test() {
  dataPage.sidenav.goTo(...)
}
```

If all authenticated pages have a side nav, move it to the base class instead

e2e/app/aou-elements/WebElement.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  public static asWebElement(page: Page, elem: ElementHandle): WebElement {
```
calbach: After reviewing more deeply, I see the potential use for the xpath utilities. However, I'm not convinced about the "WebElement / AouElement". Can you summarize in the PR description, or a doc, or a README the high level design of this framework?

Here's I'm seeing, and here's my current thinking:
- We have a "Page" concept, which wraps standard functionality / higher level page interactions. I buy the need for this.
- We have a low level xpath utility library for easier xpath creation. I think I buy the need for this *if* we're using xpath prolifically (using data-test-id more heavily reduces the need). xpath is maybe still fine for some things like picking out display strings.
- Lastly we have this WebElement / AoUElement. From what I can tell, this is effectively wrapping functionality found in ElementHandle or Page. Why are we not just directly using ElementHandle, which is the interface provided by Puppeteer? If there is common functionality we want to reuse, it may be sufficient to just use static utility helpers, rather creating an intermediate wrapper over the Puppeteer model. The advantage of using a standard API is that is a fair amount of existing documentation / support material for using it.

There are a lot of lines of code here, and a lot of things to maintain / keep in sync. I also don't want to add unnecessary wrapping / boilerplate when we can use standard interfaces. The cost might be justified in certain cases, but we want to document when/why we're making those decisions. I think a high level README.md would help here. If you're basing this off a pattern in another web testing framework, please make that clear.

gjuggler:
Top-level project tidiness request, before I jump into the code: please file a Jira ticket to associate with this PR. Include a reasonable title & short description of what motivated the work, and maybe mention what we agreed upon yesterday for acceptance criteria (e.g., good enough for general test development, even if it's imperfect and we anticipate future moderately-sized changes / improvements to the framework).

I'm moving towards asking for a Jira ticket tracking any PR larger than a very small size, and this one definitely justifies one. (Even if it wasn't included in a sprint, a Jira ticket helps us track the goals / motivation of dev work, and also gives us a place to attach follow-up PRs as needed.)

e2e/app/FeaturedWorkspacesPage.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.page = page;
+  }
+
+  public async isLoaded(): Promise<boolean> {
+    await super.isLoaded(FIELD_LABEL.TITLE);
+    await waitUntilFindTexts(this.puppeteerPage, FIELD_LABEL.HEADER);
+    return true;
+  }
+
+  public async waitForReady(): Promise<FeaturedWorkspaces> {
```
gjuggler: [Potential improvement / follow-up] I see many nearly-identical implementations of this waitForReady method. This should probably be a method implemented on the base class.

Maybe the reason you didn't do it that way was because of the desire to have a specific type of component in the returned promise?

I think you can get around this in Typescript with this (pun intended): https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types
```
export class BasePage {
  public async waitForReady(): Promise<this> {
    await this.isLoaded();
    await this.waitForSpinner();
    return this;
  }
}
```

(Disclaimer: I haven't actually tested the above.)

e2e/app/HomePage.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(page: Page) {
+    super(page);
+    this.page = page;
+  }
+
+  public async isLoaded(): Promise<boolean> {
+    await super.isLoaded(FIELD_LABEL.TITLE);
+    await new Link(this.puppeteerPage).withLabel(FIELD_LABEL.SEE_ALL_WORKSPACES);
+    await waitUntilFindTexts(this.puppeteerPage, FIELD_LABEL.HEADER);
+    return true;
```
gjuggler: I'm not sure that this method really fits in alongside the other methods. I *think* the purpose here is to wait for the page to fully load, but the name "isLoaded" suggests that it would be checking whether the page is *currently* loaded – which is definitely not the case.

For example: is there ever any scenario where this.isLoaded will resolve with a false value? If not, then I think it's meaningless to be resolving with a boolean at all.

Proposal: rename to waitForLoad, for more accurate name and consistency with other methods on the BasePage subclasses.

e2e/app/ProfilePage.ts
```diff
@@ -0,0 +1,58 @@
...
+  ABOUT_YOU: 'About You',
+  INSTITUTION: 'Institution',
+  ROLE: 'Role',
+  DISCARD_CHANGES: 'Discard Changes',
+  SAVE_PROFILE: 'Save Profile',
+};
+
+export default class Profile extends AuthenticatedPage {
+
+  public page: Page;
```
gjuggler: Another unnecessary reference to Page object – remove in favor of this.puppeteerPage.

e2e/app/UserAdminPage.ts
```diff
@@ -0,0 +1,40 @@
...
+import {SideNav} from './page-mixin/SideNav';
+const configs = require('../resources/workbench-config.js');
+
+export const FIELD_LABEL = {
+  TITLE: 'User Admin Table',
+};
+
+export default class UserAdmin extends AuthenticatedPage {
+
+  public page: Page;
```
gjuggler: Remove extra reference to Page object.

e2e/tslint.json
```diff
@@ -27,7 +27,7 @@
         "max-classes-per-file": true,
         "max-line-length": [true,
             {
-                "limit": 120,
+                "limit": 130,
```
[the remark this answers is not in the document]
aweng98: 👍

e2e/tests/aou-elements-verify.spec.ts
aweng98: removed `aou-elements-verify.spec.ts`

e2e/app/aou-elements/xpath-defaults.ts
aweng98: reworked `xpath-defaults`. almost no boilplate duplicate code.

e2e/app/page-mixin/SideNav.ts
aweng98: mixin is gone. functions moved to `BasePage`

e2e/tests/workspace/workspace-create.spec.ts
```diff
@@ -1,12 +1,8 @@
...
-import Home from '../../app/home';
-import WorkspacePage from '../../app/workspace-page';
-import { getCursorValue } from '../../driver/elementHandle-util'
+import {WorkspacesPage} from '../../app/WorkspacesPage';
 
 const Chrome = require('../../driver/ChromeDriver');
-const faker = require('faker/locale/en_US');
 
-jest.setTimeout(300000);
+jest.setTimeout(5 * 60 * 1000);
```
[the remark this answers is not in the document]
aweng98: no, test don't need be that long. changed all tests to be 2 min long max.

e2e/tests/workspace/workspace-create.spec.ts
```diff
@@ -20,91 +16,12 @@ describe('Workspace create:', () => {
...
-
-    // CREATE WORKSPACE button should be disabled
-    cursor = await getCursorValue(page, createButton);
-    expect(cursor).toEqual('not-allowed');
-
-    const words = faker.lorem.word();
-    const q2 = await workspace.question2ScientificReason();
-    await q2.type(words);
+    const workspacesPage = new WorkspacesPage(page);
+    await workspacesPage.createWorkspace(workspaceName, 'Use All of Us free credits',);
```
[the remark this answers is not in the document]
aweng98: okay 👍

e2e/tests/user/registration.spec.ts
```diff
@@ -72,44 +71,20 @@ describe('User registration tests:', () => {
...
-    const expectedH4Headers = [
-      'Your All of Us research account has been created!',
-      'Check your contact email for instructions on getting started.',
-      `Your contact email is: ${process.env.CONTACT_EMAIL}`
-    ];
-    expect(h4List).toEqual(expectedH4Headers);
-
-    const resendButton = await page.waitForXPath('//button[.="Resend Instructions"]');
-    expect(resendButton).toBeTruthy();
+    const congratMessage = 'Congratulations!';
```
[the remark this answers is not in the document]
aweng98: Done

e2e/app/DataPage.ts
```diff
@@ -1,21 +1,30 @@
...
   public async waitUntilPageReady() {
     await waitUntilTitleMatch(this.puppeteerPage, 'Data Page');
-    await this.puppeteerPage.waitForXPath(tabsSelector.showAllTab, {visible: true});
+    await this.puppeteerPage.waitForXPath(TAB_SELECTOR.showAllTab, {visible: true});
     await this.waitForSpinner();
   }
 
 }
+
+export const DataPage = SideNav(Data);
```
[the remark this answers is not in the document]
aweng98: mixin no more.

e2e/app/FeaturedWorkspacesPage.ts
aweng98: 👍 Done per suggestion

e2e/app/HomePage.ts
aweng98: `isLoaded` and `waitForReady` serve different purposes. `isLoaded` returns true if target page is indeed the right DOM, doesn't care about spinner or readiness. Updated `isLoaded` and `waitForReady`.

e2e/app/ProfilePage.ts
aweng98: side-effect of mixin. it is all being removed.

aweng98:
@calbach @gjuggler The latest commit contains lots of changes. PR is now ready for a second round of review. THANK YOU!

e2e/app/UserAdminPage.ts
aweng98: Done

e2e/tests/workspace/workspace-ui.spec.ts
```diff
@@ -1,12 +1,12 @@
...
-import AouElement from '../../driver/AouElement';
+import WebElement from '../../app/aou-elements/WebElement';
+import WorkspaceCard from '../../app/page-mixin/WorkspaceCard';
+import {WorkspacesPage} from '../../app/WorkspacesPage';
 
 const Chrome = require('../../driver/ChromeDriver');
 jest.setTimeout(60 * 1000);
 
-describe('Edit Workspace page', () => {
+describe('Workspace', () => {
```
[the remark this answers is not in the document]
aweng98: TBD test jsdoc needs to be added.

e2e/app/aou-elements/WebElement.ts
aweng98: explanation and reasons in JIRA ticket hopefully answer some of your questions here. The main reason for wrapper `WebElement` is I want to use the dot `.` notation to access useful functions instead import static functions.

e2e/app/page-mixin/SideNav.ts
aweng98: felt satisfying write very confusing and challenging JS code. 😉

e2e/tests/user/registration.spec.ts
```diff
@@ -72,44 +71,20 @@ describe('User registration tests:', () => {
 
     // Step 4: Enter demographic survey (All Survey Fields are optional)
     // Should be on Demongraphic Survey page
-    const demograpicsPageHeader = 'Demographics Survey (All Survey Fields are optional)';
-    expect(await waitForText(page, 'h3', demograpicsPageHeader)).toBeTruthy();
-
     await createAccountPage.fillOutDemographicSurvey();
+    await page.waitFor(3000);
```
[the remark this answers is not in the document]
aweng98: removed some sleeps except few that I need is needed to make tests run less flaky.

e2e/tests/home/homepage.spec.ts
```diff
@@ -1,51 +1,142 @@
...
+    expect(requestsFailed.length).toEqual(0);
+    // Expect zero page error
+    expect(pageErrors.length).toEqual(0);
   });
 
-  test('Sign in with cookies', async () => {
-    await page.setCookie(...cookies);
-    const workspaces = new WorkspacePage(page);
-    await workspaces.goURL();
+  test('Workspace cards have same UI size', async () => {
```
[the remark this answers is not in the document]
aweng98: if workspace card were ever not rendered correctly, say size == 0, this test would fail immediately. 
also I was planning to add tests to check access-level, timestamp, names.

e2e/resources/data/user-registration-fields.js
```diff
@@ -1,9 +1,9 @@
 const faker = require('faker/locale/en_US');
```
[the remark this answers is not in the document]
aweng98: Done

e2e/app/page-mixin/SideNav.ts
```diff
@@ -0,0 +1,83 @@
...
+        await this.page.waitForXPath(this.angleIconXpath, {timeout: 2000});
+      }
+    }
+
+    public async getUserName(): Promise<unknown> {
+      const xpath = `//*[child::clr-icon[@shape="angle"]/*[@role="img"]]`;
+      const username = (await this.page.$x(xpath))[0];
+      const p = await username.getProperty('innerText');
+      const value = await p.jsonValue();
+      console.log('p value = ' + value);
```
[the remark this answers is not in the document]
aweng98: Done

e2e/app/HomePage.ts
```diff
@@ -0,0 +1,55 @@
...
+  public async waitForReady(): Promise<Home> {
+    await this.isLoaded();
+    await this.waitForSpinner();
+    return this;
+  }
+
+  /**
+   * navigate to Home page URL
+   */
+  public async goToURL(): Promise<void> {
```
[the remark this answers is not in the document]
aweng98: Done

e2e/app/WorkspacesPage.ts
```diff
@@ -0,0 +1,150 @@
...
+  public page: Page;
+  constructor(page: Page) {
+    super(page);
+    this.page = page;
+  }
+
+   /**
+    * navigate to My Workspaces URL
+    */
+  public async goToURL(): Promise<void> {
```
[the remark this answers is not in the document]
aweng98: I think I missed this in my latest commit. I'll check on this to be sure it is not 3 times.

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,226 @@
+import {ElementHandle, Page} from 'puppeteer';
```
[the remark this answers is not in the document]
aweng98: kebab-case is good.
```
"rules": {
        "file-name-casing": [
            true,
            "kebab-case"
        ]
```

e2e/app/CreateAccountPage.ts
```diff
@@ -0,0 +1,226 @@
...
+  INSTITUTION_NAME: 'Institution Name',
+  ARE_YOU_AFFILIATED: 'Are you affiliated with an Academic Research Institution',
+  RESEARCH_BACKGROUND: 'describe your research background, experience, and research interests',
+  EDUCATION_LEVEL: 'Highest Level of Education Completed', // Highest Level of Education Completed
+  YEAR_OF_BIRTH: 'Year of Birth',
+};
+
+export default class CreateAccount extends BasePage {
+
+  public page: Page;
```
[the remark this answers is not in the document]
aweng98: Gone.

e2e/app/FeaturedWorkspacesPage.ts
aweng98: this is in `authenticated-page.ts` now.
```
abstract async isLoaded(): Promise<boolean>

async waitForReady(): Promise<this> {
    await this.isLoaded();
    await this.waitUntilNoSpinner();
    return this;
  }
```
All extending classes implement its own version of `isLoaded()`.

e2e/app/HomePage.ts
aweng98: forgot, but I shall rename `waitForReady` to `waitForLoad` in next commit while I address new feedback on latest commit.

e2e/app/aou-elements/web-component.ts
```diff
@@ -0,0 +1,70 @@
...
+import Checkbox from './checkbox';
+import Label from './label';
+import Link from './link';
+import RadioButton from './radiobutton';
+import Select from './select';
+import TextOptions from './text-options';
+import Textarea from './textarea';
+import Textbox from './textbox';
+
+export default class WebComponent {
```
gjuggler: opt naming nit (discussed in person): consider renaming this to BaseComponent, for consistency w/ the BasePage naming scheme. Even if there is a slight difference, since BasePage is absract, I think the parallel would be helpful.

aweng98: I think you meant `WebElement`, right? I've renamed it to `BaseElement`.

e2e/app/aou-elements/web-component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  private readonly label: TextOptions;
```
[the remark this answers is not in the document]
aweng98: agree. I renamed it to `labelTextOptions`.

e2e/app/aou-elements/xpath-finder.ts
```diff
@@ -0,0 +1,114 @@
...
+  const selector = widgetxpath.clickableXpath(label);
+  return page.waitForXPath(selector, options);
+}
+
+/**
+ * Find SELECT element with a specified label.
+ * @param {string} label
+ */
+export async function findSelect(page: Page, opts?: TextOptions, waitOptions?: WaitForSelectorOptions): Promise<ElementHandle> {
+  if (opts.ancestorNodeLevel === undefined) {
```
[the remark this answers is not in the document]
aweng98: I changed so it is required for all `TextOptions` param.

e2e/app/admin-page.ts
```diff
@@ -0,0 +1,24 @@
+import {Page} from 'puppeteer';
+import AuthenticatedPage from './authenticated-page';
+
+export const PAGE = {
+  TITLE: 'User Admin Table',
+};
+
+export default class AdminPage extends AuthenticatedPage {
```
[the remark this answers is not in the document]
aweng98: Done

e2e/app/create-account-page.ts
```diff
@@ -1,187 +1,206 @@
...
+  async fillOutUserInformation() {
+    const newUserName = await this.fillInFormFields(defaultFieldValues);
     await (await this.getResearchBackgroundTextarea()).type(faker.lorem.word());
+    // a different dropdown selection depending on Yes or No radiobutton was selected
+    await (await this.areYouAffiliatedRadioButton(true)).click();
     await (await this.getInstitutionNameInput()).type(faker.company.companyName());
-    await this.selectInstitution(registrationFields.institutionAffiliation.EARLY_CAREER_TENURE_TRACK_RESEARCHER);
-    await this.puppeteerPage.waitFor(1000);
+    await this.selectInstitution(INSTITUTION_AFFILIATION.EARLY_CAREER_TENURE_TRACK_RESEARCHER);
+    await this.page.waitFor(1000);
```
calbach: Per my previous comment, I would really like that we eliminate any arbitrary sleeps in the tests, or else have clear comments about why they are absolutely necessary.

e2e/app/aou-elements/xpath-finder.ts
```diff
@@ -0,0 +1,114 @@
...
+  return page.waitForXPath(selector, options);
+}
+
+/**
+ * Find SELECT element with a specified label.
+ * @param {string} label
+ */
+export async function findSelect(page: Page, textOptions: TextOptions, waitOptions?: WaitForSelectorOptions): Promise<ElementHandle> {
+  if (textOptions.ancestorNodeLevel === undefined) {
+    textOptions.ancestorNodeLevel = 2;
```
calbach: Per prior comment - this seems arbitrary, please add a comment explaining why this would default to 2

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -0,0 +1,70 @@
+import TextOptions from './text-options';
+
+function textXpathHelper(opts: TextOptions) {
+  let txt;
```
calbach: nit: this local var isn't really needed, you could just return directly inline

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (opts.text) {
+    txt = `text()="${opts.text}" or @aria-label="${opts.text}" or @placeholder="${opts.text}"`;
+  } else if (opts.textContains) {
+    txt = `contains(text(),"${opts.textContains}") or contains(@aria-label,"${opts.textContains}") or contains(@placeholder,"${opts.textContains}")`;
+  } else if (opts.normalizeSpace) {
+    txt = `contains(normalize-space(), "${opts.normalizeSpace}")`;
+  }
```
calbach: nit: else throw an error?

e2e/app/aou-elements/select-component.ts
```diff
@@ -0,0 +1,69 @@
...
+    await this.open(2);
+    const selector = this.componentXpath() + `//li[@class='p-dropdown-item'][normalize-space(.)="${textValue}"]`;
+    const selectValue = await this.page.waitForXPath(selector, { visible: true });
+    await selectValue.click();
+  }
+
+  async getSelectedValue(): Promise<unknown> {
+    const selector = this.componentXpath() + '/label';
+    const displayedValue = await this.page.waitForXPath(selector, { visible: true });
+    const jValue = await (await displayedValue.getProperty('innerText')).jsonValue();
```
calbach: nit: local var doesn't really do anything, just return?

e2e/app/aou-elements/select-component.ts
```diff
@@ -0,0 +1,69 @@
+import {Page} from 'puppeteer';
+
+export default class SelectComponent {
+
+  private readonly page: Page;
+  private readonly name: string;
+
+  constructor(page: Page, name?: string) {
```
calbach: Typescript constructors can be written like this:

```
constructor(
  private readonly page: Page,
  private readonly name?: string) {}
```

Then you can also remove the above type declarataions as well

e2e/app/aou-elements/select-component.ts
```diff
@@ -0,0 +1,69 @@
...
+      const is = await this.isOpen();
+      if (!is) {
+        await this.toggleOpenClose();
+      } else {
+        return;
+      }
+      if (retries < 0) {
+        return;
+      }
+      retries -= 1;
```
calbach: nit: --?

e2e/app/aou-elements/link.ts
```diff
@@ -0,0 +1,26 @@
...
+import {findClickable} from './xpath-finder';
+
+export default class Link extends BaseElement {
+
+  constructor(aPage: Page) {
+    super(aPage);
+  }
+   
+  async withLabel(aElementName: string, options?: WaitForSelectorOptions, throwErr?: boolean): Promise<ElementHandle> {
+    throwErr = throwErr || true;
```
calbach: nit: can use a param default above instead:

```
... , throwErr = true) ... {
```

e2e/app/aou-elements/textarea.ts
```diff
@@ -0,0 +1,26 @@
...
+import BaseElement from './base-element';
+import {findTextarea} from './xpath-finder';
+
+export default class Textarea extends BaseElement {
+
+  constructor(aPage: Page) {
+    super(aPage);
+  }
+   
+  async withLabel(textOptions: TextOptions, waitOptions?: WaitForSelectorOptions, throwErr?: boolean): Promise<ElementHandle> {
```
calbach: I'm confused by the pattern here. Is there a reason we want to allow someone to create a `Textarea` instance before connecting it to an actual element? i.e. should this be a static helper and `this.element` be a required constructor param?

e2e/app/aou-elements/base-element.ts
```diff
@@ -0,0 +1,229 @@
...
+    const cursor = await this.getComputedStyle('cursor');
+    return cursor === 'not-allowed';
+  }
+
+  /**
+   * Finds element's size.
+   */
+  async size(): Promise<{ width: number; height: number }> {
+    const box = await this.element.boundingBox();
+    if (!box) { return { width: 0, height: 0 }; }
```
calbach: opt_nit: In general I would not recommend 1-line if statements AND brackets. I'd do one or the other (usually in the codebase we use:
```
 if (...) {
 ...
}
```

e2e/app/aou-elements/base-element.ts
```diff
@@ -0,0 +1,229 @@
...
+  async getAttribute(attributeName: string): Promise<string | null> {
+    if (this.element == null) { throw new Error('The element is undefined.'); }
+    const elem = this.element.asElement();
+    const attributeValue = await this.page.evaluate(
+       (link, attr) => link.getAttribute(attr), elem, attributeName);
+    return attributeValue;
+  }
+
+  /*
+  const handle = await page.evaluateHandle((elem, attr) => {
```
calbach: Please remove commented out code

e2e/app/aou-elements/base-element-interface.ts
```diff
@@ -0,0 +1,30 @@
+import {ClickOptions, ElementHandle, WaitForSelectorOptions} from 'puppeteer';
+
+export interface BaseElementInterface {
```
calbach: Maybe I missed some discussion of this - but is there a reason to separate this interface from the base class? Will this be used independently somehow?

e2e/tests/misc/sidenav.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+const configs = require('../../resources/workbench-config');
+
+export const HELP_DESK = {
+  ASK_QUESTION: 'Ask a question about the Researcher Workbench',
+  REPORT_DATA_PRIVACY_CONCERN: 'Report a data privacy concern',
+  TELL_US_ABOUT_PUBLICATION: 'Tell us about an upcoming publication',
+  REQUEST_ADDITIONAL_BILLING_CREDITS: 'Request additional billing credits',
+};
+
+jest.setTimeout(2 * 60 * 1000);
```
calbach: Does this affect other test cases? Add a comment?

aweng98: Added this comment in all `.spec.ts` file. As new tests are added in future in same suite, timeout should be increased.
`// set timeout globally per suite, not per test.`

e2e/app/create-account-page.ts
aweng98: 1 second pause is needed, else test becomes flaky because Puppeteer is very fast. I added a comment.
`// need pause 1 second for dropdown to disappear, so it is not blocking click on elements below.`

e2e/app/aou-elements/xpath-finder.ts
aweng98: 👍 I added comment. `2` is chosen because it is the majority of cases.
```
// ancestor node level is used to find the closest common parent for the label element and input element.
// For most cases, closest parent element is two level up from label. Thus for the default value 2.
```

e2e/app/aou-elements/xpath-defaults.ts
aweng98: not needed because textOptions are predefined, use un-predefined string leads to compile error.

e2e/app/aou-elements/xpath-defaults.ts
aweng98: 👍

e2e/app/aou-elements/select-component.ts
aweng98: sure

e2e/app/aou-elements/select-component.ts
aweng98: removed. made two lines of code to make clear.
```
const innerText = await displayedValue.getProperty('innerText');
return await innerText.jsonValue();
```

e2e/app/aou-elements/select-component.ts
aweng98: cool!

e2e/app/aou-elements/base-element-interface.ts
aweng98: It was created for something else which has changed, so it is used only by `BaseElement` for now. I kept it in case it becomes useful in future. I don't have any strong feeling about keep or remove. either works.

e2e/app/aou-elements/base-element.ts
aweng98: 👍

e2e/app/aou-elements/base-element.ts
aweng98: 👍

e2e/app/aou-elements/link.ts
aweng98: 👍

e2e/app/aou-elements/textarea.ts
aweng98:
```
let textarea = new Textarea(this.page);
const element1 = await textarea.withLabel({normalizeSpace: FIELD_LABEL.SCIENTIFIC_APPROACHES});

In same block or function scope,  can reuse variable name textarea without creating a new instance.

const element2 = await textarea.withLabel({normalizeSpace: FIELD_LABEL.ANTICIPATED_FINDINGS});
```

calbach: I don't think reusing the variable is helping with readability at all. This seems clearer to me:

```
const element1 = await TextArea.forLabel(this,page, {{normalizeSpace: FIELD_LABEL.SCIENTIFIC_APPROACHES});
const element2 = await TextArea.forLabel(this,page, {{normalizeSpace: FIELD_LABEL.ANTICIPATED_FINDINGS});
```

e2e/app/aou-elements/xpath-defaults.ts
calbach: I don't totally follow what you're saying, but I don't think this is correct. Clearly I could pass an empty TextOptions here. The compiler would not fail.

e2e/app/create-account-page.ts
calbach: For example, in this case - could we instead block until the dropdown element is no longer visible, rather than sleeping?

e2e/app/aou-elements/base-element-interface.ts
gjuggler: +1, I also wrinkled my nose at this but forgot to comment. In general, "in case it becomes useful for the future" is a personal warning flag for https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it. Asking myself that is often a really good gut-check to make sure I'm not over-thinking something, or building functionality that will later die on the vine.

e2e/app/aou-elements/xpath-finder.ts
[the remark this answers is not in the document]
gjuggler: Thanks.

Small follow-up / no action required: an alternative approach here, especially if you wanted to have 'ancestorNodeLevel' default to 2 for most methods, would be to use a default value for the parameter (nice feature from Typescript):

async function findSelect(page: Page, opts: TextOptions = { ancestorNodeLevel: 2 }, ...)

Default param values are powerful but can be overused to negative effect, so I tend to avoid them unless it's obviously the right thing to do.

e2e/app/base-page.ts
```diff
@@ -0,0 +1,220 @@
+import {ElementHandle, Page, Response} from 'puppeteer';
+
+export default abstract class BasePage {
```
gjuggler: [action required] As discussed offline, please add some class-level comments to this base class. It really does help a lot!

PR #3207, opened 2020-02-29
[RW-3726][RISK=NO] Use merged yaml as single swagger input file

jaycarlton:
Essentially re-creating https://github.com/all-of-us/workbench/pull/2970/ but from a more recent. starting point and incorporating https://github.com/all-of-us/workbench/pull/3206. A couple of notes & learnings this time around.

* I tried merging everything into the existing `workbench.yaml` file, but that caused gradle to get into a really broken state. I think it has to do with stale references or something, and now the file is both an input and output from the same process. Or a dpendency cycle. I gave up and kept a new name `workbench-api.yaml`, as the git diff is kind of a lost cause anyway, and the SNR drops considerably after the switch to OpenAPI 3 😢(which is considered a "breaking change").
* There are trailing newlines in many of the text blocks, but I didn't fix that here because it's not a regression (at least compared to `merged.yaml`, which is what counts). I have some ideas for follow-on tools for beautification.
* diff in Beyond Compare 4 caught a property that had been removed very recently, so I've been pretty happy with that investment.

---
**PR checklist**


api/config/config_local.json
```diff
@@ -89,7 +89,6 @@
     "unsafeAllowDeleteUser": true,
     "enableVpcFlowLogs": true,
     "enableVpcServicePerimeter": true,
```
calbach: This looks like a bad rebase or merge - please fix. Or else you need to rebase this PR onto master.

api/src/main/resources/workbench-api.yaml
```diff
@@ -0,0 +1,7370 @@
...
+  "/v1/status-alert":
+    get:
+      tags:
+        - status-alert
+      description: Returns status alert information if any exists.
+      operationId: getStatusAlert
+      responses:
+        200:
+          description: 'A status alert information object.
+
```
calbach: I don't agree this is an existing issue, seems more like a problem with the tool conversion - here is the previous code: https://github.com/all-of-us/workbench/blob/c0a2f5627aeb12f35369e70a3d65f688165dfad5/api/src/main/resources/workbench.yaml#L149-L150

That said, I don't feel strongly whether this gets fixed in a follow-up or here.

api/config/config_local.json
calbach: Ah I see - it is based on a different branch - reviewing that now.

calbach: Update 2: OK - that one is already merged. In that case - please update this PR to be based on master instead of that feature branch.

blrubenstein:
FYI, I am about to merge my common-api removal PR, which will probably end up leading to a rebase. Let me know if you want help on resolving any conflicts. (I see Greg mentioned it above)

api/config/config_local.json
jaycarlton: Oh wow, you can change the base without closing the PR. That's really handy.

api/src/main/resources/workbench-api.yaml
jaycarlton: Yeah, it's only existing w/r/t merged.yaml as a base. Ruby's parser seems to want to do that.

It should be a simple one-off string replacement (if this is the last time we use that Ruby Yaml client).

api/build.gradle
```diff
@@ -485,47 +476,23 @@ dependencies {
...
   // plus any custom templates included in our repo under the swaggerTemplateDir.
-  generatedCompile "org.springframework.boot:spring-boot-starter-data-rest:${SPRING_BOOT_VERSION}"
-  generatedCompile "io.springfox:springfox-swagger2:${SPRINGFOX_VERSION}"
-  generatedCompile "io.springfox:springfox-swagger-ui:${SPRINGFOX_VERSION}"
-  generatedCompile "com.squareup.okhttp:okhttp:${OKHTTP_VERSION}"
-  generatedCompile "com.squareup.okhttp:logging-interceptor:${OKHTTP_VERSION}"
-  generatedCompile "com.google.code.gson:gson:${GSON_VERSION}"
-  generatedCompile "joda-time:joda-time:${JODA_VERSION}"
-  generatedCompile "com.fasterxml.jackson.datatype:jackson-datatype-joda:${JACKSON_VERSION}"
+  jvmCodegenDependencies "org.springframework.boot:spring-boot-starter-data-rest:${SPRING_BOOT_VERSION}"
```
calbach: Why the rename? Is this fundamentally different from the other targets we have (which follow a consistent naming pattern?) - `compile`, `testCompile`, `toolsCompile`

jaycarlton: I was confused by both the "compile" part and the past tense of "generated". I thought from the comment all these were just for the codegen step. I'm thinking of compilation narrowly as going from source to bytecode; is there a broader sense I should be thinking of?

I filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4577 to pull the codegen stuff into its own gradle file and tidy up the remnants so this is all easier to navigate.

calbach: No, it's going from source to bytecode. These are the dependencies required to compile the generated Java code, i.e. the generated source files contain imports of these 3p libraries. The codegen tooling is not something we're compiling ourselves, we're just using a gradle plugin to run that.

PR #3212, opened 2020-03-03
[risk=no] Fix some consistency issues in project.rb flags

calbach:
- Hyphenate flag names
- snakecase ruby vars
- drop "=" from flag syntax
- Rename one flag for clarity

Just noticed this when running some of the tools.

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1229,7 +1229,7 @@ def fetch_workspace_details(cmd_name, *args)
   op.opts.project = TEST_PROJECT
 
   op.add_typed_option(
-      "--workspace-project-id [project-id]",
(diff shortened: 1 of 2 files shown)


api/libproject/devstart.rb
```diff
@@ -1229,9 +1229,9 @@ def fetch_workspace_details(cmd_name, *args)
   op.opts.project = TEST_PROJECT
 
   op.add_typed_option(
-      "--projectId=[projectId]",
+      "--workspace-project-id [project-id]",
```
ericsong: `project-id` -> `workspace-project-id`?

calbach: Done.

PR #3211, opened 2020-03-03
[RW-4242][Risk=moderate] Remove common api

blrubenstein:
Description:
This removes the common api directory. 

I want to move quickly on this because it could end up with a bunch of rebases if it sits too long. 

---
**PR checklist**

- [x] This PR meets the Acceptance Criteria in the JIRA story
- [x] The JIRA story has been moved to Dev Review
- [x] This PR includes appropriate unit tests
- [x] I have run and tested this change locally
- [x] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers
- [x] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later

PR #3216, opened 2020-03-04
[RW-4548] Merge storage enums classes

als364:
In #3211 the defunct `common-api` directory was merged into the regular api directory. The `CommonStorageEnums` module from `common-api` was similar in function to the `DbStorageEnums` module, so I merged them so we wouldn't have two. Just copied all the functions from `CommonStorageEnums` into `DbStorageEnums`.


api/src/test/java/org/pmiops/workbench/db/model/StorageEnumsTest.java
```diff
@@ -75,16 +75,16 @@ private Object storageToEnum(Short shortValue, Class enumClass)
   }
 
   /**
-   * Retrieve all Enum classes in DbStorageEnums and CommonStorageEnums
+   * Retrieve all Enum classes in DbStorageEnums and DbStorageEnums
```
ericsong: Nit: this comment should be revised to just `DbStorageEnums`

als364: Looooooool. I did a global stringy search/replace. This is what I get. Thank you!

jmthibault79: Lines 23 and 87 have similar dupes

jmthibault79: And here's a dumb little PR to fix https://github.com/all-of-us/workbench/pull/3259

PR #3217, opened 2020-03-05
[RW-4546][risk=low] Bump deploy/circle to use new development image

calbach:
(Updated: I forgot we separate Dockerfiles for local dev and for Circle/deploys)

- Upgrade our CircleCI base image out of the stone age. This was necessary for installing a newer Ruby than 2.3, which is now required by the jira-ruby gem.
- Export a variable for the gradle cache directory in the build image
- Mount/cache a consistent location for the gradle cache (across circle, local dev, deploy) in `/.gradle`
- Built and pushed a new buildimage: update references accordingly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
           command: ./gradlew spotlessCheck
       - save_cache:
           paths:
-            - ~/.gradle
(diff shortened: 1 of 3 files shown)


calbach:
Ah. I screwed up. Fixing.

PR #3230, opened 2020-03-09
[RW-4572][RISK=NO]Workspace Edit after accessing Workspace About page does not display all Sub category for Race and Age

NehaBroad:
Workspace Edit is not displaying all the options/sub categories for RACE and AGE group  rather the one just selected, if the user selects the workspace Edit option from Workspace About page.

Solution: Use clone object instead.

Screen Shot 2020-03-09 at 11 46 56 AM

---
**PR checklist**


ui/src/app/utils/research-purpose.tsx
```diff
@@ -43,9 +44,9 @@ export function getSelectedResearchPurposeItems(researchPurpose: ResearchPurpose
 }
 
 export function getSelectedPopulations(researchPurpose: ResearchPurpose) {
-  const categories = SpecificPopulationItems.filter(specificPopulationItem => specificPopulationItem
-    .subCategory.filter(item => researchPurpose.populationDetails.includes(item.shortName)).length > 0);
-  categories.forEach(category => category.subCategory = category.subCategory
+  const categories = fp.cloneDeep(SpecificPopulationItems.filter(specificPopulationItem => specificPopulationItem
```
blrubenstein: Why do we need to deep clone this? Now that its a filter instead of a forEach?

NehaBroad: the logic within forEach was updating the actually SpecificPopulationItems's subCateogries. It was removing all the subCateogories that were not selected.

blrubenstein: I understand that bit, I was just wondering about the deep clone that its doing. Do we need to generate the copy?

blrubenstein:
👍 to adding a unit test

ui/src/app/utils/research-purpose.tsx
NehaBroad: had an offline discussion with @blrubenstein , we are using deep clone approach here because we want to Filter subCategories from Categories object, i.e we want to filter out nested Object hence making a new copy and manipulating it makes more sene

PR #3238, opened 2020-03-10
[risk=no] Fix circleci: revert gradle home, bump gradle revision

calbach:
Reverts gradle home changes per https://github.com/all-of-us/workbench/pull/3217 .

I can't currently explain why that didn't work, but it was just a nice-to-have and should be safe to revert. build.gradle needed to be modified in this PR in order to invalidate the previous cache version (based on build.gradle checksum).

In this case we actually had a bad cached artifact. To force a new rebuild of the cache, I used this commit: https://github.com/all-of-us/workbench/pull/3238/commits/f6cb6aa677b8ad65fc70f37ce4ca7454dd0299fa

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 
 defaults: &defaults
   docker:
-    - image: allofustest/workbench:buildimage-0.0.15
+    - image: allofustest/workbench:buildimage-0.0.16
   working_directory: ~/workbench
 java_defaults: &java_defaults
   <<: *defaults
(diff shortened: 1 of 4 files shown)


calbach:
Caching is a feature that we have configured in our Circle config. We could remove all caching if we wanted to do that, though based on my experience here that would mean executing flaky Maven downloads on every CI run which would be a nightmare.

You can read about how it works here: https://circleci.com/docs/2.0/caching/, I won't try to summarize that. Beyond changing the config and executing Circle runs, there does not appear to be any other way of manipulating the cache.

Per the PR description, one special thing I did here was to drop the cache restore step temporarily to allow for a clean cache rebuild: https://github.com/all-of-us/workbench/commit/f6cb6aa677b8ad65fc70f37ce4ca7454dd0299fa . This is something we could do again in the future if we experience a similar issue (this is the first time we've had this particular issue that I can recall).

PR #3239, opened 2020-03-10
[RW-4475][Risk=no] Working disabled UI

blrubenstein:
Description:
The UI isn't quite like the mocks, because we haven't updated the logged out headers yet. 
Screen Shot 2020-03-10 at 2 30 48 PM

I manually tested by disabling my user in test, and then hitting the UI.
---
**PR checklist**

### DIFF
--- a/ui/src/app/guards/disabled-guard.service.ts
+++ b/ui/src/app/guards/disabled-guard.service.ts
@@ -1,14 +1,13 @@
 import {Injectable} from '@angular/core';
 import {ActivatedRouteSnapshot, CanActivate, CanActivateChild, Router, RouterStateSnapshot} from '@angular/router';
 
-import {SignInService} from 'app/services/sign-in.service';
 import {profileApi} from 'app/services/swagger-fetch-clients';
 import {convertAPIError} from 'app/utils/errors';
 import {ErrorCode} from 'generated/fetch';
 
 @Injectable()
 export class DisabledGuard implements CanActivate, CanActivateChild {
-  constructor(private signInService: SignInService, private router: Router) {}
+  constructor(private router: Router) {}
 
   async canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
     return profileApi().getMe().then(_ => true).catch(error => {
--- a/ui/src/app/guards/disabled-guard.service.ts
+++ b/ui/src/app/guards/disabled-guard.service.ts
@@ -1,0 +1,29 @@
+import {Injectable} from '@angular/core';
+import {ActivatedRouteSnapshot, CanActivate, CanActivateChild, Router, RouterStateSnapshot} from '@angular/router';
+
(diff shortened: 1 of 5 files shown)


ui/src/app/pages/user-disabled.tsx
```diff
@@ -0,0 +1,57 @@
+import {Component} from '@angular/core';
+import * as React from 'react';
+
+import {StyledAnchorTag} from 'app/components/buttons';
+import {FlexColumn} from 'app/components/flex';
+import {Header} from 'app/components/headers';
+import {backgroundStyleTemplate, SIGNED_OUT_HEADER_IMAGE, SignInStep, StepToImageConfig} from 'app/pages/login/sign-in';
```
blrubenstein: I wasn't sure where to put this page. It is designed to look like the logged out pages, but I don't really like importing from different modules like this. Any ideas? (I also ended up copying some other code, that could probably be abstracted some more?

ui/src/app/guards/disabled-guard.service.ts
```diff
@@ -0,0 +1,29 @@
...
+import {SignInService} from 'app/services/sign-in.service';
+import {profileApi} from 'app/services/swagger-fetch-clients';
+import {convertAPIError} from 'app/utils/errors';
+import {ErrorCode} from 'generated/fetch';
+
+@Injectable()
+export class DisabledGuard implements CanActivate, CanActivateChild {
+  constructor(private signInService: SignInService, private router: Router) {}
+
+  async canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
```
gjuggler: nit: I don't think you're using any async syntax, so you can drop the 'async' here and below.

I personally think the async syntax is nicer than nested, but I don't think we have a style requirement to use that (yet...)

ui/src/app/pages/user-disabled.tsx
gjuggler: It could be abstracted some more, but it's not too bad as-is. The most fragile part looks like the backgroundStyleTemplate method call... I don't think including the bg image is critical to the disabled page, so I might be inclined to leave that out in the name of maintainability, but keep the header image (which could more easily be pulled out into a separate constants file, but also isn't a critical to-do IMO).

blrubenstein: The background image is actually a relatively clear part of Lou's mock/Karthik's design. I think they wanted it to look like the signin page, so that it was a similar page for users who sign in and find themself disabled:
https://projects.invisionapp.com/d/main#/console/12820961/408493657/inspect

ui/src/app/guards/disabled-guard.service.ts
blrubenstein: Want to bring it up at the eng sync? Then if everyone agrees, we can add it to our style guide markdown.

PR #3242, opened 2020-03-11
[no ticket][risk=no] Rename all classes ending in Enum

jaycarlton:
This is. technically a contract violation but we don't depend on the enum class.names in the actual requests, but I think it's worth doing. There a couple of reasons:
* we don't name all of them *Enum, so it's more consistent
* in typed languages we don't need the type in the name
* I find it harder to read,  as I have to mentally ignore the Enum suffix

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/db/model/DbWorkspace.java
```diff
@@ -422,18 +422,18 @@ public void setDisseminateResearchSet(Set<Short> disseminateResearchSet) {
   }
 
   @Transient
-  public Set<DisseminateResearchEnum> getDisseminateResearchEnumSet() {
+  public Set<ResearchPublicationOutlet> getDisseminateResearchEnumSet() {
```
calbach: While I agree these naming changes are an improvement, I'd take consistency with the DB schema over this. If you're going to change it, then you should probably rename the corresponding table/columns

jaycarlton: We have a [precedent](https://github.com/all-of-us/workbench/blob/b0239f338588c3ea9248b817b768fd60ce9eafed/api/src/main/java/org/pmiops/workbench/db/model/DbUser.java#L149) for renaming the entity property but leaving the column as-is to avoid a cosmetic migration.

FWIW it would be good for me to go through a migration before tackling the liquibase script, so I know what I'm doing a bit better. I'd like to do the "soft" renames on this PR and follow up with any DB changes though.

jaycarlton: Some of these names are bad for more than one reason though, and it doesn't make sense to me to rename them twice.

This was just a preview. method names are fixed up now. FWIW I can't spell disseminate, and it makes very little sense to have an enum be a verb.

jaycarlton: @gjuggler looking at this briefly again. I think for the workspace and research purpose names, maybe a spreadsheet would help and we can do things consistently (including updating the API and DB schema). Though I would want to use multiple PRs for that. So I guess the tradeoff is global consistency vs small, incremental PRs.

gjuggler: A spreadsheet-guided refactor would be a good way to pursue realignment of names here. I personally wouldn't consider it high-yield, given how heavily-entrenched some of these names are (including in the DB) and the arguably minimal benefit to our velocity / product at the other end.

Ultimately I think we have to be prepared for some inconsistency with how things are shown in the UI and how we name them internally, since product or stakeholders may want to adjust wording at any point in time. IMO it's not always worth going back and renaming everything, as long as we're clear where the disconnect lives. (Of course, sometimes it is worth it – for example, if we decided a workspace should now be called a project, that's a fundamental shift to a fundamental object type.)

For shining some light on these inconsistencies, a simple comment in workbench-api.yaml like "This enum corresponds to the question X in the workspace creation form" might be enough to help guide maintainers.

PR #3246, opened 2020-03-11
[RW-4360][risk=no] Create tool to load institutional affiliations from a file

ericsong:
Tool to save Institutional Affiliations specified in a JSON file.

We don't really have great tooling to support a "local" project environment through `project.rb` so I'm offering a gradle command which can be run in the meantime.

I also added a test config we can use for most of our environments which should cover most users from Broad, Verily, and VUMC. This file can be found at `api/config/test_configuration.json`

Local Usage
`./gradlew loadInstitutions -PappArgs="[\"--import-filename=institutions.json\"]"`

Cloud Environment Usage
`./project.rb load-institutions --import-filename institutions.json --dry-run=false --project all-of-us-workbench-test`

---
**PR checklist**


jaycarlton:
FWIW. I do have local env support in auditing and metrics, as well as the new ops toolkit.  LMK if you want to collaborate on a better solution.

I think one option is to stand up a "local" environment in. GCP so it's just like the others. Might be funny w/r/t the local DBs and. such though.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+  PublicInstitutionDetailsMapperImpl.class
+})
+public class LoadInstitutions {
+
+  private static final Logger log = Logger.getLogger(LoadInstitutions.class.getName());
+
+  private static Option importFilename =
+      Option.builder()
+          .longOpt("import-filename")
+          .desc("File containing JSON for institutions to save")
```
jaycarlton: It can be helpful to have an example file and refer to it from here, e.g. "see /path/to/examples/institutions.json". These would need to be fake/anonymous.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+          .required()
+          .hasArg()
+          .build();
+  private static Option dryRunOpt =
```
jaycarlton: thinking out loud, but I wonder if it wouldn't be easier for the Ruby layer to build a JSON object of the options and just pass them to Java as one chunk. No need to change this now though.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+      Option.builder()
+          .longOpt("dry-run")
+          .desc("If specified, the tool runs in dry run mode; no modifications are made")
+          .build();
+
+  private static Options options = new Options().addOption(importFilename).addOption(dryRunOpt);
```
jaycarlton: nit: with builders, I like to see them instantiated inline unless we've got conditionals happening or they're more than a couple of levels deep. I'd inline your two option vars into this declaration.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+      boolean dryRun = opts.hasOption(dryRunOpt.getLongOpt());
+
+      try (BufferedReader reader =
+          new BufferedReader(new FileReader(opts.getOptionValue(importFilename.getLongOpt())))) {
+        ObjectMapper mapper = new ObjectMapper();
+        Institution[] institutions = mapper.readValue(reader, Institution[].class);
+
+        for (Institution institution : institutions) {
+          if (institutionService.getDbInstitution(institution.getShortName()).isPresent()) {
+            log.info("Skipping... Entry already exists for " + institution.getShortName());
```
jaycarlton: do you want to check if props have changed here? Even if you don't want to upsert, it's good to know.

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
ericsong: Hmm, that would make the gradle commands pretty unwieldy if they're used without project.rb.

ericsong:
regarding the test config file, my preference would be to keep it in this repo since developers will need it to set up their environment and it should eventually be added to our local `dev-up` script.

That, and I feel like the data inside isn't very sensitive. Anyone can take a look at the repo to know that vumc/broad/verily are developing the project.

I can move it out if we think its sensitive but I figured this was a case where the convenience outweighed the costs.

looping in @gjuggler  for his thoughts

gjuggler:
> regarding the test config file, my preference would be to keep it in this repo since developers will need it to set up their environment and it should eventually be added to our local `dev-up` script.
> 
> That, and I feel like the data inside isn't very sensitive. Anyone can take a look at the repo to know that vumc/broad/verily are developing the project.
> 
> I can move it out if we think its sensitive but I figured this was a case where the convenience outweighed the costs.
> 
> looping in @gjuggler for his thoughts

(Sorry for the delay – was focused on getting the UI PR out for review.)

The test data isn't sensitive, but prod data will be.

On the one hand, having all configs in one place would be ideal for consistency (which would suggest keeping everything in workbench-devops).

On the other hand, the argument for keeping test data easily available for local-dev provisioning is also meaningful (which suggests keeping test config here).

There's also the question of "how long will this be the solution". For prod data, this is explicitly a stop-gap solution until we have a UI that admins can use. But for local / test environments, we'll likely want to keep the config around so we can easily provision a new environment.

That leads me to think we should have split storage: test config stays here, while any prod configs should land in workbench-devops.

@jaycarlton @ericsong does that sound reasonable?

api/config/test_institutions.json
```diff
@@ -0,0 +1,19 @@
+[{
```
gjuggler: File naming nit: for other configs we have {configType}_{env}.json. Prefer that convension here, e.g. institutions_test.json

api/libproject/devstart.rb
```diff
@@ -1285,6 +1285,60 @@ def export_workspace_data(cmd_name, *args)
...
 
+def load_institutions(cmd_name, *args)
+  common = Common.new
+  ensure_docker cmd_name, args
+
+  op = WbOptionsParser.new(cmd_name, args)
+  op.opts.dry_run = true
+  op.opts.project = TEST_PROJECT
+
+  op.add_typed_option(
```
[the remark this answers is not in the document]
gjuggler: I ran into something similar in https://github.com/all-of-us/workbench/pull/3225. I didn't get the sense that our hand-rolled options class was built with all the usual bells & whistles in mind :)

ericsong:
renamed it to `institutions_dev.json` to indicate that its not exactly meant to map to an environment

api/tools/src/main/java/org/pmiops/workbench/tools/LoadInstitutions.java
jaycarlton: oh I didn't look closely. if you use those option vars elsewhere, nvm.

jaycarlton:
@gjuggler re: configs. Definitely keep prod and preprod config data in workbench-devops. 

I'm personally of the opinion that nothing related to environment config or actual data should be in the public repo (for any environment). From an operational security standpoint, it's just not something anyone  outside studying our code needs to know. It's a good idea to keep a dummy config in the public repo to demonstrate the format (and for certain kinds of testing).

I understand it's a big lift to get the whole repo into compliance with this viewpoint.

ericsong:
That's fair. I can see that argument applying for all of our environmental data. That said, maybe we can tackle that issue as a whole and bring this in with it.

PR #3252, opened 2020-03-13
[RW-4412][RISK=NO]Format Dataset builder SQL query for readability

NehaBroad:
I used hibernate's BasicFormatterImpl that formats the SQL hence now we have query from dataset as follows in notebook 

For R: 

Screen Shot 2020-03-12 at 11 01 36 PM

For python
Screen Shot 2020-03-12 at 10 58 28 PM

---
**PR checklist**


jaycarlton:
That layout looks really good! I wonder if we should declare R and Python
status for the concept is lists and interpolate then into the query. Then
users could reuse them. Separate effort.


I'm on my phone; haven't looked at the actual code.

PR #414, opened 2018-01-13
Enforce auth domain for all workspaces. (And grab access level for ac…

blrubenstein:
This should by nature of FireCloud's access controls, enforce registered access for all calls that involve a workspace. @freemabd1, please let me know if I missed any cohort builder calls that should be behind registered access?

PR #3254, opened 2020-03-14
[RW-4139][risk=no] CircleCI config change to run Puppeteer tests on master merges

aweng98:
Goal: Run Puppeteer e2e tests in CircleCI on master branch merges.

- e2e tests run on "test" env. 

- job requires "ui-deploy-to-test" and "api-deploy-to-test" succeed.


.circleci/config.yml
```diff
@@ -257,7 +257,52 @@ jobs:
...
+            echo 'export PASSWORD=$PUPPETEER_TEST_USER_PASSWORD' >> $BASH_ENV
+            echo 'export INVITATION_KEY=$PUPPETEER_TEST_REGISTRATION_KEY' >> $BASH_ENV
+            echo 'export CONTACT_EMAIL=hermione.owner@quality.firecloud.org' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Run e2e tests (Puppeteer and Jest)
+          working_directory: ~/workbench/e2e
+          command: yarn --silent test:ci
+      - store_artifacts:
+          path: e2e/logs
```
aweng98: The `logs` folder is going to be created in a follow-up PR for test results and screenshots.

.circleci/config.yml
```diff
@@ -356,6 +401,7 @@ workflows:
       - ui-build-test
       - api-bigquery-test
       - api-integration-test
+      - puppeteer-e2e-test # to test Puppeteer tests start and run. To be removed before PR merge.
```
aweng98: Remove this line. Uncomment line #414 - 419 before PR merge.

.circleci/config.yml
```diff
@@ -257,7 +257,52 @@ jobs:
...
+          command: yarn cache clean && yarn install --verbose --frozen-lockfile
+      - save_cache:
+          paths:
+            - ~/.cache/yarn
+          key: ui-cache-{{ checksum "~/workbench/e2e/yarn.lock" }}
+      - run:
+          working_directory: ~/workbench/e2e
+          name: Install Headless Chrome dependencies
+          command: |
+            sudo apt-get install -yq \
```
blrubenstein: Can we bake some of these into the dockerfile? This seems like it would be a large amount of files to install every time.

.circleci/config.yml
```diff
@@ -398,6 +450,7 @@ workflows:
             - api-deps-check
             - api-integration-test
```
blrubenstein: Could you add the puppeteer tests to the staging deploy job as well? We should probably block releases on this?

gjuggler: > Could you add the puppeteer tests to the staging deploy job as well? We should probably block releases on this?

This sounds like the right approach on the surface (at least until we can run these tests hermetically pre-merge, which would provide a strong guarantee that "any merged code passes our e2e tests"). We should probably let this burn in for awhile before potentially disrupting our release process. This sounds like a good point to discuss in the design doc (@aweng98 can you add a link to the design from RW-4139?) and potentially discuss at tomorrow's eng sync.

aweng98: My recommendation is take some time to learn more about tests, test framework and Puppeteer. Let's keep it simple and non-disruptive. Only enable this job for test-deploying and non-blocking deploy and releases.

.circleci/config.yml
calbach: I'd be surprised if there isn't a prebuilt Circle image which has these. Try circleci/node:12.16.1-browsers

.circleci/config.yml
calbach: Would prefer we don't do that, this will generate noise on unrelated PRs.

.circleci/config.yml
```diff
@@ -257,7 +257,54 @@ jobs:
...
+              libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 libxcursor1 \
+              libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 ca-certificates \
+              fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget
+      - run:
+          name: Update environment variables
+          command: |
+            echo 'export USER_NAME=$PUPPETEER_TEST_USER' >> $BASH_ENV
+            echo 'export PASSWORD=$PUPPETEER_TEST_USER_PASSWORD' >> $BASH_ENV
+            echo 'export INVITATION_KEY=$PUPPETEER_TEST_REGISTRATION_KEY' >> $BASH_ENV
+            echo 'export CONTACT_EMAIL=hermione.owner@quality.firecloud.org' >> $BASH_ENV
```
calbach: This doesn't seem like it needs to be configurable via an env var.

.circleci/config.yml
calbach: Thanks, this is good. We just don't want to maintain this long list of package installs if we can avoid it. Image size of prebuilt Circle images is not really a concern since Circle automatically caches them all.

.circleci/config.yml
```diff
@@ -257,7 +257,52 @@ jobs:
...
+      - image: circleci/node:12.16.1
+    working_directory: ~/workbench
+    steps:
+      - checkout
+      - run:
+          name: Fetch Submodules
+          command: git submodule update --init --recursive
+      - restore_cache:
+          keys:
+          - ui-cache-{{ checksum "~/workbench/e2e/yarn.lock" }}
```
calbach: Please name this key differently here and below:

e.g.

```
- e2e-cache-{{ checksum ... }}
- e2e-cache-
```

This is not caching the same thing as the above UI tests

PR #3266, opened 2020-03-18
[RW-4524][risk=no] Add in-form institution check to user creation page.

gjuggler:
This adds the last missing piece to the institution form, to check the validity of the user's email address before allowing them to progress.

From local testing, demonstrating the error icon and tooltip within the input element:

Screen Shot 2020-03-18 at 12 24 08 AM

... and a validation error at the form level, which blocks form progression:

Screen Shot 2020-03-18 at 12 24 15 AM

I also found an error in one of our existing unit tests for the form (yikes!) which was causing the test to not actually verify anything. That's fixed, and the unit tests are now exercising this entire form in what I believe is a pretty reasonable manner.

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
@@ -4,6 +4,8 @@ import org.pmiops.workbench.annotations.AuthorityRequired;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.institution.InstitutionService;
 import org.pmiops.workbench.model.Authority;
+import org.pmiops.workbench.model.CheckEmailResponse;
+import org.pmiops.workbench.model.CheckInstitutionResponse;
 import org.pmiops.workbench.model.GetInstitutionsResponse;
 import org.pmiops.workbench.model.GetPublicInstitutionDetailsResponse;
 import org.pmiops.workbench.model.Institution;
@@ -56,14 +58,33 @@ public class InstitutionController implements InstitutionApiDelegate {
     return ResponseEntity.ok(response);
   }
 
-  // note: this endpoint is publicly accessible because it is needed for account creation
-
+  /**
+   * This API is publicly-accessible since it is called during account creation.
+   * @return
+   */
   @Override
   public ResponseEntity<GetPublicInstitutionDetailsResponse> getPublicInstitutionDetails() {
     final GetPublicInstitutionDetailsResponse response =
         new GetPublicInstitutionDetailsResponse()
             .institutions(institutionService.getPublicInstitutionDetails());
     return ResponseEntity.ok(response);
+  }
+
+  /**
+   * This API is publicly-accessible since it is called during account creation.
+   * @return
+   */
+  public ResponseEntity<CheckEmailResponse> checkEmail(String email, String institutionShortName) {
+    final Institution institution =
+        institutionService
+            .getInstitution(institutionShortName)
+            .orElseThrow(
+                () ->
+                    new NotFoundException(
+                        String.format("Could not find Institution '%s'", institutionShortName)));
+    return ResponseEntity.ok(
+        new CheckEmailResponse()
+            .isValidMember(institutionService.validateInstitutionalEmail(institution, email)));
   }
 
   @Override
--- a/api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
@@ -22,6 +22,14 @@ public class InstitutionController implements InstitutionApiDelegate {
     this.institutionService = institutionService;
   }
 
+  private Institution getInstitutionImpl(final String shortName) {
+    return institutionService
+        .getInstitution(shortName)
+        .orElseThrow(
+            () ->
+                new NotFoundException(String.format("Could not find Institution '%s'", shortName)));
+  }
+
   @Override
   @AuthorityRequired({Authority.INSTITUTION_ADMIN})
   public ResponseEntity<Institution> createInstitution(final Institution institution) {
@@ -38,15 +46,7 @@ public class InstitutionController implements InstitutionApiDelegate {
   @Override
   @AuthorityRequired({Authority.INSTITUTION_ADMIN})
   public ResponseEntity<Institution> getInstitution(final String shortName) {
-    final Institution institution =
-        institutionService
-            .getInstitution(shortName)
-            .orElseThrow(
-                () ->
-                    new NotFoundException(
-                        String.format("Could not find Institution '%s'", shortName)));
-
-    return ResponseEntity.ok(institution);
+    return ResponseEntity.ok(getInstitutionImpl(shortName));
   }
 
   @Override
@@ -58,9 +58,9 @@ public class InstitutionController implements InstitutionApiDelegate {
   }
 
   /**
-   * This API is publicly-accessible since it is called during account creation.
+   * Note: this API is publicly-accessible since it is called during account creation.
    *
-   * @return
+   * @return Returns publicly-accessible details about all institutions currently in the system.
    */
   @Override
   public ResponseEntity<GetPublicInstitutionDetailsResponse> getPublicInstitutionDetails() {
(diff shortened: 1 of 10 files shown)


api/src/main/resources/workbench-api.yaml
```diff
@@ -2320,23 +2320,17 @@ paths:
       tags:
       - institution
       description: 'Gets the list of all Institutions. Requires INSTITUTION_ADMIN
-        authority.
```
gjuggler: There's a bit of "clean up the awkward single-quote whitespace" here. We should probably do this for the whole file at some point, but I couldn't help myself from fixing it for institution APIs while I was here.

api/src/main/resources/workbench-api.yaml
```diff
@@ -2423,11 +2401,33 @@ paths:
           "$ref": "#/definitions/Institution"
       responses:
         200:
-          description: 'The updated Institution.
-
-'
+          description: The updated Institution.
           schema:
             "$ref": "#/definitions/Institution"
+  "/v1/institutions/{shortName}/checkEmail/{email}/":
```
gjuggler: Today I learned: the trailing slash here is important. I'd initially left it out, and what happened was that an API call such as:

/v1/institutions/VUMC/checkEmail/gjuggler%40gmail.com

would land in our controller with parameters like:

shortName: VUMC
email: gjuggler@gmail

without the trailing ".com".

Adding the trailing slash to the Swagger helped fix this – the client now sends requests with the trailing slash, and the controller parses them correctly.

(Note to self: if we were truly paranoid here, we'd add some more extensive controller-level tests to exercise the full request parsing flow. I didn't think it was worth it in this case, but I'm happy to hear disagreement.)

api/src/main/resources/workbench-api.yaml
```diff
@@ -3884,7 +3884,7 @@ definitions:
         type: string
       approved:
         type: boolean
-        default: 
+        default:
```
gjuggler: I didn't make any change here, but I think IntelliJ just removed the trailing whitespace automatically (here and below). I'm not sure how to go about removing these changes if I'm using IJ for editing...

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
```diff
@@ -235,12 +314,21 @@ export class AccountCreationInstitution extends React.Component<Props, State> {
...
                                   </div>
                                 </div>}
-                                invalid={this.state.emailFailedValidation}
-                                onBlur={() => this.validateContactEmailInline()}
-                                onChange={email => this.updateContactEmail(email)}/>
-            {this.state.emailFailedValidation &&
-              <ErrorDiv data-test-id='invalid-email-error'>
-                Error: email address is invalid
+                                invalid={!this.isEmailValid()}
+                                onBlur={() => this.onEmailBlur()}
```
jmthibault79: Is this event guaranteed to happen before the submit button is clicked?  Consider whether a user might choose to enter their Role before their Email

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
```diff
@@ -118,23 +190,30 @@ export class AccountCreationInstitution extends React.Component<Props, State> {
...
+      validationCheck['profile.verifiedInstitutionalAffiliation.institutionalRoleOtherText'] = {
         presence: {
           allowEmpty: false,
           message: '^Institutional role text cannot be blank',
         }
       };
     }
 
-    return validate(this.state.profile, validationCheck);
+    return validate(this.state, validationCheck);
```
jmthibault79: why change this?

api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
```diff
@@ -56,8 +57,11 @@
     return ResponseEntity.ok(response);
   }
 
-  // note: this endpoint is publicly accessible because it is needed for account creation
-
+  /**
+   * This API is publicly-accessible since it is called during account creation.
+   *
+   * @return
```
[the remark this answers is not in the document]
gjuggler: Done

api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
```diff
@@ -66,6 +70,24 @@
     return ResponseEntity.ok(response);
   }
 
+  /**
+   * This API is publicly-accessible since it is called during account creation.
+   *
+   * @return
+   */
+  public ResponseEntity<CheckEmailResponse> checkEmail(final String shortName, final String email) {
```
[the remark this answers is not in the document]
gjuggler: Done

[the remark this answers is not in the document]
gjuggler: I generally tend to avoid primitive return types for API responses where possible. It dramatically reduces the additional effort for someone to add another response field in the future, which I've found to be a maintenance win.

api/src/main/java/org/pmiops/workbench/api/InstitutionController.java
```diff
@@ -66,6 +70,24 @@
...
+    final Institution institution =
```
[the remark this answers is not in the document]
gjuggler: Done

api/src/main/java/org/pmiops/workbench/institution/InstitutionService.java
```diff
@@ -35,4 +35,13 @@
    */
   boolean validateAffiliation(
       @Nullable DbVerifiedInstitutionalAffiliation dbAffiliation, String contactEmail);
+
+  /**
+   * Checks whether a given email address is a valid member of an Institution.
```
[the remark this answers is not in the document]
gjuggler: Done, did some comment cleanup here.

api/src/main/java/org/pmiops/workbench/institution/InstitutionService.java
```diff
@@ -35,4 +35,13 @@
...
+   *
+   * @param institution
```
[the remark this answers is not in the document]
gjuggler: Done

api/src/main/resources/workbench-api.yaml
```diff
@@ -2423,11 +2401,33 @@ paths:
...
+      name: email
+      type: string
+      description: Institutional contact email address to check.
+      required: true
+    get:
+      tags:
+        - institution
+      operationId: checkEmail
+      description: Checks whether the given email address is a verified member of an institution.
+      x-aou-note: This endpoint is publicly-accessible and called by clients during user registration.
```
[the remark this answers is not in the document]
gjuggler: Agreed – the DOS risk is nonzero, but we have other non-auth endpoints that access the DB as well. We should probably take some time to better document our ddos-protection story: what protections are afforded at the WAF, App Engine, Google Cloud, and application levels. Filed RW-4630 to track.

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
gjuggler: This was a kind of subtle but interesting change – by providing the entire state object to validate.js, it allows us to run checks against any property of the component's state (which is the most flexible for future additions). In this case, I wanted to add some validation on the `checkEmailResponse` property, to ensure our form submission is appropriately blocked if the email-validation check hasn't completed.

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
gjuggler: That situation is caught correctly by the form-level validation: if an institution is selected, the form won't allow progress unless there is a non-null `checkEmailResponse` object in the component state.

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
jmthibault79: nice, thanks

jmthibault79:
Thanks for the updates.

One flow remains which does not work: entering the email address before choosing (or changing) the institution.  The email verification check does not get triggered this way.

gjuggler:
> Thanks for the updates.
> 
> One flow remains which does not work: entering the email address before choosing (or changing) the institution. The email verification check does not get triggered this way.

Good catch – fixed and added a test.

PR #3267, opened 2020-03-18
[risk=no][no ticket] System metadata and pre-run warning

jaycarlton:
Since this script is intended to be used only on input files from trusted sources, there's no screening in place to determine that a command won't be destructive (and of course, with local aliases, all bets are really off). Still, in case some joker tries to put in a tool with the name `rf` and the version flag `-rf /`, we want to hedge our bets. As such, the tool now prints out all the commands to be run in a list and asks for confirmation before continuing.

Additoinally, we now capture:
* username
* hostname
* os
* timestamp
* shell

Again, the intention is for this to be used internally, and users have the option of deleting stuff they don't want included before sharing. I.e. this tool isn't sending any sensitive info anywhere except a local file.

---
**PR checklist**

### DIFF
--- a/ops/ruby/devops-toolkit/README.md
+++ b/ops/ruby/devops-toolkit/README.md
@@ -67,6 +67,42 @@ To use it, do
 bundle install
 ```
 
+## Tasks
+A variety of tasks (sub-commands) are available via the required `-t` option. Here  are their
+brief descriptions:
+- `delete-all-service-account-keys`: Delete all the keys for the servie accounts and environments
+supplied. This action will interfere with any other users of this service account, so use with care.
+Main use case is cleaning up after debugging sessions where the process with a breakpoint is killed
+instead of allowed to continue and remove the key(s) that it created.
+    - usage: `./devops.rb --task delete-all-service-account-keys
+                          --envs-file ~/inputs/target_envs.json`
+- `list-all-service-account-keys`: Lists all the known service account keys for the supplied
+service accounts & environments.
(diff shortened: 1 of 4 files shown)


ops/ruby/devops-toolkit/devops.rb
```diff
@@ -26,6 +26,7 @@ def parse_options
     parser.on('-s', '--source-uri [SOURCE-URI]', String, 'URI or fully qualified name for source asset')
     parser.on('-u', '--source-env [SOURCE-ENV]', String, 'Short name for source Environment (lowercase), for example "staging".')
     parser.on('-d', '--dry-run', 'Execute a dry run of the task')
+    parser.on('-i', '--input-tools-file [INPUT-TOOLS-FILE]', String, 'Input YAML file for tools')
```
NehaBroad: so this is the option to be used for introducing a new yaml file to AoU ?

jaycarlton: it would be used for other projects. by default it uses the one I made for the aou project. The code doesn't care what project you're working on; it just takes them from the file.

jaycarlton: The AoU input file (currently the only one) was introduced in #3243 .

ops/ruby/devops-toolkit/README.md
```diff
@@ -67,6 +67,42 @@ To use it, do
 bundle install
 ```
 
+## Tasks
```
NehaBroad: nice!

PR #3285, opened 2020-03-23
[risk=no]UserServiceTestConfiguration: transitive imports for UserServiceImpl

jmthibault79:
Description:

I ran into this again and I'm now annoyed enough to solve the problem: https://github.com/all-of-us/workbench/commit/4da7c082ebc5fb94f89f914557f2703c5c8bad88

After this change, tests of controllers/services/etc which depend on UserService will not need to be individually updated after UserService's dependencies change.

---
**PR checklist**

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
@@ -276,6 +276,7 @@ public class DataSetControllerTest {
     WorkspaceMapperImpl.class,
     ManualWorkspaceMapper.class,
     LogsBasedMetricServiceFakeImpl.class,
+    UserServiceTestConfiguration.class,
   })
   @MockBean({
     BillingProjectBufferService.class,
@@ -300,7 +301,7 @@ public class DataSetControllerTest {
     Zendesk.class,
     FreeTierBillingService.class
   })
-  static class Configuration extends UserServiceTestConfiguration {
+  static class Configuration {
 
     @Bean(END_USER_CLOUD_BILLING)
     Cloudbilling endUserCloudbilling() {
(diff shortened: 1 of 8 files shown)


api/src/test/java/org/pmiops/workbench/api/DataSetControllerTest.java
```diff
@@ -307,7 +300,7 @@
     Zendesk.class,
     FreeTierBillingService.class
   })
-  static class Configuration {
+  static class Configuration extends UserServiceTestConfiguration {
```
jaycarlton: 🤔 The problem we'll run into, though, is that Java uses single inheritance. So I can't use this same trick for the WorkspaceService or NotebookService.

jaycarlton: Oh, you can have multiple classes annotated with `@TestConfiguration`! So basically you can just have 
```
@TestConfiguration
static class UserServiceTestConfig extends UserServiceTestConfiguration {}
```
and do the same for the others. So this is definitely viable for the test cases.

I think what I would change would be to leave the main configuration class as `Configuration` and then add `UserServiceConfig`, `NotebookServiceConfig`, etc, as separate classes, so they're on equal footing.

jmthibault79: very true.  I really want this to be a Scala `trait` (mixin) but we don't have this.

jaycarlton: I vaguely recall it being possible to build a config class that you can `@Import` into another one, but it's 50/50 whether I made that up.

gjuggler: > I vaguely recall it being possible to build a config class that you can `@Import` into another one, but it's 50/50 whether I made that up.

Definitely possible – see https://github.com/all-of-us/workbench/blob/c44a41fe2834788e9fb294c93263cf75cb406f6f/api/src/integration/java/org/pmiops/workbench/BaseIntegrationTest.java#L21-L21

jmthibault79: This is much better, thank you.  I'll update to this.

jmthibault79:
Ready for re-review @als364 @gjuggler @jaycarlton

PR #3286, opened 2020-03-23
[RW-4561][risk=no] factor out pdf viewer

als364:
Factor out the PDF viewer component from the account creation component. Unfortunately, I had to put an amount of the 'look, we've viewed the last page' stuff into the PDF viewer but most of it I was able to keep in the account creation component.

---

**PR checklist**

### DIFF
--- a/ui/src/app/components/pdf-viewer.spec.tsx
+++ b/ui/src/app/components/pdf-viewer.spec.tsx
@@ -1,0 +1,27 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {Document, Page} from 'react-pdf';
+
+import {PdfViewer, Props} from 'app/components/pdf-viewer';
+
+let props = {
+  windowSize: {width: 1700, height: 0},
+  pdfPath: '/assets/documents/fake-document-path.pdf'
+};
+
+it('should load PDF pages', async() => {
+  const wrapper = shallow(<PdfViewer {...props} />).shallow();
+
+  // Initially we should have a document and no pages.
+  expect(wrapper.find(Document).length).toEqual(1);
+  expect(wrapper.find(Page).length).toEqual(0);
+
+  // Simulate the PDF document loading and calling the 'onLoadSuccess' prop to indicate we have
+  // 10 pages in the PDF.
+  const pdfDocument = wrapper.find(Document);
+  const onSuccess = pdfDocument.prop('onLoadSuccess') as (data: object) => {};
+  onSuccess({numPages: 10});
+
+  // We should now be rendering a <Page> component for each of the pages.
+  expect(wrapper.find(Page).length).toEqual(10);
+});
--- a/ui/src/app/components/pdf-viewer.spec.tsx
+++ b/ui/src/app/components/pdf-viewer.spec.tsx
@@ -4,9 +4,9 @@ import {Document, Page} from 'react-pdf';
 
 import {PdfViewer, Props} from 'app/components/pdf-viewer';
 
-let props = {
-  windowSize: {width: 1700, height: 0},
-  pdfPath: '/assets/documents/fake-document-path.pdf'
+let props: Props = {
+  pdfPath: '/assets/documents/fake-document-path.pdf',
+  windowSize: {width: 1700, height: 0}
 };
 
 it('should load PDF pages', async() => {
--- a/ui/src/app/components/pdf-viewer.spec.tsx
+++ b/ui/src/app/components/pdf-viewer.spec.tsx
@@ -2,9 +2,9 @@ import {shallow} from 'enzyme';
 import * as React from 'react';
 import {Document, Page} from 'react-pdf';
 
-import {PdfViewer, Props} from 'app/components/pdf-viewer';
+import {PdfViewer} from 'app/components/pdf-viewer';
 
-let props: Props = {
+let props = {
   pdfPath: '/assets/documents/fake-document-path.pdf',
   windowSize: {width: 1700, height: 0}
 };
--- a/ui/src/app/components/pdf-viewer.tsx
+++ b/ui/src/app/components/pdf-viewer.tsx
@@ -29,6 +29,10 @@ export const PdfViewer = withWindowSize()( class extends React.Component<Props, 
     };
   }
 
+  unusedCallback() {
+    return;
+  }
+
   render() {
     const {loading, numPages} = this.state;
 
@@ -50,18 +54,18 @@ export const PdfViewer = withWindowSize()( class extends React.Component<Props, 
                   <Page
                       renderAnnotationLayer={false}
                       renderTextLayer={false}
-                      loading=''
-                      className='tos-pdf-page'
+                      loading={''}
+                      className={'pdf-page'}
                       width={Math.max(500, this.props.windowSize.width * .75)}
                       key={`page_${index + 1}`}
                       pageNumber={index + 1}
                       inputRef={index === numPages - 1
                           ? (ref) => this.props.setLastPageRef(ref)
-                          : undefined
+                          : () => this.unusedCallback()
                       }
                       onRenderSuccess={index === numPages - 1
                           ? () => this.props.handleLastPageRender()
-                          : undefined
+                          : () => this.unusedCallback()
                       }
                   />
               ),
--- a/ui/src/app/components/pdf-viewer.tsx
+++ b/ui/src/app/components/pdf-viewer.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react';
+import {Document, Page} from 'react-pdf';
+
+import {SpinnerOverlay} from 'app/components/spinners';
+import {withWindowSize} from 'app/utils';
+
+export interface Props {
+  handleLastPageRender: () => void;
+  pdfPath: string;
+  setLastPageRef: (ref) => void;
+  windowSize: {
+    width: number,
+    height: number
+  };
+}
+
+interface State {
+  loading: boolean;
+  numPages: number;
+}
+
+export const PdfViewer = withWindowSize()( class extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      loading: true,
+      numPages: 0
+    };
+  }
+
+  render() {
+    const {loading, numPages} = this.state;
+
+    return <div style={{flex: '1 1 0', overflowY: 'auto'}}>
+      {loading && <SpinnerOverlay/>}
+      <Document data-test-id='tos-pdf-document'
+                file={this.props.pdfPath}
+                loading=''
+                onLoadSuccess={
+                  data => this.setState({numPages: data.numPages, loading: false})
+                }
+      >
+        {
+          Array.from(
+            new Array(numPages),
+            (el, index) => (
+                  // We can't set inline styles on the react-pdf Page element, so instead we set a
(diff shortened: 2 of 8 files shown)


ui/src/app/components/pdf-viewer.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  // Simulate the PDF document loading and calling the 'onLoadSuccess' prop to indicate we have
+  // 10 pages in the PDF.
+  const pdfDocument = wrapper.find(Document);
+  const onSuccess = pdfDocument.prop('onLoadSuccess') as (data: object) => {};
+  onSuccess({numPages: 10});
+
+  // We should now be rendering a <Page> component for each of the pages.
+  expect(wrapper.find(Page).length).toEqual(10);
+});
```
jaycarlton: nit: check the IDE settings to insert a newline before EOF

ui/src/app/components/pdf-viewer.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  // Initially we should have a document and no pages.
+  expect(wrapper.find(Document).length).toEqual(1);
+  expect(wrapper.find(Page).length).toEqual(0);
+
+  // Simulate the PDF document loading and calling the 'onLoadSuccess' prop to indicate we have
+  // 10 pages in the PDF.
+  const pdfDocument = wrapper.find(Document);
+  const onSuccess = pdfDocument.prop('onLoadSuccess') as (data: object) => {};
+  onSuccess({numPages: 10});
```
jaycarlton: thinking out loud: Is there a way to test actual content on the pages? This kinda feels like we're saying "there are 10 pages. How many pages are there? 10? good." Not sure how much is really getting covered by this test I guess, but it may be all you can really do here.

ui/src/app/components/pdf-viewer.spec.tsx
```diff
@@ -0,0 +1,27 @@
+import {shallow} from 'enzyme';
+import * as React from 'react';
+import {Document, Page} from 'react-pdf';
+
+import {PdfViewer, Props} from 'app/components/pdf-viewer';
+
+let props = {
+  windowSize: {width: 1700, height: 0},
+  pdfPath: '/assets/documents/fake-document-path.pdf'
```
jaycarlton: Could you have a `/test/assets/documents/pdf1.pdf` or something and test with a real doc?

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+                onLoadSuccess={
+                  data => this.setState({numPages: data.numPages, loading: false})
+                }
+      >
+        {
+          Array.from(
+            new Array(numPages),
+            (el, index) => (
+                  // We can't set inline styles on the react-pdf Page element, so instead we set a
+                  // className and specify some style overrides in src/styles.css
```
jaycarlton: aside: why isn't this the preferred way to do all the styles? Just a TSX style thing?

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+                  <Page
+                      renderAnnotationLayer={false}
+                      renderTextLayer={false}
+                      loading=''
+                      className='tos-pdf-page'
+                      width={Math.max(500, this.props.windowSize.width * .75)}
+                      key={`page_${index + 1}`}
+                      pageNumber={index + 1}
+                      inputRef={index === numPages - 1
```
jaycarlton: It's not obvious by the name `inputRef` that it has to do with last page handling. Are you just using this property assignment for its side effect of calling `setLastPageRef()`?

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+            (el, index) => (
+                  // We can't set inline styles on the react-pdf Page element, so instead we set a
+                  // className and specify some style overrides in src/styles.css
+                  <Page
+                      renderAnnotationLayer={false}
+                      renderTextLayer={false}
+                      loading=''
+                      className='tos-pdf-page'
+                      width={Math.max(500, this.props.windowSize.width * .75)}
+                      key={`page_${index + 1}`}
```
jaycarlton: nit: it feels redundant to have an index, a page number, and a key like `page_99`. Do we really need all three? If so, it would help to have unit tests to make sure they are always in sync.

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+                }
+      >
+        {
+          Array.from(
+            new Array(numPages),
+            (el, index) => (
+                  // We can't set inline styles on the react-pdf Page element, so instead we set a
+                  // className and specify some style overrides in src/styles.css
+                  <Page
+                      renderAnnotationLayer={false}
```
jaycarlton: What's the difference between `{false}` and `false`?

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+                      renderTextLayer={false}
+                      loading=''
+                      className='tos-pdf-page'
```
jaycarlton: nit: I'd spell out `terms-of-service` or make it more general like `pdf-page`. That is, is there anything TOS-specific about this component?

ui/src/app/components/pdf-viewer.tsx
```diff
@@ -0,0 +1,73 @@
...
+                      width={Math.max(500, this.props.windowSize.width * .75)}
+                      key={`page_${index + 1}`}
+                      pageNumber={index + 1}
+                      inputRef={index === numPages - 1
+                          ? (ref) => this.props.setLastPageRef(ref)
+                          : undefined
+                      }
+                      onRenderSuccess={index === numPages - 1
+                          ? () => this.props.handleLastPageRender()
+                          : undefined
```
jaycarlton: instead of `undefined`, could you define a stub method to just demonstrate that it gets called and nothing happens? That feels safer too, since you don't have to test for undefined before calling.

ui/src/app/pages/login/account-creation/account-creation-tos.tsx
```diff
@@ -82,7 +80,7 @@ export const AccountCreationTos = withWindowSize()(
       }
       this.hasCreatedIntersectionObserver = true;
       const intersectionCallback: IntersectionObserverCallback = (
-        entries: IntersectionObserverEntry[], unusedObserver: IntersectionObserver) => {
+          entries: IntersectionObserverEntry[], unusedObserver: IntersectionObserver) => {
```
jaycarlton: nit: check indent

ui/src/app/pages/login/account-creation/account-creation-tos.tsx
```diff
@@ -93,44 +91,21 @@ export const AccountCreationTos = withWindowSize()(
...
-                    } : undefined}
-                    onRenderSuccess={index === numPages - 1 ? () => this.handleLastPageRender() :
-                      undefined}
-                  />
-                ),
-              )
-            }
-          </Document>
-        </div>
+        <PdfViewer
```
jaycarlton: very nice 💯

ui/src/app/components/pdf-viewer.tsx
als364: Yes, it's non-obvious, but that's what `Page` calls it, and that's not a Workbench component.

ui/src/app/components/pdf-viewer.tsx
als364: We do, they're all required and `Page` is not a Workbench component.

ui/src/app/components/pdf-viewer.tsx
als364: ooh good call

ui/src/app/components/pdf-viewer.tsx
als364: that is how one passes values. some of the rest of these are a bit wrong.

ui/src/app/components/pdf-viewer.tsx
als364: No idea. I just factored this out.

ui/src/app/components/pdf-viewer.spec.tsx
als364: ah, looking at the test for the parent component, it seems like it tests some of this functionality, and the rest is covered by E2E tests.

ui/src/app/pages/login/account-creation/account-creation-tos.tsx
```diff
@@ -93,44 +93,21 @@ export const AccountCreationTos = withWindowSize()(
...
+            handleLastPageRender={() => this.handleLastPageRender()}
+            pdfPath={this.props.pdfPath}
+            setLastPageRef={(ref) => this.setLastPageRef(ref)}
+            windowSize={this.props.windowSize}
```
ericsong: Is it necessary to pass in the windowSize if we're using `withWindowSize()` in the component? Or we can keep this and remove the `withWindowSize()`

ui/src/app/pages/login/account-creation/account-creation-tos.tsx
```diff
@@ -82,7 +80,9 @@ export const AccountCreationTos = withWindowSize()(
       }
       this.hasCreatedIntersectionObserver = true;
       const intersectionCallback: IntersectionObserverCallback = (
-        entries: IntersectionObserverEntry[], unusedObserver: IntersectionObserver) => {
+        entries: IntersectionObserverEntry[],
```
ericsong: the rest of the stuff in `handleLastPageRender()` looks like stuff that might be able to go into the shared component.

I don't fully understand how it works but I think, ultimately, components that use the PdfViewer component just want to know when the user has read the entire PDF and not concern themselves with stuff like `IntersectionObserverCallback`.

ui/src/app/pages/login/account-creation/account-creation-tos.spec.tsx
```diff
@@ -71,14 +48,15 @@ it('should enable checkboxes and next button with user input', async() => {
   expect(getNextButton(wrapper).prop('disabled')).toBeTruthy();
 
   // Now, simulate checking both boxes, which should enable the "next" button.
+  debugger;
   getPrivacyCheckbox(wrapper).simulate('change', {target: {checked: true}});
```
calbach: Per slack discussion, it appears that in order to trigger the event property you need to call `simulate()` on the `input` rather than the `CheckBox` element.

ui/src/app/pages/login/account-creation/account-creation-tos.tsx
```diff
@@ -51,137 +49,79 @@ interface AccountCreationTosState {
...
-                understand the terms of this agreement and agree to abide by them.
-              </div>
-              <div>
-                Please read through the entire agreement to continue.
-              </div>
+    return <FlexColumn data-test-id='account-creation-tos'
+                       style={{flex: 1, padding: '1rem 3rem 0 3rem'}}>
+      <PdfViewer
+          onLastPageRender={() => this.setState({hasReadEntireTos: true})}
+          pdfPath={this.props.pdfPath}
```
ericsong: Nice! This interface looks super easy to use.

PR #3287, opened 2020-03-23
[no ticket][risk=no] sync creationTime with deid dataset creation time

freemabd:
Update the creationTime to be the create date of the generating deid CDR dataset. This date will be used by the cohort builder when querying based on the event date modifier and this will also clue us into which aou-res-curation-prod dataset was used to generate that specific cdr. This change shouldn't have any side effects. AFAICT nothing references this property in [DbCdrVersion](https://github.com/all-of-us/workbench/blob/b6d2b8adbfac3fd8386442e3c6fb3be8c23a7b96/api/src/main/java/org/pmiops/workbench/db/model/DbCdrVersion.java#L28-L28). @calbach  Please let me know if you see any potential snags in changing these dates.

Also updated the CDR playbook to make sure the creationTime is set from deid aou-res-curation-prod dataset date: [Playbook Entry](https://docs.google.com/document/d/1St6pG_EUFB9oRQUQaOSO7a9UPxPkQ5n4qAVyKF9j9tk/edit#bookmark=id.pyyn0srcju3n)

**Below are the new dates from the generated deid datasets**

> synthetic_cdr20180606 
> generated from **all-of-us-ehr-dev:synthetic_cdr20180606**
> new creationTime **2018-06-06 00:00:00Z**

> R2018Q4R1 
> generated from **aou-res-curation-prod:deid_output_20181116**
> new creationTime **2018-11-16 00:00:00Z**

> R2019Q2R1             
> generated from **aou-res-curation-prod:deid_output_20190510**
> new creationTime **2019-05-10 00:00:00Z**

> R2019Q4R3
> generated from **aou-res-curation-prod:deid_output_20191004_clean**
> new creationTime **2019-10-04 00:00:00Z**


freemabd:
> I don't think it's currently used anywhere. I didn't realize the offsets worked this way - interesting. Could you also add the new semantics to the yaml docs here
> 
> https://github.com/all-of-us/workbench/blob/45ec4131d32cff451bf99c27a7f40442e87f05f4/api/src/main/resources/workbench-api.yaml#L3405
> 
> ?

Done

PR #3288, opened 2020-03-23
[RW-4634][]risk=no]Add custom Chrome startUp flags

aweng98:
Extending Puppeteer default Chrome startUp flags (switches) with CircleCi relevant flags to improve test stability and performance.

### DIFF
--- a/e2e/driver/puppeteer-launch.ts
+++ b/e2e/driver/puppeteer-launch.ts
@@ -1,22 +1,15 @@
 import * as puppeteer from 'puppeteer';
+import {getChromeFlags} from './chrome-startup-flags';
+
+const isHeadless = !process.env.PUPPETEER_HEADLESS;
 
 const defaultLaunchOpts = {
-  headless: !process.env.PUPPETEER_HEADLESS,
+  headless: isHeadless,
   slowMo: 10,
   defaultViewport: null,
   devtools: false,
-  ignoreDefaultArgs: ['--disable-extensions'],
-  args: [
-    '--no-sandbox',
-    '--disable-setuid-sandbox',
-    '--disable-dev-shm-usage',
-    '--disable-gpu',
-    '--disable-background-timer-throttling',
-    '--disable-backgrounding-occluded-windows',
-    '--disable-renderer-backgrounding',
-    '--disable-web-security',
-    '--disable-features=IsolateOrigins,site-per-process',
-  ]
+  ignoreDefaultArgs: true,
+  args: getChromeFlags(isHeadless)
 };
 
 export default async (opts?): Promise<puppeteer.Browser> => {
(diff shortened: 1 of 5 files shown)


e2e/package.json
```diff
@@ -7,9 +7,9 @@
   "license": "BSD",
   "scripts": {
     "test": "WORKBENCH_ENV=dev jest --detectOpenHandles",
-    "test:debug": "WORKBENCH_ENV=dev PUPPETEER_HEADLESS=false jest --detectOpenHandles --runInBand",
+    "test:debug": "WORKBENCH_ENV=dev PUPPETEER_HEADLESS=false jest --detectOpenHandles --maxWorkers=2",
     "test-local": "WORKBENCH_ENV=local jest",
-    "test:ci": "cross-env NODE_ENV=development WORKBENCH_ENV=dev INCOGNITO=true jest --ci --maxWorkers=2 --forceExit",
+    "test:ci": "cross-env NODE_ENV=development WORKBENCH_ENV=dev INCOGNITO=true jest --ci --runInBand --forceExit",
```
aweng98: `--runInBand` is same as `-maxWorkers=1`: Ensure tests are run in serial order in one thread in CircleCi.

e2e/tests/user/registration.spec.ts
```diff
@@ -61,7 +61,7 @@ describe('User registration tests:', () => {
     await createAccountPage.fillOutInstitution();
     nextButton = await createAccountPage.getNextButton();
     await nextButton.waitUntilEnabled();
-    await nextButton.click();
+    await nextButton.clickWithEval();
```
aweng98: occasionally, Puppeteer `click()` on the `Next` button is not working. Changing to call to `Document.click()` function.

aweng98:
Tests passed on localhost.
Screen Shot 2020-03-23 at 12 26 08 PM

e2e/driver/puppeteer-launch.ts
```diff
@@ -1,22 +1,15 @@
...
-    '--disable-setuid-sandbox',
-    '--disable-dev-shm-usage',
-    '--disable-gpu',
-    '--disable-background-timer-throttling',
-    '--disable-backgrounding-occluded-windows',
-    '--disable-renderer-backgrounding',
-    '--disable-web-security',
-    '--disable-features=IsolateOrigins,site-per-process',
-  ]
+  ignoreDefaultArgs: true,
```
ericsong: I don't see code that's handling this argument, `ignoreDefaultArgs`. How is it able to continue working if the type was changed form a string array to a boolean?

aweng98: Because `ignoreDefaultArgs` takes param either is `boolean` or `Array<string>`.
```
ignoreDefaultArgs <boolean|Array<string>> If true, then do not use puppeteer.defaultArgs(). If an array is given, then filter out the given default arguments. Dangerous option; use with care. Defaults to false.
```

PR #3294, opened 2020-03-24
[RW-4621][risk=no] Add Zendesk integration test which runs nightly

calbach:
This only runs nightly on account of the test generating ZD sandbox requests which aren't cleaned up. Leverages Junit "Categories" for filtering on tests. I couldn't get `@Tag` filtering to work with gradle; on further review this may only be possible if you're using the new JUnit5 platform engine, so upgrading to the JUnit5 lib may not have been sufficient:

```
> Could not find method useJUnitPlatform() for arguments [build_7rmii44aljl4a99qxkq0ucme2$_run_closure20$_closure70@31450fc0] on task ':integrationTest' of type org.gradle.api.tasks.testing.Test.
```

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -496,6 +496,10 @@ dependencies {
   testCompile 'com.h2database:h2:1.4.194'
   testCompile 'org.liquibase:liquibase-core:3.5.3'
   testCompile 'org.bitbucket.radistao.test:before-after-spring-test-runner:0.1.0'
+  testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.4.0'
+  testRuntime 'org.junit.vintage:junit-vintage-engine:5.4.0'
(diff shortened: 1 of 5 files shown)


api/build.gradle
```diff
@@ -522,18 +526,28 @@ compileJava {
   options.compilerArgs << '-Xlint:unchecked'
 }
 
-task integration(type: Test) {
+task integrationTest(type: Test) {
   group = LifecycleBasePlugin.VERIFICATION_GROUP
   testClassesDirs = sourceSets.integration.output.classesDirs
   classpath = sourceSets.integration.runtimeClasspath
+  useJUnit {
+		excludeCategories 'org.pmiops.workbench.NightlyTests'
```
blrubenstein: Nit: indentation seems off here.

calbach: Ew, thanks - fixed.

PR #3298, opened 2020-03-25
[risk=medium][RW-4465][RW-4637]Enable requireInstitutionalVerification in Staging and Perf

jmthibault79:
Enable requireInstitutionalVerification in Staging and Perf
Disable requireInvitationKey in Test, Local, Perf, and Staging

Conservatively labeling "medium" risk because we're removing one barrier to access.  Mitigated by requiring another, which should be better security-wise.

Enable/disable Feature Flags: https://precisionmedicineinitiative.atlassian.net/browse/RW-4465
TODO: Stable, Prod

Remove Feature Flags: https://precisionmedicineinitiative.atlassian.net/browse/RW-4361

Description:

---
**PR checklist**

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -23,7 +23,10 @@
     "bufferCapacity": 2,
     "bufferRefillProjectsPerTask": 1,
     "defaultFreeCreditsDollarLimit": 300.0,
-    "freeTierCostAlertThresholds": [0.5, 0.75],
+    "freeTierCostAlertThresholds": [
+      0.5,
+      0.75
+    ],
     "garbageCollectionUserCapacity": 1000,
     "garbageCollectionUsers": [
       "aou-rw-gc-test-1@all-of-us-workbench-test.iam.gserviceaccount.com"
@@ -82,7 +85,7 @@
     "enableDataUseAgreement": true,
     "enableBetaAccess": true,
     "unsafeAllowSelfBypass": true,
-    "requireInvitationKey": true
+    "requireInvitationKey": false
   },
   "featureFlags": {
     "unsafeAllowDeleteUser": true,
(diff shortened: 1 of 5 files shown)


api/config/config_local.json
```diff
@@ -23,7 +23,10 @@
     "bufferCapacity": 2,
     "bufferRefillProjectsPerTask": 1,
     "defaultFreeCreditsDollarLimit": 300.0,
-    "freeTierCostAlertThresholds": [0.5, 0.75],
+    "freeTierCostAlertThresholds": [
```
jmthibault79: Welp.  The good thing about autoformatters is that they only run once.

gjuggler: Which autoformatter did you run? I'm wondering if we can document something somewhere about how this should be done, so we can have consistent adoption.

api/config/config_perf.json
```diff
@@ -85,7 +88,7 @@
     "enableDataUseAgreement": true,
     "enableBetaAccess": true,
     "unsafeAllowSelfBypass": true,
-    "requireInvitationKey": true
+    "requireInvitationKey": false
```
gjuggler: FYI @aweng98 in case this flag flip might affect our perf tests at all. I think it's unlikely to be impacted, just wanted to check.

(My thought process is that perf tests may be impacted only if both of the following are true: (1) nightly perf tests are creating new users, and (2) our API returns an error response if an invitation key is provided bu requireInvitationKey is false. My quick read of [this line](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/api/ProfileController.java#L343) is that we just ignore the invitation key when it's not required.

api/config/config_local.json
jmthibault79: IntelliJ Save Action, following our onboarding/setup doc: https://docs.google.com/document/d/1DtESBapEzvuti7xODTFPHorwmLM7LybF-6D5lhbIkLU/edit#heading=h.sjqyvxq4luy4

There are many options for Save Actions however, and I was initially quite overzealous on a different branch and had to revert some changes.

api/config/config_perf.json
jmthibault79: I spoke to Alex offline - we're ok

PR #3302, opened 2020-03-26
[RW-4621][risk=no] Fix nightly command

calbach:
Fixed some late-breaking regressions on the last PR which I apparently neglected to run afterwards. Verified locally that `./project.rb nightly-integration` now passes.


calbach:
To be clear, this is the case for 100% of the commands in this file. It's a wart of the current command system. The reason for it, which is tricky to follow from the code, is to re-execute the command within a docker context, if you're not already in docker. This mechanism ensures that most commands here are run in a consistent docker context.

PR #3309, opened 2020-03-27
[RW-4555][RISK= Moderate]  Enable Account Creation Captcha for Prod

NehaBroad:
Screen Shot 2020-03-27 at 10 38 03 AM

Ticket to clean up feature flag related code once this works on Prod https://precisionmedicineinitiative.atlassian.net/browse/RW-4673

As part of this PR:

- Replace google captcha test/local and staging to use google test keys to help with puppeteer test.

-  Modify the logic to match HostName in Api

- **note:** I have not updated Test and Prod site keys yet. Test site key will be updated once this Pr is merged in (since test is still using old captcha keys). Production key upload might need permission (?)

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
```diff
@@ -17,7 +21,21 @@
 
   final String urlPattern = "https://%s/login";
 
-  final String localHostUrlPattern = "http://%s:4200/login";
+  final List<String> nonTestCaptchaURLs =
```
calbach: I would rather drive this from a config value per environment, unless there's something I'm missing. This list approach seems brittle to me.

ui/src/environments/environment.prod.ts
```diff
@@ -19,7 +19,7 @@ export const environment: Environment = {
   trainingUrl: 'https://aou.nnlm.gov',
   inactivityTimeoutSeconds: 30 * 60,
   inactivityWarningBeforeSeconds: 5 * 60,
-  enableCaptcha: false,
+  enableCaptcha: true,
```
calbach: Existing issue, but it looks like this is redefining a value that's set in the server-side config. Probably better to pass this from the server. Fine as a follow-up if this is large, or maybe we could just wait to delete this, depending on how long we think the flag will stay around.

api/src/test/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+      mockCaptchaResponse("testkey.google.com", true);
+      captchaVerificationService.mockLoginUrl(prodAllOfUsUrl);
+      boolean captchaSuccess = captchaVerificationService.verifyCaptcha(responseToken);
+      assertThat(captchaSuccess).isFalse();
+
+      // AllOfUs test url should be using google test captcha keys
+      captchaVerificationService.mockLoginUrl(testAllOfUsUrl);
+      captchaSuccess = captchaVerificationService.verifyCaptcha(responseToken);
+      assertThat(captchaSuccess).isTrue();
+    } catch (ApiException e) {
```
calbach: I would just add `throws Exception` to the unit test method definition throughout and remove these catches, I don't think the approach here offers any benefit over that

api/src/test/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+    mockCaptchaResponse("hostName", true);
+  }
+
+  private void mockCaptchaResponse(String hostName, boolean success) {
+    CaptchaVerificationResponse response = new CaptchaVerificationResponse();
+    response.setHostname(hostName);
+    response.setSuccess(true);
+    try {
+      Mockito.when(captchaApiProvider.verify("key", responseToken)).thenReturn(response);
+    } catch (ApiException ex) {
```
calbach: again, I'd just add the throws

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
```diff
@@ -17,7 +21,21 @@
...
+
+  final List<String> testCaptchaURLs =
+      new ArrayList<String>(
+          Arrays.asList(
+              "http://localhost:4200/login",
+              "https://all-of-us-workbench-test.appspot.com",
+              "https://all-of-us-rw-staging.appspot.com/login",
+              "https://all-of-us-rw-perf.appspot.com/login"));
+
+  final String googleTestHost = "testkey.google.com";
```
calbach: Could you add some comments that indicate what the behavior is for a "test captcha"? Are these just two separate keys,  or does it just allow any captcha input to pass?

api/config/config_perf.json
```diff
@@ -113,6 +113,6 @@
     "exportObjectsPerTask": 10
   },
   "captcha": {
-    "enableCaptcha": false
+    "enableCaptcha": true
```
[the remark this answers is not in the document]
NehaBroad: https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
NehaBroad: I was not sure if just checking google test key should be enough for test/staging/perf and local. Normally for captcha verification we would check site key + host matches the url (the way we are doing it for prod and staable). But again since i am checking url against the config entry i can just create a new entry to test if the environment is using testCaptcha

ui/src/environments/environment.prod.ts
NehaBroad: hmm since it will be part of next weeks release, i can have a clean up PR ready for next week?

api/src/test/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.pmiops.workbench.captcha.api.CaptchaApi;
+import org.pmiops.workbench.captcha.model.CaptchaVerificationResponse;
+import org.pmiops.workbench.config.WorkbenchConfig;
+import org.pmiops.workbench.google.CloudStorageService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.context.annotation.Import;
+
+public class CaptchaVerificationServiceTest extends BaseControllerTest {
```
calbach: This isn't a controller, not sure it should be extending BaseControllerTest

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
```diff
@@ -36,6 +41,21 @@ public CaptchaVerificationServiceImpl(
     this.captchaApiProvider = captchaApiProvider;
   }
 
+  @VisibleForTesting
+  public void mockLoginUrl(String loginUrl) {
```
calbach: This is not ideal, should just remove these and use the same approach that is used in this test case (define a static WorkbenchConfig, then provide a prorotype scoped bean of it): https://github.com/all-of-us/workbench/blob/61036972900bae172325d6128b604d647eb05745/api/src/test/java/org/pmiops/workbench/api/SumoLogicControllerTest.java#L42

In the individual test cases, override the values directly on the static config with the config you want.

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
```diff
@@ -60,11 +80,17 @@ public boolean verifyCaptcha(String responseToken) throws ApiException {
...
-    boolean captchaHostNameMatchUI =
-        String.format(urlPattern, captchaHostname).equals(uiUrl)
-            || String.format(localHostUrlPattern, captchaHostname).equals(uiUrl);
+    // Production/Stable should  not use google test Key and the domainName should match with the
+    // one return from Captcha Response
+    if (!usingTestCaptcha) {
+      // check if the UI URL has the host as send by Captcha Response
+      captchaHostNameMatchUI = String.format(urlPattern, captchaHostname).equals(uiUrl);
+    } else {
+      captchaHostNameMatchUI = captchaHostname.equals(googleTestHost);
```
calbach: Should you just compare this to `config.server.uiBaseUrl`, or at least - extract the host from there ?

NehaBroad: i was initially checking the config.loginUrl to be one of test/login/stable but since it was basically checking just the config variable to constants i was not sure if it was adding anything.

NehaBroad: https://github.com/all-of-us/workbench/blob/fd85f5755b299cfa8ac1f092dc9752ca87ba4d44/api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java L100

ui/src/environments/environment.prod.ts
calbach: Sorry, forgot to reply ehre - yes that's fine

api/src/main/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceImpl.java
```diff
@@ -60,11 +91,20 @@ public boolean verifyCaptcha(String responseToken) throws ApiException {
...
+    // Production/Stable should  not use google test Key and the domainName should match with the
+    // one return from Captcha Response
+    if (!usingTestCaptcha) {
+      // check if the UI URL has the host as send by Captcha Response
+      captchaHostNameMatchUI = String.format(urlPattern, captchaHostname).equals(uiUrl);
+    } else {
+      // If URL is using google test key hence the host return will always be testkey.google.com
+      // hence check if the UI url in config matches one of AllOfUs urls(local/test/staging/perf)
+      captchaHostNameMatchUI =
+          captchaHostname.equals(googleTestHost) && testCaptchaURLs.contains(uiUrl);
```
NehaBroad: i have added the logic again for testing the urls that are using test google captha. Since the host from the captcha service will always be testkey.google.com, check if the ui url will be one of the environment using test captcha test/local/staging and perf. Let me know if this makes sense/ is there any other check that i can add.

calbach: Sorry, I still don't understand why you need to hardcode a list of URLs here. Can you not use the combination of `config.server.uiBaseUrl` with `usingTestCaptcha` to make the determination?

api/src/test/java/org/pmiops/workbench/captcha/CaptchaVerificationServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Before
+  public void setUp() throws IOException {
+    config = WorkbenchConfig.createEmptyConfig();
+    config.admin.loginUrl = "hostname";
+    Mockito.when(cloudStorageService.getCaptchaServerKey()).thenReturn("key");
+    try {
+      mockCaptchaResponse("hostName", true);
+    } catch (ApiException ex) {
```
calbach: This catch should no longer be necessary

PR #3315, opened 2020-03-28
[RW-4583][risk=low] Switch to Rawls cloneWorkspace

calbach:
System changes:

- Drop our custom handling of file cloning on CloneWorkspace, leverage the Firecloud cloneWorkspace file copy feature instead.

Functional changes:

- We no longer copy all files in the bucket, just in the notebooks/ subdir. This is a revert to the original behavior which was intended to mitigate accidental copies of large files. It's required for this change since Rawls clone workspace doesn't support copying all files
- Drop the size limitation on cloned file size
- Files are no longer synchronously available immediately after the clone, Rawls does this in a background thread.

Based on manual local testing:
- The propagation delay was not noticeable - tested with 50 notebook files.
- Confirmed that files outside this prefix do not get copied
- Confirmed that files in "subdirectories" do get copied

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
@@ -330,7 +330,7 @@ public class FireCloudServiceImpl implements FireCloudService {
             .namespace(toProject)
             .name(toName)
             // We copy only the notebooks/ subdirectory as a heuristic to avoid unintentionally
-            // performing a deep copy of workspace blobs.
+            // propagating copies of large data files elswhere in the bucket.
             .copyFilesWithPrefix("notebooks/")
             .authorizationDomain(
                 ImmutableList.of(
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java
```diff
@@ -315,23 +309,36 @@ public void removeOwnerFromBillingProject(
...
-    FirecloudWorkspaceIngest workspaceIngest = new FirecloudWorkspaceIngest();
-    workspaceIngest.setNamespace(toProject);
-    workspaceIngest.setName(toName);
-    checkAndAddRegistered(workspaceIngest);
+    FirecloudWorkspaceRequestClone cloneRequest =
+        new FirecloudWorkspaceRequestClone()
+            .namespace(toProject)
+            .name(toName)
+            // We copy only the notebooks/ subdirectory as a heuristic to avoid unintentionally
+            // performing a deep copy of workspace blobs.
```
jaycarlton: Do we ever want to do a deep copy?

calbach: I updated the comment, it didn't quite say what I originally intended. Deep copy is ambiguous terminology in this context.

If your question is: are there use cases for copying different subsets of files (including all / none) when copying  a workspace? Yes. Whether we support those is a product question.

jaycarlton: 👍

PR #3321, opened 2020-03-31
[RW-4691][risk=no] Await authentication before checking disabled status

calbach:
This causes the initial /getMe request to fail with a 500. There is no clear user facing issue as a result of this, but it is a confounding factor in the very least to debugging other issues.

Manually tested with a disabled and a non-disabled user.

### DIFF
--- a/ui/src/app/guards/disabled-guard.service.ts
+++ b/ui/src/app/guards/disabled-guard.service.ts
@@ -12,7 +12,9 @@ export class DisabledGuard implements CanActivate, CanActivateChild {
 
   async canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
     try {
-      // Only grab the first status, otherwise toPromise hangs forever.
+      // The user is not necessarily authenticated at this point - we have to
+      // wait for the sign-in signal before making backend requests.
+      // Only grab the first element from the observable, otherwise toPromise hangs forever.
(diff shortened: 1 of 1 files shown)


ui/src/app/guards/disabled-guard.service.ts
```diff
@@ -1,16 +1,24 @@
...
 import {ErrorCode} from 'generated/fetch';
 
 @Injectable()
 export class DisabledGuard implements CanActivate, CanActivateChild {
-  constructor(private router: Router) {}
+  constructor(private router: Router, private signInService: SignInService) {}
 
   async canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
     try {
+      // The user is not necessarily authenticated at this point - we have to
```
gjuggler: You just have read my mind – I was adding a review comment asking for a comment, then your comment showed up when I refreshed :)

I'm still a little confused as to how we have confidence that this guard will be called again once the sign-in state is changed. Is this something we handle explicitly somewhere?

calbach: The observable here blocks until the state is known, at which point it fires the first event which gets captured by this `toPromise`. Subsequently, it may fire again I believe if the user sign in state changes again, but we're no longer listening at that point.

calbach: On initial sign in - we're currently dependent on a full page reload. Therefore signed out -> signed in is not a valid state transition anyways.

PR #3333, opened 2020-04-02
[RW-4249][Risk=low] Srubenst/data page objects

blrubenstein:
Description:
This adds a getDataPageObjects endpoint to reduce the number of access checks necessary. Because of this, it also changes `RecentResource` to `WorkspaceResource` to be more correct, and adds two mappers so it only uses mappers.

---
**PR checklist**


api/src/main/resources/workbench-api.yaml
```diff
@@ -804,6 +804,20 @@ paths:
           description: Workspace deletion request accepted
           schema:
             "$ref": "#/definitions/EmptyResponse"
+  "/v1/workspaces/{workspaceNamespace}/{workspaceId}/get-data-page-items":
```
jaycarlton, line 807: We don't need `get` in the resource name, since that's in the verb.

Also, can we be more descriptive than "data page"? We don't need `items` either, as it doesn't really clarify what we're fetching.

api/src/main/resources/workbench-api.yaml
```diff
@@ -804,6 +804,20 @@ paths:
...
+    parameters:
+      - "$ref": "#/parameters/workspaceNamespace"
+      - "$ref": "#/parameters/workspaceId"
+    get:
+      tags:
+      - workspaces
+      description: Gets all items shown on the data page of the workspcae
```
jaycarlton, line 814: This definition is from the perspective of an existing UI design or implementation. It should really be written in a stand-alone way, so that a developer of a future workbench client could use it. That is, the API definition shouldn't really be in terms of the UI.

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
```diff
@@ -167,7 +167,7 @@ public void setDistinctWorkspaceLimit(int limit) {
 
   /** Gets the list of all resources recently access by user in order of access date time */
   @Override
-  public ResponseEntity<RecentResourceResponse> getUserRecentResources() {
+  public ResponseEntity<WorkspaceResourceResponse> getUserRecentResources() {
```
jaycarlton, line 170: If it's not too late, could you please split out the rename refactorings into a follow-on PR? It's a bit difficult to follow everything that's happening.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
```diff
@@ -3,14 +3,29 @@
...
+import org.pmiops.workbench.model.PrePackagedConceptSetEnum;
 import org.pmiops.workbench.utils.mappers.CommonMappers;
 
 @Mapper(
     componentModel = "spring",
     uses = {CommonMappers.class})
 public interface DataSetMapper {
 
+  @Mapping(target = "id", source = "dataSetId")
+  @Mapping(target = "conceptSets", ignore = true) // This is stored as a list of ids. Look those up in the controller for entities
```
jaycarlton, line 19: Since the mapper is a Spring bean itself, you can write a little method to do these mappings using the real DAO. 

When I use a mapper, I expect everything that can be mapped to be mapped.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
```diff
@@ -3,14 +3,29 @@
...
+  @Mapping(target = "cohorts", ignore = true) // This is stored as a list of ids. Look those up in the controller for entities
+  @Mapping(target = "domainValuePairs", ignore = true) // This is stored in a subtable, we may not want to fetch all the time
```
jaycarlton, line 21: OK. I think what we want then is another target class, such as DatasetLight. A mapper with a target class of `DataSet` should be expected to do whatever it takes to build a fully populated `DataSet`.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
```diff
@@ -3,14 +3,29 @@
...
+  @Mapping(target = "etag", source = "version")
+  DataSet dbModelToClient(DbDataset dbDataset);
+
+  default PrePackagedConceptSetEnum prePackagedConceptSetFromStorage(Short prePackagedConceptSet) {
```
jaycarlton, line 25: You shouldn't need to write this method. You can simply import the appropriate StorageEnums class and MapStruct will scan it to find a method from `Short` to `PrePackagedConceptSetEnum`.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -141,6 +161,13 @@ public WorkspacesController(
       BillingProjectBufferService billingProjectBufferService,
       WorkspaceService workspaceService,
       CdrVersionDao cdrVersionDao,
+      CohortMapper cohortMapper,
```
jaycarlton, line 164: The introduction of 7 new services at once makes me think perhaps there's an opportunity to define 2-3 high-level intermediate services.

We're going to break all of @ericsong's workspace tools again by increasing the argument list, and testing is going to get increasingly hairy.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
     return ResponseEntity.ok(recentWorkspaceResponse);
   }
 
+  private WorkspaceResource createWorkspaceResourceFromWorkspace(DbWorkspace workspace) {
```
jaycarlton, line 1047: This could be another (tiny) mapper.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+    return new WorkspaceResource()
+        .workspaceFirecloudName(workspace.getFirecloudName())
+        .workspaceId(workspace.getWorkspaceId())
+        .workspaceNamespace(workspace.getWorkspaceNamespace())
+        .workspaceBillingStatus(workspace.getBillingStatus());
+  }
+
+  @Override
+  public ResponseEntity<WorkspaceResourceResponse> getDataPageItems(
```
jaycarlton, line 1056: What about `getWorkspaceWithResources`? We try not to refer to UI constructs like individual pages/frames/panes from the backend.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+      String workspaceNamespace, String workspaceId) {
+    // This also enforces registered auth domain.
+    workspaceService.enforceWorkspaceAccessLevel(
+        workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
+
+    DbWorkspace workspace =
```
jaycarlton, line 1062: Please make this `final`.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+                        .cohortReview(cohortReview))
+            .collect(Collectors.toList()));
+    // TODO convert cohort reviews to resources
+    List<DbConceptSet> conceptSets =
+        conceptSetService.findByWorkspaceId(workspace.getWorkspaceId());
+    workspaceResources.addAll(
+        conceptSets.stream()
+            .map(conceptSetMapper::dbModelToClient)
+            .map(
+                conceptSet ->
```
jaycarlton, line 1094: This conversion can live in the ConceptSetMapper as well. We can have a `WorkspaceResource conceptSetToWorkspaceResource(DbConceptSet dbConceptSet, Workspace workspace)`. `WorkspaceMapper` has some examples of this.

We can do the same for the dataset, cohort, etc.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+  public ResponseEntity<WorkspaceResourceResponse> getDataPageItems(
+      String workspaceNamespace, String workspaceId) {
+    // This also enforces registered auth domain.
+    workspaceService.enforceWorkspaceAccessLevel(
+        workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
+
+    DbWorkspace workspace =
+        workspaceService.getRequiredWithCohorts(workspaceNamespace, workspaceId);
+    Set<DbCohort> cohorts = workspace.getCohorts();
+    WorkspaceResourceResponse workspaceResources = new WorkspaceResourceResponse();
```
jaycarlton, line 1065: Please name this `workspaceResourceResponse`, as it's not a collection.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+  @Override
+  public ResponseEntity<WorkspaceResourceResponse> getDataPageItems(
+      String workspaceNamespace, String workspaceId) {
+    // This also enforces registered auth domain.
+    workspaceService.enforceWorkspaceAccessLevel(
+        workspaceNamespace, workspaceId, WorkspaceAccessLevel.READER);
+
+    DbWorkspace workspace =
+        workspaceService.getRequiredWithCohorts(workspaceNamespace, workspaceId);
+    Set<DbCohort> cohorts = workspace.getCohorts();
```
jaycarlton, line 1064: We can actually stuff all of the rest of this code into the mappers, since it's just a fetch & hydrate block. There's no real decision-making logic going on here.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1044,76 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
+    workspaceResources.addAll(
+        cohorts.stream()
+            .map(cohortMapper::dbModelToClient)
+            .map(
+                cohort ->
+                    createWorkspaceResourceFromWorkspace(workspace)
+                        .cohort(cohort)
+                        .modifiedTime(cohort.getLastModifiedTime().toString()))
+            .collect(Collectors.toList()));
+    // TODO: Convert cohorts to resources
```
jaycarlton, line 1075: Can we bite this off all at once? How much does that blow up the ticket?

I'll be happy to help. I haven't actually written any backend code in like six weeks.

blrubenstein: That's a stale todo, this is already done :D

api/src/main/java/org/pmiops/workbench/api/UserMetricsController.java
blrubenstein: Sure. I can try to do that. We will have a bit of a weird in-between state if that's okay?

api/src/main/java/org/pmiops/workbench/cohortreview/CohortReviewMapper.java
```diff
@@ -0,0 +1,27 @@
...
+import org.pmiops.workbench.model.CohortReview;
+import org.pmiops.workbench.model.ReviewStatus;
+import org.pmiops.workbench.utils.mappers.CommonMappers;
+
+@Mapper(
+    componentModel = "spring",
+    uses = {CommonMappers.class})
+public interface CohortReviewMapper {
+  @Mapping(target = "etag", source = "version")
+  @Mapping(target = "queryResultSize", ignore = true) // used for pagination, we don't want this on all results
```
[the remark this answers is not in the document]
blrubenstein: I talked to Brian, and they're actually talking about moving all this out of the object anyway.

api/src/main/java/org/pmiops/workbench/cohortreview/CohortReviewMapper.java
```diff
@@ -0,0 +1,27 @@
...
+  @Mapping(target = "page", ignore = true) // used for pagination, we don't want this on all results
```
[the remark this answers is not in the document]
blrubenstein: I can change that. Will do.

Regarding the POJO, I don't want to spend too much time on this, given @freemabd1 says he's going to be refactoring this object soon anyway.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
blrubenstein: That is a behavior change from our current mapper, and I actually disagree in this case, because it would require more database lookups. I am okay making it a dataSetLight if we want though. The controller lookup for these objects is something @als364 recommended while we were pairing.

api/src/main/java/org/pmiops/workbench/utils/mappers/CommonMappers.java
```diff
@@ -34,6 +34,13 @@ public Long timestamp(Timestamp timestamp) {
     return null;
   }
 
+  public String timestampToString(Timestamp timestamp) {
```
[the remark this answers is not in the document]
blrubenstein: Mapstruct has a `toString` override, and it does the wrong thing. Or at least has different behavior than our current converter.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
blrubenstein: I think that sort of new service should be a new PR. I am incredibly worried about the amount of scope creep happening on this ticket.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
blrubenstein: Sure, I can do that.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
blrubenstein: It doesn't return a workspace though. Maybe `getWorkspaceResources`? Except it doesn't return all resources. (There was an hour or so brainstorming in slack, but we couldn't really come up with something I liked. If you have other ideas, I would love to hear them)

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
blrubenstein: Sure. Will do.

api/src/main/resources/workbench-api.yaml
blrubenstein: Talked offline! Renaming to `cdr-selectors`

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1010,6 +1043,70 @@ static String notebookLockingEmailHash(String bucket, String email) {
     return ResponseEntity.ok(recentWorkspaceResponse);
   }
 
+  @Override
+  public ResponseEntity<WorkspaceResourceResponse> getCdrSelectors(
+      String workspaceNamespace, String workspaceId) {
+    // This also enforces registered auth domain.
```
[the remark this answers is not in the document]
blrubenstein: Will do.

api/src/main/java/org/pmiops/workbench/cohortreview/CohortReviewMapper.java
```diff
@@ -13,12 +13,35 @@
     uses = {CommonMappers.class})
 public interface CohortReviewMapper {
   @Mapping(target = "etag", source = "version")
-  @Mapping(target = "queryResultSize", ignore = true) // used for pagination, we don't want this on all results
-  @Mapping(target = "page", ignore = true) // used for pagination, we don't want this on all results
-  @Mapping(target = "pageSize", ignore = true) // used for pagination, we don't want this on all results
-  @Mapping(target = "sortOrder", ignore = true) // used for pagination, we don't want this on all results
-  @Mapping(target = "sortColumn", ignore = true) // used for pagination, we don't want this on all results
-  @Mapping(target = "participantCohortStatuses", ignore = true) // this fetches all participants, and can be large, we don't want to fetch by default.
+  @Mapping(
```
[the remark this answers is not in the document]
blrubenstein: I am checking if there is a ticket for that now.

api/src/main/java/org/pmiops/workbench/utils/mappers/CommonMappers.java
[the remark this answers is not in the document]
jaycarlton: That may actually work if [XmlGregorianCalendar](https://docs.oracle.com/javase/7/docs/api/javax/xml/datatype/XMLGregorianCalendar.html) has the requisite precision.

api/src/main/java/org/pmiops/workbench/cohortreview/CohortReviewMapper.java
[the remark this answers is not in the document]
blrubenstein: Filed one https://precisionmedicineinitiative.atlassian.net/browse/RW-4706

api/src/main/java/org/pmiops/workbench/utils/mappers/CommonMappers.java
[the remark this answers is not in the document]
blrubenstein: Mainly it is different behavior from what we currently have. I would much rather use the same conversion method than what is already being used in the conversion, and clean up if we think we should later.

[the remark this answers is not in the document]
jaycarlton: Have you established that it's different functional behavior? Or just a different implementation? It would be neat to have a test case that runs through a bunch of timestamps to do the conversions.

The main advantage of MapStruct is that it makes the easy things trivial. If we don't trust it here, then it calls into question where we can trust it.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -100,6 +104,67 @@ void mergeResearchPurposeIntoWorkspace(
   WorkspaceResource workspaceResourceFromDbWorkspace(
       DbWorkspace dbWorkspace, WorkspaceAccessLevel accessLevel);
 
+  @Mapping(target = "workspaceFirecloudName", source = "dbWorkspace.firecloudName")
```
blrubenstein, line 107: @jaycarlton do you know if there's a way to avoid all this repeated boilerplate? It feels like it should be abstractable.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -1056,52 +1056,43 @@ static String notebookLockingEmailHash(String bucket, String email) {
...
             .map(
                 cohortReview ->
-                    workspaceMapper
-                        .workspaceResourceFromDbWorkspace(workspace, workspaceAccessLevel)
-                        .modifiedTime(cohortReview.getLastModifiedTime().toString())
-                        .cohortReview(cohortReview))
+                    workspaceMapper.workspaceResourceFromDbWorkspaceAndCohortReview(
+                        workspace,
+                        workspaceAccessLevel,
+                        cohortReviewMapper.dbModelToClient(cohortReview)))
```
blrubenstein, line 1073: @jaycarlton is there a way to do this in the mapper? Ideally it gets passed a dbModel object, and knows how to do the mapping?

api/src/main/java/org/pmiops/workbench/utils/mappers/CommonMappers.java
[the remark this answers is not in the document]
blrubenstein: I have just checked different implementation. I can check functional behavior.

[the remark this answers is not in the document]
blrubenstein: It is largely the same, with minor differences:
```
Gregorian: 2020-03-30T18:31:50.000Z
toString: 2020-03-30 18:31:50.0
```

I think gregorian is probably the right one to use, but should wait for the work on unifying date format.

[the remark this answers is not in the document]
jaycarlton: You're right; that's a functional change. Please put a comment with that difference in.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
jaycarlton: When the corresponding fields are named the same in the source and target, it just works.

api/src/main/java/org/pmiops/workbench/cdrselector/CdrSelectorService.java
```diff
@@ -0,0 +1,11 @@
+package org.pmiops.workbench.cdrselector;
+
+import org.pmiops.workbench.db.model.DbWorkspace;
+import org.pmiops.workbench.model.WorkspaceAccessLevel;
+import org.pmiops.workbench.model.WorkspaceResource;
+
+import java.util.List;
+
```
jaycarlton: A little comment about what a CDR selector is would help here.

Is `WorkspaceResource` already in use? If not, maybe we could change it (unless a CDR Selector is a subset of workspace resource).

api/src/main/java/org/pmiops/workbench/cdrselector/CdrSelectorServiceImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    workspaceResources.addAll(
+        conceptSets.stream()
+            .map(
+                dbConceptSet ->
+                    workspaceMapper.workspaceResourceFromDbWorkspaceAndConceptSet(
+                        dbWorkspace,
+                        workspaceAccessLevel,
+                        dbConceptSet))
+            .collect(Collectors.toList()));
+    List<DbDataset> datasets =
```
jaycarlton, line 71: Please put a blank line between each category.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -28,7 +32,7 @@
 @Mapper(
     componentModel = "spring",
     collectionMappingStrategy = CollectionMappingStrategy.TARGET_IMMUTABLE,
-    uses = {CommonMappers.class})
+    uses = {CommonMappers.class, CohortMapper.class, CohortReviewMapper.class, ConceptSetMapper.class, DataSetMapper.class})
```
jaycarlton: Please list these one per line like we do in import statements in the test classes.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -104,66 +108,69 @@ void mergeResearchPurposeIntoWorkspace(
...
   @Mapping(target = "cohortReview", ignore = true)
   @Mapping(target = "conceptSet", ignore = true)
   @Mapping(target = "dataSet", ignore = true)
   @Mapping(target = "notebook", ignore = true)
   // This should be set when the resource is set
-  @Mapping(target = "modifiedTime", source = "cohort.lastModifiedTime")
-  WorkspaceResource workspaceResourceFromDbWorkspaceAndCohort(
-      DbWorkspace dbWorkspace, WorkspaceAccessLevel accessLevel, Cohort cohort);
+  @Mapping(target = "modifiedTime", source = "dbCohort.lastModifiedTime")
+  WorkspaceResource workspaceResourceFromDbWorkspaceAndDbCohort(
```
jaycarlton, line 130: Our emerging naming convention for these appears to be xAndYToZ(), not zFromXAndY().

api/src/main/java/org/pmiops/workbench/workspaces/WorkspacesController.java
```diff
@@ -132,16 +133,10 @@
           .build();
 
   private final BillingProjectBufferService billingProjectBufferService;
+  private final CdrSelectorService cdrSelectorService;
```
jaycarlton, line 125: This is much cleaner. Thanks.

api/src/main/resources/workbench-api.yaml
```diff
@@ -5392,11 +5407,11 @@ definitions:
 
 '
         type: string
-  RecentResourceResponse:
+  WorkspaceResourceResponse:
```
jaycarlton, line 5410: I didn't look closely at `RecentResource` before, but I don't think it's a pattern we want to extend without careful consideration.

Instead of having an array of this type, I would've thought you'd want (instead of an overall array) a single object with an array of Cohort Reviews, Concept Sets, etc. You can certainly array-ify them on the front end if you need to.

If I'm reading this right, you've basically got a C language `union`-style structure, where at most one of each category is populated. That's not a great abstraction in a language where you don't have private fields. But I guess there's no polymorphism in Swagger.

ui/src/app/pages/data/data-page.tsx
```diff
@@ -107,38 +105,24 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
...
-      list = list.concat(convertToResources(conceptSets.items, namespace,
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.CONCEPTSET));
-      list = list.concat(convertToResources(dataSets.items, namespace,
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.DATASET));
+      const results = await workspacesApi().getCdrSelectors(namespace, id);
+
+      results.map(result => result.modifiedTime ? result.modifiedTime = new Date(Number(result.modifiedTime)).toDateString() : null);
       this.setState({
-        resourceList: list
+        existingCohortName: results.filter(resource => resource.cohort !== null && resource.cohort !== undefined)
```
jaycarlton, line 117: Yeah, this would be much easier with a single response object that has a typed list of every selector type.

api/src/main/java/org/pmiops/workbench/cdrselector/CdrSelectorService.java
blrubenstein: A CDR Selector is a subset of workspace resource.

api/src/main/resources/workbench-api.yaml
blrubenstein: Yeah, its not ideal. It would be good if we could completely change this, but that would be a breaking api change in several places. Which I am okay doing, but I think for now I'd rather roll it out using the existing structure. 

I could roll a new structure that uses this: https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ ? I don't know if that is usable in swagger codegen, but I can try. (Just call it `WorkspaceResourceV2` or something.)

blrubenstein: Discussed offline. Looks like polymorphism isn't supported in OpenApi 2, and moving to an object-aware model would be a big refactor on the front and backend, so that will spin off into its own ticket. https://precisionmedicineinitiative.atlassian.net/browse/RW-4740

blrubenstein: Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4740 to track this. Decided to wait until after OpenApi 3.0 and discussion on what a workspace is.

api/src/test/java/org/pmiops/workbench/utils/mappers/WorkspaceMapperTest.java
```diff
@@ -63,8 +68,15 @@
...
-  @MockBean({WorkspaceDao.class})
+  @Import({
+    CohortMapperImpl.class,
+    CohortReviewMapperImpl.class,
+    CommonMappers.class,
+    ConceptSetMapperImpl.class,
+    DataSetMapperImpl.class,
+    WorkspaceMapperImpl.class,
+  })
+  @MockBean({UserDao.class, WorkspaceDao.class})
```
blrubenstein, line 79: Our linter is opinionated about having these on one line. FYI.

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -3145,4 +3162,102 @@ public void copyNotebook_validateActiveBilling() {
...
+        workspaceResourceResponse.stream()
+            .filter(workspaceResource -> workspaceResource.getCohortReview() != null)
+            .collect(Collectors.toList());
+    List<WorkspaceResource> conceptSets =
+        workspaceResourceResponse.stream()
+            .filter(workspaceResource -> workspaceResource.getConceptSet() != null)
+            .collect(Collectors.toList());
+    List<WorkspaceResource> dataSets =
+        workspaceResourceResponse.stream()
+            .filter(workspaceResource -> workspaceResource.getDataSet() != null)
```
jaycarlton, line 3251: can you filter on the type code?

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -3145,4 +3162,102 @@ public void copyNotebook_validateActiveBilling() {
...
+            .collect(Collectors.toList());
+    assertThat(cohorts.size()).isEqualTo(1);
+    assertThat(cohorts.get(0).getCohort()).isEqualTo(cohort);
+    assertThat(cohortReviews.size()).isEqualTo(1);
+    compareCohortReviewFields(cohortReviews.get(0).getCohortReview(), cohortReview);
+    assertThat(conceptSets.size()).isEqualTo(1);
```
jaycarlton, line 3257: please use `assertThat(conceptSets).hasSize(1);` here and wherever you're checking a collection's size

ui/src/app/pages/data/data-page.tsx
```diff
@@ -107,38 +105,24 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
...
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.COHORT));
-      list = list.concat(convertToResources(cohortReviews.items, namespace,
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.COHORTREVIEW));
-      list = list.concat(convertToResources(conceptSets.items, namespace,
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.CONCEPTSET));
-      list = list.concat(convertToResources(dataSets.items, namespace,
-        id, accessLevel as unknown as WorkspaceAccessLevel, ResourceType.DATASET));
+      const results = await workspacesApi().getCdrSelectors(namespace, id);
+
+      results.map(result => result.modifiedTime ? result.modifiedTime = new Date(Number(result.modifiedTime)).toDateString() : null);
```
jaycarlton, line 115: if there's a ticket yet for how this may be changed, please put it in a comment.

api/src/main/java/org/pmiops/workbench/cdrselector/CdrSelectorServiceImpl.java
```diff
@@ -0,0 +1,87 @@
...
+      DataSetDao dataSetDao,
+      WorkspaceMapper workspaceMapper) {
+    this.cohortReviewService = cohortReviewService;
+    this.conceptSetService = conceptSetService;
+    this.dataSetDao = dataSetDao;
+    this.workspaceMapper = workspaceMapper;
+  }
+
+  @Override
+  public List<WorkspaceResource> getCdrSelectorsInWorkspace(
```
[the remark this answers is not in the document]
calbach: That's not something you can just casually tack on. It would involve such a fundamental redesign of the current model that it's not useful to speculate about how we might name that.

In frontend work, there's often a pattern of combining many things that are commonly request together into one endpoint to reduce rendering latency. The common thread connecting these things is that there is a page in the UI that is trying to render all of them at once. I would not overfit to that (alternatively: be transparent about overfitting) or you run the risk of being too inflexible when the product requirements inevitably evolve, e.g.: "going forward, lets split cohorts out into it's own tab" or "lets add notebooks here too".

Anyways, that's fine - I'll wait for the document.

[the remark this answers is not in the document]
blrubenstein: Hmmm... Let me actually give that a shot. I think it may be possible to have it take an array of Resource types. I will try to add that. If it is a lot of extra work, it may be worth splitting out until after discussion, but that sounds better. I pushed back against it a while ago, but thinking more, it shouldn't blow up too much scope if I just make the other types unsupported for now.

jaycarlton:
Maybe you could have it return a map from resource type to array of
workspace resources. Then you know how to down cast more easily.

But really, it may not be worth the cycles right now.

api/src/main/java/org/pmiops/workbench/dataset/DataSetMapper.java
```diff
@@ -3,14 +3,39 @@
...
+          true) // This is stored as a list of ids. Look those up in the controller for entities
+  @Mapping(
+      target = "cohorts",
+      ignore =
+          true) // This is stored as a list of ids. Look those up in the controller for entities
+  @Mapping(
+      target = "domainValuePairs",
+      ignore = true) // This is stored in a subtable, we may not want to fetch all the time
+  @Mapping(target = "etag", source = "version", qualifiedByName = "cdrVersionToEtag")
+  DataSet dbModelToClientLight(DbDataset dbDataset);
```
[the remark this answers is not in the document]
blrubenstein: The real dataset is swagger generated, so we can't do direct inheritance, but a parameterized wrapper would be doable. Will add a todo.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -84,6 +99,70 @@ void mergeResearchPurposeIntoWorkspace(
   @Mapping(target = "role", source = "acl")
   UserRole toApiUserRole(DbUser user, FirecloudWorkspaceAccessEntry acl);
 
+  @Mapping(target = "workspaceId", source = "dbWorkspace.workspaceId")
+  @Mapping(target = "workspaceFirecloudName", source = "dbWorkspace.firecloudName")
+  @Mapping(target = "workspaceBillingStatus", source = "dbWorkspace.billingStatus")
+  @Mapping(target = "permission", source = "accessLevel")
+  @Mapping(target = "cohort", source = "dbCohort")
+  // All workspaceResources have one object and all others are null. That should be
```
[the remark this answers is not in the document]
blrubenstein: That was leftover from a previous commit. Removing

api/src/main/java/org/pmiops/workbench/cdrselector/WorkspaceResourcesServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+          dataSetDao.findByWorkspaceIdAndInvalid(dbWorkspace.getWorkspaceId(), false);
+      workspaceResources.addAll(
+          datasets.stream()
+              .map(
+                  dbDataset ->
+                      workspaceMapper.dbWorkspaceAndDbDatasetToWorkspaceResource(
+                          dbWorkspace, workspaceAccessLevel, dbDataset))
+              .collect(Collectors.toList()));
+    }
+    if (resourceTypes.contains(ResourceType.COHORT_SEARCH_GROUP) || resourceTypes.contains(ResourceType.COHORT_SEARCH_ITEM) || resourceTypes.contains((ResourceType.NOTEBOOK)) || resourceTypes.contains((ResourceType.WORKSPACE))) {
```
jaycarlton, line 96: This statement is brittle, since we could add other resource types later and they might not get added here. If you really want to validate these things, I'd do a SUPPORTED_RESOURCE_TYPES.containsAll(resourceTypes) at the top.

But what might be the bigger flag here is that this endpoint only supports a subset of the enum values in `ResourceType`. I think it wold be simpler if you had a `WorkspaceResourceType` and supported all of those. Then you'd never be able to  have this error (by definition).

api/src/test/java/org/pmiops/workbench/workspaces/WorkspacesControllerTest.java
```diff
@@ -3145,4 +3162,102 @@ public void copyNotebook_validateActiveBilling() {
             workspacesController.copyNotebook(
                 workspace.getNamespace(), workspace.getId(), "z", copyNotebookRequest));
   }
+
+  private void compareCohortReviewFields(CohortReview cohortReview, CohortReview cohortReview2) {
```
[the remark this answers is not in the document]
blrubenstein: CohortReview is generated so we can't add `equals()`

api/src/main/resources/workbench-api.yaml
```diff
@@ -804,16 +804,22 @@ paths:
           description: Workspace deletion request accepted
           schema:
             "$ref": "#/definitions/EmptyResponse"
-  "/v1/workspaces/{workspaceNamespace}/{workspaceId}/cdr-selectors":
+  "/v1/workspaces/{workspaceNamespace}/{workspaceId}/workspaceResources":
```
jaycarlton, line 807: nit (if there's time): I'd probably go with just `resources` as the path segment, since we have it under `workspaces` anyway.

api/src/main/resources/workbench-api.yaml
```diff
@@ -804,16 +804,22 @@ paths:
...
     parameters:
       - "$ref": "#/parameters/workspaceNamespace"
       - "$ref": "#/parameters/workspaceId"
-    get:
+    post:
+      parameters:
+        - in: body
+          name: resourcesToFetch
```
jaycarlton, line 814: It's technically `resourceTypesToFetch`, since you're not naming individual resources.

api/src/main/resources/workbench-api.yaml
```diff
@@ -5407,6 +5413,10 @@ definitions:
 
 '
         type: string
+  WorkspaceResourcesRequest:
```
jaycarlton, line 5416: nit: for future-proofness, I might put your array in a member so you can add other kinds of criteria to this request later if needed.

ui/src/app/pages/data/data-page.tsx
```diff
@@ -110,7 +110,8 @@ export const DataPage = withCurrentWorkspace()(class extends React.Component<
       this.setState({
         isLoading: true
       });
-      const results = await workspacesApi().getCdrSelectors(namespace, id);
+      const results = await workspacesApi().getWorkspaceResources(namespace, id,
+        [ResourceType.COHORT, ResourceType.COHORTREVIEW, ResourceType.CONCEPTSET, ResourceType.DATASET]);
```
jaycarlton, line 114: I'd likely define this array someplace handy and call it `DATA_PAGE_WORKSPACE_RESORCE_TYPES` or some such, even if there's only one use. (My issue with the 2- or 3-use rule is that the next person may not bother to count the uses, and then it's like a riptide.)

api/src/main/java/org/pmiops/workbench/cdrselector/WorkspaceResourcesServiceImpl.java
blrubenstein: I think we will want to add the other resource types eventually. (I don't know some of them, @freemabd1 might know better) 

But I will move to the SUPPORTED_RESOURCE_TYPES model. I think a list of supported types is probably better than a list of unsupported types.

PR #3335, opened 2020-04-02
[RW-4561][risk=no] update DUA

als364:
localhost_4200_data-code-of-conduct
localhost_4200_data-code-of-conduct (1)

---
**PR checklist**

- [x] If this includes a new feature flag, I have created and linked new JIRA tickets to
   - [x] turn on the feature flag (RW-4697)
   - [x] remove it later (RW-4698)


ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -1,15 +1,53 @@
 import {Component} from '@angular/core';
```
als364, line 1: This is wrapped in Angular, so I can either 1) condition what content is shown within the file on the feature flag, or 2) branch what route we go to based on the feature flag in a few different places. I started out working on 1), but it would be relatively easy to switch to 2) if you, esteemed reviewer, think it would be cleaner.

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -21,70 +59,209 @@ interface Props {
...
+                     to my actions.
+                  </li>
+                </ul>
+                <div style={{...styles.bold, ...styles.smallTopMargin}}>
+                  I understand that failure to comply with these requirements may also carry
+                   financial or legal repercussions. Any misuse of the <i>All of Us</i> Research
+                   Hub, Researcher Workbench, or the <i>All of Us</i> Research Program data is
+                   taken very seriously and other sanctions may be sought.
+                </div>
+                <label style={{...styles.bold, ...styles.largeTopMargin}}>Authorized Data user Name</label>
```
calbach, line 202: user -> User

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -21,70 +59,209 @@ interface Props {
...
+                </TooltipTrigger>
+              </FlexRow>
+            </React.Fragment>
+          }
+        </FlexColumn>;
+      } else {
+        return <div style={dataUserCodeOfConductStyles.dataUserCodeOfConductPage}>
+          <DataUseAgreementContentV1/>
+          <div style={{height: '1rem'}}/>
+          {getDataUseAgreementWidgetV1.call(this,
```
calbach, line 248: Why invoke this via `call` rather than just invoking the function normally? If this is necessary, please add a comment to explain why.

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
calbach: I like this approach, unless we think there is some need to display the v2 DUA for some reason after launching the v3 DUA.

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -21,70 +59,209 @@ interface Props {
...
+        initialPublic: '',
+        page: DataUserCodeOfConductPage.CONTENT,
         submitting: false
       };
     }
 
-    submitDataUseAgreement(initials) {
+    submitDataUserCodeOfConduct(initials) {
       this.setState({submitting: true});
+      const dataUseAgreementVersion = serverConfigStore.getValue().enableV2DataUserCodeOfConduct ? 3 : 2;
```
calbach, line 89: I'd probably use V3 in the flag name instead of V2. Should hopefully be an easy search/replace. This line illustrates why the current naming could be confusing.

Alternatively: If the intent was to convey "v2" of the DUA implementation, perhaps something like enablePdfDataUser.... would be appropriate.

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -21,70 +59,209 @@ interface Props {
...
+                {submitting && <SpinnerOverlay/>}
+                <h1>Accept Data User Code of Conduct</h1>
+                <div style={{...styles.bold, ...styles.smallTopMargin}}>
+                  I
+                  <DuaTextInput style={{margin: '0 1ex'}}
+                     disabled
+                     value={profile.givenName + ' ' + profile.familyName}
+                     data-test-id='ducc-name-input'/>
+                   ("Authorized Data User") have personally reviewed this Data User Code of Conduct.
+                   I agree to follow each of the policies and prodedures it describes.
```
calbach, line 164: spelling: procedures

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -21,70 +59,209 @@ interface Props {
...
+                   User Code of Conduct may result in termination of my <i>All of Us</i> Research
+                   Program account and/or other sanctions, including, but not limited to:
+                </div>
+                <ul style={{...styles.bold, ...styles.smallTopMargin}}>
+                  <li>
+                    the posting of my name and affiliation on a publicly accessible list of
+                     violators, and
+                  </li>
+                  <li>
+                    notification of the National Institute of Health or other federal agencies as
```
calbach, line 192: Institutes

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
als364: This is preexisting - this was what was in the render function before I started this implementation. I'd prefer not to spend effort refactoring code that we have an open ticket to delete.

PR #3338, opened 2020-04-03
[RW-3128][RISK=NO] local rollback command(s)

jaycarlton:
Description:
Progress toward a general rollback capability for liquibase.
* project.rb support
* supports rollback by count & tag (though currently we aren't populating the tag column for all changesets)

It's my belief that we want to use tags, because if we could count, we wouldn't have become programmers. Seriously, though, there are failure modes where prod is at changeset 100, staging is at 103, and we want to use `count=4`. This works on staging (taking us back to 99), but blows up the prod database because we rolled back to 96.

However, we can introduce tags later.

Call it like 
```bash
# if we manually copy an ID to the tag column for changelog-128-add-columns-workspace
 ./project.rb run-liquibase --command=rollback --argument=changelog-128-add-columns-workspace

# or to use count
./project.rb run-liquibase --command=rollbackCount --argment=5
```
output (if you cancel) is [here](https://gist.github.com/jaycarlton/0ebbe49bcc81fbb8341e98df76c8970c).

[Another experiment](https://gist.github.com/jaycarlton/51c4d5f9452d85ff34817f167e54830e), adding two Liquibase tags in separate changesets to the test db nad removing them one at a time.

In order to run on a non-local environment do
```
./project.rb run-liquibase --command=status --project=all-of-us-workbench-test
```
where the `project` argument is the full project path (or just `local`)

Since this wraps the gradle file directly, you can see all the Liquibase commands by doing
```
./project.rb run-liquibase --command=tasks
```
and looking under "Liquibase tasks".

---
**PR checklist**


api/db/build.gradle
```diff
@@ -11,16 +11,16 @@ apply plugin: 'org.liquibase.gradle'
 
 def db_host = System.getenv("DB_HOST") ?: "db"
 def db_port = System.getenv("DB_PORT") ?: "3306"
+def liquibase_username = "liquibase"
 def liquibase_password = System.getenv("LIQUIBASE_DB_PASSWORD") ?: "lb-notasecret"
 
 liquibase {
     activities {
         main {
```
jaycarlton, line 19: There's still an issue with this block, at least accorrding to IntelliJ's gradle plugin

api/db/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Wed Apr 01 12:34:20 EDT 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-all.zip
```
jaycarlton, line 2: I did this to make the `db/build.gradle` project importable in IntelliJ. I'm not sure this is the right way to do it or if I still need it. [#dev-productivity thread](https://pmi-engteam.slack.com/archives/CN4AX0K9V/p1585759155001600).

api/db/run-migrations.sh
```diff
@@ -22,7 +16,9 @@ trap finish EXIT
 envsubst < "$(dirname "${BASH_SOURCE}")/create_db.sql" > $CREATE_DB_FILE
 
 echo "Creating database if it does not exist..."
+# This command is run regardless, and we rely on the commands in CREATE_DB_FILE being idempotent.
 mysql -h ${DB_HOST} --port ${DB_PORT} -u root -p${MYSQL_ROOT_PASSWORD} < ${CREATE_DB_FILE}
```
jaycarlton, line 20: I want to break this script up into one that builds the database and one that runs a liquibase command. Even better if we can drive liquibase without Bash or Ruby.

api/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu May 10 14:42:28 CDT 2018
+#Wed Apr 01 12:34:20 EDT 2020
```
jaycarlton, line 1: I don't actually know how this file works, but I don't remember making this change...

api/libproject/devstart.rb
```diff
@@ -634,6 +634,50 @@ def run_local_all_migrations()
...
+      Integer,
+      ->(opts, c) { opts.count = c},
+      "Count of changesets to roll back from end.")
+
+  op.add_typed_option(
+      "--tag [tag]",
+      String,
+      ->(opts, t) { opts.tag = t},
+      "Liquibase changeset tag to roll back to.")
+  op.add_validator ->(opts) {
```
jaycarlton, line 652: This was a pain to write, and Zeus help me if I want another option. The idea is that exactly one is provided.

Also, I want a general liquibase command. I'll do that tomorrow just for fun.

api/db/run-migrations.sh
```diff
@@ -2,15 +2,9 @@
...
-if [ -z ${2+x} ]
-then
-    context=""
-else
-    context="-Pcontexts=$2"
-fi
-
-# Ruby is not installed in our dev container and this script is short, so bash is fine.
+liquibaseCommand="${1-update}"
+runList="-PrunList=${2-main}"
```
jaycarlton, line 6: We actually don't need to provide this explicitly, since the only activity we have is `main`. That would simplify things considerably, though I don't want to sacrifice generality. It would be more robust to set `main` as the default as well, if that's supported.

api/db/run-migrations.sh
```diff
@@ -2,15 +2,9 @@
...
+liquibaseArgs="${3--PliquibaseCommandValue=}"
```
jaycarlton, line 7: This is really gross, but it works. I.e. we are handing the liquibase plugin an empty command value, and it doesn't break anything.

This script is way past my Bash expertise/tolerance. If I need to keep it, I want to port to Ruby or Python, or even pure Gradle.

api/libproject/devstart.rb
```diff
@@ -634,6 +634,50 @@ def run_local_all_migrations()
...
+  ensure_docker_sync
+  op = WbOptionsParser.new(cmd_name, args)
+  op.add_typed_option(
+      "--count=[count]",
+      Integer,
+      ->(opts, c) { opts.count = c},
+      "Count of changesets to roll back from end.")
+
+  op.add_typed_option(
+      "--tag [tag]",
```
calbach, line 648: Perhaps a more specific flag name, e.g. --to-tag ; currently it's slightly ambiguous as to whether it's rolling back TO this tag, or rolling back JUST this tag.

api/libproject/devstart.rb
```diff
@@ -634,6 +634,116 @@ def run_local_all_migrations()
...
+
+  context = GcloudContextV2.new(op)
+  context.validate
+
+  with_optional_cloud_proxy_and_db(context, nil, 'sa-key.json') do |gcc|
+    common.status('inside with_optional_cloud_proxy_and_db')
+    common.status("project: #{gcc.project}, account: #{gcc.account}, creds_file: #{gcc.creds_file}, dir: #{Dir.pwd}")
+    command = op.opts.command
+
+    Dir.chdir('db') # 'cd' isn't available to run inline apparently
```
calbach, line 722: I believe it executes a subprocess, presumably it didn't say `cd not found`, rather it didn't have any effect?

api/libproject/devstart.rb
```diff
@@ -1999,6 +2109,19 @@ def with_cloud_proxy_and_db(gcc, service_account = nil, key_file = nil)
   end
 end
 
+def with_optional_cloud_proxy_and_db(gcc, service_account = nil, key_file = nil)
```
calbach, line 2112: thanks, not a bad solution

api/libproject/devstart.rb
```diff
@@ -634,6 +634,116 @@ def run_local_all_migrations()
...
+  }
+  op.parse.validate
+
+  # Currently there's only one activity (main), and leaving out the runList arugment causes
+  # it to run that activity. However, that's not because it's detected as default or primary, but
+  # leaving out the runList is equivalent to specifying all of the activities to run in unspecified
+  # order.
+  # https://github.com/liquibase/liquibase-gradle-plugin#3-configuring-the-plugin
+  if op.opts.run_list.nil? || op.opts.run_list.empty?
+    run_list = "main"
```
calbach, line 698: Does this duplicate the effect that `gradle.properties` is having? I assume that's just setting a default

api/libproject/devstart.rb
```diff
@@ -634,6 +634,116 @@ def run_local_all_migrations()
...
+# statements and ask user for verification.
+def run_liquibase(cmd_name, *args)
+  command_to_sql = {
+      'changelogSync' => 'changelogSyncSQL',
+      'markNextChangesetRan' => 'markNextChangesetRanSQL',
+      'rollback' => 'rollbackSQL',
+      'rollbackCount' => 'rollbackCountSQL',
+      'rollbackToDate' => 'rollbackToDateSQL',
+      'update' => 'updateSQL',
+      'updateCount' => 'updateCountSql',
```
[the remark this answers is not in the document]
jaycarlton: So it's documented as `updateCountSql` [officially](https://www.liquibase.org/documentation/command_line.html), but the task in the plugin is `updateCountSql`.
```
updateCountSql - Writes SQL to apply the next <liquibaseCommandValue> change sets to STDOUT.
```
I'll add a comment bemoaning this situation.

api/libproject/devstart.rb
jaycarlton: Oh good point. I just added the gradle.properties to shut up the IJ plugin so it would let me add `db` as a gradle project. Let me play with this.

api/libproject/devstart.rb
jaycarlton: If I change that line to `common.run_inline(%W{cd db})`, I get

```
project: local, account: jay.carlton@pmi-ops.org, creds_file: , dir: /w/api
+ cd db
Traceback (most recent call last):
	8: from /w/api/project.rb:6:in `<main>'
	7: from /w/aou-utils/workbench.rb:61:in `handle_argv_or_die'
	6: from /w/aou-utils/utils/common.rb:86:in `handle_or_die'
	5: from /w/api/libproject/devstart.rb:745:in `block in <top (required)>'
	4: from /w/api/libproject/devstart.rb:717:in `run_liquibase'
	3: from /w/api/libproject/devstart.rb:2117:in `with_optional_cloud_proxy_and_db'
	2: from /w/api/libproject/devstart.rb:723:in `block in run_liquibase'
	1: from /w/aou-utils/utils/common.rb:157:in `run_inline'
/w/aou-utils/utils/common.rb:157:in `spawn': No such file or directory - cd (Errno::ENOENT)
```

calbach: Well.. consider me confused as well

PR #3342, opened 2020-04-03
[risk=no][RW-3211]Don't include free tier billing account if the user is out of credits

jmthibault79:
Description:

API portion of RW-3211: don't return free tier billing account in the account options for Create Workspace if the user is expired.

This API is only called by the UI when enableBillingLockout is enabled, so we can assume this.

If I had a normal-powered computer I would prefer to combine this with the UI, but running both UI and API servers locally is moderately painful right now.  Merging this first will let me run UI only, speeding up my dev cycle. :)

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/UserController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/UserController.java
@@ -12,8 +12,8 @@ import java.util.Optional;
 import java.util.function.Function;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.inject.Provider;
+import org.pmiops.workbench.billing.FreeTierBillingService;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.UserService;
 import org.pmiops.workbench.db.model.DbUser;
@@ -51,6 +51,7 @@ public class UserController implements UserApiDelegate {
   private final Provider<WorkbenchConfig> configProvider;
   private final UserService userService;
   private final FireCloudService fireCloudService;
+  private final FreeTierBillingService freeTierBillingService;
 
   private final Provider<Cloudbilling> cloudBillingProvider;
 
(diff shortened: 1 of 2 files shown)


jmthibault79:
Test failure is unrelated:
```
org.pmiops.*********.api.CohortBuilderControllerBQTest > countSubjectsDrugChildAgeAtEvent FAILED
    expected: 1
    but was : 0
        at org.pmiops.*********.api.CohortBuilderControllerBQTest.assertParticipants(CohortBuilderControllerBQTest.java:2099)
        at org.pmiops.*********.api.CohortBuilderControllerBQTest.countSubjectsDrugChildAgeAtEvent(CohortBuilderControllerBQTest.java:1611)
```

api/src/main/java/org/pmiops/workbench/api/UserController.java
```diff
@@ -145,31 +148,41 @@ private BillingAccount freeTierBillingAccount() {
 
   @Override
   public ResponseEntity<WorkbenchListBillingAccountsResponse> listBillingAccounts() {
+    final List<BillingAccount> billingAccounts = Lists.newArrayList();
+
+    // if we show the free credits account, it goes first
+
+    if (freeTierBillingService.userHasRemainingFreeTierCredits(userProvider.get())) {
```
calbach, line 155: This is actually not quite the right logic. Free credits are attributed to the workspace creator.  Therefore the right thing to check would be the creator's credits.  This may require a bit more thinking and maybe a quick chat with Karthik to think about how we can handle this cleanly.

i.e. perhaps only the creator can touch the billing settings - OR perhaps only the creator sees the free tier settings (unless a workspace currently free tier, then it should always display)

In all cases, I think there are valid situations where you'd want to show the free tier account in the list despite the user being over the limit, so I think we'll want to not have the logic here.

jmthibault79:
I made some incorrect assumptions about when listBillingAccounts() is called so I'll have to rethink my approach.

PR #3347, opened 2020-04-03
[noTicket][risk=moderate]Cloud interceptor should have correct Rdr Export queue name

NehaBroad:
Cloud queue( responsible for handling task to export workspace/researcher data to Rdr ) name was recently updated from rdrQueueTest to rdrExportQueue. For all the incoming request dispatched from queue, cloud task interceptor confirms that request is coming from valid queue.

This PR is to update the name of queue in CloudTaskInterceptor

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/CloudTaskInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/CloudTaskInterceptor.java
@@ -23,7 +23,7 @@ public class CloudTaskInterceptor extends HandlerInterceptorAdapter {
   public static final String QUEUE_NAME_REQUEST_HEADER = "X-AppEngine-QueueName";
   private static final String CLOUD_TASK_TAG = "cloudTask";
   public static final Set<String> VALID_QUEUE_NAME_SET =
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/interceptors/CloudTaskInterceptor.java
```diff
@@ -23,7 +23,7 @@
   public static final String QUEUE_NAME_REQUEST_HEADER = "X-AppEngine-QueueName";
   private static final String CLOUD_TASK_TAG = "cloudTask";
   public static final Set<String> VALID_QUEUE_NAME_SET =
-      new HashSet<>(Arrays.asList("rdrQueueTest"));
+      new HashSet<>(Arrays.asList("rdrExportQueue"));
```
calbach, line 26: Should we just be pulling this out of the config instead? If that's not possible, please add a comment that this needs to be kept in sync with the config/queue yaml values.

NehaBroad: Since the queue name is same for all env i am planning to extracting that in the controller itself. (Follow up PR) as of now i have added comments

PR #3349, opened 2020-04-04
[RW-4709][risk=no] Add jest-junit reporter in e2e project

aweng98:


e2e/package.json
```diff
@@ -39,7 +43,7 @@
     "faker": "^4.1.0",
     "fs-extra": "^9.0.0",
     "jest": "^25.1.0",
-    "jest-cli": "^25.1.0",
```
aweng98, line 42: I am removing `jest-cli` for now because I haven't used any of its feature.

.circleci/config.yml
```diff
@@ -298,15 +298,15 @@ jobs:
           command: git submodule update --init --recursive
       - restore_cache:
           keys:
-          - e2e-test-cache-{{ checksum "~/workbench/e2e/yarn.lock" }}
-          - e2e-test-cache-
+          - v1-e2e-test-cache-{{ checksum "~/workbench/e2e/yarn.lock" }}
```
aweng98, line 301: Adding a version number. From [CircleCI official doc](https://circleci.com/docs/2.0/caching/):
`If you would like to invalidate your cache manually, you can bump v1 to v2 in your config.yml file. In this case, you would now have a new cache key v2-npm-deps, which will trigger the storing of a new cache.`

.circleci/config.yml
```diff
@@ -317,9 +317,12 @@ jobs:
       - run:
           name: Run e2e tests (Puppeteer and Jest)
           working_directory: ~/workbench/e2e
-          command: yarn --silent test:ci
+          command: yarn test:ci
       - store_artifacts:
           path: e2e/logs
+          destination: logs
+      - store_test_results:
+          path: ~/workbench/e2e/logs
```
blrubenstein, line 325: Just to check, this is writing to circle, right? and nowhere else?

aweng98: In `e2e` project, any screenshot or/and html file are saved in `/logs` sub-directory. If you run tests on localhost, you'll find failed screenshots and html files in `/logs` dir. 
We have to set `store_test_results` key to tell CircleCI where to find saved files.

PR #3354, opened 2020-04-06
[RW-4170][Risk=moderate] Cherry pick changes to fix RDR export on Stable

NehaBroad:
Cherry Pick commit to change the name of queue used in cloudinterceptor to ensure the request is coming from valid source


gjuggler:
Just a quick note before this goes in – can you associate this PR with a Jira ticket and add a risk annotation?

PR #3361, opened 2020-04-07
[no ticket][risk=no] fix version skew

als364:
Fixed an issue with version skew on the DUCC version - I had changed the frontend value but not the backend value.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -75,7 +75,6 @@ import org.springframework.transaction.annotation.Transactional;
 public class UserServiceImpl implements UserService, GaugeDataCollector {
 
   private final int MAX_RETRIES = 3;
-  private static final int CURRENT_DATA_USE_AGREEMENT_VERSION = 2;
   private static final int CURRENT_TERMS_OF_SERVICE_VERSION = 1;
 
   private final Provider<WorkbenchConfig> configProvider;
@@ -127,6 +126,10 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
     this.complianceService = complianceService;
     this.directoryService = directoryService;
     this.institutionService = institutionService;
+  }
+
+  private int getCurrentDuccVersion() {
+    return configProvider.get().featureFlags.enableV3DataUserCodeOfConduct ? 3 : 2;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -129,6 +128,10 @@ public UserServiceImpl(
     this.institutionService = institutionService;
   }
 
+  private int getCurrentDuccVersion() {
+    return configProvider.get().featureFlags.enableV3DataUserCodeOfConduct ? 3 : 2;
```
jaycarlton, line 132: Can we just keep the version directly in either the config file (simplest for now)? In the short/medium term, I'd really like to see a table in the database for the different singable compliance documents and their active & retired versions that each document view can refer to and user signature record can link to?

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -75,7 +75,6 @@
 public class UserServiceImpl implements UserService, GaugeDataCollector {
 
   private final int MAX_RETRIES = 3;
-  private static final int CURRENT_DATA_USE_AGREEMENT_VERSION = 2;
   private static final int CURRENT_TERMS_OF_SERVICE_VERSION = 1;
```
jaycarlton, line 78: Presumably this could also bite us.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
als364: We cannot, because it's feature flagged.

jaycarlton: We can add an integer to the config file without breaking things that don't know about it, right? Or are you saying that it exists in both states at once, guarded by a feature flag? In that case, we should remove it from the flag and just have a value for the active version.

In other words, we really shouldn't have code like this line. There's just no way it's going to age well.

als364: I'm saying that if I follow your suggestion, it will break the feature until the feature flag is turned on everywhere.

jaycarlton: Oh. You could use a feature flag for `enableDataUserCodeOfConductBeta`, and then have two values like `dataUserCodeOfConductBetaVersion` and `dataUserCodeOfConductCurrentVersion`. Then that line of code could keep working when one or both version keys changes.

PR #3365, opened 2020-04-08
[RW-4736][risk=no] Puppeteer test: waitFor institution email validation

aweng98:
Changes will also resolve ticket RW-4648.

### DIFF
--- a/e2e/app/aou-elements/button.ts
+++ b/e2e/app/aou-elements/button.ts
@@ -28,12 +28,11 @@ export default class Button extends BaseElement {
    * Wait until button is clickable (enabled).
    */
   async waitUntilEnabled(): Promise<unknown> {
-    const result = await this.page.waitForFunction((e) => {
+    const retrievedValue = await this.page.waitForFunction((e) => {
       const style = window.getComputedStyle(e);
       return style.getPropertyValue('cursor') === 'pointer';
     }, {}, this.element);
-    const jval = await result.jsonValue();
-    return jval;
+    return await retrievedValue.jsonValue();
   }
 
 }
--- a/e2e/app/aou-elements/button.ts
+++ b/e2e/app/aou-elements/button.ts
@@ -1,4 +1,4 @@
-import {JSHandle, Page, WaitForSelectorOptions} from 'puppeteer';
+import {Page, WaitForSelectorOptions} from 'puppeteer';
 import TextOptions from './text-options';
 import BaseElement from './base-element';
 import {findButton} from './xpath-finder';
@@ -27,12 +27,13 @@ export default class Button extends BaseElement {
   /**
    * Wait until button is clickable (enabled).
    */
-  async waitUntilEnabled(): Promise<JSHandle> {
-    const handle = this.element.asElement();
-    return await handle.evaluateHandle((e) => {
+  async waitUntilEnabled(): Promise<unknown> {
+    const result = await this.page.waitForFunction((e) => {
       const style = window.getComputedStyle(e);
       return style.getPropertyValue('cursor') === 'pointer';
-    }, this.element);
+    }, {}, this.element);
+    const jval = await result.jsonValue();
+    return jval;
   }
 
 }
--- a/e2e/app/aou-elements/clr-icon-link.ts
+++ b/e2e/app/aou-elements/clr-icon-link.ts
@@ -1,23 +1,24 @@
 import {Page, WaitForSelectorOptions} from 'puppeteer';
 import BaseElement from './base-element';
+import TextOptions from './text-options';
 import {findIcon} from './xpath-finder';
 
 export default class ClrIconLink extends BaseElement {
    
   static async forLabel(
      page: Page,
-     aElementName: string,
+     textOptions: TextOptions,
      shape: string,
      waitOptions: WaitForSelectorOptions = {visible: true},
      throwErr = true): Promise<ClrIconLink> {
 
     let element: ClrIconLink;
     try {
-      const iconElement = await findIcon(page, aElementName, shape, waitOptions);
+      const iconElement = await findIcon(page, textOptions, shape, waitOptions);
       element = new ClrIconLink(page, iconElement);
     } catch (e) {
       if (throwErr) {
-        console.error(`FAILED finding a clr-icon: "${aElementName}".`);
+        console.error(`FAILED finding a clr-icon: "${textOptions}".`);
         throw e;
       }
     }
--- a/e2e/app/aou-elements/xpath-defaults.ts
+++ b/e2e/app/aou-elements/xpath-defaults.ts
@@ -2,11 +2,11 @@ import TextOptions from './text-options';
 
 function textXpathHelper(opts: TextOptions) {
   if (opts.text) {
-    return `text()="${opts.text}" or @aria-label="${opts.text}" or @placeholder="${opts.text}"`;
+    return `text()='${opts.text}' or @aria-label='${opts.text}' or @placeholder='${opts.text}'`;
   } else if (opts.textContains) {
-    return `contains(text(),"${opts.textContains}") or contains(@aria-label,"${opts.textContains}") or contains(@placeholder,"${opts.textContains}")`;
+    return `contains(text(),'${opts.textContains}') or contains(@aria-label,'${opts.textContains}') or contains(@placeholder,'${opts.textContains}')`;
   } else if (opts.normalizeSpace) {
-    return `contains(normalize-space(), "${opts.normalizeSpace}")`;
+    return `contains(normalize-space(), '${opts.normalizeSpace}')`;
   }
 }
 
@@ -59,10 +59,11 @@ export function clickableXpath(label: string) {
  * @param label:
  * @param shapeValue:
  */
-export function clrIconXpath(label: string, shapeValue: string) {
-  if (label === '') {
+export function clrIconXpath(opts: TextOptions, shapeValue: string) {
+  if (opts === undefined) {
     return `//clr-icon[@shape='${shapeValue}'][*[@role='img']]`; // anywhere on page
   }
   // next to a label
-  return `//*[normalize-space()='${label}']/ancestor::node()[1]//clr-icon[@shape='${shapeValue}'][*[@role='img']]`;
+  let nodeLevel = opts.ancestorNodeLevel || 1;
+  return `//*[${textXpathHelper(opts)}]/ancestor::node()[${nodeLevel}]//clr-icon[@shape='${shapeValue}'][*[@role='img']]`;
 }
--- a/e2e/app/aou-elements/xpath-defaults.ts
+++ b/e2e/app/aou-elements/xpath-defaults.ts
@@ -60,7 +60,7 @@ export function clickableXpath(label: string) {
  * @param shapeValue:
  */
 export function clrIconXpath(opts: TextOptions, shapeValue: string) {
-  if (opts === undefined) {
+  if (Object.keys(opts).length == 0) {
     return `//clr-icon[@shape='${shapeValue}'][*[@role='img']]`; // anywhere on page
   }
   // next to a label
--- a/e2e/app/aou-elements/xpath-defaults.ts
+++ b/e2e/app/aou-elements/xpath-defaults.ts
@@ -1,12 +1,16 @@
 import TextOptions from './text-options';
 
 function textXpathHelper(opts: TextOptions) {
+  let txt;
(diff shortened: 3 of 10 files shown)


e2e/app/aou-elements/clr-icon-link.ts
```diff
@@ -1,23 +1,24 @@
...
-     aElementName: string,
+     textOptions: TextOptions,
      shape: string,
      waitOptions: WaitForSelectorOptions = {visible: true},
      throwErr = true): Promise<ClrIconLink> {
 
     let element: ClrIconLink;
     try {
-      const iconElement = await findIcon(page, aElementName, shape, waitOptions);
+      const iconElement = await findIcon(page, textOptions, shape, waitOptions);
```
jaycarlton, line 17: Do you need to catch and rethrow, or will the framework & language do most of this for you?

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -2,11 +2,11 @@ import TextOptions from './text-options';
 
 function textXpathHelper(opts: TextOptions) {
   if (opts.text) {
-    return `text()="${opts.text}" or @aria-label="${opts.text}" or @placeholder="${opts.text}"`;
+    return `text()='${opts.text}' or @aria-label='${opts.text}' or @placeholder='${opts.text}'`;
   } else if (opts.textContains) {
-    return `contains(text(),"${opts.textContains}") or contains(@aria-label,"${opts.textContains}") or contains(@placeholder,"${opts.textContains}")`;
+    return `contains(text(),'${opts.textContains}') or contains(@aria-label,'${opts.textContains}') or contains(@placeholder,'${opts.textContains}')`;
```
jaycarlton, line 7: nit: space after commas

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -2,11 +2,11 @@ import TextOptions from './text-options';
 
 function textXpathHelper(opts: TextOptions) {
   if (opts.text) {
-    return `text()="${opts.text}" or @aria-label="${opts.text}" or @placeholder="${opts.text}"`;
+    return `text()='${opts.text}' or @aria-label='${opts.text}' or @placeholder='${opts.text}'`;
```
jaycarlton, line 5: nit: can you declare a string for `'${opts.text}'` and use it in all three places?

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -59,10 +59,11 @@ export function clickableXpath(label: string) {
  * @param label:
  * @param shapeValue:
  */
-export function clrIconXpath(label: string, shapeValue: string) {
-  if (label === '') {
+export function clrIconXpath(opts: TextOptions, shapeValue: string) {
```
jaycarlton, line 62: Is xpath the only way to reach this icon? I feel like it may be brittle in the face of future page reorganization and isn't the most readable representation for folks trying to investigate failed tests.

Also, I don't know what clr means.

e2e/app/authenticated-page.ts
```diff
@@ -132,7 +132,7 @@ export default abstract class AuthenticatedPage extends BasePage {
     const is = await this.isSideNavDropdownOpen();
     if (!is) {
       // click bars icon to open dropdown
-      const icon = await findIcon(this.page, '', 'bars');
+      const icon = await findIcon(this.page, {}, 'bars');
```
jaycarlton, line 135: Can you name this `barsIcon` or `hamburgerIcon`?

e2e/app/authenticated-page.ts
```diff
@@ -132,7 +132,7 @@ export default abstract class AuthenticatedPage extends BasePage {
     const is = await this.isSideNavDropdownOpen();
```
jaycarlton, line 132: nit: name this `sideNavIsOpen`

e2e/app/authenticated-page.ts
```diff
@@ -179,7 +179,7 @@ export default abstract class AuthenticatedPage extends BasePage {
 
   private async isSideNavDropdownOpen(): Promise<boolean> {
```
jaycarlton, line 180: Could you please add a description, like "the side Nav for this page is either docked to the right or open. We can determine the open state by looking for a visible Home icon" Or something. It's not obvious to someone who doesn't watch the UI closely.

Additionally, it's confusing that the method is named after the  side nav dropdown, but the implementation is in terms of a  home icon. Can we add a test data tag to the nav and check for it more directly?

e2e/app/create-account-page.ts
```diff
@@ -154,7 +154,7 @@ export default class CreateAccountPage extends BasePage {
     const emailAddress = await Textbox.forLabel(this.page, {textContains: FIELD_LABEL.INSTITUTION_EMAIL, ancestorNodeLevel: 2});
```
jaycarlton, line 154: I would pass an institution input object into this method so that you can fill it out more than one way to test edge cases, validation, etc, if you want.

e2e/app/home-page.ts
```diff
@@ -32,7 +32,7 @@ export default class HomePage extends AuthenticatedPage {
   }
 
   async getCreateNewWorkspaceLink(): Promise<ElementHandle> {
-    return findIcon(this.page, FIELD_LABEL.CREATE_NEW_WORKSPACE, 'plus-circle');
+    return findIcon(this.page, {text: FIELD_LABEL.CREATE_NEW_WORKSPACE}, 'plus-circle');
```
jaycarlton, line 35: nit: could we make an enum  with these icon type strings?

e2e/tests/user/registration-ui.spec.ts
```diff
@@ -1,7 +1,5 @@
 import BaseElement from '../../app/aou-elements/base-element';
-import SelectComponent from '../../app/aou-elements/select-component';
-import Textbox from '../../app/aou-elements/textbox';
-import CreateAccountPage, {FIELD_LABEL, INSTITUTION_ROLE_VALUE, INSTITUTION_VALUE} from '../../app/create-account-page';
+import CreateAccountPage from '../../app/create-account-page';
```
jaycarlton, line 2: Can we use module exports or something so we don't have to import from grandparent directories? It's really breakable when you do that, because it means you can't move your test directories around easily.

e2e/tests/user/registration-ui.spec.ts
```diff
@@ -90,14 +88,7 @@ describe('User registration tests:', () => {
...
-    await institutionSelect.select(INSTITUTION_VALUE.BROAD);
-
-    const emailAddress = await Textbox.forLabel(page, {textContains: FIELD_LABEL.INSTITUTION_EMAIL, ancestorNodeLevel: 2});
-    await emailAddress.type(configs.broadInstitutionEmail);
-
-    const roleSelect = new SelectComponent(page, 'describes your role');
-    await roleSelect.select(INSTITUTION_ROLE_VALUE.UNDERGRADUATE_STUDENT);
+    await createAccountPage.fillOutInstitution();
 
     await nextButton.waitUntilEnabled();
```
jaycarlton, line 93: nit: there are two different conventions on this line and the next for async functions. I would have expected to see `await nextButton.enabled();` for consistency.

e2e/app/create-account-page.ts
jaycarlton, line 154: nit: I'd rename `emailAddress` to `emailAddressTextBox` to distinguish it from the address itself.

e2e/app/authenticated-page.ts
```diff
@@ -179,7 +179,7 @@ export default abstract class AuthenticatedPage extends BasePage {
...
     try {
-      await findIcon(this.page, 'Home', 'home', {visible: true, timeout: 1000});
+      await findIcon(this.page, {text: 'Home'}, 'home', {visible: true, timeout: 1000});
```
jaycarlton, line 182: can you make a utility method that's an async boolean-valued method to wait look for and wait for an icon (or other element)? It seems like this would be a common itch to scratch.

e2e/app/authenticated-page.ts
aweng98: sorry. your feedback is awesome! I have comments for most "big" methods but don't do for all methods because I think some methods are easy to know what it suppose to do. For this method, I added what you said as comment. 

It is the easiest way to determine the open or close state by looking for visible `Home` icon. If someone is going to implement a test friendly tag in future, I'd be happy to update this code.

e2e/app/create-account-page.ts
aweng98: I can refactor in a new PR. This PR only deals with one test failure that only happens on CircleCi.

e2e/app/home-page.ts
aweng98: good suggestion. I can refactor code in a new PR.

e2e/tests/user/registration-ui.spec.ts
aweng98: don't know how to do what you suggested. but I will research around for how-to to see what other people are doing.

e2e/app/aou-elements/button.ts
```diff
@@ -27,12 +27,13 @@ export default class Button extends BaseElement {
...
-    const handle = this.element.asElement();
-    return await handle.evaluateHandle((e) => {
+  async waitUntilEnabled(): Promise<unknown> {
+    const result = await this.page.waitForFunction((e) => {
       const style = window.getComputedStyle(e);
       return style.getPropertyValue('cursor') === 'pointer';
-    }, this.element);
+    }, {}, this.element);
+    const jval = await result.jsonValue();
+    return jval;
```
aweng98, line 36: Request `GET /v1/institutions/Broad/checkEmail/aou-dev-registration@broadinstitute.org/` could takes up to several seconds to complete. Changed to `waitForFunction` for waiting function returns true, up to 30 seconds.

e2e/app/aou-elements/clr-icon-link.ts
aweng98: `jest` will throws exception too. I'm printing out `textOptions` for manual checking if fail

e2e/app/aou-elements/button.ts
```diff
@@ -27,12 +27,13 @@ export default class Button extends BaseElement {
   /**
    * Wait until button is clickable (enabled).
    */
-  async waitUntilEnabled(): Promise<JSHandle> {
-    const handle = this.element.asElement();
-    return await handle.evaluateHandle((e) => {
+  async waitUntilEnabled(): Promise<unknown> {
+    const result = await this.page.waitForFunction((e) => {
```
blrubenstein, line 31: Nit: Could you use a more descriptive variable name than `result`?

e2e/app/aou-elements/button.ts
```diff
@@ -27,12 +27,13 @@ export default class Button extends BaseElement {
...
       const style = window.getComputedStyle(e);
       return style.getPropertyValue('cursor') === 'pointer';
-    }, this.element);
+    }, {}, this.element);
+    const jval = await result.jsonValue();
```
blrubenstein, line 35: Nit: I would leave out this variable, and just `return await result.jsonValue();`

If you do need the variable, I would use a more descriptive name than jval. Maybe `{rename of result}AsJson`?

e2e/app/aou-elements/xpath-defaults.ts
```diff
@@ -1,12 +1,16 @@
 import TextOptions from './text-options';
 
 function textXpathHelper(opts: TextOptions) {
+  let txt;
```
blrubenstein, line 4: Nit: I'd just call it `text`. Or if we can have a more descriptive name describing where the text comes from? Because I am not super following what text this variable is meant to contain.

e2e/tests/user/registration-ui.spec.ts
blrubenstein: I think it should be possible to specify a root directory, so you can use absolute paths from the root rather than relative from here.

e2e/app/aou-elements/button.ts
aweng98: how about `retrievedValue`?

e2e/app/aou-elements/button.ts
aweng98: done

e2e/app/aou-elements/xpath-defaults.ts
aweng98: how about `findText`?

PR #3367, opened 2020-04-08
[RW-4500][RISK=MODERATE] Institution affiliation: distinguish between two institution agreement types

NehaBroad:
Part 2 of distinguish between two institution agreement types:

Verify institution details/validate affiliation for createAccount before directoryService.createUser to avoid creating a user which is going to throw bad request exception later resulting in an entry only in directory service and not in workbench.
**PR checklist**

### DIFF
--- a/api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
+++ b/api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
@@ -369,7 +369,7 @@ public class ProfileControllerTest extends BaseControllerTest {
         new Institution()
             .shortName("Broad")
             .displayName("The Broad Institute")
-            .emailAddresses(Collections.singletonList("bob@example.com"))
+            .emailAddresses(Collections.singletonList(CONTACT_EMAIL))
             .emailDomains(Collections.singletonList("example.com"))
             .duaTypeEnum(DuaType.RESTRICTED);
     institutionService.createInstitution(broad);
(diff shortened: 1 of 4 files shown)


gjuggler:
Neha, can you provide some add'l explanation of this change? On the surface, it seems to be against the grain of our "small controllers" convention (https://github.com/all-of-us/workbench/blob/master/api/doc/code-structure.md#controllers) to bump the verification method out of the service and into the controller. Why was this necessary?

NehaBroad:
@gjuggler have updated the description. The reason was because we were creating a user first in gsuite and then checking for valid institution later while creating an entry in Workbench DB. This ends up with no entry in workbench but an entry in GSUITE.

gjuggler:
> @gjuggler have updated the description. The reason was because we were creating a user first in gsuite and then checking for valid institution later while creating an entry in Workbench DB. This ends up with no entry in workbench but an entry in GSUITE.

Thanks – that makes sense, I appreciate the added detail.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -291,6 +290,147 @@ public void testInvitationKeyVerification_invitationKeyMismatch() {
...
+
+  @Test
+  public void testCreateAccount_Success_RESTRICTEDDUA() {
+    createUser();
+    config.featureFlags.requireInstitutionalVerification = true;
+    final Institution broad =
+        new Institution()
+            .shortName("Broad")
+            .displayName("The Broad Institute")
+            .emailAddresses(Collections.singletonList("bob@example.com"))
```
jmthibault79, line 372: please make this like use `CONTACT_EMAIL` and change the domain to a different domain

PR #3380, opened 2020-04-10
[no ticket][risk=no] jest-junit flag to support circleci split by timings

aweng98:
Update junit XML format to fix `No timing found` error.
Before
Screen Shot 2020-04-10 at 12 54 37 PM
After
Screen Shot 2020-04-10 at 12 54 20 PM

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -27,7 +27,9 @@
   },
   "jest-junit": {
     "outputDirectory": "./logs",
-    "outputName": "test-results.xml"
+    "outputName": "test-results.xml",
+    "classNameTemplate": "{filename}",
+    "suiteNameTemplate": "{filename}"
   },
   "dependencies": {},
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


e2e/package.json
```diff
@@ -27,7 +27,9 @@
   },
   "jest-junit": {
     "outputDirectory": "./logs",
-    "outputName": "test-results.xml"
+    "outputName": "test-results.xml",
+    "classNameTemplate": "{filepath}",
```
jaycarlton, line 31: I kinda wonder why the class name and suite name use the same template.

aweng98:
```
The CLI expects both filenames and classnames to be present in the timing data produced by the testing suite. By default, splitting defaults to filename, but you can specify classnames by using the --timings-type flag.
```
based on https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-by-timing-data

PR #3381, opened 2020-04-10
[no ticket][risk=no] Puppeteer test framework improvement

aweng98:
small, incremental improvement in e2e project. no tests affected.


e2e/app/aou-elements/base-element.ts
```diff
@@ -158,16 +158,31 @@ export default class BaseElement {
   }
 
   async click(options?: ClickOptions): Promise<void> {
-    await this.element.asElement().click(options);
+    return this.element.asElement().click(options);
```
aweng98, line 161: tslint: don't need `await` when `return` is used.

e2e/app/aou-elements/button.ts
```diff
@@ -27,12 +27,23 @@ export default class Button extends BaseElement {
...
    * Wait until button is clickable (enabled).
    */
-  async waitUntilEnabled(): Promise<unknown> {
-    const retrievedValue = await this.page.waitForFunction((e) => {
-      const style = window.getComputedStyle(e);
-      return style.getPropertyValue('cursor') === 'pointer';
-    }, {}, this.element);
-    return await retrievedValue.jsonValue();
+  async waitUntilEnabled(selector?: string): Promise<unknown> {
+    // works with either a xpath selector or a Element
```
aweng98, line 31: Make `waitUntilEnabled` method more flexible.

e2e/app/authenticated-page.ts
```diff
@@ -1,43 +1,16 @@
...
-
-export enum SideNavLinkIcon {
-  HOME = 'home',
-  ADMIN = 'user',
-  CONTACT_US = 'envelope',
-  USER_SUPPORT = 'help',
-  YOUR_WORKSPACES = 'applications',
-  FEATURED_WORKSPACES = 'star',
-}
-
```
aweng98, line 40: consolidate all `enum` into one file `enums.ts`.

e2e/app/workspace-card.ts
```diff
@@ -1,14 +1,7 @@
...
 import BaseElement from './aou-elements/base-element';
-const _ = require('lodash');
-
-
-export const LINK_LABEL = {
-  DUPLICATE: 'Duplicate',
-  DELETE: 'Delete',
-  EDIT: 'Edit',
-  SHARE: 'Share',
-};
```
aweng98, line 11: renamed to `workspaceAction`, it is in new file `enums.ts`

e2e/app/aou-elements/select-component.ts
```diff
@@ -28,19 +52,22 @@ export default class SelectComponent {
     return await innerText.jsonValue();
   }
 
-  // open Select dropdown with retries
-  private async open(retries: number): Promise<void> {
+  /**
+   * Open Select dropdown.
+   * @param {number} retries Retry count. Default is 1.
+   */
+  private async open(retries: number = 1): Promise<void> {
```
calbach, line 59: I typically interpret retries=1 to mean you'll retry it once. Currently it seems to mean "retry 0 times".

e2e/resources/enums.ts
```diff
@@ -0,0 +1,43 @@
+// @ts-nocheck
```
calbach, line 1: Why?

e2e/resources/enums.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export enum workspaceAccessLevel {
```
calbach, line 3: What's the rationale for moving to this file? I'm not sure I like putting all enums into a file named `enums`. It doesn't really carry much logical significance to group them all here - to me it feels a bit similar to putting all the classes in a file named `classes.ts`. I think I prefer how it was before.

e2e/resources/enums.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export enum workspaceAction {
+   DUPLICATE  = 'Duplicate',
+   DELETE = 'Delete',
+   EDIT = 'Edit',
+   SHARE = 'Share',
+}
+
+const configs = require('../resources/workbench-config');
```
calbach, line 16: Interesting, I did not realize you could just inject a `require` statement like this. Could you add a comment about why this is necessary and what the format of the data expected is?

calbach, line 16: Is this really plural? Below the code seems to reference properties on it directly, indicating `config` would be a better name

e2e/resources/fake-user.ts
```diff
@@ -15,3 +15,14 @@ export const fakeUser = {
...
   country: 'U.S.A'
 };
+
+export const makeString = (charLimit?: number) => {
+  let loremStr = faker.lorem.paragraphs();
+  if (charLimit === undefined) {
+    return loremStr;
+  }
+  if (loremStr.length > charLimit) {
+    loremStr = loremStr.slice(0, charLimit - 1);
```
calbach, line 25: slice is exclusive, so I think the -1 is not correct here

e2e/resources/enums.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export enum pageUrl {
```
calbach, line 18: If you're exporting enums, I'd keep them capitalized as they were previously.

e2e/app/aou-elements/select-component.ts
calbach: I'm not sure what to believe anymore,  looks like I implemented this a similar way to you :)

https://github.com/all-of-us/workbench/blob/81a1fc6b9b97d2dfe72c1d978621315bd71b1aea/ui/src/app/utils/retry.tsx#L8

How about `maxAttempts`?

e2e/resources/enums.ts
```diff
@@ -0,0 +1,41 @@
+export const config = require('resources/workbench-config');
+
+export enum WORKSPACE_ACCESS_LEVEL {
```
calbach, line 3: The casing should be like upper-camel: `WorkspaceAccessLevel`

e2e/resources/enums.ts
calbach: Why aren't you able to just use `import` like in the other files?

e2e/resources/enums.ts
calbach: If you feel very strongly I won't block, but I don't think this organization makes sense. These being enums is an implementation detail.

I'd suggest at least splitting this into two files: `workspaces.ts`/`workspace-utils.ts` and `navigation.ts`. Alternatively, rename this file to something that pertains to the logical grouping of these enums: maybe something like `page-identifiers` or `labels`?

PR #3374, opened 2020-04-10
[risk=low]Reinstall karma-sauce-launcher to run angular tests locally

jmthibault79:
Description:

I'm assuming this is fallout from the karma upgrade at #3373 

Following the suggestion at https://github.com/karma-runner/karma-sauce-launcher/issues/117 I can now run the angular tests locally again.

---
**PR checklist**

- [ ] This PR meets the Acceptance Criteria in the JIRA story
- [ ] The JIRA story has been moved to Dev Review
- [ ] This PR includes appropriate unit tests
- [ ] I have run and tested this change locally
- [ ] If this includes an API change, I have updated the appropriate Swagger definitions and notified API consumers
- [ ] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later

PR #3383, opened 2020-04-10
[risk=low]Reinstall wd to re-enable local angular testing

jmthibault79:
Description:

I was mistaken: #3374 did not resolve my local angular testing issues.  So I tried the other suggestion from https://github.com/karma-runner/karma-sauce-launcher/issues/117

---
**PR checklist**

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -163,6 +163,7 @@
     "tslint-eslint-rules": "^5.4.0",
     "tslint-microsoft-contrib": "^6.0.0",
     "typescript": ">=3.1.1 <3.2",
+    "wd": "^1.12.1",
     "webpack-dev-server": "3.1.11"
   }
 }
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -9192,6 +9192,19 @@ wd@^1.11.0:
     request "2.88.0"
     vargs "^0.1.0"
 
+wd@^1.12.1:
+  version "1.12.1"
+  resolved "https://registry.yarnpkg.com/wd/-/wd-1.12.1.tgz#067eb3674db00eeb9e506701f9314657c44d5a89"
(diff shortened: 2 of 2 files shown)


calbach:
@jmthibault79 could you add what command you're running and how it is failing? This stuff really shouldn't be necessary for our angular unit testing setup. This is integration test stuff that we don't touch on the Angular side (though it might be indirectly referenced in our config files).

jmthibault79:
My goal is to match what Circle runs, to avoid surprises: `yarn test --no-watch --no-progress --browsers=ChromeHeadless`

This is only something I started doing recently, so blaming recent changes like karma may be a red herring.  There may be Linux/Mac differences as well.

It's a sporadic problem, which is why I thought it was solved.  Candidates for possible triggers:
* yarn cache clean
* rm node_modules
* yarn install after modifications to package.lock or yarn.lock

jaycarlton:
Can we run chrome headless with those other things inside a Docket
container?

calbach:
It's not a bad idea, we already have such a utility but it runs a debug server which is slightly different than what Circle does. @jmthibault79 if we added a `./project.rb test-headless` which you could call locally, but was also invoked on CircleCI, would that resolve this issue? I'm just not seeing why "wd" would be necessary here.

Trying locally at master, I have no issue running the following:

```
yarn && yarn run code-gen
yarn test --no-watch --no-progress --browsers=ChromeHeadless
```

Also the fact that it only sporadically fails is suspicious, what's the error?

jmthibault79:
Naturally, I have been unable to reproduce the error since you asked to see it. But I do recall what led me to the github issue I referenced here: `Cannot find module '../build/safe-execute'`

The comment on the issue karma-runner/karma-sauce-launcher#117 (comment) suggested the problem is due to a stale `wd` cache which can be fixed by a reinstall.

jmthibault79:
Here's an initial attempt at dockerizing this: https://github.com/all-of-us/workbench/pull/3391

PR #3387, opened 2020-04-13
[RW-4749][risk=no] Add Google Analytics for Registration pages

dolbeew:
Add Google Analytics for events on registration pages.

Events based off the Registration tab of the [workbench analytics doc](https://docs.google.com/spreadsheets/d/1V_Zp0ZtN8f237LNcaGQA-WZzYJu9g8ECq80gJCV8njA/edit#gid=293698803)

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/homepage/registration-dashboard.tsx
+++ b/ui/src/app/pages/homepage/registration-dashboard.tsx
@@ -10,6 +10,7 @@ import {Spinner, SpinnerOverlay} from 'app/components/spinners';
 import {profileApi} from 'app/services/swagger-fetch-clients';
 import colors from 'app/styles/colors';
 import {reactStyles} from 'app/utils';
+import {AnalyticsTracker} from 'app/utils/analytics';
 import {navigate, serverConfigStore, userProfileStore} from 'app/utils/navigation';
 import {environment} from 'environments/environment';
 import {AccessModule, Profile} from 'generated/fetch';
@@ -49,6 +50,7 @@ const styles = reactStyles({
 });
 
 function redirectToGoogleSecurity(): void {
+  AnalyticsTracker.Registration.TwoFactorAuth();
   let url = 'https://myaccount.google.com/u/2/signinoptions/two-step-verification/enroll';
   const {profile} = userProfileStore.getValue();
   // The profile should always be available at this point, but avoid making an
@@ -62,6 +64,7 @@ function redirectToGoogleSecurity(): void {
 }
 
 function redirectToNiH(): void {
+  AnalyticsTracker.Registration.ERACommons();
   const url = environment.shibbolethUrl + '/link-nih-account?redirect-url=' +
           encodeURIComponent(
             window.location.origin.toString() + '/nih-callback?token={token}');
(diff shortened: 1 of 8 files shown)


ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -137,7 +141,10 @@ export const getRegistrationTasks = () => serverConfigStore.getValue() ? ([
     completionTimestamp: (profile: Profile) => {
       return profile.dataUseAgreementCompletionTime || profile.dataUseAgreementBypassTime;
     },
-    onClick: () => navigate(['data-code-of-conduct'])
+    onClick: () => {
+      AnalyticsTracker.Registration.EnterDUCC();
```
calbach, line 145: 🦆

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
```diff
@@ -35,6 +37,7 @@ export class AccountCreationSurvey extends React.Component<AccountCreationSurvey
     const {invitationKey, termsOfServiceVersion, onComplete} = this.props;
 
     try {
+      AnalyticsTracker.Registration.DemographicSurvey();
```
calbach, line 40: nit: For separation of responsibilities, I'd probably put this in the `onSubmit` callback below - since the `createAccount` method has no direct relationship to the demographics survey as a step

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
```diff
@@ -231,6 +232,7 @@ export const DataUserCodeOfConduct = withUserProfile()(
                       data-test-id={'submit-ducc-button'}
                       disabled={errors || submitting}
                       onClick={() => {
+                        AnalyticsTracker.Registration.AcceptDUCC();
```
calbach, line 235: You can view this page from the profile as well. Does it make sense this bump this up to a higher level callback in the registration flow instead? Since the metric is registration specific

ui/src/app/pages/login/account-creation/account-creation-survey.tsx
dolbeew: Makes sense. Done.

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
dolbeew: That does make sense, but doesn't look like there's a higher callback. The submit function just calls `window.history.back()` when complete.

I'm not too familiar with this workflow but seems like we should disable the Accept button if the user has already accepted the agreement. Currently they can view and accept as many times as they want.

ui/src/app/pages/homepage/registration-dashboard.tsx
dolbeew: 😆

ui/src/app/pages/profile/data-user-code-of-conduct.tsx
als364: There are no current plans to change this. Before I changed the workflow you could still re-sign the DUCC from the profile page. I'm not sure whether we still want to do that, that's really more of a product question. I could see either allowing or disallowing re-signing making sense.

PR #3410, opened 2020-04-15
[No Ticket][Risk=no] Make minimum width 1300px for site

blrubenstein:
Description:
This is to address the desire to audit the site for horizontal scroll bars. This disables the footer on all environments (for now), and drops the width of some workspace edit components to fit within 1280px of screen width. This makes the lowest supported web-browser width of the site 1300px. Lou has confirmed that all screens he has designed are 1280px wide. (And then we add 20px for the scroll bar)

---
**PR checklist**

### DIFF
--- a/ui/src/environments/environment.local.ts
+++ b/ui/src/environments/environment.local.ts
@@ -25,5 +25,5 @@ export const environment: Environment = {
   enablePublishedWorkspaces: false,
   enableProfileCapsFeatures: true,
   enableNewConceptTabs: true,
-  enableSignedInFooter: true
+  enableSignedInFooter: false
 };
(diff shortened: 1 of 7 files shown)


ui/src/environments/environment.local.ts
```diff
@@ -25,5 +25,5 @@ export const environment: Environment = {
   enablePublishedWorkspaces: false,
   enableProfileCapsFeatures: true,
   enableNewConceptTabs: true,
-  enableSignedInFooter: true
+  enableSignedInFooter: false
```
NehaBroad, line 28: we are hiding it from local and test?

blrubenstein: I am hiding it from local and test because we are doing UX verification on those environments leading up to launch, and while we are doing that testing I wanted to make sure that the footer wasn't there, since we aren't going to have it for launch.

PR #3415, opened 2020-04-15
[RW-4693][risk=no] Update ethics training language

dolbeew:
Updated language from 'Research Ethics Training' to '_All of Us_ Responsible Conduct of Research Training' on homepage and profile page:
Screen Shot 2020-04-15 at 4 15 55 PM
Screen Shot 2020-04-15 at 4 16 56 PM

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/homepage/registration-dashboard.tsx
+++ b/ui/src/app/pages/homepage/registration-dashboard.tsx
@@ -81,7 +81,7 @@ interface RegistrationTask {
   key: string;
   completionPropsKey: string;
   loadingPropsKey?: string;
-  title: string;
+  title: React.ReactNode;
   description: React.ReactNode;
   buttonText: string;
   completedText: string;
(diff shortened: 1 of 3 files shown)


ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -81,7 +81,7 @@ interface RegistrationTask {
   key: string;
   completionPropsKey: string;
   loadingPropsKey?: string;
-  title: string;
+  title: React.ReactNode;
```
dolbeew, line 84: Used `React.ReactNode` for the titles so they can be either a string or an element.

ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -122,7 +122,7 @@ export const getRegistrationTasks = () => serverConfigStore.getValue() ? ([
   }, {
     key: 'complianceTraining',
     completionPropsKey: 'trainingCompleted',
-    title: 'Complete Ethics training',
+    title: <span><i>All of Us</i> Responsible Conduct of Research Training</span>,
```
NehaBroad, line 125: Nit: i just came across while cleaning up some code <AouTitle/> maybe we can use it here too

dolbeew: Sorry, not sure I understand. What can we maybe use here?

blrubenstein: There's an <AoUTitle/> element (I assume) but that specifically puts the text "_All of Us_ Research Program", which won't work here.

NehaBroad: My bad by editing the comment i just ended up deleting the Tag itself. But scout is right since it has research we cannot use but **not part of this ticket** it would be great to have a <i>All Of Us</i> so that in case of change we do not have to go to All the pages and do it

PR #3433, opened 2020-04-17
increase no-output-timeout on circleci

aweng98:
Recently added workspace-create tests have longer running time. Average test run time is around 10 min.
Screen Shot 2020-04-17 at 4 59 35 PM

Today, encountered no-output timeout error for the first time on CircleCI [here](https://circleci.com/gh/all-of-us/workbench/77797?utm_campaign=chatroom-integration&utm_medium=referral&utm_source=slack)
```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

Fix: increase circleci no_output_timeout option to 15m.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -424,7 +424,6 @@ workflows:
       - ui-build-test
       - api-bigquery-test
       - api-integration-test
-      - puppeteer-e2e-test
       # Run deployment to "test" on master merges.
       - api-deploy-to-test:
           <<: *filter_master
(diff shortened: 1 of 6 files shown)


calbach:
200 seconds is insane - why does it take this long? Are the puppeteer tests depleting the billing buffer?

aweng98:
Reason is that now user has to enter a long string in textarea or facing this error. 
Screen Shot 2020-04-17 at 5 37 11 PM

I can tweak random strings to be shorter and still meet that requirement and I can also make tests run a little faster by reducing `slowMo` Puppeteer option.
What do you think it is a better approach?

jaycarlton:
@aweng98 Instead of typing character by character, why not paste something into the edit box or (better yet) use a puppeteer command especially for typing words or paragraphs?

PR #3439, opened 2020-04-18
[RW-4633][risk=no] Ignore more missing props & try treating warnings as errors

jaycarlton:
The way research purposes and workspaces work together makes things awkward in several places. All I'm doing here is stifling a few warnings.

I'm deprecating the `mergeResearchPurposeIntoWorkspace()` mapper method as I'm about to get rid of it altogether.

Also, I looked into treating warnings as errors, and it choked on our generated files. For those interested, see the [output](https://gist.github.com/jaycarlton/d3d2bc9f134969576f1e8c362c25fe1d).
 
To enable this in the build.gradle file I did 
```
gradle.projectsEvaluated {
  tasks.withType(JavaCompile) {
    options.compilerArgs << "-Xlint:all" << "-Werror"
  }
}
```
@calbach I didn't see an obvious way to split out our generated code.  It's in its own directory, but we share packages (and a compile target) with it.
---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
+++ b/api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
@@ -50,12 +50,13 @@ public interface WorkspaceMapper {
   @Mapping(target = "cdrVersionId", source = "dbWorkspace.cdrVersion")
   Workspace toApiWorkspace(DbWorkspace dbWorkspace, FirecloudWorkspace fcWorkspace);
 
-  @Mapping(target = "researchPurpose", source = "dbWorkspace")
+  @Mapping(target = "cdrVersionId", source = "cdrVersion")
+  @Mapping(target = "creator", source = "creator.username")
   @Mapping(target = "etag", source = "version", qualifiedByName = "cdrVersionToEtag")
+  @Mapping(target = "googleBucketName", ignore = true) // available via toApiWorkspace(DbWorkspace dbWorkspace, FirecloudWorkspace fcWorkspace)
   @Mapping(target = "id", source = "firecloudName")
   @Mapping(target = "namespace", source = "workspaceNamespace")
-  @Mapping(target = "creator", source = "creator.username")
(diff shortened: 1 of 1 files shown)


jmthibault79:
This is RW-4633.  I'll close my existing PR https://github.com/all-of-us/workbench/pull/3353 since you're running with this 👍

jaycarlton:
Thanks, sorry I missed that.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -92,6 +98,37 @@ WorkspaceResponse toApiWorkspaceResponse(
...
+  @Mapping(target = "disseminateResearchSet", ignore = true)
+  @Mapping(target = "firecloudName", ignore = true)
+  @Mapping(target = "firecloudUuid", ignore = true)
+  @Mapping(target = "lastAccessedTime", ignore = true)
+  @Mapping(target = "lastModifiedTime", ignore = true)
+  @Mapping(target = "name", ignore = true)
+  @Mapping(target = "published", ignore = true)
+  @Mapping(target = "researchOutcomeSet", ignore = true)
+  @Mapping(target = "reviewRequested", ignore = true)
+  @Mapping(target = "timeRequested", ignore = true)
```
NehaBroad, line 127: question: By ignoring it here we are saying look at the corresponding method in IMPL to see how to convert it from DB to other modal (and vice cersa)?

NehaBroad:
I have verified the changes on local 

- There are no warnings when starting server 

- i can save , edit and duplicate workspace

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
jaycarlton: Ignore should be very rare. It's almost always a sign that we don't have a good handle on this. Basically, if a function returns an object, it should have all required fields populated. Swagger-codegen doesn't give us a good way to set some defaults.

I have a better solution in the works.

PR #3448, opened 2020-04-21
[risk=no] Include sudo in docker-sync install docs

calbach:
Check this with Greg to confirm this was the intent. @aweng98 hit this when running through the instructions.


gjuggler:
Seems reasonable for the root-level install.

PR #3457, opened 2020-04-22
[risk=no][RW-4678]Followup: don't spam logs and make proper float comparisons

jmthibault79:
Description:

The motivation for this change was seeing a large number of logs like this:

```User [redacted] has 4.041393 in total free tier spending in BigQuery, which is less than the 4.041393 previous spending we have recorded in the DB (FreeTierBillingService.java:182)```

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
@@ -47,9 +47,9 @@ public class FreeTierBillingService {
 
   private static final Logger logger = Logger.getLogger(FreeTierBillingService.class.getName());
 
-  // somewhat arbitrary
-  private static final double COST_COMPARISON_TOLERANCE = 0.0001;
-  private static final double COST_FRACTION_TOLERANCE = 0.0001;
+  // somewhat arbitrary - 1 per million
+  private static final double COST_COMPARISON_TOLERANCE = 0.000001;
+  private static final double COST_FRACTION_TOLERANCE = 0.000001;
 
   @Autowired
   public FreeTierBillingService(
--- a/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
@@ -176,9 +176,10 @@ public class FreeTierBillingService {
 
     // this shouldn't happen, but it did (RW-4678)
     // alert if it happens again
-    if (currentCost < previousCost &&
-            // was logging many false positives
-            !withinCostTolerance(currentCost, previousCost) ) {
+    if (currentCost < previousCost
+        &&
+        // was logging many false positives
+        !withinCostTolerance(currentCost, previousCost)) {
       String msg =
           String.format(
               "User %s (%s) has %f in total free tier spending in BigQuery, "
--- a/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
@@ -3,6 +3,7 @@ package org.pmiops.workbench.billing;
 import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.QueryJobConfiguration;
 import com.google.common.collect.Sets;
+import com.google.common.math.DoubleMath;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
@@ -45,6 +46,10 @@ public class FreeTierBillingService {
   private final Provider<WorkbenchConfig> workbenchConfigProvider;
 
   private static final Logger logger = Logger.getLogger(FreeTierBillingService.class.getName());
+
+  // somewhat arbitrary
+  private static final double COST_COMPARISON_TOLERANCE = 0.0001;
+  private static final double COST_FRACTION_TOLERANCE = 0.0001;
 
   @Autowired
   public FreeTierBillingService(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java [hunk 1]
```diff
@@ -151,6 +151,12 @@ private void sendAlertsForCostThresholds(
         });
   }
 
+  // are these two cost values approximately equal?
+  // used to determine whether we should log anomalies
+  private boolean withinCostTolerance(double currentCost, double previousCost) {
```
jaycarlton, line 156: Please check if there's a library function for this. As excited as I am to see actual math for the first time in 2020, we shouldn't be doing it this way. If we keep this, we probably need to make a constant out of the tolerance.

Additionally, why are we using doubles instead of a Currency class?

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -170,7 +176,10 @@ private void maybeAlertOnCostThresholds(
 
     // this shouldn't happen, but it did (RW-4678)
     // alert if it happens again
-    if (currentCost < previousCost) {
+    if (currentCost < previousCost
+        &&
+        // was logging many false positives
+        !withinCostTolerance(currentCost, previousCost)) {
```
jaycarlton, line 182: Let's see if we can add an audit entry for this, since it's more user-oriented than system- or application-oriented.

jmthibault79: The purpose of this log is to alert us to bugs in our system, rather than reflecting an actual user condition.  Does that change your judgment of audit vs log?

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: good call re: library function.  I'll update.

re: Currency.  When implemented, that seemed like overkill for our application since we're not doing anything complex and we don't require financial-level precision.  I'm open to the change if we think it's justified.

jmthibault79: Ultimately we retrieve these values from BigQuery as doubles, so we'll never have the integer-like precision we'd really want in a Currency.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -123,8 +128,16 @@ public void checkFreeTierBillingUsage() {
     sendAlertsForCostThresholds(usersToThresholdCheck, previousUserCosts, userCosts);
   }
 
+  private int compareCosts(final double a, final double b) {
```
jaycarlton, line 131: I wrote a little utility class for this so you don't have to remember the convention. org.pmiops.workbench.utils.Comparables

If you wanted to use it here, you'd need to have a little Comparable class wrapping the fuzzyCompare method. Assuming the convention is the same. I think you'd have make a comparable FuzzyCost class to wrap Double or something though, so not a huge win.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -46,6 +47,10 @@
 
   private static final Logger logger = Logger.getLogger(FreeTierBillingService.class.getName());
 
+  // somewhat arbitrary
+  private static final double COST_COMPARISON_TOLERANCE = 0.0001;
+  private static final double COST_FRACTION_TOLERANCE = 0.0001;
```
jaycarlton, line 52: There's probably a way to figure this out based on use cases. I know from geometry angle and linear tolerance are different too. This is probably OK.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -241,7 +241,7 @@ public void checkFreeTierBillingUsage_altDollarThresholds() throws MessagingExce
         .alertUserFreeTierDollarThreshold(
             eq(user), eq(threshold), eq(costOverThreshold), eq(remaining));
 
-    // check that we do not alert twice for the 75% threshold
+    // check that we do not alert twice for the 65% threshold
```
jmthibault79, line 244: copypaste.  this test uses different thresholds.

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -555,14 +555,14 @@ public void getUserCachedFreeTierUsage() {
...
     final Map<String, Double> costs =
-        ImmutableMap.of(SINGLE_WORKSPACE_TEST_PROJECT, 1000.0, "another project", 100.0);
+        ImmutableMap.of(SINGLE_WORKSPACE_TEST_PROJECT, 1000.0, "another project", 200.0);
     doReturn(mockBQTableResult(costs)).when(bigQueryService).executeQuery(any());
 
     // we have not yet cached the new workspace costs
     assertWithinBillingTolerance(freeTierBillingService.getCachedFreeTierUsage(user1), 100.01);
 
     freeTierBillingService.checkFreeTierBillingUsage();
-    final double expectedTotalCachedFreeTierUsage = 1000.0 + 100.01;
```
jmthibault79, line 565: this was wrong!  but close enough due to loose tolerance.  No longer.

PR #3463, opened 2020-04-23
[RW-4817][Risk=no] Hardcode width on dropdown

blrubenstein:
Description:
So, this one is interesting. It looks like the PrimeReact dropdown text has a bug where it will expand out to fill the space. I am attaching a screenshot of what that does to the actual bounding boxes. 
Problem:
Screen Shot 2020-04-23 at 3 00 20 PM
Solution:
Screen Shot 2020-04-23 at 3 06 53 PM

 
We tried fixing it with `text-overflow` and with `overflow` but neither worked. The best thing we could find was pinning the width, which we do a lot on these pages anyway, so that should be okay.

Note: This solution doesn't fix the underlying issues, but does cap out the width such that the aside moves back in, so that the page doesn't get wider.

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/login/account-creation/account-creation-institution.tsx
+++ b/ui/src/app/pages/login/account-creation/account-creation-institution.tsx
@@ -29,6 +29,9 @@ import {Dropdown} from 'primereact/dropdown';
 const styles = reactStyles({
   ...commonStyles,
   wideInputSize: {
+    // We have a fixed width here because the text for dropdown options sometimes
+    // expands beyond the width of the dropdown, which leads to display bugs.
+    // For example, https://precisionmedicineinitiative.atlassian.net/browse/RW-4817
     width: 600
   },
   institutionalDuaTextBox: {
--- a/ui/src/app/pages/login/account-creation/account-creation-institution.tsx
+++ b/ui/src/app/pages/login/account-creation/account-creation-institution.tsx
@@ -29,8 +29,7 @@ import {Dropdown} from 'primereact/dropdown';
 const styles = reactStyles({
   ...commonStyles,
   wideInputSize: {
-    width: '50%',
-    minWidth: '600px'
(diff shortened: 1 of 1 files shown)


ui/src/app/pages/login/account-creation/account-creation-institution.tsx
```diff
@@ -29,8 +29,7 @@ import {Dropdown} from 'primereact/dropdown';
 const styles = reactStyles({
   ...commonStyles,
   wideInputSize: {
-    width: '50%',
-    minWidth: '600px'
+    width: 600
```
gjuggler, line 32: Please add a comment with a short description of the reasoning here and a cross-link to the Jira issue. It's a nuanced enough fix that it warrants some breadcrumbs.

blrubenstein: Got it, will do.

petesantos: The flex issue here is counter-intuitive. If you set the min-width to 600px on the first child of the top level flex container (the first column) it should resolve the issue. 

The reason the min-width needs to be set is because the min-width of a flex item defaults to 'auto'. If the text is not wrapped, as is the case with the selected item in the drop down, the browser will compute the min width for the flex item as 'auto' essentially sizing the item to the content. Setting a fixed min-width will correct for this.

You can read more here - although it is verbose
https://www.w3.org/TR/css-flexbox-1/#min-size-auto

gjuggler: Nice catch – and with a w3c reference to boot!

My assertion about the fix being "obscure" may still stand, but I was wrong about it being more complicated than the workaround in this PR – it actually feels much simpler that way.

PR #3466, opened 2020-04-23
[risk=no][RW-4815] change icon color

als364:
Screen Shot 2020-04-23 at 4 39 10 PM

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/icons.tsx
+++ b/ui/src/app/components/icons.tsx
@@ -17,8 +17,8 @@ export const styles = {
     height: '20px'
   },
 
-  warningIcon: {
-    color: colors.warning,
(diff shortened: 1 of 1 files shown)


ui/src/app/components/icons.tsx
```diff
@@ -17,8 +17,8 @@ export const styles = {
     height: '20px'
   },
 
-  warningIcon: {
-    color: colors.warning,
+  dangerIcon: {
+    color: colors.danger,
```
als364, line 21: As it turns out, `ValidationIcon` is only used in registration, so I'd rather change it one place than pass in styles in two places.

PR #3469, opened 2020-04-24
[RW-4830][RISK=NO] Export parameter focusOnUnderrepresentedPopulations to RDR

NehaBroad:
**Problem** : While exporting workspace data to RDR, workbench was not sending the parameter focusOnUnderrepresentedPopulations (which represents if researcher has selected the question _Will your study focus on any historically underrepresented populations_ while creating workspace). Because of which RDR was assuming this value to be false and ignoring the workspace Demographic information being sent.

**Solution**: Populate and send the value focusOnUnderrepresentedPopulations while exporting Workspace data to RDR:
 IF workspace has any Demographic information (race/age etc) send YES else NO

Confirmed with Wang Yu its working fine and RDR is able to store this value

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
@@ -296,6 +296,8 @@ public class RdrExportServiceImpl implements RdrExportService {
     rdrWorkspace.setFindingsFromStudy(dbWorkspace.getAnticipatedFindings());
     rdrWorkspace.setReviewRequested(dbWorkspace.getReviewRequested());
 
+    rdrWorkspace.setFocusOnUnderrepresentedPopulations(
+        dbWorkspace.getSpecificPopulationsEnum().size() > 0);
     rdrWorkspace.setWorkspaceDemographic(
         toRdrWorkspaceDemographics(dbWorkspace.getSpecificPopulationsEnum()));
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -296,6 +296,8 @@ private RdrWorkspace toRdrWorkspace(DbWorkspace dbWorkspace) {
     rdrWorkspace.setFindingsFromStudy(dbWorkspace.getAnticipatedFindings());
     rdrWorkspace.setReviewRequested(dbWorkspace.getReviewRequested());
 
+    rdrWorkspace.setFocusOnUnderrepresentedPopulations(
+        dbWorkspace.getSpecificPopulationsEnum().size() > 0);
```
gjuggler, line 300: Do we have a guarantee that the getSpecificPopulationsEnum value will be non-null? Looking at the existing code, we seem to be assuming it's non-null. Just wondering where that expectation comes from.

NehaBroad: This will non-null since populationDetails  is initialized to a HashSet  [DbWorkspace](https://github.com/all-of-us/workbench/blob/cea86f34bea7eb37a0948ad5a7c1aa948c6035f6/api/src/main/java/org/pmiops/workbench/db/model/DbWorkspace.java) L 102

If this is ever change to null, RdrExportServiceImplTest.exportWorkspace_successful will fail with null pointer exception

PR #3470, opened 2020-04-24
[RW-4674][RISK=LOW][P2] Registration form button and checkboxes resets to original state

NehaBroad:
ISSUES FIXED FOR Create account flow: 

- The TOS page was not persisting the selection i.e if user selects the terms and condition page go to the next page and select PREVIOUS Button. They had to re select the checkboxes again

- If the institution information has been filled by researcher and they move to another page (PREVIOUS OR NEXT) if they come back to the institution page, the email validation image was not being  displayed.

---
**PR checklist**


NehaBroad:
Good point, so is Previous was being set to false on click on Next button in  all the steps except TOS. I have pushed the fix for the same.
 In case of two previous and one next i have tested it is set to true. Also isPrevious is actually just used in case of TOS since in all other cases profile object is populated so the checks are on the basis of profile value

PR #3493, opened 2020-04-28
[RW-4839][RISK=NO] Change logic to 'excludeFromPublicDirectory' flag

NehaBroad:
Along with workspace details, workbench send RDR attribute excludeFromPublicDirectory , that excludes the workspace to be show up in Researcher directory. 
By default we send the value false i.e show all workspace in researcher directory

As part of this PR: Workbench will send  excludeFromPublicDirectory flag as true if the creator of the workspace has institution All of Us Program operational Use

I have tested the exportToWorkspace call with:

-  Workspace with creator with no verified institution 

- Workspace with creator with verified institution other than All of Us Program operational Use
- Workspace with creator with verified institution as All of Us Program operational Use

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
@@ -24,6 +24,7 @@ import org.pmiops.workbench.db.model.DbRdrExport;
 import org.pmiops.workbench.db.model.DbUser;
 import org.pmiops.workbench.db.model.DbWorkspace;
 import org.pmiops.workbench.db.model.RdrEntityEnums;
+import org.pmiops.workbench.institution.InstitutionService;
 import org.pmiops.workbench.model.InstitutionalRole;
 import org.pmiops.workbench.model.RdrEntity;
 import org.pmiops.workbench.model.SpecificPopulationEnum;
@@ -52,6 +53,8 @@ public class RdrExportServiceImpl implements RdrExportService {
   private RdrExportDao rdrExportDao;
   private WorkspaceDao workspaceDao;
   private UserDao userDao;
+
+  private InstitutionService institutionService;
   private WorkspaceService workspaceService;
   private final VerifiedInstitutionalAffiliationDao verifiedInstitutionalAffiliationDao;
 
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -258,10 +258,11 @@ private RdrResearcher toRdrResearcher(DbUser dbUser) {
                       ? verifiedInstitutionalAffiliation.getInstitutionalRoleOtherText()
                       : roleEnum.toString();
 
-              new ResearcherVerifiedInstitutionalAffiliation()
-                  .institutionShortName(
-                      verifiedInstitutionalAffiliation.getInstitution().getShortName())
-                  .institutionalRole(role);
+              researcher.setVerifiedInstitutionalAffiliation(
```
NehaBroad, line 261: Not related to this Story but i noticed this information was not being send as part of exportResearcher hence updated it as well

gjuggler: good catch, that seems important :)

api/src/main/java/org/pmiops/workbench/rdr/RdrExportServiceImpl.java
```diff
@@ -435,4 +436,25 @@ public void updateDbRdrExport(RdrEntity entity, List<Long> idList) {
...
+    rdrWorkspace.setExcludeFromPublicDirectory(false);
+    verifiedInstitutionalAffiliationDao
+        .findFirstByUser(creatorUser)
+        .ifPresent(
+            verifiedInstitutionalAffiliation -> {
+              rdrWorkspace.setExcludeFromPublicDirectory(
+                  verifiedInstitutionalAffiliation
+                      .getInstitution()
+                      .getShortName()
+                      .equals("AouOps"));
```
gjuggler, line 457: Please make this a file-level constant. Or, better yet, most of this method might be better placed within the institution service, e.g. `institutionService.isOperationalUser(dbUser user)`. That would centralize the logic within the institution package, and I don't think it would require adding any new deps to the institution service.

Slightly tangential thought: I'm concerned that we're at risk of putting load-bearing system weight on the `shortName` identifier, which can be changed by operational end-users via the Institution UI at any time. This is a pretty serious red flag, and we should think about ways to make sure far-flung business logic can't be impacted by someone deciding in an admin UI that "AouOps" should be capitalized differently, or have a space added, etc. etc.

I filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4844 to track – I'd suggest following up with Joel & Karthik to align on a solution for this.

PR #3494, opened 2020-04-28
[no ticket][risk=no] Fix cache key regression in circleci config

aweng98:
Fixed bugs I introduced in my last config change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ anchors:
       # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
       # so we bumped this down further to 2G.
       JAVA_TOOL_OPTIONS: -Xmx2g
+      # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
       GRADLE_OPTS: "-Dorg.gradle.daemon=false"
       TERM: dumb
 
@@ -120,7 +121,7 @@ commands:
           name: "restore ui cache"
           keys:
             - v1-ui-cache-{{ .Branch }}-{{ checksum "~/workbench/ui/yarn.lock" }}
-            - v1-ui-cache-{{ .Branch }}
+            - v1-ui-cache-master
             - v1-ui-cache-
       - run:
           name: "workbench/ui: yarn install dependencies"
@@ -131,7 +132,7 @@ commands:
           paths:
             - ~/.cache/yarn
             - ui/node_modules
-          key: v1-ui-cache-{{ .Branch }}
+          key: v1-ui-cache-{{ .Branch }}-{{ checksum "~/workbench/ui/yarn.lock" }}
 
   install_e2e_dependencies:
     description: "workbench/e2e: yarn install, save and restore cache"
@@ -140,7 +141,7 @@ commands:
           name: "restore e2e cache"
           keys:
             - v1-e2e-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
-            - v1-e2e-cache-{{ .Branch }}
+            - v1-e2e-cache-master
             - v1-e2e-cache-
       - run:
           name: "workbench/e2e: yarn install dependencies"
@@ -151,7 +152,7 @@ commands:
           paths:
             - ~/.cache/yarn
             - e2e/node_modules
-          key: v1-e2e-cache-{{ .Branch }}
+          key: v1-e2e-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
 
   manage_api_cache:
     description: "workbench/api: save or restore gradle cache"
@@ -166,61 +167,20 @@ commands:
       - when:
           condition: << parameters.save >>
           steps:
-            # splitting into multiple distinct caches. Refers to https://circleci.com/docs/2.0/caching/#cache-size
-            - save_cache:
-                name: "save api gradle/wrapper cache"
-                paths:
-                  - ~/.gradle/wrapper
-                key: v1-gradle-wrapper-{{ .Branch }}-{{ checksum "~/workbench/api/gradle/wrapper/gradle-wrapper.properties" }}
             - save_cache:
                 name: "save api gradle cache"
                 paths:
-                  - ~/.gradle/caches
+                  - ~/.gradle
                 key: v1-gradle-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
       - when:
           condition: << parameters.restore >>
           steps:
             - restore_cache:
-                name: "restore api gradle/wrapper cache"
-                keys:
-                  - v1-gradle-wrapper-{{ .Branch }}-{{ checksum "~/workbench/api/gradle/wrapper/gradle-wrapper.properties" }}
-                  - v1-gradle-wrapper-{{ .Branch }}
-                  - v1-gradle-wrapper-
-            - restore_cache:
                 name: "restore api gradle cache"
                 keys:
                   - v1-gradle-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
-                  - v1-gradle-cache-{{ .Branch }}
+                  - v1-gradle-cache-master
                   - v1-gradle-cache-
-
-  manage_maven_cache:
-    description: "workbench/api: save or restore maven cache"
-    parameters:
-      save:
-        type: boolean
-        default: false
-      restore:
-        type: boolean
-        default: false
-    steps:
-      - when:
-          condition: << parameters.save >>
-          steps:
-            - save_cache:
-                name: "save build cache"
-                paths:
-                  - ~/.m2
-                  - ~/workbench/api/build/exploded-api/WEB-INF/lib/
-                key: v1-build-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
-      - when:
-          condition: << parameters.restore >>
-          steps:
-            - restore_cache:
-                name: "restore build cache"
-                keys:
-                  - v1-build-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
-                  - v1-build-cache-{{ .Branch }}
-                  - v1-build-cache-
 
   run_e2e_test:
     description: "Run puppeteer e2e integration tests"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -19,16 +19,21 @@ anchors:
...
+    resource_class: medium+
     environment:
       # As best I can tell (dmohs, 7 Feb '17), this is the only way to set a memory limit that Java
       # processes executed within CircleCI's docker containers will respect. Very helpful resource:
       # https://circleci.com/blog/how-to-handle-java-oom-errors/
       #
       # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
       # so we bumped this down further to 2G.
-      JAVA_TOOL_OPTIONS: -Xmx2g
+      # JAVA_TOOL_OPTIONS: -Xmx2g
```
aweng98, line 30: I don't think reducing memory size for Gradle is the right approach. Gradle needs decent size of memory to be "happy" and "fast". Reducing it is not what we should do. Instead, use a larger container `medium+` (default was `medium`) and assign gradle a larger memory size. So far, all my testing, I have not had an issue with OOM anymore.

.circleci/config.yml
```diff
@@ -167,47 +166,52 @@ commands:
...
   run_e2e_test:
     description: "Run puppeteer e2e integration tests"
     steps:
+      - run:
+          name: Update Puppeteer test environment variables
+          command: |
+            echo 'export USER_NAME=$PUPPETEER_TEST_USER' >> $BASH_ENV
+            echo 'export PASSWORD=$PUPPETEER_TEST_USER_PASSWORD' >> $BASH_ENV
+            echo 'export INVITATION_KEY=$PUPPETEER_TEST_REGISTRATION_KEY' >> $BASH_ENV
+            source $BASH_ENV
```
aweng98, line 211: Bug:
Still need this way to get env. variables to work. :(

.circleci/config.yml
```diff
@@ -131,7 +134,6 @@ commands:
             # If that fails, then check for the latest from any branch
             - << parameters.key_prefix >>-{{ .Branch }}
             - << parameters.key_prefix >>-master
-            - << parameters.key_prefix >>
```
aweng98, line 134: If `checksum` fails (which never happens), then new cache is created and will be stored. Don't use this fall back, clutter code.

.circleci/config.yml
```diff
@@ -167,47 +166,52 @@ commands:
...
       - when:
           condition: << parameters.save >>
           steps:
+            # splitting into multiple distinct caches. Refers to https://circleci.com/docs/2.0/caching/#cache-size
             - save_cache:
                 paths:
-                  - ~/.gradle/caches
                   - ~/.gradle/wrapper
-                  - ~/.m2
-                  - ~/workbench/api/build/exploded-api/WEB-INF/lib/
```
aweng98, line 190: I am not sure either dir existed to have anything to cache. Am I wrong?

.circleci/config.yml
```diff
@@ -19,16 +19,21 @@ anchors:
...
       # so we bumped this down further to 2G.
-      JAVA_TOOL_OPTIONS: -Xmx2g
+      # JAVA_TOOL_OPTIONS: -Xmx2g
+
+      # Configure the JVM and Gradle: give it a bigger memory to avoid OOM errors
+      # https://circleci.com/docs/2.0/java-oom/#using-java-environment-variables-to-set-memory-limits
+      _JAVA_OPTIONS: "-Xmx3g"
       # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
-      GRADLE_OPTS: "-Dorg.gradle.daemon=false"
+      GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
```
aweng98, line 36: If I don't specify `worker.max`, it defaults to number of cpu. I don't want that.

.circleci/config.yml
```diff
@@ -167,47 +166,52 @@ commands:
...
         type: boolean
         default: false
-      key_prefix:
-        type: string
-        default: v1-api-cache-{{ checksum "~/workbench/api/build.gradle" }}
     steps:
       - when:
           condition: << parameters.save >>
           steps:
+            # splitting into multiple distinct caches. Refers to https://circleci.com/docs/2.0/caching/#cache-size
```
aweng98, line 181: I splitted one cache to two separate caches. See url for reason.

.circleci/config.yml
calbach: The counter argument is that we should not need a large (and more expensive) machine to run our unit tests. Have you experimented with changing the resource class independent of the other changes you've made here? If this one difference gets us from 15m -> 6m, great. If it has no effect on latency, then leaving it at the default seems preferred.

Please update the comment here and remove commented out code.

.circleci/config.yml
calbach: Why not? Also please add a comment for why 2 was chosen into the code

.circleci/config.yml
```diff
@@ -116,11 +117,9 @@ commands:
   install_ui_dependencies:
     description: "workbench/ui: yarn install, save and restore cache"
     parameters:
-      version:
-        type: string
       key_prefix:
         type: string
-        default: << parameters.version >>-yarn-ui-cache-{{ checksum "~/workbench/ui/yarn.lock" }}
+        default: v1-yarn-ui-cache-{{ checksum "~/workbench/ui/yarn.lock" }}
```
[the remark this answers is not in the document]
calbach: I missed this on the previous review - checksum should not be in the key prefix. It should be the suffix. The restore backup keys don't make sense with this model. Here is how it was supposed to work:

```
restore_cache:
  keys:
    - v1-api-cache-{checksum} # Restore from exact match
    - v1-api-cache- # Restore from any recent cached version, as a backup
```

If you want to incorporate the branch into the cache key, the branch name should come before the checksum, like:

```
v1-api-cache-{branch}-{checksum}
v1-api-cache-{branch}
v1-api-cache-master
```

.circleci/config.yml
calbach: What is the size of the cache? Are we actually nearing 500MB?

.circleci/config.yml
calbach: The deployment job puts stuff in exploded-api
~/.m2 may contain maven caches, but unsure

You should be able SSH into a circle job and check to confirm

.circleci/config.yml
```diff
@@ -222,26 +226,51 @@ commands:
...
       - manage_api_cache:
           restore: true
       - run:
-          name: Validate Swagger definitions
+          name: Run Java unit tests
           working_directory: ~/workbench/api
-          command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+          # Refers to https://circleci.com/docs/2.0/language-java/#sample-configuration
+          command: |
+            cd ~/workbench/api/src/test/java/org/pmiops/workbench/api
```
calbach, line 241: This is not correct. `api` is one subdirectory of many that contain unit tests

I would change this to `pushd src/test`, then `popd` later

.circleci/config.yml
```diff
@@ -222,26 +226,51 @@ commands:
...
       - ensure_branch_has_changes:
           dir_name: "api"
       - manage_api_cache:
           restore: true
       - run:
-          name: Validate Swagger definitions
+          name: Run Java unit tests
           working_directory: ~/workbench/api
-          command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+          # Refers to https://circleci.com/docs/2.0/language-java/#sample-configuration
```
calbach, line 239: Can you add more to this comment? The shell script is pretty hairy and it's not clear what it's actually doing. Presumably it has something to do with parallelizing the tests

.circleci/config.yml
```diff
@@ -222,26 +226,51 @@ commands:
...
+          when: always
+      - store_test_results:
+          path: ~/workbench/api/test-results
+      - store_artifacts:
+          path: ~/workbench/api/test-results
+          destination: test-results
+      - manage_api_cache:
+          save: true
+
+  api-lint-swagger-check:
```
calbach, line 263: I'd prefer that we move these steps back into the API test, then add a `when: always` to " Run Java unit tests". This fixes the ordering concerns that we had previously. Unless there are major latency tradeoffs here.

aweng98:
@calbach Hi, just want to give a quick update on where this PR is at. I reverted changes to the `api-build-test` job (no more `parallelism`, no more `medium+` image). To test all possible combinations to avoid using larger docker image and OOM proves to be a very time-consuming task. I have to stop. So, this PR is ready for final review. Thanks.

.circleci/config.yml
jaycarlton: I looked into this on one of your other PRs, and it seemed like the nicer syntax won't work in your situation. I can't remember details though.

.circleci/config.yml
calbach: The cache is best effort. In theory all jobs should succeed regardless of whether they had received an ideal or even relevant cache. gradle wapper properties very rarely changes, so I would probably favor the simplicity of the previous singular cache approach. If we do get stuck with a cache issue, incrementing from "v1" is a nice manual out to have available that you've added.

.circleci/config.yml
```diff
@@ -115,99 +114,127 @@ commands:
...
+                  - v1-gradle-wrapper-{{ .Branch }}
+                  - v1-gradle-wrapper-
+            - restore_cache:
+                name: "restore api gradle cache"
+                keys:
+                  - v1-gradle-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
+                  - v1-gradle-cache-{{ .Branch }}
+                  - v1-gradle-cache-
+
+  manage_maven_cache:
```
calbach, line 196: I would also fold this back into the API cache unless size is a concern.

.circleci/config.yml
```diff
@@ -27,7 +27,6 @@ anchors:
       # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
       # so we bumped this down further to 2G.
       JAVA_TOOL_OPTIONS: -Xmx2g
-      # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
```
calbach, line 30: This comment still seems relevant

.circleci/config.yml
aweng98: Thank you, your help and review is greatly appreciated.

PR #3498, opened 2020-04-29
[RW-4793][risk=no] Fix disable user CLI

calbach:
- Update bit-rotted API URL reference, this was the primary regression
- Make the arguments all required (usability issue)
- Rename the command and update docs to be more accurate

Future work, not included here:
- This CLI still suffers from not telling you which required flag you forgot (a general issue with all our commands). I don't attempt to solve that here.
- The command leaves you signed in as your admin email account. This is not ideal, but probably needs broader changes

Example usage:
```
./project.rb update-user-disabled-status --project all-of-us-workbench-test --account calbach@fake-research-aou.org --user calbach2@fake-research-aou.org --disabled true
```

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -1275,27 +1275,28 @@ def update_user_registered_status(cmd_name, args)
     "Disabled state to set: true/false."
   )
   op.add_option(
-    "--account [account]",
+    "--account [admin_account_email]",
     ->(opts, v) { opts.account = v},
-    "Account to perform update registered status as."
+    "Workbench admin account to perform update registered status as."
   )
   op.add_option(
-    "--user [user]",
+    "--user [target_email]",
     ->(opts, v) { opts.user = v},
     "User to grant or revoke registered access from."
   )
+  op.add_validator ->(opts) {
+    raise ArgumentError unless (opts.project and opts.disabled != nil and opts.account and opts.user)
+  }
   op.parse.validate
 
-  common.run_inline %W{gcloud auth login}
+  common.run_inline %W{gcloud auth login #{op.opts.account}}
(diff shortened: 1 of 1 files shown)


jaycarlton:
Could you please document the differences between this tool and the user admin page's capabilities (once it's fixed)? When  should you use each one, and do either of them trigger an email  to the user about the  change?

api/libproject/devstart.rb
```diff
@@ -1261,48 +1261,49 @@ def backfill_billing_project_owners(cmd_name, *args)
...
   )
   op.add_option(
     "--disabled [disabled]",
     ->(opts, v) { opts.disabled = v},
-    "Disabled state to set: true/false."
+    "Disabled state to set: true/false"
   )
   op.add_option(
-    "--account [account]",
+    "--account [admin_account_email]",
```
als364, line 1278: Argh! This must have been the thing missing from the command that the script didn't tell us about. Didn't think to pull the script's help in the middle of the night.

calbach:
@jaycarlton based on context it seems like you're asking about the workspace admin page. My understanding is that this page would have a button to disable all users on a workspace. I've added a description here of what the tool does, but I won't try to keep it synchronized with the exact buttons we decide to expose on this page. In general, the admin UI allows you to disable users, which is the same thing that is done here. In particular, the user admin page allows you to disable individual users, which is identical to what this CLI supports.

If there's a different time to use each of these, this will be better to put into the playbook.

api/libproject/devstart.rb
calbach: If it's any consolation, the script's help docs were pretty inaccurate. Also it was hitting the wrong URL so it would not have worked without code modifications

PR #3507, opened 2020-04-29
[RW-4864][risk=no] Fix dua completion logic for registration tasks

calbach:
Switch the registration dashboard to look at the signed DUA version instead of just the DUA completion time. This brings the UI up to date with the API's behavior on this access module. Without this, users who have completed an old DUA will be stuck on the registration homepage with no actions.

- Pull out logic for determining current DUA version, since it's now in two spots. Notes about how we could refactor in comments.
- Add regression test for this situation.

To test this, I put a user into a state where they've completed DUA v2 with no bypass in test.

Currently on test:
image

With this PR:
image

### DIFF
--- a/ui/src/app/pages/homepage/homepage.spec.tsx
+++ b/ui/src/app/pages/homepage/homepage.spec.tsx
@@ -49,6 +49,7 @@ describe('HomepageComponent', () => {
       projectId: 'aaa',
       publicApiKeyForErrorReports: 'aaa',
       enableEraCommons: true,
+      enableV3DataUserCodeOfConduct: true
     });
   });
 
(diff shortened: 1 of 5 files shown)


jaycarlton:
Seems like the front end knows s but too much about how the timestamps work
for the DUCC. Like the knowledge about completion vs bypass times should
happen in one place on the server.

But this is just a drive by observation.

calbach:
I agree, but I would like to have a plan for how DUCC version changes are going to work before I do a bunch of refactors to the API. I haven't seen this yet. The problem that was introduced by the DUCC follows:

1. New DUCC version can be pushed globally
2. The current access modules logic uses the convention of a "completionTime" per access requirement
3. There is no system by which you can clear a "completionTime" when we push a new DUA, nor is it clear this is desirable.

An alternate way of fixing this bug would be to conditionally return a duaCompletionTime IFF duaSignedVersion === active. However, I'm not crazy about this approachbecause it would be making a non-obvious transformation to the database value of the same name at serving time.

jaycarlton:
@calbach I've started a draft on my scheme for a more generalized system that wold handle some of these things. What you have is probably the most reasonable version for now.

PR #3510, opened 2020-04-30
[no ticket][risk=no] restore auth debugging line

jaycarlton:
Restore an apparently useful log statement.
---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
+++ b/api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
@@ -170,7 +170,9 @@ public class AuthInterceptor extends HandlerInterceptorAdapter {
         .setAuthentication(
             new UserAuthentication(user, OAuth2Userinfo, token, UserType.RESEARCHER));
 
-    // TODO: setup this in the context, get rid of log statement
+    // This log line is currently the the only reliable way to associate a particular App Engine request log
+    // with the authenticated user identity, which is critical information for debugging.
+    // TODO(jaycarlton) replace this log line with a UserInfo entry in a dedicated Stackdriver Auth log.
     log.log(Level.INFO, "{0} logged in", OAuth2Userinfo.getEmail());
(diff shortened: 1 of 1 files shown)


jaycarlton:
@calbach fyi

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -171,7 +171,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
             new UserAuthentication(user, OAuth2Userinfo, token, UserType.RESEARCHER));
 
     // TODO: setup this in the context, get rid of log statement
```
gjuggler, line 173: Might be worth converting this vague (and misleading?) TODO into an explainer comment describing why this log line is important. Channeling CH, it sounds like this is important to verify which user an authenticated request came from.

calbach, line 173:
```suggestion
    // TODO: Remove log statement and attach userinfo as a Stackdriver key/value instead.
    // This log line is currently the the only reliable way to associate a particular App Engine request log
    // with the authenticated user identity, which is critical information for debugging.
```

jaycarlton: @calbach if you can sketch out which properties you want in a Stackdriver log (and if you want it in JSON, which I recommend for filtering and making logs-based metrics), what keys & values would be helpful, I'll be happy to do the logging part.

jaycarlton: merging this in the meantime

calbach: Just email is fine to start to keep parity here. I don't have other ideas off the top of my head (authority and data access level are also interesting)

PR #3479, opened 2020-04-24
[RW-4645][risk=no] CircleCI config upgrade to version 2.1

aweng98:
Upgrade CircleCI yaml to version 2.1, leveraging new v2.1 features like `commands`, `parameters`, `when`. Re-organized common actions into `commands` for reuse by `jobs`.

High-level structures:
Screen Shot 2020-04-25 at 4 54 29 PM

Notes:
-  Kept all ` jobs` names unchanged to avoid adding new confusion. and number of `jobs` are not changed.
-  Use pipeline `parameters` for set/access common configs. 
-  `command` `git_checkout` combines `- checkout` and `- git submodules update`.
-  new cache `command` s. `ui_dependencies`, `e2e_dependencies` and `api_cache` with parameters.
- New anchor `filter_pr`, for running workflow jobs on non-master and non-release (tags) branches. In other words, run workflow's jobs for PR branches after pushed to origin.

I am not able test out changes in `deploy-staging`, `deploy-perf`, `api-deploy-to-test` and `ui-deploy-to-test` because these jobs are filtered by `filter_master`.

PR #3528, opened 2020-05-01
[risk=no] Fix keyfile path for staging/perf deploys

calbach:
The SA key activation was recently factored out. The refactored version writes the key in the `api` subdir, while the previous approach wrote the file to the `deploy` subdir.

https://github.com/all-of-us/workbench/pull/3479/files#diff-1d37e48f9ceff6d8030570cd36286a61L325

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -457,7 +457,7 @@ jobs:
               --git-version "${CIRCLE_TAG}" \
               --app-version "${CIRCLE_TAG}" \
               --circle-url "https://circleci.com/workflow-run/${CIRCLE_WORKFLOW_ID}" \
-              --key-file circle-sa-key.json \
+              --key-file ~/workbench/api/circle-sa-key.json \
               --promote
       - manage_api_cache:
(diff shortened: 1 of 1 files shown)


calbach:
Added explanation to PR description

.circleci/config.yml
```diff
@@ -457,7 +457,7 @@ jobs:
               --git-version "${CIRCLE_TAG}" \
               --app-version "${CIRCLE_TAG}" \
               --circle-url "https://circleci.com/workflow-run/${CIRCLE_WORKFLOW_ID}" \
-              --key-file circle-sa-key.json \
+              --key-file ~/workbench/api/circle-sa-key.json \
```
jaycarlton, line 460: @aweng98 for another day, but I see a lot of duplication of the working dir `~/workbench`. Can that be a top-level parameter (or container-level)?

Also is there a more canonical location for some of these than the user's home directory? I have a hard time understanding who the user is with Docker, but that's probably just a lack of familiarity.

PR #3529, opened 2020-05-03
[no ticket][risk=no] MapStruct Best Practices Doc

jaycarlton:
Started writing this up while working on the WorkspaceMapper and related test code for [RW-4803](https://precisionmedicineinitiative.atlassian.net/browse/RW-4803). It makes more sense as its own PR though.

---
**PR checklist**


jaycarlton:
FYI @als364 and @blrubenstein.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+Second, you don't have to rely on developers to perform awkward null
+checks for every field, as you get that for free. The generated code is very uniform and thus
+quite readable.
+
+Third, and most important, is that once you've defined a mapper, maintenance is very easy. New fields
+will simply be mapped for you (assuming they exist on both the source and the target type). If there's
+a new target field that doesn't have a sourcce, youll gget a compile-time warning reminding you. (It's
+important to stay on top of those). 
+
+### Examples
```
als364, line 24: I would prefer to have this section at the very end of the doc or possibly in a subdirectory for stories and other protracted examples. It distracts from the actual reason I come looking for this file - the coding standards we want to uphold when using this library.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+    }
+}
+```
+
+To fix the name mapping we need to add a method to build a full name from first & last names. Let's pretend our
+system didn't already have a good one, and add it directly in the mapper itself for brevity. To do
+this, we need to create a `default` method on the interface, which is a relatively new facility in
+Java to make life easier in this kind of situation.
+
+Our default method is itself another mapper, though you want to be careful when the source type of
```
als364, line 184: For what reason do you want to be careful? What does being careful or not being careful look like? What happens when you are not careful?

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+that target property.
+```java
+@Mapping(target = "billingAccountName", ignore = true)
+```
+
+While there are some places this is necessary, it should really be a last resort, for a couple of
+reasons.
+
+First, we plan to use mappers to replace existing chunks of code in a clean way. We don't want to
+have to docment that mapping function X leaves the created object in state Y with respect to fields
```
als364, line 521: docment -> document. does your mac hate u?

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+  @Mapping(source = "employeeDbEntity", target = "name")
+  @Mapping(source = "employeeDbEntity.departmentCode", target = "department")
+  @Mapping(source = "employeeDbEntity.hourlyRate", target = "salary")
+  EmployeeModel toModel(EmployeeDbEntity employeeDbEntity);
+
+  default String toNickname(EmployeeDbEntity employeeDbEntity) {
+    return String.format("%s %s", employeeDbEntity.getFirstName(), employeeDbEntity.getLastName());
+  }
+
+  default double toAnnalSalary(double hourlyRate) {
```
als364, line 403: annal -> annual

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+
+While there are some places this is necessary, it should really be a last resort, for a couple of
+reasons.
+
+First, we plan to use mappers to replace existing chunks of code in a clean way. We don't want to
+have to docment that mapping function X leaves the created object in state Y with respect to fields
+P, Q, and R. Many of our objects are relatively complex, and it's not obvious when looking at a call
+to `fooMapper.toModel()` that the output model is incomplete in some way(s).
+
+Second, if a target property is difficuult to satisfy, this is a signal that either our class design
```
als364, line 525: difficuult -> difficult. maybe it took all the u's and put them here.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+could be better, or we've left out a prereqisite.
```
als364, line 526: prereqisite -> prerequisite

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+them. Then simply specify defaults for those at the call site as appropriate.
+
+If the type of the target property has a reasonable defalt (such as zero), and `null` isn't a valid
+state for it, then using `NullValuePropertyMappingStrategy.SET_TO_DEFAULT` is an option.
+
+## Handling Default Values
+### Handling database defalts
+If the target class is a Hibernate model, and one or more fields have defaults
+defined in the database schema, we do not want to overwrite that field with `null`
+or any defauld value listed in the mapper's annotations. To get around this, we need
```
als364, line 538: defauld -> default

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+doing this, or the application should specify it. It's mainly
+useful for dealing with awkward legacy structure that you can't change.
+
+One place where this seems to be the least bad option is with
+the read-only `etag` column in the `workspace` table:
+```java
+// Use default CDR version if null. There are really no other good options (i.e. a zero version is just wrong).
+@Mapping(source = "workspace.etag", target = "version", qualifiedByName = "etagToCdrVersion", defaultValue = "1")
+```
+## Semi-advanced topics
```
als364, line 563: This is the section I most want to see in this document and it's 563 lines in. Could you move this closer to the beginning of the document?

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+
+### Mapping to a subset of target properties
+If you need to map from an object with sub-objects to a target type that's flat, there are two
+stratgegies. The first is to use dot indexing in the `@Mapping` annotation, for example
```
als364, line 567: stratgegies -> strategies

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+not sure what the convention should be on whether a target Hibernate entity should be saved inside
+the mapping method or not.
+
+## False positive matches
+Although the rules used in the codegen are reliable, they may not be intuitive when you're dealing
+with large classes with many similar fields of the same types. In particular for this project,
+be very careful with various combinations of `Workspace`, `DbWorkspace`, and `FirecloudWorkspace`. 
+- Timestamp vs SqlTimestamp. SqlTimestamp is slightly special, and many of the generated time conversions are not _quite_ what we
+want. For example, we spun our own Sql Timestamps to String mapper
+- Application-specific order dependence
```
als364, line 608: It seems like this sentence isn't quite finished

api/docs/mapstruct.md
jaycarlton: yeah. My cat was sleeping on the keyboard and I wiped all the hair and whatever oil she likes to secrete off, and the u key has never been the same since.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+First, by defining a mapper, you're adding some executable documentation to your types, and being
+explicit about how they should map.
+ 
+Second, you don't have to rely on developers to perform awkward null
+checks for every field, as you get that for free. The generated code is very uniform and thus
+quite readable.
+
+Third, and most important, is that once you've defined a mapper, maintenance is very easy. New fields
+will simply be mapped for you (assuming they exist on both the source and the target type). If there's
+a new target field that doesn't have a sourcce, youll gget a compile-time warning reminding you. (It's
```
blrubenstein, line 21: Nit: There's some double letters here.

api/docs/mapstruct.md
blrubenstein: I think my preference here is to put the examples where they are relevant. A full, complete example should go at the bottom (I agree with Ariel) and maybe have a link here. But I would love to see examples scattered throughout with each standard.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+such a mapper is a common type like String or Timestamp.
+
+```java
+@Mapper
+public interface EmployeeMapper {
+
+  @Mapping(source = "employeeDbEntity", target = "name")
+  EmployeeModel toModel(EmployeeDbEntity employeeDbEntity);
+
+  default String toName(EmployeeDbEntity employeeDbEntity) {
```
blrubenstein, line 194: These are exactly the sort of small examples I like.

api/docs/mapstruct.md
jaycarlton: Pulling it out into its own tutorial makes sense.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+```java
+@Mapper
+public interface SampleEnumMapper {
+  default Department toDepartment(int departmentCode) {
+    return Department.fromDepartmentCode(departmentCode);
+  }
+}
+```
+
+I'm sure there's a way to do this without creating another mapper class. Anyway, this gives us
```
blrubenstein, line 340: I mean, I am sure you could do it the same way you did name in the same class, but that is less reusable.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+```
+## Tips & Strategies
+
+### Stub it so it builds, and go one property at a time
+For big target classes, put in some temporary `ignore` statements or other hacks just to get it
+to the point that it generates the implemmentation class and actually compiles without errors. Then
+proceed one property at a time until all the warnings are gone.
+ 
+### Proofread the generated code
+First, you should *always* look at the generated code and step through it in the debugger. This is
```
blrubenstein, line 482: It would be great to mention some sample file names and where they appear in the filetree here. (Filetree optional, since it could move and make our documentation stale)

api/docs/mapstruct.md
blrubenstein: I am not sure I actually agree here. I do think we could probably move the big, early example down, but my preference is for the basics and then the advanced topics, so that people new to mapstruct don't get thrown in the deep end before they know the basics.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+        employeeModel.setAddress( employeeDbEntity.getAddress() );
+
+        return employeeModel;
+    }
+}
+```
+Our `@Mapping` instruction let MapStruct know which source property (an int) to match to the target
+property (`Department`). There was only one such method, so it was selected.
+
+To see what happens when there's ambiguity, add another method to `SampleEnumMapper`:
```
jmthibault79, line 365: Add how to resolve ambiguity here.  See `qualifiedByName` in `VerifiedInstitutionalAffiliationMapper`

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+## Handling missing target properties
+### Avoid `ignore = true` unless absolutely necessary
+It's annoying to see warnings about missing target properties, and for good reason. The quick way to
+stifle these is to use the `ignore` parameter, which tells MapStruct to elide the setter code for
+that target property.
+```java
+@Mapping(target = "billingAccountName", ignore = true)
+```
+
+While there are some places this is necessary, it should really be a last resort, for a couple of
```
jmthibault79, line 517: I disagree on "last resort".  Certain patterns require it, such as a DB model class with a DB-specific primary key field which isn't present in the API model class.

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+First, you should *always* look at the generated code and step through it in the debugger. This is
+because, while the codegen should be deterministic, most of us don't go around with its order of
+operator precedence in our head. Particularly if you have many input parameters with multiple fields
+with common names, you may find that `foo.id` is getting used instead of `bar.id` becase one's a 
+`long` and the other an `int`. Or something. 
+
+In general, the library gives very reasonable error
+messages and is very cranky at compile time. This is A Good Thing.
+
+## Write tests for mappers
```
calbach, line 491: Future idea: It would be nice if these Java warning for missed properties could be asserted in a simple resuable test

api/docs/mapstruct.md
jaycarlton: Thanks. It was fun and educational to go through the exercise of writing it and convincing myself I knew what I was talking about.

api/docs/mapstruct.md
jaycarlton: A Mapper can use another Mapper, so as long as you avoid cycles it's still reusable.

api/docs/mapstruct.md
jaycarlton: I yanked them from the repo. I'd need to make a new project and put them there. In the end I decided this article could be the source of truth.

If you just make a new file in the existing mappers directory (or anywhere) it'll work. Then put the test case in the same package under the test tree.

api/docs/mapstruct.md
jaycarlton: Even better, we can set up a [`@MapperConfig`](https://mapstruct.org/documentation/stable/api/org/mapstruct/MapperConfig.html) class and set the [unmappedTargetPolicy](https://mapstruct.org/documentation/stable/api/org/mapstruct/MapperConfig.html#unmappedTargetPolicy--) on it globally to [ERROR](https://mapstruct.org/documentation/1.0/api/org/mapstruct/ReportingPolicy.html) (as well as specify Spring and a couple other things that are tedious to set everywhere, like the null source policy.

Filed https://precisionmedicineinitiative.atlassian.net/browse/RW-4926

api/docs/mapstruct.md
```diff
@@ -0,0 +1,608 @@
...
+methods, such as `DbStorageEnums` can be used to good effect.
+
+### Using the `@AfterMapping` annotation
+Sometimes it's desirable to perform some fixup or post-mapping operations. While this should not be
+common, it's a handy tool to have around. In the case of target types that are generated from
+`swagger-codegen`, often the default constructor is lacking. For example it does not initialize
+array members to empty arrays but leaves them as null. We can fix this with an appropriate
+after mapping method that initializes those fields if they're not already.
+
+### Using DAOs
```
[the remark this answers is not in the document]
jaycarlton: I'll just remove this section as I have no experience. I know @ericsong mentioned this as one of the powerful features/advantages of the Spring component model version.

Depending on  how much existing code could move to mappers, I agree that read-only use cases are safer, though I could see a convention someday where we do all our saving in `@AfterMapping` methods so we don't forget.

README.md
```diff
@@ -1,11 +1,23 @@
...
-[![CircleCI Build Status](https://circleci.com/gh/all-of-us/workbench.svg)](https://circleci.com/gh/all-of-us/workflows/workbench)
+## Introduction
+The [All of Us Research Program](https://allofus.nih.gov/) Researcher Workbench is a platform for researchers to work with program
+data in a secure and convenient way. It provides users with an environment for running Jupyter notebooks
+in R or Python that can access data in the program's Curated Data Repository. For information on research
+opportunities see the [official website](https://allofus.nih.gov/get-involved/opportunities-researchers).
 
-[Documentation on API Structure](https://github.com/all-of-us/workbench/blob/master/api/doc/code-structure.md)
+While the code in this repository is available according to the terms of our  [licence](LICENSE.txt), not all features
+are available for external use. In particular, it's not crrenlty possible for third parties to build and
```
calbach, line 11: currently

PR #3534, opened 2020-05-04
[RW-4877][RISK=NO]Run RDR export at a specific time

NehaBroad:
Since i cannot test this on local i want to verify on TEST env that by following https://cloud.google.com/appengine/docs/flexible/nodejs/scheduling-jobs-with-cron-yaml

cron job will run at a specified time every day (right now at 1:00pm)
If this will work i will push another PR to correct the time (10:00pm EST)


NehaBroad:
@gjuggler  has already PR up for this ..closing this PR

PR #3535, opened 2020-05-04
[RW-4869][risk=no] CreateWorkspaceRequest

jaycarlton:
Break out a new `CreateWorkspaceRequest` to clarify required fields for creation vs fields populated later.

---
**PR checklist**


jaycarlton:
I'm also hoping to break up the actual `Workspace` model itself and have sub-objects for billing, and firecloud metadata in addition to research purpose.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
```diff
@@ -199,4 +203,25 @@ WorkspaceResource dbWorkspaceAndDbDatasetToWorkspaceResource(
   default String cdrVersionId(CdrVersion cdrVersion) {
     return String.valueOf(cdrVersion.getCdrVersionId());
   }
+
```
calbach, line 206: I thought simplicity in mappers was one of your motivations you mentioned for this. Am I missing something here? The trade offs I see overall are:

+ Improves clarity of which fields are required during creation, and offers more breathing room to clearly document implications of a field on create
- Duplicates a subset of the fields in the workspace object, which will need to be kept synchronized by hand and documented twice

The alternative to this approach would be to improve the documentation on the workspace object, and add the read-only labels as needed. This is my preference still, since the workspace object is really not that unusual. It has many read-only fields, yes, but this is pretty standard in a REST API.

jaycarlton: The problems I was running into had to do with fields that were not only optional, but not read at all during `createWorkspce`, but that were thereafter available (read-only for the purposes of that API). I've not yet run down whether any of those read-only fields are writable for updateWorkspace or any other places taking that object.

I agree that the duplication in the Swagger is undesirable. The hope is that we can consolidate somewhat with OpenApi 3. YAML anchors may not work well here.

As for simplifying the mapper, there are mechanisms to get around this, but I don't really like them. The root cause always smelled like a bad object design, but I realize API payloads are special in that regard, and we don't always expect them to be 1:1 w/r/t entities.

api/src/main/resources/workbench-api.yaml
```diff
@@ -3726,10 +3726,46 @@ definitions:
...
+      billingAccountName:
+        type: string
+        description: >
+          Name of Google Cloud Platform account responsible for billing for this
+          workspace. This will be used to locate the registerd billing account and
+          determine its status and type.
+      cdrVersionId:
+        type: string
+        description: Version staring for AoU Curated Data Repository
+      creator:
```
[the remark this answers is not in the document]
jaycarlton: good catch

api/src/main/resources/workbench-api.yaml
```diff
@@ -3726,10 +3726,46 @@ definitions:
...
+        type: string
+        description: Username of user creating this workspace
+      dataAccessLevel:
```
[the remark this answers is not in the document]
jaycarlton: This is provided by the client today. Diff of sent vs received payload attached.
[workspace_in_vs_out.pdf](https://github.com/all-of-us/workbench/files/4594247/workspace_in_vs_out.pdf)

api/src/main/resources/workbench-api.yaml
```diff
@@ -3782,15 +3818,21 @@ definitions:
     properties:
       id:
         type: string
+        readOnly: true
```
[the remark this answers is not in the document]
jaycarlton: In Java, this adds a `readOnly` attribute to the `@ApiModelProperty` annotation on the getter.
```java
   /**
   * Firecloud workspace identifier. Generated by simplifying the workspace's name or a random string if nothing is left after simplification. 
   * @return id
  **/
  @ApiModelProperty(readOnly = true, value = "Firecloud workspace identifier. Generated by simplifying the workspace's name or a random string if nothing is left after simplification. ")


  public String getId() {
    return id;
  }
```

Documented as _Allows a model property to be designated as read only._ So I guess just a decorator, possibly affecting some tooling.

In Typescript, there's apparently a builtin [`readonly` modifier](https://www.typescriptlang.org/docs/handbook/classes.html#readonly-modifier), which indicates that the property must be specified in the constructor (or left forever `null`). Which is exactly what we want.

```typescript
    /**
     * Entity tag for optimistic concurrency control. To be set during a read-modify-write to ensure that the client has not attempted to modify a changed resource.
     */
    readonly etag?: string;
```

api/src/test/java/org/pmiops/workbench/utils/TestMockFactory.java
```diff
@@ -29,60 +31,84 @@
 import org.pmiops.workbench.model.WorkspaceAccessLevel;
 import org.pmiops.workbench.notebooks.model.ClusterStatus;
 import org.pmiops.workbench.notebooks.model.ListClusterResponse;
+import org.springframework.stereotype.Service;
 
+@Service
 public class TestMockFactory {
```
jaycarlton, line 37: I'm not really happy with the way this class works. It's not really freeing me from most of the burdens of setting up the test configuration, for example. And it's not really a factory.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
calbach: Right you are. This is also a bug which is currently innocuous but would be problematic once we actually have multiple data access tiers.

api/src/main/java/org/pmiops/workbench/utils/WorkspaceMapper.java
calbach: My preferences looking at all of this:
- readonly annotations seem very useful to apply to our API models
- I'd prefer the standard REST pattern for our create methods. I'm not sold on the benefits of this approach, and I'd want a strong argument to stray from the standard pattern i.e. what makes this endpoint so unique that it needn't conform to standard REST patterns (and the Google API style guide).

To play devil's advocate a bit, there is some precedent to use a different model here for symmetry with Firecloud's underlying CreateWorkspace API (WorkspaceIngest), though I dislike that model for the same reasons.

RE anchors: I buy we might be able to avoid some duplication in the yaml, but that's only really half (or less) of the concern. There's duplication in the IDL, but then there's duplication in the generated code (tests, multiple overlapping models in the UI, etc). For example, we currently reuse the same model for the create / clone / edit pages on the UI. These are highly similar UI components which share a lot of code. The test stubs all use a single Workspace model throughout.

jaycarlton: I agree `CreateWorkspace` wasn't much of a win. The blast radius is huge and I think you're right that smart APIs and dumb models is the way to go.

I wouldn't want to use the existing Firecloud API pattern, as (1) it's probably going to change and (2) it's an implementation detail to our API client(s).

A bit about how I wound up here:

The test stubs were being sloppy and passing in post-created objects to the createWorkspace method, making it maddening to debug the tests. I think I still want separate stub methods for those, though many of the places in `WorkspacesControllerTest` that se `createWorkspace()` don't actually need to, and would be fine with a `static final Workspace`.

The reason I dug into the tests was to debug changes to `WorkspaceMapper`, which were confusing me for the same reason: so many fields are nullable vs read-only.

There's yet another branch where I have `readOnly` and `required` fields marked on the `Workspace` model. It sounds like that's the preferred path, all things considered. From an OOP standpoint I'm disgruntled, but I'll allow that it's hard to have cross-language patterns for reuse or inheritance.

So, closing this PR. May cannibalize some of the test fixes and any documentation additions. Will add `required` and `readOnly` to Workspace parameters and fixup front end code.

jaycarlton: > I thought simplicity in mappers was one of your motivations you mentioned for this.
The mapper changes are on another branch, and not yet updated for `CreateWorkspaceRequest`.

PR #3536, opened 2020-05-04
[RW-4888][risk=no] Add helptext to notebooks

als364:
This text is now on both the notebook preview page and the actual notebook page.
Screen Shot 2020-05-04 at 3 38 21 PM

---
**PR checklist**

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -142,9 +142,12 @@ const routes: Routes = [
                       // use the (urldecoded) captured value nbName
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
+                      // The iframe we use to display the Jupyter notebook does something strange
+                      // to the height calculation of the container, which is normally set to auto.
+                      // Setting this flag sets the container to 100% so that no content is clipped.
+                      contentFullHeightOverride: true,
                       helpContent: 'notebookStorage',
                       notebookHelpSidebarStyles: true,
-                      notebookRouterOutletContainerStyles: true,
                       minimizeChrome: true
                     }
                   }, {
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -142,6 +142,9 @@ const routes: Routes = [
                       // use the (urldecoded) captured value nbName
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
+                      helpContent: 'notebookStorage',
+                      notebookHelpSidebarStyles: true,
+                      notebookRouterOutletContainerStyles: true,
                       minimizeChrome: true
                     }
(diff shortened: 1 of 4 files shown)


ui/src/app/pages/workspace/workspace-wrapper/component.html
```diff
@@ -29,7 +29,7 @@ <h3 class="modal-title">Error:</h3>
                     [shareFunction]="handleShareAction"
                     [sidebarOpen]="sidebarOpen"
                     [notebookStyles]="notebookStyles"></app-help-sidebar>
-  <div [style.margin-right]="helpContent ? '45px' : 0" [style.height]="helpContent ? 'auto': '100%'">
+  <div [style.margin-right]="helpContent ? '45px' : 0" [style.height]="helpContent && !notebookRouterOutletContainerStyles ? 'auto': '100%'">
```
calbach, line 32: Please add a comment somewhere to explain what this is doing, or else rename property to be more clearly tied to the display implications per above comment

ui/src/app/app-routing.module.ts
```diff
@@ -142,6 +142,9 @@ const routes: Routes = [
                       // use the (urldecoded) captured value nbName
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
+                      helpContent: 'notebookStorage',
+                      notebookHelpSidebarStyles: true,
+                      notebookRouterOutletContainerStyles: true,
```
calbach, line 147: Is there a more semantic name that can be used here, e.g. "notebookFullscreenStyle" (or something similar - I don't follow the goal for the property). I'm not following how this relates to a router outlet.

ui/src/app/app-routing.module.ts
```diff
@@ -150,7 +153,7 @@ const routes: Routes = [
                     data: {
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
-                      helpContent: 'preview',
+                      helpContent: 'notebookStorage',
```
calbach, line 156: Perhaps this would be explained by the other comments, but why is the new property not set here?

als364: Screen Shot 2020-05-04 at 3 02 55 PM
`height: auto` behaves strangely when we're actually displaying the notebook. The content of the div is large enough that it ought to make the container div grow, but it's not doing that.

ui/src/app/app-routing.module.ts
als364: How about something like `fullHeightOverride`?

calbach: OK I see what it's doing now. Maybe `containerFullHeightOverride` or `contentFullHeightOverride`? 

Comment in workspace-wrapper wouldn't hurt as well to explain that it's needed primarily due to the notebooks page iframe

PR #3539, opened 2020-05-05
[RW-4889][risk=no] Folder icon on help sidebar

als364:
Screen Shot 2020-05-05 at 11 52 01 AM
Screen Shot 2020-05-05 at 11 52 09 AM

---
**PR checklist**

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -38,6 +38,7 @@ import {environment} from 'environments/environment';
 import {DisabledGuard} from './guards/disabled-guard.service';
 import {InteractiveNotebookComponent} from './pages/analysis/interactive-notebook';
 import {BreadcrumbType, NavStore} from './utils/navigation';
+import {NOTEBOOK_HELP_CONTENT} from "./components/help-sidebar";
 
 
 declare let gtag: Function;
@@ -148,7 +149,7 @@ const routes: Routes = [
                       // to the height calculation of the container, which is normally set to auto.
                       // Setting this flag sets the container to 100% so that no content is clipped.
                       contentFullHeightOverride: true,
-                      helpContent: 'notebookStorage',
+                      helpContent: NOTEBOOK_HELP_CONTENT,
                       notebookHelpSidebarStyles: true,
                       minimizeChrome: true
                     }
@@ -158,7 +159,7 @@ const routes: Routes = [
                     data: {
                       pathElementForTitle: 'nbName',
                       breadcrumb: BreadcrumbType.Notebook,
-                      helpContent: 'notebookStorage',
+                      helpContent: NOTEBOOK_HELP_CONTENT,
                       notebookHelpSidebarStyles: true,
                       minimizeChrome: true
                     }
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -105,7 +105,8 @@ const routes: Routes = [
                   breadcrumb: BreadcrumbType.Workspace,
                   helpContent: 'about'
                 }
-              }, {
+              },
+              {
                 path: 'edit',
                 component: WorkspaceEditComponent,
                 data: {
@@ -114,7 +115,8 @@ const routes: Routes = [
                   breadcrumb: BreadcrumbType.WorkspaceEdit,
                   helpContent: 'edit'
                 }
-              }, {
+              },
+              {
                 path: 'duplicate',
                 component: WorkspaceEditComponent,
                 data: {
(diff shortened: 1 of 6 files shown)


ui/src/app/app-routing.module.ts
```diff
@@ -105,7 +105,8 @@ const routes: Routes = [
                   breadcrumb: BreadcrumbType.Workspace,
                   helpContent: 'about'
                 }
-              }, {
+              },
```
als364, line 108: it was bugging me that I couldn't collapse stuff. I can revert this if you, esteemed reviewer, would prefer.

ui/src/app/components/help-sidebar.tsx
```diff
@@ -180,23 +181,24 @@ const icons = [{
...
-}, {
-  id: 'annotations',
-  disabled: false,
-  faIcon: faEdit,
-  label: 'Annotations Icon',
-  page: 'reviewParticipantDetail',
-  style: {fontSize: '20px', marginLeft: '3px'},
-  tooltip: 'Annotations',
-}];
+  }, {
```
als364, line 184: this is a whitespace change

ui/src/app/components/help-sidebar.tsx
```diff
@@ -428,27 +434,31 @@ export const HelpSidebar = fp.flow(withCurrentWorkspace(), withUserProfile())(
...
-                {icon.id === 'dataDictionary'
-                  ? <a href={dataDictionaryUrl} target='_blank'>
-                      <FontAwesomeIcon icon={icon.faIcon} style={icon.style} />
-                    </a>
-                  : icon.faIcon === null
-                    ? <img src={proIcons[icon.id]} style={icon.style} />
-                    : <FontAwesomeIcon icon={icon.faIcon} style={icon.style} />
-                }
+          {icons(helpContent).map((icon, i) =>
+            (!icon.page || icon.page === helpContent) &&
```
als364, line 438: this is also largely a whitespace change

ui/src/app/pages/workspace/workspace-wrapper/component.ts
```diff
@@ -42,6 +42,9 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
   helpContent = 'data';
   sidebarOpen = false;
   notebookStyles = false;
+  // The iframe we use to display the Jupyter notebook does something strange
```
als364, line 45: forgot to put this in the last PR - so putting it here

ui/src/app/components/help-sidebar.tsx
```diff
@@ -164,14 +164,15 @@ const iconStyles = {
...
 // TODO uncomment 'thunderstorm' icon when cluster configuration function is ready
-const icons = [{
-  id: 'help',
-  disabled: false,
-  faIcon: faInfoCircle,
-  label: 'Help Icon',
-  page: null,
-  style: {fontSize: '21px'},
-  tooltip: 'Help Tips',
+const icons = (helpContent) => {
```
gjuggler, line 167: I'd add a type annotation for the function param at least. Even better, a one-line comment describing the 'helpContent' parameter and why it's needed.

[optional] If you're game for as mass-rename, I think s/helpContent/helpContentKey/ in this file would help reduce confusion throughout, since AFAICT this is really a key into the foreign JSON file which contains the actual content.

ui/src/app/components/help-sidebar.tsx
```diff
@@ -164,14 +164,15 @@ const iconStyles = {
...
+  return [{
+    id: 'help',
+    disabled: false,
+    faIcon: helpContent === 'notebookStorage' ? faFolderOpen : faInfoCircle,
```
gjuggler, line 171: [opt] Magic string code smell – for safety, it may help to break this out into a top-level constant, with a comment clarifying that it needs to match the strings used in app/app-routing-module.ts and assets/json/help-sidebar.json.

I also would love to see a comment here clarifying why we switch to a folder icon when the "notebookStorage" help content is being displayed. Something to the effect of "Because the 'notebookStorage' help text is about workspace storage, we switch the icon to be a folder instead of a question mark."

(Unrelated comment: do we know of any technical reason why we are loading the help-sidebar content via a static JSON file rather than including it as an object within our Typescript code? The latter approach would add some level of type safety and would keep the content closer to the code which uses it.)

lstrano-broad:
Icon looks good - thx

ui/src/app/components/help-sidebar.tsx
als364: I don't know of any technical reason why we're loading the content this way, aside from that the content is not really code. I'm sure there's some historical thing lost to the centuries.

PR #3550, opened 2020-05-07
[RW-3942][risk=moderate] Add readonly database user (1/2)

calbach:
- Adds a new database user: `dev-readonly`, intended for developer readonly access. Our other usernames are: `workbench` (i.e. the application), `root`, and `liquibase`.
- Adds a new flag to `./project.rb connect-to-cloud-db` of `--db-user`. Defaults to the new readonly user.

If this looks good, I will need to execute the following:

- Apply a manual process to backfill the new username and generated passwords into all vars.env files in all environments
- ~~Manually run the create_db script in all environments to backfill the new user. Alternatively may issue this as manual SQL commands, will figure this out on the ticket.~~
--> **TIL**: create_db.sql gets applied on all migrations

### DIFF
--- a/api/db/create_db.sql
+++ b/api/db/create_db.sql
@@ -1,16 +1,20 @@
 CREATE DATABASE IF NOT EXISTS ${DB_NAME} CHARACTER SET utf8 COLLATE utf8_general_ci;
 
+CREATE USER IF NOT EXISTS '${DEV_READONLY_DB_USER}'@'%' IDENTIFIED BY '${DEV_READONLY_DB_PASSWORD}';
 CREATE USER IF NOT EXISTS '${WORKBENCH_DB_USER}'@'%' IDENTIFIED BY '${WORKBENCH_DB_PASSWORD}';
 CREATE USER IF NOT EXISTS '${LIQUIBASE_DB_USER}'@'%' IDENTIFIED BY '${LIQUIBASE_DB_PASSWORD}';
 -- In case the users already exist, change their passwords.
+SET PASSWORD FOR '${DEV_READONLY_DB_USER}'@'%' = '${DEV_READONLY_DB_PASSWORD}';
 SET PASSWORD FOR '${WORKBENCH_DB_USER}'@'%' = '${WORKBENCH_DB_PASSWORD}';
 SET PASSWORD FOR '${LIQUIBASE_DB_USER}'@'%' = '${LIQUIBASE_DB_PASSWORD}';
+
+-- Grant readonly access to all tables, captures Workbench, CDR, and Liquibase.
+GRANT SELECT, CREATE TEMPORARY TABLES ON * TO '${DEV_READONLY_DB_USER}'@'%';
 
 GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON ${DB_NAME}.* TO '${WORKBENCH_DB_USER}'@'%';
 GRANT SELECT, INSERT, UPDATE, DELETE, DROP, ALTER, CREATE, INDEX, REFERENCES, CREATE TEMPORARY TABLES, CREATE VIEW ON ${DB_NAME}.* TO '${LIQUIBASE_DB_USER}'@'%';
 
--- Give wildcard permission to cdr databases for workbench
--- cdr* is the older and/or local naming convention, synth_r_*, r_* is new
-GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON `cdr%`.* TO '${WORKBENCH_DB_USER}'@'%';
+-- Give wildcard permission to cdr databases for workbench/readonly
+-- synth_r_*, r_* is the CDR database naming convention for synthetic and prod respectively
 GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON `r_%`.* TO '${WORKBENCH_DB_USER}'@'%';
 GRANT SELECT, INSERT, UPDATE, DELETE, CREATE TEMPORARY TABLES ON `synth_r_%`.* TO '${WORKBENCH_DB_USER}'@'%';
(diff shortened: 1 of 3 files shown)


api/libproject/devstart.rb
```diff
@@ -2020,20 +2022,36 @@ def connect_to_cloud_db(cmd_name, *args)
...
+  end
+
   CloudSqlProxyContext.new(gcc.project).run do
-    password = op.opts.root ? env["MYSQL_ROOT_PASSWORD"] : env["WORKBENCH_DB_PASSWORD"]
-    user = op.opts.root ? "root" : env["WORKBENCH_DB_USER"]
     common.run_inline %W{
-      mysql --host=127.0.0.1 --port=3307 --user=#{user}
-      --database=#{env["DB_NAME"]} --password=#{password}},
-      password
+      mysql --host=127.0.0.1 --port=3307 --user=#{op.opts.db_user}
```
jaycarlton, line 2052: Can you put a prominent notice right above here that the session is or isn't read-only?

api/libproject/devstart.rb
```diff
@@ -1277,6 +1277,8 @@ def write_db_creds_file(project, cdr_db_name, root_password, workbench_password)
       db_creds_file.puts "MYSQL_ROOT_PASSWORD=#{root_password}"
       db_creds_file.puts "WORKBENCH_DB_USER=workbench"
       db_creds_file.puts "WORKBENCH_DB_PASSWORD=#{workbench_password}"
+      db_creds_file.puts "DEV_READONLY_DB_USER=dev-readonly"
```
jaycarlton, line 1280: Maybe for a later pass, but do we have a policy about which systems we need named accounts per user for and which ones we can share? If at some point we wanted to audit who was running which statements, it would be helpful to associate these sessions with names.

Or maybe there's another way to capture that outside of mysql.

api/db/create_db.sql
```diff
@@ -1,16 +1,20 @@
 CREATE DATABASE IF NOT EXISTS ${DB_NAME} CHARACTER SET utf8 COLLATE utf8_general_ci;
 
+CREATE USER IF NOT EXISTS '${DEV_READONLY_DB_USER}'@'%' IDENTIFIED BY '${DEV_READONLY_DB_PASSWORD}';
```
jaycarlton, line 3: nit: I think this would be a bit more readable if the CREATE/PASSWORD/GRANT statements were grouped by user instead of by statement type.

api/libproject/devstart.rb
```diff
@@ -2020,20 +2022,36 @@ def connect_to_cloud_db(cmd_name, *args)
   common = Common.new
   op = WbOptionsParser.new(cmd_name, args)
   op.add_option(
```
jaycarlton, line 2024: If I were looking at the help text quickly, I'd probably be looking for an option like `--read-only` or `--access=read-only` rather than `--db-user` (which is a bit of an implementation detail).

api/libproject/devstart.rb
```diff
@@ -2020,20 +2022,36 @@ def connect_to_cloud_db(cmd_name, *args)
...
   op.parse.validate
   gcc.validate
+
+  if op.opts.db_user.nil?
+    op.opts.db_user = "dev-readonly"
+  end
+
   env = read_db_vars(gcc)
+  db_password = nil
+  case op.opts.db_user
```
jaycarlton, line 2038: I'd probably make a map from user name to env variable name and do a lookup instead of a `case` here.

api/libproject/devstart.rb
jaycarlton, line 2052: nit: might want the ability to override this port from a citizenship standpoint

api/libproject/devstart.rb
```diff
@@ -2490,12 +2508,12 @@ def create_project_resources(gcc)
 end
 
 def setup_project_data(gcc, cdr_db_name)
-  root_password, workbench_password = random_password(), random_password()
+  root_password, workbench_password, readonly_password = random_password(), random_password(), random_password()
```
jaycarlton, line 2511: Please break this into three assignments, one on each line, or make a method that outputs an array of random passwords so it's like `root_password, workbench_password, readonly_password = *random_passwords(3)`. I stared at it for a good minute before I realized it was a multi-assignment.

calbach: sure

api/libproject/devstart.rb
calbach: done

api/libproject/devstart.rb
calbach: In general we don't have a critical dependency on mysql users for auditing nor for authentication (except as defense-in-depth), since CloudSQL has its own mechanisms outside of MySQL for managing this (i.e. cloud sql proxy). The current convention is to add users is when we need different functional permissions for different things and there are good cut points to allow granting least privilege (mostly to mitigate security exploits / bugs).

For example, if we had a separate GAE service connecting to the same DB for a slightly different purpose (e.g. different subset of tables), another SQL user would probably make sense.

I could see an argument for a `dev-writer` user to allow it to evolve separately of `workbench`, and for logging purposes on the off-chance this surfaces somewhere useful.

Going for something like one user per developer I think would be a pain to maintain and would not gain us much.

api/db/create_db.sql
calbach: I'd rather keep all the user initialization first before executing any other GRANT statements

api/libproject/devstart.rb
calbach: I added more to the flag doc. The problem with that approach is that we have existing functionality which supports signing in as --root instead. So the --readonly proposal means we now have mutually exclusive flags with confusing interactions, since the implementations are all driven by the database user e.g. --readonly, --db-user or , --readonly and --root

I'd prefer to leave this general and add documentation. Also - I feel readonly should be the default, so I don't like it as a flag. --no-readonly maybe (suffers from same issues).

api/libproject/devstart.rb
calbach: done

calbach: Probably. I can send a follow-up PR - it's not related to this change

calbach:
Thinking this through more carefully; this should be  two parter. First I will add the user, then I will update the CLI. The new user will roll out with the deployments, and I don't want the CLI to be broken in the interim.

calbach:
If CI passes, I will merge since this is a subset of the original approved changes.

PR #3554, opened 2020-05-07
[RW-4908][risk=no] More aggressively clear ?workbenchAccessTasks=true

gjuggler:
This PR attempts to clean up some of our user registration navigation endpoints, to clear out the `?workbenchAccessTasks=true` query param if it has ever been set. Some users encountered this issue when walking through registration, where the "access tasks" dashboard never went away.

Unfortunately, the way the registration tasks component is loaded & shown as part of the homepage components makes the fix here quite ugly. (I'd attempted to motivate these two components to be entirely different routes during its original implementation, but didn't succeed.)

While walking through the registration flow, I also noticed that DUCC uses browser.back() instead of navigating to the homepage component. I think the latter is more appropriate, and also would have fixed the issue in RW-4908.

I'm not 100% sure this PR is even worth merging – it could provide us defense in depth against RW-4908 regressions, but it may be introducing too much add'l complexity.

---
**PR checklist**

- [NO SORRY] This PR includes appropriate unit tests
- [NO] I have run the E2E tests on ths change against my local UI + API server with `yarn test-local`

### DIFF
--- a/ui/src/app/pages/homepage/registration-dashboard.tsx
+++ b/ui/src/app/pages/homepage/registration-dashboard.tsx
@@ -402,7 +402,7 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
              data-test-id='success-message'>
           You successfully completed all the required steps to access the Researcher Workbench.
           <Button style={{marginLeft: '0.5rem'}}
-                  onClick={() => window.location.reload()}>Get Started</Button>
+                  onClick={() => navigate([''])}>Get Started</Button>
         </div>
       }
       {this.state.twoFactorAuthModalOpen && <Modal width={500}>
--- a/ui/src/app/pages/homepage/registration-dashboard.tsx
+++ b/ui/src/app/pages/homepage/registration-dashboard.tsx
@@ -402,7 +402,15 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
              data-test-id='success-message'>
           You successfully completed all the required steps to access the Researcher Workbench.
           <Button style={{marginLeft: '0.5rem'}}
-                  onClick={() => navigate([''])}>Get Started</Button>
+                  onClick={() => {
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/homepage/registration-dashboard.tsx
```diff
@@ -398,7 +398,15 @@ export class RegistrationDashboard extends React.Component<RegistrationDashboard
              data-test-id='success-message'>
           You successfully completed all the required steps to access the Researcher Workbench.
           <Button style={{marginLeft: '0.5rem'}}
-                  onClick={() => window.location.reload()}>Get Started</Button>
+                  onClick={() => {
+                    // Quirk / hack note: the goal here is to send the user to the homepage once they've completed
+                    // all access modules. Normally we would just navigate(['']) to do this. However, because
+                    // of the way this dashboard is rendered *within* the homepage component, a call to
+                    // navigate is not enough to trigger the normal homepage to load. As a workaround, we
```
als364, line 405: Does `navigate(['']); window.location.reload()` work or does that revert the url to include the search again?

gjuggler: The latter – I tried that first, and unfortunately the window.location.reload() uses the original window.location, apparently before navigate(['']) has any effect on changing it.

als364: rats.

Is this related to RW-4915 or just discovered around the same time

gjuggler: Unrelated, just discovered around the same time.

gjuggler:
> It looks like this somehow broke the UI tests - good to go when you fix these (but please alert me to the changes that fix them so I can see what they are)

Turns out I was using an incorrect call to navigate. Adding a forward slash to the argument fixes the broken test.

als364:
aha! yay! the tests caught something!

PR #3560, opened 2020-05-08
[no ticket][risk=no] Handy bash Ops snippets

jaycarlton:
Some bash one-liners from January & February that are handy for working with Stackdriver Monitoring. 

---
**PR checklist**


ops/bash/create-dashboard.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+curl -X POST "https://monitoring.googleapis.com/v1/projects/$(gcloud config get-value project)/dashboards" \
```
jmthibault79, line 3: Looks like this uses the default project you have set in your local gcloud config.  I have found default projects to be cumbersome when working in many different projects (like we do) and I often find myself doing the wrong thing.

So my preference is to avoid them.  Do you have a good system for working with them?

ops/bash/README.md
```diff
@@ -0,0 +1,11 @@
+# Bash Ops Snippets
+This directory has a smattering of bash goodies for working with GCP and Stackdriver Monitoring
+especially.
+
+## Prerequisites
+* `bash`
+* `gcloud`
+* `curl`
```
jmthibault79, line 8: also `jq`

ops/bash/create-dashboard.sh
calbach: +1 - this should be avoided. The biggest risk is that you set this to prod, forget, start working on something entirely unrelated and implicitly start mutating production

jaycarlton: Fair enough, though no worse than the Cloud Console. In almost all cases, the target name will include a project identifier, but I agree it's not a spectacular practice when test and production share the same account.

PR #3562, opened 2020-05-09
[RW-4904][risk=no] cron job schedule updates

jaycarlton:
Based on conversation on the spreadsheet I've made a few schedule adjustments. I've also tried to elaborate on the descriptions where I felt those weren't obvious at a glance.

I'll have to check on Monday if there's a good way to test this locally. I suspect I'll need to deploy to test and keep an eye on it.

*TIL*
- the AppEngine section in the Cloud Console has a nifty page describing the cron jobs, which displays the description from the cron.
---
**PR checklist**

### DIFF
--- a/api/src/main/webapp/WEB-INF/cron_default.yaml
+++ b/api/src/main/webapp/WEB-INF/cron_default.yaml
@@ -4,24 +4,44 @@ cron:
   schedule: every 3 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of compliance training status (via Moodle API)'
+- description: |
+    Daily sync of compliance training status (via Moodle API). This is just a backup to the primary
+    flow. This would only ever catch the following scenarios:
+    1. The user's compliance training expires. The time scale on this expiration is O(years),
+    so syncing nightly is entirely acceptable.
+
+    2. The user completes compliance and doesn't return to the dashboard. We'd progress them in the
+    background. This won't make much of a difference and would have the same effect as if they had
+    navigated back to the dashboard.
+
+    3. Somehow the user manages to "uncomplete" training in Moodle. I'm not sure this is possible,
+    but if so, it's certainly an edge case where nightly would be acceptale latency
   url: /v1/cron/bulkSyncComplianceTrainingStatus
   schedule: every 24 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of eRA Commons linkage status (via FireCloud API)'
+- description: >
+    Daily sync of eRA Commons linkage status (via FireCloud API).  I expect this cron to never
+    detect anything. This is effectively checking for whether any changes happened in the
+    background between the Terra account and its linkage to eRA commons. During onboarding,
+    this won't matter since it's checked live from the dashboard. After onboarding, I don't even
+    know whether it's possible to "unlink", and if a user did this, whether that even matters for
+    our registration policy. Certainly it's good to check, but none of this is time sensitive.
+    The current 24h interval is fine.
   url: /v1/cron/bulkSyncEraCommonsStatus
   schedule: every 24 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of 2FA set up status (via Google Directory Service)'
+- description: >
+    Update our cache of users' two-factor authentication settings on their GSuite accounts via
+    Google Directory Service.
   url: /v1/cron/bulkSyncTwoFactorAuthStatus
   schedule: every 24 hours
   timezone: UTC
   target: api
 - description: 'Daily audit of gcp resources that users have access to'
   url: /v1/cron/bulkAuditProjectAccess
-  schedule: every 24 hours
+  schedule: every 1 hours
   timezone: UTC
   target: api
 - description: 'If the AoU Billing Project buffer is not full, refill with one or more projects.'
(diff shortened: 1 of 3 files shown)


api/src/main/webapp/WEB-INF/cron_default.yaml
```diff
@@ -4,24 +4,44 @@ cron:
...
   url: /v1/cron/bulkSyncComplianceTrainingStatus
   schedule: every 24 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of eRA Commons linkage status (via FireCloud API)'
+- description: >
+    Daily sync of eRA Commons linkage status (via FireCloud API).  I expect this cron to never
+    detect anything. This is effectively checking for whether any changes happened in the
+    background between the Terra account and its linkage to eRA commons. During onboarding,
+    this won't matter since it's checked live from the dashboard. After onboarding, I don't even
```
calbach, line 27: My commentary here is way too ambiguous to be checked in. @gjuggler may be able to provide a better description. But before I wrote this I'd need to refresh on exactly what the intended behavior is. In particular, I believe the current registration system as designed, only cares that a linkage was established at one point, for verification - this is purely a stand-in for ID verification. If so, this background sync would actually behave incorrectly if a user unlinked their account, or let their linkage expire.

However, for controlled tier, there might be new policies which require an active eRA commons linkage, in case this stricter behavior would be required, including a linkage refresh.

api/src/main/webapp/WEB-INF/cron_default.yaml
```diff
@@ -4,24 +4,44 @@ cron:
...
+    know whether it's possible to "unlink", and if a user did this, whether that even matters for
+    our registration policy. Certainly it's good to check, but none of this is time sensitive.
+    The current 24h interval is fine.
   url: /v1/cron/bulkSyncEraCommonsStatus
   schedule: every 24 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of 2FA set up status (via Google Directory Service)'
+- description: >
+    Update our cache of users' two-factor authentication settings on their GSuite accounts via
```
calbach, line 36: nit: would probably say "database cache" to avoid ambiguity here. If I read this as-is I would probably assume this meant memcache or other ephemeral storage.

api/src/main/webapp/WEB-INF/cron_default.yaml
```diff
@@ -41,24 +61,31 @@ cron:
...
 - description: 'Find and alert users that have exceeded their free tier billing usage'
   url: /v1/cron/checkFreeTierBillingUsage
-  schedule: every 3 hours
+  schedule: every 30 minutes
   timezone: UTC
   target: api
-- description: 'Find billing projects associated with deleted workspaces and transfer ownership to Garbage Collection Service Accounts'
+- description: >
+    Find billing projects associated with deleted workspaces and transfer ownership to
+    designated "Garbage Collection" Service Accounts. This process works around the lack of API
```
calbach, line 69: That would have been true when this API was created. However, this API now exists and we have a ticket to eliminate this cron and move to the new method: https://precisionmedicineinitiative.atlassian.net/browse/RW-3627. I would update this to reflect that and probably link the ticket here

api/src/main/webapp/WEB-INF/cron_default.yaml
```diff
@@ -4,24 +4,44 @@ cron:
   schedule: every 3 hours
   timezone: UTC
   target: api
-- description: 'Daily sync of compliance training status (via Moodle API)'
+- description: |
+    Daily sync of compliance training status (via Moodle API). This is just a backup to the primary
```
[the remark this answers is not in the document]
jaycarlton: Sounds good. FWIW the UI is pretty responsive and reflows the text well.

api/src/main/webapp/WEB-INF/cron_default.yaml
```diff
@@ -4,24 +4,44 @@ cron:
...
+    flow. This would only ever catch the following scenarios:
+    1. The user's compliance training expires. The time scale on this expiration is O(years),
+    so syncing nightly is entirely acceptable.
+
+    2. The user completes compliance and doesn't return to the dashboard. We'd progress them in the
+    background. This won't make much of a difference and would have the same effect as if they had
+    navigated back to the dashboard.
+
+    3. Somehow the user manages to "uncomplete" training in Moodle. I'm not sure this is possible,
```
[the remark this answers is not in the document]
jaycarlton: Is the Queen going to be inspecting the cron dashboard? 😉

api/src/main/webapp/WEB-INF/cron_default.yaml
jaycarlton: I don't want to go too far down a rabbit hole, but it sounds like we're collecting this information but never acting on it. It just logs things to INFO but doesn't ring any bells elsewhere as far as I can see.

gjuggler: No, it's not wrong to be digging into the details here. [This line](https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/api/OfflineUserController.java#L142) should be updating user status (including registered-tier membership) based on the updated eRA Commons linkage state synced from Terra.

PR #3563, opened 2020-05-09
[RW-4933][risk=no] Revert docker_run change

calbach:
Regression: `run_inline` executes and then continues, `exec` replaces the current process. Revert this change from https://github.com/all-of-us/workbench/pull/3548/files


calbach:
It doesn't have any flags

PR #3565, opened 2020-05-11
[no ticket][risk=no] Allow RDR export cron failures to throw a server error

gjuggler:
This is a small follow-up based on Jay's observation that exceptions in the RDR export cronjob handler aren't causing the service to return a 500 response.

Screenshot demonstrating this issue (see the line with a severe error log but a 204 response):

Screen Shot 2020-05-11 at 10 14 00 AM

TESTING: I haven't tested locally. I'm willing to try and set up some unit tests for RDR export controller (no test file exists yet), but it will take some DI finagling to get working correctly, so I've punted on this for a first cut of this PR.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
@@ -19,7 +19,6 @@ import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.config.WorkbenchLocationConfigService;
 import org.pmiops.workbench.exceptions.ServerErrorException;
 import org.pmiops.workbench.rdr.RdrExportService;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.RestController;
 
@@ -27,7 +26,7 @@ import org.springframework.web.bind.annotation.RestController;
  * This offline process is responsible to daily sync up with RDR by creating/pushing multiple Cloud
  * Task tasks with eligible user Ids and workspace Ids to cloud task queue.
  *
- * None of the actual RDR communication occurs from within this cron job handler, so this cron
+ * <p>None of the actual RDR communication occurs from within this cron job handler, so this cron
  * may succeed even though the actual export tasks to RDR are failing. See
  * CloudTaskRdrExportController for the handlers which do the actual RDR export work.
  *
@@ -76,7 +75,8 @@ public class OfflineRdrExportController implements OfflineRdrExportApiDelegate {
         groupIdsAndPushTask(rdrExportService.findAllWorkspacesIdsToExport(), EXPORT_USER_PATH);
       } catch (Exception ex) {
         throw new ServerErrorException(
-            String.format("Error creating RDR export Cloud Tasks for workspaces: %s", ex.getMessage()));
+            String.format(
+                "Error creating RDR export Cloud Tasks for workspaces: %s", ex.getMessage()));
       }
     }
     return ResponseEntity.noContent().build();
@@ -113,7 +113,8 @@ public class OfflineRdrExportController implements OfflineRdrExportApiDelegate {
     }
   }
 
-  private void createAndPushTask(List<Long> ids, String queuePath, String taskUri) throws IOException {
+  private void createAndPushTask(List<Long> ids, String queuePath, String taskUri)
+      throws IOException {
     List<String> idsAsString = ids.stream().map(id -> id.toString()).collect(Collectors.toList());
     Gson gson = new Gson();
     String daysJson = gson.toJson(ids);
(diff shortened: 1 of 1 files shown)


gjuggler:
@jaycarlton and @NehaBroad here's a non-urgent, targeted fix to improve the error response handling of the RDR export cron. There's no rush – I'll be sure to wait for Neha's to have a look before this merges.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -63,14 +68,15 @@
       try {
         groupIdsAndPushTask(rdrExportService.findAllUserIdsToExport(), EXPORT_RESEARCHER_PATH);
       } catch (Exception ex) {
-        log.severe(
-            String.format("Error while exporting researcher data to RDR: %s", ex.getMessage()));
+        throw new ServerErrorException(
```
jaycarlton, line 71: In a controller you might want to throw a more specific error here and let the handler translate it into a 500, but I'm not sure if that happens with crons. This is fine.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -63,14 +68,15 @@
...
+            String.format("Error creating RDR export Cloud Tasks for users: %s", ex.getMessage()));
       }
       try {
         groupIdsAndPushTask(rdrExportService.findAllWorkspacesIdsToExport(), EXPORT_USER_PATH);
       } catch (Exception ex) {
-        log.severe(
-            String.format("Error while exporting workspace data to RDR: %s", ex.getMessage()));
+        throw new ServerErrorException(
+            String.format(
+                "Error creating RDR export Cloud Tasks for workspaces: %s", ex.getMessage()));
```
jaycarlton, line 79: If you use this [constructor](https://docs.oracle.com/javaee/7/api/javax/ws/rs/ServerErrorException.html#ServerErrorException-javax.ws.rs.core.Response-java.lang.Throwable-) that takes the throwable itself, you keep the stack trace.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -127,8 +134,10 @@ private void createAndPushTask(List<Long> ids, String queuePath, String taskUri)
     } catch (IOException ex) {
```
jaycarlton, line 134: drive-by nit: Above, I'd declare a `TaskBuilder` and then use `taskBuilder.build()` in the cal to `createTask()`. Or just inline the builder.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
gjuggler: Thanks – done

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
gjuggler: Done

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -17,13 +17,18 @@
...
 
 /**
- * This offline process is responsible to daily sync up with RDR by creating/pushing multiple task
- * with eligible user Ids and workspace Ids to cloud task queue
+ * This offline process is responsible to daily sync up with RDR by creating/pushing multiple Cloud
+ * Task tasks with eligible user Ids and workspace Ids to cloud task queue.
+ *
+ * <p>None of the actual RDR communication occurs from within this cron job handler, so this cron
+ * may succeed even though the actual export tasks to RDR are failing. See
+ * CloudTaskRdrExportController for the handlers which do the actual RDR export work.
```
jaycarlton, line 31: So we need to trigger an alert on the tasks failing in the queue. I bet I can get that from logs somewhere, assuming the tasks live in our GCP-space.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
```diff
@@ -83,7 +86,7 @@
    * @param idList : Lis of Ids
    * @param taskUri: The destination URL the task will be calling with group of 10 ids
    */
-  private void groupIdsAndPushTask(List<Long> idList, String taskUri) {
+  private void groupIdsAndPushTask(List<Long> idList, String taskUri) throws IOException {
```
jaycarlton, line 89: nit: Should the IDs be a Set? Alternatively, just take in a `Collection<Long>`.

api/src/main/java/org/pmiops/workbench/api/OfflineRdrExportController.java
gjuggler: Yes – this will be an important metric to track & alert on too.

PR #3568, opened 2020-05-11
[RW-3942][risk=low] Add a new standalone command to create/update the DB (1.5/2)

calbach:
On further review, create_db.sql isn't run during deployments. Add a standalone command to apply this per environment as needed. This would only be needed when user creation, grants, or passwords need updating.

```
./project.rb create-or-update-workbench-db --project ...
```

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -2230,7 +2230,13 @@ end
 
 Common.register_command({
   :invocation => "create-or-update-workbench-db",
-  :description => "Creates or updates the Workbench database and users",
+  :description => "Creates or updates the Workbench database and users.\n" \
+                  "This can be used in the event that new users or permissions " \
+                  "are added to create_db.sql, as this is not currently rerun " \
+                  "during deployment. This process is separate from Liquibase " \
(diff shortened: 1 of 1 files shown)


api/libproject/devstart.rb
```diff
@@ -2221,6 +2221,25 @@ def create_workbench_db()
   )
 end
 
+def create_or_update_workbench_db(cmd_name, args)
+  ensure_docker cmd_name, args
+  with_cloud_proxy_and_db_env(cmd_name, args) do |ctx|
+    create_workbench_db
```
jaycarlton, line 2227: So this is just another entry point to what we aleready have, which makes sense.

One confusing thing is that this always runs a had-coded create_db.sql script. Which makes it feel like a wart in some ways. It's also scary to run something called `create_workbench_db` when you're just doing an update.

calbach: Fair point on naming, renamed.

The choice for SQL and XML templates in our dev tooling is to use env substitution. For better or worse, that is the convention. I've added some comments to make the implicit parameterization that's happening here hopefully more clear.

PR #3569, opened 2020-05-11
[RW-3942][risk=low] Leverage new readonly user in developer CLI tooling (2/2)

calbach:
This was already reviewed/approved during https://github.com/all-of-us/workbench/pull/3550, which was partially reverted. Adds a new flag to `./project.rb connect-to-cloud-db of --db-user.` Defaults to the new readonly user.

**Blocked on**: https://precisionmedicineinitiative.atlassian.net/browse/PD-5108

### DIFF
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -2079,11 +2079,10 @@ def connect_to_cloud_db_binlog(cmd_name, *args)
     common.status "\n" + "*" * 80
     common.status "Listing available journal files: "
 
-    # "root" is required for binlog access.
-    password = env["MYSQL_ROOT_PASSWORD"]
+    password = env["DEV_READONLY_DB_PASSWORD"]
     run_with_redirects(
       "echo 'SHOW BINARY LOGS;' | " +
-      "mysql --host=127.0.0.1 --port=3307 --user=root " +
+      "mysql --host=127.0.0.1 --port=3307 --user=dev-readonly " +
       "--database=#{env['DB_NAME']} --password=#{password}", password)
     common.status "*" * 80
 
@@ -2097,7 +2096,7 @@ def connect_to_cloud_db_binlog(cmd_name, *args)
     # Work out of /tmp for easy local file redirection. We don't want binlogs
     # winding up back in Workbench source control accidentally.
     run_with_redirects(
-      "export MYSQL_HOME=$(with-mysql-login.sh root #{password}); " +
+      "export MYSQL_HOME=$(with-mysql-login.sh dev-readonly #{password}); " +
       "cd /tmp; /bin/bash", password)
   end
 end
--- a/api/libproject/devstart.rb
+++ b/api/libproject/devstart.rb
@@ -2031,7 +2031,7 @@ def connect_to_cloud_db(cmd_name, *args)
   op.parse.validate
   gcc.validate
 
-  if op.opts.db_user.nil?
+  if op.opts.db_user.nil? || op.opts.db_user.empty?
     op.opts.db_user = "dev-readonly"
   end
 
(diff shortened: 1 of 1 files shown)


api/libproject/devstart.rb
```diff
@@ -2022,20 +2022,42 @@ def connect_to_cloud_db(cmd_name, *args)
   common = Common.new
   op = WbOptionsParser.new(cmd_name, args)
   op.add_option(
-    "--root",
-    ->(opts, _) { opts.root = true },
-    "Connect as root")
+    "--db-user [user]",
+    ->(opts, v) { opts.db_user = v },
+    "Optional database user to connect as, defaults to 'dev-readonly'. " +
+    "To perform mutations use 'workbench'. Avoid using 'root' unless " +
```
jaycarlton, line 2028: For later, if we're thinking of a "break glass" mode in certain responses, maybe this is one of those. I wonder if there's a way to capture/log a justification when signing in as root.

api/libproject/devstart.rb
```diff
@@ -2022,20 +2022,42 @@ def connect_to_cloud_db(cmd_name, *args)
...
+    "absolutely necessary.")
   gcc = GcloudContextV2.new(op)
   op.parse.validate
   gcc.validate
+
+  if op.opts.db_user.nil?
```
jaycarlton, line 2034: nit: to catch blank strings too you can do `db_user.to_s.empty?`.

api/libproject/devstart.rb
```diff
@@ -2022,20 +2022,42 @@ def connect_to_cloud_db(cmd_name, *args)
...
+    op.opts.db_user = "dev-readonly"
+  end
+
   env = read_db_vars(gcc)
+  user_to_password = {
+    "dev-readonly" => env["DEV_READONLY_DB_PASSWORD"],
+    "workbench" => env["WORKBENCH_DB_PASSWORD"],
+    "root" => env["MYSQL_ROOT_PASSWORD"]
+  }
+  unless user_to_password.has_key? op.opts.db_user
```
jaycarlton, line 2044: nit: Rubocop's style guide prefers parenthesis when there are arguments.

api/libproject/devstart.rb
```diff
@@ -2022,20 +2022,42 @@ def connect_to_cloud_db(cmd_name, *args)
...
+    exit 1
+  end
+  db_password = user_to_password[op.opts.db_user]
+
   CloudSqlProxyContext.new(gcc.project).run do
-    password = op.opts.root ? env["MYSQL_ROOT_PASSWORD"] : env["WORKBENCH_DB_PASSWORD"]
-    user = op.opts.root ? "root" : env["WORKBENCH_DB_USER"]
+    if op.opts.db_user == "dev-readonly"
+      common.status ""
+      common.status "Database session will be read-only; use --db-user to change this"
```
jaycarlton, line 2054: does `mysql` not let them know that during startup?

api/libproject/devstart.rb
```diff
@@ -2022,20 +2022,42 @@ def connect_to_cloud_db(cmd_name, *args)
...
+    Common.new.error(
+      "invalid --db-user provided, wanted one of #{user_to_password.keys}, got '#{op.opts.db_user}'")
+    exit 1
+  end
+  db_password = user_to_password[op.opts.db_user]
+
   CloudSqlProxyContext.new(gcc.project).run do
-    password = op.opts.root ? env["MYSQL_ROOT_PASSWORD"] : env["WORKBENCH_DB_PASSWORD"]
-    user = op.opts.root ? "root" : env["WORKBENCH_DB_USER"]
+    if op.opts.db_user == "dev-readonly"
```
jaycarlton, line 2052: nit: if we add another readonly user, we'll need to adjust this. Maybe make an array of read only users and check inclusion in it. But again, I wonder if this blurb is redundant.

api/libproject/devstart.rb
calbach: You requested this..

> Can you put a prominent notice right above here that the session is or isn't read-only?

https://github.com/all-of-us/workbench/pull/3550#discussion_r421469642

The MySQL CLI is printed when you are launching the command, which includes the database user. However, this is somewhat buried in the startup logs and is not at all obvious. The CLI prompt does not give any indication of a readonly session, since there is nothing special about the session itself - this is just a property of the user.

api/libproject/devstart.rb
calbach: Did we decide to follow that style guide? This file uses them somewhat interchangeably, but I'll make the change since I tend to prefer parens

api/libproject/devstart.rb
calbach: done

PR #3570, opened 2020-05-12
[RW-4934][risk=no] Task to archive Stackdriver configuration as JSON or YAML

jaycarlton:
This task visits all the passed-in environments and lists assets of various types and saves them to JSON. Note that if there are no filters, it's about a 25 MB zip file.

Files are written to directories generated by their formal GCP identifiers, which conveniently include the project names. I use a sortable timestamp (sans slashes and colons) for the file names.

Run this with
```
./devops.rb --task=backup-config --envs-file=monitoring_env_targets.json --output-dir=~/scratch/config_dump
```

The input file should be [this file](https://github.com/all-of-us/workbench-devops/blob/master/ops/inputs/monitoring_env_targets.json) or a subset of it for our projects. It's a good idea to make a copy and drop `local` so it doesn't run all the `test` environment fetches twice.

Output archive is on [Google Drive](https://drive.google.com/file/d/1JAGISpR8jUFyGORunytP27-VdFmrlUrb/view?usp=sharing).

I just realized Stackdriver names dashboards using the project number and not the project ID. I'm thinking about changing things so that we have a directory for each environment's short name above `projects` so at least they stay together.

Screen Shot 2020-05-13 at 11 23 06 AM

Note that if you run it in IntelliJ you get nice hyperlinks to the files created.
Screen Shot 2020-05-13 at 12 16 24 PM

---
**PR checklist**


jaycarlton:
@petesantos I thought this might be a fun entry point into Ruby so I made you primary. Also curious to see if you can easily make it work by following the installation instructions.

jaycarlton:
I finally got around to verifying round trips, and it looks like I need to use the internal `to_json` instead of doing `JSON.pretty_generate()`. I'm saving snake_case instead of camelCase, and various extra field are getting saved that the API doesn't really want to see. This makes sense because I was hashing before writing.
[json_diffs.pdf](https://github.com/all-of-us/workbench/files/4623048/json_diffs.pdf)

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
```diff
@@ -1,65 +1,227 @@
...
 
-      @logger.info("Total counts for #{env.short_name}: #{counts}")
+      logger.info("Total counts for #{@current_env.short_name}: #{counts}")
+    end
+  end
+
+  DO_BACKUP_ALL = false # TODO(jaycarlton): make this an option
+
+  # For every environment in the envs_file, fetch all items of interest from the appropriate
+  # Stackdriver APIs. (Note that in future we may wish to broaden this to other )
```
calbach, line 74: other..

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
```diff
@@ -1,65 +1,227 @@
...
+        backup_all_metrics
+      else
+        backup_custom_metrics
+        backup_user_logs_based_metrics
+      end
+      backup_dashboards
+      backup_group_members
+      backup_monitored_resources
+      backup_notification_channels
+      backup_notification_channels
```
calbach, line 88: I assume adding this twice was not intentional

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
```diff
@@ -1,65 +1,227 @@
...
+  def backup_custom_metrics
+    custom_metrics = @metric_client.list_metric_descriptors(@project_path, {filter: CUSTOM_METRIC_FILTER})
+    backup_assets(custom_metrics)
+  end
+
+  def backup_dashboards
+    dashboards = @dashboard_client.list_dashboards(@current_env.formatted_project_number)
+    backup_assets(dashboards)
+  end
+  def backup_user_logs_based_metrics
```
calbach, line 108: nit: newline

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
```diff
@@ -1,65 +1,227 @@
...
+  end
+
+  def backup_notification_channels
+    channels = @notification_channel_client.list_notification_channels(@project_path)
+    backup_assets(channels)
+  end
+
+  # Backup a collection of Stackdriver assets
+  def backup_assets(assets)
+    assets.each(&method(:backup_asset))
```
calbach, line 138: Interesting, first time seeing this syntax

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
```diff
@@ -1,65 +1,227 @@
...
+  # TODO(jaycarlton): explicitly provide constructor parameters for the clients so refactoring
+  # becomes easier
+  def visit_envs
+    visitor.visit do |env|
+      @current_env = env
+
+      # Stackdriver clients. Note that somehow the initialization order matters here. I suspect it
+      # has to do with library-wide state getting pulled into the constrctors.
+      @metric_client = Google::Cloud::Monitoring::Metric.new
+      @project_path = Google::Cloud::Monitoring::V3::MetricServiceClient.project_path(@current_env.project_id)
```
calbach, line 201: project_path is initialized twice here

ops/ruby/devops-toolkit/README.md
```diff
@@ -70,6 +70,14 @@ bundle install
 ## Tasks
 A variety of tasks (sub-commands) are available via the required `-t` option. Here  are their
 brief descriptions:
+- `backup-config`: Generate JSON or YAML files for each Stackdriver asset (other GCP assets TBD),
```
calbach, line 73: This is separably from the implementation details of the tool, but where do you intend to archive these? It sounds like it's too large to want to put this into any of our git repos.

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
jaycarlton: good catch

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
jaycarlton: Yeah the IDE did that for me on a refactor.

ops/ruby/devops-toolkit/README.md
jaycarlton: So there was a discussion a while back where @gjuggler mentioned that have achieved "configuration as code" for Stackdriver, and I've been looking for a way to make that an actual reality. https://precisionmedicineinitiative.atlassian.net/browse/RW-4946 is the task for the periodic archival. The simplest approach is to just make a zip file and post it every day and delete the Nth oldest, so they just rotate.

If I can output straight to BigQuery that would be even better.

ops/ruby/devops-toolkit/tasks/monitoring_assets.rb
jaycarlton: 👍

jaycarlton:
> Separately, if we're archiving these configs, it follows naturally that we may want to clone these to a different environment, diff them against the live version, or restore to a prior configuration. At this point we have built Terraform V2 in Ruby. I'd suggest looking into Terraform before any more substantial investment here.

I built the tools to clone dashboards, metrics, and some other things across environments first. That's why this uses the Ruby GCP libraries: in theory I'd get an easier object model to work with on the client. The main reason for doing this now is that it will make setting up all the alerts in a consistent way easier and safer. 

As it turns out I'm a bit disappointed by the documentation and conventions in the libraries, but it's becoming really nice now.

I'm definitely really interested in Terraform, but I didn't want to get too excited about it before determining if it's ok with security. Lots of much simpler and common things seem to be uphill struggles, and I didn't think I had the bandwidth for another side quest, and I don't yet have a sense for the timeline to get that running.

ops/ruby/devops-toolkit/README.md
gjuggler: My understanding of config-as-code is that the git repo is the canonical archive of config, same as our code. Maybe I'm missing something here, though. Is there something different about our Stackdriver monitoring setup that makes it necessary to pull things back down from Google and archive them?

jaycarlton: You have to have a systematic way of keeping up with it. The first step is mirroring what Stackdriver has locally, then archiving that, then restoring it. To put it into git is doable, though you'd want to strip out non-custom things more aggressively.

I'm not sure what form the "config" was expected to take if not this JSON or XML version. Either we start with that or build it with a template (which I may do next).

If the goal is to make the stored config the source of truth, then there needs to be some kind of reloading or two-way synchronizing happening periodically.

The work on the fetching pattern should be useful on the uploading step too though.

jaycarlton: If we want to put it in git, I can make a place in workbench-devops and do another PR there. I'd change the name structure to remove the multiple-files-per-config-named-after-timestamp and let git do the hard work on diffing, etc. (Sorting the arbitrarily ordered keys when writing the files will help here).

Having it directly in git is clearly the ideal from a config-as-code standpoint.

Moving from here to Terraform templates would be another increment of work.

jaycarlton:
I've cut monitoring resource descriptors, which were a very large list, since those are constant (or at least maintained by Stackdriver), and not on our side to maintain or restore.

PR for the first configs is at https://github.com/all-of-us/workbench-devops/pull/29.

ops/ruby/devops-toolkit/README.md
calbach: The "code-as-config" model would typically treat the code as the source of truth, with a one way sync from code -> deployed system. There would be no two-way sync. This simplification also allows you to templatize/parameterize these configs so that you're not managing N fully attenuated configs for N environments. Some one-time bootstrapping of these configs is fine, but that's probably a oneoff manual process.

I have the same question as Greg, which I don't think was answered above, which is why it is necessary to sync from Stackdriver -> code config, or maintain a mechanism for performing any kind of two-way sync.

jaycarlton: People like to fiddle with graphs and dashboards, and adjust levels and queries interactively. It would be very painful to lock down the console and only allow text edits.

With this tool, we can periodically (or right after a fiddling session) pull the configs, write them on top of what's in github, and PR the differences.

It would never have occurred to me to do a one-way only sync with an interactive, graphical tool (with such a relatively low-level API). Yes, the config is the source of truth, but prior to its existence we need a systematic way to pull it and arrange it.

Of course going up is just in the inverse of coming down. You still need to organize the config in some manner, and having it in the same form that the API wants is a plus. But since the input of the up direction is a subset of the output from the down direction, it was natural to do the down direction first. Plus, it let me get a handle on how much stuff we were working with.

Finally, this tool should work for every GCP project that uses Stackdriver, not just ours, so a self-bootstrapping capability is a big plus.

calbach: Sorry, has the overall strategy you're describing already been discussed and agreed upon? Or are we having this discussion now from first principles? If this PR is aligned with what you and @gjuggler have already discussed, I won't block this. @gjuggler you can approve now if this is the case.

I don't think the code is the source of truth in the model you're describing. It's derived/pulled from the live config defined in the UI. If we're treating the UI as source of truth, I'd prefer not to check these configs into git as they must lag behind the UI by definition and are likely to go out of sync. If we want to drive all of the monitoring configs using a config-as-code approach, I think the way I'd expect this to work would be:
- prefer to isolate any "fiddling" to explicit personal/sandbox/experimental dashboards or to freeform query editors
- Use a query editor to build dashboards / alert policies
- When you're done, pull the JSON/yaml representation of this out, and convert it into a templated form as a PR, such that they are appropriately parameterized for all relevant environments
- Right before merging the PR, push the canonical configs to all environments
- You should expect any changes you make to the production configs outside of git (via the UI) may be blown away on the next monitoring config push

If we're using the UI as the source of truth, and periodically archiving our configs outside of source control, that sounds fine if that's the pragmatic way forward.

jaycarlton:
This is just one piece of functionality that can enable or support or serve
as a prototype for several strategies. I did this piece first so I could at
least have a backup in case of disaster in later stages.

PR #3572, opened 2020-05-12
[no ticket][risk=no] Puppeteer test code refactor

aweng98:
Description:
* Moving, renaming and grouping similar methods. Removing unused methods.
* Adding new class `Container.ts` for Dialog/Modal class files. Making it possible for finding elements inside the dialog/modal object. This is the main change that generated a cascade of changes in this PR. Because almost every type of element has to have a `Container` parameter in class constructors.

Test coverage remains essentially unchanged.
All e2e tests ran passed on localhost.
Screen Shot 2020-05-12 at 1 27 24 PM


e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
   }
 
-  protected readonly page: Page;
-  protected css: string;
-  protected xpath: string;
-  protected element: ElementHandle;
+  protected _element: ElementHandle;
+  protected readonly _xpath;
+  protected readonly _page: Page;
+  protected readonly _container: Container;
```
aweng98, line 19: Use a leading underscore `_` for class fields names. Main reason for change naming convention is because old class field `page` can be easily confused with jest-puppeteer global `page` property.

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,77 @@
...
+import {JSHandle, Page} from 'puppeteer';
 import Checkbox from 'app/element/checkbox';
 import RadioButton from 'app/element/radiobutton';
 import TextOptions from 'app/element/text-options';
 import Textarea from 'app/element/textarea';
 import Textbox from 'app/element/textbox';
+import GoogleLoginPage from 'app/page/google-login';
+import HomePage from 'app/page/home-page';
+
+export async function signIn(page: Page) {
```
aweng98, line 11: `signIn` method moved here from`app-utils.ts`

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,77 @@
...
+  const homePage = new HomePage(page);
+  await homePage.waitForLoad();
+}
+
+/**
+ * <pre>
+ * Wait until spinner stops spinning, usually it indicate page is ready.
+ * </pre>
+ */
+export async function waitUntilNoSpinner(page: Page): Promise<void> {
```
aweng98, line 25: `waitUntilNoSpinner` method moved here from`base-page.ts`. It feels right to me to have it inside `test-utils.ts`.

e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
 
-  constructor(aPage: Page, aElement?: ElementHandle) {
-    this.page = aPage;
-    this.element = aElement || undefined;
+  constructor(pageOptions: {puppeteerPage: Page, container?: Container},
+              selector: {xpath?: string, testId?: string} = {}) {
+    this._page = pageOptions.puppeteerPage;
+    this._container = pageOptions.container || undefined;
+    if (selector.testId !== undefined) {
+      this._xpath = `//*[@data-test-id="${selector.testId}"]`;
```
aweng98, line 26: added new code to support `data-test-id` usage in future.

e2e/puppeteer-custom-environment.ts
```diff
@@ -23,7 +23,7 @@ class PuppeteerCustomEnvironment extends PuppeteerEnvironment {
     case 'test_fn_failure':
       console.log('test failed: \t', event.test.name);
       if (state.currentlyRunningTest.invocations > retryTimes) {
-        const testName = state.currentlyRunningTest.name.replace(/\s/g, ''); // remove whitespaces
+        const testName = fp.startCase(state.currentlyRunningTest.name).replace(/(^\w|\s)/g, '');
```
aweng98, line 26: make screenshot filename clearer. e.g.
`Test "User can create a workspace using all inputs" failed. Saved screenshot logs/screenshot/serCanCreateAWorkspaceUsingAllInputs_1589299608904.png.`

e2e/app/element/base-element.ts
jaycarlton: The issue I see with this convention is that if you have two levels of parent classes, you'll need two leading underscores. But you don't know, when defining the parent class, at what level in the hierarchy it will wind up. It could be several.

@petesantos is this the usual convention? I see this answer as indicating underscores serve to make fields [effectively private](https://stackoverflow.com/questions/45167199/why-are-underscores-added-to-fields-of-a-typescript-class-in-angular4), which is a different consideration.

e2e/utils/test-utils.ts
jaycarlton: Utils classes in general tend to become like the junk drawer in your kitchen. It tends not to be a clear abstraction that's easy to maintain.

e2e/puppeteer-custom-environment.ts
jaycarlton: "screenshot/serCanCreateAWorkspaceUsingAllInputs" did you lose a U?

e2e/app/component/dialog.ts
```diff
@@ -41,17 +36,17 @@ export default class Dialog {
...
   private async findDialog() {
     if (this.dialogElement === undefined) {
-      this.dialogElement = await this.page.waitForSelector(SELECTOR.dialogRoot, {visible: true});
+      this.dialogElement = await this.page.waitForXPath(this.xpath, {visible: true});
     }
   }
 
   private getButtonSelector() {
-    return `${SELECTOR.dialogRoot} ${SELECTOR.button}`;
+    return `${this.xpath}//button | ${this.xpath}//*[@role="button"]`;
```
jaycarlton, line 49: There's some redundancy here because the `{this.xpath}` is repeated. I'd concatenate the string "${this.xpath}" with either side or else move the conditional into the descendant (after the xpath), if that's possible. There's probably also a way to join path segments richly.

e2e/app/component/ellipsis-menu.ts
```diff
@@ -5,7 +5,7 @@ export default class EllipsisMenu {
 
   readonly rootXpath = '//*[@id="popup-root"]';
 
-  constructor(private readonly page: (Page),
+  constructor(private readonly page: Page,
```
jaycarlton, line 8: What doe the parentheses do?

e2e/app/component/navigation.ts
```diff
@@ -1,6 +1,6 @@
 import {ElementHandle, Page} from 'puppeteer';
 import {clrIconXpath} from 'app/element/xpath-defaults';
-import {findIcon} from 'app/element/xpath-finder';
+import ClrIconLink from 'app/element/clr-icon-link';
```
jaycarlton, line 3: What does the CLR stand for? I don't think we necessarily have to reuse names that are confusing in the main app.

e2e/app/component/navigation.ts
```diff
@@ -84,15 +84,15 @@ export default class Navigation {
     const isOpen = await Navigation.sideNavIsOpen(page);
     if (!isOpen!) {
         // click bars icon to open dropdown
-      const barsIcon = await findIcon(page, {}, 'bars');
+      const barsIcon = await ClrIconLink.forLabel({puppeteerPage: page}, {}, 'bars');
```
jaycarlton, line 87: is it a Jest page or a Puppeteer page?

e2e/app/component/select-menu.ts
```diff
@@ -39,7 +45,7 @@ export default class SelectMenu {
       if (maxAttempts <= 0) {
         return null;
       }
-      return await this.page.waitFor(2000).then(clickAndCheck); // two seconds pause and retry
+      return await this._page.waitFor(2000).then(clickAndCheck); // two seconds pause and retry
```
jaycarlton, line 48: So would `this.page` ever have evaluated to the global `page` here? Or is it just a change in name to avoid confusing humans (which I support)?

e2e/app/component/select-menu.ts
```diff
@@ -98,17 +104,19 @@ export default class SelectMenu {
   }
 
   private dropdownXpath(): string {
-    if (this.label === undefined) {
-      return '//*[contains(concat(" ", normalize-space(@class), " "), " p-dropdown ")]';
+    if (this._label === undefined) {
```
jaycarlton, line 107: This looks really complex and prone to breakage by refactoring the dropdown component. I feel like there may be a way to simplify this by adding "handles" on the component (such as data tests IDs). Is that possible here?

The other problem is that the test class knows much more about the implementation than I'd be comfortable with if this is avoidable.

e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
+  protected readonly _page: Page;
+  protected readonly _container: Container;
 
-  constructor(aPage: Page, aElement?: ElementHandle) {
-    this.page = aPage;
-    this.element = aElement || undefined;
+  constructor(pageOptions: {puppeteerPage: Page, container?: Container},
+              selector: {xpath?: string, testId?: string} = {}) {
+    this._page = pageOptions.puppeteerPage;
+    this._container = pageOptions.container || undefined;
```
jaycarlton, line 24: If you have to check everything for undefined, then you're not in a very happy place. It's better to design invariants into the objects that ensure they are always in a well-defined state.

Also, I'm a bit worried that we have the same base class for a contained vs not-contained element. What about having a high-level base class that matches all elements, and an intermediate one derived from that for elements in containers? Then you don't need all these undefined members.

e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
+
+  async findElement(waitOptions: WaitForSelectorOptions = { visible: true }): Promise<ElementHandle | null> {
+    if (this._element !== undefined) {
+      return this._element;
+    }
+    if (this._xpath !== undefined) {
+      this._element = await this._page.waitForXPath(this._xpath, waitOptions);
     } else {
-      this.element = null;
+      this._element = null;
```
jaycarlton, line 55: So now we're introducing `null` in addition to the convention of using `undefined`. It's going to be easy to write bugs by confusing these, considering how many places are explicitly checking for `undefined`.

Also, please document the lazy evaluation here. I think it's safe, though it could affect timings potentially if used in a tight loop.

e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
+    }
+    return this._element;
+  }
+
+  async findElements(): Promise<ElementHandle[] | null> {
```
jaycarlton, line 60: Please document why I would use this instead of `findElement`. How would I know there's an array to be found? Please test that, for a single-element instance, the first (and only) returned element is the same as I'd get from `findElements()`.

e2e/app/element/base-element.ts
```diff
@@ -7,69 +8,69 @@ import {ClickOptions, ElementHandle, Page, WaitForSelectorOptions} from 'puppete
...
-    return this.element;
+    return null;
   }
 
-  async retryFindElement(): Promise<ElementHandle | null> {
-    if (this.xpath != null) {
-      return this.findByXpath(this.xpath);
-    } else if (this.css != null) {
-      return this.findByCss(this.css);
+  async findDescendantElements(descendantXpath: string, appendSelector: boolean = false): Promise<ElementHandle[] | null> {
```
jaycarlton, line 67: We're almost getting into territory of writing a prettier interface to xpath itself. Does such a library already exist?

We'll want lots of test cases to make sure the edge cases are covered for these types of methods.

e2e/app/element/base-element.ts
```diff
@@ -84,10 +85,10 @@ export default class BaseElement {
    *  return await handle.jsonValue();
    */
   async getProperty(propertyName: string): Promise<unknown> {
-    if (this.element == null) {
+    if (this._element == null) {
```
jaycarlton, line 88: Why not make this a `string?`-valued function instead of throwing? You can always throw later if the property name was critical.

e2e/app/element/button.ts
```diff
@@ -1,27 +1,20 @@
...
 import TextOptions from './text-options';
 import BaseElement from './base-element';
-import {findButton} from './xpath-finder';
+import * as xpathDefaults from './xpath-defaults';
 
 export default class Button extends BaseElement {
 
   static async forLabel(
-     page: Page,
+     pageOptions: {puppeteerPage: Page, container?: Container},
```
jaycarlton, line 10: The return value isn't called out in this signature.

e2e/app/element/checkbox.ts
```diff
@@ -1,27 +1,24 @@
...
-    let element: Checkbox;
-    try {
-      const checkboxElement = await findCheckbox(page, textOptions, waitOptions);
-      element = new Checkbox(page, checkboxElement);
-    } catch (e) {
-      if (throwErr) {
-        console.error(`FAILED finding Checkbox: "${JSON.stringify(textOptions)}".`);
-        throw e;
-      }
+    if (textOptions.ancestorNodeLevel === undefined) {
```
jaycarlton, line 14: Why not just do `textOptions.ancestorNodeLevel |= 1;`?

e2e/app/element/checkbox.ts
```diff
@@ -40,7 +37,7 @@ export default class Checkbox extends BaseElement {
     if (!is) {
       await this.focus();
       await this.clickWithEval();
-      await this.page.waitFor(500);
+      await this._page.waitFor(500);
```
jaycarlton, line 40: This 500 should be parameterized somewhere.

e2e/app/element/clr-icon-link.ts
```diff
@@ -1,28 +1,21 @@
 import {Page, WaitForSelectorOptions} from 'puppeteer';
 import BaseElement from './base-element';
+import Container from './container';
 import TextOptions from './text-options';
-import {findIcon} from './xpath-finder';
+import * as xpathDefaults from './xpath-defaults';
```
jaycarlton, line 5: I know in Java we don't like to import `*` from a package. Not sure what the rule here is.

e2e/app/element/container.ts
```diff
@@ -0,0 +1,19 @@
+export default class Container {
+
+  private readonly _xpath: string;
+
+  constructor(selector: {xpath?: string, testId?: string}) {
```
jaycarlton, line 5: Is it possible to have factory methods, one taking an xpath and another taking  testId? Or get the xPath given a testId and just continue on?

Having two constructor parameters that are mutually exclusive is a confusing pattern, especially when both of them look like strings.

If you had a helper `testIdToXPath(testId?: string)` you could call that before constructing one of these.

e2e/app/element/container.ts
```diff
@@ -0,0 +1,19 @@
+export default class Container {
```
jaycarlton, line 1: This is a really important class, so we will need some documentation I think.

e2e/app/element/container.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (selector.testId !== undefined) {
+      this._xpath = `//*[@data-test-id="${selector.testId}"]`;
+    } else if (selector.xpath !== undefined) {
+      this._xpath = selector.xpath;
+    } else {
+      throw new Error('Require selector xpath or testId.');
+    }
+  }
+
+  get xpath() {
```
jaycarlton, line 15: So `Container` does not extend `BaseElement`, yet defines its own `xpath` accessor. This means we have composition going on at the same time as inheritance, so the tree is also a bush kind of... it's a mixed metaphor, unless I'm misunderstanding.

I think if you're going to go this route, then you have to be really careful, probably write a document about the scheme, and make sure you teach folks how it's intended to be used. For example if a Container is an Element but not a Base Element, but BaseElement is the parent for fall the elements, then is there a contradiction.

e2e/app/element/label.ts
```diff
@@ -1,27 +1,20 @@
 import {Page, WaitForSelectorOptions} from 'puppeteer';
+import Container from './container';
 import TextOptions from './text-options';
 import BaseElement from './base-element';
-import {findLabel} from './xpath-finder';
+import * as xpathDefaults from './xpath-defaults';
 
 export default class Label extends BaseElement {
 
   static async forLabel(
```
jaycarlton, line 9: This confuses me. Seeing `Label.forLabel()` would be odd.

e2e/app/element/xpath-defaults.ts
```diff
@@ -27,9 +27,8 @@ export function labelXpath(opts: TextOptions) {
  * @param label
  */
 export function buttonXpath(opts: TextOptions) {
```
jaycarlton, line 29: I thought the linter was requiring return types. Anyway, it would be good to see those here.

e2e/app/page/authenticated-page.ts
```diff
@@ -21,27 +20,33 @@ export default abstract class AuthenticatedPage extends BasePage {
...
    */
   async waitForLoad(): Promise<this> {
     if (!await this.isLoaded()) {
-      await savePageToFile(this.page, 'PageIsNotLoaded');
-      await takeScreenshot(this.page, 'PageIsNotLoaded');
-      throw new Error('Page isLoaded() failed.');
+      await savePageToFile(this._page);
+      await takeScreenshot(this._page);
+      const title = await this._page.title();
+      throw new Error(`${title}: page not loaded correctly.`);
```
jaycarlton, line 48: Please give the path to the screenshot and html in the error message (on separate lines would help).

e2e/app/page/authenticated-page.ts
```diff
@@ -59,50 +64,10 @@ export default abstract class AuthenticatedPage extends BasePage {
    */
   async getUserName(): Promise<unknown> {
```
jaycarlton, line 65: nit: should be `getUsername()`

e2e/app/page/base-page.ts
```diff
@@ -39,187 +38,16 @@ export default abstract class BasePage {
...
+    return this._page.$eval(`${cssSelector}`, elem => elem.textContent.trim())
   }
 
   /**
    * Find texts in DOM.
    * @param txt
    */
   async findText(txt: string): Promise<boolean> {
-    const indx = await (await this.page.content()).indexOf(txt);
+    const indx = await (await this._page.content()).indexOf(txt);
```
jaycarlton, line 49: Please rename this to `containsText()` so it looks more like a predicate. You could also make a small `FindResults` class that has a boolean for found and an index for the position (if found). Or just stick with the integer return value and call it `findStartIndex(text: string): Promise<integer?>` (if that's allowed).

e2e/app/page/create-account-page.ts
```diff
@@ -185,12 +176,11 @@ export default class CreateAccountPage extends BasePage {
...
-    const demograpicsSurveyPageHeader = 'Optional Demographics Survey';
-    await this.waitForTextExists(demograpicsSurveyPageHeader);
+    await waitUntilPageContainsText(this._page, 'Optional Demographics Survey');
     // Find and check on all checkboxes with same label: Prefer not to answer
     const targetXpath = '//*[normalize-space(text())="Prefer not to answer"]/ancestor::node()[1]/input[@type="checkbox"]';
-    await this.page.waitForXPath(targetXpath, { visible: true });
-    const checkboxes = await this.page.$x(targetXpath);
+    await this._page.waitForXPath(targetXpath, { visible: true });
+    const checkboxes = await this._page.$x(targetXpath);
     for (const ck of checkboxes) {
```
jaycarlton, line 184: I was curious and looked up how you could do these asynchronously just for fun. Something like:

```
await Promise.all(checkboxes.map(ck => ck.click()));
```

e2e/app/page/featured-workspaces-page.ts
```diff
@@ -14,7 +16,10 @@ export default class FeaturedWorkspacesPage extends AuthenticatedPage {
 
   async isLoaded(): Promise<boolean> {
     try {
-      await this.waitForTextExists(PAGE.HEADER);
+      await Promise.all([
+        waitUntilPageContainsText(this._page, PAGE.HEADER),
+        waitUntilNoSpinner(this._page)
+      ]);
       return true;
     } catch (e) {
```
jaycarlton, line 24: Why are you swallowing this exception message? Maybe log it at least.

e2e/app/page/google-login.ts
```diff
@@ -50,24 +52,24 @@ export default class GoogleLoginPage extends BasePage {
...
     }
 
     const emailInput = await this.email();
     await emailInput.focus();
     await emailInput.type(userEmail);
 
-    const nextButton = await this.page.waitForXPath(SELECTOR.NextButton);
-    await this.clickAndWait(nextButton);
+    const nextButton = await this._page.waitForXPath(SELECTOR.NextButton);
+    await BaseElement.asBaseElement(this._page, nextButton).clickAndWait();
```
jaycarlton, line 72: are all `BaseElement`s clickable? If you expect a button there as part of the contract you're testing, why not create a `Button`?

e2e/app/page/google-login.ts
```diff
@@ -84,22 +86,16 @@ export default class GoogleLoginPage extends BasePage {
...
   async load(): Promise<void> {
     const url = config.uiBaseUrl + config.loginUrlPath;
-    try {
-      await this.page.goto(url, {waitUntil: ['networkidle0', 'domcontentloaded'], timeout: 0});
-    } catch (err) {
-      console.error('Google login page not found. ' + err);
-      await takeScreenshot(this.page, 'GoogleLoginPageNotFound');
-      throw err;
-    }
+    await this._page.goto(url, {waitUntil: ['networkidle0', 'domcontentloaded'], timeout: 120000});
```
jaycarlton, line 98: please define these two string constants with better names. I realize they're external.

That is a huge default timeout. I'd rather default to reasonable ones and only give more slack if we can't avoid it.

e2e/app/page/google-login.ts
```diff
@@ -118,32 +114,34 @@ export default class GoogleLoginPage extends BasePage {
...
       }
       await this.enterEmail(user);
+      await this._page.waitFor(1000); // reduce chances of getting Google login recaptcha
       await this.enterPassword(pwd);
       await this.submit();
     } catch (err) {
-      await takeScreenshot(this.page, 'FailedLoginPage');
-      await savePageToFile(this.page, 'FailedLoginPage');
-      throw err;
+      await takeScreenshot(this._page);
```
jaycarlton, line 127: This code is duplicated from what we saw before. Maybe make a common method or class you can use for cleaning up failed tests and reporting on them consistently.

e2e/app/page/google-login.ts
```diff
@@ -118,32 +114,34 @@ export default class GoogleLoginPage extends BasePage {
...
         throw err;
       });
-      await this.clickAndWait(googleLoginButton);
+      await BaseElement.asBaseElement(this._page, googleLoginButton).clickAndWait();
 
       if (!user || user.trim().length === 0) {
         console.warn('Login user email: value is empty!!!')
       }
       await this.enterEmail(user);
+      await this._page.waitFor(1000); // reduce chances of getting Google login recaptcha
```
jaycarlton, line 123: @gjuggler Have you been in touch with the recaptcha folks at all on what the best practices for e2e test suites is? This seems like Whack-a-Mole, though I don't know a better way to do it. (unless we send the captchas to a Slack channel or something).

e2e/app/page/google-login.ts
```diff
@@ -118,32 +114,34 @@ export default class GoogleLoginPage extends BasePage {
...
-        ]);
-      }
+    // Sometimes, user is prompted with "Enter Recovery Email" page. Handle the page if found.
+    await Promise.race([
+      waitUntilMatchPageTitle(this._page, 'Homepage', 30000),
+      this._page.waitForXPath('//input[@type="email" and @aria-label="Enter recovery email address"]', {visible: true, timeout: 30000}),
+    ]);
+    const elementHandles = await this._page.$x('//input[@type="email" and @aria-label="Enter recovery email address"]');
+    if (elementHandles.length > 0) {
+      await elementHandles[0].type(config.broadInstitutionEmail);
```
jaycarlton, line 139: nit: Nothing in the public repo (here) should be Broad- or Verily- or VUMC-specific. Those details can go in workbench-devops and be pulled in during build or launch.

e2e/app/page/google-login.ts
```diff
@@ -118,32 +114,34 @@ export default class GoogleLoginPage extends BasePage {
...
+      await Promise.all([
+        this._page.waitForNavigation(),
+        this._page.keyboard.press(String.fromCharCode(13)), // press Enter key
```
jaycarlton, line 142: Please define a constant for this and/or add a `hitEnter()` method to the base class.

e2e/app/page/home-page.ts
```diff
@@ -24,19 +26,17 @@ export default class HomePage extends AuthenticatedPage {
   async isLoaded(): Promise<boolean> {
     try {
       await Promise.all([
-        this.waitUntilTitleMatch(PAGE.TITLE),
-        this.waitForTextExists(PAGE.HEADER),
-        Link.forLabel(this.page, LABEL_ALIAS.SEE_ALL_WORKSPACES),
-        this.waitUntilNoSpinner(),
+        waitUntilMatchPageTitle(this._page, PAGE.TITLE, 60000),
+        waitUntilNoSpinner(this._page, 60000)
```
jaycarlton, line 30: We could use an invisible element or flag to indicate the loadedness of the page for these tests, though I guess you want to rely on user-visible behaviors.

e2e/driver/page-wait.ts
```diff
@@ -1,32 +1,23 @@
 import {Page} from 'puppeteer';
 
-
+// Experienmental method
 export async function waitUntilNetworkIdle(page: Page) {
```
jaycarlton, line 4: Please rename this to something like `waitUntilLoadedRace()`. Or something.

e2e/app/component/dialog.ts
```diff
@@ -1,36 +1,31 @@
...
+    console.log('dialog: \n' + modalText);
+    return modalText.toString();
   }
 
   async clickButton(buttonLabel: ButtonLabel): Promise<void> {
     const selector = this.getButtonSelector();
-    await this.page.waitForSelector(selector, {visible: true});
-    const buttons = await this.page.$$(selector);
+    await this.page.waitForXPath(selector, {visible: true});
+    const buttons = await this.page.$x(selector);
```
petesantos, line 28: Could the selector be modified to include the text content or aria-label of the button? And if so would that eliminate the for loop and reduce some complexity?
I am seeing a 'findButton' function in xpath-finder.ts and it looks like there are text options that can be passed in - would this work?

e2e/jest.test-setup.ts
```diff
@@ -6,8 +6,9 @@ const userAgent = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/5
 beforeEach(async () => {
   await page.setUserAgent(userAgent);
   // Refer to https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetdefaultnavigationtimeouttimeout
-  await page.setDefaultNavigationTimeout(90000);
-  await page.setDefaultTimeout(60000);
+  await page.setDefaultNavigationTimeout(70000);
```
jaycarlton, line 9: There are lots of places throughout where you have timeouts configured. Could they be consolidated here?

Additionally, if we could keep track of places where we "wasted" extra timeout time, then we might be able to find the long tent poles to performance. But if the worst offenders encourage us to increase all of them, then we might not have very tight bounds on things.

In other words, if we tracked how long every step of the e2e suite takes, we could potentially pinpoint regressions.

e2e/puppeteer-custom-environment.ts
```diff
@@ -22,23 +22,45 @@ class PuppeteerCustomEnvironment extends PuppeteerEnvironment {
...
+  async savePageToFile(fileName) {
+    const logDir = 'logs/html';
+    await fs.ensureDir(logDir);
+    const htmlFile = `${logDir}/${fileName}`;
+    const htmlContent = await this.global.page.content();
+    return new Promise((resolve, reject) => {
+      fs.writeFile(htmlFile, htmlContent, 'utf8', error => {
+        if (error) {
+          console.error(`Failed to save html file. ` + error);
+          reject(false);
```
jaycarlton, line 55: please document what the arguments to `reject` and `resolve` mean.

e2e/tests/home/home-page.spec.ts
```diff
@@ -19,7 +16,7 @@ describe('Home page ui tests', () => {
     let width;
     let height;
     for (const card of cards) {
-      const cardElem = new BaseElement(page, card.asElementHandle());
+      const cardElem = BaseElement.asBaseElement(page, card.asElementHandle());
```
jaycarlton, line 19: I don't understand this convention. Why not use a constructor?

e2e/tests/workspace/workspace-clone.spec.ts
```diff
@@ -84,12 +85,21 @@ describe('Clone workspace', () => {
       // strips out dash from workspace name
       expect(workspaceDataUrl).toContain(cloneWorkspaceName.replace(/-/g, ''));
 
+      // starting a new incognito page
```
jaycarlton, line 88: We probably want an `IncognitoBrowserWindow` class. Though the additional complexity here could be considerable.

e2e/tests/workspace/workspace-clone.spec.ts
```diff
@@ -84,12 +85,21 @@ describe('Clone workspace', () => {
...
+      await page.deleteCookie(...await page.cookies());
       await jestPuppeteer.resetBrowser();
-      const newPage = await browser.newPage();
+      await page.waitFor(2000);
+      const newBrowser = await browser.createIncognitoBrowserContext();
+      const newPage = await newBrowser.newPage();
+      const userAgent = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.138 Safari/537.36';
```
jaycarlton, line 94: please define this in one of the top-level config places.

e2e/app/component/dialog.ts
petesantos: There is a buttonXpath function in xpathDefaults, could that be used and/or modified to accept a container. That would help to ensure that we are always looking for buttons in the same way

e2e/utils/css-utils.ts
```diff
@@ -1,6 +1,10 @@
 import {Page} from 'puppeteer';
 
-export async function clickByXpath(page: Page, xpathSelector: string) {
+/**
+ * Summary: Helper functions when directly dealing with CSS selectors.
+ */
+
+export async function clickByEvalXpath(page: Page, xpathSelector: string) {
```
jaycarlton, line 7: Ideally we could have an abstraction where you just call `click()` and we pick the strategy behind the scenes.

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,98 @@
...
+  // wait maximum 90 seconds for spinner disappear if spinner existed
+  const spinAnimationSelector = 'svg[style*="spin"], .spinner:empty';
+  // const startTime = performance.now();
+  try {
+    if (jValue) {
+      await page.waitFor((selector) => {
+        return document.querySelectorAll(selector).length === 0;
+      }, {timeout: customTimeout}, spinAnimationSelector);
+    }
+  } catch (err) {
```
jaycarlton, line 48: Is this rethrowing necessary to get the test to report the failure in the right way?

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,98 @@
...
 import Textbox from 'app/element/textbox';
+import GoogleLoginPage from 'app/page/google-login';
+import HomePage, {LABEL_ALIAS} from 'app/page/home-page';
+import Link from '../app/element/link';
+import {waitForText} from './wait-utils';
+
+export async function signIn(page: Page) {
+  const loginPage = new GoogleLoginPage(page);
+  await loginPage.login();
+  // this element exists in DOM after user has logged in
```
jaycarlton, line 15: Is there something more user-facing, and less likely to change, that we could use to detect a successful login?

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,98 @@
...
+  }
+}
+
+export async function handleRecaptcha(page: Page) {
+  const css = '[id="recaptcha-anchor"][role="checkbox"]';
+  await page.frames().find(async (frame) => {
+    for(const childFrame of frame.childFrames()) {
+      const recaptcha = await childFrame.$$(css);
+      if (recaptcha.length > 0) {
+        await recaptcha[0].click();
```
jaycarlton, line 63: Is this trying to pass the Turing Test by clicking where it says "I am not a robot."? Maybe we should add wobbling to the mouse movement and click off center instead of right in the middle.

I don't suppose there's any way to whitelist our test accounts for captcha bypass.

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,98 @@
...
+    return;
+  }
+
+  // Click Self-Bypass button to continue
+  const selfBypass = await page.waitForXPath(`${selfBypassXpath}//div[@role="button"]`, {visible: true});
+  await selfBypass.click();
+  try {
+    await waitUntilNoSpinner(page);
+  } catch (timeouterr) {
+    // wait more if 60 seconds wait time wasn't enough.
```
jaycarlton, line 89: I feel like we should file a bug if it takes anywhere near this long.

e2e/utils/test-utils.ts
```diff
@@ -1,9 +1,98 @@
...
+      const recaptcha = await childFrame.$$(css);
+      if (recaptcha.length > 0) {
+        await recaptcha[0].click();
+        return;
+      }
+    }
+  });
+}
+
+export async function newUserRegistrationSelfBypass(page: Page) {
```
jaycarlton, line 70: The thing here is, we're writing code to test a non-supported workflow, but not actually testing the supported (and breakable) workflow our end users will see.

I think instead of a bypass button we should have trivial content for these users (including test users and internal devs) to consume/do in order to get access.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
+import {Page} from 'puppeteer';
+
+/**
+ * Summary: Helper functions when directly dealing with XPath selectors.
+ */
+
+export async function waitForProperty(page: Page, xpathSelector: string, propertyName: string, propertyValue: string, maxTimeOut: number = 30000) {
```
jaycarlton, line 7: nit: name `waitForPropertyEquality()`.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  return page.waitForFunction(xpath => {
+    const element = document.evaluate(
```
jaycarlton, line 9: Is there another way to call this method without so many parameters (like with an options struct)?

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * Wait until document title to match a regular expression.
+ * @param {string} title
+ */
+export async function waitUntilMatchPageTitle(page: Page, title: string, maxTimeOut: number = 30000) {
+  return page.waitForFunction(t => {
+    const actualTitle = document.title;
+    return actualTitle.includes(t);
```
jaycarlton, line 45: nit: inline `actualTitle` if you don't need to do anything else with it.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+    const href = window.location.href;
+    return href.includes(txt);
+  }, {timeout: maxTimeOut}, urlParitalString);
+}
+
+/**
+ * Wait until document title to match a regular expression.
+ * @param {string} title
+ */
+export async function waitUntilMatchPageTitle(page: Page, title: string, maxTimeOut: number = 30000) {
```
jaycarlton, line 42: rename `tite` to `titleRegex` please.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+    const actualTitle = document.title;
+    return actualTitle.includes(t);
+  }, {timeout: maxTimeOut}, title);
+}
+
+/**
+ * Look for visible texts on the page.
+ * @param {string} texts Partial texts to look for
+ */
+export async function waitUntilPageContainsText(page: Page, texts: string, maxTimeOut: number = 30000) {
```
jaycarlton, line 53: I really wish we had a way to say "file a bug or otherwise notify/aggregate with yellow flags after 8 seconds, but let the test continue, unless it takes more than 30 seconds".

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Wait for the element found from the selector has a particular attribute value pair
+ * @param {string} cssSelector - The selector for the element on the page
+ * @param {string} attribute - The attribute name
+ * @param {string} value - The attribute value to match
+ */
+export async function waitUntilContainsAttributeValue(page: Page, cssSelector, attribute, value, maxTimeOut: number = 30000) {
+  return page.waitForFunction((selector, attributeName, attributeValue) => {
+    const element = document.querySelector(selector);
+    return element.attributes[attributeName] && element.attributes[attributeName].value === attributeValue;
```
jaycarlton, line 73: could `element` be `undefined` or `null`?

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Wait for the elements count to be greater than a particular value
+ * @param {string} cssSelector - The css selector for the element
+ * @param {number} greaterThanCount - The expected number of elements
+ */
+export async function waitForNumberElementsIsBiggerThan(page: Page, cssSelector: string, greaterThanCount: number, maxTimeOut: number = 30000) {
+  return page.waitForFunction((selector, count) => {
+    return document.querySelectorAll(selector).length > count;
+  }, {timeout: maxTimeOut}, cssSelector, greaterThanCount);
```
jaycarlton, line 86: nit: I don't know what the style guide says, but I'd instinctively break after `},` almost always. @petesantos  any opinion here?

e2e/app/element/base-element.ts
petesantos: The _ prefix in JS was a convention that was meant to indicate a member was private.

I am a bit confused about the intention of this naming convention. Is it intended to clearly separate these variables from globals? Or to clearly identify where in the prototype chain these are coming from? If it is for clarity I would prefer a more explicit naming convention.

Looking at the code without the context of this PR I would not know what the _ means unless it was an accepted convention within the code base.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  }, {timeout: maxTimeOut}, cssSelector, attribute, value);
+}
+
+
+/**
+ * Wait for the elements count to be greater than a particular value
+ * @param {string} cssSelector - The css selector for the element
+ * @param {number} greaterThanCount - The expected number of elements
+ */
+export async function waitForNumberElementsIsBiggerThan(page: Page, cssSelector: string, greaterThanCount: number, maxTimeOut: number = 30000) {
```
jaycarlton, line 83: I'd probably make this `waitForMinElementCount()` and use an inclusive lower bound instead of the strict greater than. It seems awkward to subtract one from the minimum in my head.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  }, {timeout: maxTimeOut}, cssSelector, expectedCount);
+}
+
+/**
+ * Is there a element and visible located by CSS selector?
+ *
+ * @param page Puppeteer.Page
+ * @param selector CSS selector
+ */
+export async function visible(page: Page, cssSelector: string) {
```
jaycarlton, line 107: please indicate return types, here and everywhere.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  return page.evaluate(css => {
+    const element = document.querySelector(css);
+    return !!(element && (element.offsetWidth || element.offsetHeight || element.getClientRects().length))
```
jaycarlton, line 110: there's repeated code here from the wait until hidden function. Please factor that out so both call `isElementVisible()` in a synchronous way on an element.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  }, cssSelector)
+}
+
+/**
+ * Wait for exact text to match.
+ * @param page
+ * @param selector: {css, xpath}
+ * @param expectedText
+ */
+export async function waitForText(page: Page, selector: {xpath?: string, css?: string}, expectedText: string, maxTimeOut: number = 30000) {
```
jaycarlton, line 120: nit: one arg per line would read better. Might wait for an automatic linter though.

e2e/utils/wait-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+export async function visible(page: Page, cssSelector: string) {
+  return page.evaluate(css => {
+    const element = document.querySelector(css);
+    return !!(element && (element.offsetWidth || element.offsetHeight || element.getClientRects().length))
+  }, cssSelector)
+}
+
+/**
+ * Wait for exact text to match.
```
jaycarlton, line 115: if you're using `includes()`, then it's not an exact match IIUC. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

It looks like one branch uses `includes` and the other uses `===`. There might be a reason to do that, but it wasn't obvious.

e2e/app/page/authenticated-page.ts
aweng98: `savePageToFile` and `takeScreenshot` functions already log that.
e.g.
```
console.log utils/save-file-utils.ts:23
    Saved file logs/html/CreateWorkspace_2020-05-18T21:58:24.461Z.html

  console.log utils/save-file-utils.ts:41
    Saved screenshot logs/screenshot/CreateWorkspace_2020-05-18T21:58:24.593Z.png
```

e2e/.env.sample
```diff
@@ -1,6 +1,4 @@
-# Login user email
+# Login user email. Override by set process.env.USER_NAME property.
```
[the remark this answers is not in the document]
jaycarlton: The preferred name is Username, since users can't receive email at their researchallofus.org address.

e2e/app/component/select-menu.ts
jaycarlton: I'd prefer renaming the member in the foreign class on import and using the best (local) name in your own class.

e2e/app/component/select-menu.ts
jaycarlton: Have you tried adding the `data-test-id`s yourself? I think that's a reasonable, low-risk thing you could do. Just add it where it's feasible and file tickets for those components that aren't test-friendly. Also a best practice doc on the test IDs and another one on xpath might be helpful.

The goal is for anyone looking at a failed test to be able to understand the code at a glance.

e2e/app/element/base-element.ts
jaycarlton: I can buy that.

e2e/app/element/base-element.ts
jaycarlton: Please rename `findElement()` to `findFirstElement()` and `findElements()` to `findAllElements()` (so they differ by more than one character).

Yes, when the test harness gets this complex, it needs its own test suite. Plus, figuring out how to test it will show you ways to make it simpler and more flexible.

e2e/app/page/google-login.ts
jaycarlton: Is this because you can't wait for recaptcha itself to load? Anywhere I see a hardcoded wait, I get nervous.

e2e/app/page/google-login.ts
jaycarlton: Nothing in the system under test is Broad-specific, so the email name could just be `newUserContactEmail`.

You should keep email addresses and passwords in a config or environment file outside your test suite so folks outside AoU can use it too (in theory).

e2e/app/page/workspace-edit-page.ts
```diff
@@ -279,83 +280,88 @@ export default class WorkspaceEditPage extends AuthenticatedPage {
...
-      await this.getWorkspaceNameTextbox();
-      await new SelectMenu(this.page, LABEL_ALIAS.SELECT_BILLING).getSelectedValue();
-      await this.getCreateWorkspaceButton();
+      await Promise.all([
+        waitUntilMatchPageTitle(this._page, PAGE.TITLE),
+        this.getWorkspaceNameTextbox(),
+        new SelectMenu({puppeteerPage: this._page}, {label: LABEL_ALIAS.SELECT_BILLING}).getSelectedValue(),
+        this.getCreateWorkspaceButton(),
+        waitUntilNoSpinner(this._page)
+      ]);
```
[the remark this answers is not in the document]
jaycarlton: That works. You can make a reusable error messaging/logging library to for this so it's easy to use.

e2e/utils/wait-utils.ts
jaycarlton: Pretty sure it can go on the next line.

I think annotating the return types can also allow certain types of automatic analysis and make life easier for IDEs as well.

e2e/utils/wait-utils.ts
jaycarlton: Looks better, though I like `waitUntil` more than `waitFor` for readability.

e2e/README.md
```diff
@@ -7,27 +7,44 @@
 ##### Clone github project
 - `git clone git@github.com:all-of-us/workbench.git`
 
-##### Set up login credential
-- copy `.env.sample` file, save new file as `.env` (do not commit `.env` file to version control).
-- edit new `.env` file to provide workbench login credential.
-  ** note: we have a 2FA-bypassed test user created to run tests in test env. if you don't know it, ask others for test user credential.
-
-##### install node libraries
+##### Install node libraries
```
jaycarlton, line 10: Above, where you don't have a space after `####`, it's not rendering correctly.
Screen Shot 2020-05-19 at 12 14 37 PM

e2e/app/component/dialog.ts
```diff
@@ -1,36 +1,31 @@
...
   async getContent(): Promise<string> {
     await this.findDialog();
-    const modalText = await this.page.evaluate((selector) => {
-      const modalElement = document.querySelector(selector);
-      return modalElement.innerText;
-    }, SELECTOR.dialogRoot);
-    console.log('dialog: ' + modalText);
-    return modalText;
+    const modalText = await (await this.dialogElement.getProperty('innerText')).jsonValue();
+    console.log('dialog: \n' + modalText);
```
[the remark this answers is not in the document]
jaycarlton: I'd make this a `console.debug()` probably.

e2e/README.md
aweng98: fixed

e2e/app/component/dialog.ts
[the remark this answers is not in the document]
aweng98: sounds good.

e2e/utils/wait-utils.ts
aweng98: make sense. I've added return types for all methods in this class.

e2e/app/element/base-element.ts
aweng98: I like new names too. Done with rename.

e2e/app/component/dialog.ts
```diff
@@ -1,36 +1,32 @@
...
   async clickButton(buttonLabel: ButtonLabel): Promise<void> {
-    const selector = this.getButtonSelector();
-    await this.page.waitForSelector(selector, {visible: true});
-    const buttons = await this.page.$$(selector);
+    const selector = this.getButtonSelector(buttonLabel.toString());
+    await this.puppeteerPage.waitForXPath(selector, {visible: true});
+    const buttons = await this.puppeteerPage.$x(selector);
     for (const button of buttons) {
       const propValue = await button.getProperty('textContent');
       if (await propValue.jsonValue() === buttonLabel) {
```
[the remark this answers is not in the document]
aweng98: hmm, don't know why you are not seeing the code. This method has been refactored, no longer has a loop.

e2e/app/component/dialog.ts
```diff
@@ -1,36 +1,32 @@
...
 
-  constructor(private readonly page: Page) {
+  private dialogElement: ElementHandle;
 
+  constructor(private readonly puppeteerPage: Page, selector: {xpath?: string, testId?: string}) {
+    super(selector);
   }
 
   async getContent(): Promise<string> {
     await this.findDialog();
```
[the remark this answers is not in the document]
aweng98: Good points. Hopefully, there won't be need to add many new methods in future.
I could call `getDialog` in `constructor` but I really don't want to do that because I don't control when user start to create an instance of Dialog object. If instance is created before Dialog is visible, then it will error.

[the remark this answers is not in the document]
jaycarlton: +1: if you need to initialize something, then it shouldn't be named `findDialog()` maybe `initDialog()` if that's the case. But this is an anti-pattern anyway, and violates RAII.

If the pattern is that we need to do some kind of action to bind this class to a real, visible dialog, then that should be very explicit. But that's semantic coupling though.

PR #3579, opened 2020-05-12
[risk=moderate][RW-4972] Change checkEmail endpoint from GET to POST with email body

jmthibault79:
Description:

---
**PR checklist**

### DIFF
--- a/api/src/main/resources/workbench-api.yaml
+++ b/api/src/main/resources/workbench-api.yaml
@@ -2361,13 +2361,7 @@ paths:
         description: Institutional contact email address to check.
         required: true
         schema:
-          type: object
-          required:
-            - value
-          properties:
-            value:
-              type: string
-              description: Institutional contact email address to check.
+          "$ref": "#/definitions/CheckEmailRequest"
 
       responses:
         200:
@@ -5210,6 +5204,13 @@ definitions:
         default: false
     example:
       isTaken: false
+
+  CheckEmailRequest:
+    type: object
+    properties:
+      contactEmail:
+        description: The Institutional contact email address to check.
+        type: string
 
   CheckEmailResponse:
     type: object
--- a/api/src/main/resources/workbench-api.yaml
+++ b/api/src/main/resources/workbench-api.yaml
@@ -2361,7 +2361,14 @@ paths:
         description: Institutional contact email address to check.
         required: true
         schema:
-          type: string
+          type: object
+          required:
+            - value
+          properties:
+            value:
+              type: string
+              description: Institutional contact email address to check.
+
       responses:
         200:
           description: Whether the email is a valid member of the institution.
--- a/api/src/main/resources/workbench-api.yaml
+++ b/api/src/main/resources/workbench-api.yaml
@@ -2341,23 +2341,49 @@ paths:
           description: The updated Institution.
           schema:
             "$ref": "#/definitions/Institution"
-  "/v1/institutions/{shortName}/checkEmail/{email}/":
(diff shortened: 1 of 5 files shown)


api/src/main/resources/workbench-api.yaml
```diff
@@ -2341,25 +2341,58 @@ paths:
...
       description: Checks whether the given email address is a verified member of an institution.
       x-aou-note: This endpoint is publicly-accessible and called by clients during user registration.
       security: []
+      parameters:
+      - in: body
+        name: contactEmail
+        description: Institutional contact email address to check.
+        required: true
+        schema:
+          type: object
```
jmthibault79, line 2364: I would prefer to use `type: string` here but something gets lost in translation.  The Typescript wrapper adds double quotes to the string it passes to the endpoint (so, a valid JSON String) but the Java endpoint here treats the JSON String as a **Java** String without parsing the quotes away.

An alternative solution would be to manually strip the quotes.  But that feels too reliant on Swagger implementation so I'd be concerned that this might change with an upgrade.

api/src/main/resources/workbench-api.yaml
```diff
@@ -2341,25 +2341,58 @@ paths:
...
+            value:
+              type: string
+              description: Institutional contact email address to check.
+
+      responses:
+        200:
+          description: Whether the email is a valid member of the institution.
+          schema:
+            "$ref": "#/definitions/CheckEmailResponse"
+  "/v1/institutions/{shortName}/checkEmail/{email}/":
```
jmthibault79, line 2377: keeping the old endpoint so as not to immediately break anything reliant on it.  https://precisionmedicineinitiative.atlassian.net/browse/RW-4948 removes it.

ui/src/testing/stubs/institution-api-stub.ts
```diff
@@ -1,5 +1,6 @@
 import {
   CheckEmailResponse,
+  ContactEmail,
```
jmthibault79, line 3: Swagger generated this class for us without asking.  neat!

ui/src/app/pages/login/account-creation/account-creation-institution.tsx
```diff
@@ -170,8 +170,7 @@ export class AccountCreationInstitution extends React.Component<Props, State> {
     }
 
     try {
-      const result = await institutionApi().checkEmail(institutionShortName, contactEmail,
-        {signal: this.aborter.signal});
+      const result = await institutionApi().checkEmail(institutionShortName, {value: contactEmail}, {signal: this.aborter.signal});
```
jmthibault79, line 173: If we find a way to make `type: string` work in Swagger, we wouldn't need to change UI code at all.

api/src/main/resources/workbench-api.yaml
gjuggler: Yeah, I'm not too surprised there's weirdness in trying to have a bare string for a POST body. I think what you've done is the best approach, just go with the flow and make it a simple JSON object.

[opt] The vast majority of our `in: body` definitions have a named `schema: $ref`. For consistency & explicitness, I'd recommend creating a CheckEmailRequest object here rather than inlining the schema.

api/src/main/resources/workbench-api.yaml
gjuggler: Nice approach! I appreciate the attention to detail (and I'm sure Alex does too).

api/src/main/resources/workbench-api.yaml
jmthibault79: ok

PR #3584, opened 2020-05-14
[no ticket][risk=no] Implementing BQ integration tests for DataSet Builder

freemabd:
This PR adds BQ integration tests for DataSetController
1) Implemented new test class DataSetControllerBQTest
2) Removed test from DataSetControllerTest that mocked and verified sql only, since the test class above runs queries against BQ.


api/src/bigquerytest/java/org/pmiops/workbench/api/DataSetControllerBQTest.java
```diff
@@ -0,0 +1,504 @@
...
+                createDataSetRequest(
+                    Arrays.asList(dbConditionConceptSet, dbProcedureConceptSet),
+                    Arrays.asList(dbCohort1),
+                    Arrays.asList(Domain.CONDITION, Domain.PROCEDURE),
+                    false,
+                    PrePackagedConceptSetEnum.NONE))
+            .getBody()
+            .getCode();
+
+    String query = extractPythonQuery(code, 1);
```
NehaBroad, line 327: should we check for assert for Comments like in previous test

freemabd: Thanks for catching this. I extracted a private method that each test calls to capture the same asserts for all.

PR #3588, opened 2020-05-15
[risk=no] Delete the ephemeral ingest dataset on publish-cdr

calbach:
I previously thought this wasn't possible, but I was just misunderstanding the error message. `bq rm -rf` is needed to kill datasets which still contain child tables. Leave the TTL setting in case there's a partial cancel/failure.


calbach:
Verified against test env

PR #3590, opened 2020-05-18
[risk=low] Add Q4R4 to preprod CDR configs

calbach:
Already published per https://precisionmedicineinitiative.atlassian.net/browse/PD-5122

@freemabd to confirm the database index used here.

### DIFF
--- a/api/config/cdr_versions_preprod.json
+++ b/api/config/cdr_versions_preprod.json
@@ -10,5 +10,17 @@
     "releaseNumber": 3,
     "numParticipants": 224143,
     "cdrDbName": "r_2020q1_2"
+  },
+  {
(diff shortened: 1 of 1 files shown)


api/config/cdr_versions_preprod.json
```diff
@@ -10,5 +10,17 @@
     "releaseNumber": 3,
     "numParticipants": 224143,
     "cdrDbName": "r_2020q1_2"
+  },
+  {
+    "cdrVersionId": 2,
+    "isDefault": false,
```
freemabd, line 16: Should this be default?

calbach: Yeah, I think you're right. Done.

PR #3595, opened 2020-05-18
[RW-4919][risk=low] Fix yarn and gradle caches on deploy

calbach:
- Fix the gradle cache location. I had attempted to fix this a while back, but had to rollback: https://github.com/all-of-us/workbench/pull/3238 . This should allow for incremental progress and faster subsequent deploys (e.g. stable push may be slow, preprod/prod fast). Should mitigate issues from 504 gradle installs, as incremental progress is now cached.
- Add yarn cache for consecutive UI builds (for yarn installs)


calbach:
Note: I ran the deploy in dry run mode several times to confirm and observed the cache working. But yes - we'll need to see how it impacts the actual runtimes.

PR #3598, opened 2020-05-19
[RW-4976][risk=no] Running Puppeteer tests on local UI in CircleCI

aweng98:
**PR checklist**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,9 @@
 version: 2.1
 
+# https://circleci.com/orbs/registry/
+orbs:
+  browser-tools: circleci/browser-tools@1.0.0
+
 # -------------------------
 #   PARAMETERS
 # -------------------------
@@ -7,29 +11,48 @@ parameters:
   circleci_service_account:
     type: string
     default: "circle-deploy-account@all-of-us-workbench-test.iam.gserviceaccount.com"
+  workbench_image:
+    type: string
+    default: "allofustest/workbench:buildimage-0.0.17"
+  mysql_image:
+    type: string
+    default: "circleci/mysql:5.7"
 
 # -------------------------
 #   ANCHORS
 # -------------------------
 anchors:
+  java_env: &java_env
+    # As best I can tell (dmohs, 7 Feb '17), this is the only way to set a memory limit that Java
+    # processes executed within CircleCI's docker containers will respect. Very helpful resource:
+    # https://circleci.com/blog/how-to-handle-java-oom-errors/
+    #
+    # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
+    # so we bumped this down further to 2G.
+    JAVA_TOOL_OPTIONS: -Xmx2g
+    # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
+    GRADLE_OPTS: "-Dorg.gradle.daemon=false"
+    TERM: dumb
+    MYSQL_ROOT_PASSWORD: ubuntu
+
+  mysql_env: &mysql_env
+    MYSQL_ROOT_PASSWORD: ubuntu
+    MYSQL_USER: ubuntu
+    MYSQL_PASSWORD: ubuntu
+
+  e2e_env: &e2e_env
+    CI: true
+    NODE_ENV: development
+
   defaults: &defaults
     docker:
-      - image: allofustest/workbench:buildimage-0.0.17
+      - image: << pipeline.parameters.workbench_image >>
     working_directory: ~/workbench
 
   java_defaults: &java_defaults
     <<: *defaults
     environment:
-      # As best I can tell (dmohs, 7 Feb '17), this is the only way to set a memory limit that Java
-      # processes executed within CircleCI's docker containers will respect. Very helpful resource:
-      # https://circleci.com/blog/how-to-handle-java-oom-errors/
-      #
-      # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
-      # so we bumped this down further to 2G.
-      JAVA_TOOL_OPTIONS: -Xmx2g
-      # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
-      GRADLE_OPTS: "-Dorg.gradle.daemon=false"
-      TERM: dumb
+      <<: *java_env
 
   # See https://circleci.com/docs/2.0/workflows/#git-tag-job-execution
   # Job runs for master branch only
@@ -97,7 +120,7 @@ commands:
           name: Halt job and succeed early if no code changes on master branch
 
   ensure_branch_has_changes:
-    description: "Ensure branch has changes to code in specified directory"
+    description: Ensure branch has changes to code in << parameters.dir_name >> directory
     parameters:
       dir_name:
         type: enum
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -488,6 +533,71 @@ jobs:
...
+    <<: *defaults
+    environment:
+      <<: *e2e_env
+      WORKBENCH_ENV: local
+    resource_class: medium+
+    working_directory: ~/workbench
+    parallelism: 1
+    steps:
+      - run:
+          name: Check if it is a pull request
```
calbach, line 561: Better to just skip on master. This logic is not consistent with the other tests, and causes problems because it won't run again when a PR is opened.

Specifically consider the following PR flow:
- git push # this logic runs, skips since there's no PR yet
- create PR # nothing happens, Circle continues running the other steps as normal, puppeteer will show green since it skipped
- (may or may not happen, depending on PR) git push # follow-up commit, now it will run properly

.circleci/config.yml
```diff
@@ -196,13 +214,72 @@ commands:
...
+          working_directory: ~/workbench/api
+          command: ./project.rb run-local-migrations
+      - run:
+          name: Launch local API server
+          working_directory: ~/workbench/api
+          # tail -f is important here, it keeps this process running indefinitely
+          command: ./project.rb start-local-api && tail -f build/dev-appserver-out/dev_appserver.out
+          background: true
+      - run:
+          name: Wait for local API server to start
```
calbach, line 249: Note: one optimization you could make, if these things take a while, is to start the servers early, then put the dockerize wait as late as possible. i.e.

```
- start_api
- start_ui
- setup_for_puppeteer
- wait_api
- wait_ui
- puppeteer
```

There's no real need to block immediately

.circleci/config.yml
jaycarlton: FWIW there's an option we can turn on in Circle to only kick off workflows for changes on branches that are in pull requests if that's desirable.

.circleci/config.yml
aweng98: That's good strategy. But I think it's best for me that I don't make this change to keep focus only getting local ui + "test" api servers to work.
I will incorporate that in future PR.

.circleci/config.yml
```diff
@@ -488,6 +533,70 @@ jobs:
...
+    parallelism: 1
+    steps:
+      - run:
+          name: Skip job if this is not a pull request
+          command: |
+            echo "PR: $CI_PULL_REQUEST"
+            if [ ${CIRCLE_BRANCH} != "" ] && [ ${CIRCLE_BRANCH} == "master" ] ; then
+              echo "Not a PR, skipping Puppeteer tests"
+              circleci-agent step halt
+            fi
```
aweng98, line 567: Good thinking on using master branch for skip filtering. Updated command accordingly.

.circleci/config.yml
aweng98: any objection to turn on this flag?
Screen Shot 2020-05-19 at 3 15 03 PM
https://app.circleci.com/settings/project/github/all-of-us/workbench/advanced?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fall-of-us%2Fworkbench

calbach: Interestingly, appears to already be on. I don't know who turned that on and when. So sounds like it should work as you expect.

Still, we're doing the same semantic check on this line, so it's good to just check master branch here for consistency as you've done: https://github.com/all-of-us/workbench/blob/8fc54652be3b1b322f0ea3e316b8d4455dd9de1b/.circleci/config.yml#L109

aweng98: oh, I turned it on after seeing Jay's comment. Honestly, I always thought it was on before.
For job `puppeteer-e2e-local-ui`, the command to skip job if it is on master branch is:
```
- run:
          name: Skip job if this is not a pull request
          command: |
            echo "PR: $CI_PULL_REQUEST"
            if [ ${CIRCLE_BRANCH} != "" ] && [ ${CIRCLE_BRANCH} == "master" ] ; then
              echo "Not a PR, skipping Puppeteer tests"
              circleci-agent step halt
            fi
```
Is that correct?
Another question:  How can `${CIRCLE_BRANCH} == ""`? I haven't seen that property with a empty string before.

.circleci/config.yml
```diff
@@ -84,7 +101,7 @@ commands:
           command: git submodule update --init --recursive
 
   halt_job_noncode_changes:
-    description: "Halt job and succeed early if no code changes on master branch"
+    description: "Skip job if no code changes on master branch"
```
calbach, line 104: nit: I prefer the previous description

.circleci/config.yml
```diff
@@ -140,9 +157,9 @@ commands:
       - restore_cache:
           name: "restore e2e cache"
           keys:
-            - v1-e2e-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
-            - v1-e2e-cache-master
-            - v1-e2e-cache-
+            - v1-e2e-dependency-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
```
calbach, line 160: nit: this feels implicit to me - what else would we be caching? I don't think the change is needed

.circleci/config.yml
```diff
@@ -488,6 +533,70 @@ jobs:
...
+      - run:
+          name: Skip job if this is not a pull request
+          command: |
+            echo "PR: $CI_PULL_REQUEST"
+            if [ ${CIRCLE_BRANCH} != "" ] && [ ${CIRCLE_BRANCH} == "master" ] ; then
+              echo "Not a PR, skipping Puppeteer tests"
+              circleci-agent step halt
+            fi
+      - checkout_init_git
+      - halt_job_noncode_changes
```
calbach, line 569: this checks the api directory by default. That doesn't seem desirable here

.circleci/config.yml
```diff
@@ -488,6 +533,70 @@ jobs:
...
+      - browser-tools/install-browser-tools
+      - start_local_ui
+      - run_e2e_test
+
+  # NOT READY!!
```
calbach, line 574: Did you mean to leave this in? Or are you actively debugging this and hoping to merge it with this PR? I dislike dead code so I would probably just leave this in a branch for now if it's not connected to anything.

.circleci/config.yml
calbach: It can be empty when a tag is pushed, which happens during deploys.

.circleci/config.yml
aweng98: reverted.

.circleci/config.yml
aweng98: reverted. thought it would be good to match command name `install_e2e_dependencies`.

.circleci/config.yml
aweng98: It is removed now. So close to have it working. I am saving it in a local file.

.circleci/config.yml
aweng98: `halt_job_noncode_changes` has no parameter. It skips job if only `*.md` or `*.pdf` files changed. I think you mean `ensure_branch_has_changes` command. But it has no default value for `dir` parameter. This was the change requested from you in last PR.

.circleci/config.yml
```diff
@@ -84,20 +101,20 @@ commands:
...
-                echo "No code changes in non-master branch. halting job."
+                echo "No code changes on non-master branch. halting job."
                 circleci step halt
             fi
-          name: Halt job and succeed early if no code changes on master branch
+          name: Halt job and succeed early if no code changes on non-master branch
 
   ensure_branch_has_changes:
-    description: "Ensure branch has changes to code in specified directory"
+    description: Halt job and succeed early if no code changes in << parameters.dir_name >> directory on non-master branch
```
aweng98, line 117: to match `halt_job_noncode_changes` description.

.circleci/config.yml
```diff
@@ -84,20 +101,20 @@ commands:
           command: git submodule update --init --recursive
 
   halt_job_noncode_changes:
-    description: "Halt job and succeed early if no code changes on master branch"
+    description: "Halt job and succeed early if no code changes on non-master branch"
```
aweng98, line 104: I believe it should be worded as `non-master`.

.circleci/config.yml
calbach: Oops, yep - way off. Thanks.

PR #3601, opened 2020-05-19
[risk=low] Add or update equals() and hashCode() for Institution* DB models

jmthibault79:
Description:

I need `DbVerifiedInstitutionalAffiliation.equals()` for the tool I'm building for RW-4911, and perhaps these could be more generally useful.

Local testing:
* Hitting a variety of API endpoints continues to work
* I can create a user with a verified institution in the UI
* I receive the User Instructions email for my institution

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -172,6 +172,29 @@ public class DbInstitution {
     }
     DbInstitution that = (DbInstitution) o;
 
+    final boolean domainsEq = Objects.equal(emailDomains, that.emailDomains);
+    System.out.println("DBInst emailDomains eq=" + domainsEq);
+    if (!domainsEq) {
+      for (DbInstitutionEmailDomain domain : emailDomains) {
+        System.out.println("domain = " + domain);
+      }
+      for (DbInstitutionEmailDomain domain : that.emailDomains) {
+        System.out.println("that domain = " + domain);
+      }
+    }
+
+    final boolean addrsEq = Objects.equal(emailAddresses, that.emailAddresses);
+    System.out.println("DBInst emailAddresses eq=" + addrsEq);
+    if (!addrsEq) {
+      for (DbInstitutionEmailAddress addr : emailAddresses) {
+        System.out.println("addr = " + addr);
+      }
+
+      for (DbInstitutionEmailAddress addr : that.emailAddresses) {
+        System.out.println("that addr = " + addr);
+      }
+    }
+
     return Objects.equal(shortName, that.shortName)
         && Objects.equal(displayName, that.displayName)
         && Objects.equal(organizationTypeEnum, that.organizationTypeEnum)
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionUserInstructions.java
```diff
@@ -67,7 +67,7 @@ public boolean equals(Object o) {
 
     DbInstitutionUserInstructions that = (DbInstitutionUserInstructions) o;
 
-    return Objects.equals(institution.getInstitutionId(), that.institution.getInstitutionId())
+    return Objects.equals(institution, that.institution)
```
jmthibault79, line 70: This is now an equivalence comparison (recursing through the Institution model fields) rather than "is this exactly the same row in the DB"

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -67,12 +67,14 @@ public boolean equals(Object o) {
 
     DbInstitutionEmailAddress that = (DbInstitutionEmailAddress) o;
 
-    return Objects.equals(institution, that.institution)
-        && Objects.equals(emailAddress, that.emailAddress);
+    // don't refer to the institution itself - addresses can still be equal if unassigned or
```
calbach, line 70: Can it? This object basically tracks an association between an email and an institution. The previous approach seems like it may have been more correct. If the issue is that we don't want to depend on the database's assigned ID during comparison, I'd just make that be the rule instead

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -157,4 +158,36 @@ public DbInstitution setEmailAddresses(
 
     return this;
   }
+
+  // logical equality: data members without the ID field
```
calbach, line 162: Comment should explain why the ID field is omitted

api/src/main/java/org/pmiops/workbench/db/model/DbVerifiedInstitutionalAffiliation.java
```diff
@@ -91,14 +93,17 @@ public boolean equals(Object o) {
 
     DbVerifiedInstitutionalAffiliation that = (DbVerifiedInstitutionalAffiliation) o;
 
-    return Objects.equals(user, that.user)
+    // TODO: DbUser doesn't have a well-defined equals() so we use the username instead
+
+    return Objects.equals(user.getUsername(), that.user.getUsername())
```
calbach, line 98: seems `that.user` could be null here

jmthibault79:
I'm having surprising difficulty making this work.  I'm going to drop this for now in favor of another approach.

jmthibault79:
Returning to this...

I am seeing inequality for DbInstitutionEmailAddress and DbInstitutionEmailDomain that doesn't make any sense to me.  So far I am unable to trigger it via unit test - only against a live environment.  The problem can be seen (along with a lot of debug prints) via:

input: https://github.com/all-of-us/workbench-devops/blob/master/ops/inputs/affiliations/alpha_beta_batch_1.csv 

```
cp alpha_beta_batch_1.csv workbench/api
./project.rb populate-user-affiliations \
--import-filename alpha_beta_batch_1.csv \
--user-type=RESEARCHERS \
--project all-of-us-rw-prod
```

This is safe to run because it defaults to Dry Run mode.

jmthibault79:
Closing in favor of #3618 which is working much better.

PR #3604, opened 2020-05-19
[RW-4925][risk=low] Add session expired, sign in again pages; extract public layout

calbach:
- This adds two new static content pages, see screenshots below.
- Users are redirected here in two scenarios:
  1. An active session expires: redirected to `/session-expired`
  2. The user closes their tabs and comes back later, after the session expires: redirect to `/sign-in-again`
- This extracts a common pattern from `cookie-policy` and `user-disabled` for a public page layout. This resulted in standardization between these two pages (see screenshot).
- Note: there are some styling changes made to the user disabled page, including removal of the responsive background image. The background image does not seem necessary for a static error message page, so I prefer consistency here.
- This also makes a functional change across the 4 functional pages, which is to make the logo clickable.
- There are no mocks for this, so I will request out of band review from Lou/Karthik on this.

Screenshot 2020-05-20 at 2 56 46 PM
Screenshot 2020-05-20 at 2 56 38 PM
Screenshot 2020-05-19 at 10 19 28 PM


jaycarlton:
I would probably not say that we "cannot determine" if the session is still valid. Maybe just say the session was closed due to an idle timeout. Also, does it need to be its own page, or can it be an overlay on the sign-in page to save a navigation step?

calbach:
@jaycarlton your suggested alternate text is not really accurate. Please first review the context on the ticket and make any copy suggestions there.

RE "overlay": it doesn't need to be done in any particular way. Redirecting to a static page is a simple solution which does not require much design input. I would say that's the correct answer here given the priority / effort we're assigning to this issue. Having some alternate UX on the signin page is something we could do, but we'd need mocks and generalized pattern for that. It's also certainly more complicated to build and maintain.

calbach:
Note: actively revising some of this per discussion with Karthik and Lou

calbach:
Updated per Karthik/Lou

ui/src/app/pages/signed-in/component.ts
```diff
@@ -88,17 +89,20 @@ export class SignedInComponent implements OnInit, OnDestroy, AfterViewInit {
       });
     });
 
-    this.signInService.isSignedIn$.subscribe(signedIn => {
+    this.signOutNavigateSub = this.signInService.isSignedIn$.subscribe(signedIn => {
       if (!signedIn) {
         this.signOut();
       }
     });
+    this.subscriptions.push(this.signOutNavigateSub);
```
petesantos, line 97: What does the `subscriptions` array do in this component?

calbach: It's purely for component teardown purposes: https://github.com/all-of-us/workbench/blob/56a6c34fb66952234d4607d15773c404ca44e8f9/ui/src/app/pages/signed-in/component.ts#L215-L217

calbach:
Sorry, made one more refactor per Karthik's request to include the standard 'G' icon signin button everywhere. Please take another quick look when you have the chance @petesantos , thanks.

PR #3606, opened 2020-05-20
[risk=no][RW-4921] Audit Log Entry API

jaycarlton:
Add an API to query a workspace's history in the audit dataset. This is a low-level query, so all fields are returned typed and ordered like they are in the query (as opposed to combined into a higher-level semantic structure). The same query service will be used for the user admin endpoint.

I added some tests, as well as some utility code to deal with `FieldValueList` structures more easily. These are now usable via `Optional` chaining.

I'm using Joda time now as sot of a head start on RW-4682, generated by using `format: date-time` on a `string` field.

One interesting thing about this format is that swagger-codegen's REST/JSON translation actually writes an `int64` value of the microseconds since epoch, even though it generates `DateTime` and `Date` in Java and TypeScript, respectively.

To call the API, log in as a user with the right authority and do something like
```
curl -X GET 'http://localhost:8081/v1/admin/audit/workspace/?limit=99&workspaceNamespace=aou-rw-local1-89786e50'   \
--header "Authorization: Bearer `gcloud auth print-access-token`" \
| jq
```

An example output entry looks like this:
```json
  {
      "actionId": "b937413e-ff66-4e7b-a639-f7947730b2c0",
      "actionType": "CREATE",
      "agentId": 1,
      "agentType": "USER",
      "agentUsername": "ericsong@fake-research-aou.org",
      "eventTime": 1579293096978,
      "newValue": "\"1\"",
      "previousValue": null,
      "targetId": 3,
      "targetProperty": "etag",
      "targetType": "WORKSPACE"
    }
```

I've added a freeform map for useful things that, depending on the endpoint needing to make the query, might make sense. The two things I have now are the workspace's database ID (which is needed for further queries and interpreting the result set potentially) and the query itself (for use in BigQuery). 

I felt that including the query would help with debugging and teaching users to use the BQ Console, so that they have a launch point for further queries. There should be no security issue in doing this, as the dataset names are currently public in the config files and no one without the right access can run the query anyway.

---
**PR checklist**


api/config/config_local.json
```diff
@@ -105,7 +105,9 @@
     "enableEventDateModifier": false
   },
   "actionAudit": {
-    "logName": "workbench-action-audit-local"
+    "logName": "workbench-action-audit-local",
```
jaycarlton, line 108: There's some duplication here, but it's because I can't guarantee we won't need to rename a table separately from the dataset or log. For example, adding a column to the JSON in the log means we technically need a new BQ schema for the table.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,158 @@
...
+import org.pmiops.workbench.workspaces.WorkspaceService;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ActionAuditQueryServiceImpl implements ActionAuditQueryService {
+
+  private BigQueryService bigQueryService;
+  private Provider<WorkbenchConfig> workbenchConfigProvider;
+  private WorkspaceService workspaceService;
+  public static final String WORKSPACE_EVENTS_QUERY_STRING_FORMAT =
```
jaycarlton, line 31: I'm really close to shopping for a template engine for this. I'm already up to three ordered substitutions and I don't have any time constraints, ordering options, etc.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7280,3 +7305,93 @@ definitions:
...
+        description: >
+          A member of the org.pmiops.workbench.actionaudit.AgentType enum (not generated). Since
+          this is typed as a string in BigQuery (and not constrained to that type), we don't try
+          to narrow the type in the admin API.
+        type: string
+      eventTime:
+        description: >
+          Timestamp recorded at time of the event. Called timestamp in the BigQuery table.
+        type: string
+        format: date-time
```
jaycarlton, line 7375: This is the first attempt at the more generic date & time handling. Maps to Joda `DateTime` in Java and Date in TypeScript.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
gjuggler: Maybe look to https://cloud.google.com/bigquery/docs/parameterized-queries ? I think we use this on the CB side of the house, so there should be examples in our codebase ([like this](https://github.com/all-of-us/workbench/blob/49b293e10583d5bbb047daf87f70bc6d45aa24cf/api/src/main/java/org/pmiops/workbench/cdr/ConceptBigQueryService.java#L57-L58)).

jaycarlton:
@als364 FYI: this endpoint and a similar one are going to be exposed for use in admin pages and/or tools. It's not in the mock yet.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+          + "  jsonPayload.agent_email AS agent_username,\n"
+          + "  jsonPayload.action_id AS action_id,\n"
+          + "  jsonPayload.action_type AS action_type,\n"
+          + "  jsonPayload.target_type AS target_type,\n"
+          + "  CAST(jsonPayload.target_id AS INT64) AS target_id,\n"
+          + "  jsonPayload.target_property AS target_property,\n"
+          + "  jsonPayload.prev_value AS prev_value,\n"
+          + "  jsonPayload.new_value AS new_value\n"
+          + "FROM %s\n"
+          + "WHERE jsonPayload.target_id = %d AND\n"
```
freemabd, line 38: I would use a named parameter for any values injected from the client. We use the QueryParameterUtil  to help with this in CB land. Example: https://github.com/all-of-us/workbench/blob/c516872973fe331af315a0d4ef1f99da76347455/api/src/main/java/org/pmiops/workbench/cohortbuilder/SearchGroupItemQueryBuilder.java#L401-L401

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    final ImmutableMap<String, String> metadata =
+        ImmutableMap.of(
+            "workspaceDatabaseId", Long.toString(workspaceDatabaseId), "query", queryString);
+
+    return new AuditLogEntriesResponse().logEntries(logEntries).queryMetadata(metadata);
+  }
+
+  private AuditLogEntry fieldValueListToAditLogEntry(FieldValueList row) {
+    final AuditLogEntry entry = new AuditLogEntry();
+    FieldValues.getString(row, "action_id").ifPresent(entry::setActionId);
```
freemabd, line 85: Nice. I like this pattern.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.cloud.bigquery.FieldValue;
+import com.google.cloud.bigquery.FieldValue.Attribute;
+import com.google.cloud.bigquery.FieldValueList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.joda.time.DateTime;
+
+/** Utility class for working with FieldValueLists, FieldValues, and Fields */
+public final class FieldValues {
```
freemabd, line 13: Nice. BigQueryService is doing something similar to this class. I think we could move this functionality into the BigQueryService and replace it's methods that are very similar, like getString(List<FieldValue> row, int index), getBoolean(List<FieldValue> row, int index).. etc. 
https://github.com/all-of-us/workbench/blob/c516872973fe331af315a0d4ef1f99da76347455/api/src/main/java/org/pmiops/workbench/api/BigQueryService.java#L126-L126

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
```diff
@@ -0,0 +1,56 @@
...
+    return getTimestampMicroseconds(row, fieldName)
+        .map(micro -> micro / MICROSECONDS_IN_MILLISECOND)
+        .map(DateTime::new);
+  }
+
+  public static FieldValue toPrimitiveFieldValue(Object value) {
+    return FieldValue.of(Attribute.PRIMITIVE, value);
+  }
+
+  public static FieldValueList buildFieldValueList(FieldList schemaFieldList, List<Object> values) {
```
freemabd, line 51: Is this method useful outside of ActionAuditQueryServiceTest? Otherwise i would move this and the method above into the test itself.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
```diff
@@ -0,0 +1,56 @@
...
+    return FieldValues.getOptional(row, fieldName).map(FieldValue::getTimestampValue);
+  }
+
+  public static Optional<DateTime> getDateTime(FieldValueList row, String fieldName) {
+    return getTimestampMicroseconds(row, fieldName)
+        .map(micro -> micro / MICROSECONDS_IN_MILLISECOND)
+        .map(DateTime::new);
+  }
+
+  public static FieldValue toPrimitiveFieldValue(Object value) {
```
freemabd, line 47: Could be private since nothing is accessing this method outside of this class.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
```diff
@@ -0,0 +1,56 @@
...
+  public static Optional<FieldValue> getOptional(FieldValueList row, String fieldName) {
+    final FieldValue value = row.get(fieldName);
+    if (value.isNull()) {
+      return Optional.empty();
+    } else {
+      return Optional.of(value);
+    }
+  }
+
+  public static Optional<String> getString(FieldValueList row, String fieldName) {
```
freemabd, line 29: Also, since this reusable can we add methods for all the possible BQ sql types?

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+          + "  jsonPayload.action_type AS action_type,\n"
+          + "  jsonPayload.target_type AS target_type,\n"
+          + "  CAST(jsonPayload.target_id AS INT64) AS target_id,\n"
+          + "  jsonPayload.target_property AS target_property,\n"
+          + "  jsonPayload.prev_value AS prev_value,\n"
+          + "  jsonPayload.new_value AS new_value\n"
+          + "FROM %s\n"
+          + "WHERE jsonPayload.target_id = %d AND\n"
+          + "  jsonPayload.target_type = 'WORKSPACE'\n"
+          + "ORDER BY event_time, agent_id, action_id\n"
```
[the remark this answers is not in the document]
jaycarlton: Definitely. The plan is to take in severe optional quiet params. I thought this was a reasonable stopping point until we design the UI. Happy to add them now though.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+          + "LIMIT %d";
+
+  public ActionAuditQueryServiceImpl(
+      BigQueryService bigQueryService, Provider<WorkbenchConfig> workbenchConfigProvider) {
+    this.bigQueryService = bigQueryService;
+    this.workbenchConfigProvider = workbenchConfigProvider;
+  }
+
+  @Override
+  public AuditLogEntriesResponse queryEventsForWorkspace(long workspaceDatabaseId, long limit) {
```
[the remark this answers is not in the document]
jaycarlton: You mean at the service level? Sure. Maybe 1000 would work. Beyond that we'd need pagination, which is simple here since it only grows by appending.

Auditable events are relatively rare, and the ending should be restricted to admin users.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
jaycarlton: Yeah. I kinda like static helpers to live in a class named after the thing, and don't really expect static methods on services. But if you only need this when you already have BigQueryService, that makes sense too.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
jaycarlton: If you give a mouse a 🍪...

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
jaycarlton: I think it could be useful for testing. I'll annotate it visiblefortesting or make it private.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
jaycarlton: Yes. It's useful anytime you want to move a BQ result. I can make another utility class for these.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
jaycarlton: Brilliant! I was wondering what those were for. Also, now I can make an enough with my column names and friendly names and and setter methods on the model class like `AuditQueryParam.AGENT_ID.map(row, logentry)`. Similar to `WorkspaceTargetProperty`.

api/src/main/java/org/pmiops/workbench/utils/FieldValues.java
jaycarlton: I decided to stick with a utility class so that I can use the methods inside an enum class (which can't inject BigQueryService). There are no dependencies on BigQuery itself.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7280,3 +7329,95 @@ definitions:
...
+      now.
+    required:
+    - logEntries
+    - queryMetadata
+    properties:
+      logEntries:
+        type: array
+        items:
+          "$ref": "#/definitions/AuditLogEntry"
+      queryMetadata:
```
gjuggler, line 7348: I'm generally not a fan of unspecified objects / fields in APIs. There really isn't too much additional cost in specifying field names and types, and IMO there's a huge benefit in API specificity.

There's also no other use of "additionalProperties" in this file, which suggests an argument by consistency, that we have a strong pattern to fully specify our API response objects.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7280,3 +7329,95 @@ definitions:
         format: int64
       conceptName:
         type: string
+  AuditLogEntriesResponse:
+    type: object
+    description: >
+      Represents the results of a query into the audit dataset in BigQery without
+      filtering or transformation. This response type is a stopgap until we develop
```
gjuggler, line 7336: This comment might feel better placed in a design document or Jira ticket rather than API documentation.

I'm not sure I fully understand why this is considered a stopgap – can you give any examples of higher-level queries or semantic mapping that you would want to build? From my viewpoint, it covers just about everything I would have imagined we'd want for the admin pages.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7280,3 +7329,95 @@ definitions:
...
+      All fields are treated as optional here, as there are no non-null constraints in the BigQery
+      dataset.
+
+      It's possible that due to some error now or in the past that some columns might not match
+      their expected values, though we take pains to avoid that. Regardless, nothing in this type
+      definition is stricter than the BigQuery schema.
+    type: object
+    properties:
+      actionId:
+        description: Unique ID for this action. Used for grouping similar events under a
```
gjuggler, line 7367: Thanks for the detailed documentation on these fields! Super helpful.

api/src/main/resources/workbench-api.yaml
```diff
@@ -933,6 +933,55 @@ paths:
           description: A list of workspaces
           schema:
             "$ref": "#/definitions/WorkspaceListResponse"
+  "/v1/admin/audit/workspace":
```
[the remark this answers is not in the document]
jaycarlton: There could certainly be useful queries that involve both a user and a workspace (or various numbers of them). Though I don't know I've left room for that.

api/src/main/resources/workbench-api.yaml
```diff
@@ -933,6 +933,55 @@ paths:
...
+            is good enough for now.
+          in: query
+          type: integer
+          format: int64
+          required: false
+          default: 0
+        - name: beforeExclusive
+          description: >
+             Upper bound on timestamp of most recent result (exclusive). Specified as milliseconds since Epoch
+             (though declared as a string). Defaults to current timestamp.
```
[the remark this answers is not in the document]
jaycarlton: OK. I need to update this one. There's a Spring bug where it doesn't really handle string timestamps correctly, so I changed it back to a long but neglected to update the description.

api/src/main/resources/workbench-api.yaml
jaycarlton: I generally agree. There's a historical problem here, because the audit system uses the DB id and the API does not.

The reason to do it here is to allow reusing the response object with several different kinds of queries. For example, the user query endpoint might take in a user ID, but the output is in terms of the database identifier.

Now, in the special case of this query, the DB ID is easy to extract from the result set, since it only shows up in one column. In the case of hte User query, I'm probably going to include both rows where the user is the agent as well as the target, so it's not obvious which ID should be used.

The client will likely be interested in distinguishing which column is describing the user in question.

Now, why not just have a `databaseId` field for both the user and the workspace (and others)? That gets you a little way down the road, but then you can easily have queries with more than one input (in principle).

The best solution not involving an arbitrary map is separate response types for each query type that look very similar but all contain a list of rows. That's cheap enough I suppose.

The balancing factor was that this was all going to be temporary until we built more high-level "storytelling" APIs for different types of things for different use cases. But that colud be a while.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+import org.pmiops.workbench.model.AuditLogEntry;
+import org.pmiops.workbench.utils.FieldValues;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ActionAuditQueryServiceImpl implements ActionAuditQueryService {
+
+  private static final int MICROSECONDS_IN_MILLISECOND = 1000;
+
+  enum Parameters {
```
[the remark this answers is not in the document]
jaycarlton: Yes. Originally I was thinking we'd have all the column names as enum constants too and have some slick metadata to help. With ~5 parameters it's borderline helpful.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionAuditQueryServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    final TableResult tableResult = bigQueryService.executeQuery(queryJobConfiguration);
+
+    final List<AuditLogEntry> logEntries =
+        StreamSupport.stream(tableResult.iterateAll().spliterator(), false)
+            .map(this::fieldValueListToAditLogEntry)
+            .collect(ImmutableList.toImmutableList());
+
+    final ImmutableMap<String, String> metadata =
+        ImmutableMap.of(
+            "workspaceDatabaseId", Long.toString(workspaceDatabaseId),
```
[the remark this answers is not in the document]
jaycarlton: That's true, in that we could fill it in in the controller. But then I need one more type if I'm also going to return the query, or I need to mutate the returned value (which I guess is OK).

api/src/main/resources/workbench-api.yaml
gjuggler: I think I like the solution in your second to last paragraph best – separate out the user-and-workspace specific response types, but use a common AuditLogEntries data definition for each.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
gjuggler: Makes sense, but it sounds a bit like YAGNI, as I don't know of any concrete use cases that would lead us to include such a query in the API.

On the balance, I think aligning with the existing collection hierarchy will be clearer and more consistent with existing API paths.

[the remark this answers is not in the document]
jaycarlton: That's fine I guess. I didn't want to be married to the namespace as the identifier though, which was the other reason for preferring query params.

PR #3613, opened 2020-05-21
[RW-4863][risk=no] Filter CB prep_ tables on publish

calbach:
Tested manually in test environment. This is the new convention for any extra tables Cohort Builder adds to the CDR dataset.


calbach:
@gjuggler PTAL

PR #3615, opened 2020-05-22
[risk=no] Enable gradle output caching

calbach:
Output caching is apparently not enabled by default in Gradle. The only caching we currently do is for externally downloaded dependencies. For me, this shaved 10-30s on a cached dev-up.

Numbers to capture from the log before/after:
- Loading configs complete (52s)                                                                                                                                                                                     
- Total dev-env setup time: 67s 
- API server startup complete (39s)

### DIFF
--- a/api/gradle.properties
+++ b/api/gradle.properties
@@ -1,0 +1,2 @@
+org.gradle.caching=true
+org.gradle.console=plain


calbach:
I believe this is actually auto-enabled on the new version of gradle we upgraded to

PR #3617, opened 2020-05-24
[no ticket][risk=no] Puppeteer framework xpath refactor

aweng98:
Current test framework does not yet support finding elements in a Dialog or Modal window. This is a blocker issue for DataSet and Cohort test cases.

Following are framework changes that enable finding in an easy way.

Every custom element wrapper class (Button, Link,etc.) requires a Puppeteer Page object and a XPath expression. To ensure this "behavior",  I created new `Container.ts` class with a simple constructor:
```
constructor(protected readonly page: Page, protected xpath: string)
```
The `Container.ts` class is the base class where all custom element wrapper classes(Button, dialog, etc.) extends from. 
Example:
`class BaseElement extends Container`
`class Dialog extends Container`
`class CreateCriteriaModal extends Dialog`, etc.

Two additional must have changes to make easier to write XPath expression (not knowing AoU app UI and from scratch is not easy for newcomers).
* `text-options.ts` became `XPathOptions`. Rename file in order to make it clear that options are used to for construction of xpath expression.
```
export interface XPathOptions {
  type?: ElementType;
  name?: string;
  containsText?: string;
  normalizeSpace?: string;
  ancestorLevel?: number;
  iconShape?: string;
}

```
* New (factory method) `xPathOptionToXpath` function in `xpath-defaults.ts`. This new function constructs and returns a xpath expression from `XPathOptions` object. Returned xpath expression includes parent node path if `container: Container` function parameter is set.
  Example:
 `Dialog.ts` `waitForButton` function:
```
  async waitForButton(buttonLabel: ButtonLabel): Promise<ElementHandle> {
     const selector = xPathOptionToXpath({name: buttonLabel, type: ElementType.Button}, this);
     return this.page.waitForXPath(selector, {visible: true});
  }
```

Further, existing functions code must be updated in order for compile project without error.

Couple added benefits from refactoring:
* Cleaner `xpath-defaults.ts` class with reduced code complexity. One new `xPathOptionToXpath()` function replaced all following functions:
- `textXpathHelper()`
- `labelXpath()`
- `buttonXpath()`
- `inputXpath()`
- `imageXpath()`
- `clickableXpath()`
- `clrIconXpath()`

Unused `xpath-finders.ts` functions were also removed:
- `findClickable()`
- `findSelect()`
- `findCheckbox()`
- `findButton()`
- `findTextarea()`
- `findLabel()`
- `findTextbox()`
- `findRadiobutton()`
- `findIcon()`
- `findImage()`


e2e/app/component/create-criteria-modal.ts
```diff
@@ -0,0 +1,37 @@
...
+
+
+export default class CreateCriteriaModal extends Dialog {
+
+  constructor(page: Page, xpath: string = defaultXpath) {
+    super(page, xpath);
+  }
+
+  async waitForPhysicalMeasurementCriteriaLink(criteriaType: PhysicalMeasurementsCriteria): Promise<ClrIconLink> {
+    return ClrIconLink.forLabel(this.page, {type: ElementType.Icon, name: criteriaType, iconShape: 'slider'}, this);
```
aweng98, line 34: Created this class to demonstrate finding icon-link in modal window. similar to existing function that finds button in dialog.

e2e/app/component/dialog.ts
```diff
@@ -1,57 +1,66 @@
...
-    for (const button of buttons) {
-      const propValue = await button.getProperty('textContent');
-      if (await propValue.jsonValue() === buttonLabel) {
-        return await button.click();
-      }
-    }
-    throw new Error(`Failed to find button with label ${buttonLabel}`);
+    const button = await this.waitForButton(buttonLabel);
+    return button.click();
+  }
```
aweng98, line 37: extracted find button step out of `clickButton` function.

e2e/app/page/create-account-page.ts
```diff
@@ -37,8 +38,8 @@ export const EDUCATION_LEVEL_VALUE = {
 };
 
 export const LABEL_ALIAS = {
-  READ_UNDERSTAND_PRIVACY_STATEMENT: 'I have read, understand, and agree to the All of Us Program Privacy Statement.',
-  READ_UNDERSTAND_TERMS_OF_USE: 'I have read, understand, and agree to the Terms of Use described above.',
+  READ_PRIVACY_STATEMENT: 'I have read, understand, and agree to the All of Us Program Privacy Statement',
+  READ_TERMS_OF_USE: 'I have read, understand, and agree to the Terms of Use described above',
```
aweng98, line 42: lint check: Had to shorten the label alias in orde to keep length of line under 140 characters. Also for some other long alias.

e2e/utils/test-utils.ts
```diff
@@ -45,14 +45,13 @@ export async function waitWhileLoading(page: Page, timeOut: number = 60000): Pro
...
+      // 1 second to give page time finish rendering
+      await page.waitFor(1000);
     }
   } catch (err) {
     throw new Error(err);
   }
-  // final 1 second wait for page render to finish
-  if (jValue) {
-    await page.waitFor(1000);
-  }
```
aweng98, line 55: moved up to `if (jValue)` block.

e2e/app/component/select-menu.ts
```diff
@@ -1,12 +1,21 @@
...
+import Container from 'app/container';
 import BaseElement from 'app/element/base-element';
+import {xPathOptionToXpath} from 'app/element/xpath-defaults';
+import {ElementType, XPathOptions} from 'app/xpath-options';
+import {Page} from 'puppeteer';
 
-export default class SelectMenu {
+export default class SelectMenu extends Container {
+
+  static async forLabel(page: Page, xOpt: XPathOptions = {}, container?: Container): Promise<SelectMenu> {
```
als364, line 9: why is this called `forLabel`? Should it perhaps be called `waitForLabel`?

e2e/app/page/create-account-page.ts
als364: You could always break the string over two lines.

e2e/app/component/select-menu.ts
aweng98: For consistency. Element wrapper classes (Button, Link, etc.) have `forLabel` static function.

e2e/app/component/create-criteria-modal.ts
```diff
@@ -0,0 +1,37 @@
+import {Page} from 'puppeteer';
+import ClrIconLink from 'app/element/clr-icon-link';
+import {ElementType} from 'app/xpath-options';
+import Dialog from './dialog';
+
+const defaultXpath = '//*[contains(concat(" ", normalize-space(@class), " "), " crit-modal-container ")]';
+
+export enum PhysicalMeasurementsCriteria {
+  BMI = 'BMI',
```
jaycarlton, line 9: Please alphabetize these. Also, maybe document what this is for (and if it's supposed to include all the possibilities).

e2e/app/component/create-criteria-modal.ts
jaycarlton: It looks like this class isn't used at all yet? If not, can we not have an example that's also in use? Or maybe wait to merge it?

e2e/app/component/dialog.ts
jaycarlton: might want to do this in one line. Would that be `await this.waitForButton(buttonLabel).then(b => b.click())`?

e2e/app/component/dialog.ts
```diff
@@ -1,57 +1,66 @@
...
+
+  async waitForButton(buttonLabel: ButtonLabel): Promise<ElementHandle> {
+    const selector = xPathOptionToXpath({name: buttonLabel, type: ElementType.Button}, this);
```
jaycarlton, line 40: What about renaming `xPathOptionToXpath()` to `buildXPathSelector()`?

e2e/app/component/dialog.ts
```diff
@@ -1,57 +1,66 @@
...
+    await this.page.waitForXPath(this.xpath, {visible: false});
   }
 
-  private async findDialog() {
+  async findDialog(): Promise<ElementHandle> {
     if (this.dialogElement === undefined) {
-      this.dialogElement = await this.page.waitForSelector(SELECTOR.dialogRoot, {visible: true});
+      this.dialogElement = await this.page.waitForXPath(this.xpath, {visible: true});
     }
+    return this.dialogElement;
```
jaycarlton, line 52: style-wise, I wonder if we need to prefix everything with `this.`. For C++, Java, Ruby, etc, we tend not to (though other styles have emerged over the years, like `m_page` 🤢.

e2e/app/component/dialog.ts
```diff
@@ -1,57 +1,66 @@
...
 }
 
-export default class Dialog {
-  private dialogElement: ElementHandle;
+export default class Dialog extends Container {
 
-  constructor(private readonly page: Page) {
+  private dialogElement: ElementHandle;
 
+  constructor(page: Page, xpath: string = Selector.defaultDialog) {
```
jaycarlton, line 23: One thing I'm not 100% sure about is the protocol here. You instantiate the Dialog object before you've determined that a dialog is ready or exists at all. I kinda feel like the `Dialog` should be something you wait on, and then it exposes properties that dialogs have.

e2e/app/component/select-menu.ts
jaycarlton: It's confusing, because there's no label in the signature or in the body of the function.

Also, can you declare it in the base class and use `override`?

e2e/app/container.ts
```diff
@@ -0,0 +1,16 @@
+import {Page} from 'puppeteer';
+
+/**
+ * This is the super base class.
+ * Every element needs a Page object and a xpath for locating the element.
+ */
+export default class Container {
```
jaycarlton, line 7: I don't think `Container` is quite right. Aren't there leaf elements that are xpath-able that don't contain things?

e2e/app/element/base-element.ts
```diff
@@ -1,77 +1,53 @@
...
 /**
  * BaseElement represents a web element in the DOM.
  * It implements useful methods for querying and interacting with this element.
  */
-export default class BaseElement {
+export default class BaseElement extends Container {
 
-  static asBaseElement(page: Page, elem: ElementHandle): BaseElement {
-    return new BaseElement(page, elem);
+  static asBaseElement(page: Page, elementHandle: ElementHandle): BaseElement {
```
jaycarlton, line 10: Generally methods starting with `as` aren't static. Can this go in the Page class?

e2e/app/element/xpath-defaults.ts
```diff
@@ -1,77 +1,81 @@
...
-  const nodeLevel = `ancestor::node()[${opts.ancestorNodeLevel}]`;
-  if (opts.inputType !== undefined) {
-    return `${labelXpath(opts)}/${nodeLevel}${numSlashes}input[@type='${opts.inputType}']`;
+  // optional function parameters check
+  if (type === 'icon') {
+    if (iconShape === undefined) {
+      throw new Error(`Incorrect XPathOptions configuration for Icon: set "iconShape" value.`);
+    }
+  } else {
+    if (name === undefined && normalizeSpace === undefined && containsText === undefined) {
```
jaycarlton, line 18: Generally, you should not try to catch programming errors at runtime like this. It's OK to let low-level errors show up and fix those. There are just too many things that could be undefined to chase down.

e2e/app/page/create-account-page.ts
```diff
@@ -47,6 +48,34 @@ export const LABEL_ALIAS = {
   INSTITUTION_EMAIL: 'Your institutional email address',
 };
 
+export const FIELD = {
```
jaycarlton, line 51: More descriptive name would help here.

e2e/app/xpath-options.ts
```diff
@@ -0,0 +1,20 @@
...
+  type?: ElementType;
+  name?: string;
+  containsText?: string;
+  normalizeSpace?: string;
+  ancestorLevel?: number;
+  iconShape?: string;
+}
+
+export enum ElementType {
+  Button = 'button',
```
jaycarlton, line 11: enums are typically ALL_CAPS unless there's a typescript standard that says otherwise.

e2e/app/component/create-criteria-modal.ts
aweng98: no, not used by any class/function in this PR. To add a usage example, it would cause new files to be added in this PR, outside of scope. It is just for an example on how to find Icon-Link in a modal container.

e2e/app/component/dialog.ts
aweng98: Sorry, I'm confused by this comment. Can you elaborate? Class constructor does nothing else except instantiate `page` and `xpath` class fields. It is not yet finding dialog. No action is happening here.

e2e/app/component/dialog.ts
aweng98: yes, I can. I like the new name.

e2e/app/component/select-menu.ts
als364: My expectation is that a function name tells me what a function does. This doesn't have a verb in it. Perhaps this should be called `getSelectMenu`, and the function getting the wrapped element in radio buttons should be `getRadioButton`, and so on.

e2e/app/container.ts
als364: How about something like `ComponentWrapper`? That's what we're actually locating here

aweng98: Thank you. I will consider renaming in future refactor PRs.

e2e/app/element/base-element.ts
```diff
@@ -1,77 +1,53 @@
...
 
-  protected readonly page: Page;
-  protected css: string;
-  protected xpath: string;
   protected element: ElementHandle;
 
-  constructor(aPage: Page, aElement?: ElementHandle) {
-    this.page = aPage;
-    this.element = aElement || undefined;
+  constructor(protected readonly page: Page, xpath?: string) {
```
petesantos, line 18: I am curious what the use cases are for this / how you envision its usage. Also - Should the constructor ever be directly called?

aweng98:
@petesantos Thank you. I will address most of your feedback in new PR.

PR #3618, opened 2020-05-28
[risk=low] Refactor Institutional email matching model objects

jmthibault79:
Description:

The bidirectional mapping between DbInstitution and DbInstitutionEmailDomain/Address caused a lot of confusing and subtle problems.  Now the mapping is present only on the email side of the relationship (the Many side).  Also:
* new mappers and DAOs 
* add equals() and hashCode() to DB model classes

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
@@ -1,12 +1,13 @@
 package org.pmiops.workbench.db.dao;
 
 import java.util.Set;
+import org.pmiops.workbench.db.model.DbInstitution;
 import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
 import org.springframework.data.repository.CrudRepository;
 
 public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
 
-  Set<DbInstitutionEmailDomain> getByInstitutionId(final long institutionId);
+  Set<DbInstitutionEmailDomain> getByInstitution(final DbInstitution institution);
 
-  long deleteByInstitutionId(final long institutionId);
+  long deleteByInstitution(final DbInstitution institution);
 }
--- a/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
@@ -1,0 +1,12 @@
+package org.pmiops.workbench.db.dao;
+
+import java.util.Set;
+import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
+
+  Set<DbInstitutionEmailDomain> getByInstitutionId(final long institutionId);
+
+  long deleteByInstitutionId(final long institutionId);
+}
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -12,9 +12,7 @@ import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
-import javax.persistence.OneToMany;
 import javax.persistence.Table;
-import org.jetbrains.annotations.NotNull;
 import org.pmiops.workbench.model.DuaType;
 import org.pmiops.workbench.model.OrganizationType;
 
@@ -28,8 +26,6 @@ public class DbInstitution {
   private Short organizationTypeEnum;
   private String organizationTypeOtherText;
   private Short duaTypeEnum;
-  @NotNull private Set<DbInstitutionEmailDomain> emailDomains = Sets.newHashSet();
-  @NotNull private Set<DbInstitutionEmailAddress> emailAddresses = Sets.newHashSet();
 
   public DbInstitution() {}
 
@@ -92,69 +88,6 @@ public class DbInstitution {
 
   public DbInstitution setOrganizationTypeOtherText(String organizationTypeOtherText) {
     this.organizationTypeOtherText = organizationTypeOtherText;
-    return this;
-  }
-
-  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL)
-  @NotNull
-  public Set<DbInstitutionEmailDomain> getEmailDomains() {
-    return emailDomains;
-  }
-
-  /**
-   * Effectively: do an in-place this.emailDomains = emailDomains
-   *
-   * <p>Hibernate doesn't like it when you reassign collections. Instead, modify in-place. First,
-   * call retainAll() to subset DB rows to those we wish to keep: the intersection of old and new.
-   * Then, call addAll() to add the diff(new - old) rows.
-   *
-   * <p>https://stackoverflow.com/questions/5587482/hibernate-a-collection-with-cascade-all-delete-orphan-was-no-longer-referenc
-   *
-   * @param emailDomains the new collection of domains for this Institution
-   */
-  public DbInstitution setEmailDomains(final Collection<DbInstitutionEmailDomain> emailDomains) {
-    final Set<DbInstitutionEmailDomain> attachedDomains =
-        Optional.ofNullable(emailDomains)
-            .map(Collection::stream)
-            .orElse(Stream.empty())
-            .map(domain -> domain.setInstitution(this))
-            .collect(Collectors.toSet());
-    // modifies this set so that its value is the intersection of the two sets
-    this.emailDomains.retainAll(attachedDomains);
-    this.emailDomains.addAll(Sets.difference(attachedDomains, this.emailDomains));
-
-    return this;
-  }
-
-  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL)
-  @NotNull
-  public Set<DbInstitutionEmailAddress> getEmailAddresses() {
-    return emailAddresses;
-  }
-
-  /**
-   * Effectively: do an in-place this.emailAddresses = emailAddresses
-   *
-   * <p>Hibernate doesn't like it when you reassign collections. Instead, modify in-place. First,
-   * call retainAll() to subset DB rows to those we wish to keep: the intersection of old and new.
-   * Then, call addAll() to add the diff(new - old) rows.
-   *
-   * <p>https://stackoverflow.com/questions/5587482/hibernate-a-collection-with-cascade-all-delete-orphan-was-no-longer-referenc
-   *
-   * @param emailAddresses the new collection of addresses for this Institution
-   */
-  public DbInstitution setEmailAddresses(
-      final Collection<DbInstitutionEmailAddress> emailAddresses) {
-    final Set<DbInstitutionEmailAddress> attachedAddresses =
-        Optional.ofNullable(emailAddresses)
-            .map(Collection::stream)
-            .orElse(Stream.empty())
-            .map(address -> address.setInstitution(this))
-            .collect(Collectors.toSet());
-    // modifies this set so that its value is the intersection of the two sets
-    this.emailAddresses.retainAll(attachedAddresses);
-    this.emailAddresses.addAll(Sets.difference(attachedAddresses, this.emailAddresses));
-
     return this;
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -1,12 +1,5 @@
 package org.pmiops.workbench.db.model;
 
-import com.google.common.collect.Sets;
-import java.util.Collection;
-import java.util.Optional;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
@@ -1,5 +1,6 @@
 package org.pmiops.workbench.db.model;
 
+import com.google.common.base.Objects;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
@@ -82,5 +83,31 @@ public class DbInstitution {
   public DbInstitution setOrganizationTypeOtherText(String organizationTypeOtherText) {
     this.organizationTypeOtherText = organizationTypeOtherText;
     return this;
+  }
+
+  // omit ID field from equality so equivalent objects match regardless
+  // of whether they are actually present in the DB
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    DbInstitution that = (DbInstitution) o;
+
+    return Objects.equal(shortName, that.shortName)
+        && Objects.equal(displayName, that.displayName)
+        && Objects.equal(organizationTypeEnum, that.organizationTypeEnum)
+        && Objects.equal(organizationTypeOtherText, that.organizationTypeOtherText)
+        && Objects.equal(duaTypeEnum, that.duaTypeEnum);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(
+        shortName, displayName, organizationTypeEnum, organizationTypeOtherText, duaTypeEnum);
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
@@ -1,11 +1,14 @@
 package org.pmiops.workbench.db.model;
 
 import java.util.Objects;
+import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 
 @Entity
@@ -13,7 +16,7 @@ import javax.persistence.Table;
 public class DbInstitutionEmailAddress {
 
   private long institutionEmailAddressId;
-  private long institutionId;
+  private DbInstitution institution;
   private String emailAddress;
 
   public DbInstitutionEmailAddress() {}
@@ -30,13 +33,14 @@ public class DbInstitutionEmailAddress {
     return this;
   }
 
-  @Column(name = "institution_id", nullable = false)
-  public long getInstitutionId() {
-    return institutionId;
+  @ManyToOne(cascade = CascadeType.ALL)
+  @JoinColumn(name = "institution_id", nullable = false)
+  public DbInstitution getInstitution() {
+    return institution;
   }
 
-  public DbInstitutionEmailAddress setInstitutionId(long institutionId) {
-    this.institutionId = institutionId;
+  public DbInstitutionEmailAddress setInstitution(DbInstitution institution) {
+    this.institution = institution;
     return this;
   }
 
@@ -65,11 +69,11 @@ public class DbInstitutionEmailAddress {
     DbInstitutionEmailAddress that = (DbInstitutionEmailAddress) o;
 
     return Objects.equals(emailAddress, that.emailAddress)
-        && Objects.equals(institutionId, that.institutionId);
+        && Objects.equals(institution, that.institution);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(emailAddress, institutionId);
+    return Objects.hash(emailAddress, institution);
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
@@ -1,14 +1,11 @@
 package org.pmiops.workbench.db.model;
 
 import java.util.Objects;
-import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
 import javax.persistence.Table;
 
 @Entity
@@ -16,7 +13,7 @@ import javax.persistence.Table;
 public class DbInstitutionEmailAddress {
 
   private long institutionEmailAddressId;
-  private DbInstitution institution;
+  private long institutionId;
   private String emailAddress;
 
   public DbInstitutionEmailAddress() {}
@@ -33,14 +30,13 @@ public class DbInstitutionEmailAddress {
     return this;
   }
 
-  @ManyToOne(cascade = CascadeType.ALL)
-  @JoinColumn(name = "institution_id", nullable = false)
-  public DbInstitution getInstitution() {
-    return institution;
+  @Column(name = "institution_id", nullable = false)
+  public long getInstitutionId() {
+    return institutionId;
   }
 
-  public DbInstitutionEmailAddress setInstitution(DbInstitution institution) {
-    this.institution = institution;
+  public DbInstitutionEmailAddress setInstitutionId(long institutionId) {
+    this.institutionId = institutionId;
     return this;
   }
 
@@ -67,12 +63,12 @@ public class DbInstitutionEmailAddress {
 
     DbInstitutionEmailAddress that = (DbInstitutionEmailAddress) o;
 
-    return Objects.equals(institution, that.institution)
-        && Objects.equals(emailAddress, that.emailAddress);
+    return Objects.equals(emailAddress, that.emailAddress)
+        && Objects.equals(institutionId, that.institutionId);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(institution, emailAddress);
+    return Objects.hash(emailAddress, institutionId);
   }
 }
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
@@ -1,7 +1,6 @@
 package org.pmiops.workbench.db.model;
 
 import java.util.Objects;
-import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
@@ -33,7 +32,7 @@ public class DbInstitutionEmailAddress {
     return this;
   }
 
-  @ManyToOne(cascade = CascadeType.ALL)
+  @ManyToOne
   @JoinColumn(name = "institution_id", nullable = false)
   public DbInstitution getInstitution() {
     return institution;
--- a/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
+++ b/api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
@@ -50,7 +50,8 @@ public class DbInstitutionEmailAddress {
     return this;
   }
 
-  // logical equality: data members without the ID field
+  // omit ID field from equality so equivalent objects match regardless
+  // of whether they are actually present in the DB
 
   @Override
   public boolean equals(Object o) {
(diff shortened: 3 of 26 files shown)


api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -95,66 +85,29 @@ public DbInstitution setOrganizationTypeOtherText(String organizationTypeOtherTe
     return this;
   }
 
-  @OneToMany(mappedBy = "institution", cascade = CascadeType.ALL)
```
jmthibault79, line 98: Remove this error prone logic.  It's now distributed among mappers, InstitutionServiceImpl, and the DbInstitutionEmailDomain/DbInstitutionEmailAddress classes.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
```diff
@@ -95,66 +85,29 @@ public DbInstitution setOrganizationTypeOtherText(String organizationTypeOtherTe
...
-    final Set<DbInstitutionEmailDomain> attachedDomains =
-        Optional.ofNullable(emailDomains)
-            .map(Collection::stream)
-            .orElse(Stream.empty())
-            .map(domain -> domain.setInstitution(this))
-            .collect(Collectors.toSet());
-    // modifies this set so that its value is the intersection of the two sets
-    this.emailDomains.retainAll(attachedDomains);
-    this.emailDomains.addAll(Sets.difference(attachedDomains, this.emailDomains));
+  // omit ID field from equality so equivalent objects match regardless
```
jmthibault79, line 88: An earlier version of this was at #3601

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
```diff
@@ -67,12 +68,12 @@ public boolean equals(Object o) {
 
     DbInstitutionEmailAddress that = (DbInstitutionEmailAddress) o;
 
-    return Objects.equals(institution, that.institution)
-        && Objects.equals(emailAddress, that.emailAddress);
+    return Objects.equals(emailAddress, that.emailAddress)
```
jmthibault79, line 71: reordered so the more-likely-to-differ field comes first

api/src/main/java/org/pmiops/workbench/db/model/DbVerifiedInstitutionalAffiliation.java
```diff
@@ -91,15 +95,19 @@ public boolean equals(Object o) {
 
     DbVerifiedInstitutionalAffiliation that = (DbVerifiedInstitutionalAffiliation) o;
 
-    return Objects.equals(user, that.user)
+    // TODO: DbUser doesn't have a well-defined equals() so we use the username instead
```
jmthibault79, line 98: open to suggestions for a better pattern here (and below)

api/src/main/java/org/pmiops/workbench/institution/InstitutionMapper.java
```diff
@@ -1,64 +1,32 @@
...
   DbInstitution modelToDb(Institution modelObject);
 
-  // userInstructions will be populated by setUserInstruction afterMapping
+  // these fields will be populated by @AfterMapping populateFromAuxTables
+  @Mapping(target = "emailDomains", ignore = true)
+  @Mapping(target = "emailAddresses", ignore = true)
   @Mapping(target = "userInstructions", ignore = true)
   Institution dbToModel(DbInstitution dbObject, @Context InstitutionService institutionService);
 
-  default List<String> toModelDomains(Set<DbInstitutionEmailDomain> dbDomains) {
```
jmthibault79, line 30: moved this logic to the email domain / address mappers

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -29,27 +32,40 @@
 
   private static final Logger log = Logger.getLogger(InstitutionServiceImpl.class.getName());
 
+  private final String OPERATIONAL_USER_INSTITUTION_SHORT_NAME = "AouOps";
```
jmthibault79, line 35: moved from slightly below here

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -70,26 +86,26 @@
 
   @Override
   public Optional<Institution> getInstitution(final String shortName) {
-    return getDbInstitution(shortName)
```
jmthibault79, line 73: replaced this method with its contents

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionDaoTest.java
```diff
@@ -25,183 +21,54 @@
...
   public void test_findOneByDisplayName() {
-    assertThat(institutionDao.findOneByDisplayName("The Broad Institute"))
-        .hasValue(institutionWithoutEmailPatterns);
-    assertThat(institutionDao.findOneByDisplayName("The National Institutes of Health"))
-        .hasValue(institutionWithEmailPatterns);
-    assertThat(institutionDao.findOneByDisplayName("Verily, LLC")).isEmpty();
-  }
-
-  @Test
-  public void test_updateAllNewEmailAddresses() {
```
jmthibault79, line 129: these are no longer relvant

jmthibault79:
Technically I "ran local e2e tests" but they were mostly timeout failures so I'm not sure how useful that was.

jmthibault79:
In addition to general reliability, this should help with:
* https://precisionmedicineinitiative.atlassian.net/browse/RW-4994
* https://precisionmedicineinitiative.atlassian.net/browse/RW-4258

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.Set;
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailDomain;
+import org.springframework.data.repository.CrudRepository;
+
+public interface InstitutionEmailDomainDao extends CrudRepository<DbInstitutionEmailDomain, Long> {
+
+  Set<DbInstitutionEmailDomain> getByInstitution(final DbInstitution institution);
+
+  long deleteByInstitution(final DbInstitution institution);
```
jaycarlton, line 12: How does the return value work? Do you get 0 if nothing is deleted or the ID if something was?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jaycarlton: Are all of our Db objects using the same semantics for `equals()`?

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jaycarlton: I usually try to use `expected == actual`. But I'm not that consistent (unless `expected` is a constant`).

api/src/main/java/org/pmiops/workbench/db/model/DbVerifiedInstitutionalAffiliation.java
jaycarlton: Why not add one in `DbUser`? If `equals` is taken, you could use `isEffectivelyEqualTo()` or something.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import org.pmiops.workbench.db.model.DbInstitution;
+import org.pmiops.workbench.db.model.DbInstitutionEmailAddress;
+import org.pmiops.workbench.model.Institution;
+import org.pmiops.workbench.utils.mappers.MapStructConfig;
+
+@Mapper(config = MapStructConfig.class)
+public interface InstitutionEmailAddressMapper {
+  default Set<DbInstitutionEmailAddress> modelToDb(
+      final Institution modelInstitution, final DbInstitution dbInstitution) {
+    return Optional.ofNullable(modelInstitution.getEmailAddresses()).orElse(Collections.emptyList())
```
jaycarlton, line 18: I'd try to guard against the list being nullable.

We noticed in [#workbench-techdesign](https://pmi-engteam.slack.com/archives/GP4CG4TFF/p1590099355134900) that some of our generated models are getting their lists initialized to new ArrayLists when they're declared, but others (sadly including yours) aren't. It would be great to get away from the need to do this check everywhere for a null collection.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
```diff
@@ -0,0 +1,35 @@
...
+        .distinct()
+        .map(
+            address ->
+                new DbInstitutionEmailAddress()
+                    .setEmailAddress(address)
+                    .setInstitution(dbInstitution))
+        .collect(Collectors.toSet());
+  }
+
+  default List<String> dbToModel(Set<DbInstitutionEmailAddress> dbAddresses) {
```
jaycarlton, line 29: You should be able to just define a mapping from `DbInstitutionEmailAddress` to `String` and let MapStruct generate this for you. It's pretty smart with collection mappings.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
```diff
@@ -0,0 +1,33 @@
...
+import org.pmiops.workbench.model.Institution;
+import org.pmiops.workbench.utils.mappers.MapStructConfig;
+
+@Mapper(config = MapStructConfig.class)
+public interface InstitutionEmailDomainMapper {
+  default Set<DbInstitutionEmailDomain> modelToDb(
+      final Institution modelInstitution, final DbInstitution dbInstitution) {
+    return Optional.ofNullable(modelInstitution.getEmailDomains()).orElse(Collections.emptyList())
+        .stream()
+        .distinct()
```
jaycarlton, line 20: For mappers that write to a DB object, maybe document whether or not the caller can immediately call `save()` afterwards, or if anything changes in the database as a side effect. Currently we're not doing that anywhere, though.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
```diff
@@ -0,0 +1,33 @@
...
+        .map(
+            domain ->
+                new DbInstitutionEmailDomain().setEmailDomain(domain).setInstitution(dbInstitution))
+        .collect(Collectors.toSet());
+  }
+
+  default List<String> dbToModel(final Set<DbInstitutionEmailDomain> dbDomains) {
```
jaycarlton, line 27: As above, you should just be able to clue in MapStruct now to convert the `DbInstitutionEmailDomain` to a `String` and it will write your collection mapping for you.

please name this something like `dbDomainToDomainNameList()`.

Also, why a list and not a set?

api/src/main/java/org/pmiops/workbench/institution/InstitutionMapper.java
```diff
@@ -23,42 +14,17 @@
...
-    return Optional.ofNullable(modelAddresses).orElse(Collections.emptySet()).stream()
-        .map(address -> new DbInstitutionEmailAddress().setEmailAddress(address))
-        .collect(Collectors.toSet());
-  }
-
   @AfterMapping
-  default void setUserInstruction(
+  default void populateFromAuxTables(
       @MappingTarget Institution target, @Context InstitutionService institutionService) {
+    target.setEmailDomains(institutionService.getEmailDomains(target.getShortName()));
```
jaycarlton, line 26: I believe you can do this in your main mapping with attributes like `@Mapping(source = "institutionService.emailDomains", target = "emailDomains"`, unless I've missed something.

api/src/main/java/org/pmiops/workbench/institution/InstitutionService.java
```diff
@@ -52,6 +52,25 @@ boolean validateAffiliation(
...
   boolean validateInstitutionalEmail(Institution institution, String contactEmail);
 
+  /**
+   * Retrieve an ordered list of the email domains which this institution uses to match user contact
+   * email, or an empty list if none. Throws NotFoundException if the Institution does not exist.
+   *
+   * @param institutionShortName the short name (key) used to refer to this institution in the API
+   * @return the list of email domains associated with this institution, if any
+   */
+  List<String> getEmailDomains(String institutionShortName);
```
jaycarlton, line 62: So my thing is, why not just fetch an institution and then ask it for its email domains and addresses? I would expect the institution object to know (or expose how to get) this, but seeing it also on the service makes me wonder what the source of truth is.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: For later consideration: After experimenting in SQL  yesterday, I think we need an environment-agnostic way to refer to this special institution, or to find it. (Maybe a method we can call to get its ID for queries).

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -70,26 +86,26 @@
...
             () ->
                 new NotFoundException(String.format("Could not find Institution '%s'", shortName)));
   }
 
-  private Optional<DbInstitution> getDbInstitution(final String shortName) {
-    return institutionDao.findOneByShortName(shortName);
-  }
-
   @Override
   public Institution createInstitution(final Institution institutionToCreate) {
```
jaycarlton, line 104: Why are you declaring method parameters final? I'm a pretty big finalphile, but I don't generally do that.

In the case of calling `save()`, you should technically be nullling out its argument or replacing it so you don't accidentally reuse it. Though in practice it's less dangerous than the doc makes it sound.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -113,15 +129,19 @@ public void deleteInstitution(final String shortName) {
...
               // create new DB object, but mark it with the original's ID to indicate that this is
               // an update
+
               final DbInstitution dbObjectToUpdate =
-                  institutionMapper.modelToDb(institutionToUpdate).setInstitutionId(dbId);
-              return institutionMapper.dbToModel(institutionDao.save(dbObjectToUpdate), this);
+                  institutionDao.save(
+                      institutionMapper.modelToDb(institutionToUpdate).setInstitutionId(dbId));
+              populateAuxTables(institutionToUpdate, dbObjectToUpdate);
+              return institutionMapper.dbToModel(dbObjectToUpdate, this);
```
jaycarlton, line 144: Please make a private wrapper in the service like `toModel(dbObjectToUpdate)` so readers don't see this scary passing of `this` as a second argument.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -217,4 +249,34 @@ public boolean validateOperationalUser(DbInstitution institution) {
     return institution != null
         && institution.getShortName().equals(OPERATIONAL_USER_INSTITUTION_SHORT_NAME);
   }
+
+  private void populateAuxTables(
```
jaycarlton, line 253: nit: "AuxTables" isn't really an abstraction.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -217,4 +249,34 @@ public boolean validateOperationalUser(DbInstitution institution) {
...
+      setInstitutionUserInstructions(
+          new InstitutionUserInstructions()
+              .institutionShortName(modelInstitution.getShortName())
+              .instructions(userInstructions));
+    }
+  }
+
+  private void setInstitutionEmailDomains(
+      final Institution modelInstitution, final DbInstitution dbInstitution) {
+    institutionEmailDomainDao.deleteByInstitution(dbInstitution);
```
jaycarlton, line 269: Please document that this is replacing rather than appending or merging to the domains. Here and below.

api/tools/src/main/java/org/pmiops/workbench/tools/institutions/LoadInstitutions.java
```diff
@@ -28,6 +30,8 @@
   InstitutionMapperImpl.class,
   InstitutionUserInstructionsMapperImpl.class,
   PublicInstitutionDetailsMapperImpl.class,
+  InstitutionEmailDomainMapperImpl.class,
```
jaycarlton, line 33: FWIW, I'd be fine lumping all the institution mapping functions into one Mapper. It would simplify dependencies, and there's hardly any additional cost to pulling functions you don't need.

api/src/test/java/org/pmiops/workbench/institution/InstitutionUserInstructionsMapperTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Test(expected = BadRequestException.class)
+  public void test_modelToDb_emptyInstructions() {
+    final DbInstitution broad =
+        new DbInstitution().setShortName("Broad").setDisplayName("The Broad Institute");
+    when(service.getDbInstitutionOrThrow(broad.getShortName())).thenReturn(broad);
+
+    final InstitutionUserInstructions userInstructions =
+        new InstitutionUserInstructions()
+            .instructions("")
```
jaycarlton, line 81: 🔮 I see this potentially blowing up into a much larger set of metadata on the institution that will need to be managed and tended with care.

als364:
jay got to this faster than me so I don't have much in the way of comments.

jaycarlton:
_jay got to this faster than me so I don't have much in the way of comments._

Sorry about that. I didn't check how new it was.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitution.java
jmthibault79: I did not audit other DB objects.  In the case of DbInstitution and friends, it is necessary for things like determining Set membership and comparisons between already-persisted and newly-created objects.

api/src/main/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDao.java
jmthibault79: a count of how many were deleted

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
jmthibault79: Do you have examples for the better models ?

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
jmthibault79: In this case I want the list to be sorted so I'd have to add a step for that anyway

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
jmthibault79: The model has a (sorted) list

api/src/main/java/org/pmiops/workbench/institution/InstitutionMapper.java
jmthibault79: How would that work?  You'd have to specify the source parameter to institutionService.getEmailDomains() as well.  I have not seen examples like this.

api/src/main/java/org/pmiops/workbench/institution/InstitutionService.java
jmthibault79: The mapper itself needs this in order to create the Institution

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: I don't have a specific reason to make them final.  I also don't have a specific reason not to - so my default is to make all things final unless I have a good reason.

The argument to save here is never stored in a variable so there's nothing to null out

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: what is scary about it?

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
jmthibault79: got this to work returning a SortedSet.  Thanks.

api/src/main/java/org/pmiops/workbench/db/model/DbVerifiedInstitutionalAffiliation.java
jmthibault79: done

jmthibault79: nope - my attempt at full DbUser.equals() resulted in
```
Caused by: org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: org.pmiops.workbench.db.model.DbUser.authorities, could not initialize proxy - no Session
```
So I chose the easier username-only path

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jmthibault79: I'm confused here.  This is not a test so we don't have the concept of expected.  This also needs to use equals() instead of ==

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
jmthibault79: update: I was able to do this.  thanks.

api/src/main/java/org/pmiops/workbench/db/model/DbInstitutionEmailAddress.java
jaycarlton: yeah never mind. it's an old convention from a different language anyway.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapper.java
jaycarlton: This showed up on a Friday afternoon and the best we could say about it was that it was "odd". I can take a look.

I still need to get back on OpenApi3 now that we're launched.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
```diff
@@ -0,0 +1,35 @@
...
+    nullValueMappingStrategy = NullValueMappingStrategy.RETURN_DEFAULT)
+public interface InstitutionEmailDomainMapper {
+  default Set<DbInstitutionEmailDomain> modelToDb(
+      final Institution modelInstitution, final DbInstitution dbInstitution) {
+    return Optional.ofNullable(modelInstitution.getEmailDomains()).orElse(Collections.emptyList())
+        .stream()
+        .distinct()
+        .map(
+            domain ->
+                new DbInstitutionEmailDomain().setEmailDomain(domain).setInstitution(dbInstitution))
```
jaycarlton, line 26: If you define a mapping from domain and dbInst to DbInstitutionEmailDomain, MapStruct should be able to write this code I think.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
```diff
@@ -0,0 +1,35 @@
...
+        .collect(Collectors.toSet());
+  }
+
+  SortedSet<String> dbDomainsToStrings(final Set<DbInstitutionEmailDomain> dbDomains);
+
+  default String dbDomainToString(final DbInstitutionEmailDomain dbDomain) {
```
jaycarlton, line 32: nit: I wonder if there's a way to do this inline in a mapping annotation. No need to chase that rabbit if you have other prey.

api/src/main/java/org/pmiops/workbench/institution/InstitutionMapper.java
jaycarlton: Oh, I see.

By the way, I _think_ you can put `InstitutionService.class` in your `uses` list so you don't need to use `@Context`.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: I heard there's a plan for a flag somewhere, so that's fine.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: Oh, I misread it. The thing being saved is anonymous, so you're good.

My default is to make everything final that could possibly be final, with the exception of method parameters, and that's only for readability.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jaycarlton: I'm scarred by [COM programming](https://www.cs.umd.edu/~pugh/com/). I distrust any fancy footwork with `this`.

Basically this class already has a reference to the institution mapper, so it can pass whatever is needed into that method. I don't expect methods that I call to need a reference to me in general.

api/src/main/java/org/pmiops/workbench/institution/InstitutionUserInstructionsMapper.java
```diff
@@ -1,37 +1,36 @@
...
 
 @Mapper(config = MapStructConfig.class)
 public interface InstitutionUserInstructionsMapper {
-  @Mapping(target = "institutionUserInstructionsId", ignore = true)
-  @Mapping(target = "institution", ignore = true) // set by setFields()
-  @Mapping(target = "userInstructions", ignore = true) // set by setFields()
-  DbInstitutionUserInstructions modelToDb(
-      InstitutionUserInstructions modelObject, @Context InstitutionService institutionService);
+  default DbInstitutionUserInstructions modelToDb(
+      InstitutionUserInstructions modelObject, InstitutionService institutionService) {
```
jaycarlton, line 16: style-wise: I would just pass the short name in as the second parameter instead of the whole service. It's a "law of demeter" thing: You hand the ATM your debit card, not your jacket with the pocket with the wallet with the card in it.  As a bonus, this would allow other services to use the mapper more easily.

Also, do you not need `@Context` here?

api/src/test/java/org/pmiops/workbench/db/dao/InstitutionEmailDomainDaoTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Test
+  public void test_getByInstitution_empty() {
+    assertThat(institutionEmailDomainDao.getByInstitution(testInst)).isEmpty();
+    assertThat(institutionEmailDomainDao.count()).isEqualTo(0L);
+  }
+
+  @Test
+  public void test_getByInstitution_multiple() {
+    final DbInstitutionEmailDomain one =
```
jaycarlton, line 41: nit: the spelled out numbers as names confuse me. They make me use the wrong side of my brain.

api/src/test/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapperTest.java
```diff
@@ -0,0 +1,117 @@
...
+            .shortName("Broad")
+            .displayName("The Broad Institute")
+            .emailAddresses(rawAddresses);
+
+    // does not need to match the modelInst; it is simply attached to the DbInstitutionEmailAddress
+    final DbInstitution dbInst = new DbInstitution();
+
+    final Set<DbInstitutionEmailAddress> dbAddresses = mapper.modelToDb(modelInst, dbInst);
+
+    assertThat(dbAddresses.size()).isEqualTo(sortedDistinctAddresses.size());
```
jaycarlton, line 48: nit: use `hasSize()` on `dbAddresses`

api/src/test/java/org/pmiops/workbench/institution/InstitutionEmailAddressMapperTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    for (final DbInstitutionEmailAddress dbAddress : dbAddresses) {
```
jaycarlton, line 50: nit: you can do this without a loop if you like by asserting with contains all and doing the stream operation for all match.

api/src/main/java/org/pmiops/workbench/institution/InstitutionEmailDomainMapper.java
jmthibault79: updated

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
jmthibault79: easy enough, updated

jmthibault79:
Puppeteer error was a timeout: https://app.circleci.com/pipelines/github/all-of-us/workbench/1316/workflows/4da861dd-901c-4c78-aeae-66c454528096/jobs/84021/tests

PR #3619, opened 2020-05-28
[RW-4788][risk=no] Remove Zendesk integration on review request

calbach:
Also drops nightly tests, since this was the only test in that category.

Tested locally.


calbach:
Snyk error appears to be unrelated to the changes in this PR.

calbach:
Nevermind, Snyk was right - it was just a very convoluted implicit dependency I managed to break. The Zendesk module was transitively pulling in a newer Jackson, so I bumped out own version to match that.

calbach:
The issue is that apparently Snyk cannot handle the current way that we pass constants into our package versions in build.gradle. I verified Snyk goes green if I hardcode the version instead - this is  a false positive.

I filed a ticket to track this: https://precisionmedicineinitiative.atlassian.net/browse/RW-5007

PR #3621, opened 2020-05-28
[risk=no] Add additional metadata fields for user/workspace export

calbach:
These were requested by Karthik.

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -11,6 +11,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


calbach:
@jaycarlton ptal

calbach:
Looks like Jay is out today, switching to @jmthibault79

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -230,11 +238,34 @@ private WorkspaceExportRow toWorkspaceExportRow(DbWorkspace workspace) {
...
+            .findFirstByUser(user)
+            .map(via -> via.getInstitution().getDisplayName())
+            .orElse("");
+
     WorkspaceExportRow row = new WorkspaceExportRow();
     row.setCreatorContactEmail(user.getContactEmail());
     row.setCreatorUsername(user.getUsername());
+    row.setInstitution(verifiedInstitutionName);
     row.setCreatorFirstSignIn(
         user.getFirstSignInTime() == null ? "" : dateFormat.format(user.getFirstSignInTime()));
```
jmthibault79, line 252: nit: I'd prefer the nullable/map/orElse pattern here but I'm not gonna block on that

api/tools/src/main/java/org/pmiops/workbench/tools/WorkspaceExportRow.java
```diff
@@ -38,7 +50,7 @@ public void setCreatorFirstSignIn(String creatorFirstSignIn) {
   }
 
   @CsvBindByName(column = "First Sign In")
-  @CsvBindByPosition(position = 2)
+  @CsvBindByPosition(position = 3)
```
jmthibault79, line 53: welp this is certainly tedious.  assuming there's no better way or you'd have done it?

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
calbach: Agreed, done

api/tools/src/main/java/org/pmiops/workbench/tools/WorkspaceExportRow.java
calbach: Indeed - I didn't do the initial implementation or review, but from a quick scan around some of the documentation, it seems that custom names + custom ordering is not very well supported by this CSV library. I would not be surprised if this is the best way (at least, for this library). Will consider digging deeper if we decide to extend the lifespan of this tool further - it may be going away soon'ish with new metrics export plans.

jaycarlton: FWIW, this was not intended to be a long-lived implementation.

api/tools/src/main/java/org/pmiops/workbench/tools/WorkspaceExportRow.java
```diff
@@ -62,7 +122,7 @@ public void setProjectId(String projectId) {
   }
 
   @CsvBindByName(column = "Workspace Project ID")
-  @CsvBindByPosition(position = 4)
+  @CsvBindByPosition(position = 9)
```
jaycarlton, line 125: One option is to just write an array of JSON objects and use an off-the-shelf json->csv converter.

PR #3623, opened 2020-05-29
[RW-4383][risk=no] Puppeteer Cohort test case

aweng98:
Screen Shot 2020-05-29 at 11 23 52 AM

### DIFF
--- a/e2e/app/component/create-criteria-modal.ts
+++ b/e2e/app/component/create-criteria-modal.ts
@@ -1,28 +1,38 @@
-import {Page} from 'puppeteer';
 import ClrIconLink from 'app/element/clr-icon-link';
 import {ElementType} from 'app/xpath-options';
-import Dialog from './dialog';
+import {Page} from 'puppeteer';
+import {waitForNumericalString} from 'utils/waits-utils';
+import {xPathOptionToXpath} from 'app/element/xpath-defaults';
+import Dialog, {ButtonLabel} from './dialog';
+import SelectMenu from './select-menu';
 
 const defaultXpath = '//*[contains(concat(" ", normalize-space(@class), " "), " crit-modal-container ")]';
 
 export enum PhysicalMeasurementsCriteria {
+  BloodPressure = 'Blood Pressure',
   BMI = 'BMI',
-  HEIGHT = 'Height',
-  WEIGHT = 'Weight',
-  WAIST_CIRCUMFERENCE = 'Waist Circumference',
-  HIP_CIRCUMFERENCE = 'Hip Circumference',
-  HEART_RATE = 'Heart Rate',
-  BLOOD_PRESSURE = 'Blood Pressure',
-  PREGNANT_ENROLLMENT = 'Pregnant at enrollment',
-  WHEELCHAIR_USER = 'Wheelchair user at enrollment',
-  BP_HYPOTENSIVE = 'Hypotensive (Systolic <= 90 / Diastolic <= 60)', // Blood Pressure
-  BP_NORMAL = 'Normal (Systolic <= 120 / Diastolic <= 80)', // Blood Pressure
-  BP_PREHYPERTENSIVE = 'Pre-Hypertensive (Systolic 121-139 / Diastolic 81-89)', // Blood Pressure
-  BP_HYPERTENSIVE = 'Hypertensive (Systolic >= 140 / Diastolic >= 90)', // Blood Pressure
-  HR_NO_IRREGULARITY = 'No-irregularity detected', // Heart Rhythm Status
-  HR_IRREGULARITY = 'Irregularity detected', // Heart Rhythm Status
+  BPHypotensive = 'Hypotensive (Systolic <= 90 / Diastolic <= 60)', // BP prefix for Blood Pressure
+  BPNormal = 'Normal (Systolic <= 120 / Diastolic <= 80)', // Blood Pressure
+  BPPrehypertensive = 'Pre-Hypertensive (Systolic 121-139 / Diastolic 81-89)', // Blood Pressure
+  BPHypertensive = 'Hypertensive (Systolic >= 140 / Diastolic >= 90)', // Blood Pressure
+  HeartRate = 'Heart Rate',
+  Height = 'Height',
+  HipCircumference = 'Hip Circumference',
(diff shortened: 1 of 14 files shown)


e2e/app/element/xpath-defaults.ts
```diff
@@ -14,56 +14,47 @@ export function xPathOptionToXpath(xOpts: XPathOptions, container?: Container):
     if (iconShape === undefined) {
       throw new Error(`Incorrect XPathOptions configuration for Icon: set "iconShape" value.`);
     }
-  } else {
-    if (name === undefined && normalizeSpace === undefined && containsText === undefined) {
-      throw new Error(`Incorrect XPathOptions configuration for label name: 
-      Cannot find a text parameter: set "name", "containsText" or "normalizeSpace" value.`);
-    }
   }
```
aweng98, line 17: Removing this check because the Add Criteria dialog has input fields which don't have labels. 🤷‍♂️
Minor tweaking for xpath expressions to accommodate new type of UI fields in Add Criteria dialog.

e2e/app/page/data-page.ts
```diff
@@ -38,27 +38,27 @@ export default class DataPage extends AuthenticatedPage {
...
-    const tabs = await this.page.$x(selector);
-    for (const tab of tabs) {
-      const contentProp = await tab.getProperty('textContent');
-      if (await contentProp.jsonValue() === tabName) {
-        return tab.click();
-      }
-      await tab.dispose();
-    }
-    throw new Error(`Failed to find page tab with name ${tabName}`);
+  async openTab(tabName: LabelAlias): Promise<void> {
```
aweng98, line 50: rename `selectTab` to `openTab` because I think `select` word can be confused with menu select.

e2e/utils/waits-utils.ts
```diff
@@ -75,6 +75,23 @@ export async function waitForPropertyEquality(
...
 }
 
+export async function waitForNumericalString(page: Page, xpath: string): Promise<string> {
+  await page.waitForXPath(xpath, {visible: true});
+
+  const numbers =  await page.waitForFunction( xpathSelector => {
+    const node = document.evaluate(xpathSelector, document.body, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
+    if (node !== null) {
+      const txt = node.textContent.trim();
+      const re = new RegExp(/^\d{1,3}(,\d{3})*(\.\d+)?$/);
```
aweng98, line 85: Cohort build page is not ready until Participants Count is calculated after adding new criterias. Participants Count numbers contain comma character(s). This function waits until Participants Count changes to a digits string with comma(s). It is used to get Counts and to indicate page is ready for next actions.

e2e/app/component/tiered-menu.ts
```diff
@@ -1,12 +1,42 @@
...
   constructor(page: Page, xpath: string = defaultXpath) {
     super(page, xpath);
   }
 
+  /**
+   * Select menu item(s).
+   * @param {ElementHandle} menuParentElement
+   * @param {string | string[]} selectMenuItems
+   */
+  async select(selectMenuItems: string[]): Promise<void> {
```
als364, line 18: It looks like what this function does is to find and click all items in a list of items - can you change the name of the function to indicate that that is what it does?

e2e/utils/waits-utils.ts
als364: This seems fragile. Participant counts may not always be over 1000. Could you instead check that the content changes?

e2e/app/component/create-criteria-modal.ts
```diff
@@ -31,7 +41,53 @@ export default class CreateCriteriaModal extends Dialog {
   }
 
   async waitForPhysicalMeasurementCriteriaLink(criteriaType: PhysicalMeasurementsCriteria): Promise<ClrIconLink> {
-    return ClrIconLink.forLabel(this.page, {type: ElementType.Icon, name: criteriaType, iconShape: 'slider'}, this);
+    return ClrIconLink.forLabel(this.page, {type: ElementType.Icon, name: criteriaType, iconShape: 'slider', ancestorLevel: 2}, this);
```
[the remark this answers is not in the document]
aweng98: `forLabel` function name was chosen for this reason at the beginning, a common static function that finds an element matching a label name. An element can be type of Icon, Select, Textbox, Textarea, etc.

e2e/app/component/tiered-menu.ts
aweng98: user probably doesn't care about there is an internal helper function for finding menuitems. do you recommend a better name? I'd have to rename other `select()` functions too to keep naming convention standard.
https://github.com/all-of-us/workbench/blob/4091a0588aabe68831499fe337a16c33a2ce4e06/e2e/app/component/select-menu.ts#L26

e2e/utils/waits-utils.ts
aweng98: good catch. I updated regex to catch numbers less than one thousand. I need to check that calcualated count turns into a number.

e2e/app/component/tiered-menu.ts
als364: Developers are the users here and as a developer I'd prefer to have meaningful function names. I would call this `clickMenuItems`, and for other types of inputs I would similar functions things like `clickRadioButton` or `clickCheckBox`.

e2e/app/component/create-criteria-modal.ts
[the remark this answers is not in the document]
als364: Right, but almost none of those have labels. Could this instead be called `findByName`?

e2e/app/component/tiered-menu.ts
aweng98: Sounds good.
I renamed following three function names (for consistency):
- `selectAction` to `clickAction` in `EllipsisMenu.ts`.
- `select` to `clickMenuItem` in `SelectMenu.ts`.
- `select` to `clickMenuItem` in `TieredMenu`.

e2e/app/component/create-criteria-modal.ts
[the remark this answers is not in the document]
als364: Yeah, go ahead and do that in a separate PR, that's gonna be a big change

PR #3626, opened 2020-05-29
[RW-4906][risk=low] bulkAuditProjectAccess improvements

calbach:
- Bug fix: actually paginate the project results from CRM. Previous code only examined the first page..
- Bug fix: for default projects (no org/folder), project.parent is null. Handle this
- Behavior: continue after we hit an error. This will allow us to quantify whether we're having issues for all users, or just one.
- Logs: actually log which user we're having auth issues for. Swallowing the exception at a low level and throwing a runtime exception instead is not useful in this situation.


calbach:
Puppeteer is a known flake - unrelated to this PR.

Also verified via local cron run that project parent is nullable, but adding my test user to a default project.

jaycarlton:
Once this is working again in prod, we need to flip the Stackdriver alert back on. Stackdriver's absent metric trigger doesn't seem to work if the signal hasn't been present since the condition was set up.

PR #3627, opened 2020-05-29
[RW-4521][risk=low] Create API for updating verified institution

als364:
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -388,9 +388,10 @@ public class ProfileController implements ProfileApiDelegate {
     }
 
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
-      request.getProfile().setVerifiedInstitutionalAffiliation(
-          profileService.validateInstitutionalAffiliation(request.getProfile())
-      );
+      request
+          .getProfile()
+          .setVerifiedInstitutionalAffiliation(
+              profileService.validateInstitutionalAffiliation(request.getProfile()));
     }
 
     final Profile profile = request.getProfile();
@@ -687,8 +688,7 @@ public class ProfileController implements ProfileApiDelegate {
     updateInstitutionalAffiliations(updatedProfile, user);
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
       updatedProfile.setVerifiedInstitutionalAffiliation(
-        profileService.validateInstitutionalAffiliation(updatedProfile)
-      );
+          profileService.validateInstitutionalAffiliation(updatedProfile));
     }
 
     userService.updateUserWithConflictHandling(user);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -8,7 +8,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.logging.Level;
@@ -32,7 +31,6 @@ import org.pmiops.workbench.db.model.DbDemographicSurvey;
 import org.pmiops.workbench.db.model.DbInstitutionalAffiliation;
 import org.pmiops.workbench.db.model.DbPageVisit;
 import org.pmiops.workbench.db.model.DbUser;
-import org.pmiops.workbench.db.model.DbVerifiedInstitutionalAffiliation;
 import org.pmiops.workbench.exceptions.BadRequestException;
 import org.pmiops.workbench.exceptions.ConflictException;
 import org.pmiops.workbench.exceptions.EmailException;
@@ -390,7 +388,7 @@ public class ProfileController implements ProfileApiDelegate {
     }
 
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
-      verifyInstitutionalAffiliation(request.getProfile());
+      profileService.verifyInstitutionalAffiliation(request.getProfile());
     }
 
     final Profile profile = request.getProfile();
@@ -560,28 +558,6 @@ public class ProfileController implements ProfileApiDelegate {
     }
   }
 
-  private void verifyInstitutionalAffiliation(Profile profile) {
-    String userName = profile.getUsername();
-    String contactEmail = profile.getContactEmail();
-    DbVerifiedInstitutionalAffiliation dbVerifiedAffiliation =
-        verifiedInstitutionalAffiliationMapper.modelToDbWithoutUser(
-            profile.getVerifiedInstitutionalAffiliation(), institutionService);
-    if (!institutionService.validateAffiliation(dbVerifiedAffiliation, contactEmail)) {
-      final String msg =
-          Optional.ofNullable(dbVerifiedAffiliation)
-              .map(
-                  affiliation ->
-                      String.format(
-                          "Cannot create user %s: contact email %s is not a valid member of institution '%s'",
-                          userName, contactEmail, affiliation.getInstitution().getShortName()))
-              .orElse(
-                  String.format(
-                      "Cannot create user %s: contact email %s does not have a valid institutional affiliation",
-                      userName, contactEmail));
-      throw new BadRequestException(msg);
-    }
-  }
-
   private void checkUserCreationNonce(DbUser user, String nonce) {
     if (Strings.isNullOrEmpty(nonce)) {
       throw new BadRequestException("missing required creationNonce");
@@ -707,6 +683,9 @@ public class ProfileController implements ProfileApiDelegate {
     user.setLastModifiedTime(now);
 
     updateInstitutionalAffiliations(updatedProfile, user);
+    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
+      profileService.verifyInstitutionalAffiliation(updatedProfile);
+    }
 
     userService.updateUserWithConflictHandling(user);
 
@@ -718,14 +697,16 @@ public class ProfileController implements ProfileApiDelegate {
 
   @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
   @Override
-  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(Long userId,
-      VerifiedInstitutionalAffiliation verifiedInstitution) {
-      DbUser dbUser = userDao.findUserByUserId(userId);
-      Profile profile = profileService.getProfile(dbUser);
+  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(
+      Long userId, VerifiedInstitutionalAffiliation verifiedInstitution) {
+    DbUser dbUser = userDao.findUserByUserId(userId);
+    Profile profile = profileService.getProfile(dbUser);
 
-      profile.setVerifiedInstitutionalAffiliation(verifiedInstitution);
-      this.updateProfile(profile);
-      return ResponseEntity.ok(new EmptyResponse());
+    profile.setVerifiedInstitutionalAffiliation(verifiedInstitution);
+    profileService.verifyInstitutionalAffiliation(profile);
+    this.updateProfile(profile);
+
+    return ResponseEntity.ok(new EmptyResponse());
   }
 
   private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user) {
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -703,7 +703,6 @@ public class ProfileController implements ProfileApiDelegate {
     Profile profile = profileService.getProfile(dbUser);
 
     profile.setVerifiedInstitutionalAffiliation(verifiedInstitution);
-    profileService.verifyInstitutionalAffiliation(profile);
     this.updateProfile(profile);
 
     return ResponseEntity.ok(new EmptyResponse());
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -388,7 +388,9 @@ public class ProfileController implements ProfileApiDelegate {
     }
 
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
-      profileService.verifyInstitutionalAffiliation(request.getProfile());
+      request.getProfile().setVerifiedInstitutionalAffiliation(
+          profileService.validateInstitutionalAffiliation(request.getProfile())
+      );
     }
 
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -567,28 +558,6 @@ private void verifyCaptcha(String captchaToken) {
     }
   }
 
-  private void verifyInstitutionalAffiliation(Profile profile) {
```
als364, line 570: pushed this down into the Service layer. Ideally more things would be there but that would take a LOT of time for a relatively time sensitive PR.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -723,6 +695,20 @@ private boolean userHasEverLoggedIn(
...
 
+  @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
+  @Override
+  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(
+      Long userId, VerifiedInstitutionalAffiliation verifiedInstitution) {
+    DbUser dbUser = userDao.findUserByUserId(userId);
+    Profile profile = profileService.getProfile(dbUser);
+
+    profile.setVerifiedInstitutionalAffiliation(verifiedInstitution);
+    profileService.verifyInstitutionalAffiliation(profile);
```
als364, line 706: note to self: this is probably redundant

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -723,6 +695,20 @@ private boolean userHasEverLoggedIn(
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
+  @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
+  @Override
+  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(
```
als364, line 700: I think ideally I'd run all updates like this directly through `updateProfile` or something like `adminUpdateProfile` rather than something specific to verified institution. I'll address that in RW-4980 where we'll take a holistic look at what we want the API structure to look like for the user admin

api/src/main/java/org/pmiops/workbench/profile/ProfileMapper.java
```diff
@@ -19,7 +20,8 @@
       CommonMappers.class,
       DemographicSurveyMapper.class,
       InstitutionalAffiliationMapper.class,
-      PageVisitMapper.class
+      PageVisitMapper.class,
+      VerifiedInstitutionalAffiliationMapper.class
```
als364, line 24: note to self: unnecessary

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -36,7 +43,12 @@
   @Autowired VerifiedInstitutionalAffiliationMapper verifiedInstitutionalAffiliationMapper;
 
   @TestConfiguration
-  @MockBean({FreeTierBillingService.class})
+  @MockBean({
```
als364, line 46: note to self: revert

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -36,7 +43,12 @@
...
+      FreeTierBillingService.class,
+      InstitutionDao.class,
+      InstitutionService.class,
+      UserService.class,
+  })
```
als364, line 51: I think I want to move the tests for `verifyInstitutionalAffiliation` into this test class since that function is now in the Service layer

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -29,14 +40,18 @@
...
+  @MockBean
+  private VerifiedInstitutionalAffiliationMapper mockVerifiedInstitutionalAffiliationMapper;
+
   @Autowired ProfileService profileService;
   @Autowired UserDao userDao;
-  @Autowired VerifiedInstitutionalAffiliationMapper verifiedInstitutionalAffiliationMapper;
 
   @TestConfiguration
-  @MockBean({FreeTierBillingService.class})
+  @MockBean({FreeTierBillingService.class, InstitutionService.class, UserService.class})
```
gjuggler, line 54: Why is InstitutionService included as a MockBean here and also above? I think you can remove one or the other, no?

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -77,4 +92,203 @@ public void testReturnsLastAcknowledgedTermsOfService() {
...
+            .contactEmail("kibitz@broadinstitute.org");
+
+    DbInstitution dbInstitution = new DbInstitution();
+    dbInstitution.setShortName("Broad");
+    dbInstitution.setDisplayName("The Broad Institute");
+
+    when(mockInstitutionDao.findOneByShortName("Broad")).thenReturn(Optional.of(dbInstitution));
+
+    when(mockInstitutionService.validateAffiliation(
+            any(DbVerifiedInstitutionalAffiliation.class), anyString()))
```
gjuggler, line 117: [opt] I feel like there could be some meaningful argument capture here, to ensure that this method is called with the correct institutional affiliation shortName. This would require building a non-empty DbVerifiedInstitutionalAffiliation at L122, which feels like a good idea anyway (since in the real world, we'll expect that Db model object to be non-empty).

This connection, between the mapper return value and the institutionService call, is IMO the main piece of logic that's being tested in this positive case – so it seems potentially useful to explicitly check.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -318,13 +316,6 @@ private void validateUpdatedProfile(Profile updatedProfile, Profile prevProfile)
       // See RW-1488.
       throw new BadRequestException("Changing username is not supported");
     }
-    final VerifiedInstitutionalAffiliation updatedAffil =
```
gjuggler, line 321: We're not changing the fact that non-admin profile updates aren't allowed to change affiliation, right? Removing this line breaks that enforcement.

I think I know why you removed this, because we're using a unified code path for non-admin and admin profile updates. But we need to account for the fact that self-profile updates have different business logic than admin-driven profile updates.

A couple options:
 - Bump this logic into a `checkUserProfileUpdate` method, whose job is basically to reject any requests where a user is trying to update things that they shouldn't be allowed to. This fits nicely with a higher-level goal of pushing more of the core upsert logic into a service, but keeping the top-level "can I do this" logic in the controller itself.
 - Pass down some additional call context to methods like this, e.g. "boolean isAdmin" (Seems clunky)

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -723,6 +699,19 @@ private boolean userHasEverLoggedIn(
...
 
+  @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
+  @Override
+  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(
+      Long userId, VerifiedInstitutionalAffiliation verifiedInstitution) {
+    DbUser dbUser = userDao.findUserByUserId(userId);
+    Profile profile = profileService.getProfile(dbUser);
+
+    profile.setVerifiedInstitutionalAffiliation(verifiedInstitution);
+    this.updateProfile(profile);
```
gjuggler, line 710: I don't think this call will work to update any user that isn't the currently-authenticated user. See the first line of updateProfile: `DbUser user = userProvider.get();`.

I think you'll need to extract a new method, `updateProfileForUser(DbUser user)`, in order to make the correct call here.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -714,6 +686,10 @@ private boolean userHasEverLoggedIn(
     user.setLastModifiedTime(now);
 
     updateInstitutionalAffiliations(updatedProfile, user);
+    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
+      updatedProfile.setVerifiedInstitutionalAffiliation(
```
gjuggler, line 690: AFAICT, calling `setVerifiedInstitutionalAffiliation` on the updatedProfile object has no effect, since this object is not persisted or returned in the HTTP response. Is it necessary?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
gjuggler: Makes total sense – I'd love to brainstorm ideas here. As I'm going through this PR I'm starting to build some mental model of how profile vs. user APIs and models might be cleaned up for the future.

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -77,4 +92,203 @@ public void testReturnsLastAcknowledgedTermsOfService() {
...
+            .institutionDisplayName("The Broad Institute")
+            .institutionalRoleEnum(InstitutionalRole.OTHER)
+            .institutionalRoleOtherText("Kibitzing");
+
+    Profile profile =
+        new Profile()
+            .verifiedInstitutionalAffiliation(affiliation)
+            .contactEmail("kibitz@broadinstitute.org");
+
+    DbInstitution dbInstitution = new DbInstitution();
```
gjuggler, line 110: I'm usually accepting of a fair amount of repetition in test code, but I think these 4 lines to create a mock Broad institution are used in every single test case. That's my threshold of "bump it into the setup and kill a few dozen lines" in unit tests.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
als364: We can't control the signature of the public controller methods without changing the API and that seems like overkill for something like this. so I'll do the first one.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
als364: yep, i totally read this wrong before. argh

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -723,6 +699,19 @@ private boolean userHasEverLoggedIn(
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
+  @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
```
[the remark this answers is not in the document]
als364: yep, planning on doing that as part of the holistic api design work that will happen in RW-4980

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
als364: Yes, let's talk about that later.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -723,6 +699,19 @@ private boolean userHasEverLoggedIn(
...
+  @Override
+  public ResponseEntity<EmptyResponse> updateVerifiedInstitutionalAffiliation(
+      Long userId, VerifiedInstitutionalAffiliation verifiedInstitution) {
```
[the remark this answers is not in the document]
als364: good call.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -70,4 +88,62 @@ public Profile getProfile(DbUser user) {
...
+    Optional<DbInstitution> institution =
+        institutionDao.findOneByShortName(
+            verifiedInstitutionalAffiliation.getInstitutionShortName());
+    if (!institution.isPresent()) {
+      throw new NotFoundException(
+          String.format(
+              "Could not find institution %s in database",
+              verifiedInstitutionalAffiliation.getInstitutionShortName()));
+    }
+    if (!institution
```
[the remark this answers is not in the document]
als364: I probably wrote this when I was still sorting out the differences between the db model and the api model.

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
als364: Something like:

```
create mock DbVerifiedInstitutionalAffiliation with appropriate values
tell mapper to return mock when called
tell institution service to return true when called with mock (or when called with anything?)
spy on institutionService validate call to make sure the DbVerifiedInstitutionalAffiliation has the right shortname
```

Wouldn't that be redundant? Did you maybe mean the profileService call?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
als364: hm. I don't think we have a pattern of checking authority on the backend outside of `@AuthorityRequired` annotations and I don't want to introduce one. 🙁 maybe I'll fork the controller now and factor out as much stuff from updateProfile as I can.

gjuggler: I don't think forking the controller should be needed. The crux of this issue (similar to the one further down about `userProvider.get()`) is that we're attempting to use the same top-level controller method for two different purposes now.

Instead, one could create a private `updateProfileImpl` method which is called by the two top-level controller methods. With that change in place, to resolve this issue you could create a `checkUserProfileUpdate` method which is only called by the `updateProfile` top-level controller.

e.g. something like the following would be very clear

```
public ResponseEntity<Void> updateProfile(Profile updatedProfile) {
  validateUserProfileUpdate(updatedProfile);
  updateProfileImpl(updatedProfile, userProvider.get());
}

public ResponseEntity<Void> updateVerifiedInstitutionalAffiliation(long userId, VerifiedInstitutionalAffiliation verifiedInstitution) {
  [...]
  updateProfileImpl(updatedProfile, userDao.findUserByUserId(userId));
}
```

and eventually updateProfileImpl should be factored into ProfileService.

als364: moved into updatedProfile.

api/src/main/resources/workbench-api.yaml
```diff
@@ -1079,6 +1079,31 @@ paths:
...
             "$ref": "#/definitions/ErrorResponse"
+  "/v1/admin/users/{userId}/updateVerifiedInstitutionalAffiliation":
+    parameters:
+    - "$ref": "#/parameters/userId"
+    post:
+      tags:
+      - profile
+      consumes:
+      - application/json
+      description: Updates a user's verified institution. Requires ACCESS_CONTROL_ADMIN authority.
```
gjuggler, line 1090: Nit: it might be worth mentioning here that institutional validation is executed by this API method, e.g. a request will fail if the indicated user is not a valid member of the new affiliated institution.

als364:
Snyk failure seems unrelated - I didn't touch the UI at all

PR #3628, opened 2020-05-29
[RW-4972][risk=low] Update to latest welder hash

calbach:
This is the latest default version for Leo. Will discuss having a better method for automatically getting these updates in the future, e.g. a floating image label: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L166


calbach:
Thank you for testing! I did a quick manual test as well, and didn't see any issues.

PR #3629, opened 2020-05-29
[RW-4906][risk=no] bulkProjectAudit: handle 0 projects case, add regression test

calbach:
This caused an NPE in the test environment for users with access to 0 projects.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
@@ -94,7 +94,9 @@ public class CloudResourceManagerServiceImpl implements CloudResourceManagerServ
                     .list()
                     .setPageToken(pageToken.orElse(null))
                     .execute();
-            projects.addAll(resp.getProjects());
+            if (resp.getProjects() != null) {
+              projects.addAll(resp.getProjects());
+            }
 
             // The API does not specify null or empty string; treat both as empty to be safe.
             pageToken =
--- a/api/src/integration/java/org/pmiops/workbench/google/CloudResourceManagerServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/CloudResourceManagerServiceImplIntegrationTest.java
@@ -23,14 +23,25 @@ public class CloudResourceManagerServiceImplIntegrationTest extends BaseIntegrat
   // This is a single hand created user in the fake-research-aou.org gsuite.
   // It has one project that has been shared with it, AoU CRM Integration Test
   // in the firecloud dev domain.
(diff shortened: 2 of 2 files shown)


api/src/main/java/org/pmiops/workbench/google/CloudResourceManagerServiceImpl.java
```diff
@@ -94,7 +94,9 @@ private CloudResourceManager getCloudResourceManagerServiceWithImpersonation(DbU
                     .list()
                     .setPageToken(pageToken.orElse(null))
                     .execute();
-            projects.addAll(resp.getProjects());
+            if (resp.getProjects() != null) {
```
jaycarlton, line 97: nit: I'd do `Optional.ofNullable(resp.getProjects()).ifPresent(projects::addAll);`

calbach: I'm aware that it can be written this way, but do you really find this to be more readable? Personally I find it a bit convoluted. I don't object to this style in certain cases e.g. if it simplifies a large block of code with repeated setters, but I don't think we want to place prohibition on simple if statements in the codebase.

jaycarlton: Ideally I'd change `getProjects()` to never be null, or be optional (though I wouldn't do optional<List<thing>>).

I feel like the intent is a bit clearer, and there are fewer places for a maintainer to add a bug. `if`s beget more `if`s sometimes.

Not a big win in this case.

calbach: Agree that would be better, and it's what I originally expected hence the bug. It's a 3p library we don't control though.

Avoiding repetition of the value is a good argument for switching the approach, though I still don't love the side-effect tacked onto the end of the expression with the `ifPresent` approach. Perhaps something like this instead:

```
projects.addAll(CollectionUtils.emptyIfNull(resp.getProjects()));
```

or 

```
projects.addAll(Optional.ofNullable(resp.getProjects()).orElse(Collections.emptyList()));
```

jaycarlton: _ I still don't love the side-effect tacked onto the end of the expression with the ifPresent approach_
Yeah, `foreach` and `ifPresent` do sneak side effects into otherwise functional expressions. I wasn't raised functional fundamentalist. I still prefer `ifPresent()` when chained after methods that return `Optional` though. It's a nice counterpart to `orElseThrow()`, `orElseGet()`, and `orElse()`, all of which come at the end of the chain.

Of your two options I like the first one better. One thing to note (likely inconsequential here) is that in your scheme you call `putAll` regardless of whether you need to or not. As long as there are no side effects, that's OK.

PR #3630, opened 2020-06-01
[no ticket][risk=no] Puppeteer function rename

aweng98:
Renamed `forLabel` function name to `findByName` to be clearer.


aweng98:
@als364 Hello, this is the PR for rename. can you help with review since you know the context?

PR #3636, opened 2020-06-02
RW-5019 adding validation for cohort def on save

freemabd:
1) Adding validation for cohort definition on save
2) Adding test case for above

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/CohortsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/CohortsController.java
@@ -2,6 +2,8 @@ package org.pmiops.workbench.api;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/pmiops/workbench/api/CohortsController.java
```diff
@@ -136,6 +139,16 @@ private void checkForDuplicateCohortNameException(String newCohortName, DbWorksp
         workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
     DbWorkspace workspace = workspaceService.getRequired(workspaceNamespace, workspaceId);
 
+    try {
+      // validate the cohort definition
+      new Gson().fromJson(cohort.getCriteria(), SearchRequest.class);
```
NehaBroad, line 144: Is there any chance that cohort could be null? Should we also handle nullPointer?

api/src/test/java/org/pmiops/workbench/api/CohortsControllerTest.java
```diff
@@ -368,6 +412,22 @@ public void testGetCohortsInWorkspace() throws Exception {
     assertThat(cohorts.size()).isEqualTo(2);
   }
 
+  @Test
```
NehaBroad, line 415: we can add @Test(expected = ServerErrorException.class) rather than fail

api/src/main/java/org/pmiops/workbench/api/CohortsController.java
freemabd: UI doesn't allow for saving a null cohort and i think swagger would throw an exception if the request body was null.

PR #3641, opened 2020-06-04
parallelism for api-build-test job

aweng98:
827 unit tests in `api/src/test/java/org/pmiops/workbench/**/*.java` and `*.kt`.

Screen Shot 2020-06-04 at 6 59 45 PM

Without parallelism, `api-unit-test` job takes some time to finish, range from 15 - 25 min long.
With parallelism 4, job time is reduced to under 10 min every time.

Larger parallelism does not dramatically further reduces test duration time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,8 +30,8 @@ anchors:
     # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
     # so we bumped this down further to 2G.
     JAVA_TOOL_OPTIONS: -Xmx2g
-    # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
-    GRADLE_OPTS: "-Dorg.gradle.daemon=false"
+    # See: https://support.circleci.com/hc/en-us/articles/360021812453-Common-Android-memory-issues
+    GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process
     TERM: dumb
     MYSQL_ROOT_PASSWORD: ubuntu
 
@@ -294,9 +294,6 @@ jobs:
     # We're hitting OOM failures as of 1 May 2020.  Increase to 6GB RAM by specifying a medium+ machine.
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     # resource_class: medium+
-    environment:
-      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process
-      JAVA_TOOL_OPTIONS: -Xmx1536m
     steps:
       - checkout_init_git
       - ensure_branch_has_changes:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ jobs:
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     # resource_class: medium+
     environment:
-      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=1 -Dkotlin.incremental=false
+      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process
       JAVA_TOOL_OPTIONS: -Xmx1536m
     steps:
       - checkout_init_git
@@ -322,7 +322,7 @@ jobs:
               | sed 's/\.[^.]*$//' \
               | circleci tests split --split-by=timings --timings-type=classname --index=$CIRCLE_NODE_INDEX)
             GRADLE_ARGS=$(echo $CLASSNAMES | awk '{for (i=1; i<=NF; i++) print "--tests", $i}')
-            java -version && gradle -v && gradle --no-daemon test $GRADLE_ARGS --stacktrace
+            java -version && gradle -v && gradle test $GRADLE_ARGS
       - run:
           # Lint last; it's more important to surface test failures early.
           name: Java linting
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ jobs:
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     # resource_class: medium+
     environment:
-      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false
+      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=1 -Dkotlin.incremental=false
       JAVA_TOOL_OPTIONS: -Xmx1536m
     steps:
       - checkout_init_git
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,6 +295,7 @@ jobs:
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     # resource_class: medium+
     environment:
+      GRADLE_OPTS: -Dorg.gradle.daemon=false -Dkotlin.incremental=false
       JAVA_TOOL_OPTIONS: -Xmx1536m
     steps:
       - checkout_init_git
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,9 @@ commands:
           command: |
             if [ ${CIRCLE_BRANCH} != "" ] &&
               [ ${CIRCLE_BRANCH} != "master" ] &&
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
+              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) \
+                | grep -qvE '(.md)|(.pdf)|(.png)|(.svg)|(Test.java)|(Test.kt)|(.mp4)|(.log)|(.txt)'; then
+
                 echo "No code changes on non-master branch. halting job."
                 circleci step halt
             fi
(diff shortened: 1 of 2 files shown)


calbach:
RE bigquery: those are BigQuery integration tests, they do not run as part of the unit test suite.

.circleci/config.yml
```diff
@@ -285,31 +285,46 @@ commands:
...
+          #dir_name: "api"
       - manage_api_cache:
           restore: true
       - run:
           name: Validate Swagger definitions
           working_directory: ~/workbench/api
-          command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" == 0 ]; then
+              ./project.rb validate-swagger --project-prop verboseTestLogging=yes
```
jaycarlton, line 304: Does this command work currently? ISTR @jmthibault79 has a ticket on it. It doesn't take too long thogh.

.circleci/config.yml
```diff
@@ -285,31 +285,46 @@ commands:
...
+            CLASSNAMES=$(circleci tests glob "src/test/java/org/pmiops/workbench/**/*.java" \
+              | sed 's@src/test/java/@@' \
+              | sed 's@/@.@g' \
+              | sed 's/.\{5\}$//' \
+              | circleci tests split --split-by=timings --timings-type=classname)
+
+            GRADLE_ARGS=$(echo $CLASSNAMES | awk '{for (i=1; i<=NF; i++) print "--tests",$i}')
+            ./gradlew --no-daemon test $GRADLE_ARGS
       - run:
           # Lint last; it's more important to surface test failures early.
```
jaycarlton, line 319: Can we lint in parallel to the Unit Tests? I have a mental note to move this up.

.circleci/config.yml
```diff
@@ -285,31 +285,46 @@ commands:
...
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" == 0 ]; then
+              ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+            fi
       - run:
           name: Run Java unit tests
           working_directory: ~/workbench/api
-          command: ./project.rb test
+          command: |
+            CLASSNAMES=$(circleci tests glob "src/test/java/org/pmiops/workbench/**/*.java" \
```
jaycarlton, line 310: I would pull everything under `src/test/java`. If we added another package there, we'd still want those here I think.

Also, some comment or description attribute here would help. I'm not p on my `sed`isms.

.circleci/config.yml
```diff
@@ -285,31 +285,46 @@ commands:
...
+    parallelism: 6
     # We're hitting OOM failures as of 1 May 2020.  Increase to 6GB RAM by specifying a medium+ machine.
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     resource_class: medium+
     <<: *java_defaults
     steps:
       - checkout_init_git
-      - ensure_branch_has_changes:
-          dir_name: "api"
+      #- ensure_branch_has_changes:
```
jaycarlton, line 295: why is this commented out?

.circleci/config.yml
```diff
@@ -285,31 +285,46 @@ commands:
 # -------------------------
 jobs:
   api-build-test:
```
jaycarlton, line 287: Would it be possible to rename this to `api-unit-test`?

.circleci/config.yml
```diff
@@ -124,7 +126,9 @@ commands:
           command: |
             if [ ${CIRCLE_BRANCH} != "" ] &&
               [ ${CIRCLE_BRANCH} != "master" ] &&
-              [ $(git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep << parameters.dir_name >>/ | wc -l | xargs) == 0 ]; then
+              [ $(git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep << parameters.dir_name >>/ | wc -l | xargs) == 0 ] &&
+              git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(Test.kt)|(Test.java)|(.yaml)'; then
+
```
aweng98, line 131: Don't halt job if tests and .yaml files were found in commits.

.circleci/config.yml
```diff
@@ -360,13 +385,14 @@ jobs:
               --version circle-ci-test \
               --key-file circle-sa-key.json \
               --promote
-      - manage_api_cache:
-          save: true
```
aweng98, line 364: gradle cache saving will never happen in `*-deploy-to-*` jobs because they requires other jobs that will save gradle cache if cache does not exists.

.circleci/config.yml
```diff
@@ -284,11 +288,15 @@ commands:
 #        JOBS
 # -------------------------
 jobs:
-  api-build-test:
+  api-unit-test:
```
aweng98, line 291: renamed from `api-build-test` to `api-unit-test`

aweng98:
I have a good news. I have a workaround for Gradle OOM issue.
`GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process`

I'm removing `resource_class: medium+` for all api-* jobs. And, I am going to monitor builds for OOM to make sure workaround is working.

.circleci/config.yml
jaycarlton: nice! @calbach will this break anything?

calbach: I doubt this will break anything. Outside of this file, these names just affect display strings on PRs and in the Circle UI - there should currently be no other programmatic dependencies on this.

aweng98:
> Nice finds on the resource settings! Do you have any notes on how you iterated on this? i.e. did you run any tools to identify the source of memory utilization, or was it mostly experimentation with flags you found suggested in various articles?

No tools, mostly by experimentation and using Gradle `--stacktrace` option. Like right away, after including Kotlin tests, OOM started happen for CircleCI job on `medium` resource_class. When I change to use `medium+`, jobs were running okay again. I had config`max.workers=1` with trace on. Thru the trace logs, I saw there were two Gradle threads being used. One thread was started later before running Kotlin tests. So I was pretty sure Kotlin tests were related to OOM or maybe caused OOM.

jaycarlton:
There are only 4 kotlin unit tests I'm aware of. They shouldn't be any more memory intensive than java tests though.

They could go back to Java if it's a major problem. Either way maybe file a bug.

aweng98:
4 `.kt` test suites, which has 13 tests total.

.circleci/config.yml
```diff
@@ -298,18 +303,36 @@ jobs:
...
+            if [ "$CIRCLE_NODE_INDEX" == 0 ]; then
+              ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+            fi
       - run:
           name: Run Java unit tests
           working_directory: ~/workbench/api
-          command: ./project.rb test
+          # See: https://circleci.com/docs/2.0/language-java/#sample-configuration
+          command: |
+            CLASSNAMES=$(circleci tests glob "src/test/java/**/*Test.java" "src/test/java/**/*Test.kt" \
```
jaycarlton, line 315: Can we write this in a higher-level language like Ruby or Python and just deploy it to the circle box and call it here? 

Otherewise, can you please put comments describing the algorithm here?

.circleci/config.yml
```diff
@@ -393,7 +417,7 @@ jobs:
   api-bigquery-test:
     # We're hitting OOM failures as of 13 May 2020.  Increase to 6GB RAM by specifying a medium+ machine.
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
-    resource_class: medium+
+    # resource_class: medium+
```
jaycarlton, line 420: I'd adjust this separately from the main PR(s).

.circleci/config.yml
jaycarlton: are you saying one job overrides the other's cache? That seems odd.

api/src/main/java/org/pmiops/workbench/profile/AddressMapper.java
```diff
@@ -11,6 +11,6 @@
   Address dbAddressToAddress(DbAddress dbAddress);
 
   @Mapping(target = "id", ignore = true)
-  @Mapping(target = "user", ignore = true) // set by UserService.createUser
+  @Mapping(target = "user", ignore = true) // set by UserService.createUser.
```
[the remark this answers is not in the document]
aweng98: yes, to run the api-unit-test job. I will undo change before merge.

.circleci/config.yml
aweng98: no, not saying about cache override. 
cache save only happens once if none exists and it happens on whatever the job that finishes first. deploy-to-* jobs requires other jobs run and pass first, so cache save will always be certain not happening here.

.circleci/config.yml
```diff
@@ -107,7 +107,9 @@ commands:
           command: |
             if [ ${CIRCLE_BRANCH} != "" ] &&
               [ ${CIRCLE_BRANCH} != "master" ] &&
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
+              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) \
+                | grep -qvE '(.md)|(.pdf)|(.png)|(.svg)|(Test.java)|(Test.kt)|(.mp4)|(.log)|(.txt)'; then
```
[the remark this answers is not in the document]
aweng98: puppeteer job calls this. It doesn't make sense to me to run puppeteer tests if commits include only tests. Since there is no product code change, I think running puppeteer tests would be waste of time and resources.

.circleci/config.yml
```diff
@@ -30,8 +30,8 @@ anchors:
     # In Feb 2017, this was set to 3G. But in Feb 2019 (RW-2194) we started seeing OOM errors,
     # so we bumped this down further to 2G.
     JAVA_TOOL_OPTIONS: -Xmx2g
-    # https://docs.gradle.org/6.3/userguide/gradle_daemon.html#sec:disabling_the_daemon
-    GRADLE_OPTS: "-Dorg.gradle.daemon=false"
+    # See: https://support.circleci.com/hc/en-us/articles/360021812453-Common-Android-memory-issues
+    GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process
```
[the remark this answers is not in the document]
aweng98: PR started small for parallelism only. You raised questions and I tried to help to address few. :)

.circleci/config.yml
```diff
@@ -298,18 +303,36 @@ jobs:
...
           working_directory: ~/workbench/api
-          command: ./gradlew spotlessCheck
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" == 0 ]; then
+              ./gradlew spotlessCheck
+            fi
       - store_test_results:
-           path: ~/workbench/api/build/test-results/test
+          path: ~/workbench/api/build/test-results/test
+      - store_artifacts:
```
jaycarlton, line 333: comment here: artifacts give circle the timing info on tests.

PR #3645, opened 2020-06-04
[RW-4947][risk=no] Refactor standard number inputs to use NumberInput component

dolbeew:
- Fix NumberInput component to return input of type number instead of text
- Replace all instances of standard number inputs to use NumberInput

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/inputs.tsx
+++ b/ui/src/app/components/inputs.tsx
@@ -65,6 +65,17 @@ export const styles = {
     verticalAlign: 'middle',
   },
 
+  inputStyle: {
+    backgroundColor: colors.white,
+    borderColor: inputBorderColor,
+    borderRadius: 3,
+    borderStyle: 'solid',
+    borderWidth: 1,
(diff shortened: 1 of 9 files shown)


ui/src/app/components/inputs.tsx
```diff
@@ -220,15 +220,15 @@ export const FormValidationErrorMessage = withStyle({
   fontSize: 12
 })('div');
 
-export const TextInput = React.forwardRef(({style = {}, onChange, onBlur, ...props}:
-      {style?: React.CSSProperties, onChange: Function, onBlur: Function, [key: string]: any},
+export const TextInput = React.forwardRef(({style = {}, type = 'text', onChange, onBlur, ...props}:
```
NehaBroad, line 223: Should textInput be used for anything else except type text

dolbeew: Good point. I refactored to use a regular number input instead of TextInput.

PR #3646, opened 2020-06-05
[RW-4918][RW-4767][risk=no] Validate against whitespace-only strings on workspace create/edit/clone

calbach:
Notes:
- Refactored validate.js usage to take better advantage built-in capabilities, including `allowEmpty: false`
- Switched to a conditional validation approach, rather than including this in the validation check
- Does not implement any API-level check for non-whitespace strings, this is probably something we should have but is out of scope.

Future work:
- We should rework the error messages to utilize the validate.js error messaging framework. Currently there's a fair amount of duplication going on in this file.

### DIFF
--- a/ui/src/app/pages/workspace/workspace-edit.spec.tsx
+++ b/ui/src/app/pages/workspace/workspace-edit.spec.tsx
@@ -393,4 +393,16 @@ describe('WorkspaceEdit', () => {
 
     expect(getSaveButtonDisableMsg(wrapper, 'otherPopulationDetails')).toBeDefined();
   });
+
+  it ('should show error message when other disseminate checked but empty', async() => {
+    const wrapper = component();
+    wrapper.find('[data-test-id="OTHER-checkbox"]').at(1).simulate('change', { target: { checked: true } });
+    const validInput = fp.repeat(8, 'a');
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/workspace/workspace-edit.spec.tsx
```diff
@@ -392,13 +381,23 @@ describe('WorkspaceEdit', () => {
...
     wrapper.find('[data-test-id="other-specialPopulation-text"]').first().simulate('change', {target: {value: inValidInput}});
 
-    otherSpecificPopulationError = getSaveButtonDisableMsg(wrapper, 'otherSpecificPopulation');
-    expect(fp.first(otherSpecificPopulationError)).toBe('Other specific population must be true');
+    expect(getSaveButtonDisableMsg(wrapper, 'otherPopulationDetails')).toBeDefined();
+  });
+
+  it ('should show error message when other disseminate checked but empty', async() => {
+    const wrapper = component();
+    wrapper.find('[data-test-id="OTHER-checkbox"]').at(1).simulate('change', { target: { checked: true } });
```
als364, line 394: I wish our data-test-ids were a bit more consistent in casing/style, but that's a problem for another PR

PR #3654, opened 2020-06-09
run java-unit-tests in parallel in CircleCI

aweng98:
Rename `java-build-test` job to `java-unit-test` for clear identification and run all tests in parallel on 4 CircleCI nodes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,7 +284,8 @@ commands:
 #        JOBS
 # -------------------------
 jobs:
-  api-build-test:
+  api-unit-test:
+    parallelism: 4
     # We're hitting OOM failures as of 1 May 2020.  Increase to 6GB RAM by specifying a medium+ machine.
     # Refers to https://circleci.com/docs/2.0/configuration-reference/#resource_class
     resource_class: medium+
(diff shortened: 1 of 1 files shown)


aweng98:
Break a large PR (https://github.com/all-of-us/workbench/pull/3641) into smaller separate PRs. This PR enable parallelism for Java unit tests execution in CircleCI.

.circleci/config.yml
```diff
@@ -298,18 +299,43 @@ jobs:
       - run:
           name: Validate Swagger definitions
           working_directory: ~/workbench/api
-          command: ./project.rb validate-swagger --project-prop verboseTestLogging=yes
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" == 0 ]; then
+              ./project.rb validate-swagger --project-prop verboseTestLogging=yes
```
jaycarlton, line 304: FYI, if you're testing this, note that this command isn't doing anything right now. https://precisionmedicineinitiative.atlassian.net/browse/RW-5073. No change is needed here, though.

.circleci/config.yml
```diff
@@ -298,18 +299,43 @@ jobs:
...
+          #   Append "--tests" in front of file name.
+          command: |
+            CLASSNAMES=$(circleci tests glob "src/test/java/**/*Test.java" "src/test/java/**/*Test.kt" \
+              | cut -c 1- \
+              | sed 's@src/test/java/@@' \
+              | sed 's@/@.@g' \
+              | sed 's/\.[^.]*$//' \
+              | circleci tests split --split-by=timings --timings-type=classname --index=$CIRCLE_NODE_INDEX)
+            GRADLE_ARGS=$(echo $CLASSNAMES | awk '{for (i=1; i<=NF; i++) print "--tests", $i}')
+            java -version && gradle -v && gradle test $GRADLE_ARGS
```
jaycarlton, line 324: nit: Can you move the java & gradle version checks to their own step if they're necessary? I guess I was thinking that's specified by the image, so maybe we don't need it here too?

.circleci/config.yml
```diff
@@ -561,7 +587,7 @@ workflows:
       # Always run basic test/lint/compilation (open PRs, master merge).
       # Note: by default tags are not picked up.
       - api-local-test
-      - api-build-test
+      - api-unit-test
       - ui-build-test
```
jaycarlton, line 591: nit: is there a better name for `ui-build-test`?

.circleci/config.yml
```diff
@@ -298,18 +299,43 @@ jobs:
...
           working_directory: ~/workbench/api
-          command: ./project.rb test
+          # See: https://circleci.com/docs/2.0/language-java/#sample-configuration
+          # Short explanation on commands:
+          #   Find all unit tests recursively in "api/src/test/java/" directories.
+          #   Strips off "src/test/java/" substring;
+          #   Replaces all path separator "/" with ".";
+          #   Removes file extensions.
+          #   Append "--tests" in front of file name.
+          command: |
```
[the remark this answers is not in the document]
aweng98: Done.

.circleci/config.yml
aweng98: removed `java -version && gradle -v` substring. I thought it would be good to have for info, but you're right, it is already in docker image. No need to show here.

.circleci/config.yml
aweng98: 🤔 `ui-bvt`? At IBM, `BVT` is universally accepted short name for "Build Verification Tests". But word of caution, I don't find any team in Broad use it.

.circleci/config.yml
aweng98: good to know.

.circleci/config.yml
jaycarlton: I think build is implied. I don't know of any non-build tests we'd be doing here (except linting).

What do you call this suite? Maybe `angular-test`?

calbach: It checks for both compilation and runs the unit tests. Full compilation is not required to run the unit tests, so I don't agree that it's implicit, though I'd be fine with just ui-test. angular-test would be inaccurate.

calbach: I see the other was renamed to unit. That would be fine here as well, `ui-unit-test`

PR #3656, opened 2020-06-09
Addition of .gradle directory to store in api gradle cache

aweng98:
Break up a large PR (#3641) into smaller separate PRs. This PR enable updates api gradle cache and its version in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,7 +158,7 @@ commands:
           name: "restore e2e cache"
           keys:
             - v1-e2e-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
-            - v1-e2e-cache-{{ .Branch }}
+            - v1-e2e-cache-master
             - v1-e2e-cache-
       - run:
           name: "workbench/e2e: yarn install dependencies"
@@ -199,7 +199,7 @@ commands:
                 name: "restore api gradle cache"
                 keys:
                   - v2-gradle-cache-{{ .Branch }}-{{ checksum "~/workbench/api/build.gradle" }}
-                  - v2-gradle-cache-{{ .Branch }}
+                  - v2-gradle-cache-master
                   - v2-gradle-cache-
 
   run_e2e_test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -158,7 +158,7 @@ commands:
           name: "restore e2e cache"
           keys:
             - v1-e2e-cache-{{ .Branch }}-{{ checksum "~/workbench/e2e/yarn.lock" }}
-            - v1-e2e-cache-master
+            - v1-e2e-cache-{{ .Branch }}
```
calbach, line 161: Why this change? I think I prefer master here. As I understand it, using the branch name here only helps in the following scenario:

- I'm working on a feature branch
- I push one set of changes to yarn.lock
- On that same feature branch, I push a different set of changes to yarn.lock

In this scenario, unless you are making a ton of changes to the lockfile, you're probably getting the same performance as just restoring from the master cache.

Alternatively, if I'm working on a new feature branch and change yarn.lock (more likely), my cache hit may be worse in this model, since I'd pull in any recent cache entry, possibly from an unrelated feature branch.

.circleci/config.yml
```diff
@@ -186,18 +186,21 @@ commands:
           steps:
             - save_cache:
                 name: "save api gradle cache"
+                # See https://docs.gradle.org/current/userguide/directory_layout.html#dir:project_root
+                # Save Project-specific cache and Global cache (for everything that’s not project-specific).
                 paths:
+                  - .gradle
```
calbach, line 192: Good find. Does this help? Did you confirm there are contents in this directory by SSHing into the Circle VM?

aweng98: I did `ls -al` in `api` dir. cmd outputs confirms `.gradle` dir exists after gradle job.

.circleci/config.yml
aweng98: yeah you're right. I didn't think about that. Reverted change. Also, I realized same mistake I made in e2e cache.

aweng98: I spoke too soon about e2e cache mistake. I had `v1-e2e-cache-master`, nothing to fix.

PR #3658, opened 2020-06-10
[risk=no] Refactor docker-compose to leverage anchors

calbach:
Behavioral change:
- In a few cases, this adds the GOOGLE_APPLICATION_CREDENTIALS. Based on manual testing of commands, this looks like it had no effect, so I'll prefer just making the configs consistent.
- Remove some unused services: mysql-cloud, cloud-sql-proxy. @freemabd in case you're aware of some reason we still need these

Manually tested:
- Ran `./project.rb` commands with coverage of the various docker-compose services

Future work:
- Several of these services are redundant and could be further merged / removed

### DIFF
--- a/api/docker-compose.yaml
+++ b/api/docker-compose.yaml
@@ -1,16 +1,22 @@
-version: "3"
+version: "3.4"
+x-api-defaults: &api-defaults
+  build:
+    context: ./src/dev/server
+  image: allofustest/workbench-dev-api:local
+  user: ${UID}
+  working_dir: /w/api
+  environment:
+    - GOOGLE_APPLICATION_CREDENTIALS=/w/api/sa-key.json
+  env_file:
+    - db/vars.env
+  volumes:
+    - src-sync:/w:nocopy
+    - gradle-cache:/.gradle
+    - ~/.config:/.config:cached
+    - ~/.gsutil:/.gsutil:cached
 services:
   scripts:
-    build:
-      context: ./src/dev/server
-    image: allofustest/workbench-dev-api:local
-    user: ${UID}
-    working_dir: /w/api
-    volumes:
-      - src-sync:/w:nocopy
-      - gradle-cache:/.gradle
-      - ~/.config:/.config:cached
-      - ~/.gsutil:/.gsutil:cached
+    <<: *api-defaults
   db:
     image: mysql:5.7.27
     env_file:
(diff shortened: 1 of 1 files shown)


api/docker-compose.yaml
```diff
@@ -20,165 +26,52 @@ services:
...
       - db
-    build:
-      context: ./src/dev/server
-    image: allofustest/workbench-dev-api:local
-    user: ${UID}
     working_dir: /w/api/db
-    volumes:
-      - src-sync:/w:nocopy
-      - gradle-cache:/.gradle
     entrypoint: ['with-uid.sh', 'wait-for-it', 'db:3306', '-t', '30', --]
```
jaycarlton, line 43: Maybe parameterize the port number and timeout?

api/docker-compose.yaml
```diff
@@ -20,165 +26,52 @@ services:
...
-      - gradle-cache:/.gradle
-      - ~/.config:/.config:cached
-      - ~/.gsutil:/.gsutil:cached
     command: ./project.rb start-api-and-incremental-build
-    environment:
-      - GOOGLE_APPLICATION_CREDENTIALS=/w/api/sa-key.json
-    env_file:
-      - db/vars.env
     ports:
       - 127.0.0.1:8081:8081
```
jaycarlton, line 34: Maybe make a named constant for the two different port numbers 8081 and 8001. If order is arbitrary, maybe list them alphabetically.

api/docker-compose.yaml
```diff
@@ -20,165 +26,52 @@ services:
     ports:
       - 127.0.0.1:3306:3306
   api:
+    <<: *api-defaults
```
jaycarlton, line 29: I'm a bit surprised you need to use an anchor. I would've expected `docker-compose` to have a high-level concept for this kind of reuse.

api/docker-compose.yaml
```diff
@@ -20,165 +26,52 @@ services:
...
-      context: ./src/dev/server
-    image: allofustest/workbench-dev-api:local
-    user: ${UID}
-    working_dir: /w
-    volumes:
-      - .:/w:cached
-      - ~/.config:/.config:cached
-      - ~/.gsutil:/.gsutil:cached
-    command: |
-      cloud_sql_proxy
```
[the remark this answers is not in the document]
freemabd: I also use cloud_sql_proxy to connect to test mysql db when using sequelPro

api/docker-compose.yaml
calbach: Looks like the way to do this in docker-compose would be to push the ports into a `.env` file. local constants/variables don't seem to be possible in docker-compose, except maybe by further abusing anchors.

AFAICT the suggestion isn't really related to the PR, so I don't intend to make a non-trivial change like that here.

api/docker-compose.yaml
calbach: See above comment. Also the port is fine, but the timeout is a constant. Creating knobs for things that don't need to change generally just adds noise. If the intent was to avoid duplication with the other services here, see the PR description: my preference is to just eliminate the services with duplicate functionality. If there's a legitimate need for multiple services with this entrypoint, using an anchor can also work to avoid double-encoding of this.

api/docker-compose.yaml
```diff
@@ -1,16 +1,22 @@
...
-    image: allofustest/workbench-dev-api:local
-    user: ${UID}
-    working_dir: /w/api
-    volumes:
-      - src-sync:/w:nocopy
-      - gradle-cache:/.gradle
-      - ~/.config:/.config:cached
-      - ~/.gsutil:/.gsutil:cached
+    <<: *api-defaults
   db:
```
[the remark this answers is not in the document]
calbach: There's only one database here, and it feels like a pretty big stretch to think this could be referring to BigQuery. Is there a specific use case you are concerned about? Just reading this file?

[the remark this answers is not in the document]
jaycarlton: Fair enough. Just seems like a database should have a name other than database. The confusion isn't so much with this file or even with developers, but when discussing application components with product folks or other stakeholders.

calbach:
Puppeteer e2e is broken due to https://precisionmedicineinitiative.atlassian.net/browse/RW-5080 . Merging

PR #3659, opened 2020-06-10
[risk=low] Upgradle to 4.10.3

calbach:
- Bump local dev image from 4.3.1 -> 4.10.3
- Bump Circle image from 4.10.2 -> 4.10.3, these shouldn't have diverged (I probably was responsible)

I want this upgrade as part of [RW-1794](https://precisionmedicineinitiative.atlassian.net/browse/RW-1794). 4.3.1 has some undesirable behavior w.r.t. daemons/locking which seem to be addressed in later versions.

This is a stacked PR against https://github.com/all-of-us/workbench/pull/3658 , I will merge that one first - then rebase this on master.

### DIFF
--- a/api/docker-compose.yaml
+++ b/api/docker-compose.yaml
@@ -10,7 +10,7 @@ x-api-defaults: &api-defaults
   # it from the provided build context or use a cached local version.
   # When making changes to this image, you can modify this tag to force all devs
   # to rebuild.
-  image: allofustest/workbench-dev-api:local
+  image: allofustest/workbench-dev-api:local-1
   build:
     context: ./src/dev/server
   user: ${UID}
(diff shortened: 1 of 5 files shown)


api/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: "3.4"
 x-api-defaults: &api-defaults
   build:
     context: ./src/dev/server
-  image: allofustest/workbench-dev-api:local
+  image: allofustest/workbench-dev-api:local-1
```
jmthibault79, line 5: Does this image exist on Dockerhub?  I don't have access to it.

calbach: It doesn't - I just added a comment above this line in my other PR which you should now be able to see here. Please take a read and let me know if this addresses your question.

calbach:
puppeteer-e2e-local-ui still broken due to FC dev project limits. Merging

PR #3660, opened 2020-06-11
[RW-5080][risk=no] Temporarily turn off FC dev project buffering

calbach:
**Please approve/merge in the morning before merging any other PRs**

I manually applied this for now, but that will be overwritten with the next PR merge.

Confirmed this had the intended effect:
image

### DIFF
--- a/api/config/config_test.json
+++ b/api/config/config_test.json
@@ -20,7 +20,7 @@
     "projectNamePrefix": "aou-rw-test-",
     "exportBigQueryTable": "all-of-us-workbench-test-bd.billing_data.rw_test_firecloud_view",
     "retryCount": 2,
-    "bufferCapacity": 100,
+    "bufferCapacity": 0,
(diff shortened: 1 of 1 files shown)


jmthibault79:
Correct - no change to existing projects.  I would also have been OK with merge-yourself in this case.

calbach:
@jaycarlton as I mentioned in the PR description:
- I already applied it to test manually, hence why I indicated this should be merged before any other PRs so my change wasn't overwritten.
- Yes, we do have a process for that and it's been tried. Looks like it may not be documented in the playbook though. I will make a note to add this.

There's nothing to purge, the buffer is entry which is how I noticed the issue.

calbach:
nvm, we had a playbook entry. I improved it a bit: https://docs.google.com/document/d/1hbYNWzdxuRybTfW3M70hWZfr_TXRVd5rOMR9kZtwAFY/edit#heading=h.dmqxllxdcmlq

PR #3663, opened 2020-06-11
[RW-4383][risk=no] Puppeteer DataSet test

aweng98:
New Puppeteer tests that covers following features:
* Create DataSet without Export to Notebook
* Export dataset to notebook in Python language
* Edit a DataSet
* Rename a DataSet
* Delete a DataSet
* Create new cohort

Additional changes in some existing files were required in order to support writing of new tests.

Screen Shot 2020-06-12 at 6 13 26 PM

Screen Shot 2020-06-12 at 6 15 21 PM

### DIFF
--- a/e2e/app/element/base-element.ts
+++ b/e2e/app/element/base-element.ts
@@ -234,8 +234,8 @@ export default class BaseElement extends Container {
    */
   async clickAndWait(): Promise<void> {
     await Promise.all([
-      this.page.waitForNavigation({waitUntil: ['domcontentloaded', 'networkidle0']}),
       this.click(),
+      this.page.waitForNavigation({waitUntil: ['load', 'domcontentloaded', 'networkidle0']}),
     ]);
   }
 
--- a/e2e/app/element/base-element.ts
+++ b/e2e/app/element/base-element.ts
@@ -13,7 +13,7 @@ export default class BaseElement extends Container {
     return baseElement;
   }
 
-  protected element: ElementHandle;
+  private element: ElementHandle;
 
   constructor(protected readonly page: Page, xpath?: string) {
     super(page, xpath);
@@ -47,7 +47,7 @@ export default class BaseElement extends Container {
    * @param {string} descendantXpath Be sure to begin xpath with a dot. e.g. ".//div".
    */
   async findDescendant(descendantXpath: string): Promise<ElementHandle[]> {
-    return this.element.$x(descendantXpath);
+    return (await this.asElementHandle()).$x(descendantXpath);
   }
 
   /**
@@ -60,10 +60,7 @@ export default class BaseElement extends Container {
    *  return await handle.jsonValue();
    */
   async getProperty(propertyName: string): Promise<unknown> {
-    if (this.element == null) {
-      throw new Error('The element is undefined.');
-    }
-    const p = await this.element.asElement().getProperty(propertyName);
+    const p = await (await this.asElementHandle()).getProperty(propertyName);
     return p.jsonValue();
   }
 
@@ -72,10 +69,7 @@ export default class BaseElement extends Container {
    * @param attribute name
    */
   async getAttribute(attributeName: string): Promise<string | null> {
-    if (this.element == null) {
-      throw new Error('The element is undefined.');
-    }
-    const elem = this.element.asElement();
+    const elem = await this.asElementHandle();
     const attributeValue = await this.page.evaluate(
        (link, attr) => link.getAttribute(attr), elem, attributeName);
     return attributeValue;
@@ -87,9 +81,6 @@ export default class BaseElement extends Container {
    * @param attribute name
    */
   async hasAttribute(attributeName: string): Promise<boolean> {
-    if (this.element == null) {
-      throw new Error('The element is undefined.');
-    }
     const value = await this.getAttribute(attributeName);
     return value !== null;
   }
@@ -111,7 +102,7 @@ export default class BaseElement extends Container {
    * @param {ElementHandle} element
    */
   async isVisible(): Promise<boolean> {
-    const boxModel = await this.element.boxModel();
+    const boxModel = await (await this.asElementHandle()).boxModel();
     return boxModel !== null;
   }
 
@@ -119,14 +110,15 @@ export default class BaseElement extends Container {
    * Check both boxModel and style for visibility.
    */
   async isDisplayed() {
+    const elemt = await this.asElementHandle();
     const isVisibleHandle = await this.page.evaluateHandle((e) =>
     {
       const style = window.getComputedStyle(e);
       return (style && style.display !== 'none' &&
          style.visibility !== 'hidden' && style.opacity !== '0');
-    }, this.element);
+    }, elemt);
     const jValue = await isVisibleHandle.jsonValue();
-    const boxModelValue = await this.element.boxModel();
+    const boxModelValue = await elemt.boxModel();
     if (jValue && boxModelValue) {
       return true;
     }
@@ -134,17 +126,17 @@ export default class BaseElement extends Container {
   }
 
   async click(options?: ClickOptions): Promise<void> {
-    return this.element.asElement().click(options);
+    return (await this.asElementHandle()).click(options);
   }
 
   async type(text: string, options?: { delay: number }): Promise<void> {
     await this.focus();
     await this.clear();
-    return this.element.asElement().type(text, options);
+    return (await this.asElementHandle()).type(text, options);
   }
 
   async pressKeyboard(key: string, options?: { text?: string, delay?: number }): Promise<void> {
-    return this.element.asElement().press(key, options);
+    return (await this.asElementHandle()).press(key, options);
   }
 
   async pressReturnKey(): Promise<void> {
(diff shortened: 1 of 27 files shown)


jaycarlton:
We started a campaign to correct the spelling of `DataSet` to `Dataset`, but it's only a third of the way through.

e2e/app/element/base-element.ts
```diff
@@ -60,10 +60,7 @@ export default class BaseElement extends Container {
    *  return await handle.jsonValue();
    */
   async getProperty(propertyName: string): Promise<unknown> {
-    if (this.element == null) {
-      throw new Error('The element is undefined.');
-    }
-    const p = await this.element.asElement().getProperty(propertyName);
+    const p = await (await this.asElementHandle()).getProperty(propertyName);
```
calbach, line 63: In the future, if large refactoring changes like this can be split out from the PR, that would be very helpful for giving quicker review turnaround. To review 500+ LOC all at once, I need to set aside a significant amount of time. In comparison, ~400 lines of mechanical refactoring changes alone is easy - and 100 lines of new tests is easy.

e2e/app/page/data-page.ts
```diff
@@ -50,30 +56,91 @@ export default class DataPage extends AuthenticatedPage {
...
+    const newNameTextbox = new Textbox(this.page, '//*[@id="new-name"]');
+    await newNameTextbox.type(newDatasetName);
+
+    const descriptionTextarea = await Textarea.findByName(this.page, {containsText: 'Description:'}, dialog);
+    await descriptionTextarea.type('Puppeteer automation rename dataset.');
+
+    const renameButton = await Button.findByName(this.page, {normalizeSpace: 'Rename Dataset'}, dialog);
+    await Promise.all([
+      renameButton.click(),
+      dialog.waitUntilDialogIsClosed(120000), // Rename Dataset can be very slow sometimes.
```
calbach, line 141: That doesn't make sense - I suspect something else was going on when you observed this. Is this reproducible?

e2e/app/page/dataset-build-page.ts
```diff
@@ -0,0 +1,99 @@
...
+  async selectConceptSets(conceptSetNames: string[]): Promise<void> {
+    for (const conceptSetName of conceptSetNames) {
+      const xpath = this.getCheckboxXpath(LabelAlias.SelectConceptSets, conceptSetName);
+      const checkbox = new Checkbox(this.page, xpath);
+      await checkbox.check();
+      await waitWhileLoading(this.page);
+    }
+  }
+
+  async selectValues(selectAll: boolean = false, values?: string[]): Promise<void> {
```
calbach, line 63: nit: I don't think the boolean type should be necessary in this situation; type I believe should just be implicit based ont he default value

calbach, line 63: It seems like this should really be two different methods. Overloading both capabilities here is making the interaction a bit confusing in my opinion

e2e/app/page/dataset-save-modal.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(page: Page) {
+    super(page);
+  }
+
+  /**
+   * Handle Save or Update dialog.
+   * @param {boolean} isExportToNotebook If True, select Export To Notebook checkbox.
+   * @param {boolean} isUpdate If true, click Update button. If False, click Save button.
+   * @param {string} notebookName New notebook name to be created or select an existing notebook.
+   * @param {String} language Notebook programming language.
```
calbach, line 19: These should probably be constants

e2e/app/page/dataset-save-modal.ts
```diff
@@ -0,0 +1,50 @@
...
+                       {isExportToNotebook?: boolean, isUpdate?: boolean, notebookName?: string, language?: string} = {}): Promise<string> {
+    const {isExportToNotebook = false, isUpdate = false, notebookName, language = 'Python'} = opts;
+    const newDatasetName = makeRandomName();
+
+    const nameTextbox = await this.waitForTextbox('Dataset Name');
+    await nameTextbox.type(newDatasetName);
+
+    // Export to Notebook checkbox is checked by default
+    const exportCheckbox = await this.waitForCheckbox('Export to notebook');
+      // Not Export to Notebook
```
calbach, line 31: This comment seems to be the opposite of what the following code is doing

e2e/app/page/workspaces-page.ts
```diff
@@ -88,8 +88,8 @@ export default class WorkspacesPage extends WorkspaceEditPage {
     await (await editPage.getWorkspaceNameTextbox()).type(workspaceName);
     await (await editPage.getWorkspaceNameTextbox()).tabKey();
 
-    // select Synthetic Data Set 3
-    await editPage.selectDataSet('3');
+    // select the default Synthetic Dataset
```
calbach, line 91: The default will already be selected. If you intend to test the default behavior, I would not touch the selector

e2e/tests/datasets/export-to-notebook.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+   /**
+    * Create new Dataset, export to notebook in Python language
+    * Finally delete Dataset.
+    */
+  test('Export dataset to notebook in Python programming language', async () => {
+    const workspaceCard = await findWorkspace(page);
+    await workspaceCard.clickWorkspaceName();
+
+      // Click Add Datasets button
```
calbach, line 21: nit: fix indentation for all comments in this file. These shouldn't be indented deeper than the code

e2e/tests/datasets/dataset-create.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    const addIcon = await ClrIconLink.findByName(page, {containsText: nameValue, iconShape: 'plus-circle'}, modal);
+    await addIcon.click();
+
+    // Click FINISH button. Criteria dialog closes.
+    await modal.clickButton(ButtonLabel.Finish);
+    await cohortBuildPage.getTotalCount();
+
+    // Save new cohort.
+    const cohortName = await cohortBuildPage.saveCohortAs();
+    await waitForText(page, 'Cohort Saved Successfully');
```
calbach, line 89: Side-note, not for this PR: we should really make this case insensitive at least, if we're depending on text like this

e2e/tests/datasets/export-to-notebook.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      // Verify Notebook preview. Not going to start the Jupyter notebook.
+    const notebookPreviewPage = new NotebookPreviewPage(page);
+    await notebookPreviewPage.waitForLoad();
+    const currentPageUrl = await page.url();
+    expect(currentPageUrl).toContain(`notebooks/preview/${newNotebookName}.ipynb`);
+
+    const code = await notebookPreviewPage.getFormattedCode();
+    expect(code).toContain('import pandas');
+    expect(code).toContain('import os');
+
```
calbach, line 49: Test description says to delete the dataset - I don't see that part

aweng98: Missed it. Added delete Dataset and Notebook.

e2e/app/page/dataset-build-page.ts
```diff
@@ -0,0 +1,106 @@
...
+      await checkbox.check();
+      await waitWhileLoading(this.page);
+    }
+  }
+
+  /**
+   * Check or uncheck the Select All checkbox.
+   * @param {boolean} selectAll
+   */
+  async selectAllValues(selectAll: boolean = false): Promise<void> {
```
calbach, line 67: nit: I would expect the default to be true, given the method name. Alternatively, I would probably just remove the default here.

e2e/app/page/dataset-build-page.ts
```diff
@@ -0,0 +1,106 @@
...
+    const xpath = xPathOptionToXpath({name: LabelAlias.SelectValues, ancestorLevel: 3, type: ElementType.Checkbox});
+    const selectAllCheckbox = new Checkbox(this.page, xpath);
+    selectAll ? await selectAllCheckbox.check() : await selectAllCheckbox.unCheck();
+    await waitWhileLoading(this.page);
+  }
+
+  async selectValues(values?: string[]): Promise<void> {
```
calbach, line 74: nit: probably make the parameter non-optional now, I think the below code breaks if it's null/undefind

e2e/jest.test-setup.ts
```diff
@@ -27,6 +27,7 @@ beforeAll(async () => {
        || host === 'accounts.youtube.com'
        || host === 'static.zdassets.com'
        || host === 'play.google.com'
+       || host === 'static.zdassets.com'
```
calbach, line 30: This is a repeat of line 28

e2e/app/page/dataset-build-page.ts
aweng98: right, default set to `true`.

e2e/app/page/dataset-build-page.ts
aweng98: right, param is set as non-optional.

e2e/jest.test-setup.ts
aweng98: 🙂 removed.

PR #3665, opened 2020-06-12
[RISK=NO][RW-5065] Migrate additional route and add a guard

petesantos:
Description:

I migrated the data user code of conduct route over to react router.

1. Removed the existing route from the Angular router
2. Added a sign in guard - at the moment both sign in guards will get called, further work on teasing out the guards and stores will be needed
3. Added the concept of "atoms" to the code base - basically a subscribe-able container
   1. This allows us to iteratively migrate away from the Angular services
   2. This could be a pattern that replaces some of the current angular services / stores
4. Added an `authStore` and `routeDataStore` - a couple of barebones stores to be able to check the auth status and set the page title/pass data to the route respectively

---
**PR checklist**

(Failing tests also failed on master)


ui/src/app/utils/stores.tsx
```diff
@@ -0,0 +1,23 @@
+import {useEffect, useState} from 'react';
+import {atom, Atom} from './subscribable';
+
+interface RouteDataStore {
+  title?: string;
+}
+
+export const routeDataStore = atom <RouteDataStore>({});
```
calbach, line 8: nit: rm space?

ui/src/app/utils/stores.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface AuthStore {
+  authLoaded: boolean;
+  isSignedIn: boolean;
+}
+
+export const authStore = atom<AuthStore>({authLoaded: false, isSignedIn: false});
+
+export function useStore<T>(theStore: Atom<T>) {
+  const [value, setValue] = useState(theStore.get());
```
calbach, line 18: I think it makes sense and I like the symmetry of `useStore`, but seeing `useState` in a helper function is hurting by brain a bit. We can discuss this offline if it's faster, but:

Do you happen to know the mechanism by which React associates this global call to a component? Is it just tracking a stack of render calls - and binds any `useState` call to the current component in that stack? Or setting aside the implementation details, is that the right mental model?

I would probably add a method doc to this one to explain the intended usage, i.e. that it should only be called from a react function component (I believe)

ui/src/app/utils/stores.tsx
```diff
@@ -0,0 +1,23 @@
...
+  useEffect(() => {
+    return theStore.subscribe(v => setValue(v)).unsubscribe;
+  }, [theStore, setValue]);
```
calbach, line 21: I looked at the useEffect docs. This line is still does not compute.

`theStore` is not a property or state, AFAICT. What does it mean to filter `useEffect` to only run when it "changes"? setValue might make some sense(?); my best guess is that it basically daisy-chains the above subscription setup to itself.

Can you add a comment, or if there's a more general concept I'm missing - point me to the right docs?

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,43 @@
+import * as fp from 'lodash/fp';
+
+export interface Subscribable<T> {
```
calbach, line 3: Could you add some basic docs to these interfaces? Method-level docs are probably not needed given the simplicity

ui/src/app/components/app-router.tsx
```diff
@@ -1,26 +1,43 @@
...
 export const AppRouter = ({children}): React.ReactElement => <BrowserRouter><SubRoute>{children}</SubRoute></BrowserRouter>;
 
 export const RouteLink = ({path, style = {}, children}): React.ReactElement => <Link style={{...style}} to={path}>{children}</Link>;
 
+// To compensate for Angular, while keeping true to the declarative/compnentized nature of the router
+// We will utilize a redirect component that uses the Angular navigation.
+// Upon completing the migration this can be replaced with a react-router Redirect component
+const NavRedirect = ({path}) => {
+  navigate([path]);
+  return <div/>;
```
calbach, line 38: nit: Do you need the div? Does Fragment not work here?

ui/src/app/components/app-router.tsx
```diff
@@ -1,26 +1,43 @@
...
+export const withFullHeight = WrappedComponent => ({...props}) => {
+  return <div style={{height: '100%'}}><WrappedComponent {...props} /></div>;
+};
+
 export const SubRoute = ({children}): React.ReactElement => <Switch>{children}</Switch>;
 export const AppRouter = ({children}): React.ReactElement => <BrowserRouter><SubRoute>{children}</SubRoute></BrowserRouter>;
 
 export const RouteLink = ({path, style = {}, children}): React.ReactElement => <Link style={{...style}} to={path}>{children}</Link>;
 
+// To compensate for Angular, while keeping true to the declarative/compnentized nature of the router
```
calbach, line 33: Not a real word anyways, but spelling: componentized

ui/src/app/utils/stores.tsx
```diff
@@ -0,0 +1,23 @@
+import {useEffect, useState} from 'react';
+import {atom, Atom} from './subscribable';
+
+interface RouteDataStore {
```
calbach, line 4: This gets killed after migrating off Angular right? Maybe leave a comment here so we can remember what to rip out afterwards.

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return {
+    subscribe: fn => {
+      subscribers = fp.concat(subscribers, [fn]);
+      return {
+        unsubscribe: () => {
+          subscribers = fp.without([fn], subscribers);
+        }
+      };
+    },
+    next: (...args) => {
```
calbach, line 26: I'm surprised the type system is OK with this. Should this have the more constrained method signature found in the interface?

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  get: () => T;
+  set: (value: T) => void;
+  subscribe: Subscribable<T>['subscribe'];
+  reset: () => void;
+}
+
+export function subscribable<T>(): Subscribable<T> {
+  let subscribers = [];
+  return {
+    subscribe: fn => {
```
calbach, line 18: Is this based on something else? There's some interesting design considerations here. Is this tailored for a particular constrained use case - or if not, do you have rationale on the following?

- Whether or not to immediately deliver the most recent element to new subscribers. For something like an event system, this wouldn't make sense - but for a store, it can make the initial connection boilerplate a bit more homogeneous (one subscribe() path, rather than a get + subscribe).
- Synchronous vs asynchronous handling on `next()`, i.e. current approach vs invoking subscribers in a `setTimeout(, 0)` to pop this out of the stack. This provides some incapsulation between the subscribers, and avoids dependency on synchronous `next()` behaviors.

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,43 @@
...
+      subscribers = fp.concat(subscribers, [fn]);
+      return {
+        unsubscribe: () => {
+          subscribers = fp.without([fn], subscribers);
+        }
+      };
+    },
+    next: (...args) => {
+      fp.forEach(fn => fn(...args), subscribers);
```
calbach, line 27: This immediately throws if any of the subscribers throw. If you keep this synchronous per above, I would have some try/catch protection around each callback, since this is potentially a large list of mixed subscribers - an error in one should not be able to take down the whole subscribable.

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,43 @@
+import * as fp from 'lodash/fp';
+
```
calbach, line 2: Some simple unit tests around the Atom to codify some of its more nuanced behaviors could be nice: synchronous vs asynchronous behaviors, subscriptions, behavior on initial data seeding

ui/src/app/utils/stores.tsx
petesantos: On the `useState` inside of `useStore` - I believe that is the standard pattern if you want your custom hook to run the render method of your component. Calling hooks from within hooks is an acceptable practice according to the [rules of hooks](https://reactjs.org/docs/hooks-rules.html)(just to put your mind at ease). With some more exposure to hooks I think this will become clearer.

I haven't looked at the implementation of `useState`, but you suggested sounds correct and is the right mental model. `useState` is a functional approach to `setState`

I can add some docs to this. Ideally, any custom hook functions should be prefixed with "use" and there will be no functions in the codebase that are named use_Something_ that are not react hooks.  I thought about putting this in a separate file called "hooks" - we could then put all of our custom hooks in there. Would that be better/clearer than adding comments? I can still add comments describing what the hook does and the file could have a top level comment about how to use the hooks.

ui/src/app/utils/subscribable.tsx
petesantos: added typing

ui/src/app/utils/stores.tsx
calbach: Thanks, this makes more sense after reading the React Hooks docs through.

Let me check my understanding:

- `theStore` in all current use cases is not a prop or state, but this function should treat all incoming parameters as if they might be props
- Presumably the `useEffect` filter works on non-prop/state values. I'm imagining in this specific case where `theStore` is not a prop, `[theStore]` is equivalent to `[]` ?
- Accordingly, we expect this `useEffect()` to run once on initial mount, then never run again

ui/src/app/utils/subscribable.tsx
```diff
@@ -0,0 +1,58 @@
...
+      };
+    },
+    next: (newValue?: T, oldValue?: T) => {
+      fp.forEach((fn: subscriber) => setTimeout(() => fn(newValue, oldValue), 0), subscribers);
+    }
+  };
+}
+
+// A simple state container inspired by clojure atoms.
+export function atom<T>(initialValue: T): Atom<T> {
```
als364, line 48: I googled this pattern cause I hadn't heard it called 'atom' before. It looks like this is a clojurescript pattern, and that [someone ported it](https://github.com/derrickbeining/react-atom) to react. Do you think this port is worth using, or can our needs be served by something as simple as what you've implemented here?

ui/src/app/utils/subscribable.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+    testAtom.set({value: 10});
+
+    await complete1();
+    await complete2();
+    await complete3();
+  });
+
+  it('should not call an unsubscribed function', async() => {
+    const testAtom = atom({value: 1});
+    let value = 1;
```
calbach, line 79: nit: for the purposes of counting the invocations, I think starting with 0 would be more readable

ui/src/app/utils/subscribable.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+    testAtom.set({value: 10});
+
+    await complete1();
+    await complete3();
+    await delay(50);
+
+    expect(value).toBe(3);
+  });
+
+  it('should not call any functions when all have unsubscribed', async() => {
```
calbach, line 102: optional: one more test case which would be nice is to have one of the subscribers throw an error, and verify the others still succeed

ui/src/app/utils/subscribable.spec.tsx
petesantos: done

ui/src/app/utils/subscribable.spec.tsx
petesantos: added

ui/src/app/components/app-router.tsx
```diff
@@ -1,26 +1,43 @@
...
+};
+
 export const SubRoute = ({children}): React.ReactElement => <Switch>{children}</Switch>;
 export const AppRouter = ({children}): React.ReactElement => <BrowserRouter><SubRoute>{children}</SubRoute></BrowserRouter>;
 
 export const RouteLink = ({path, style = {}, children}): React.ReactElement => <Link style={{...style}} to={path}>{children}</Link>;
 
+// To compensate for Angular, while keeping true to the declarative/componentized nature of the router
+// We will utilize a redirect component that uses the Angular navigation.
+// Upon completing the migration this can be replaced with a react-router Redirect component
```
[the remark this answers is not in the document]
petesantos: done - thanks

ui/src/app/utils/stores.tsx
petesantos: Yes, useEffect will work on any value. There could be a case where there is a new value derived via a computation between state and props and only want to run the effect in that case. In the vast majority of cases this will run once. But in the event there are multiple stores of the same type, and a component swaps out the store in the hook, the proper cleanup (unsubscribe) and setup (subscribe) will take place for the newly designated store.

PR #3675, opened 2020-06-16
[no ticket][risk=no] Add support for plain-text request body in Swagger

gjuggler:
**Background**
While working on RW-4257 (integrating with a new Shibboleth service), I encountered a need to send a POST request whose message body is a primitive string.

I tried following the instructions in the Swagger 2.0 docs (https://swagger.io/docs/specification/2-0/describing-request-body/ under section "Primitive Body"). When I tried to execute a request using the generated API client, I would see an error: 'Content type "text/plain" is not supported'. Kt turns out this is a longstanding Swagger-codegen issue (https://github.com/swagger-api/swagger-codegen/issues/8288) which hasn't been fixed.

**Guide to reviewer**
This PR implements the proposed workaround (add a new 'else if' branch to support plain-text body). I tested the fix in the context of Shibboleth service integration, and this change allows a plain-text request body to be successfully sent to downstream services.

---
**PR checklist**

### DIFF
--- a/api/src/main/resources/swaggercodegen/libraries/okhttp-gson/ApiClient.mustache
+++ b/api/src/main/resources/swaggercodegen/libraries/okhttp-gson/ApiClient.mustache
@@ -1,0 +1,1372 @@
+{{!
+
+This is an AoU-specific extension of the default Swagger-codegen okhttp API implementation template.
+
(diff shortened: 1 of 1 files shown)


gjuggler:
Before merging, I confirmed that the diff between the new file and the original template (from https://github.com/swagger-api/swagger-codegen/blob/v2.2.3/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache) is what we expect.

PR #3677, opened 2020-06-16
[risk=medium][RW-4980] Admin Update Profile endpoint

jmthibault79:
Description:

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ProfileAuditorImpl.kt
+++ b/api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ProfileAuditorImpl.kt
@@ -64,7 +64,7 @@ constructor(
                             agentIdMaybe = userProvider.get().userId,
                             agentEmailMaybe = userProvider.get().username,
                             targetType = TargetType.PROFILE,
-                            targetIdMaybe = userProvider.get().userId,
+                            targetIdMaybe = previousProfile.userId,
                             targetPropertyMaybe = it.key,
                             previousValueMaybe = it.value.previousValue,
                             newValueMaybe = it.value.newValue)
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -541,7 +541,15 @@ public class ProfileController implements ProfileApiDelegate {
   }
 
   @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
-  public ResponseEntity<Void> adminUpdateProfile(DbUser user, Profile updatedProfile) {
+  @Override
+  public ResponseEntity<Profile> adminUpdateProfile(Profile updatedProfile) {
+    final String username = updatedProfile.getUsername();
+    final DbUser user =
+        userService
+            .getByUsername(username)
+            .orElseThrow(
+                () -> new NotFoundException("Could not find user account for " + username));
+
     // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
 
@@ -562,7 +570,7 @@ public class ProfileController implements ProfileApiDelegate {
     final Profile appliedUpdatedProfile = profileService.getProfile(user);
     profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
 
-    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+    return ResponseEntity.ok(appliedUpdatedProfile);
   }
 
   @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -540,6 +540,7 @@ public class ProfileController implements ProfileApiDelegate {
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
+  @AuthorityRequired(Authority.ACCESS_CONTROL_ADMIN)
   public ResponseEntity<Void> adminUpdateProfile(DbUser user, Profile updatedProfile) {
     // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -23,6 +23,7 @@ import org.pmiops.workbench.actionaudit.auditors.ProfileAuditor;
 import org.pmiops.workbench.annotations.AuthorityRequired;
 import org.pmiops.workbench.auth.UserAuthentication;
 import org.pmiops.workbench.auth.UserAuthentication.UserType;
+import org.pmiops.workbench.billing.FreeTierBillingService;
 import org.pmiops.workbench.captcha.CaptchaVerificationService;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.UserDao;
@@ -130,6 +131,7 @@ public class ProfileController implements ProfileApiDelegate {
   private final DemographicSurveyMapper demographicSurveyMapper;
   private final DirectoryService directoryService;
   private final FireCloudService fireCloudService;
+  private final FreeTierBillingService freeTierBillingService;
   private final InstitutionService institutionService;
   private final PageVisitMapper pageVisitMapper;
   private final ProfileAuditor profileAuditor;
@@ -152,6 +154,7 @@ public class ProfileController implements ProfileApiDelegate {
       DemographicSurveyMapper demographicSurveyMapper,
       DirectoryService directoryService,
       FireCloudService fireCloudService,
+      FreeTierBillingService freeTierBillingService,
       InstitutionService institutionService,
       PageVisitMapper pageVisitMapper,
       ProfileAuditor profileAuditor,
@@ -171,6 +174,7 @@ public class ProfileController implements ProfileApiDelegate {
     this.demographicSurveyMapper = demographicSurveyMapper;
     this.directoryService = directoryService;
     this.fireCloudService = fireCloudService;
+    this.freeTierBillingService = freeTierBillingService;
     this.institutionService = institutionService;
     this.mailServiceProvider = mailServiceProvider;
     this.pageVisitMapper = pageVisitMapper;
@@ -543,6 +547,12 @@ public class ProfileController implements ProfileApiDelegate {
     profileService.validateUpdatedProfile(updatedProfile, previousProfile);
 
     profileService.adminUpdateProfile(updatedProfile);
+
+    // update the user's free-tier-active status, if this changes
+    if (!updatedProfile.getFreeTierDollarQuota().equals(previousProfile.getFreeTierDollarQuota())) {
+      freeTierBillingService.setFreeTierDollarOverride(
+          user, updatedProfile.getFreeTierDollarQuota());
+    }
 
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
       profileService.saveVerifiedInstitutionalAffiliation(user, updatedProfile);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -522,20 +522,34 @@ public class ProfileController implements ProfileApiDelegate {
   public ResponseEntity<Void> updateProfile(Profile updatedProfile) {
     DbUser user = userProvider.get();
 
-    // Save current profile for audit trail. Continue to use the userProvider (instead
-    // of info on previousProfile) to ensure addition of audit system doesn't change behavior.
-    // That is, in the (rare, hopefully) condition that the old profile gives incorrect information,
-    // the update will still work as well as it would have.
+    // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
-    final VerifiedInstitutionalAffiliation updatedAffil =
-        updatedProfile.getVerifiedInstitutionalAffiliation();
-    final VerifiedInstitutionalAffiliation prevAffil =
-        previousProfile.getVerifiedInstitutionalAffiliation();
-    if (!Objects.equals(updatedAffil, prevAffil)) {
+
+    if (!Objects.equals(
+        updatedProfile.getVerifiedInstitutionalAffiliation(),
+        previousProfile.getVerifiedInstitutionalAffiliation())) {
       throw new BadRequestException("Cannot update Verified Institutional Affiliation");
     }
 
     profileService.updateProfileForUser(user, updatedProfile, previousProfile);
+
+    return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
+  }
+
+  public ResponseEntity<Void> adminUpdateProfile(DbUser user, Profile updatedProfile) {
+    // Save current profile for audit trail.
+    final Profile previousProfile = profileService.getProfile(user);
+
+    validateUpdatedProfile(updatedProfile, previousProfile);
+
+    profileService.adminUpdateProfile(updatedProfile);
+
+    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
+      saveVerifiedInstitutionalAffiliation(user, updatedProfile);
+    }
+
+    final Profile appliedUpdatedProfile = profileService.getProfile(user);
+    profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
 
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
@@ -550,6 +564,9 @@ public class ProfileController implements ProfileApiDelegate {
     if (verifiedAffiliation == null) {
       throw new BadRequestException("Cannot delete Verified Institutional Affiliation.");
     }
+
+    // if the Institution's Display Name differs from that of the VerifiedInstitutionalAffiliation
+    // passed in here, use the Institution's Display Name
 
     Optional<Institution> institution =
         institutionService.getInstitution(verifiedAffiliation.getInstitutionShortName());
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -540,12 +540,12 @@ public class ProfileController implements ProfileApiDelegate {
     // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
 
-    validateUpdatedProfile(updatedProfile, previousProfile);
+    profileService.validateUpdatedProfile(updatedProfile, previousProfile);
 
     profileService.adminUpdateProfile(updatedProfile);
 
     if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
-      saveVerifiedInstitutionalAffiliation(user, updatedProfile);
+      profileService.saveVerifiedInstitutionalAffiliation(user, updatedProfile);
     }
 
     final Profile appliedUpdatedProfile = profileService.getProfile(user);
--- a/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ProfileController.java
@@ -544,11 +544,7 @@ public class ProfileController implements ProfileApiDelegate {
   @Override
   public ResponseEntity<Profile> adminUpdateProfile(Profile updatedProfile) {
     final String username = updatedProfile.getUsername();
-    final DbUser user =
-        userService
-            .getByUsername(username)
-            .orElseThrow(
-                () -> new NotFoundException("Could not find user account for " + username));
+    final DbUser user = userService.getByUsernameOrThrow(username);
 
     // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
--- a/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
+++ b/api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
@@ -295,4 +295,28 @@ public class FreeTierBillingService {
     return Optional.ofNullable(user.getFreeTierCreditsLimitDollarsOverride())
         .orElse(workbenchConfigProvider.get().billing.defaultFreeCreditsDollarLimit);
   }
+
+  /**
+   * Set a custom Free Tier dollar limit override for this user. If this is greater than the user's
+   * total cost, set their workspaces to active. Note: lowering the limit below total cost will not
+   * set the workspaces to inactive. checkFreeTierBillingUsage() will do this as part of the next
+   * cron run.
+   *
+   * @param user the user as represented in our database
+   * @return the US dollar amount, represented as a double
+   */
+  public boolean setFreeTierDollarOverride(DbUser user, double dollarLimit) {
+    // TODO: prevent setting this limit directly except in this method?
+    user.setFreeTierCreditsLimitDollarsOverride(dollarLimit);
+
+    // may be redundant: enable anyway
+    boolean enableWorkspaces =
+        Optional.ofNullable(getCachedFreeTierUsage(user))
+            .map(usage -> compareCosts(usage, dollarLimit) < 0)
+            .orElse(true);
+    if (enableWorkspaces) {
+      updateFreeTierWorkspacesStatus(user, BillingStatus.ACTIVE);
+    }
+    return enableWorkspaces;
+  }
 }
--- a/api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
+++ b/api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
@@ -14,7 +14,6 @@ import org.pmiops.workbench.billing.FreeTierBillingService;
 import org.pmiops.workbench.config.WorkbenchConfig;
 import org.pmiops.workbench.db.dao.InstitutionDao;
 import org.pmiops.workbench.db.dao.UserDao;
-import org.pmiops.workbench.db.dao.UserService;
 import org.pmiops.workbench.db.dao.UserTermsOfServiceDao;
 import org.pmiops.workbench.db.dao.VerifiedInstitutionalAffiliationDao;
 import org.pmiops.workbench.db.model.DbDemographicSurvey;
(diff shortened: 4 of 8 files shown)


api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ProfileAuditorImpl.kt
```diff
@@ -64,7 +64,7 @@ constructor(
                             agentIdMaybe = userProvider.get().userId,
                             agentEmailMaybe = userProvider.get().username,
                             targetType = TargetType.PROFILE,
-                            targetIdMaybe = userProvider.get().userId,
```
jmthibault79, line 67: Now shared between updateProfile and adminUpdateProfile - so we can't assume the target is the caller

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -337,6 +343,16 @@ private void validateUpdatedProfile(Profile updatedProfile, Profile prevProfile)
       // See RW-1488.
       throw new BadRequestException("Changing username is not supported");
     }
+    if (updatedProfile.getVerifiedInstitutionalAffiliation() == null
+        && prevProfile.getVerifiedInstitutionalAffiliation() != null) {
+      throw new BadRequestException("Cannot delete Verified Institutional Affiliation.");
+    }
+
+    if (!prevProfile.getGivenName().equalsIgnoreCase(updatedProfile.getGivenName())
```
jmthibault79, line 351: moved from updateProfileForUser to make this common between update and adminUpdate

Planning to make this more explicit as part of RW-5104

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -667,16 +683,12 @@ private boolean userHasEverLoggedIn(
   public ResponseEntity<Void> updateProfile(Profile updatedProfile) {
     DbUser user = userProvider.get();
 
-    // Save current profile for audit trail. Continue to use the userProvider (instead
-    // of info on previousProfile) to ensure addition of audit system doesn't change behavior.
```
jmthibault79, line 671: can't use the userProvider in the common code because it may refer to the admin user instead of the update target

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -685,15 +697,33 @@ private boolean userHasEverLoggedIn(
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
-  private void updateProfileForUser(DbUser user, Profile updatedProfile, Profile previousProfile) {
```
jmthibault79, line 688: github is confused.  this still exists below

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -685,15 +697,33 @@ private boolean userHasEverLoggedIn(
...
+    // Save current profile for audit trail.
+    final Profile previousProfile = profileService.getProfile(user);
+
     validateUpdatedProfile(updatedProfile, previousProfile);
 
-    if (!userProvider.get().getGivenName().equalsIgnoreCase(updatedProfile.getGivenName())
-        || !userProvider.get().getFamilyName().equalsIgnoreCase(updatedProfile.getFamilyName())) {
-      userService.setDataUseAgreementNameOutOfDate(
-          updatedProfile.getGivenName(), updatedProfile.getFamilyName());
+    profileService.adminUpdateProfile(updatedProfile);
```
jmthibault79, line 706: adminUpdateProfile uses a mostly-different path here, so as not to affect other endpoints much.  RW-5104 will unify them.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -685,15 +697,33 @@ private boolean userHasEverLoggedIn(
...
+
+    // update the user's free-tier-active status, if this changes
+    if (!updatedProfile.getFreeTierDollarQuota().equals(previousProfile.getFreeTierDollarQuota())) {
```
jmthibault79, line 709: new code: billing un-lock-out

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -685,15 +697,33 @@ private boolean userHasEverLoggedIn(
...
+      freeTierBillingService.setFreeTierDollarOverride(
+          user, updatedProfile.getFreeTierDollarQuota());
     }
 
+    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
```
jmthibault79, line 714: users can't update their own affiliation

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -717,32 +747,40 @@ private void updateProfileForUser(DbUser user, Profile updatedProfile, Profile p
     user.setDemographicSurvey(dbDemographicSurvey);
     user.setLastModifiedTime(now);
 
-    updateInstitutionalAffiliations(updatedProfile, user);
```
jmthibault79, line 720: this is the old-style flow so it only applies when the FF is false

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -754,6 +792,9 @@ private void updateProfileForUser(DbUser user, Profile updatedProfile, Profile p
       throw new BadRequestException("Cannot delete Verified Institutional Affiliation.");
     }
 
+    // if the Institution's Display Name differs from that of the VerifiedInstitutionalAffiliation
```
jmthibault79, line 795: VerifiedInstitutionalAffiliation has a non-authoritative displayName, to make `get` more descriptive

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -28,7 +29,7 @@
   private final InstitutionService institutionService;
   private final ProfileMapper profileMapper;
   private final UserDao userDao;
-  private final UserService userService;
```
jmthibault79, line 31: was not used

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -135,4 +136,14 @@ public void validateInstitutionalAffiliation(Profile profile) {
       throw new BadRequestException(msg);
     }
   }
+
+  public void adminUpdateProfile(Profile updatedProfile) {
```
jmthibault79, line 140: ProfileController.adminUpdateProfile() will be pushed down to here as part of RW-5104

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -862,6 +864,20 @@ public void updateProfile_removeVerifiedInstitutionalAffiliationForbidden() {
     profileController.updateProfile(profile);
   }
 
+  @Test(expected = BadRequestException.class)
+  public void adminUpdateProfile_removeVerifiedInstitutionalAffiliationForbidden() {
```
jmthibault79, line 868: many of these tests are duplicates of existing updateProfile as admin versions

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -874,6 +890,39 @@ public void updateVerifiedInstitutionalAffiliation_removeForbidden() {
     profileController.updateVerifiedInstitutionalAffiliation(dbUser.getUserId(), null);
   }
 
+  @Test
+  public void adminUpdateProfile_disable() {
```
jmthibault79, line 894: new tests

AuthDomainController.updateUserDisabledStatus() is the standalone version of the API

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -939,14 +988,20 @@ public void updateName_alsoUpdatesDua() {
     assertThat(duas.get(0).isUserNameOutOfDate()).isTrue();
   }
 
-  @Test(expected = BadRequestException.class)
-  public void updateGivenName_badRequest() {
```
jmthibault79, line 943: moved lower and renamed to `updateProfile_badRequest_familyName_long()`

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -522,16 +526,12 @@ private boolean userHasEverLoggedIn(
...
-    // the update will still work as well as it would have.
+    // Save current profile for audit trail.
     final Profile previousProfile = profileService.getProfile(user);
-    final VerifiedInstitutionalAffiliation updatedAffil =
-        updatedProfile.getVerifiedInstitutionalAffiliation();
-    final VerifiedInstitutionalAffiliation prevAffil =
-        previousProfile.getVerifiedInstitutionalAffiliation();
-    if (!Objects.equals(updatedAffil, prevAffil)) {
+
+    if (!Objects.equals(
```
jmthibault79, line 532: not actually a change

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -46,7 +46,6 @@
   private final InstitutionService institutionService;
   private final ProfileAuditor profileAuditor;
   private final ProfileMapper profileMapper;
-  private final Provider<DbUser> userProvider;
```
jmthibault79, line 49: was not used

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -172,12 +169,6 @@ public void validateInstitutionalAffiliation(Profile profile) {
   public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile previousProfile) {
     validateUpdatedProfile(updatedProfile, previousProfile);
 
-    if (!userProvider.get().getGivenName().equalsIgnoreCase(updatedProfile.getGivenName())
```
jmthibault79, line 175: moved to validateUpdatedProfile() because it also applies to the adminUpdateProfile()

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,32 +192,40 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
       validateInstitutionalAffiliation(updatedProfile);
+    } else {
+      updateInstitutionalAffiliations(updatedProfile, user);
     }
 
-    userService.updateUserWithConflictHandling(user);
+    user = userService.updateUserWithConflictHandling(user);
+
     if (requireInstitutionalVerification) {
-      DbVerifiedInstitutionalAffiliation updatedDbVerifiedAffiliation =
```
jmthibault79, line 213: this is now saveVerifiedInstitutionalAffiliation() which is called in two places

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -334,7 +344,22 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
+
+    // TODO RW-5104: move these DB null fixes to mapper or other common code
+
+    user.getPageVisits().forEach(dbPageVisit -> dbPageVisit.setUser(user));
+    user.getInstitutionalAffiliations().forEach(affiliation -> affiliation.setUser(user));
+    Optional.ofNullable(user.getDemographicSurvey()).ifPresent(survey -> survey.setUser(user));
+    Optional.ofNullable(user.getAddress()).ifPresent(addr -> addr.setUser(user));
+
+    // need to retain some DbUser fields which are not present in Profile
+    final DbUser existingUser = userDao.findUserByUserId(updatedProfile.getUserId());
```
jmthibault79, line 358: these will probably also into the mapper for RW-5104 (which needs further refinement btw)

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/ProfileAuditorImpl.kt
jaycarlton: Good catch.

Thinking out loud though, I wonder if it would ever make sense to have a targetUserProvider for administrative methods. I.e. it would provide the administrator doing the call (if one exists), otherwise null (or we can try the OptionalProvider, but CH wasn't too excited about that last time).

In other words, I rely on Spring context for things like this: info that changes globally.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jaycarlton: Can you please write up a 1-2 page plan for how the calls are going to work and where the responsibilities are going to lie now?

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jaycarlton: why not include this logic in the ProfileService? If anyone calls `profileService.adminUpdateProfile(updatedProfile)` from somewhere else, this fixup code won't be executed.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
jaycarlton: So is the idea that a profile update never touches the verified inst affl? If that's true, then I don't understand why it's part of this controller method.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -540,6 +540,35 @@ private boolean userHasEverLoggedIn(
...
+      freeTierBillingService.setFreeTierDollarOverride(
+          user, updatedProfile.getFreeTierDollarQuota());
+    }
+
+    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
+      profileService.saveVerifiedInstitutionalAffiliation(user, updatedProfile);
+    }
+
+    final Profile appliedUpdatedProfile = profileService.getProfile(user);
+    profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
```
jaycarlton, line 567: ⚠️ You need to update the `ProfileTargetProperty` enum so that the audit log will pick up changes to institutional affiliation (assuming you don't want to make Institution its own target type and CHANGE_AFFILIATION_TO and CHANGE_AFFILIATION_FROM into action types). I would've probably done it the hard way, but we should be able to query for these either way. 

The key point to consider is whether consumers of the log should consider all these changes "profile changes" or more serious admin changes.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -295,4 +295,28 @@ public double getUserFreeTierDollarLimit(DbUser user) {
...
+    // TODO: prevent setting this limit directly except in this method?
+    user.setFreeTierCreditsLimitDollarsOverride(dollarLimit);
+
+    // may be redundant: enable anyway
+    boolean enableWorkspaces =
+        Optional.ofNullable(getCachedFreeTierUsage(user))
+            .map(usage -> compareCosts(usage, dollarLimit) < 0)
+            .orElse(true);
+    if (enableWorkspaces) {
+      updateFreeTierWorkspacesStatus(user, BillingStatus.ACTIVE);
```
jaycarlton, line 318: I feel like architecturally this is an event that needs to take place as a result of the event that changes the limit I think. I.e. all the edits that could affect workspace status should either fire an event or set off a business logic function that makes this determination.

If you put this check here, and then we realize later we need additional business logic to evaluate the status, or we have another method someplace that affects the usage or dollar limit, we'll get out of sync.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,32 +192,40 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
     user.setDemographicSurvey(dbDemographicSurvey);
     user.setLastModifiedTime(now);
 
-    updateInstitutionalAffiliations(updatedProfile, user);
     boolean requireInstitutionalVerification =
         workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification;
     if (requireInstitutionalVerification) {
       validateInstitutionalAffiliation(updatedProfile);
+    } else {
```
jaycarlton, line 199: If there's no way to roll back institutional affiliations at this point, or even run without them, then it's cleaner to ditch the feature flag.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,32 +192,40 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
-      this.verifiedInstitutionalAffiliationDao.save(updatedDbVerifiedAffiliation);
+      saveVerifiedInstitutionalAffiliation(user, updatedProfile);
     }
 
     final Profile appliedUpdatedProfile = getProfile(user);
     profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
   }
 
+  public void saveVerifiedInstitutionalAffiliation(DbUser user, Profile updatedProfile) {
+    DbVerifiedInstitutionalAffiliation updatedDbVerifiedAffiliation =
```
jaycarlton, line 214: `DbAffiliation` 🙏

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,32 +192,40 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
-                  verifiedInstitutionalAffiliation.getVerifiedInstitutionalAffiliationId()));
-      this.verifiedInstitutionalAffiliationDao.save(updatedDbVerifiedAffiliation);
+      saveVerifiedInstitutionalAffiliation(user, updatedProfile);
     }
 
     final Profile appliedUpdatedProfile = getProfile(user);
     profileAuditor.fireUpdateAction(previousProfile, appliedUpdatedProfile);
   }
 
+  public void saveVerifiedInstitutionalAffiliation(DbUser user, Profile updatedProfile) {
```
jaycarlton, line 213: Please split this per the [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). I.e. I'd expect this method to take in only a DbUser and an Affiliation ID. Since that's all it needs according to the name, the rest is confusing.

Also "updated profile" is a description that only makes sense within a function. Updated how?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,32 +192,40 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
+    DbVerifiedInstitutionalAffiliation updatedDbVerifiedAffiliation =
+        verifiedInstitutionalAffiliationMapper.modelToDbWithoutUser(
+            updatedProfile.getVerifiedInstitutionalAffiliation(), institutionService);
```
jaycarlton, line 216: nit: `profileToDb()` is fine, since the signature implies that there's no user provided, and there's no DbProfile.

But this feels like going the long way around somehow...

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton, line 213: So we have a function here that takes in two things, looks up something else, updates that, and returns void. It doesn't really feel active.

Returning the saved affiliation could help with debugging if nothing else.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,7 +285,7 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
     }
   }
 
-  private void validateUpdatedProfile(Profile updatedProfile, Profile prevProfile)
+  public void validateUpdatedProfile(Profile updatedProfile, Profile prevProfile)
```
jaycarlton, line 288: I feel like having to expose something like this is a code smell. This should be a guarantee of the profile service: it always returns valid profiles or throws. I shouldn't have to remember to call a validator (or which validator to call).

Granted, you inherited this issue and the night is young.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -334,7 +344,22 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
 
-  public List<Profile> listAllProfiles() {
-    return userService.getAllUsers().stream().map(this::getProfile).collect(Collectors.toList());
+  public void adminUpdateProfile(Profile updatedProfile) {
+    final DbUser user = profileMapper.profileToDbUser(updatedProfile);
+
+    // TODO RW-5104: move these DB null fixes to mapper or other common code
+
+    user.getPageVisits().forEach(dbPageVisit -> dbPageVisit.setUser(user));
+    user.getInstitutionalAffiliations().forEach(affiliation -> affiliation.setUser(user));
```
jaycarlton, line 353: Why does the affiliation object need to point back to the user if it's 1:1 now?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -334,7 +344,22 @@ private void validateStringLength(String field, String fieldName, int max, int m
     }
   }
 
-  public List<Profile> listAllProfiles() {
-    return userService.getAllUsers().stream().map(this::getProfile).collect(Collectors.toList());
+  public void adminUpdateProfile(Profile updatedProfile) {
+    final DbUser user = profileMapper.profileToDbUser(updatedProfile);
```
jaycarlton, line 348: There should be a getOrThrow variant that will 404.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -334,7 +344,22 @@ private void validateStringLength(String field, String fieldName, int max, int m
...
+
+    // TODO RW-5104: move these DB null fixes to mapper or other common code
+
+    user.getPageVisits().forEach(dbPageVisit -> dbPageVisit.setUser(user));
```
jaycarlton, line 352: I don't understand what's going on with this. How did the user get page visits that don't know she is the user?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton, line 348: OK, so you're making a new user every time? That doesn't sound like we are using the ORM system right. Can't we make edits on the existing one and save that?

If you ran into some issue, please comment on that before trying to manually fixup the profile. That seems really error prone and difficult to maintain.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: This just seems odd.

I believe the mapper changes are a requirement if you feel we have to do it this way. It's way too easy to have a bug where an admin change causes us to lose state in the User table (if I'm reading this right).

api/src/main/resources/workbench-api.yaml
```diff
@@ -1163,6 +1163,33 @@ paths:
           description: User doesn't have the ACCESS_CONTROL_ADMIN authority
           schema:
             "$ref": "#/definitions/ErrorResponse"
+
+  "/v1/admin/users":
+    patch:
```
jaycarlton, line 1168: oh nice

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -848,6 +851,22 @@ public void updateVerifiedInstitutionalAffiliation_update() {
         .isEqualTo(verifiedInstitutionalAffiliation);
   }
 
+  @Test(expected = BadRequestException.class)
+  public void updateProfile_changeVerifiedInstitutionalAffiliationForbidden() {
+    config.featureFlags.requireInstitutionalVerification = true;
+
+    final VerifiedInstitutionalAffiliation affiliation = createVerifiedInstitutionalAffiliation();
+    createAccountRequest.getProfile().setVerifiedInstitutionalAffiliation(affiliation);
```
jaycarlton, line 859: I really only like to use class members for state that's not mutable, and create things like requests that have to be dynamic in the test methods themselves. As a reader, I don't know what side effects createUser() has, or what's happened to createAccountRequest in the meantime.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
jaycarlton: ⚠️ Do we want to allow the administrator to change every single field in the  Profile? Are there any fields that only a user should set (or be able to see)? I feel like that might be true today (and we should verify that), but I don't want to hang our hat on that assumption.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
jaycarlton: Send these two APIs to the Thunderdome then.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -1015,6 +1103,74 @@ public void updateFamilyName_badRequest() {
     profileController.updateProfile(profile);
   }
 
+  @Test(expected = BadRequestException.class)
+  public void adminUpdateProfile_badRequest_nullAddress() {
+    createUser();
```
jaycarlton, line 1108: can we just move createUser to setup()?

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -1015,6 +1103,74 @@ public void updateFamilyName_badRequest() {
...
+    Profile profile = profileController.getMe().getBody();
+    profile.setAddress(null);
+    profileController.adminUpdateProfile(profile);
+  }
+
+  @Test(expected = BadRequestException.class)
+  public void adminUpdateProfile_badRequest_nullCountry() {
```
jaycarlton, line 1115: ☠️  It will be important, as long as we have dual entry points, to write a regression test that sets every user-settable field on the profile and make sure the same thing happens in both updateProfile() and adminUpdateProfile().

It's going to be bloated, but we've basically duplicated one of the hairiest methods we have now.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -1015,6 +1103,74 @@ public void updateFamilyName_badRequest() {
...
+    createUser();
+    Profile profile = profileController.getMe().getBody();
+    profile.getAddress().country(null);
+    profileController.adminUpdateProfile(profile);
+  }
+
+  @Test(expected = BadRequestException.class)
+  public void adminUpdateProfile_badRequest_nullState() {
```
jaycarlton, line 1123: Can you please put a test for the case where nothing needs to update?

jmthibault79:
This work was based on a flawed understanding of the requirements of this ticket.  Will open a new PR when ready.

PR #3678, opened 2020-06-16
[RW-4957][risk=no] UI to update verified institutional affiliation

als364:
Screen Shot 2020-06-16 at 2 04 00 PM

A UI for the API developed in #3627. Plus a small change on the API side to make that API actually work 🙃 

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -705,11 +710,26 @@ private void updateProfileForUser(DbUser user, Profile updatedProfile, Profile p
...
     updateInstitutionalAffiliations(updatedProfile, user);
-    if (workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification) {
+    boolean requireInstitutionalVerification =
+        workbenchConfigProvider.get().featureFlags.requireInstitutionalVerification;
+    if (requireInstitutionalVerification) {
       profileService.validateInstitutionalAffiliation(updatedProfile);
     }
 
     userService.updateUserWithConflictHandling(user);
+    if (requireInstitutionalVerification) {
```
als364, line 720: this whole block is kind of janky and I expect the entire thing to get deleted by RW-4980.

ui/src/app/pages/login/account-creation/common.tsx
```diff
@@ -158,3 +160,21 @@ export const OptionalDropDownSection = (props) => {
   </DropDownSection>;
 };
 
+export const getRoleOptions = (institutions: Array<PublicInstitutionDetails>, institutionShortName: string):
```
jmthibault79, line 163: Hooray for reuse.  Should this go up a level or two?  A little weird for Profile Edit to be referring to this file.  Maybe some kind of institution/common

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -43,14 +47,15 @@ const freeCreditLimitOptions = [
   {label: '$800', value: 800}
 ];
 
-const ReadonlyInputWithLabel = ({label, content, dataTestId, inputStyle = {}}) => {
+const TextInputWithLabel = ({label, content, dataTestId, onChange = (value) => {}, disabled= false, inputStyle = {}}) => {
```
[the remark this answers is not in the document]
als364: oh snap. I had no idea lol. At this point I just assume we don't have any components for anything. I will refactor

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -136,32 +145,134 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
+    );
+  }
+
+  isSaveDisabled(errors) {
+    const {oldProfile, updatedProfile} = this.state;
+    return fp.isEqual(oldProfile, updatedProfile) || errors;
+  }
+
+  setVerifiedInstitutionOnProfile(institutionShortName: string) {
+    const {verifiedInstitutionsByShortname} = this.state;
```
petesantos, line 185: Consider trying to perform the state change in one operation with flow:
```suggestion
    this.setState(fp.flow(
        fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionShortName'],institutionShortName),
        fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionDisplayName'], fp.find(['shortName', institutionShortName], sortedInstitutions).displayName),
        fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionRoleEnum'], undefined),
        fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionalRoleOtherText'], undefined)
    ))
```

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -136,32 +145,134 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
+
+  validateVerifiedInstitutionalAffiliation() {
+    const {updatedProfile} = this.state;
+    if (updatedProfile) {
+      if (updatedProfile.verifiedInstitutionalAffiliation) {
+        return updatedProfile.verifiedInstitutionalAffiliation;
+      }
+      return false;
+    }
+    return false;
```
petesantos, line 223: Take a look at [optional chaining ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) with a ternary here - it may clean this up a bit

If that works, it may work for a few of the functions below as well

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -136,32 +145,134 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
+          ['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionDisplayName'],
+        verifiedInstitutionsByShortname.get(institutionShortName).displayName
+        )
+    );
+    this.setState(fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionRoleEnum'], undefined));
+    this.setState(fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionalRoleOtherText'], undefined));
+  }
+
+  setInstitutionalRoleOnProfile(institutionalRoleEnum: InstitutionalRole) {
+    this.setState(fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionalRoleEnum'], institutionalRoleEnum));
```
petesantos, line 203: Similar thought as above wrt using fp.flow

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -136,32 +145,134 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
+    this.setState(fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionalRoleOtherText'], undefined));
+  }
+
+  updateVerifiedInstitutionalAffiliation() {
+    const {updatedProfile} = this.state;
```
petesantos, line 208: nit: consider further destructuring
```suggestion
const {updatedProfile: {userId, verifiedInstitutionalAffiliation}} = this.state
```

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -114,10 +121,12 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
     this.state = {
       institutionsLoadingError: '',
       loading: true,
-      profile: null,
+      oldProfile: null,
```
petesantos, line 117: I am a little confused on the introduction of oldProfile. It seems to replace saveDisabled, but the saveDisabled state is still getting set in a few places, but never read. The oldProfile also does not get updated. A few questions:

1. When the profile is updated and saved, it seems like the save button will still be enabled (I may be wrong) - is this intentional? 
2. Would it be unreasonable to revert back to one profile state, and setting `saveDisabled` to `false` in `setVerifiedInstitutionOnProfile`?

ui/src/app/pages/login/account-creation/common.tsx
```diff
@@ -158,3 +160,21 @@ export const OptionalDropDownSection = (props) => {
...
+    Array<{label: string, value: InstitutionalRole}> => {
+  if (isBlank(institutionShortName)) {
+    return [];
+  }
+
+  const selectedOrgType = institutions.find(
```
petesantos: Do we have a standard around using native find (and other array functions), or utilize lodash find in these cases?
My personal preference is to keep things consistent with lodash, which also has other benefits like auto currying.

Also, a few nits - could this be changed to destructure out `organizationTypeEnum` (reassigning the name in the destructure). Could `shortName` be destructured out as well?

ui/src/app/utils/institutions.tsx
```diff
@@ -0,0 +1,21 @@
...
+import {InstitutionalRole, PublicInstitutionDetails} from 'generated/fetch';
+import {isBlank} from './index';
+
+export const getRoleOptions = (institutions: Array<PublicInstitutionDetails>, institutionShortName: string):
+    Array<{ label: string, value: InstitutionalRole }> => {
+  if (isBlank(institutionShortName)) {
+    return [];
+  }
+
+  const selectedOrgType = institutions.find(
```
petesantos, line 11: Do we have a standard around using native find (and other array functions), or utilize lodash find in these cases?
My personal preference is to keep things consistent with lodash, which also has other benefits like auto currying.

Also, a few nits - could this be changed to destructure out `organizationTypeEnum` (reassigning the name in the destructure). Could `shortName` be destructured out as well?

ui/src/app/pages/admin/admin-user.tsx
als364: you can't use the updatedProfile if you do that. but I can do two lines of destructuring.

ui/src/app/pages/admin/admin-user.tsx
als364: Our version of TypeScript doesn't support that.

ui/src/app/pages/admin/admin-user.tsx
petesantos: I don't see updatedProfile being used in this method outside of the property access, that is why I suggested the destructure in that way - is there something I am missing?

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -136,32 +145,134 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
-          return {
-            'label': institution.displayName ? institution.displayName : institution.shortName,
-            'value': {displayName: institution.displayName, shortName: institution.shortName}
-          };
-        },
-        institutionsResponse.institutions
-      );
-      this.setState({verifiedInstitutionOptions: options});
+      const institutionsResponse = await institutionApi().getPublicInstitutionDetails();
+      const institutions = institutionsResponse.institutions;
```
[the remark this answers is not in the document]
petesantos: It is good to think about the time complexity. My concern is around the code complexity - adding additional state and code to manage that state. My personal preference is for simpler code that is easier to maintain over time, even if that means in some cases it is not the most optimized. Part of my rationale for that is to prevent premature optimization and instead opt for simpler code. I feel removing the Map, reduce and additional state is worth the tradeoff. I do think your concern is valid, but if it does not cause a noticeable performance issue I would prefer managing less state in the component and if a performance issue does arise in the future we could address it.

Also, looking at this a bit more closely, it looks like the `find` would only occur on a selection change (it should not occur on a re-render - unless I am missing something). I ran a simple performance check in the browser comparing the find against the initial state setup with reduce (10,000 items). The reduce takes 1.8 - 3ms, whereas the find, in the worst case, takes about 0.6 - 0.8ms. You could run the find ~2-5 times in the amount of time it takes to run the reduce (which will occur on each mount). I can give you the code I ran if you are interested.

To me that makes it unclear which would be more performant in the real world - but again I would prefer to not avoid thinking too much about the optimization unless a performance problem became evident.

All that said - if it is too time consuming to make the change right now, I am ok with leaving it as is. I just wanted to better explain my position.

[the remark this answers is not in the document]
als364: Thank you for running a test! Because of email validation, the find will end up running on all renders. If it's a matter of milliseconds even for ten thousand things then I'll do the simpler code. This is what I get for assuming a new JS framework works like an ancient JS framework 🙂

ui/src/app/pages/admin/admin-user.tsx
als364: This works, but not if I combine it with other state changes for some reason

ui/src/app/pages/admin/admin-user.tsx
als364: After a successful save, I change the oldProfile to the updatedProfile so that the check of whether the Save button is enabled works properly.

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -132,36 +148,202 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
+  }
+
+  async setVerifiedInstitutionOnProfile(institutionShortName: string) {
+    const {verifiedInstitutionOptions} = this.state;
+    await this.setState({loading: true});
+    await this.setState(fp.flow(
+      fp.set(['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionShortName'], institutionShortName),
+      fp.set(
+          ['updatedProfile', 'verifiedInstitutionalAffiliation', 'institutionDisplayName'],
+          verifiedInstitutionOptions.find(
```
petesantos, line 257: nit: Do we need to grab the property after the find when setting the `institutionDisplayName`

ui/src/app/pages/admin/admin-user.tsx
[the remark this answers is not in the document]
petesantos: Thanks for doing that!

ui/src/app/pages/admin/admin-user.tsx
als364: eek! yes

als364:
the E2E tests are failing because they're timing out. Merging.

PR #3680, opened 2020-06-17
[RW-4398][risk=no] Puppeteer new Concept Sets tests

aweng98:
Screen Shot 2020-06-17 at 9 38 11 AM

New `conceptset-create.spec.ts` test and supporting page objects files.

### DIFF
--- a/e2e/app/component/concept-domain-card.ts
+++ b/e2e/app/component/concept-domain-card.ts
@@ -38,11 +38,7 @@ export default class ConceptDomainCard extends Container {
     */
   async getParticipantsCount(): Promise<string> {
     const selector = `${this.getXpath()}//*[contains(normalize-space(text()), "participants in domain")]`;
-    const elemt = await this.page.waitForXPath(selector, {visible: true});
-    const textContent = await elemt.getProperty('textContent');
-    const value = (await textContent.jsonValue()).toString();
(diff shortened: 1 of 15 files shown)


e2e/app/component/concept-domain-card.ts
```diff
@@ -0,0 +1,68 @@
...
+    const selectConceptButton = await this.getSelectConceptButton();
+    await selectConceptButton.click();
+    await this.page.waitForXPath('//*[@data-test-id="conceptTable"]', {visible: true});
+    await waitWhileLoading(this.page);
+  }
+
+   /**
+    * Find displayed number of Participants in this domain.
+    */
+  async getParticipantsCount(): Promise<string> {
```
jmthibault79, line 39: these 2 functions are nearly identical, so we can use common code

aweng98: yes. made a common function
```
private async extractConceptsCount(selector: string): Promise<string> {
    const elemt = await this.page.waitForXPath(selector, {visible: true});
    const textContent = await elemt.getProperty('textContent');
    const value = (await textContent.jsonValue()).toString();
    const regex = new RegExp(/\d{1,3}(,?\d{3})*/); // Match numbers with comma
    return (regex.exec(value))[0];
  }
```
new function is called in `getConceptsCount()` and `getParticipantsCount()`.

PR #3682, opened 2020-06-18
[RW-4696][risk=no] Update footer per mocks and fix flow issues

calbach:
Demo @ https://calbach-dot-all-of-us-workbench-test.appspot.com

- Update footer to match latest mocks
  - The updated content reduced the width below our minimum, which fixed several flow issues
- Fix overlap between footer and right-hand help bar
- Add some extra margin-bottom in a few places where the footer was crowding the page content
- Hide the footer in two new places: TOS and DUCC
  - TOS: the account creation TOS change can just be hardcoded, due to the nature of the component
  - DUCC: needed to extend the React app router HOC to also include `minimizeChrome`, which is a route config data we have in the Angular router
- Put modals on top of the side bar / footer. Previously, the help bar was not behind the modal overlay - which was a bug

Feedback on the route data changes would be appreciated @petesantos , not sure if you were planning something else for this kind of use case.

### DIFF
--- a/ui/src/app/cohort-search/modal/modal.component.tsx
+++ b/ui/src/app/cohort-search/modal/modal.component.tsx
@@ -55,7 +55,7 @@ const styles = reactStyles({
     visibility: 'hidden',
     transform: 'scale(1.1)',
     transition: 'visibility 0.25s linear, opacity 0.25s 0s, transform 0.25s',
-    zIndex: 102,
+    zIndex: 105,
   },
   panelLeft: {
     display: 'none',
--- a/ui/src/app/components/modals.tsx
+++ b/ui/src/app/components/modals.tsx
@@ -15,7 +15,9 @@ const styles = reactStyles({
 
   overlay: {
     backgroundColor: Color(colors.dark).alpha(0.85).toString(), padding: '1rem', display: 'flex',
-    position: 'fixed', left: 0, right: 0, top: 0, bottom: 0, overflowY: 'auto'
+    position: 'fixed', left: 0, right: 0, top: 0, bottom: 0, overflowY: 'auto',
+    // Keep z-index in sync with popups.tsx.
+    zIndex: 105
   },
 
   modalTitle: {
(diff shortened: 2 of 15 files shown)


ui/src/app/pages/signed-in/component.css
```diff
@@ -43,3 +43,8 @@ body {
 .create-account__degree-select .p-multiselect-header {
   display: none;
 }
+
+/* Must be higher than the side helpbar index. See help-bar.tsx. */
+app-footer {
```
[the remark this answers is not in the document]
calbach: Yep... after clicking around a bit more of course I uncovered several other tug of wars. I think the state of the world is OK for now with my latest updates. The tension here is that we have high level structural flow issues on the page. These are not easily solved without getting everything onto React, then refactoring. I've filed https://precisionmedicineinitiative.atlassian.net/browse/RW-5110 to track this.

ui/src/app/cohort-search/modal/modal.component.tsx
```diff
@@ -55,7 +55,7 @@ const styles = reactStyles({
     visibility: 'hidden',
     transform: 'scale(1.1)',
     transition: 'visibility 0.25s linear, opacity 0.25s 0s, transform 0.25s',
-    zIndex: 102,
+    zIndex: 105,
```
calbach, line 58: @dolbeew FYI

ui/src/app/components/modals.tsx
```diff
@@ -15,7 +15,9 @@ const styles = reactStyles({
 
   overlay: {
     backgroundColor: Color(colors.dark).alpha(0.85).toString(), padding: '1rem', display: 'flex',
-    position: 'fixed', left: 0, right: 0, top: 0, bottom: 0, overflowY: 'auto'
+    position: 'fixed', left: 0, right: 0, top: 0, bottom: 0, overflowY: 'auto',
+    // Keep z-index in sync with popups.tsx.
+    zIndex: 105
```
calbach, line 20: @dolbeew FYI - this should fix the issue where the right-hand-side help bar sits above the modal overlay e.g. on saving a cohort or dataset, and was still visible/clickable. I assume this was a bug.

calbach:
I will merge since I did not make substantial changes here. I'm only enabling this in test/local for now. Will flip on tomorrow or Monday in other envs if there are no major issues discovered.

PR #3683, opened 2020-06-18
[no ticket][risk=no] Set maximum number of tests run in parallel

aweng98:
Reduce maximum number of Puppeteer tests to run in parallel on localhost, thus, reducing their probability of failing.


aweng98:
@als364 This change helps with reducing tests failures on localhost, pointed to local dev server.

als364:
oh sweet thank you!

PR #3692, opened 2020-06-19
[RW-4696][risk=no] Sticky the dataset builder footer

calbach:
fixed --> sticky, to avoid bad double-footer interactions

I checked the codebase for fixed position, but missed this as I didn't understand the analyze button was implemented as a footer.

stick-dataset


calbach:
Ran puppeteer tests locally. All passed except:

```
Waiting for page title contains "Redirect Notice" failed.
```

I'm not quite sure of the logic, but I think it may just autoredirect in certain cases. We may want to adjust that test to also accept just landing on the login page.

Either way, it's not related to the footer.

PR #3695, opened 2020-06-21
[RW-4696][risk=no] Bump the z-index of the floating dropdown in the share modal

calbach:
Footer-related-fix `#3` and counting. The z-index changes inadvertently caused issues with the sharing role dropdown, which gets rendered as a floating fixed element. After auditing the other dialogs in the app, this appears to be the only regression caused by the modal z index changes.

I could not write a regression test using enzyme. Instead, I have created a puppeteer test to cover the sharing flow. I've verified that it serves as a regression test for this issue, as well as for RW-5013.


calbach:
Thank you for testing!

PR #84, opened 2017-08-21
Separating out project name from firecloud name in workspace table.

danqrodney:
Generating project name and firecloud name when creating workspace (based on name).

Fixing data access level conversion to use toString() instead of name().

PR #3696, opened 2020-06-21
[RW-4696][RW-5013][risk=no] Initial workspace sharing puppeteer test

calbach:
This is intended as a regression test for two separate bugs:

- https://github.com/all-of-us/workbench/pull/3694
- https://github.com/all-of-us/workbench/pull/3695

### DIFF
--- a/e2e/app/component/share-modal.ts
+++ b/e2e/app/component/share-modal.ts
@@ -1,8 +1,8 @@
 import {ElementHandle, Page} from 'puppeteer';
 import {WorkspaceAccessLevel} from 'app/page-identifiers';
 import {ElementType} from 'app/xpath-options';
+import Dialog from 'app/component/dialog';
 import Button from 'app/element/button';
-import Dialog from 'app/element/dialog';
 import Textbox from 'app/element/textbox';
 import ClrIcon from 'app/element/clr-icon-link';
 
@@ -52,7 +52,7 @@ export default class ShareModal extends Dialog {
   private collabRowXPath(username: string): string {
     // We dip into child contents to find the collab user row element parent.
     // .//div filters by a relative path to the parent row.
-    return `{this.getXPath()}//*[@data-test-id="collab-user-row" and .//div[` +
+    return `${this.getXpath()}//*[@data-test-id="collab-user-row" and .//div[` +
         `@data-test-id="collab-user-email" and contains(text(),"${username}")]]`;
   }
 
--- a/e2e/app/component/share-modal.ts
+++ b/e2e/app/component/share-modal.ts
@@ -32,7 +32,7 @@ export default class ShareModal extends Container {
     // This is pretty limited - here we assume that the first collab with "Reader"
     // is the user we care about, which may not be the case. Ideally we'd have a
     // cleaner selection here.
-    const roleInput = await this.waitForFirstReaderRoleDropdown();
+    const roleInput = await this.waitForRoleSelectorForUser(username);
     await roleInput.click();
 
     const ownerOpt = await this.waitForRoleOption(level);
@@ -44,12 +44,17 @@ export default class ShareModal extends Container {
 
   async removeUser(username: string): Promise<void> {
     const rmCollab = await this.page.waitForXPath(
-      `//*[contains(text(),"${username}")]/clr-icon[@shape="minus-circle"]`);
+      `${this.collabRowXPath(username)}//clr-icon[@shape="minus-circle"]`);
     await rmCollab.click();
 
     await this.clickButton(ButtonLabel.Save);
     await this.waitUntilDialogIsClosed();
     return;
+  }
+
+  private collabRowXPath(username: string): string {
+    return `//*[@data-test-id="collab-user-row" and .//div[` +
+        `@data-test-id="collab-user-email" and contains(text(),"${username}")]]`;
   }
 
   async clickButton(buttonLabel: ButtonLabel): Promise<void> {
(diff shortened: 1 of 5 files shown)


aweng98:
@calbach This is great. I am so happy to see you're trying out Puppeteer.
My suggestions are in following. Let me know what you think of them.

(second and third xpaths are "copy" of the first xpath, with less condition)
```
If user id and access-level are known
'//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and starts-with(text(), "OWNER") and contains(text(), "alexweng@fake-research-aou.org")]'


If user id is known (="alexweng@fake-research-aou.org")
'//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and contains(text(), "alexweng@fake-research-aou.org")]'

If access-level is known (= "OWNER")
'//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and starts-with(text(), "OWNER")]'
```
Unfortunately, Puppeteer supports xpath 1.0, not 2.0 or 3.0. so cannot use `ends-with()` for user id string.


Test code:
```
      const infoLink = await ClrIconLink.findByName(page, {containsText: 'Collaborators', iconShape: 'info-standard'});
      const cls = (await infoLink.getProperty('className'));
      expect(cls).toBe('is-solid');

      const shareButton = await Button.findByName(page, {name: 'Share'});
      const butnText = await shareButton.getTextContent();
      expect(butnText).toBe('Share');

      const user = 'puppettester1@fake-research-aou.org';
      const accessLevel = 'OWNER';

      const selector1 = `//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and contains(text(), "${user}")]`;
      const elemt1 = await page.waitForXPath(selector1);
      const value1 = await (await elemt1.getProperty('textContent')).jsonValue();
      expect(value1).toContain(accessLevel);

      const selector2 = `//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and starts-with(text(), "${accessLevel}")]`;
      const elemt2 = await page.waitForXPath(selector2);
      const value2 = await (await elemt2.getProperty('textContent')).jsonValue();
      expect(value2).toContain(user);

      const selector3 = `//*[descendant::h3[text()="Collaborators:"]]/*[starts-with(@data-test-id,"workspaceUser-") and starts-with(text(), "${accessLevel}") and contains(text(), "${user}")]`;
      const elemt3 = await page.waitForXPath(selector3);
      expect(elemt3).toBeTruthy();
      const value3 = await (await elemt2.getProperty('textContent')).jsonValue();
      console.log(`textContent = ${value3}`);
```

calbach:
Thanks! Changes some of the xpath selectors per Slack discussion.

e2e/app/component/share-modal.ts
```diff
@@ -0,0 +1,107 @@
...
+export enum ButtonLabel {
+  Cancel = 'Cancel',
+  Save = 'Save'
+}
+
+const Selector = {
+  defaultDialog: '//*[@role="dialog"]',
+}
+
+export default class ShareModal extends Container {
```
aweng98, line 19: Can you change to `ShareModal extends Dialog` so common methods are not duplicated?

e2e/app/component/share-modal.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  async waitForAddCollaboratorIcon(): Promise<ClrIcon> {
+    return ClrIcon.findByName(this.page, {iconShape: 'plus-circle'}, this);
+  }
+
+  async waitForRoleSelectorForUser(username: string): Promise<Textbox> {
+    const box = new Textbox(
+      this.page, `${this.collabRowXPath(username)}//input[@type="text"]`);
+    await box.waitForXPath({visible: true});
```
aweng98, line 84: nit: Line #84 is unnecessary here because the `click()` takes care the finding element too.
```
async click(options?: ClickOptions): Promise<void> {
    return (await this.asElementHandle()).click(options);
  }
```

e2e/app/component/share-modal.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Creates an xpath to a share "row" for a given user within the modal. This
+   * can be combined with a child selector to pull out a control for a user.
+   */
+  private collabRowXPath(username: string): string {
+    // We dip into child contents to find the collab user row element parent.
+    // .//div filters by a relative path to the parent row.
+    return `//*[@data-test-id="collab-user-row" and .//div[` +
```
aweng98, line 59: nit: If this class extends `Dialog`, you can also adds `${this.getXpath()}` at the beginning of xpath to ensure element is located in dialog.

e2e/app/page/workspace-about-page.ts
```diff
@@ -33,4 +36,37 @@ export default class WorkspaceAboutPage extends AuthenticatedPage{
     return waitForAttributeEquality(page, {xpath: selector}, 'aria-selected', 'true');
   }
 
+  async findUserInCollaboratorList(username: string): Promise<WorkspaceAccessLevel> {
+    // At least one collab should eventually render, i.e. the current user.
+    const collabXPath = `//*[starts-with(@data-test-id,"workspaceUser-")]`;
+    await this.page.waitForXPath(collabXPath);
```
aweng98, line 42: `waitForXPath()` needs `{visible: true}` option here to ensure element is visible.

e2e/tests/workspace/workspace-share.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      const aboutPage = new WorkspaceAboutPage(page);
+      await aboutPage.waitForLoad();
+
+      // This test is not hermetic - if the collaborator is already on this
+      // workspace, just remove them before continuing.
+      let accessLevel = await aboutPage.findUserInCollaboratorList(config.collaboratorUsername);
+      if (accessLevel !== null) {
+        await (await aboutPage.openShareModal()).removeUser(config.collaboratorUsername);
+        waitWhileLoading(page);
```
aweng98, line 30: `waitWhileLoading` needs `await`.

e2e/app/component/share-modal.ts
calbach: Hmm.. is that obvious? I'm not sure I would have expected that calling `click()` indirectly also waits for the element to become visible.

e2e/app/page/workspace-about-page.ts
calbach: Done

e2e/tests/workspace/workspace-share.spec.ts
calbach: oops, thanks

e2e/app/component/share-modal.ts
calbach: Done

e2e/app/component/share-modal.ts
calbach: Done

calbach:
Hmm.. I'll come back to this tomorrow. I'm stuck on clicking the dropdown "role" menu when sharing. When I run it locally in debug mode it works. Then when I run it in headless mode it fails (instead, it adds the user as a Writer or Reader, not an Owner). Let me know if you have any ideas on debugging the difference between these two modes.

calbach:
OK, finally got it working locally.

Oddly, I got it to a point where it was only working when I took a screenshot before clicking the menu. I determined this probably had something to do with the screen size / rendering being evaluated, so I decided to set a viewport for the test, which magically resolved the issue.

At some point, I also found that hover before click of the React select menu helped, but now that no longer seems necessary.

PR #3704, opened 2020-06-23
[RW-5125][risk=no] Update grep regex patters in CircleCI

aweng98:
The command `halt_job_noncode_changes` is called by `puppeteer-*` jobs only. Update its' regex expression to skip `puppeteer-*` job.

Don't run e2e tests if the commit includes non-product code change. I think running e2e tests would be waste of time and resources.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,18 +100,19 @@ commands:
           name: Update git submodules
           command: git submodule update --init --recursive
 
-  halt_job_noncode_changes:
-    description: "Halt job and succeed early if no code changes on non-master branch"
+  halt_if_code_unchanged:
+    description: "Halt job if no code changed"
     steps:
       - run:
           command: |
-            if [ ${CIRCLE_BRANCH} != "" ] &&
-              [ ${CIRCLE_BRANCH} != "master" ] &&
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
-                echo "No code changes on non-master branch. halting job."
-                circleci step halt
+            if ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvFf .circleci/e2e-job-ignore-patterns.txt; then
+              echo "Application code have not changed. halting job."
+              circleci step halt
+            else
+              git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -vFf .circleci/e2e-job-ignore-patterns.txt
+              echo "job continuing."
             fi
-          name: Halt job and succeed early if no code changes on non-master branch
+          name: Halt job if no code changed
 
   ensure_branch_has_changes:
     description: Halt job and succeed early if no code changes in << parameters.dir_name >> directory on non-master branch
@@ -559,7 +560,7 @@ jobs:
     parallelism: 3
     steps:
       - checkout_init_git
-      - halt_job_noncode_changes
+      - halt_if_code_unchanged
       - browser-tools/install-browser-tools
       - run_e2e_test
 
@@ -573,16 +574,11 @@ jobs:
     working_directory: ~/workbench
     parallelism: 1
     steps:
-      - run:
-          name: Skip job if this is not a pull request
-          command: |
-            echo "PR: $CI_PULL_REQUEST"
-            if [ ${CIRCLE_BRANCH} != "" ] && [ ${CIRCLE_BRANCH} == "master" ] ; then
-              echo "Not a PR, skipping Puppeteer tests"
-              circleci-agent step halt
-            fi
       - checkout_init_git
-      - halt_job_noncode_changes
+      - run:
+          name: Halt job if not a pull request
+          command: bash .circleci/pr-skip-ci.sh
+      - halt_if_code_unchanged
       - browser-tools/install-browser-tools
       - start_local_ui
       - run_e2e_test
@@ -597,7 +593,6 @@ workflows:
       - ui-unit-test
       - api-bigquery-test
       - api-integration-test
-      # Disable e2e tests https://precisionmedicineinitiative.atlassian.net/browse/RW-5080 is resolved.
       # - puppeteer-e2e-local-ui
       # Run deployment to "test" on master merges.
       - api-deploy-to-test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -593,7 +593,7 @@ workflows:
       - ui-unit-test
       - api-bigquery-test
       - api-integration-test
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -107,7 +107,9 @@ commands:
           command: |
             if [ ${CIRCLE_BRANCH} != "" ] &&
               [ ${CIRCLE_BRANCH} != "master" ] &&
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
+              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) \
+                | grep -qvE '(.md)|(.pdf)|(.png)|(.svg)|(Test.java)|(Test.kt)|(.mp4)|(.log)|(.txt)'; then
```
jaycarlton, line 111: I kinda want to take this logic out of this file and put it in a script that we just call here.

It's conceivable that a .txt file could influence the behavior of the product, though I don't know of any current examples of this.

Eventually I'd like to move to a model where the first job determines what future jobs to run. That is, we wouldn't see a green checkmark for halted jobs; it would just never have run at all.

Is there a way to halt with another symbol (like it does for skipped tests)? Seems like I looked at one point.

.circleci/config.yml
```diff
@@ -107,7 +107,9 @@ commands:
           command: |
             if [ ${CIRCLE_BRANCH} != "" ] &&
               [ ${CIRCLE_BRANCH} != "master" ] &&
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
+              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) \
```
jaycarlton, line 110: Have you tested what happens when we make a PR against a non-master branch? I've occasionally felt the need to do this (e.g. when lots of reviewers are on vacation). Do we just not run CircleCI jobs?

.circleci/config.yml
jaycarlton, line 111: Can you put all these extensions into an environment var called `E2E_IGNORE_PATTERNS`? Can we use this approach for all our jobs?

.circleci/config.yml
```diff
@@ -100,18 +100,19 @@ commands:
...
-              ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvE '(.md$)|(.pdf$)'; then
-                echo "No code changes on non-master branch. halting job."
-                circleci step halt
+            if ! git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -qvFf .circleci/e2e-job-ignore-patterns.txt; then
+              echo "Application code have not changed. halting job."
+              circleci step halt
+            else
+              git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) | grep -vFf .circleci/e2e-job-ignore-patterns.txt
+              echo "job continuing."
             fi
```
aweng98, line 114: Checking for PR branch `if [ ${CIRCLE_BRANCH} != "" ] && [ ${CIRCLE_BRANCH} != "master" ]` is now replaced by `pr-skip-ci.sh`.

The `grep` regex patterns is in `e2e-job-ignore-patterns.txt`.

Screen Shot 2020-06-23 at 7 46 14 PM

e2e/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "cross-env WORKBENCH_ENV=dev jest --maxWorkers=3",
     "test:debug": "cross-env WORKBENCH_ENV=dev HEADLESS=false jest --detectOpenHandles",
     "test-local": "cross-env WORKBENCH_ENV=local jest --maxWorkers=2",
-    "test:ci": "cross-env NODE_ENV=development CI=true jest --ci --maxWorkers=2 --forceExit",
+    "test:ci": "cross-env NODE_ENV=development CI=true jest --ci --maxWorkers=1",
```
aweng98, line 12: Unrelated to CircleCI config change, but it's related to running e2e tests in CircleCI. Abort job early if more than 3 tests have failed.

.circleci/config.yml
```diff
@@ -604,11 +604,11 @@ workflows:
           requires:
             - ui-unit-test
       # On master merges, run Puppeteer e2e tests after ui and api deployed to "test" env successfully.
-      # - puppeteer-e2e-test:
-          # filters: *filter_only_master_branch
-          # requires:
-            # - api-deploy-to-test
-            # - ui-deploy-to-test
+      - puppeteer-e2e-test:
+          filters: *filter_only_master_branch
```
jaycarlton, line 608: Awesome! I didn't see where you checked this in.

jaycarlton: Oh, it's the same PR.

.circleci/pr-skip-ci.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+
+# This script makes CircleCI exit a non pull-request job.
+
+echo "CIRCLE_PULL_REQUEST: ${CIRCLE_PULL_REQUEST}"
+regexp="[[:digit:]]\+$"
```
jaycarlton, line 7: Are you scanning the PR description for a Jira Ticket number?

I thought CircleCI already had a away to disable it for all but pull requests, but I remember we decided not to do that anymore for some reason.

.circleci/config.yml
jaycarlton: Very nice. I like how you didn't have to update a docker image to add your script. I was worried it would be worse.

.circleci/e2e-job-ignore-patterns.txt
```diff
@@ -0,0 +1,10 @@
+.md
+.jpg
```
jaycarlton, line 2: If you're doing any screenshot comparison tests (which are kind of a pain), then some of these formats might need to be included. Likewise, if an image's dimensions change, it could make a button inaccessible, etc. So this may be a little strict.

.circleci/pr-skip-ci.sh
aweng98: No, not scanning PR description for ticket number.
The CircleCI setting to pull on PR only is still enabled, no change there.
Screen Shot 2020-06-24 at 11 43 17 AM

.circleci/e2e-job-ignore-patterns.txt
aweng98: Good point. Screenshot comparison testing is on the longer term road map.

.circleci/config.yml
aweng98: So, the `else` part in shell script is just to show you that non-matching files are the reason job is continuing. I can remove it if you like.

PR #3708, opened 2020-06-24
[RW-4929][risk=low] Add flag control & GSuite email extraction for dev auto-registration

gjuggler:
This PR attempts to formalize our dev user registration flow, which had previously been inlined into AuthInterceptor.

Breaking this function out into its own service should make it clearer to any readers what's going on.

The only slightly clever thing going on here is with institutional affiliation: to support the follow-up work (where we want to auto-load a few dev institutions into a blank-slate dev database), this PR adds some logic to dev user registration which finds the first matching institution, if available, and uses that for the verified institutional affiliation. This felt to me like the right thing to do in preparation for the future, but I'm open to alternative views.

---
**PR checklist**

- [No – this adds a long-lived server config variable] If this includes a new feature flag, I have created and linked new JIRA tickets to (a) turn on the feature flag and (b) remove it later

### DIFF
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -87,7 +87,7 @@
     "enableDataUseAgreement": true,
     "enableBetaAccess": false,
     "unsafeAllowSelfBypass": true,
-    "unsafeAllowUserCreationFromOauthData": true,
+    "unsafeAllowUserCreationFromGSuiteData": true,
     "requireInvitationKey": false
   },
   "featureFlags": {
--- a/api/config/config_local.json
+++ b/api/config/config_local.json
@@ -87,6 +87,7 @@
     "enableDataUseAgreement": true,
     "enableBetaAccess": false,
     "unsafeAllowSelfBypass": true,
+    "unsafeAllowUserCreationFromOauthData": true,
     "requireInvitationKey": false
   },
   "featureFlags": {
--- a/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
@@ -1,6 +1,7 @@
 package org.pmiops.workbench.google;
 
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth8.assertThat;
 
 import java.time.Clock;
 import java.util.Map;
@@ -53,13 +54,11 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
                 });
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(aouMeta).containsEntry("Institution", "All of Us Research Workbench");
-    assertThat(service.getContactEmail(userName).get()).isEqualTo("notasecret@gmail.com");
+    assertThat(service.getContactEmail(userName)).hasValue("notasecret@gmail.com");
     assertThat(
-            service
-                .getContactEmailFromGSuiteEmail(
-                    userName + "@" + config.googleDirectoryService.gSuiteDomain)
-                .get())
-        .isEqualTo("notasecret@gmail.com");
+            service.getContactEmailFromGSuiteEmail(
+                userName + "@" + config.googleDirectoryService.gSuiteDomain))
+        .hasValue("notasecret@gmail.com");
 
     service.deleteUser(userName);
     assertThat(service.isUsernameTaken(userName)).isFalse();
--- a/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
@@ -53,7 +53,12 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
                 });
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(aouMeta).containsEntry("Institution", "All of Us Research Workbench");
-    assertThat(service.getContactEmailAddress(userName)).isEqualTo("notasecret@gmail.com");
+    assertThat(service.getContactEmail(userName)).isEqualTo("notasecret@gmail.com");
+    assertThat(
+            service.getContactEmailFromGSuiteEmail(
+                userName + "@" + config.googleDirectoryService.gSuiteDomain))
+        .isEqualTo("notasecret@gmail.com");
+
     service.deleteUser(userName);
     assertThat(service.isUsernameTaken(userName)).isFalse();
   }
--- a/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
@@ -35,9 +35,10 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
 
   @Test
   public void testCreateAndDeleteTestUser() {
-    String userName = String.format("integration.test.%d", Clock.systemUTC().millis());
-    service.createUser("Integration", "Test", userName, "notasecret@gmail.com");
-    assertThat(service.isUsernameTaken(userName)).isTrue();
+    String userPrefix = String.format("integration.test.%d", Clock.systemUTC().millis());
+    String username = userPrefix + "@" + config.googleDirectoryService.gSuiteDomain;
+    service.createUser("Integration", "Test", username, "notasecret@gmail.com");
+    assertThat(service.isUsernameTaken(userPrefix)).isTrue();
 
     // As of ~6/25/19, customSchemas are sometimes unavailable on the initial call to Gsuite. This
     // data is likely not written with strong consistency. Retry until it is available.
@@ -46,7 +47,7 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
             .execute(
                 c -> {
                   Map<String, Map<String, Object>> schemas =
-                      service.getUserByUsername(userName).getCustomSchemas();
+                      service.getUser(username).getCustomSchemas();
                   if (schemas == null) {
                     throw new RuntimeException("custom schemas is still null");
                   }
@@ -54,14 +55,10 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
                 });
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(aouMeta).containsEntry("Institution", "All of Us Research Workbench");
-    assertThat(service.getContactEmail(userName)).hasValue("notasecret@gmail.com");
-    assertThat(
-            service.getContactEmailFromGSuiteEmail(
-                userName + "@" + config.googleDirectoryService.gSuiteDomain))
-        .hasValue("notasecret@gmail.com");
+    assertThat(service.getContactEmail(username)).hasValue("notasecret@gmail.com");
 
-    service.deleteUser(userName);
-    assertThat(service.isUsernameTaken(userName)).isFalse();
+    service.deleteUser(username);
+    assertThat(service.isUsernameTaken(userPrefix)).isFalse();
   }
 
   private static RetryTemplate retryTemplate() {
--- a/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
+++ b/api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
@@ -53,10 +53,12 @@ public class DirectoryServiceImplIntegrationTest extends BaseIntegrationTest {
                 });
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(aouMeta).containsEntry("Institution", "All of Us Research Workbench");
-    assertThat(service.getContactEmail(userName)).isEqualTo("notasecret@gmail.com");
+    assertThat(service.getContactEmail(userName).get()).isEqualTo("notasecret@gmail.com");
     assertThat(
-            service.getContactEmailFromGSuiteEmail(
-                userName + "@" + config.googleDirectoryService.gSuiteDomain))
+            service
+                .getContactEmailFromGSuiteEmail(
+                    userName + "@" + config.googleDirectoryService.gSuiteDomain)
+                .get())
         .isEqualTo("notasecret@gmail.com");
 
     service.deleteUser(userName);
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -312,7 +312,10 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
   }
 
   @Override
-  public DbUser createUser(final Userinfoplus oAuth2Userinfo, final String contactEmail) {
+  public DbUser createUser(
+      final Userinfoplus oAuth2Userinfo,
+      final String contactEmail,
+      DbVerifiedInstitutionalAffiliation dbVerifiedAffiliation) {
     return createUser(
         oAuth2Userinfo.getGivenName(),
         oAuth2Userinfo.getFamilyName(),
@@ -326,7 +329,7 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
         null,
         null,
         null,
-        null);
+        dbVerifiedAffiliation);
   }
 
   @Override
--- a/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
@@ -313,13 +313,14 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
 
   @Override
   public DbUser createUser(
-      final Userinfoplus oAuth2Userinfo,
+      final Userinfoplus userInfo,
       final String contactEmail,
       DbVerifiedInstitutionalAffiliation dbVerifiedAffiliation) {
     return createUser(
-        oAuth2Userinfo.getGivenName(),
-        oAuth2Userinfo.getFamilyName(),
-        oAuth2Userinfo.getEmail(),
+        userInfo.getGivenName(),
+        userInfo.getFamilyName(),
+        // This GSuite primary email address is what RW refers to as `username`.
+        userInfo.getEmail(),
         contactEmail,
         null,
         null,
@@ -336,7 +337,7 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
   public DbUser createUser(
       String givenName,
       String familyName,
-      String userName,
+      String username,
       String contactEmail,
       String currentPosition,
       String organization,
@@ -350,7 +351,7 @@ public class UserServiceImpl implements UserService, GaugeDataCollector {
     DbUser dbUser = new DbUser();
     dbUser.setCreationNonce(Math.abs(random.nextLong()));
     dbUser.setDataAccessLevelEnum(DataAccessLevel.UNREGISTERED);
-    dbUser.setUsername(userName);
+    dbUser.setUsername(username);
     dbUser.setContactEmail(contactEmail);
     dbUser.setCurrentPosition(currentPosition);
     dbUser.setOrganization(organization);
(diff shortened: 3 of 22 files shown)


api/src/main/java/org/pmiops/workbench/google/DirectoryService.java
```diff
@@ -1,25 +1,29 @@
 package org.pmiops.workbench.google;
```
gjuggler, line 1: Changes in this file were mostly opportunistic (e.g. s/emailAddress/email/ in method names), but the key need was to add a method supportting conversion from the full GSuite email address (gjordan@fake-research-aou.org) to the username (gjordan).

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -356,4 +356,10 @@ private void validateInstitution(Institution institutionRequest) {
               });
     }
   }
+
+  public Optional<Institution> getFirstMatchingInstitution(final String contactEmail) {
```
gjuggler, line 360: I decided this wasn't worth a unit test case... feel free to disagree!

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -110,16 +114,15 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
     }
 
     final String token = authorizationHeader.substring("Bearer".length()).trim();
-    final Userinfoplus OAuth2Userinfo = userInfoService.getUserInfo(token);
+    final Userinfoplus userInfo = userInfoService.getUserInfo(token);
```
gjuggler, line 117: Opportunistic rename – the original var didn't conform to style guidelines, and userInfo is sufficiently precise.

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
```diff
@@ -47,95 +60,114 @@ public void handle() {}
   public void handle() {}
 }
 
+@RunWith(SpringRunner.class)
```
gjuggler, line 63: A lot of the changes in this file were related to Spring-ifying this test class, and pulling some very common logic into simple helper methods.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
```diff
@@ -53,7 +53,14 @@ public void testCreateAndDeleteTestUser() {
                 });
     // Ensure our two custom schema fields are correctly set & re-fetched from GSuite.
     assertThat(aouMeta).containsEntry("Institution", "All of Us Research Workbench");
-    assertThat(service.getContactEmailAddress(userName)).isEqualTo("notasecret@gmail.com");
+    assertThat(service.getContactEmail(userName).get()).isEqualTo("notasecret@gmail.com");
```
jaycarlton, line 56: nit: If you do
```
import static com.google.common.truth.Truth8.assertThat;
```
in addition to the main `assertThat()`, you can use 
```
assertThat(service.getContactEmail(userName)).hasValue("notasecret@gmail.com");
```

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -312,12 +312,15 @@ public DbUser createServiceAccountUser(String username) {
...
   @Override
-  public DbUser createUser(final Userinfoplus oAuth2Userinfo) {
+  public DbUser createUser(
+      final Userinfoplus oAuth2Userinfo,
+      final String contactEmail,
+      DbVerifiedInstitutionalAffiliation dbVerifiedAffiliation) {
     return createUser(
         oAuth2Userinfo.getGivenName(),
         oAuth2Userinfo.getFamilyName(),
         oAuth2Userinfo.getEmail(),
```
jaycarlton, line 322: Maybe add a comment that this "email" is what we now call `username`.

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -326,7 +329,7 @@ public DbUser createUser(final Userinfoplus oAuth2Userinfo) {
         null,
         null,
         null,
-        null);
+        dbVerifiedAffiliation);
```
jaycarlton, line 332: aside: a builder would be nice for this class.

api/src/main/java/org/pmiops/workbench/google/DirectoryService.java
```diff
@@ -1,25 +1,29 @@
...
+  Optional<String> getContactEmail(String username);
+
   /**
-   * Looks up a user by username and returns their stored contact email address. If no contact email
-   * is stored in G Suite, then null is returned.
+   * Looks up a user by GSuite email address and returns their stored contact email address, if
+   * available.
    */
-  String getContactEmailAddress(String username);
+  Optional<String> getContactEmailFromGSuiteEmail(String gSuiteEmailAddress);
```
jaycarlton, line 26: this is the only place I've seen `gSuiteEmailAddress` instead of `username`. Is the idea here that there may not actually be a user yet?

api/src/main/java/org/pmiops/workbench/google/DirectoryServiceImpl.java
```diff
@@ -189,12 +189,26 @@ public boolean isUsernameTaken(String username) {
...
-    return (String)
-        getUserByUsername(username)
-            .getCustomSchemas()
-            .get(GSUITE_AOU_SCHEMA_NAME)
-            .get(GSUITE_FIELD_CONTACT_EMAIL);
+  public Optional<String> getContactEmail(String username) {
+    return Optional.ofNullable(getUserByUsername(username))
+        .map(
+            user ->
+                (String)
```
jaycarlton, line 196: nit: If we need to do much with this API, a helper class might...help. For example
```
Optional<String> getStringField(customSchemaMap, schemaName, fieldName)
```

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -356,4 +356,10 @@ private void validateInstitution(Institution institutionRequest) {
...
+    return getInstitutions().stream()
+        .filter(institution -> validateInstitutionalEmail(institution, contactEmail))
```
jaycarlton, line 362: nit: I prefer very small names for lambda arguments, especially if they go out of scope in the same line.

api/src/main/java/org/pmiops/workbench/google/DirectoryService.java
gjuggler: Mostly for consistency and clarity within this file, since `username` within this service has an existing meaning (which is the GSuite email without the domain suffix). It looks like it would be possible to drop that meaning from this service – IMO not a huge win, I'll opt to leave as-is.

jaycarlton: In that case, I'd rename `username` in this class then to `usernamePrefix` or some such.

FWIW I rarely absorb code a file at a time. It's 80% hopping via IDE linkage between occurrences. Email already means two things. If username means two things, then we're in trouble.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
```diff
@@ -356,4 +356,10 @@ private void validateInstitution(Institution institutionRequest) {
               });
     }
   }
+
+  public Optional<Institution> getFirstMatchingInstitution(final String contactEmail) {
+    return getInstitutions().stream()
```
jaycarlton, line 361: super nit: might be faster to do this via a query directly on the DAO, but with a small # of institutions it doesn't much matter.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
jaycarlton: nit: `Userinfoplus` isn't the most descriptive class name. Maybe `oauthInfo`? "user" is implied I think.

api/src/main/java/org/pmiops/workbench/user/DevUserRegistrationService.java
```diff
@@ -0,0 +1,8 @@
+package org.pmiops.workbench.user;
+
+import com.google.api.services.oauth2.model.Userinfoplus;
+import org.pmiops.workbench.db.model.DbUser;
+
+public interface DevUserRegistrationService {
```
jaycarlton, line 6: nit: not sure if "dev" is an approved abbrev.

api/src/main/java/org/pmiops/workbench/user/DevUserRegistrationService.java
```diff
@@ -0,0 +1,8 @@
...
+  DbUser createUserFromUserInfo(Userinfoplus userInfo);
```
jaycarlton, line 7: Might as well use the full class name `createUserFromUserinfoplus` or leave it out, i.e. `createUser()`.

api/src/main/java/org/pmiops/workbench/user/DevUserRegistrationServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    Institution institution =
+        institutionService
+            .getFirstMatchingInstitution(contactEmail)
+            .orElseThrow(
+                () ->
+                    new BadRequestException(
+                        String.format(
+                            "Contact email %s does not match any institutions. Cannot register new dev user.",
+                            contactEmail)));
+    VerifiedInstitutionalAffiliation verifiedAffiliation =
```
jaycarlton, line 63: It seems awkward that this service needs to know anything about institutions. It seems like the UserService should have all the machinery necessary to affiliate the user.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -155,8 +159,13 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
     }
     DbUser user = userDao.findUserByUsername(userName);
```
jaycarlton, line 160: nit: you can use the new `Optional<DbUser> getByUsername(String username)` on `UserService` to avoid the null comparison.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -155,8 +159,13 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
     if (user == null) {
-      // TODO(danrodney): start populating contact email in Google account, use it here.
-      user = userService.createUser(OAuth2Userinfo);
+      if (workbenchConfigProvider.get().access.unsafeAllowUserCreationFromGSuiteData) {
+        DbUser dbUser = devUserRegistrationService.createUserFromUserInfo(userInfo);
+        log.info(String.format("Dev user '%s' has been re-created.", dbUser.getUsername()));
```
jaycarlton, line 164: for later: it would be good to audit this...though if it never happens in production 🤷 .

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
```diff
@@ -47,95 +59,114 @@ public void handle() {}
...
     workbenchConfig.googleDirectoryService = new GoogleDirectoryServiceConfig();
     workbenchConfig.googleDirectoryService.gSuiteDomain = "fake-domain.org";
     workbenchConfig.auth = new AuthConfig();
     workbenchConfig.auth.serviceAccountApiUsers = new ArrayList<>();
     workbenchConfig.auth.serviceAccountApiUsers.add("service-account@appspot.gserviceaccount.com");
-    this.interceptor =
-        new AuthInterceptor(
-            userInfoService, fireCloudService, Providers.of(workbenchConfig), userDao, userService);
-    this.user = new DbUser();
+    user = new DbUser();
```
jaycarlton, line 102: You could make this a bean I think.

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
```diff
@@ -47,95 +59,114 @@ public void handle() {}
...
-  @Mock private UserDao userDao;
+  @MockBean private UserInfoService userInfoService;
+  @MockBean private FireCloudService fireCloudService;
+  @MockBean private UserDao userDao;
+  @MockBean private UserService userService;
+  @MockBean private DevUserRegistrationService devUserRegistrationService;
+
+  private static WorkbenchConfig workbenchConfig;
+
   @Mock private HttpServletRequest request;
```
jaycarlton, line 75: nit: new convention is to have a `mock` prefix for mock instance variables, since the call site can be a long way from the declaration.

api/src/test/java/org/pmiops/workbench/user/DevUserRegistrationServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+    when(institutionService.getFirstMatchingInstitution("gregory.jordan.123@gmail.com"))
+        .thenReturn(Optional.of(new Institution().shortName("Google")));
+    when(institutionService.getDbInstitutionOrThrow(eq("Google"))).thenReturn(dbInstitution);
+    when(userService.createUser(
+            eq(userInfo), eq("gregory.jordan.123@gmail.com"), dbAffiliationCaptor.capture()))
+        .thenReturn(dbUser);
+
+    service.createUserFromUserInfo(userInfo);
+
+    assertThat(dbAffiliationCaptor.getValue().getInstitution().getShortName()).isEqualTo("Google");
```
jaycarlton, line 71: Are there no other side effects to verify here?

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
gjuggler: If we had a clear style rule one way or another, I'd follow it! For now I'm sticking with my general aesthetic, which is longer-but-clearer parameter names.

api/src/main/java/org/pmiops/workbench/institution/InstitutionServiceImpl.java
gjuggler: Good point – I'll avoid premature optimization in this case (especially since we do a full read of institutions in a few other places in this service).

api/src/main/java/org/pmiops/workbench/user/DevUserRegistrationService.java
gjuggler: I think dev is pretty well-established usage – e.g. `project.rb dev-up`

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
gjuggler: Not sure why?

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
gjuggler: 👍  I think creating an API style guide will be a helpful place to document conventions like this, especially when they're a change from much of our existing code.

api/src/test/java/org/pmiops/workbench/user/DevUserRegistrationServiceTest.java
gjuggler: The Mockito stubs above pretty much cover the function under test line-by-line... I couldn't really think of anything else that wouldn't be super redundant.

api/src/main/java/org/pmiops/workbench/user/DevUserRegistrationService.java
gjuggler: Yeah, I agree this isn't super consistent. Just `createUser` seems like an all around improvement – done.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
gjuggler: I made a good-faith attempt to use Optional patterns here, but I couldn't come up with a solution that wasn't more awkward than the current conditional logic. The challenge is that we want to *either* provide a non-null DbUser object if the flag allows it, *or* return false, which isn't easily expressed within chained optional lambdas.

There was a bug in my existing logic though, which I did fix (pushing new commit in a minute). So there's that...

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
gjuggler: I agree, but we don't control that class. It's part of Google's OAuth2 implementation.

api/config/config_local.json
```diff
@@ -87,6 +87,7 @@
     "enableDataUseAgreement": true,
     "enableBetaAccess": false,
     "unsafeAllowSelfBypass": true,
+    "unsafeAllowUserCreationFromGSuiteData": true,
```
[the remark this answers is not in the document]
gjuggler: Unsafe has the same meaning as the other unsafe flag: unsafe to have this flag turned on in production. Which is why we have unit tests to help prevent this from accidentally happening.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
```diff
@@ -155,8 +159,13 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
     }
     DbUser user = userDao.findUserByUsername(userName);
     if (user == null) {
-      // TODO(danrodney): start populating contact email in Google account, use it here.
-      user = userService.createUser(OAuth2Userinfo);
+      if (workbenchConfigProvider.get().access.unsafeAllowUserCreationFromGSuiteData) {
```
calbach, line 162: Thanks for this improvement. Will reference this in the post-mortem.

api/src/integration/java/org/pmiops/workbench/google/DirectoryServiceImplIntegrationTest.java
gjuggler: Good call, done. (Is there a good way to get IntelliJ to offer to auto-import the Truth8 version when I want to use this pattern?)

api/src/main/java/org/pmiops/workbench/google/DirectoryService.java
gjuggler: I was about to punt on this, but when I looked through even this file I found there was some inconsistency in how `username` alone was being used. I took a broader pass at cleaning up the naming in existing methods too (turns out most of these are only used in one place, so the surface area was relatively small). Feels like an improvement, as long as I didn't break anything. PTAL.

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
jaycarlton: Oh  if this class doesn't use a user provider then I guess t that doesn't make sense.

api/src/test/java/org/pmiops/workbench/interceptors/AuthInterceptorTest.java
jaycarlton: I could spend time on that. In most  instances folks are ok making little improvements like that without a guide. I've just been changing them all whenever I edit a test class.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
jaycarlton: I understand,  but I'm saying  changing the var  name to  be more descriptive at the  expense of matching the type would help.

api/src/main/java/org/pmiops/workbench/interceptors/AuthInterceptor.java
jaycarlton: Yeah, it's hard to write awkward logic with Optionals, which is kind of the point.

FWIW you can do `DbUser user  = getByUsername(String username).orElseGet(() -> userService.createServiceAccountUser(userName));` to clean up one  conditional. Or better yet put a `getOrCreateServiceAccount(String username)` on the `UserService`.

api/src/test/java/org/pmiops/workbench/user/DevUserRegistrationServiceTest.java
jaycarlton: OK. I was going to  see if any other fields on the  created user could  be validated, but it's mocks all  the way down.

PR #3714, opened 2020-06-25
[RW-5127][risk=no] Revert DUCC routing changes

calbach:
This got a bit messy, some of the stuff was still needed. I ripped out the DUCC changes only, but left in some of the infra. This PR should be reverted and fixed forward with the real fix.


calbach:
Only issue on previous run was a trivial lint error, which I've fixed in the latest commit. Going to merge this without letting Circle finish since we are actively broken - I will personally deal with any fallout if the unit tests wind up actually failing for an unforeseen reason.

PR #3718, opened 2020-06-27
[no ticket][risk=no] Puppeteer disable test retries for local test development

aweng98:
Issue: the auto-retry in the jest-circus config is annoying for local development. It has be set manually to 0 retries when working on the test on localhost.


aweng98:
@calbach FYI. PR to address one annoyance. I'll ask for PR reviewer next week.

PR #3725, opened 2020-06-29
[no ticket][risk=no] Puppeteer refactor: base-element.ts edit

aweng98:
Attempt to make code more readable: Rewrote asyn/await calls with [Promises chaining](https://javascript.info/promise-chaining).

Rename `pressReturnKey()` function to `pressReturn`.
Rename `tabKey()` function to `pressTab`.

Test were not changed.

### DIFF
--- a/e2e/app/component/navigation.ts
+++ b/e2e/app/component/navigation.ts
@@ -35,7 +35,7 @@ export default class Navigation {
     const findMenuItem = async (): Promise<ElementHandle | null> => {
       await Navigation.openNavMenu(page);
       const angleIconXpath = buildXPath({type: ElementType.Icon, iconShape: 'angle'});
-      await page.waitForXPath(angleIconXpath, {timeout: 2000});
+      await page.waitForXPath(angleIconXpath, {timeout: 2000, visible: true});
       const appLinkXpath = `//*[@role="button" and @tabindex="0"]//span[contains(., "${destinationApp}")]`;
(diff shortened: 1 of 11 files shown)


e2e/app/component/navigation.ts
```diff
@@ -35,7 +35,7 @@ export default class Navigation {
     const findMenuItem = async (): Promise<ElementHandle | null> => {
       await Navigation.openNavMenu(page);
       const angleIconXpath = buildXPath({type: ElementType.Icon, iconShape: 'angle'});
-      await page.waitForXPath(angleIconXpath, {timeout: 2000});
+      await page.waitForXPath(angleIconXpath, {timeout: 2000, visible: true});
```
NehaBroad, line 38: nit: we can have an input as a common variable and use it just so in case it changes in future you have to change in one place

e2e/app/element/base-element.ts
```diff
@@ -252,22 +280,23 @@ export default class BaseElement extends Container {
...
   }
 
   /**
    * Returns ElementHandle.
    */
-  async asElementHandle(): Promise<ElementHandle | null> {
-    await this.waitForXPath();
-    return this.element.asElement();
+  async asElementHandle(): Promise<ElementHandle> {
+    return this.waitForXPath();
```
NehaBroad, line 299: We are doing return without await because you need to return Promise rather than value?

PR #3726, opened 2020-06-29
[RW-5057][risk=no] move conceptSetMapper into ConceptSetService

freemabd:
This PR moves
1. conceptService into the conceptSetService
2. conceptSetMapper into the conceptSetService

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -22,28 +22,23 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -126,13 +117,11 @@
       NotebooksService notebooksService,
       Provider<DbUser> userProvider,
       @Qualifier(DatasetConfig.DATASET_PREFIX_CODE) Provider<String> prefixProvider,
-      WorkspaceService workspaceService,
-      ConceptSetMapper conceptSetMapper) {
```
jaycarlton, line 130: Ah. You're moving the usage of the mapper, not the mapper itself. I was confused for a second.

PR #3727, opened 2020-06-29
[RW-5144][risk=no] Puppeteer test: workspace shared as READER

aweng98:
Add a new test `Workspace READER cannot share edit or delete workspace` in `workspace-share.spec.ts`.

### DIFF
--- a/e2e/app/component/share-modal.ts
+++ b/e2e/app/component/share-modal.ts
@@ -1,8 +1,6 @@
 import {ElementHandle, Page} from 'puppeteer';
 import {LinkText, WorkspaceAccessLevel} from 'app/text-labels';
-import {ElementType} from 'app/xpath-options';
 import Dialog from 'app/component/dialog';
-import Button from 'app/element/button';
 import Textbox from 'app/element/textbox';
 import ClrIcon from 'app/element/clr-icon-link';
 
@@ -14,11 +12,14 @@ export default class ShareModal extends Dialog {
   }
 
   async shareWithUser(username: string, level: WorkspaceAccessLevel): Promise<void> {
+    /*
     if (!this.page.viewport()) {
       throw Error(
         'testing of the share modal requires a viewport to be set on account ' +
           'on odd rendering issues with the react-select component');
     }
+    */
+
     const searchBox = await this.waitForSearchBox();
     await searchBox.type(username);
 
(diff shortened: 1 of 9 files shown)


e2e/app/component/share-modal.ts
```diff
@@ -14,11 +12,14 @@ export default class ShareModal extends Dialog {
...
 
   async shareWithUser(username: string, level: WorkspaceAccessLevel): Promise<void> {
+    /*
     if (!this.page.viewport()) {
       throw Error(
         'testing of the share modal requires a viewport to be set on account ' +
           'on odd rendering issues with the react-select component');
     }
+    */
+
```
aweng98, line 22: works without setting viewport. I think finding element with `{visible: true}` is the reason.

e2e/app/component/share-modal.ts
```diff
@@ -56,16 +56,6 @@ export default class ShareModal extends Dialog {
...
-  async clickButton(buttonLabel: LinkText): Promise<void> {
-    const button = await this.waitForButton(buttonLabel);
-    await button.waitUntilEnabled();
-    return button.click();
-  }
-
-  async waitForButton(buttonLabel: LinkText): Promise<Button> {
-    return Button.findByName(this.page, {containsText: buttonLabel, type: ElementType.Button}, this);
-  }
-
```
aweng98, line 68: Dupl. functions. `clickButton()` and `waitForButton()` are in `Dialog` class.

e2e/app/component/share-modal.ts
gjuggler: If that's the case, please remove the commented code! :)

e2e/tests/workspace/workspace-share.spec.ts
```diff
@@ -49,5 +55,69 @@ describe('Share workspace', () => {
...
+
+      // Duplicate action is available for click.
+      await menu2.isDisabled(EllipsisMenuAction.Duplicate);
+
+      // Make sure the Search input-field in Share modal is disabled.
+      await workspaceCard2.clickWorkspaceName();
+      await (new DataPage(newPage)).openTab(TabLabelAlias.About);
+      const aboutPage = new WorkspaceAboutPage(newPage);
+      await aboutPage.waitForLoad();
+      const modal2 = await aboutPage.openShareModal();
```
gjuggler, line 112: [no action] TIL that you can open the share modal even if you're a read-only user. Is this somehow expected / required for the product? The only thing I can think of is that this is the path for a read-only user to remove themself from the workspace... but it does feel like an awkward flow if that's the case. More common / expected (e.g. for things like Google Groups) would be a button that says "Remove me from this workspace".

aweng98: I thought it was odd to find enabled Share button in About page. Maybe the purpose of Share modal is to show information about this workspace who else are OWNER/READER/WRITER.

PR #3734, opened 2020-06-30
[RW-5139][risk=low] Switch profile validation to only validate changed fields

gjuggler:
Here is a proposed fix for RW-5139 (a bug where empty profile address fields are preventing admins from changing institutional affiliations due to always-on validation logic).

Two separate issues were involved here:
(1) Users with a non-empty address entry, but invalid fields (e.g. city='')
(2) Users with a null address entry.

We have many instances of each in prod.

The solution here is based on a few principles / requirements:
* When a profile is being updated, we only care about actively validating changed fields.
* When a new profile is being created, all fields should be validated.
* Profile validation code should be simple, flexible, and unified.
* This is a P1 bug so we shouldn't boil the ocean (any more than necessary) in pursuit of a fix.

I took a stab here at using a generic object differ (Javers seems to be standard for this) to make it more straightforward to detect field-by-field changes. I think the end result strikes a nice balance, but I'm not dead set in this direction and am open to feedback.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -270,12 +270,8 @@ private DbUser initializeUserIfNeeded() {
 
     final Profile profile = request.getProfile();
 
-    // We don't include this check in validateAndCleanProfile since some existing user profiles
-    // may have empty addresses. So we only check this on user creation, not update.
-    Optional.ofNullable(profile.getAddress())
-        .orElseThrow(() -> new BadRequestException("Address must not be empty"));
-
-    profileService.validateAndCleanProfile(profile);
+    // Run all profile validation by passing in a null 'previous' object.
```
jaycarlton, line 273: nit: I'd add a wrapper around this in the service so you could do `profileService.validateProfile(profile);`. Ideally, we should very rarely need to write out `null`.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -227,23 +235,16 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
-
+  public void cleanProfile(Profile profile) throws BadRequestException {
     // Cleaning steps, which provide non-null fields or apply some cleanup / transformation.
     profile.setDemographicSurvey(
         Optional.ofNullable(profile.getDemographicSurvey()).orElse(new DemographicSurvey()));
     profile.setInstitutionalAffiliations(
         Optional.ofNullable(profile.getInstitutionalAffiliations()).orElse(new ArrayList<>()));
-    // We always store the username as all lowercase.
-    profile.setUsername(profile.getUsername().toLowerCase());
+    if (profile.getUsername() != null) {
```
jaycarlton, line 244: should we not just bail here if the username is missing? Either throw IllegalStateException or return a failure code.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
+   * @param updatedProfile
+   * @param prevProfile
+   * @throws BadRequestException
+   */
+  public void validateProfile(Profile updatedProfile, Profile prevProfile)
+      throws BadRequestException {
+    cleanProfile(updatedProfile);
+
+    boolean isNewObject = prevProfile == null;
+    Diff diff = javers.compare(prevProfile, updatedProfile);
```
jaycarlton, line 383: Did you look at `ModelBackedTargetProperty` and the `TargetPropertyExtractor`? It works via one of the target property enums like `ProprtyTargetProperty`.

I'd prefer not to have separate logic for essentially the same thing both here and in the audit code.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
+   *
+   * @param updatedProfile
+   * @param prevProfile
+   * @throws BadRequestException
+   */
+  public void validateProfile(Profile updatedProfile, Profile prevProfile)
+      throws BadRequestException {
+    cleanProfile(updatedProfile);
+
+    boolean isNewObject = prevProfile == null;
```
jaycarlton, line 382: If we could avoid nullable parameters, that would be better. Optional works, but even optional parameters are discouraged.

I think the breakdown is also a bit confusing, because `validateProfile()` calls `cleanProfile()`, which makes changes. I generally expect `validate` methods to just check something and either return a boolean or throw something.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
+    if (field.length() < min) {
+      if (min == 1) {
+        throw new BadRequestException(String.format("%s cannot be left blank.", fieldName));
+      } else {
+        throw new BadRequestException(
+            String.format("%s is under character minimum. (%d)", fieldName, min));
+      }
+    }
+  }
+
```
jaycarlton, line 307: Can we just add validation methods to the ProfileTargetProperty enum and cycle through those instead of putting a bunch of if's here? That way we won't have to remember to update them all.

Where we have
```
enum class ProfileTargetProperty
constructor(override val propertyName: String, override val extractor: (Profile) -> String?) : ModelBackedTargetProperty<Profile>
```
we can just add another parameter like `String? -> boolean`

You could also add another interface for this like `ValidatingTargetProperty<Profile>`.

The downside is we have to make everything into a string.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
-      throw new BadRequestException(String.format("%s cannot be left blank!", fieldName));
+  /**
+   * Returns a list of PropertyChange entries from a diff that match either a field or any of its
+   * subfields. For example, a pathPrefix of 'address' will match a field change for both 'address'
+   * and for 'address.zipCode'.
+   */
+  private List<Change> getChangesWithPrefix(final Diff diff, final String pathPrefix) {
+    return diff.getChanges(
+        change -> {
+          if (change instanceof PropertyChange) {
```
jaycarlton, line 359: Why not use `&&` after the `instanceof` check?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
+    if (!getChangesWithPrefix(diff, "areaOfResearch").isEmpty() || isNewObject) {
+      validateAreaOfResearch(updatedProfile);
+    }
+    if (!getChangesWithPrefix(diff, "verifiedInstitutionalAffiliation").isEmpty() || isNewObject) {
+      validateInstitutionalAffiliation(updatedProfile);
+    }
+
+    if (!isNewObject) {
+      // We disallow changes in certain fields.
+      if (!getChangesWithPrefix(diff, "username").isEmpty()) {
```
jaycarlton, line 409: Can you make a list of Strings and iterate over tit since it's the same condition really? Better yet, can you simply tell Javers not to look at these fields?

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -31,22 +35,52 @@
...
+          .institutionShortName("Broad")
+          .institutionDisplayName("The Broad Institute")
+          .institutionalRoleEnum(InstitutionalRole.ADMIN);
+
+  private static final Profile VALID_PROFILE =
+      new Profile()
+          .username("jdoe123")
+          .contactEmail("jdoe123@gmail.com")
+          .address(
+              new Address()
```
jaycarlton, line 73: It's easier to interpret and maintain if the fields look more like real data. If they're all asdf, we won't catch errors that swap them.

It's also much easier to pick up how the class is supposed to be used if there's recognizable state in it.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -286,50 +287,132 @@ private void updateInstitutionalAffiliations(Profile updatedProfile, DbUser user
...
 
-  private void validateStringLength(String field, String fieldName, int max, int min) {
-    if (field == null) {
-      throw new BadRequestException(String.format("%s cannot be left blank!", fieldName));
+  /**
+   * Returns a list of PropertyChange entries from a diff that match either a field or any of its
+   * subfields. For example, a pathPrefix of 'address' will match a field change for both 'address'
+   * and for 'address.zipCode'.
+   */
+  private List<Change> getChangesWithPrefix(final Diff diff, final String pathPrefix) {
```
jaycarlton, line 356: Is there not a way to associate a schema with a Diff? It seems somewhat risky (in terms of unintended changes in behavior) to use reflection, though I can see the other argument.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -227,23 +235,16 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
...
-    // Validation steps, which yield a BadRequestException if errors are found.
-    String userName = profile.getUsername();
-    if (userName == null || userName.length() < 3 || userName.length() > 64) {
-      throw new BadRequestException(
-          "Username should be at least 3 characters and not more than 64 characters");
-    }
-    validateStringLength(profile.getGivenName(), "Given Name", 80, 1);
-    validateStringLength(profile.getFamilyName(), "Family Name", 80, 1);
-
+  public void cleanProfile(Profile profile) throws BadRequestException {
```
jaycarlton, line 238: I like the name change. It makes expect changes.

I would prefer a more immutable style, though, where we take in a profile and return a new one, especially since these aren't ORM objects.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: I thought of a couple of things to be careful about would be any dynamic properties that return different results every time or properties derived from other properties, which would be double-evaluated. Neiither are an issue for a simple codegen'd model class I suppose.

gjuggler:
> Tested locally:
> 
> * removed address from DB
> * attempted to Admin Update the user's Verified Institutional Affiliation
> * 500 in UI due to Address NPE

Can you try with the latest commit? I didn't realize we made assumptions of non-null address also in the updateProfileForUser method. Turns out that method had no unit tests at all (yikes). I'll add a more specific regression test at that level too.

gjuggler:
> > Tested locally:
> > 
> > * removed address from DB
> > * attempted to Admin Update the user's Verified Institutional Affiliation
> > * 500 in UI due to Address NPE
> 
> Can you try with the latest commit? I didn't realize we made assumptions of non-null address also in the updateProfileForUser method. Turns out that method had no unit tests at all (yikes). I'll add a more specific regression test at that level too.

Quick update: the last commit didn't seem to fix things. Still working on this, I'll ping you when I'm more confident it's ready.

aweng98:
Puppeteer test failed. UI changed, test fixed and merged yesterday. So rebase will resolve the test failure.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: Oh, I was thinking we only wrote fields that changed. nvm.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: We can  make a `cloneProfile()` easily with MapStruct. You just have a method that takes and returns a Profile.

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
```diff
@@ -55,12 +92,12 @@
   @MockBean
   private VerifiedInstitutionalAffiliationMapper mockVerifiedInstitutionalAffiliationMapper;
 
-  private static final FakeClock CLOCK = new FakeClock(Instant.parse("2000-01-01T00:00:00.00Z"));
+  @MockBean private VerifiedInstitutionalAffiliationDao mockVerifiedInstitutionalAffiliationDao;
```
jmthibault79, line 95: I often trip up on @MockBean vs @Autowired vs @Import ... do we have a doc?

I have probably asked this before.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -53,6 +60,7 @@
   private final UserTermsOfServiceDao userTermsOfServiceDao;
   private final VerifiedInstitutionalAffiliationDao verifiedInstitutionalAffiliationDao;
   private final VerifiedInstitutionalAffiliationMapper verifiedInstitutionalAffiliationMapper;
+  private final Javers javers;
```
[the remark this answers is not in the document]
gjuggler: I lurve alphabetization – dunno why I didn't notice we're doing that here.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -185,31 +210,30 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
     user.setAreaOfResearch(updatedProfile.getAreaOfResearch());
     user.setProfessionalUrl(updatedProfile.getProfessionalUrl());
     user.setAddress(addressMapper.addressToDbAddress(updatedProfile.getAddress()));
-    user.getAddress().setUser(user);
+    // Address may be null for users who were created before address validation was in place. See
+    // RW-5139.
+    Optional.ofNullable(user.getAddress()).ifPresent(address -> address.setUser(user));
```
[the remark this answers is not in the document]
gjuggler: I don't think this quite works since the ifPresent subject is what we want to call the method on, not the argument.

api/src/test/java/org/pmiops/workbench/profile/ProfileServiceTest.java
gjuggler: Eek, yeah it would be great to have some docs on this. I've definitely developed an aesthetic that seems to work reliably for most cases, it'll be good to capture that. Tracked in https://precisionmedicineinitiative.atlassian.net/browse/RW-5159

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jmthibault79: oops.  yes.

PR #3737, opened 2020-07-01
[risk=no][rw-4945] Add URL validation to account creation and profile

petesantos:
Description:

Add validation to the professional URL field in the account creation and profile pages.
I also updated the error messages in the profile page when the save button is disabled to enumerate the errors as it done on other pages.

I looked into URL sanitization, but could not find a library I was satisfied with and also feel some requirements around sanitization should be defined - not only for URL's but all inputs. Also, it is not clear to me what we want to guard against on the client side (URL or HTML sanitization?) and how we want to sanitize (e.g. should we immediately change the input on a change? On a blur? Or not update the field, but save the sanitized value? Will there be server side sanitization, and if so should we converge on a library? etc...)

I added two additional tech debt tickets:
RW-5160 Converge on validation error handling
RW-5161 Spike on user input validation

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/login/account-creation/account-creation.tsx
+++ b/ui/src/app/pages/login/account-creation/account-creation.tsx
@@ -410,14 +410,18 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
       }
     }
 
-    const urlResult = validationData.professionalUrl ? validate({website: validationData.professionalUrl}, {
-      website: {
-        url: {
-          message: '^Proffesional URL %{value} is not a valid URL'}
-      }
-    }) : {};
+    const urlError = validationData.professionalUrl
+      ? validate({website: validationData.professionalUrl}, {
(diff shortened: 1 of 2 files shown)


ui/src/app/pages/login/account-creation/account-creation.tsx
```diff
@@ -409,14 +409,26 @@ export class AccountCreation extends React.Component<AccountCreationProps, Accou
...
+
+    return fp.isEmpty(errors) ? undefined : errors;
   }
 
   render() {
     const {
       profile: {
         givenName, familyName,
-        contactEmail, username, areaOfResearch, professionalUrl,
+        contactEmail, username, areaOfResearch, professionalUrl = '',
```
petesantos, line 431: Defaulting `professionalUrl` prevents a react error from showing up in the console from having an input field initialized as undefined

PR #3738, opened 2020-07-01
[RW-5062][risk=no] Implement full DataSetMapper

freemabd:
This PR addresses:
1. Implementing `DataSetMapper` to allow full db model to client conversion
2. Removing `DataSetController#TO_CLIENT_DATA_SET` and `DataSetController#TO_CLIENT_DOMAIN_VALUE ` functions and using `DataSetMapper`
3. Implementing `DataSetMapperTest`
4. Updating all affected test classes that use `DataSetMapper`
5. Creation of `BigQueryTableInfo` enum to replace `ConceptSetDao#DOMAIN_TO_TABLE_NAME`

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/cohorts/CohortService.java
+++ b/api/src/main/java/org/pmiops/workbench/cohorts/CohortService.java
@@ -1,19 +1,28 @@
 package org.pmiops.workbench.cohorts;
 
 import java.util.List;
+import java.util.stream.Collectors;
 import org.pmiops.workbench.db.dao.CohortDao;
 import org.pmiops.workbench.db.model.DbCohort;
+import org.pmiops.workbench.model.Cohort;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service
 public class CohortService {
   private CohortDao cohortDao;
+  private CohortMapper cohortMapper;
 
-  public CohortService(CohortDao cohortDao) {
+  @Autowired
+  public CohortService(CohortDao cohortDao, CohortMapper cohortMapper) {
     this.cohortDao = cohortDao;
+    this.cohortMapper = cohortMapper;
   }
 
-  public List<DbCohort> findAll(List<Long> cohortIds) {
-    return (List<DbCohort>) cohortDao.findAll(cohortIds);
+  public List<Cohort> findAll(List<Long> cohortIds) {
+    return ((List<DbCohort>) cohortDao.findAll(cohortIds))
+        .stream()
+            .map(dbCohort -> cohortMapper.dbModelToClient(dbCohort))
+            .collect(Collectors.toList());
   }
 }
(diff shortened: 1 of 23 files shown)


api/src/main/java/org/pmiops/workbench/cohorts/CohortService.java
```diff
@@ -1,19 +1,28 @@
...
     this.cohortDao = cohortDao;
+    this.cohortMapper = cohortMapper;
   }
 
-  public List<DbCohort> findAll(List<Long> cohortIds) {
-    return (List<DbCohort>) cohortDao.findAll(cohortIds);
+  public List<Cohort> findAll(List<Long> cohortIds) {
+    return ((List<DbCohort>) cohortDao.findAll(cohortIds))
+        .stream()
+            .map(dbCohort -> cohortMapper.dbModelToClient(dbCohort))
```
NehaBroad, line 25: Can this be .map(cohortMapper::dbModelToClient)

api/src/test/java/org/pmiops/workbench/dataset/DataSetMapperTest.java
```diff
@@ -0,0 +1,194 @@
...
+    assertThat(dbDataset.getLastModifiedTime().toInstant().toEpochMilli())
+        .isEqualTo(dataSet.getLastModifiedTime());
+    assertThat(dbDataset.getPrePackagedConceptSet())
+        .isEqualTo(
+            DbStorageEnums.prePackagedConceptSetsToStorage(dataSet.getPrePackagedConceptSet()));
+  }
+
+  private void assertDbModelToClient(
+      DataDictionaryEntry dataDictionaryEntry, DbDataDictionaryEntry dbDataDictionaryEntry) {
+    assertThat(dbDataDictionaryEntry.getCdrVersion().getCdrVersionId())
```
NehaBroad, line 173: Rather than individual elements since DataDictionaryEntry has equal and hashcode can we just have assertThat(dbDataDictionaryEntry).isEqual(dataDictionaryEntry)?

api/src/main/java/org/pmiops/workbench/cohorts/CohortService.java
freemabd: Done.

api/src/test/java/org/pmiops/workbench/dataset/DataSetMapperTest.java
freemabd: dbDataDictionaryEntry and dataDictionaryEntry are different types. The equal and hashcode defined in DataDictionaryEntry is only for comparison to objects of type DataDictionaryEntry not DbDataDictionaryEntry

NehaBroad: Oh man i just missed that :D i thought its the same type DataDictionaryEntry

PR #3743, opened 2020-07-03
[no ticket][risk=no] Puppeteer new Concept Set tests

aweng98:
Add two Concept Set tests:
* Add to existing Concept Set in same workspace.
* Copy Concept Set to another workspace.


e2e/app/text-labels.ts
```diff
@@ -13,43 +13,47 @@ export enum WorkspaceAccessLevel {
 }
 
 export enum EllipsisMenuAction {
```
aweng98, line 15: Alphabetize list.

e2e/app/page/data-page.ts
```diff
@@ -162,8 +170,11 @@ export default class DataPage extends AuthenticatedPage {
    * @param {string} conceptsetName
    */
   async deleteConceptSet(conceptsetName: string): Promise<string> {
-    const datasetCard = await DataResourceCard.findCard(this.page, conceptsetName);
-    const menu = datasetCard.getEllipsis();
+    const card = await DataResourceCard.findCard(this.page, conceptsetName);
```
dolbeew, line 173: nit: can we call this something like `conceptSetCard` to be consistent with how the other cards are named?

aweng98: Done. Thank you.

PR #3747, opened 2020-07-06
[RW-5162][risk=no] creating new criteria generation script and adding in ICD9 codes

jscherdin:
- this creates a new criteria generation scripts
- updating ICD9 source codes and adds in the new way of generating counts

### DIFF
--- a/api/db-cdr/generate-cdr/make-bq-criteria-tables.sh
+++ b/api/db-cdr/generate-cdr/make-bq-criteria-tables.sh
@@ -1,0 +1,534 @@
+#!/bin/bash
+
+# This generates the criteria tables for the CDR
+
+# PREP: upload all prep tables
+
+# ./project.rb generate-cb-criteria-tables --bq-project aou-res-curation-output-prod --bq-dataset SR2019q4r3
+
+set -ex
+
+export BQ_PROJECT=$1        # project
+export BQ_DATASET=$2        # dataset
+export DATA_BROWSER=$3      # data browser flag
+export DRY_RUN=$4           # dry run
+
+if [ "$DRY_RUN" == true ]
+then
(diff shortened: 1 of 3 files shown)


api/db-cdr/generate-cdr/make-bq-criteria-tables.sh
```diff
@@ -0,0 +1,550 @@
+#!/bin/bash
```
freemabd, line 1: Are replacing `generate-cb-criteria-tables.sh` with this file? If so should we delete `generate-cb-criteria-tables.sh`?

jscherdin: It will eventually replace it. It is a slow re-write. Once this new file is complete, we can drop the old one

api/libproject/devstart.rb
```diff
@@ -1007,7 +1007,7 @@ def generate_cb_criteria_tables(cmd_name, *args)
   op.parse.validate
 
   common = Common.new
-  common.run_inline %W{docker-compose run --rm db-make-bq-tables ./generate-cdr/generate-cb-criteria-tables.sh #{op.opts.bq_project} #{op.opts.bq_dataset} #{op.opts.data_browser} #{op.opts.dry_run}}
+  common.run_inline %W{docker-compose run --rm db-make-bq-tables ./generate-cdr/make-bq-criteria-tables.sh #{op.opts.bq_project} #{op.opts.bq_dataset} #{op.opts.data_browser} #{op.opts.dry_run}}
```
freemabd, line 1010: Ok, if it's a slow rewrite should we revert this change in case we have to regenerate tables in prod or synth?

jscherdin: yeah. I will make that change

PR #3748, opened 2020-07-06
[risk=no] Fix mounted paths for CDR building tools

calbach:


calbach:
I'm not quite sure about how to test these without disrupting cloud resources. Let me know if you have a sample CLI I can try here.

calbach:
Thanks!

Ignoring puppeteer failure as this change has no interaction with that test.

PR #3753, opened 2020-07-07
[noTicket][RISK=LOW] Use config flag in UI to show Research Review Prompt

NehaBroad:
As part of the story https://precisionmedicineinitiative.atlassian.net/browse/RW-2400
A prompt to review research purpose is displayed if the workspace attribute needs_research_purpose_prompt is set to true.

This PR adds Flag check in UI to restrict the prompt on PROD/STABLE/STAGING/PERF enviornment

---
**PR checklist**

### DIFF
--- a/ui/src/app/guards/workspace-guard.service.ts
+++ b/ui/src/app/guards/workspace-guard.service.ts
@@ -14,7 +14,6 @@ export class WorkspaceGuard implements CanActivate, CanActivateChild {
     private router: Router) {}
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
-    console.log(serverConfigStore.getValue().enableResearchReviewPrompt);
     if (serverConfigStore.getValue().enableResearchReviewPrompt && route.routeConfig.path === 'data' ||
         route.routeConfig.path === 'notebooks') {
       this.workspaceService.getWorkspace(route.params.ns, route.params.wsid).subscribe(workspace => {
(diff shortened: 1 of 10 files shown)


ui/src/app/guards/workspace-guard.service.ts
```diff
@@ -13,7 +14,8 @@ export class WorkspaceGuard implements CanActivate, CanActivateChild {
     private router: Router) {}
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
-    if (route.routeConfig.path === 'data' ||
+    console.log(serverConfigStore.getValue().enableResearchReviewPrompt);
```
calbach, line 17: rm

ui/src/app/guards/workspace-guard.service.ts
```diff
@@ -13,7 +14,8 @@ export class WorkspaceGuard implements CanActivate, CanActivateChild {
...
+    if (serverConfigStore.getValue().enableResearchReviewPrompt && route.routeConfig.path === 'data' ||
```
calbach, line 18: Is this the right if statement? Seems like this should be further down @ line 21

NehaBroad: Since enableResearchReviewPrompt is false there is no point even to have the api call this.workspaceService.getWorkspace to check for accessLevel or anything and just return 
Observable.from([true])

calbach: I see - I didn't realize this was a new guard added only for this new review prompt feature. Lets discuss more, I have performance concerns with this approach and I think there may be some inconsistent behavior, but it's all feature flagged with this PR so it's not urgent.

PR #3754, opened 2020-07-08
[RW-5013][risk=no] Recycle backfill script for owner fix

calbach:
Backfill script to fix desynchronized workspace / billing project ACLs. We encountered one bug which caused a bunch of known workspaces with this issue, though theoretically this issue can also occur during normal operation.

Required pulling in the billing project members Firecloud API from orchestration.

Removed a highly similar script while I was here, reusing much of the same code.


api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
```diff
@@ -0,0 +1,202 @@
...
+      // ShareWorkspace first updates the Workspace ACL, then the billing project role. For this
+      // reason, the workspace ACL is the source of truth, as it will have been set properly in the
+      // event of a partial sharing failure.
+      for (String user : Sets.difference(billingProjectOwners, workspaceOwners)) {
+        dryLog(
+            dryRun,
+            String.format(
+                "'%s': '%s', inconsistency: workspace role '%s', project role '%s'",
+                w.getNamespace(), user, workspaceRoles.get(user), billingProjectRoles.get(user)));
+        if (billingProjectOwners.contains(user)) {
```
ericsong, line 143: wouldn't this always be true if `user` is in the set of `billingProjectOwners` - `workspaceOwners`? `user` is always an element within `billingProjectOwners` which is not in `workspaceOwners`

api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
```diff
@@ -0,0 +1,202 @@
...
+              billingApi.removeUserFromBillingProject(w.getNamespace(), "owner", user);
+            } catch (ApiException e) {
+              log.log(Level.WARNING, "failed to remove user from project", e);
+            }
+          }
+          dryLog(
+              dryRun,
+              String.format("removed user '%s' from billing project '%s'", user, w.getNamespace()));
+          ownersRemoved++;
+        } else {
```
ericsong, line 156: re: above. I believe this case will never be hit. I think what the logic covers is the set difference of `workspaceOwners` - `billingProjectOwners`

api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
calbach: Great catch - thanks. There were two bugs here:

1. I accidentally imported from the Elasticsearch client library... that was not intentional. Should have been Guava.
2. I'd thought `Sets.difference` was the symmetric difference. Fixed to actually call the `symmetricDifference`, which was my intent.

calbach:
Added one final change - the empty list handling on the Java flag did not work as I was hoping. I won't block on reapproval given the size of the change but feel free to comment if you get the chance.

calbach:
Note: there are still some non-1ppw workspaces hanging around. These are causing the script to thrash on add/remove, so the recommended usage for now is --dry-run, then apply to targeted billing project IDs

PR #3758, opened 2020-07-08
[RW-5150][risk=no] removing answer_concept_id from survey query

freemabd:
removing answer_concept_id from survey query

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/dataset/DataSetServiceImpl.java
+++ b/api/src/main/java/org/pmiops/workbench/dataset/DataSetServiceImpl.java
@@ -474,7 +474,7 @@ public class DataSetServiceImpl implements DataSetService, GaugeDataCollector {
       List<DbConceptSet> conceptSetsSelected,
       String cohortQueries,
       CdrBigQuerySchemaConfig bigQuerySchemaConfig) {
-    //    validateConceptSetSelection(domain, conceptSetsSelected);
+    validateConceptSetSelection(domain, conceptSetsSelected);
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/pmiops/workbench/dataset/DataSetServiceImpl.java
```diff
@@ -515,16 +515,26 @@ private QueryJobConfiguration buildQueryJobConfigForDomain(
...
-          clause ->
-              queryBuilder
-                  .append(" WHERE \n(")
-                  .append(domainConceptIdInfo.getStandardConceptIdColumn())
-                  .append(clause)
-                  .append(" OR \n")
-                  .append(domainConceptIdInfo.getSourceConceptIdColumn())
-                  .append(clause)
-                  .append(")"));
+      if (Domain.SURVEY.equals(domain)) {
```
NehaBroad, line 518: Can we have it like :
   if (conceptSetSqlInClauseMaybe.isPresent()) {
        queryBuilder
            .append(" WHERE \n(")
            .append(domainConceptIdInfo.getStandardConceptIdColumn())
            .append(conceptSetSqlInClauseMaybe.get());
        if (Domain.SURVEY.equals(domain)) {
          queryBuilder.append(" OR \n")
              .append(domainConceptIdInfo.getSourceConceptIdColumn())
              .append(conceptSetSqlInClauseMaybe.get());
        }
        queryBuilder.append(")");

      }

To avoid duplication of code?

freemabd: Changed to remove any duplication.

PR #3759, opened 2020-07-08
[RW-5240][risk=no] Puppeteer Workspace Edit test

aweng98:
Editing Workspace test.

### DIFF
--- a/e2e/app/page/workspace-about-page.ts
+++ b/e2e/app/page/workspace-about-page.ts
@@ -69,4 +69,26 @@ export default class WorkspaceAboutPage extends AuthenticatedPage{
     await modal.waitUntilVisible();
     return modal;
   }
+
+  async getCdrVersion(): Promise<string> {
+    const cdrVerXpath = '//*[@data-test-id="cdrVersion"]';
+    const cdr = await this.page.waitForXPath(cdrVerXpath, {visible: true});
+    const cdrValue = await (await cdr.getProperty('innerText')).jsonValue();
+    return cdrValue.toString();
+  }
+
+  async getCreationDate(): Promise<string> {
+    const creationDateXpath = '//*[@data-test-id="creationDate"]/*[last()]';
+    const creationDate = await this.page.waitForXPath(creationDateXpath, {visible: true});
+    const dateValue = await (await creationDate.getProperty('innerText')).jsonValue();
+    return dateValue.toString();
+  }
+
+  async getLastUpdatedDate(): Promise<string> {
+    const lastUpdatedDateXpath = '//*[@data-test-id="lastUpdated"]/*[last()]';
+    const lastUpdatedDate = await this.page.waitForXPath(lastUpdatedDateXpath, {visible: true});
+    const dateValue = await (await lastUpdatedDate.getProperty('innerText')).jsonValue();
+    return dateValue.toString();
+  }
+
 }
--- a/e2e/jest-puppeteer.config.js
+++ b/e2e/jest-puppeteer.config.js
@@ -5,7 +5,7 @@
 
 const fp = require('lodash/fp');
 const puppeteer = require('puppeteer');
-const isHeadless = process.env.HEADLESS !== 'false';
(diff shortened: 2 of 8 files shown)


e2e/app/page/workspace-about-page.ts
```diff
@@ -69,4 +69,26 @@ export default class WorkspaceAboutPage extends AuthenticatedPage{
     await modal.waitUntilVisible();
     return modal;
   }
+
+  async getCdrVersion(): Promise<string> {
+    const cdrVerXpath = '//*[@data-test-id="cdrVersion"]';
+    const cdr = await this.page.waitForXPath(cdrVerXpath, {visible: true});
+    const cdrValue = await (await cdr.getProperty('innerText')).jsonValue();
```
jmthibault79, line 76: why double await ?

e2e/jest-puppeteer.config.js
```diff
@@ -5,7 +5,7 @@
 
 const fp = require('lodash/fp');
 const puppeteer = require('puppeteer');
-const isHeadless = process.env.HEADLESS !== 'false';
+const isHeadless = (process.env.HEADLESS || 'true') === 'true';
```
jmthibault79, line 8: so we're defaulting to Headless if this value is missing.  is that what we want?

e2e/tests/workspace/workspace-edit.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+   * - Verify landing in DATA tab.
+   * - Verify Workspace Information in ABOUT tab.
+   */
+  test('User as OWNER can edit workspace', async () => {
+    const workspaceCard = await findWorkspace(page);
+    await (workspaceCard.getEllipsis()).clickAction(EllipsisMenuAction.Edit);
+
+    const workspacesPage = new WorkspacesPage(page);
+    // Change question #2 answer
+    await performActions(page, testData.defaultAnswersResearchPurposeSummary);
```
jmthibault79, line 28: how does this work?

e2e/app/page/workspace-about-page.ts
aweng98: both are async calls. the inner call has to finish first before outer call starts. I could write it differently with `.then()` (promises chaining) if you think that's a better to read.

e2e/jest-puppeteer.config.js
aweng98: yes, that's the intention. it's like that since day 1. I changed code to make it explicitly clear.

e2e/tests/workspace/workspace-edit.spec.ts
aweng98: Basically, iterate thru a json object consisted of elements and performs action (type/check/select) according to the type (textinput/checkbox/radiobutton) of element. (saves me from dupl. code)

```
export const defaultAnswersResearchPurposeSummary = [
  {
    id: FIELD.RESEARCH_PURPOSE_SUMMARY.scientificQuestionsIntentToStudyTextarea,
    value: makeString(100)
  }, {
    id: FIELD.RESEARCH_PURPOSE_SUMMARY.scientificApproachesToUseTextarea,
    value: makeString(100)
  }, {
    id: FIELD.RESEARCH_PURPOSE_SUMMARY.anticipatedFindingsFromStudyTextarea,
    value: makeString(100)
  }
];
```

e2e/app/page/workspace-about-page.ts
jmthibault79: OK

No need to change

e2e/jest-puppeteer.config.js
jmthibault79: great!

e2e/tests/workspace/workspace-edit.spec.ts
jmthibault79: oh ok - it fills in those 3 text fields.  cool.

aweng98:
Thank you.

PR #3763, opened 2020-07-08
[no ticket][risk=no] Puppeteer enable rename dataset test code

aweng98:
Bug https://precisionmedicineinitiative.atlassian.net/browse/RW-5079 fixed.

### DIFF
--- a/e2e/app/element/base-element.ts
+++ b/e2e/app/element/base-element.ts
@@ -151,10 +151,33 @@ export default class BaseElement extends Container {
       });
   }
 
-  async type(text: string, options?: { delay: number }): Promise<this> {
-    await this.focus();
-    await this.clear();
-    await this.asElementHandle().then((handle: ElementHandle) => handle.type(text, options));
+  /**
+   * Clear existing value in textbox then type new text value.
+   * @param {string} The texts string.
+   * @param options The typing options.
+   */
+  async type(textValue: string, options?: { delay: number }): Promise<this> {
+    const clearAndType =  async (txt: string, opts?: { delay: number }): Promise<string> => {
+      await this.click();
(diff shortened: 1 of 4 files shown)


e2e/app/element/base-element.ts
```diff
@@ -151,10 +151,33 @@ export default class BaseElement extends Container {
...
-  async type(text: string, options?: { delay: number }): Promise<this> {
-    await this.focus();
-    await this.clear();
-    await this.asElementHandle().then((handle: ElementHandle) => handle.type(text, options));
+  /**
+   * Clear existing value in textbox then type new text value.
+   * @param {string} The texts string.
+   * @param options The typing options.
+   */
+  async type(textValue: string, options?: { delay: number }): Promise<this> {
```
aweng98, line 159: While running `dataset-create.spec` to make sure it runs okay, I observed `clear()` func acts flaky. Old value in text input was not always cleared before type new text. 
My fix is to add one retry.

e2e/app/element/base-element.ts
```diff
@@ -177,12 +200,26 @@ export default class BaseElement extends Container {
...
+  async clear(options: ClickOptions = { clickCount: 3 }): Promise<void> {
     const elemt = await this.asElementHandle();
-    await elemt.click({ clickCount: 3 });
-    await elemt.press('Backspace');
+    await elemt.focus();
+    await elemt.click(options);
+    await this.page.keyboard.press('Backspace');
+  }
+
+  async clearTextInput(): Promise<void> {
```
aweng98, line 212: clearText alternative: injecting page function.

aweng98:
Thank you.

PR #3764, opened 2020-07-09
[risk=no] Include timestamp in metrics export

calbach:
Per Karthik's request, include timestamp. I've just hardcoded it to CST since I didn't see a clear way to handle time zones in Google sheets. Karthik confirmed this was fine.


calbach:
Great suggestion - done.

PR #3768, opened 2020-07-09
[no ticket][risk=no] Puppeteer: Rework jest.test-setup file

aweng98:


e2e/jest.test-setup.ts
```diff
@@ -1,35 +1,37 @@
...
-    // to improve page load performance, block requests if it is not for application functions.
+    // to improve page load performance, block network requests unrelated to application.
     try {
       if (host === 'www.google-analytics.com'
-       || host === 'accounts.youtube.com'
-       || host === 'static.zdassets.com'
-       || host === 'play.google.com'
-       || request.url().endsWith('content-security-index-report')) {
-        await request.abort()
+         || host === 'accounts.youtube.com'
```
jaycarlton, line 30: I think you can use `fp.includes()` to see if the host in in one of the names you have here.

e2e/jest.test-setup.ts
```diff
@@ -39,7 +41,15 @@ beforeAll(async () => {
...
 });
 
-// Runs this afterAll() before test's afterAll().
-afterAll(async () => {
+/**
+ * At the end of each test completion, do:
+ * - Disable network interception.
+ * - Delete broswer cookies.
+ * - Reset global page and browser variables.
+ */
```
jaycarlton, line 49: Do you also need an afterAll? Or would that not let you inspect the browser after a failure?

aweng98: I don't think need `afterAll` at this time. We don't set up local server or DB anything like that.

e2e/jest.test-setup.ts
aweng98: That's a good tip. I'll test out in future PR.

PR #3770, opened 2020-07-10
[no ticket][risk=no] circleci config change for pupppeteer-e2e-test job

aweng98:
skip check has a bug - job were skipped when PR contains changed `.sh`, `.rb` files. I need time to find solution. remove the check in meantime.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -560,7 +560,6 @@ jobs:
     parallelism: 3
     steps:
       - checkout_init_git
-      - halt_if_code_unchanged
       - browser-tools/install-browser-tools
       - run_e2e_test
 


.circleci/config.yml
```diff
@@ -560,7 +560,6 @@ jobs:
     parallelism: 3
     steps:
       - checkout_init_git
-      - halt_if_code_unchanged
```
calbach, line 563: what about the check on puppeteer local? Isn't that more of a concern here?

aweng98: I'm sure what went wrong exactly. `puppeteer-e2e-local-ui` job has been running okay but not for the `puppeteer-e2e-test` job.

calbach: If we're running on the master branch, we shouldn't be skipping.

When this runs on master commits, it will always generate 0 diffs:
`git diff --name-only $(git merge-base origin/master ${CIRCLE_BRANCH}) `

Previously, the "halt" command would first check whether this branch was master or a release - if so, don't skip. I don't know when that was removed, but it should probably be added back. This change achieves the same thing, though indirectly (since this job *only* runs on master merge).

aweng98:
Sorry, it was me removed that check in this job because this job only runs on master merge (checked in workflow).
Thank you.

PR #3774, opened 2020-07-10
[RW-5013][risk=no] Backfill tool fixes: id lists and 1ppw projects

calbach:
- Fix yet another issue with the "id list" flag approach
- Filter out 1ppw projects. These cause thrashing with the tool, as it can lead to  adding/removing the same person from a billing project repeatedly (one project may be encountered multiple times during iteration).

This tool has already run successfully, these changes are included for posterity.

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
@@ -97,7 +97,7 @@ public class FixDesynchronizedBillingProjectOwners {
         workspacesApi.listWorkspaces(FIRECLOUD_LIST_WORKSPACES_REQUIRED_FIELDS);
     log.info(String.format("found %d workspaces", workspaces.size()));
 
-    // 1PPW workspaces still exist and need to be filtered out - this script depends on the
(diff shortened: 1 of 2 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/FixDesynchronizedBillingProjectOwners.java
```diff
@@ -91,8 +92,27 @@ private static void clean(
...
     int ownersRemoved = 0;
     int ownersAdded = 0;
-    for (FirecloudWorkspaceResponse resp :
-        workspacesApi.listWorkspaces(FIRECLOUD_LIST_WORKSPACES_REQUIRED_FIELDS)) {
+
+    List<FirecloudWorkspaceResponse> workspaces =
+        workspacesApi.listWorkspaces(FIRECLOUD_LIST_WORKSPACES_REQUIRED_FIELDS);
+    log.info(String.format("found %d workspaces", workspaces.size()));
+
+    // 1PPW workspaces still exist and need to be filtered out - this script depends on the
```
jmthibault79, line 100: non-1PPW, rather

calbach: Fixed, thanks

PR #3777, opened 2020-07-13
[RW-5111][risk=no] GA footer, support tracker callback on links

calbach:
- Add GA tracking on footer links
- Support an analytics callback function on `StyledAnchorTag`
- Renamed GA enum to follow more standard naming conventions (originally, I'd thought I might export this.. happy to revert if this is too much for the PR)

Alternatives considered on property name:
- `onClick`: this seemed to ambiguous in combination with the standard onclick function (which runs first, etc)
- `sideEffect`: there's nothing GA specific about the callback parameter. However, this also left things a bit ambiguous as to when the side effect might actually run and I don't have any other ideas about what else could use this callback

### DIFF
--- a/ui/src/app/components/buttons.tsx
+++ b/ui/src/app/components/buttons.tsx
@@ -303,13 +303,16 @@ export const Link = ({disabled = false, style = {}, children, ...props}) => {
   >{children}</Clickable>;
 };
 
-export const StyledAnchorTag = ({href, style = {}, children, ...props}) => {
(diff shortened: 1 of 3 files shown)


calbach:
Oops, missed the unregistered links. I've fixed those, and also updated some of the labels (specifically, behavior is different for the "contact us" link when signed out vs signed in). Confirming the new labels with product.

calbach:
Snyk complain is due to lack of rebasing. Going to ignore - confirmed this is the only complaint.

Confirmed this approach with Karthik/Shimon .

PR #3780, opened 2020-07-13
[risk=no] Add some initial React style guidance around function components

calbach:


ui/docs/style-guide.md
```diff
@@ -65,6 +65,11 @@ We are working towards abstracting styles (and possibly components) where it mak
     - Declare all exports at the bottom of the file so there is one place to look at to determine what a file is exposing versus having to scan the file for export statements
     - Only export something when it must be referenced by something externally. Try to limit what gets exported.
     - Use aliases to keep variables namespaced appropriately to their scope. Ex. Props can just be Props in this file but can be prefixed with the class name when it is exposed publicly.
+    - Prefer function components for simple, stateless components: [example](
```
petesantos, line 68: Is this meant as: 
1. when creating a simple, stateless components use function components
or 
2. Try to keep your components simple and with as little state as possible and function components can help with this?

calbach: It was the former. I rephrased to match this.

Though I do agree with (2) as well. I think it would be worth eventually adding a section here which covers best practices like minimizing state and potentially ties in React hooks too, but my focus was more narrow with this PR.

petesantos: Agreed - I would like to have a conversation with the team about function vs class components and best practices at some point

PR #3781, opened 2020-07-14
[RW-5166][risk=no] inject DataSetMapper into DataSetServiceImpl

freemabd:
This PR addresses:
1. moving DataSetMapper from DataSetController to DataSetServiceImpl
2. moving DataDictionaryEntryDao from DataSetController to DataSetServiceImpl
3. updating affected test cases.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/DataSetController.java
@@ -10,10 +10,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import java.sql.Date;
-import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
-import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -30,17 +28,11 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.pmiops.workbench.cdr.CdrVersionService;
 import org.pmiops.workbench.dataset.BigQueryTableInfo;
-import org.pmiops.workbench.dataset.DataSetMapper;
 import org.pmiops.workbench.dataset.DataSetService;
 import org.pmiops.workbench.dataset.DatasetConfig;
-import org.pmiops.workbench.db.dao.DataDictionaryEntryDao;
 import org.pmiops.workbench.db.model.DbCdrVersion;
-import org.pmiops.workbench.db.model.DbDataDictionaryEntry;
-import org.pmiops.workbench.db.model.DbDataset;
 import org.pmiops.workbench.db.model.DbUser;
-import org.pmiops.workbench.db.model.DbWorkspace;
 import org.pmiops.workbench.exceptions.BadRequestException;
-import org.pmiops.workbench.exceptions.ConflictException;
 import org.pmiops.workbench.exceptions.NotFoundException;
 import org.pmiops.workbench.firecloud.FireCloudService;
 import org.pmiops.workbench.firecloud.model.FirecloudWorkspaceResponse;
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/api/DataSetController.java
```diff
@@ -124,18 +107,14 @@
...
+            .getWorkspaceId();
     dataSetRequest.setWorkspaceId(workspaceId);
-    DbDataset datasetToSave = dataSetMapper.dataSetRequestToDb(dataSetRequest, null);
-    datasetToSave.setCreationTime(now);
-    datasetToSave.setCreatorId(userProvider.get().getUserId());
-    datasetToSave.setInvalid(false);
-    DbDataset savedDataSet = dataSetService.saveDataSet(datasetToSave);
-    return ResponseEntity.ok(dataSetMapper.dbModelToClient(savedDataSet));
+    return ResponseEntity.ok(
+        dataSetService.saveDataSet(dataSetRequest, userProvider.get().getUserId()));
```
NehaBroad, line 117: Not part of this story, but we should add validation for createRequest to be sure we have cohort/conceptsets and values information in data set Request

api/src/main/java/org/pmiops/workbench/dataset/DataSetService.java
```diff
@@ -37,11 +44,13 @@ DbDataset cloneDataSetToWorkspace(
 
   List<DbCohort> getCohortsForDataset(DbDataset dataSet);
 
-  List<DbDataset> getDataSets(ResourceType resourceType, long resourceId);
+  List<DataSet> getDataSets(ResourceType resourceType, long resourceId);
```
NehaBroad, line 47: Should this also be Optional?

api/src/main/java/org/pmiops/workbench/dataset/DataSetServiceImpl.java
```diff
@@ -683,19 +732,18 @@ public DbDataset cloneDataSetToWorkspace(
...
-  public Optional<DbDataset> getDbDataSet(DbWorkspace dbWorkspace, Long dataSetId) {
-    return Optional.of(dataSetDao.findOne(dataSetId));
+  public Optional<DataSet> getDbDataSet(Long dataSetId) {
+    return Optional.of(dataSetMapper.dbModelToClient(dataSetDao.findOne(dataSetId)));
   }
 
   @Override
   public void markDirty(ResourceType resourceType, long resourceId) {
-    List<DbDataset> dbDataSetList = getDataSets(resourceType, resourceId);
+    List<DbDataset> dbDataSetList = getDbDataSets(resourceType, resourceId);
```
NehaBroad, line 746: Again we can create another story: But should this throw an exception if no data sets are found for resourceType and Id ?

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
freemabd: Isn't that what `validateDataSetCreateRequest(dataSetRequest)` on line 109 does?

api/src/main/java/org/pmiops/workbench/dataset/DataSetService.java
freemabd: Not sure return type of `List` needs to be `Optional`... it either has items on it or it's empty... no chance of a `NullPointer`

api/src/main/java/org/pmiops/workbench/dataset/DataSetServiceImpl.java
freemabd: Good idea, we could also return `false` here instead of an exception -> https://github.com/all-of-us/workbench/blob/9c1af4c461c124a8ced075f168ce055d8ebeb405/api/src/main/java/org/pmiops/workbench/api/DataSetController.java#L381-L381

api/src/main/java/org/pmiops/workbench/api/DataSetController.java
NehaBroad: Ahh i missed that. Thanks!

PR #3785, opened 2020-07-14
[risk=medium][RW-4980] Add an Admin updateAccount endpoint to power new Admin UI

jmthibault79:
Description:

Add a new Admin endpoint to ProfileController which enables combinations of these actions:
* Auth module bypass (replicates existing functionality, but now allows multiple changes)
* Verified Institutional Affiliation update (replicates recent existing functionality)
* Update user contactEmail - new, validates against affiliation if it exists
* Update free tier dollar limit override - new, also re-enables the user's workspaces if appropriate

This PR is intended as a minimal change to unblock UI work.  Ideas about larger refactorings of Profile/Controller/Service should wait until after this is completed.  Those are coming soon.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -279,9 +279,8 @@ public Double getCachedFreeTierUsage(DbUser user) {
    * @return whether the user has remaining credits
    */
   public boolean userHasRemainingFreeTierCredits(DbUser user) {
-    return Optional.ofNullable(getCachedFreeTierUsage(user))
-        .map(usage -> compareCosts(getUserFreeTierDollarLimit(user), usage) > 0)
-        .orElse(true);
+    final double usage = Optional.ofNullable(getCachedFreeTierUsage(user)).orElse(0.0);
```
jmthibault79, line 282: This seems clearer

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -201,6 +223,7 @@ public void updateProfileForUser(DbUser user, Profile updatedProfile, Profile pr
 
     Timestamp now = new Timestamp(clock.instant().toEpochMilli());
 
+    user.setContactEmail(updatedProfile.getContactEmail());
```
jmthibault79, line 226: admins can change this, but users cannot.  See the validateProfile() methods family.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -231,6 +235,9 @@ public void setUp() throws IOException {
             .country(COUNTRY)
             .zipCode(ZIP_CODE));
 
+    // TODO: this needs to be set in createAccountAndDbUserWithAffiliation() instead of here.  Why?
```
jmthibault79, line 238: not a change, just an observation of the current state

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -1147,14 +1155,333 @@ public void testUpdateProfile_updateDemographicSurvey() {
...
 
     Profile result = profileController.createAccount(createAccountRequest).getBody();
+
+    // initialize the global test dbUser
     dbUser = userDao.findUserByUsername(PRIMARY_EMAIL);
+
+    // TODO: why is this necessary instead of initializing in setUp() ?
     dbUser.setEmailVerificationStatusEnum(EmailVerificationStatus.SUBSCRIBED);
-    userDao.save(dbUser);
+    dbUser = userDao.save(dbUser);
```
jmthibault79, line 1482: these two lines help with round-tripping consistency

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/FreeTierAuditor.kt
```diff
@@ -0,0 +1,5 @@
+package org.pmiops.workbench.actionaudit.auditors
+
+interface FreeTierAuditor {
```
jmthibault79, line 3: This was a fun exercise for Baby's First Kotlin, but it might be more appropriately part of User or Profile Auditor

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
```diff
@@ -11,5 +11,6 @@ enum class ActionType {
     DUPLICATE_TO,
     COLLABORATE,
     DELETE,
-    DETECT_HIGH_EGRESS_EVENT
+    DETECT_HIGH_EGRESS_EVENT,
+    FREE_TIER_DOLLAR_OVERRIDE
```
jmthibault79, line 15: not sure whether this makes sense as a ProfileTargetProperty or elsewhere

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/ProfileTargetProperty.kt
```diff
@@ -20,5 +20,6 @@ constructor(override val propertyName: String, override val extractor: (Profile)
     INSTITUTIONAL_AFFILIATIONS("institutional_affiliations",
             { it.institutionalAffiliations?.joinToString(", ") }),
     DEMOGRAPHIC_SURVEY("demographic_survey", { it.demographicSurvey?.toString() }),
-    ADDRESS("address", { it.address?.toString() });
+    ADDRESS("address", { it.address?.toString() }),
+    FREE_TIER_DOLLAR_QUOTA("free_tier_dollar_quota", { it.freeTierDollarQuota?.toString() });
```
jmthibault79, line 24: not sure whether this makes sense as a ProfileTargetProperty or elsewhere

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -159,6 +160,10 @@ public void validateInstitutionalAffiliation(Profile profile) {
           "Institutional role description cannot be empty when institutional role is set to Other");
     }
 
+    validateInstitutionalAffiliationAgainstEmail(profile);
+  }
+
+  private void validateInstitutionalAffiliationAgainstEmail(Profile profile) {
```
jmthibault79, line 166: broken out here so it can be called when either the VIA or the email change

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -187,11 +192,28 @@ public void validateInstitutionalAffiliation(Profile profile) {
...
+   * @param updatedProfile
+   * @param previousProfile
+   */
+  public void adminUpdateProfileForUser(
+      DbUser user, Profile updatedProfile, Profile previousProfile) {
+    updateProfileForUser(user, updatedProfile, previousProfile, true);
+  }
+
+  private void updateProfileForUser(
+      DbUser user, Profile updatedProfile, Profile previousProfile, boolean userIsAdmin) {
```
jmthibault79, line 211: I'm not super happy about the boolean flag here - open to ideas

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -381,47 +415,73 @@ private void validateAreaOfResearch(Profile profile) throws BadRequestException
...
    */
   @VisibleForTesting
-  public void validateProfile(@Nonnull Profile updatedProfile, @Nullable Profile prevProfile)
+  public void validateProfile(
+      @Nonnull Profile updatedProfile, @Nullable Profile previousProfile, boolean userIsAdmin) {
+
+    final boolean isNewProfile = previousProfile == null;
+    final Diff diff = javers.compare(previousProfile, updatedProfile);
+
+    validateProfileForCorrectness(updatedProfile, isNewProfile, diff);
```
jmthibault79, line 428: Split into "this is a correct Profile" vs "user or admin can change this" for clarity

api/src/test/java/org/pmiops/workbench/actionaudit/auditors/FreeTierAuditorTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.pmiops.workbench.actionaudit.TargetType
+import org.pmiops.workbench.actionaudit.targetproperties.ProfileTargetProperty
+import org.pmiops.workbench.db.model.DbUser
+import org.springframework.test.context.junit4.SpringRunner
+import java.time.Clock
+import java.time.Instant
+import javax.inject.Provider
+
+@RunWith(SpringRunner::class)
+class FreeTierAuditorTest {
```
jmthibault79, line 24: using ProfileAuditorTest as a model

api/src/test/java/org/pmiops/workbench/api/WorkspacesControllerTest.java
```diff
@@ -260,6 +260,8 @@
   @Autowired private CohortAnnotationDefinitionController cohortAnnotationDefinitionController;
   @Autowired private WorkspacesController workspacesController;
 
+  @MockBean FreeTierBillingService mockFreeTierBillingService;
```
jmthibault79, line 263: Because FreeTierBillingService now has a FreeTierAuditor dependency, this class needs to `@Import` it to run.  Instead, I took the opportunity to change this from a sorta-mock to a proper mock.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
jmthibault79: not currently.  My assumption is that these test classes will change quite a lot, so I plan to revisit then or soon after

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/FreeTierAuditor.kt
jmthibault79: sounds good

api/src/main/java/org/pmiops/workbench/actionaudit/targetproperties/ProfileTargetProperty.kt
jmthibault79: ok

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: That seems reasonable.  One snag here is that sometimes the user object is "without authorities" and I don't understand the implications of that well enough

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
```diff
@@ -579,6 +580,12 @@ private boolean userHasEverLoggedIn(
     return ResponseEntity.ok(new EmptyResponse());
   }
 
+  @Override
+  @AuthorityRequired({Authority.ACCESS_CONTROL_ADMIN})
+  public ResponseEntity<Profile> editUserInformation(EditUserInformationRequest request) {
+    return ResponseEntity.ok(profileService.editUserInformation(request));
```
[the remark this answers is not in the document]
jmthibault79: I'm undecided on whether to move `ACCESS_CONTROL_ADMIN` endpoints like these to a new FooAdminController/Service.  Certainly this will share the fate of others like it, whatever we decide on.  Before then, these will get pushed down to ProfileService/Impl like the others.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -381,47 +415,73 @@ private void validateAreaOfResearch(Profile profile) throws BadRequestException
...
-      if (!getChangesWithPrefix(diff, "username").isEmpty()) {
-        // See RW-1488.
-        throw new BadRequestException("Changing username is not supported");
-      }
-      if (!getChangesWithPrefix(diff, "contactEmail").isEmpty()) {
-        // See RW-1488.
-        throw new BadRequestException("Changing contact email is not currently supported");
-      }
+  private void validateChangesAllowedByAdmin(Diff diff) {
+    if (fieldChanged(diff, "username")) {
```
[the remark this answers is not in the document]
jmthibault79: could you clarify?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -432,10 +492,48 @@ public void validateProfile(@Nonnull Profile updatedProfile, @Nullable Profile p
...
+   *     validation.
+   * @return the Profile of the user, after updates
+   */
+  public Profile editUserInformation(EditUserInformationRequest request) {
+    final DbUser dbUser = userService.getByUsernameOrThrow(request.getUsername());
+    final Profile originalProfile = getProfile(dbUser);
+
+    Optional.ofNullable(request.getFreeCreditsLimit())
+        // only set override if it's different
+        .filter(newLimit -> !newLimit.equals(originalProfile.getFreeTierDollarQuota()))
```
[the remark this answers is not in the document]
jmthibault79: this is a different code path than the diffs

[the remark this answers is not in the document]
jmthibault79: There is also a subtlety here which is a bit complicated and would probably benefit from a longer comment.  I'll work on that.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -432,10 +492,48 @@ public void validateProfile(@Nonnull Profile updatedProfile, @Nullable Profile p
...
+        .ifPresent(
+            freeCreditsLimit ->
+                freeTierBillingService.setFreeTierDollarOverride(dbUser, freeCreditsLimit));
+
+    Optional.ofNullable(request.getAccessBypassRequests())
```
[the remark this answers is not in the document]
jmthibault79: it's not a required field in the JSON.  I can make it one, since an empty list has the same semantics as no list here.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -432,10 +492,48 @@ public void validateProfile(@Nonnull Profile updatedProfile, @Nullable Profile p
...
+        .ifPresent(
+            requests ->
+                requests.forEach(
+                    bypass -> userService.updateBypassTime(dbUser.getUserId(), bypass)));
+
+    // refetch from the DB
+    Profile updatedProfile = getProfile(userService.getByUsernameOrThrow(request.getUsername()));
+    Optional.ofNullable(request.getContactEmail()).ifPresent(updatedProfile::setContactEmail);
```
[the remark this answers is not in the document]
jmthibault79: this will get caught by validateContactEmail()

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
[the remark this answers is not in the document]
jaycarlton: I prefer a separate `UserAdminController`, leaving the profile stuff to just user-editable fields. Otherwise it's going to be too easy to accidentally give non-administrators access via some random codepath. I've already got changes to do that on one of my branches.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7787,3 +7814,29 @@ definitions:
...
+    properties:
+      username:
+        description: The full system-assigned ID of a user, including email domain
+        type: string
+      freeCreditsLimit:
+        description: When set, overrides the default free credits dollar limit for this user.
+        type: number
+        format: double
+      contactEmail:
+        description: When set, replaces the user's contact email.  Must validate against the user's Institutional Affiliation, if one exists.
```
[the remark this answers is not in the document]
jmthibault79: They don't need to be set together - either can validate against the previous-profile value if the other is unset here

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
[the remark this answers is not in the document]
jaycarlton: [Action type names](https://github.com/all-of-us/workbench/blob/master/api/docs/action-audit.md#action-types) should not have targets in them. You could just reuse `EDIT`. `BYPASS` to me works for any kind of limit or other blocker that the administrator is removing, and I could see it being helpful to query all such changes together.

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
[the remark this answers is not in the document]
jaycarlton: I'm a bit concerned that "user information" is not a well-scoped abstraction. What _isn't_ user information? How about `updateAccountAdminProperties()` or something?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: Can we just start calling it Affiliation? I.e. `validateEmailInAffiliatedDomain()`?

jaycarlton: Also, I try not to throw exceptions in services (especially exceptions intended for clients). Also, below, there's a 'validate' method that returns a boolean and doesn't throw:
```
!institutionService.validateAffiliation(dbVerifiedAffiliation, contactEmail)
```

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jaycarlton: Aren't they just lazily fetched the first time you try to grab them?

```
  // Authorities (special permissions) are granted using api/project.rb set-authority.
  @ElementCollection(fetch = FetchType.LAZY)
  @CollectionTable(name = "authority", joinColumns = @JoinColumn(name = "user_id"))
  @Column(name = "authority")
  public Set<Short> getAuthorities() {
    return authorities;
  }
```

This is getting at my main source of confusion: does the DAO or the Entity drive the train here?

api/src/main/resources/workbench-api.yaml
```diff
@@ -7787,3 +7814,29 @@ definitions:
...
+        type: string
+      verifiedInstitutionalAffiliation:
+        description: When set, replaces the user's verified institutional email.  Must validate against the user's contact email.
+        "$ref": '#/definitions/VerifiedInstitutionalAffiliation'
+      accessBypassRequests:
+        description: When set, updates the user's bypass status for one or more access modules.
+        type: array
+        items:
+          "$ref": '#/definitions/AccessBypassRequest'
```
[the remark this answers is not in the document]
jmthibault79: There's no nullable boolean here - this is a confusion between data at rest and the action to change the data.

AccessBypassRequest holds the intent to make a change: turn a bypass on or off (possibly redundantly).  The bypass state itself is a nullable timestamp on DbUser, with the semantics of `isBypassed = (timestamp != null)`.

So the default of "never set" is equivalent to false.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -180,8 +188,9 @@
     CaptchaVerificationService.class,
     UserServiceImpl.class,
     UserServiceTestConfiguration.class,
+    FreeTierBillingService.class,
```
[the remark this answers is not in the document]
jmthibault79: Will do.  I typically use Edit -> Sort Lines

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -231,6 +240,9 @@ public void setUp() throws IOException {
             .country(COUNTRY)
             .zipCode(ZIP_CODE));
 
+    // TODO: this needs to be set in createAccountAndDbUserWithAffiliation() instead of here.  Why?
+    // profile.setEmailVerificationStatus(EmailVerificationStatus.SUBSCRIBED);
```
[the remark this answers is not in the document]
jmthibault79: This is the command which logically fits here but does not work properly.  Is there a better way to make that clear?

I could also just remove it.

api/src/test/java/org/pmiops/workbench/api/ProfileControllerTest.java
```diff
@@ -1147,14 +1138,382 @@ public void testUpdateProfile_updateDemographicSurvey() {
...
+  public void test_editUserInformation_no_change() {
+    final Profile original = createAccountAndDbUserWithAffiliation();
+
+    // valid user but no fields updated
+    final EditUserInformationRequest request =
+        new EditUserInformationRequest().username(PRIMARY_EMAIL);
+    final Profile retrieved = profileService.editUserInformation(request);
+
+    // RW-5257 Demo Survey completion time is incorrectly updated
+    retrieved.setDemographicSurveyCompletionTime(null);
```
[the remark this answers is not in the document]
jmthibault79: I want to assert a round trip but I can't because of RW-5257: this line is necessary to correct that.  Is there a better way to show this?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -381,47 +415,73 @@ private void validateAreaOfResearch(Profile profile) throws BadRequestException
...
-    if (!getChangesWithPrefix(diff, "username").isEmpty() || isNewObject) {
-      validateUsername(updatedProfile);
+      // only validate if the new profile has an affiliation - some older users do not
+      if (profile.getVerifiedInstitutionalAffiliation() != null) {
+        validateInstitutionalAffiliationAgainstEmail(profile);
+      }
     }
-    if (!getChangesWithPrefix(diff, "contactEmail").isEmpty() || isNewObject) {
-      validateContactEmail(updatedProfile);
+    if (isNewProfile || fieldChanged(diff, "givenName")) {
```
jaycarlton, line 450: This would be cleaner if you passed in `isNewProfile` to `fieldChanged()`. Actually I'd probably have a method that absorbed all the `if`s as well so there's not so much branching. Like

`boolean validateIfChanged(Diff diff, String fieldName, Consumer<Profile> validateFunction);` 

Bonus points for putting all that into an enum, or using a map from field names to validators.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jaycarlton: In the Swagger I mean. It's actually marked `required` even though we can't change it  😢 .

```
  Profile:
    type: object
    required:
    - username
    - dataAccessLevel
```

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -432,10 +492,48 @@ public void validateProfile(@Nonnull Profile updatedProfile, @Nullable Profile p
...
+   * Updates the user metadata referenced by the fields of EditUserInformationRequest.
+   *
+   * @param request the fields to update. Fields left null here will not be updated. Contact Email
+   *     and Verified Institutional Affiliation updates will trigger a check for affiliation
+   *     validation.
+   * @return the Profile of the user, after updates
+   */
+  public Profile editUserInformation(EditUserInformationRequest request) {
+    final DbUser dbUser = userService.getByUsernameOrThrow(request.getUsername());
+    final Profile originalProfile = getProfile(dbUser);
```
jaycarlton, line 514: I'd bundle these two calls into a `ProfileService#getByUsername(String username)` if you don't need the DbUser.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jaycarlton: Can you compare with a tolerance (like $0.01) so that we're not affected by roundoff problems?

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jaycarlton: 💥  `null` should be deprecated along with `if`.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7787,3 +7814,29 @@ definitions:
...
+        description: The full system-assigned ID of a user, including email domain
+        type: string
+      freeCreditsLimit:
+        description: When set, overrides the default free credits dollar limit for this user.
+        type: number
+        format: double
+      contactEmail:
+        description: When set, replaces the user's contact email.  Must validate against the user's Institutional Affiliation, if one exists.
+        type: string
+      verifiedInstitutionalAffiliation:
```
jaycarlton, line 7836: Can we just name this property `affiliation` with the intention to rename `VerifiedInstitutionalAffiliation` soon?

api/src/main/resources/workbench-api.yaml
```diff
@@ -7787,3 +7814,29 @@ definitions:
...
+        description: When set, replaces the user's verified institutional email.  Must validate against the user's contact email.
```
jaycarlton, line 7837: I'm confused: does the affiliation own the contact email or is it independent?

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -295,4 +296,34 @@ public double getUserFreeTierDollarLimit(DbUser user) {
...
+
+  /**
+   * Set a custom Free Tier dollar limit override for this user. If this is greater than the user's
+   * total cost, set their workspaces to active. Note: lowering the limit below total cost will not
+   * set the workspaces to inactive. checkFreeTierBillingUsage() will do this as part of the next
+   * cron run.
+   *
+   * @param user the user as represented in our database
+   * @param dollarLimit the US dollar amount, represented as a double
+   * @return whether the user's total cost is now below their limit
```
jaycarlton, line 308: If I see a method starting with `set` that returns a `boolean`, I generally assume it returns true if the value was successfully set. This contract is confusing to me.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -295,4 +296,34 @@ public double getUserFreeTierDollarLimit(DbUser user) {
...
+   */
+  public boolean setFreeTierDollarOverride(DbUser user, double dollarLimit) {
+    final Double previousLimitMaybe = user.getFreeTierCreditsLimitDollarsOverride();
+
+    // TODO: prevent setting this limit directly except in this method?
+    user.setFreeTierCreditsLimitDollarsOverride(dollarLimit);
+    user = userDao.save(user);
```
jaycarlton, line 315: I am surprised to see this being set in this service as opposed to the UserAdminService or ProfileService or someplace. I guess you'd have to invert the dependency for that to happen though.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -187,11 +192,28 @@ public void validateInstitutionalAffiliation(Profile profile) {
...
+  }
+
+  /**
+   * Updates a profile for a given user as Admin and persists all information to the database.
+   *
+   * @param user
+   * @param updatedProfile
+   * @param previousProfile
+   */
+  public void adminUpdateProfileForUser(
```
jaycarlton, line 205: Why do we need the "forUser" suffix? I think that was there previously but I never got what it meant.

So Admins can't create profiles, right? At least not in their capacity as admins.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -381,47 +415,73 @@ private void validateAreaOfResearch(Profile profile) throws BadRequestException
...
+      validateInstitutionalAffiliation(profile);
     }
-    if (!getChangesWithPrefix(diff, "verifiedInstitutionalAffiliation").isEmpty() || isNewObject) {
-      validateInstitutionalAffiliation(updatedProfile);
+  }
+
+  private void validateChangesAllowedByUser(Diff diff) {
+    if (fieldChanged(diff, "username")) {
+      // See RW-1488.
+      throw new BadRequestException("Changing username is not supported");
```
jaycarlton, line 470: By exiting early, we can't validate multiple fields and give the client a single exception with a list of issues with the profile. Granted no one has asked for that yet.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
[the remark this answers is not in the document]
jmthibault79: edit works for me

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: We throw all over this Service.  It's useful for me to do so here as well.

validateAffiliation() works for me

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: Oh I see, thanks.  This seems much more tractable, then.

jmthibault79: This logic is now gated on the new DbUser.hasAuthority()

api/src/main/java/org/pmiops/workbench/api/ProfileController.java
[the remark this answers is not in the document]
jmthibault79: "User Information" comes from the UI control.  I agree that it's vague and should be changed.  Something like AdminAccountProperties, yeah.  Let's see how that looks.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jmthibault79: Looks like we don't use `readOnly` properties at all.

api/src/main/resources/workbench-api.yaml
```diff
@@ -7787,3 +7814,29 @@ definitions:
           timestamp (epoch millis) at workspace creation
         type: integer
         format: int64
+  EditUserInformationRequest:
```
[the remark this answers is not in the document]
jmthibault79: Profile would be overkill here, and much more difficult to get right.  (That was the main difficulty of a previous iteration of this work.)  This fits the semantics of what we're trying to do here.  I have updated the name to reference the Account concept, per other comments.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: I do need the DbUser

api/src/main/resources/workbench-api.yaml
jmthibault79: copypaste error.  fixed.

api/src/main/resources/workbench-api.yaml
jmthibault79: I like that.

We should remove all of the old-style affiliations before doing so, but I think we can get to this soon.  New ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-5326

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: good point.  We don't need this return value.  It's a void now.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: (NOTE: I have removed the admin version of this method in favor of checking authorities)

I'm not sure.  @gjuggler this was your originally, right?  Any concern with renaming `updateProfileForUser()` to "updateProfile" ?

gjuggler: IIRC this method originally lived at the controller level, where there is also an updateProfile() method (which handles the case when "my current user" is the target user being updated). Now that it's in the service it makes sense to remove the suffix for consistency with other methods in that file.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jmthibault79: OK.  The comment I am referring to is now the Javadoc for the method FreeTierBillingService.maybeSetDollarLimitOverride()

and it does now check vs a tolerance

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: Update: it returns a boolean again, and works in the way you describe.  Method is now called maybeSetDollarLimitOverride() and has more internal logic

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: Thanks.  Updated.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
```diff
@@ -11,5 +11,5 @@ enum class ActionType {
     DUPLICATE_TO,
     COLLABORATE,
     DELETE,
-    DETECT_HIGH_EGRESS_EVENT
+    DETECT_HIGH_EGRESS_EVENT,
```
jaycarlton, line 14: nit: I don't know if we have a decision on trailing commas in these kinds of lists. Seems like it might not work everywhere in Java.

api/src/main/java/org/pmiops/workbench/actionaudit/auditors/UserServiceAuditor.java
```diff
@@ -21,4 +22,7 @@ void fireAdministrativeBypassTime(
       Optional<Instant> newBypassTime);
 
   void fireAcknowledgeTermsOfService(DbUser targetUser, Integer termsOfServiceVersion);
+
+  void fireFreeTierDollarQuotaAction(
```
jaycarlton, line 26: nit: general convention would be something like `fireEditFreeTierQuota()`. Auditors always fire actions, so that part of the name doesn't narrow it down much.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -136,10 +138,14 @@ private int compareCostFractions(final double a, final double b) {
     return DoubleMath.fuzzyCompare(a, b, COST_FRACTION_TOLERANCE);
   }
 
-  private boolean expiredByCost(final DbUser user, final double currentCost) {
+  private boolean costAboveLimit(final DbUser user, final double currentCost) {
     return compareCosts(currentCost, getUserFreeTierDollarLimit(user)) > 0;
   }
 
+  private boolean costsDiffer(final double a, final double b) {
```
jaycarlton, line 145: It might help to have a small `MyCurrency` class to wrap the double, from which you could move to a standard representation using integral cents. Just to keep money from getting subtracted from weights or something.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -295,4 +300,42 @@ public double getUserFreeTierDollarLimit(DbUser user) {
...
+   * @return whether an override was set
+   */
+  public boolean maybeSetDollarLimitOverride(DbUser user, double newDollarLimit) {
+    final Double previousLimitMaybe = user.getFreeTierCreditsLimitDollarsOverride();
+
+    if (previousLimitMaybe != null
+        || costsDiffer(
+            newDollarLimit, workbenchConfigProvider.get().billing.defaultFreeCreditsDollarLimit)) {
+
+      // TODO: prevent setting this limit directly except in this method?
```
jaycarlton, line 325: Best way to do that is with a separate little service that only exposes one entry point.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
```diff
@@ -295,4 +300,42 @@ public double getUserFreeTierDollarLimit(DbUser user) {
...
+      user.setFreeTierCreditsLimitDollarsOverride(newDollarLimit);
```
jaycarlton, line 326: Looks like we're not setting the modified time on the user row? Please confirm that's what's wanted.

api/src/main/java/org/pmiops/workbench/db/model/DbUser.java
```diff
@@ -301,6 +301,13 @@ public void setAuthorities(Set<Short> newAuthorities) {
     this.authorities = newAuthorities;
   }
 
+  @Transient
+  public boolean hasAuthority(Authority required) {
+    // DEVELOPER is the super-authority which subsumes all others
```
jaycarlton, line 306: We should never have an enumeration where some states include others. This is icky. DEVELOPER should be its own, separate concept then.

Additionally, DEVELOPER is a bad name for this. I don't have or want authorities on PROD to do things like this.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -381,47 +395,72 @@ private void validateAreaOfResearch(Profile profile) throws BadRequestException
...
+    validateProfileForCorrectness(updatedProfile, isNewProfile, diff);
+
+    if (userProvider.get().hasAuthority(Authority.ACCESS_CONTROL_ADMIN)) {
+      validateChangesAllowedByAdmin(diff);
+    } else {
+      validateChangesAllowedByUser(diff);
+    }
+  }
+
+  private void validateProfileForCorrectness(Profile profile, boolean isNewProfile, Diff diff)
```
jaycarlton, line 416: I don't like the `isNewProfile` flag here, because the Diff should already be able to tell us that. E.g. something like `boolean isNewObject(Diff diff)`. Then each validate method could just take in a Diff and a string for the field name, or we could have a `validateFields(Diff diff, Set<String> fieldNames)`. there are lots of `if` statements here currently, which are easily breakable by maintainers.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jaycarlton: I don't believe that's by policy, it's just that we haven't yet. I have it on a branch someplace, but that didn't go in for other reasons.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
jaycarlton: But now there's an order dependence: you can't change one without changing the other if there's a new domain and a new institution, right? 

Also, I'm really surprised no one has asked for separate contact emails and institutional emails yet.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
jaycarlton: Right. We can't distinguish a state with no history from a state where it was bypassed and removed by looking at the database. If history matters, then a boolean isn't really enough.

It's in the audit trail, though, and not something that should really be tackled on this PR.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
jmthibault79: Not necessarily - a contact email could potentially be valid for more than one institution.

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: That's effectively what we're doing already, at the Service or Controller level.  I'm imagining something stricter which we probably can't do, like "only FreeTierBillingService can call this DbUser accessor"

api/src/main/java/org/pmiops/workbench/billing/FreeTierBillingService.java
jmthibault79: good catch.  fixed.

api/src/main/java/org/pmiops/workbench/db/model/DbUser.java
jmthibault79: This replicates the existing Controller auth behavior.

I would support removing it, but that's going to require a separate ticket and a larger conversation at a minimum.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: ok fine: isNewProfile is now a function of the Diff.

I'm keeping the ifs because I like how they make the intent explicit.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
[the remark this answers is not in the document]
jmthibault79: I tried this out and it didn't seem do anything.  It didn't remove the generated Profile.setUsername() or even prevent us from calling it - which we do, by the way: ProfileService.java:261

It also doesn't seem to *hurt* anything, so we could add it to mark our intention.

api/src/main/java/org/pmiops/workbench/actionaudit/ActionType.kt
jmthibault79: I've been adding these various places - especially `@Import({...})` lists in tests

api/src/main/java/org/pmiops/workbench/db/dao/UserServiceImpl.java
```diff
@@ -1022,4 +1024,13 @@ public void updateBypassTime(long userDatabaseId, AccessBypassRequest accessBypa
             "There is no access module named: " + accessBypassRequest.getModuleName().toString());
     }
   }
+
+  @Override
+  public boolean hasAuthority(long userId, Authority required) {
```
jmthibault79, line 1029: This is new, and it works because it calls the query with JOIN FETCH

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: The annotation is Controller-only, but I added a UserService.hasAuthority() here which does the job

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
jmthibault79: Good point!  I filed this: https://precisionmedicineinitiative.atlassian.net/browse/RW-5334

api/src/main/resources/workbench-api.yaml
```diff
@@ -1128,6 +1128,33 @@ paths:
...
+      description: 'Updates a subset of a user''s account metadata, as defined by the AccountPropertyUpdate object.  Requires ACCESS_CONTROL_ADMIN authority.'
+      parameters:
+        - in: body
+          name: request
+          description: The admin edit request to perform on a user
+          schema:
+            "$ref": "#/definitions/AccountPropertyUpdate"
+      operationId: updateAccountProperties
+      responses:
+        200:
```
jaycarlton, line 1146: aside: we should start using a yaml partial to bring these common codes in if we want to declare them everywhere. I'm not sure there's much percentage in doing it for all the endpoints.

api/src/main/resources/workbench-api.yaml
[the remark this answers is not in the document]
jaycarlton: No I mean my verified institutional email is brassrat@mit.edu but I want transactional email to go to prince.adam@gmail.com. As in I only use the institutional email for official purposes (kind of like how I have a broadinstitute email that I never check).

api/src/test/java/org/pmiops/workbench/billing/FreeTierBillingServiceTest.java
```diff
@@ -329,27 +335,112 @@ public void checkFreeTierBillingUsage_workspaceMissingCreatorNoNPE() {
...
+    // we update the user again and should see this new last modified time
+    final Instant time3 = START_INSTANT.plusSeconds(2000);
+    CLOCK.setInstant(time3);
+
+    assertThat(freeTierBillingService.maybeSetDollarLimitOverride(user, 100.0)).isTrue();
+    verify(mockUserServiceAuditor)
+        .fireSetFreeTierDollarLimitOverride(user.getUserId(), 200.0, 100.0);
+    assertWithinBillingTolerance(freeTierBillingService.getUserFreeTierDollarLimit(user), 100.0);
+    assertThat(userDao.findUserByUserId(user.getUserId()).getLastModifiedTime())
+        .isEqualTo(new Timestamp(time3.toEpochMilli()));
```
jaycarlton, line 363: I've been using `.isWithin(100.0).of(time_millis)` for millisecond comparisons. Have to cast the arg to `assertThat()` to `double` to make that work. I guess we could make an instant-valued version...

PR #3801, opened 2020-07-17
[RW-5264, RW-5266][risk=no] Reporting PoC payload objects and service

jaycarlton:
Initial classes and services for reporting. Highlights:
* A ReportingService class responsible for gathering the information and building a payload for BigQuery
* Value classes (generated for convenience in Swagger) for the columns we need. Currently just researcher adn workspce tables (2-3 columns each)
* a method to capture a timestamp to use as a partition key (potentially)
* endpoint for kicking off the upload, to serve as cron target
* unit test class

---
**PR checklist**

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/reporting/ReportingService.java
+++ b/api/src/main/java/org/pmiops/workbench/reporting/ReportingService.java
@@ -1,7 +1,11 @@
 package org.pmiops.workbench.reporting;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.pmiops.workbench.model.ReportingSnapshot;
 
 public interface ReportingService {
+  @VisibleForTesting
   ReportingSnapshot getSnapshot();
+
+  void uploadSnapshot();
 }
--- a/api/src/main/java/org/pmiops/workbench/reporting/ReportingService.java
+++ b/api/src/main/java/org/pmiops/workbench/reporting/ReportingService.java
@@ -1,0 +1,7 @@
+package org.pmiops.workbench.reporting;
+
+import org.pmiops.workbench.model.ReportingSnapshot;
+
+public interface ReportingService {
+  ReportingSnapshot getSnapshot();
+}
--- a/api/src/main/resources/workbench-api.yaml
+++ b/api/src/main/resources/workbench-api.yaml
@@ -7732,10 +7732,11 @@ definitions:
   ReportingResearcher:
     description: >
       Payload for the researcher table in the BigQuery reporting dataset.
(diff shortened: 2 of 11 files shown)


api/src/main/resources/workbench-api.yaml
```diff
@@ -7711,3 +7729,60 @@ definitions:
         description: Datetime at creation of this user's accont
         type: string
         format: date-time
+  ReportingResearcher:
+    description: >
+      Payload for the researcher table in the BigQuery reporting dataset.
+    type: object
+    properties:
+      username:
+        description: Username (GSuite)
```
jmthibault79, line 7738: Please clarify whether this includes the domain

api/src/main/resources/workbench-api.yaml
```diff
@@ -7711,3 +7729,60 @@ definitions:
         description: Datetime at creation of this user's accont
         type: string
         format: date-time
+  ReportingResearcher:
```
jmthibault79, line 7732: Which elements are required for each of these structures?  We can add that in a `required` section.

api/src/main/java/org/pmiops/workbench/reporting/ReportingService.java
```diff
@@ -0,0 +1,11 @@
+package org.pmiops.workbench.reporting;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.pmiops.workbench.model.ReportingSnapshot;
+
+public interface ReportingService {
+  @VisibleForTesting
+  ReportingSnapshot getSnapshot();
```
jmthibault79, line 8: these are both `public` in the impl

jaycarlton: right, It's only in the interface (thus public when overridden) for testing. But this is already changed in the next batch of code.

api/src/main/resources/workbench-api.yaml
jaycarlton: This isn't part of the API at all, though all environments will be saddled with it. I want to replace it with Immutables or maybe Kotlin data classes so we don't have to put it here.

api/src/main/resources/workbench-api.yaml
jaycarlton: yes it does, though now that I think about it, we'll be documenting the schema separately from this file so maybe I should pull these descriptions from here.

We should probably make a new yaml file for internal, backend-only codegen, but it's not worth it compared to just using Immutables.

PR #3803, opened 2020-07-17
[RW-5089][risk=no] Centralize Zendesk urlset and correct sandbox URLs

calbach:
- ZD sandbox (used in lower environments), has different article IDs than in prod
- I decided to build the URLs statically to allow for static references in code

## Notes about Zendesk URLs

- the locale infix "/en-us/" is not necessary (except for the community forum URL). Better to leave this unset.
- the article ID is the only necessary part of the URL suffix, though when you click the URL via Zendesk it will include a sluggified title. My preference is to omit this, as those titles may change over time

## Alternative considered
Just linking the URLs directly into the environment.ts, but this felt like it might be overly complicated, for example:

```
cat environment.prod.ts
{
  ...,
  zendeskUrls: zendesk.prodUrls,
  ...

cat environment.stable.ts
{
  ...,
  zendeskUrls: zendesk.sandboxUrls,
  ...
```

where zendesk would need to be a hermetic package which environment.ts depends on. Could be convinced to explore this more.

### DIFF
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -20,9 +20,9 @@ import {getRegistrationTasksMap, RegistrationDashboard} from 'app/pages/homepage
 import {profileApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors, {addOpacity} from 'app/styles/colors';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
-import {supportUrls} from 'app/utils/zendesk';
 import {AnalyticsTracker} from 'app/utils/analytics';
 import {fetchWithGlobalErrorHandler} from 'app/utils/retry';
+import {supportUrls} from 'app/utils/zendesk';
 import {
   Profile,
 } from 'generated/fetch';
--- a/ui/src/app/pages/homepage/homepage.tsx
+++ b/ui/src/app/pages/homepage/homepage.tsx
@@ -20,9 +20,9 @@ import {getRegistrationTasksMap, RegistrationDashboard} from 'app/pages/homepage
 import {profileApi, workspacesApi} from 'app/services/swagger-fetch-clients';
 import colors, {addOpacity} from 'app/styles/colors';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
+import {supportUrls} from 'app/utils/zendesk';
 import {AnalyticsTracker} from 'app/utils/analytics';
 import {fetchWithGlobalErrorHandler} from 'app/utils/retry';
-import {environment} from 'environments/environment';
 import {
   Profile,
 } from 'generated/fetch';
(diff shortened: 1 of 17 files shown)


ui/src/app/utils/zendesk.ts
```diff
@@ -1,4 +1,69 @@
...
+  tableOfContents: string,
+  researchPurpose: string,
+}
+
+/**
+ * A set of support URLs for the current environment. These need to be
+ * parameterized per environment since the Zendesk support content IDs are not
+ * stable across the prod and sandbox Zendesk instances.
+ */
+export const supportUrls: ZendeskUrls = ((baseUrl) => {
```
gjuggler, line 20: nit: What is the benefit of wrapping everything in an IIFE here? The only export is a ZendeskUrls object. Not actually a big difference, but it might be clearer to do:

```
const baseUrl = environment.zendeskHelpCenterUrl;

... most of function body ...

let supportUrls: ZendeskUrls = urls[prodHost];
if (urls[host]) {
  supportUrls = urls[host];
}
export supportUrls;
```

ui/src/environments/environment-type.ts
```diff
@@ -43,7 +43,6 @@ export interface Environment {
   // The base URL for the Zendesk help center / user forum.
   // Example value: https://aousupporthelp.zendesk.com/hc/
   zendeskHelpCenterUrl: string;
```
gjuggler, line 45: Rather than trying to parse this URL and implicitly identify the Zendesk environment we're binding to, it may be simpler (and more true to the intent) to have 'prod' and 'sandbox' as the allowable env variables here, and use that to key the `urls` map above.

ui/src/app/pages/homepage/homepage.tsx
```diff
@@ -22,6 +22,7 @@ import colors, {addOpacity} from 'app/styles/colors';
 import {hasRegisteredAccessFetch, reactStyles, ReactWrapperBase, withUserProfile} from 'app/utils';
 import {AnalyticsTracker} from 'app/utils/analytics';
 import {fetchWithGlobalErrorHandler} from 'app/utils/retry';
+import {environment} from 'environments/environment';
```
gjuggler, line 25: This change feels in the opposite direction from changes in all the other files (where we remove an `environment` dependency and add a `zendesk` dependency). See my comment on `environment.ts` below... it may be cleanest to make the env variable a simple prod vs. sandbox signal, and have all access to Zendesk URLs go through the zendesk module. Which would turn this reference into:

`href={supportUrls.helpCenterUrl}`

ui/src/environments/environment-type.ts
calbach: Good suggestion. Done. I think this way is clearer. I also moved the widget key out, since it follows from the environment type.

ui/src/app/pages/homepage/homepage.tsx
calbach: Done per other suggestion. The reason this looked different is b/c the Zendesk URL was simply hardcoded on this page originally, i.e. in worse shape than the other links - otherwise I wouldn't have needed to change this

ui/src/app/utils/zendesk.ts
calbach: Primarily stylistic preference. Generally I prefer to minimize the footprint in the package context. The state and helpers I've created (for the sole purpose of initializing these URLs) are scoped as tightly as possible and don't pollute the top level context of this file - the top-level approach leaves open the possibility of references / closures from any of the functions defined in the package. Of course, we wouldn't create any such references here, but as a first-time reader of this file I'd need to think/search to rule this out.

Visually, I also find it easier as a reader to scan this package to understand what is being exported, and what's just the internal code details. The visual hierarchy is easier to scan than picking out the `export` keyword from several top-level statements.

PR #3805, opened 2020-07-18
[RW-5295][risk=low] Consolidate field masking, fix mapper

calbach:
Bug: getWorkspaces() was not returning googleBucketName. See https://github.com/all-of-us/workbench-snippets/pull/50#discussion_r454547600 for where this mattered. Our UI currently doesn't care.

Fix1: consolidate the field masks (request more from getWorkspaces). There should be no performance delta here given the metadata we're requesting.
Fix2: have the mapper use the Firecloud workspace

I used a custom mapper as I didn't see another way, this is pretty ugly - mapstruct experts please advise.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
+++ b/api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
@@ -6,6 +6,7 @@ import org.mapstruct.CollectionMappingStrategy;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
 import org.mapstruct.MappingTarget;
+import org.mapstruct.factory.Mappers;
 import org.pmiops.workbench.cohortreview.CohortReviewMapper;
 import org.pmiops.workbench.cohorts.CohortMapper;
 import org.pmiops.workbench.conceptset.ConceptSetMapper;
@@ -61,10 +62,14 @@ public interface WorkspaceMapper {
   @Mapping(target = "researchPurpose", source = "dbWorkspace")
   Workspace toApiWorkspace(DbWorkspace dbWorkspace);
 
-  @Mapping(target = "workspace", source = "dbWorkspace")
-  @Mapping(target = "accessLevel", source = "firecloudWorkspaceResponse")
-  WorkspaceResponse toApiWorkspaceResponse(
-      DbWorkspace dbWorkspace, FirecloudWorkspaceResponse firecloudWorkspaceResponse);
+  default WorkspaceResponse toApiWorkspaceResponse(
+      DbWorkspace dbWorkspace, FirecloudWorkspaceResponse firecloudWorkspaceResponse) {
+    FirecloudMapper fcMapper = Mappers.getMapper(FirecloudMapper.class);
+    return new WorkspaceResponse()
+        .accessLevel(
+            fcMapper.fcWorkspaceResponseToApiWorkspaceAccessLevel(firecloudWorkspaceResponse))
+        .workspace(toApiWorkspace(dbWorkspace, firecloudWorkspaceResponse.getWorkspace()));
+  };
 
   @Mapping(target = "timeReviewed", ignore = true)
   @Mapping(target = "populationDetails", source = "specificPopulationsEnum")
--- a/api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
+++ b/api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
@@ -6,7 +6,6 @@ import org.mapstruct.CollectionMappingStrategy;
 import org.mapstruct.Mapper;
 import org.mapstruct.Mapping;
 import org.mapstruct.MappingTarget;
-import org.mapstruct.factory.Mappers;
 import org.pmiops.workbench.cohortreview.CohortReviewMapper;
 import org.pmiops.workbench.cohorts.CohortMapper;
 import org.pmiops.workbench.conceptset.ConceptSetMapper;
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
```diff
@@ -61,10 +62,14 @@
   @Mapping(target = "researchPurpose", source = "dbWorkspace")
   Workspace toApiWorkspace(DbWorkspace dbWorkspace);
 
-  @Mapping(target = "workspace", source = "dbWorkspace")
-  @Mapping(target = "accessLevel", source = "firecloudWorkspaceResponse")
-  WorkspaceResponse toApiWorkspaceResponse(
-      DbWorkspace dbWorkspace, FirecloudWorkspaceResponse firecloudWorkspaceResponse);
+  default WorkspaceResponse toApiWorkspaceResponse(
```
calbach, line 65: I switched to a default mapper because apparently `@Mapping` cannot support multivariate mappers: https://github.com/mapstruct/mapstruct/issues/621

api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
```diff
@@ -61,10 +62,14 @@
...
+      DbWorkspace dbWorkspace, FirecloudWorkspaceResponse firecloudWorkspaceResponse) {
+    FirecloudMapper fcMapper = Mappers.getMapper(FirecloudMapper.class);
```
calbach, line 67: Now that I'm in a default mapper, I needed to map the access level. This is the mechanism I could come up with for achieving this - I didn't see another built-in mechanism to inject the mapper, though I'm thinking perhaps it's be possible if I just split this mapper out from this class.

jaycarlton:
Let's pair on this. I spent a good amount of time in February wrestling with the [workspace mapper](https://github.com/all-of-us/workbench/pull/3465/files#diff-e5967f0cb03794422311275e49891c41), before setting it aside for a bit. The main issue is that the object has gotten so large that it's hard to keep track of little dependencies in what needs to be set where.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -242,10 +242,10 @@ private WorkspaceResponse getWorkspaceImpl(DbWorkspace dbWorkspace) {
     return workspaceResponse;
   }
 
-  private Map<String, FirecloudWorkspaceResponse> getFirecloudWorkspaces(List<String> fields) {
+  private Map<String, FirecloudWorkspaceResponse> getFirecloudWorkspaces() {
     // fields must include at least "workspace.workspaceId", otherwise
     // the map creation will fail
-    return fireCloudService.getWorkspaces(fields).stream()
```
jaycarlton, line 248: 👍 to not doing this; it's pretty opaque and likely doesn't  save  much. I really like objects in Java  to be densely populated.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
```diff
@@ -187,10 +187,10 @@ public DbWorkspace get(String ns, String firecloudName) {
 
   @Override
   public List<WorkspaceResponse> getWorkspacesAndPublicWorkspaces() {
-    Map<String, FirecloudWorkspaceResponse> fcWorkspaces =
-        getFirecloudWorkspaces(ImmutableList.of("accessLevel", "workspace.workspaceId"));
+    Map<String, FirecloudWorkspaceResponse> fcWorkspaces = getFirecloudWorkspaces();
```
jaycarlton, line 190: nit: I'd name this like  `uuidToFirecloudWorkspace`. I'd abbreviate  workspace before  firecloud because of local context.

api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
jaycarlton: Another useful trick  is the `@AfterMapping` facility, which works well,  but is somewhat less elegant IMHO. Don't think  you necessarily need it here.

api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
jaycarlton: It might be cleaner to declare `WorkspaceResponse toApiWorkspaceResponse(Workspace w, AccessLevel a)` and then build/extract the access level (or cajole the mapper to do it) when calling that. It's a more natural signature to expose, and I think we can more simply implement `WorkspaceResponse toApiWorkspaceResponse(DbWorkspace dbWorkspace, FirecloudWorkspaceResponse firecloudWorkspaceResponse)`.

jaycarlton:
Now that @ericsong is back, I'm happy to return the MapStruct guru hat.

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Existing issue, but sure

api/src/main/java/org/pmiops/workbench/workspaces/WorkspaceServiceImpl.java
calbach: Sorry to break to it to you, but it's still sparse after this change. We're just accepting the underlying service's default partial field settings, and making it standard across different entry points. We can debate the best server side representation of the partial workspace, but it is necessary to request partial data from Firecloud for performance reasons.

api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
calbach: I think this is what you were suggesting. I think I agree this is slightly cleaner - at least it allows me to remove the mapper fetch.

api/src/main/java/org/pmiops/workbench/utils/mappers/WorkspaceMapper.java
calbach: Right, but this is a default mapper. AFAICT the `@Mapping` annotation on a default mapper does nothing as default mappers don't even result in generated code.

jaycarlton: I fiddled with it for 20 minutes or so but didn't get much further. If a simple default method is working then I'd stick with that until we can overhaul the WorkspaceMapper somewhat.

calbach: Thanks for pulling it down and testing.

calbach:
I added a regression test to cover this case. Feel free to review though it's small enough that I probably won't wait on reapproval

PR #3813, opened 2020-07-21
[RW-5033][risk=no] Scripts to generate a randomish multisample VCF

als364:
Usage:
```
aschwart@failboat:~/Documents/dev/workbench/api$ ./project.rb randomize-vcf --vcf ~/Documents/dev/misc/synthetic_vcf_testing/NA12878_204126160130_R01C01.vcf.gz --number-of-copies 10 --output-dir ~/Documents/dev/misc/synthetic_vcf_testing/
aschwart@failboat:~/Documents/dev/workbench/api$  ./project.rb combine-vcfs --dir ~/Documents/dev/misc/synthetic_vcf_testing/ --output ~/Documents/dev/misc/synthetic_vcf_testing/combined.vcf
```

---
**PR checklist**


api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -0,0 +1,121 @@
...
+  public void closeTool() {
+    if ( vcfWriter != null ) {
+      vcfWriter.close();
+    }
+  }
+
+  @VisibleForTesting
+  protected static VariantContext randomizeVariant(VariantContext variant) {
+    // This initializes most of the VariantContextBuilder fields to what they were in the original variant.
+    // We just want to change the alleles, genotypes, and quality score.
```
calbach, line 77: stale comment I think - don't want to change the alleles

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -0,0 +1,121 @@
...
+    // If there's error, fuzz the error.
+    if (variant.hasLog10PError()) {
+      variantContextBuilder.log10PError(random.nextDouble() * variant.getLog10PError());
+    }
+
+    return variantContextBuilder.make();
+  }
+
+  protected static Genotype randomizeGenotype(VariantContext variantContext, Genotype genotype) {
+    GenotypeBuilder genotypeBuilder = new GenotypeBuilder();
```
calbach, line 100: nit: looks like these builder methods can be chained

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -0,0 +1,121 @@
...
+    genotypeBuilder.copy(genotype);
+    genotypeBuilder.name(appendSuffixToSampleName(genotype.getSampleName()));
+    genotypeBuilder.alleles(randomizeAlleles(variantContext, genotype.getAlleles()));
+    return genotypeBuilder.make();
+  }
+
+  @VisibleForTesting
+  protected static List<Allele> randomizeAlleles(VariantContext variantContext, List<Allele> genotypeAlleles) {
+    // The alleles list on the VariantContext has first the reference and then all possible alternates.
+    // For each genotype, we pick from among those possible alternates (or we put a no-call.)
```
calbach, line 112: opt: another way we could do this would be to take the global rate of no-call in the VCF file (just look it up manually and hardcode it here), then pick no call with the corresponding probability. This would probably be better since the current approach will result in strictly fewer no-calls w.r.t. the original VCF, and only in the same positions.

I was curious so I looked at the frequencies of each genotype in the sample VCF:
```
  27629 ./.
1518493 0/0
 106755 0/1
 117349 1/0
 138615 1/1
   1123 2/2
```

Perhaps this randomizer should also favor homozygous reference as well.

.gitignore
```diff
@@ -25,9 +26,7 @@ cron-emulator-err.log
 /api/src/generated
 /api/tools/build/*
 /api/bin
-/common-api/.swagger-codegen/*
```
als364, line 28: We don't have common-api anymore.

api/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu May 10 14:42:28 CDT 2018
+#Tue Jul 21 11:03:10 EDT 2020
```
als364, line 1: auto generated when refreshing gradle after creating the subproject

api/genomics/src/test/java/org/pmiops/workbench/genomics/RandomizeVcfTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class RandomizeVcfTest {
+  private static final VCFFileReader reader =
+      new VCFFileReader(new File("src/test/resources/NA12878_204126160130_R01C01.toy.vcf.gz"));
+  private static final VariantContext variantContext = reader.iterator().next();
+  private static final RandomizeVcf randomizeVcf = new RandomizeVcf();
+
+  @BeforeClass
+  public static void setUp() {
+    randomizeVcf.sampleNameSuffix = "0";
```
calbach, line 22: nit: I'd probably just expose two constructors here, 1 default constructor, 1 that takes the sampleNameSuffix, used by the test (and `@VisibleForTesting`)

api/genomics/src/test/java/org/pmiops/workbench/genomics/RandomizeVcfTest.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Test
+  public void TestRandomizeVariant() {
```
calbach, line 26: nit: method names should be lower camel

api/genomics/src/test/java/org/pmiops/workbench/genomics/RandomizeVcfTest.java
```diff
@@ -0,0 +1,63 @@
...
+    VariantContext randomizedVariant = randomizeVcf.randomizeVariant(variantContext);
+    // For each sample/GT in the variant...
+    randomizedVariant
+        .getGenotypes()
```
calbach, line 30: opt nit: you could probably reduce some nesting here by using stream `flatMap`. This basically lets you expand the current stream with more elements, i.e. map a small collection of genotypes into a larger collection alleles

```
.getGenotypes()
.stream()
.flatMap(Genotype::getAlleles)
.forEach(... assertion ...)
```

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -0,0 +1,165 @@
...
+  @Argument(
+      doc = "Output vcf name.",
+      fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+      shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME)
+  protected File outputVcf;
+
+  @Argument(doc = "Sample name suffix.", fullName = "SAMPLE_NAME_SUFFIX", shortName = "S")
+  protected String sampleNameSuffix;
+
+  private Random random = new Random();
```
calbach, line 40: I recommend allowing the test to set a fixed seed to make the test more deterministic. See my comment in the test file for how you could plumb that through in a test constructor

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -0,0 +1,165 @@
...
+            .collect(Collectors.toList());
+    GenotypesContext randomizedGenotypesContext =
+        GenotypesContext.create(new ArrayList<>(randomizedGenotypes));
+
+    variantContextBuilder.genotypes(randomizedGenotypesContext);
+
+    // We want kind of random error. If there's no error, have no error for the new variant as well.
+    // If there's error, fuzz the error.
+    if (variant.hasLog10PError()) {
+      variantContextBuilder.log10PError(random.nextDouble() * variant.getLog10PError());
```
calbach, line 92: This works for now, we're not too concerned with errors at the moment - I'd probably put a comment with possible improvements or deficiencies. Specifically:

- I'm not sure about applying a fractional scaling to a log10 value. The net effect on the original error value will be amplified. Though I'm not familiar enough with these error ranges to know whether the output is something that's still within reason.
- this currently should strictly reduce the value; most likely we would want to fuzz it in either direction, or just generate a new value altogether here, unrelated to the original.

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcfEntryPoint.java
```diff
@@ -0,0 +1,7 @@
+package org.pmiops.workbench.genomics;
+
+public class RandomizeVcfEntryPoint {
+  public static void main(String[] argv) {
+    new RandomizeVcf().instanceMain(argv);
```
calbach, line 5: opt: personal preference, but I think this could just go into the RandomizeVcf file directly, rather than having the separate entrypoint file. I don't feel strongly.

.circleci/config.yml
```diff
@@ -321,15 +333,18 @@ jobs:
           #   src/test/java/org/pmiops/*********/monitoring/LogsBasedMetricsServiceTest.java src/test/java/org/pmiops/*********/monitoring/MonitoringServiceTest.java
           # Example of Gradle command after formatting:
           #   gradle test --tests org.pmiops.*********.monitoring.LogsBasedMetricsServiceTest --tests org.pmiops.*********.monitoring.MonitoringServiceTest
+          #
+          # We exclude the :genomics:test task below because otherwise gradle will attempt to run all the regular API unit tests under the genomics
+          # submodule and will be unable to find them and will barf
```
calbach, line 338: I don't understand - why would the below glob pickup the genomics test files? Or is that not the issue?

als364: That is not the issue. 

`gradle test`, by default, runs all subtasks of `gradle test`. This, generally, makes sense - it's common for repos to have, like, a `unitTest` and `integrationTest` subtask and you'd want `gradle test` to run them all.  In this case, `gradle test` runs `:test`, `:appengine-cron-emulator:test`, and `:genomics:test`, since the latter two are submodules of the top-level gradle project.

The issue is that running this command like this passes `$GRADLE_ARGS` to every one of those subtasks with no regard for whether those tests exist in those projects. You end up getting `gradle test :gradle:test --tests org.pmiops.workbench.DoesNotExistInThisClasspathTest`. 

What I'm doing here isn't excluding the genomics test modules from being handed to `gradle test`, I'm excluding the genomics test subtask from executing. This command won't grab the genomics tests anyways because of how `circleci tests glob` works - it's not a regex on paths, it just starts looking from `working_directory`, and since the genomics tests are not under `api/src/test/java` they won't get picked up.

Hope that all makes sense. I've created another CircleCI step above to run just the genomics tests.

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
als364: Heh, I'm also not too sure about the usual error ranges. Might just remove this and put a follow-up ticket to investigate error fuzzing.

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcfEntryPoint.java
als364: I'll give a try; if it works I'll get rid of these. I _do_ need the combine entry point cause I'm not extending anything from Picard

api/genomics/src/test/java/org/pmiops/workbench/genomics/RandomizeVcfTest.java
als364: oh that would make more sense wouldn't it

api/genomics/src/test/java/org/pmiops/workbench/genomics/RandomizeVcfTest.java
als364: Oh yeah, that's true, thank you

api/libproject/devstart.rb
```diff
@@ -2580,3 +2580,71 @@ def start_api_and_incremental_build(cmd_name, args)
...
+    basename = File.basename(op.opts.vcf, ".vcf.gz")
+    out = "#{op.opts.out}#{basename}.#{i}.vcf"
+    app_args = "-PappArgs=['-V#{op.opts.vcf}','-O#{out}','-S#{i}']"
+    Common.new.run_inline %W{./gradlew -p genomics randomizeVcf} + [app_args]
+  end
+end
+
+Common.register_command({
+  :invocation => "randomize-vcf",
+  :description => "Given an example vcf and a number of copies to make, generates that many " +
```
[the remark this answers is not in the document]
als364: Yes, `VariantContextWriter`s automatically create an index on file close. I'll update this comment. IMO a README is still worth it; I'll drop it in `api/genomics`

api/libproject/devstart.rb
```diff
@@ -2580,3 +2580,71 @@ def start_api_and_incremental_build(cmd_name, args)
...
+          "folder."
+  )
+  op.add_option(
+      "--output [output]",
+      -> (opts, o) {opts.output = o},
+      "Path to output the combined vcf"
+  )
+  op.parse.validate
+
+  app_args = "-PappArgs=["
```
calbach, line 2635: opt: We have several other commands that need to prep args like this, I think the pattern could be reused here - or it may not be relevant if the flag construction for GATK is specialized:
https://github.com/all-of-us/workbench/blob/0ccccd01288d7faeaf3787e677ecde7019274482/api/libproject/devstart.rb#L1341

probably there's enough instances to pull it into a helper eventually.

.circleci/config.yml
calbach: I see, thanks. My preferred fix would be `gradle test` --> `gradle :test` on this line (no -x), this only picks up the root test task which would have been more accurate to begin with but didn't matter until now. This way you're not needing to list out every subproject that has tests. I tested locally that this works.

I would also clarify the above comment that it pertains specifically to the test filtering. If you filter down to 0 tests for a particular gradle test target, that's an error. This means it's impractical to use the test filter in combination with a multi-test gradle task.

api/libproject/devstart.rb
als364: the flag construction for GATK is different than we do it and the flag construction for picard is differently different...

PR #3819, opened 2020-07-22
[RW-5305][risk=no] links for audit

jaycarlton:
I added links to and from the audit and admin pages for users and workspaces. For users, we just have the main table, but there's a table entry now to link to that user's audit.

Also added two contextual links on the audit cards themselves. For user agents, there's a link from the username to the user's audit page, and for workspaces, the namespace property links to that workspace. There are several more places this trick can be done. I'm looking for a more generic way to do what I want though, as I wound up duplicating the HideableCell ll code largely.

The user admin page doesn't load on test, so I was only able to see it locally.

---
**PR checklist**

Screen Shot 2020-07-30 at 5 26 43 PM
Screen Shot 2020-07-30 at 5 26 20 PM
Screen Shot 2020-07-30 at 5 25 41 PM


ui/src/app/pages/admin/user-audit.tsx
```diff
@@ -38,11 +38,17 @@ const getNextAuditPath = (subject: string) => {
   return `/admin/user-audit/${subject}`;
 };
 
+// Single-user admin page isn't available yet, so go to the main users list page.
+const getAdminPageUrl = (subject: string) => {
```
petesantos, line 42: TOL: These "feel" far away from where they are getting used. What are your thoughts on 
1. Since we have the initialSubject and know the query at this time, running the initial queries in these components and passing the data into the AuditPageComponent
2. Changing getAdminPageURL to something like `onAdminClick` or similar - this may be a bit clearer on where and how this property/function is being used

ui/src/app/pages/admin/admin-users.tsx
```diff
@@ -133,17 +133,18 @@ export const AdminUsers = withUserProfile()(class extends React.Component<
   }
 
   convertProfilesToFields(profiles: Profile[]) {
-
+    const usernameWithoutDomain = p => p.username.substring(0, p.username.indexOf('@'));
```
petesantos, line 136: Could the `usernameWithoutDomain` utility function be used here?

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -21,26 +22,61 @@ const HideableCell = (props: {content: string}) => {
...
-const PropertyChangeListEntry = (props: {targetProperty?: string, previousValue?: string, newValue?: string}) => {
-  const {targetProperty, previousValue, newValue} = props;
-  return <React.Fragment>
+const isWorkspaceNamespace = (targetType?: string, targetProperty?: string) => {
+  return targetType === 'WORKSPACE' && targetProperty === 'namespace';
+};
+
+const PossibleLinkCell = (props: {targetType?: string, targetProperty?: string, value?: string}) => {
+  const {targetType, targetProperty, value} = props;
+  if (isWorkspaceNamespace(targetType, targetProperty)) {
```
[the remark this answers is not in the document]
jaycarlton: My style is generally lots of tiny named functions. I don't really like the ternary operator in any language, or the if statement for  that matter. It usually means I've missed an abstraction somewhere. This bit  is pretty hacky though.

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -51,17 +87,28 @@ const PropertyChangeListView = (props: { propertyChanges: AuditTargetPropertyCha
     <div style={styles.propertyCell}>Previous Value</div>
     <div style={styles.propertyCell}>New Value</div>
     {propertyChanges.map((propertyChange, index) =>
-        <PropertyChangeListEntry {...propertyChange} key={index}/>)}
+        <PropertyChangeListEntry targetProperty={propertyChange.targetProperty}
```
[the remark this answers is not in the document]
jaycarlton: ooh  nice

ui/src/app/pages/admin/user-audit.tsx
jaycarlton: for #1, You'd  need a callback  to fetch  new entries on the next subject. How about if we kick that can one PR down the road? I have another batch of changes  right after this. My thought process  was that the user & workspace pages wouldn't need to make any API calls themselves, and would each only contribute what's  unique about their own query. So if we want to  re-fetch  with another subject or another time range, you'd have to call back to the outer component one or two levels up.

Are you saying onAdminClick could have the navigation code directly? I think I like that, assuming I can route from point A to point B from code captured at point C. It's "leaking" knowledge that we're using  the same router in all three places, but that's OK.

ui/src/app/pages/admin/admin-users.tsx
jaycarlton: good idea

ui/src/app/pages/admin/user-audit.tsx
petesantos: I am ok kicking  #1 down a bit.

onAdminClick would handle the navigation - it could still accept a parameter. I think it would be clearer at the top level that this is a function being used in a button.

ui/src/app/components/admin/audit-card-list-view.tsx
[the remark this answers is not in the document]
petesantos: I like tiny functions as well. In JS my personal preference is to in-line them unless I am using them in more than one place.

Agreed on the conditionals as well. The reason I prefer ternary is because it is a "this or that" operator. With `if` statements it makes it easier for a future developer to add additional conditions and nested conditionals rather than finding that better abstraction. The change from a ternary to a more complex conditional may stick out a bit more and raise a flag for someone - but that may just be me.

PR #3824, opened 2020-07-23
[RW-3389][risk=no] Avoid increasing query count for each user when fetching Profiles

jaycarlton:
We had an approach where we got all the users and called getProfile() on each, but getProfile() isn't designed to be called sequentially, as it makes some 4-ish more queries.

Stopgap solution is to query each table I need once and manually join them in code. I Improved the profile mapper and a couple of other mappers and things. Added a dao method or two to make this easier.

The long-term solution might be [entity inheritance](https://precisionmedicineinitiative.atlassian.net/browse/RW-5340).

---
**PR checklist**


jaycarlton:
CirrcleCI issue is  known, and fix is expected from vendor shortly.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -501,10 +506,44 @@ public void validateNewProfile(Profile profile) throws BadRequestException {
     validateProfileForCorrectness(dummyProfile, profile);
   }
 
+  // Get all the profiles. Best we can do without surgery on the entity classes is to do one query
+  // per table and join them in code.
   public List<Profile> listAllProfiles() {
-    return userService.getAllUsers().stream().map(this::getProfile).collect(Collectors.toList());
+    final Set<DbUser> usersHeavy = userService.findAllUsersWithAuthoritiesAndPageVisits();
```
gjuggler, line 512: nit: I'm not sure the `heavy` suffix helps clarify anything here. Is this important to a reader, and/or does it map to any well-defined concept or pattern in the codebase?

jaycarlton: We don't have an established convention, but I thought it important to indicate that this set of users had more fields populated than the default DbUser. The convention at HubSpot would be `hydratedUsers`.

api/src/main/java/org/pmiops/workbench/profile/ProfileService.java
```diff
@@ -393,15 +397,16 @@ private boolean fieldChanged(Diff diff, String field) {
...
-    final Predicate<Change> newProfilePred =
-        change ->
-            change instanceof NewObject
-                && change
-                    .getAffectedGlobalId()
-                    .getTypeName()
-                    .equals(Profile.class.getCanonicalName());
 
-    return diff.getChanges(newProfilePred).size() > 0;
+    return diff.getChanges(
```
jmthibault79, line 401: Why change this?  The linter-required formatting makes this version harder to follow

jaycarlton: I thought lambdas were preferred in this kind of usage and thought you might not have realized you can use a lambda instead of declaring a predicate. Arguably we could improve it by  declaring the list of changes and then taking its size. I believe  IntelliJ suggested this change as well.

jmthibault79: Makes sense.  Lambda + separated size check sounds like the best option here.  Thanks.

It's certainly not a priority to make that change.

PR #3827, opened 2020-07-24
DRAFT: use a client-supplied token instead of the login token

jmthibault79:
Description:

Early draft for comments.

Inspired by https://github.com/DataBiosphere/terra-ui/pull/686 as a step toward Capcha-avoidance in automated testing for https://precisionmedicineinitiative.atlassian.net/browse/RW-5108 - seems like a much easier change for our codebase.

Basic local test: observe that setting `window.clientSuppliedToken` to nonsense in the JS Console will cause network calls to fail and the app to not work properly.  Setting it to a legitimate Access Token for my user caused everything to work again.  Unsetting it also caused the app to work again, reverting to the standard auth flow. 

---
**PR checklist**

### DIFF
--- a/e2e/tests/workspace/workspace-edit.spec.ts
+++ b/e2e/tests/workspace/workspace-edit.spec.ts
@@ -2,14 +2,14 @@ import DataPage, {TabLabelAlias} from 'app/page/data-page';
 import WorkspacesPage from 'app/page/workspaces-page';
 import {EllipsisMenuAction} from 'app/text-labels';
 import * as testData from 'resources/data/workspace-data';
-import {findWorkspace, performActions, signIn} from 'utils/test-utils';
+import {findWorkspace, performActions, experimentalTestSignIn} from 'utils/test-utils';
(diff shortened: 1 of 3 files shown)


jmthibault79:
Question for the more knowledgeable: is the use of `global` and `window` appropriate here, or is there a better way to expose this kind of functionality to JS clients?

jmthibault79:
This doesn't currently persist across reloads, so localstorage sounds worth pursing.  Thanks.

Yes the real version will go behind an env-flag

jmthibault79:
It is necessary to use the Application Default Credentials flow to make this work.  `gcloud auth login` is not sufficient:
```
gcloud auth application-default login --no-launch-browser
gcloud auth application-default print-access-token
```

Access the JS Console in the UI and apply the token:
```
window.useToken('yourtoken')
```

Note that this token does not persist across reload.  To check whether the token is still be using (and not the real auth flow):
```
window.clientSuppliedToken
```

PR #3832, opened 2020-07-25
[no ticket][risk=no] Puppeteer: parameterize getPropValue function

aweng98:
Code improvement:
- Parameterize `getPropValue<T>()` function in `element-utils.ts` so it can be used widely to reduce code duplication.
- Get rid of `Promise<unknown>` as functions' return type.

Unrelated code mistake fix:
- Removed `await` from non-async calls like `page.url()`.

No change to tests and functions' behavior.

### DIFF
--- a/e2e/jest.test-setup.ts
+++ b/e2e/jest.test-setup.ts
@@ -15,8 +15,8 @@ const timeOut = parseInt(process.env.PUPPETEER_TIMEOUT, 10) || 90000;
 beforeEach(async () => {
   await page.setUserAgent(userAgent);
   // See https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetdefaultnavigationtimeouttimeout
-  await page.setDefaultNavigationTimeout(navTimeOut);
-  await page.setDefaultTimeout(timeOut);
+  page.setDefaultNavigationTimeout(navTimeOut);
+  page.setDefaultTimeout(timeOut);
 });
 
 /**
--- a/e2e/tests/cohorts/cohorts-create.spec.ts
+++ b/e2e/tests/cohorts/cohorts-create.spec.ts
@@ -121,7 +121,7 @@ describe('User can create new Cohorts', () => {
     const dataPage = new DataPage(page);
 
     // Save url
-    const workspaceDataUrl = await page.url();
+    const workspaceDataUrl = page.url();
 
     // Click Add Cohorts button
     const addCohortsButton = await dataPage.getAddCohortsButton();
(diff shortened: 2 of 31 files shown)


e2e/jest.test-setup.ts
```diff
@@ -15,8 +15,8 @@ const timeOut = parseInt(process.env.PUPPETEER_TIMEOUT, 10) || 90000;
 beforeEach(async () => {
   await page.setUserAgent(userAgent);
   // See https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagesetdefaultnavigationtimeouttimeout
-  await page.setDefaultNavigationTimeout(navTimeOut);
-  await page.setDefaultTimeout(timeOut);
```
aweng98, line 19: non-async functions, `await` not needed.

e2e/tests/cohorts/cohorts-create.spec.ts
```diff
@@ -121,7 +121,7 @@ describe('User can create new Cohorts', () => {
     const dataPage = new DataPage(page);
 
     // Save url
-    const workspaceDataUrl = await page.url();
+    const workspaceDataUrl = page.url();
```
aweng98, line 124: `page.url()` is also non-async call. removed `await` here and several other places.

e2e/utils/element-utils.ts
```diff
@@ -7,24 +7,24 @@ import {ElementHandle, JSHandle, Page} from 'puppeteer';
  * @param {string} attribute Attribute name.
  */
 export const getAttrValue = async (page: Page, element: ElementHandle, attribute: string): Promise<string> => {
-  return page.evaluate((link, attr) => link.getAttribute(attr), element, attribute);
+  return page.evaluate((link, attr) => {return link.getAttribute(attr)}, element, attribute);
```
gjuggler, line 10: What does `link` argument name mean in this context? Maybe it should be renamed to `el` for consistency with `attr`.

Am I right in assessing that this function is basically designed to have Puppeteer evaluate the ElementHandle into an actual Element which is then used to return the attribute value?

e2e/utils/element-utils.ts
```diff
@@ -7,24 +7,24 @@ import {ElementHandle, JSHandle, Page} from 'puppeteer';
...
+  return value as T;
 }
 
 /*
  * Wait until the existing element changed in DOM.
  * @param {Page} page Instance of Puppeteer page object.
  * @param {ElementHandle} element Element.
  */
-export async function waitUntilChanged(page: Page, element: ElementHandle): Promise<JSHandle> {
+export const waitUntilChanged = async (page: Page, element: ElementHandle): Promise<JSHandle> => {
```
gjuggler, line 28: What is the purpose of this change? This seems like the opposite syntax compared to the method above.

e2e/utils/element-utils.ts
aweng98: 👍 renamed `link` to `elemt` for consistency.
and yes to the second question, the `evaluate` function does just that.

e2e/utils/element-utils.ts
aweng98: forgot to change it back. I was experimenting.

PR #3833, opened 2020-07-27
[no ticket][risk=no] Puppeteer: modal clickButton function with optional waitFor parameters

aweng98:
Sometimes, click a button in modal should close the modal or/and causes page to change/navigate away. User has to remember to call `waitUntilClose` or/and `this.page.navigation` functions explicitly in order to keep test steps and UI in sync. The refactored`clickButton` function will make it easy to read and write lesser code when click a button in modal.
 
New function signature is `clickButton(buttonLabel: LinkText, waitOptions: {waitForNav?: boolean, waitForClose?: boolean} = {})`.

All tests verified to pass.
Screen Shot 2020-07-27 at 2 58 00 PM

### DIFF
--- a/e2e/app/component/modal.ts
+++ b/e2e/app/component/modal.ts
@@ -48,26 +48,17 @@ export default class Modal extends Container {
     const { waitForNav = false, waitForClose = false } = waitOptions;
     const button = await this.waitForButton(buttonLabel);
     await button.waitUntilEnabled();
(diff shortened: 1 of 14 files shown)


aweng98:
`puppeteer-e2e-local-ui` job fails due to unrelated changes. I'm going ahead with merge.

PR #3835, opened 2020-07-28
[RW-5343][risk=no] filter on action types

jaycarlton:
Simple filter on action types for audit cards. If any matching action type is unchecked, the card is not shown. The approach can be readily extended to other attributes of combinations of attributes on actions.

Formatting isn't the best, but it works. I also moved away from UPPER_SNAKE_CASE to Title Case labels, which I think is much easier on the eyes.

Filter labels include a count of how many cards match. Since a card can have many action types, the sum of the numbers in the filter may be more than the number of cards available (i.e. some match multiple filter categories, but all actions on a card must be checked in order for it to be shown.

TODO:
[] some better formatting and alignment
[] visual effect of some kind to show a card being removed or added
[] generalized component to filter
Screen Shot 2020-08-06 at 2 04 57 PM
Screen Shot 2020-08-06 at 2 11 38 PM

 a collection of objects based on a function to evaluate each one

---
**PR checklist**


ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,115 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+        <div style={{fontWeight: 600, color: colors.accent}}>Action Types</div>
+        {fp.entries(activeActionTypes).map((actionTypeValuePair: [string, FilterEntry], index) =>
+          <CheckBox
+            key={index}
+            id={`${actionTypeValuePair[0]}_active`}
+            onChange={() => toggleSelectedAction(actionTypeValuePair[0])}
+            label={actionTypeValuePair[1].displayName}
+            checked={actionTypeValuePair[1].isActive}
+            style={{margin: '0.25rem'}}/>)}
+        </div>
```
petesantos, line 248: This might be a bit cleaner with some additional desctructuring.
Also, keeping the code consistent with lodash may help future development and prevent confusion when reading through the code.

I am not a fan of the double `toPairs` in my solution, but it does keep it in lodash/fp... I will think through it a bit more to see if there is a better solution. 

If you don't want to do that - do take a look at the destructuring bit. It may be a bit cleaner than the array index references. The destructure should work with the native map code as well.

```javascript
fp.flow(
          fp.toPairs,
          fp.toPairs,
          fp.map(([index, [id, entry]]: [number, [string, FilterEntry]]) =>
            <CheckBox
              key={index}
              id={`${id}_active`}
              onChange={() => toggleSelectedAction(id)}
              label={entry.displayName}
              checked={entry.isActive}
              style={{margin: '0.25rem'}}/>)
        )(activeActionTypes)}
```

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,115 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
   return (
-      <div style={{margin: '1rem', width: '30rem'}}>
-        {actionsSorted.map((action, index) => (<AuditActionCard key={index} action={action}/>))}
-      </div>
+      <React.Fragment>
+        <ActionTypeFilter activeActionTypes={activeActionTypes}
+                          updateFilter={updateFilterCallback}
+        />
+        <div style={{margin: '1rem', width: '30rem'}}>
+          {fp.any(shouldShowAction)(actions)
```
petesantos, line 315: I know we differ in opinion on who should be responsible for conditional renders (parent or component).
After searching the code base I only saw 2-3 instances of the `show` property, but many of the `&& <` string (I believe that accurately indicates a conditional render)

Here is a suggestion that may simplify things a bit / improve readability:

```javascript
  const filteredActions = fp.flow(
    fp.toPairs,
    fp.map(([index, action]) => shouldShowAction(action) && <AuditActionCard key={index} action={action}/>),
    fp.without([false])
  )(actions);

  return (
      <React.Fragment>
        <ActionTypeFilter activeActionTypes={activeActionTypes}
                          updateFilter={updateFilterCallback}
        />
        <div style={{margin: '1rem', width: '30rem'}}>
          {filteredActions.length ? filteredActions : <div  style={infoMessageStyle}>All cards filtered out.</div>}
        </div>
      </React.Fragment>
  );
```

Followed by a removal of the show prop in the `AuditActionCard` component

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,115 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+        isActive: true,
+        displayName: `${toTitleCase(propertyPath)} (${actionTypeToCardCount[propertyPath] || 0})`
+      };
+    })(actionTypes);
+    setActiveActionTypes(actionTypeToShow);
+  }, [actionTypes]);
+
+  useEffect(() => {
+    const result: { [key: string]: number } = {};
+    fp.forEach(action => {
```
petesantos, line 281: Could this be done with a map - preventing the need to mutate the object?

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,115 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+      fp.flatMap((action: AuditAction) => action.eventBundles),
+      fp.map((eventBundle) => eventBundle.header.actionType),
+      fp.sortBy(a => a),
+      fp.sortedUniq)(actions));
+    setActionTypes(result);
+  }, [actions]);
+
+  useEffect(() => {
+    const actionTypeToShow: Object = {};
+    fp.forEach((propertyPath: string) => {
```
petesantos, line 269: Could this be a map rather than a `forEach` with object mutation?

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,115 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+    </React.Fragment>);
+};
+
+export const AuditActionCardListView = (props: { actions:  AuditAction[]}) => {
   const {actions} = props;
+  const [actionTypes, setActionTypes] = useState([]);
+  const [activeActionTypes, setActiveActionTypes] = useState({});
+  const [actionTypeToCardCount, setActionTypeToCardCount] = useState({});
+
+  useEffect(() => {
```
petesantos, line 258: Given that this effect runs on the prop change, and there is only one prop, it may be possible to eliminate all of the useEffect code and just compute all of the needed values. This would eliminate all 'state' in the react sense.

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,127 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+};
+
+export const AuditActionCardListView = (props: { actions:  AuditAction[]}) => {
+  console.log('render AuditActionCardListView');
   const {actions} = props;
+  const initialFilterState: FilterState = {};
+  const [activeActionTypes, setActiveActionTypes] = useState(initialFilterState);
+
+  // FIXME: handle situation when one card has the same action multiple times
+  const actionTypeToCardCount: {[key: string]: number} = fp.flow(
```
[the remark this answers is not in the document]
petesantos: I am not seeing this issue - I am seeing results.  Is this happening on all queries?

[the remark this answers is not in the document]
jaycarlton: No, the issue is when a card has multiple eventBundles with the same action type. Frequently happens with Collaborate when you share a workspace with multiple people. We end up counting that card several times for the repeated action type.

I think we just need to map the cards to tuples of unique action types, and then use that to build the map. I know how to do that part, but I'm not sure how to transform `[[A, B, C], [A, A, A], [D]]` into `{A: 2, B: 1, C: 1, D: 1}`.

ui/src/app/components/admin/audit-card-list-view.tsx
petesantos: It may be better to change what the structure of `activeActionTypes` looks like in `AuditActionCardListView` to make it easier to consume. That may be better for a future PR

ui/src/app/components/admin/audit-card-list-view.tsx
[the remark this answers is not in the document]
petesantos: I think I resolved this with:
```javascript
  const getActionTypes = (action: AuditAction) => fp.flow(
    fp.map((e: AuditEventBundle) => e.header.actionType),
    fp.uniq
  )(action.eventBundles);
```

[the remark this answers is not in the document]
jaycarlton: Oh I see. Once we have the unique count per card, we can just flatten that and count each one. I was thinking we had to   keep track of the original tuples somehow. Thanks.

ui/src/app/components/admin/audit-card-list-view.tsx
```diff
@@ -193,17 +210,109 @@ const AuditActionCard = (props: { action: AuditAction }) => {
...
+        displayName: `${toTitleCase(actionName)} (${count || 0})`
+      }]),
+      fp.fromPairs
+    )(actions);
+
+    setActiveActionTypes(actionTypeToShow);
+  }, [actions]);
+
+  const updateFilterCallback = (actionType: string, isActive: boolean) => {
+    const newActiveActionTypes = fp.clone(activeActionTypes);
```
jaycarlton, line 287: Is there a slicker way to do this? In Java, there's a builder pattern that will let you do `Builder.from(originalObj).withId(2).build` with no  mutations.

petesantos: Missed this comment when I approved - let me take a look

jaycarlton: It's just a curiosity, so no worries.

petesantos: I saw you merged, but if you are looking for an alternative this should work:
```javascript
 const updateFilterCallback = (actionType: string, isActive: boolean) => {
    setActiveActionTypes(fp.set([actionType, 'isActive'], isActive, activeActionTypes));
  };
```

PR #3838, opened 2020-07-30
[RW-5229][risk=no] Restrict special characters that break concept search

dolbeew:
Prevent calling the api when an invalid MySQL string been entered since they will break the api call.

Also restricting  `[ ] |`  since they will cause errors in the matching regex once the results are returned.
( `]` actually won't cause an error but seems weird to restrict `[` and not `]` )

Shows a warning message if the user tries to search with an invalid string.

Unclosed quote:
Screen Shot 2020-08-05 at 11 10 14 AM

Unclosed paren:
Screen Shot 2020-08-05 at 11 10 26 AM

Too many closing parens:
Screen Shot 2020-08-05 at 11 10 43 AM

Trailing + or -:
Screen Shot 2020-08-05 at 11 11 09 AM

Forbidden special character:
Screen Shot 2020-08-05 at 11 11 32 AM

The messages will also stack if there are multiple errors in the search string:
Screen Shot 2020-08-05 at 11 29 24 AM

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/data/concept/concept-homepage.tsx
+++ b/ui/src/app/pages/data/concept/concept-homepage.tsx
@@ -191,8 +191,8 @@ interface State {
   domainErrors: Domain[];
   // True if the getDomainInfo call fails
   domainInfoError: boolean;
-  // True if special characters that break search have been entered
-  forbiddenCharactersEntered: boolean;
+  // List of error messages to display if the search input is invalid
+  inputErrors: Array<string>;
   // If concept metadata is still being gathered for any domain
   loadingDomains: boolean;
   // If we are still searching concepts and should show a spinner on the table
@@ -240,7 +240,7 @@ export const ConceptHomepage = withCurrentWorkspace()(
         currentSearchString: '',
         domainErrors: [],
         domainInfoError: false,
-        forbiddenCharactersEntered: false,
+        inputErrors: [],
         loadingDomains: true,
         countsLoading: false,
         searching: false,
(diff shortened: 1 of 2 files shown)


dolbeew:
I have added a validate function that should catch _most_ invalid MySQL strings, at least the ones outlined above by Brian.

I wasn't able to find anything off the shelf that would work for our purposes or find any issues with unicode characters, although I'm sure there are edge-cases out there that will still fail.

Keep in mind also that this is a temporary solution. In the future, we will be aligning how search works for Cohort Builder and Concept Search, so this is likely to change again.

ui/src/app/pages/data/concept/concept-homepage.tsx
```diff
@@ -310,14 +318,69 @@ export const ConceptHomepage = withCurrentWorkspace()(
...
+        inputErrors.add('There is an unclosed ( in the search string');
+      }
+      if (unclosedQuotes) {
+        // unclosed quote
+        inputErrors.add('There is an unclosed " in the search string');
+      }
+      this.setState({inputErrors: Array.from(inputErrors)});
+      return inputErrors.size === 0; // return true if no errors
+    }
+
```
freemabd, line 373: I guess since you have the ability to embed `"` and `( )` you have to parse each character instead of using regex. I know this is a product decision on unifying this search with CB and DSB, but we should make this reusable by other components.

dolbeew: AFAIK there's no way to detect an imbalance of quotes or parens using regex. The [SO solution](https://stackoverflow.com/questions/26507087/escape-string-for-use-in-mysql-fulltext-search) we discussed seems to just strip out anything except letters, numbers and underscores. 

Yeah, whenever we unify the tools, will definitely make this (or whatever validation we end up using) reusable.

jscherdin: I think we just need to write up what all apps currently do and then present it to the PMs and see if we can get consensus

PR #3839, opened 2020-08-01
[RW-5329][RW-5333][risk=low] Bump aou Leo image to 1.0.4

calbach:
- Switches us back to main bigrquery package
- Installs new workbench-snippets library
- Upgrades R to 4.0
- Additional package updates in Leo base images, see changelogs under https://github.com/DataBiosphere/terra-docker for more details

Plan is to send out user communications along with this upgrade, but we're not expecting severe user impact from these upgrades. Per ongoing designs, we will plan to have a way of running older images for reproducibility in the future.

Manually tested: ran with this image locally, verified main dataset builder flows


calbach:
Puppeteer failures are unrelated existing issues that Alex is investigating. Merging.

PR #3840, opened 2020-08-03
[risk=no][RW-5342] Add useNavigation hook

petesantos:
Description:

---
**PR checklist**

### DIFF
--- a/ui/src/app/components/admin/audit-page-component.tsx
+++ b/ui/src/app/components/admin/audit-page-component.tsx
@@ -51,7 +51,7 @@ export interface AuditPageProps {
 const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, buttonLabel, queryText}) => {
   const [auditSubject, setAuditSubject] = useState(initialAuditSubject);
   const [downloadSqlFile, setDownloadSqlFile] = useState(false);
-  const [navSubject, setNavSubject] = useNavigation();
+  const [nextSubjectRequested, setNextSubjectRequested] = useNavigation();
 
   useEffect(() => {
     if (downloadSqlFile && !fp.isEmpty(queryText)) {
@@ -64,7 +64,7 @@ const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, but
 
   const onAuditClick = () => {
     setAuditSubject(auditSubject.toLowerCase().trim());
-    setNavSubject(true);
+    setNextSubjectRequested(true);
   };
 
   const getBigQueryConsoleUrl = () => {
@@ -85,7 +85,7 @@ const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, but
   };
 
   return <React.Fragment>
-    {navSubject && <Navigate to={getNextAuditPath(auditSubject)}/>}
+    {nextSubjectRequested && <Navigate to={getNextAuditPath(auditSubject)}/>}
     <TextInputWithLabel
       containerStyle={{display: 'inline-block'}}
       style={{width: '15rem', margin: '1rem'}}
--- a/ui/src/app/components/admin/audit-page-component.tsx
+++ b/ui/src/app/components/admin/audit-page-component.tsx
@@ -1,5 +1,5 @@
 import {AuditActionCardListView} from 'app/components/admin/audit-card-list-view';
-import {Navigate} from 'app/components/app-router';
+import {Navigate, useNavigation} from 'app/components/app-router';
 import {Button} from 'app/components/buttons';
 import {NumberInput, TextInputWithLabel} from 'app/components/inputs';
 import {TooltipTrigger} from 'app/components/popups';
@@ -50,14 +50,8 @@ export interface AuditPageProps {
 
 const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, buttonLabel, queryText}) => {
   const [auditSubject, setAuditSubject] = useState(initialAuditSubject);
-  const [loadNextSubject, setLoadNextSubject] = useState(false);
   const [downloadSqlFile, setDownloadSqlFile] = useState(false);
-
-  useEffect(() =>  {
-    if (loadNextSubject) {
-      setLoadNextSubject(false);
-    }
-  }, [loadNextSubject]);
+  const [navSubject, setNavSubject] = useNavigation();
 
   useEffect(() => {
     if (downloadSqlFile && !fp.isEmpty(queryText)) {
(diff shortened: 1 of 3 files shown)


ui/src/app/components/admin/audit-page-component.tsx
```diff
@@ -50,14 +50,8 @@ export interface AuditPageProps {
...
   const [auditSubject, setAuditSubject] = useState(initialAuditSubject);
-  const [loadNextSubject, setLoadNextSubject] = useState(false);
   const [downloadSqlFile, setDownloadSqlFile] = useState(false);
-
-  useEffect(() =>  {
-    if (loadNextSubject) {
-      setLoadNextSubject(false);
-    }
-  }, [loadNextSubject]);
+  const [navSubject, setNavSubject] = useNavigation();
```
jaycarlton, line 54: I didn't read this name as a boolean and  was confused. Maybe something like `doNavigate` or `canNavigate`?

ui/src/app/components/admin/audit-page-component.tsx
```diff
@@ -91,7 +85,7 @@ const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, but
   };
 
   return <React.Fragment>
-    {loadNextSubject && <Navigate to={getNextAuditPath(auditSubject)}/>}
+    {navSubject && <Navigate to={getNextAuditPath(auditSubject)}/>}
```
jaycarlton, line 88: Can we  change `<Navigate/>` to take this in as a property like `enabled` so we don't need the &&? I guess I  was thinking the `Navigate` element would be absorbed into a hook as well, but I don't know that I understood.

ui/src/app/components/app-router.tsx
```diff
@@ -59,3 +59,16 @@ export const Navigate = ({to}): React.ReactElement => {
   const location = useLocation();
   return <Redirect to={{pathname: to, state: {from: location}}}/>;
 };
+
+// Resets the state after change to prevent a navigation loop
```
jaycarlton, line 63: It would help to  have a little more documentation for what this does and doesn't do for you.  Maybe name it  `useGatedNavigation`?

ui/src/app/components/admin/audit-page-component.tsx
petesantos: This is the approach I initially took with the hook and had some type issues with it. I wasn't thrilled with the approach and after doing some more reading it seems that having a hook return a component for render is not recommended as it has negative effects on the state management.

I personally do like the `&&` (JS short-circuiting) approach as it is used in the react world quite a bit. Also I feel it is very explicit about state and what will render. In this case navigation occurs when the node is rendered. I feel it makes the code clear when the state is directly tied to whether the component will render - this pattern would apply to all components. Pushing it down into a component and letting the component decide whether it will render requires a bit more investigation when reading the code at a later date.

That said - I could make it look like this (and I do like the look of this/how it reads):
```javascript
   <Navigate when={nextSubjectRequested} to={getNextAuditPath(auditSubject)}/>
```

Overall it would be nice to have a coding style recommendation in place for this. Having `when` be a property we use for conditional rendering of a component could be nice

ui/src/app/components/app-router.tsx
petesantos: Changed to `useNavigationState` and updated the comment

ui/src/app/components/admin/audit-page-component.tsx
petesantos: Changed to `nextSubjectRequested` - is that clearer?

ui/src/app/components/app-router.tsx
```diff
@@ -59,3 +59,19 @@ export const Navigate = ({to}): React.ReactElement => {
   const location = useLocation();
   return <Redirect to={{pathname: to, state: {from: location}}}/>;
 };
+
+// This is essentially a setState for navigation.
+// When the navigation state is set to true it will trigger a render.
+// At this point the state should be used to render the 'Navigate' component, triggering a route change
+// This hook will automatically set the state back to false, preventing the need for an additional 'useEffect' per nav item
+export const useNavigationState = (): [boolean, Function] => {
```
jaycarlton, line 67: There's really nothing navigation-specific here. Could  this be more of a `useToggle`?

ui/src/app/components/admin/audit-page-component.tsx
```diff
@@ -50,14 +50,8 @@ export interface AuditPageProps {
...
   const [auditSubject, setAuditSubject] = useState(initialAuditSubject);
-  const [loadNextSubject, setLoadNextSubject] = useState(false);
   const [downloadSqlFile, setDownloadSqlFile] = useState(false);
-
-  useEffect(() =>  {
-    if (loadNextSubject) {
-      setLoadNextSubject(false);
-    }
-  }, [loadNextSubject]);
+  const [nextSubjectRequested, setNextSubjectRequested] = useNavigationState();
```
jaycarlton, line 54: Can we just call this `subjectRequested`? The first  one is also a next one, so I don't know that `next`  really signifies anything.

Also,  do we really need this other bit? It seems like we've got a boolean tracking whether another property is non-null and/or changed. Can we not express  what we need with `useEffect` on `auditSubject`?

petesantos: I updated the name

The auditSubject state is needed to update the value in the text box, otherwise it would never change. The boolean state for navigation is more to check whether the user clicked on the "Audit" button. Without that information we have not way of knowing whether the user intended to run a new audit.

The additional state is a different paradigm from our previous navigation - which was basically a straight up callback. This is where react-router's philosophy leaks out a bit. In react-router's view of the world, it wants everything to be a react component, including navigation. In order navigate, we need to render a component.

ui/src/app/components/app-router.tsx
petesantos: I like that a lot better - thanks!

ui/src/app/components/admin/audit-page-component.tsx
```diff
@@ -91,7 +85,7 @@ const UserInput = ({initialAuditSubject, auditSubjectType, getNextAuditPath, but
   };
 
   return <React.Fragment>
-    {loadNextSubject && <Navigate to={getNextAuditPath(auditSubject)}/>}
+    {subjectRequested && <Navigate to={getNextAuditPath(auditSubject)}/>}
```
jaycarlton, line 88: this still  seems iffy somehow, to have to evaluate an external variable instead of having this be a prop on Navigate. Other elements have a `show` property, though if we're using an element to describe a state change, we're already well outside my comfort zone. I want to think of elements as nouns or objects, with possible side effects at render time. But having a new class of elements that are not only stateful but  mutate elements around them seems awkward. So something like  `<MyNavigate go={subjectRequested} url={auditUrl}/>`.

ui/src/app/components/admin/audit-page-component.tsx
jaycarlton: OK, so a purist could have one piece of state to manage/describe the input box's contents, and  update  the next subject when that was committed, but that's no simpler than what you have here.

PR #3843, opened 2020-08-03
[risk=no] Pin Firefox to 78.0 to fix puppeteer

calbach:
The latest Firefox version (79.0) has issues installing via the CircleCI orb. Pin to 78.0 for now.


calbach:
Going to merge since this at least fixes the first issue. I'm not sure if these other breakages are flakes or real - either way I am going to merge so we can address that separately.

PR #3845, opened 2020-08-03
[no ticket][risk=no] Make randomize fast

als364:
Usage:
```
./project.rb randomize-vcf --vcf [vcf location] --number-of-copies [number] --output-path [output path]
```

### DIFF
--- a/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
+++ b/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
@@ -35,9 +35,6 @@ public class RandomizeVcf extends VariantWalker {
       shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME)
   protected File outputVcf;
 
-  //  @Argument(doc = "Sample name suffix.", fullName = "SAMPLE_NAME_SUFFIX", shortName = "S")
-  //  protected String sampleNameSuffix;
-
   @Argument(doc = "Number of samples to generate.", fullName = "NUM_SAMPLES", shortName = "N")
   protected int samples;
 
--- a/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
+++ b/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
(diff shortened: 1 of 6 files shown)


api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -50,9 +51,9 @@ public RandomizeVcf() {
   }
 
   @VisibleForTesting
-  protected RandomizeVcf(String sampleNameSuffix, Random random) {
+  protected RandomizeVcf(int samples, Random random) {
     super();
-    this.sampleNameSuffix = sampleNameSuffix;
```
ericsong, line 55: nit: maybe add a prefix to indicate that its an int? I thought it was a list while reading through

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -142,6 +141,8 @@ protected Genotype randomizeGenotype(VariantContext variantContext, Genotype gen
     double genotypeTypeIndex = random.nextDouble();
     if (genotypeTypeIndex < .0145) {
       // double no-call
+      alleles.add(Allele.NO_CALL);
```
ericsong, line 144: are these `NO_CALL`s fixing a different issue?

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -93,8 +92,8 @@ protected VariantContext randomizeVariant(VariantContext variant) {
     variantContextBuilder.alleles(variant.getAlleles());
 
     List<Genotype> randomizedGenotypes =
-        variant.getGenotypes().stream()
-            .map(genotype -> randomizeGenotype(variant, genotype))
+        IntStream.rangeClosed(1, this.samples)
+            .mapToObj(i -> randomizeGenotype(variant, variant.getGenotype(0), i))
```
ericsong, line 96: I don't understand a lot of the GATK models but I thought this part seemed a little suspicious since we're only ever grabbing the first genotype out of the variant. Is that ok?

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
als364: How about `numberOfSamples`?

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
als364: Yes. All we want to randomize is the alleles. We're starting from a single-sample vcf anyways so there will only ever be one genotype in the variant. I can specify that in the comments for the invocation and add a comment here, if you think that would help.

api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
als364: Yes - `VcfWriter` gets unhappy when some samples have no alleles when there are samples that do have alleles.

PR #3848, opened 2020-08-03
[RW-5341][risk=low] Optionally include demographics in metrics export

calbach:
There is active work to move to a nightly reporting pipeline. For now though, this script is still being run manually to generate metrics. There is a high priority request for this information - however I do not wish to make it part of the standard export, since the nature of the data is sensitive.

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -300,26 +300,26 @@ public class ExportWorkspaceData {
     row.setDegrees(
         Optional.ofNullable(user.getDegreesEnum()).orElse(ImmutableList.of()).stream()
             .map(Degree::toString)
-            .collect(Collectors.joining(",")));
+            .collect(Collectors.joining(",\n")));
     DbDemographicSurvey demo = user.getDemographicSurvey();
     if (includeDemographics && demo != null) {
       row.setRace(
           Optional.ofNullable(demo.getRaceEnum()).orElse(ImmutableList.of()).stream()
               .map(Race::toString)
-              .collect(Collectors.joining(",")));
+              .collect(Collectors.joining(",\n")));
       row.setEthnicity(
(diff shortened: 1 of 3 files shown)


jaycarlton:
Thanks for jumping on this. I probably would've just done an ad hoc query and then merged it inside sheets with some horrendous formulas, or left the join as an exercise for the reader.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -269,6 +297,39 @@ private WorkspaceExportRow toWorkspaceExportRow(DbUser user) {
             .map(dateFormat::format)
             .orElse(""));
     row.setCreatorRegistrationState(user.getDataAccessLevelEnum().toString());
+    row.setDegrees(
+        Optional.ofNullable(user.getDegreesEnum()).orElse(ImmutableList.of()).stream()
+            .map(Degree::toString)
+            .collect(Collectors.joining(",")));
```
ericsong, line 303: nit: the rest of the multi value columns are separated with a newline as well

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -269,6 +297,39 @@ private WorkspaceExportRow toWorkspaceExportRow(DbUser user) {
...
+    DbDemographicSurvey demo = user.getDemographicSurvey();
```
ericsong, line 304: nit: i think you can just move this object into the conditional and drop the null check

calbach: `includeDemographics` is driven by a flag - not null'ness of the demographics value. So this doesn't work - unless I'm misunderstanding the suggestion.

api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
calbach: Thanks, missed that.

PR #3849, opened 2020-08-04
[risk=no] Regenerate gradlew per last version upgrade

calbach:
I missed this in https://github.com/all-of-us/workbench/pull/3659 .

The gradle wrapper files can be generated via `gradle wrapper`. This should be rerun after a gradle version upgrade.


calbach:
I added more detail to the PR description. The gradle version within docker was correct, but I failed to update gradlew properties, so if you used `./gradlew` you were still on the old version.

PR #3853, opened 2020-08-04
[RW-5321][risk=no] updating dao to include sex at birth

freemabd:
updating dao/test cases to include sex at birth


api/src/main/java/org/pmiops/workbench/cdr/dao/CBCriteriaDao.java
```diff
@@ -233,9 +233,9 @@
 
   @Query(
       value =
-          "select * from cb_criteria where domain_id = 'PERSON' and type in ('GENDER', 'RACE', 'ETHNICITY') and parent_id != 0 order by name asc",
+          "select * from cb_criteria where domain_id = 'PERSON' and type in ('GENDER', 'RACE', 'ETHNICITY', 'SEX') and parent_id != 0 order by name asc",
       nativeQuery = true)
-  List<DbCriteria> findGenderRaceEthnicity();
+  List<DbCriteria> findGenderRaceEthnicitySexAtBirth();
```
NehaBroad, line 238: Do you think the name findAllDemographics/findDempgraphics will be better?

freemabd: Done.

PR #3860, opened 2020-08-05
triallelic randomize

als364:
In this array, we don't look for the ref in triallelic sites, we just no-call if it's not one of the alts. So I'm now handling randomizing at triallelic sites differently than biallelic sites. I'm combining the no-call probability with the homref probability and just writing 1/2 and 2/1 where I'd otherwise write 0/1 or 1/0.

### DIFF
--- a/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
+++ b/api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
@@ -140,30 +140,49 @@ public class RandomizeVcf extends VariantWalker {
      */
     List<Allele> alleles = new ArrayList<>();
     double genotypeTypeIndex = random.nextDouble();
-    if (genotypeTypeIndex < .0145) {
-      // double no-call
-      return alleles;
-    } else if (genotypeTypeIndex < .8095) {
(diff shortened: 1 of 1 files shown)


api/genomics/src/main/java/org/pmiops/workbench/genomics/RandomizeVcf.java
```diff
@@ -140,30 +140,49 @@ protected Genotype randomizeGenotype(VariantContext variantContext, Genotype gen
...
+      } else if (genotypeTypeIndex < .9268) {
+        // 1/0 het
+        alleles.add(variantContext.getAlternateAllele(0));
+        alleles.add(variantContext.getReference());
+      } else {
+        // homvar
+        alleles.add(variantContext.getAlternateAllele(0));
+        alleles.add(variantContext.getAlternateAllele(0));
+      }
     } else {
```
ericsong, line 166: nit: I think it would read a little quicker if we specified that the size should be equal to 2. I was trying to figure this out earlier while I was reading.

als364: that makes sense, I'll flip it around!

PR #3866, opened 2020-08-08
[RW-5369][risk=low] Up-gradle to 5.6.3

calbach:
Upgrade gradle to 5.6.3. I ran some migration tools to detect issues, and ran the dev server locally without issues. Surprisingly I only needed one change, which was to add a settings.gradle file.

This is a pre-requisite to using the openapi generator, which is a prerequisite to using the new Open API 3 Leo APIs. https://precisionmedicineinitiative.atlassian.net/browse/RW-4847


api/gradlew.bat
```diff
@@ -1,3 +1,19 @@
+@rem
```
jaycarlton, line 1: If we don't support windows or cygwin, do we still need this file and the cygwin section above?

api/gradlew.bat
```diff
@@ -14,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
jaycarlton, line 33: Interesting. Not where I would've expected these to go.

api/build.gradle
```diff
@@ -881,3 +881,25 @@ compileTestKotlin {
     jvmTarget = "1.8"
   }
 }
+
+// This is a hack to populate the (empty) output directory for the Kotlin
```
calbach, line 885: Please review this. It should resolve the errors I was seeing on Circle.

jaycarlton:
I'd prefer to avoid checking in the batch and cygwin file until we are ready to officially support Windows. It took me a little while to find my bearings when I looked at the gradle stuff last, and I  could see a maintainer getting confused by files we don't use.

api/build.gradle
```diff
@@ -881,3 +881,25 @@ compileTestKotlin {
...
+//
+// Without this workaround, compilation fails with:
+// """
+// A problem was found with the configuration of task ':compileGeneratedJava'.
+// > Directory '/usr/local/google/home/calbach/aou/workbench/api/build/classes/kotlin/generated'
+//   specified for property 'compileGeneratedKotlinOutputClasses' does not exist.
+// """
+task prepareKotlinGeneratedDir() {
+  doLast {
+    mkdir "$projectDir/build/classes/kotlin/generated"
```
jaycarlton, line 902: Looks good. Will this fail if the directory is already there somehow? Does it need to be removed on  gradle clean?

calbach: No, it doesn't fail. I've now declared it as an output, so I think gradle clean will just delete it.

calbach:
I removed gradlew.bat. The cygwin stuff is baked into the autogenerated gradlew file - I don't have any desire to modify this file.

calbach:
I will plan to merge this tomorrow after I start work. Feel free to re-review before then, I made a few small adjustments to handle the memory issues I was seeing.

PR #3868, opened 2020-08-10
[noTicket][risk=low] Fix UI for Research Purpose reminder

NehaBroad:
This PR:

- Fix UI for research purpose to look like the following 
Screen Shot 2020-08-10 at 1 12 51 AM

- Also workspace-gaurd (that checks every route) was making call to workspace-service to get workspace details, replace that with currentWorkspaceStore to improve performance 

---
**PR checklist**

### DIFF
--- a/ui/src/app/guards/workspace-guard.service.ts
+++ b/ui/src/app/guards/workspace-guard.service.ts
@@ -3,24 +3,23 @@ import {
   ActivatedRouteSnapshot, CanActivate, CanActivateChild, Router,
   RouterStateSnapshot
 } from '@angular/router';
-import {serverConfigStore} from 'app/utils/navigation';
-import {WorkspacesService} from 'generated';
+import {currentWorkspaceStore, serverConfigStore} from 'app/utils/navigation';
+import {WorkspaceAccessLevel} from 'generated/fetch';
 import {Observable} from 'rxjs/Observable';
 
 @Injectable()
 export class WorkspaceGuard implements CanActivate, CanActivateChild {
   constructor(
-    private workspaceService: WorkspacesService,
     private router: Router) {}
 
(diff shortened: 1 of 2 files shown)


ui/src/app/guards/workspace-guard.service.ts
```diff
@@ -3,24 +3,24 @@ import {
...
-    private workspaceService: WorkspacesService,
     private router: Router) {}
 
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
     if (serverConfigStore.getValue().enableResearchReviewPrompt && route.routeConfig.path === 'data' ||
         route.routeConfig.path === 'notebooks') {
-      this.workspaceService.getWorkspace(route.params.ns, route.params.wsid).subscribe(workspace => {
-        if (workspace.accessLevel === 'OWNER'
-            && workspace.workspace.researchPurpose.needsReviewPrompt) {
+      currentWorkspaceStore.subscribe(workspace => {
```
calbach, line 18: This subscription is never released, so these callbacks will build up indefinitely.

Here, I would suggest `const workspace = await currentWorkspaceStore.first().toPromise();`

`first()` only emits the first workspace in the store

calbach, line 18: Next: this doesn't work during a context switch from workspace A --> workspace B. It would pickup workspace A, whereas you want B here.

I would include here a filter on the workspaceStore to ensure it actually matches the workspace you are guarding for.

e.g. `await currentWorkspaceStore.filter(/* match the ns / ws id */).first().toPromise();`

calbach, line 18: I suggest using an async function with await here. It seems desirable to block this guard until you have evaluated the workspace. Right now this is a background process. That said, this might not be feasible since I'm not sure where the workspace load gets triggered during the rendering flow. Try it out - if not, the current way is probably OK.

PR #3870, opened 2020-08-11
[risk=no][RW-5314] Script for validating cohort extraction process

als364:
Also present in a workspace Ariel has on test - ping her for access.


als364:
I looked at that doc - it has support for reading basic metadata and for walking through the vcf line-by-line, but not for actually executing bcftools commands.

Yes, I can make it accept a sample name.

PR #3874, opened 2020-08-12
[RW-5322][risk=no] move ds_linking table to cloud sql

freemabd:
This PR moves the ds_linking table into our cloudsql CDR instance. Its more efficient to query this table in cloudsql rather than BQ. The following changes include

1. Updated the CDR indexing scripts to include `ds_linking` table (make-bq-data.sh and make-bq-data-dump.sh)
2. Added `ds_linking` table to our local liquibase build for development purposes
3. Created `DSLinkingDao` and db model `DbDSLinking` with new test case `DSLinkingDaoTest`
4. Updated `DataSetServiceImpl` to use `DSLinkingDao`
5. Bumped all cloudsql CDR versions to allow for using the `ds_linking` table

### DIFF
--- a/api/db-cdr/generate-cdr/bq-schemas/ds_linking.json
+++ b/api/db-cdr/generate-cdr/bq-schemas/ds_linking.json
@@ -1,0 +1,27 @@
+[
+  {
+    "mode": "REQUIRED",
+    "name": "ID",
+    "type": "INTEGER"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "DENORMALIZED_NAME",
+    "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "OMOP_SQL",
+    "type": "STRING"
+  },
+  {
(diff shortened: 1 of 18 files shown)


api/db-cdr/generate-cdr/bq-schemas/ds_linking.json
```diff
@@ -0,0 +1,27 @@
...
+    "name": "OMOP_SQL",
+    "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "JOIN_VALUE",
+    "type": "STRING"
+  },
+  {
+    "mode": "REQUIRED",
```
NehaBroad, line 23: Should the database field have nullable false constraint for domain?

freemabd: domain is what the dataset builder uses to lookup the sql chunks. It is invalid to have a null domain.

PR #3875, opened 2020-08-12
[no ticket][risk=no] Use 24h display for audit event dates

gjuggler:
This caused me quite some confusion when investigating RW-5373 (I though the event was 06:35, it was actually 18:35).

I haven't validated this fix locally, but I'm pretty sure it's correct per the [momentjs docs](https://momentjs.com/docs/#/parsing/string-formats/).


jaycarlton:
good catch.

PR #3876, opened 2020-08-13
[no ticket][risk=no] update mapper packages in CB

freemabd:
created mapper packages for all CB/CS/DS components

- org.pmiops.workbench.cohortbuilder.mapper
- org.pmiops.workbench.cohortreview.mapper
- org.pmiops.workbench.conceptset.mapper
- org.pmiops.workbench.dataset.mapper


api/src/bigquerytest/java/org/pmiops/workbench/api/CohortBuilderControllerBQTest.java
```diff
@@ -33,7 +33,7 @@
 import org.pmiops.workbench.cohortbuilder.CohortBuilderServiceImpl;
 import org.pmiops.workbench.cohortbuilder.CohortQueryBuilder;
 import org.pmiops.workbench.cohortbuilder.SearchGroupItemQueryBuilder;
-import org.pmiops.workbench.cohortbuilder.mappers.CohortBuilderMapperImpl;
+import org.pmiops.workbench.cohortbuilder.mapper.CohortBuilderMapperImpl;
```
NehaBroad, line 36: Question: I know this isnt part of this PR but why do we need Impl class in Test why not just the interface?

freemabd: Because the Impl class needs to be injected to the real service, since this is an end to end integration test

PR #3878, opened 2020-08-13
[RW-5356] [risk=no] Touch-ups to workspace export

gjuggler:
A couple changes that came out of running the metrics export this week.

1) Don't attempt to fetch notebooks on deleted workspaces. This avoids the error spam and probably speeds up runtime a bit.
2) Clean up the logic for identifying non-workspace users. In prior exports, there seems to have been some extra rows at the end, where users who had workspaces were incorrectly being treated as non-workspace users.

---
**PR checklist**

### DIFF
--- a/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
+++ b/api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
@@ -12,9 +12,11 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.function.Function;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import javax.persistence.EntityManagerFactory;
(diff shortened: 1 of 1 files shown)


api/tools/src/main/java/org/pmiops/workbench/tools/ExportWorkspaceData.java
```diff
@@ -180,22 +183,26 @@ public CommandLineRunner run(
...
+          Streams.stream(userDao.findAll())
+              .collect(Collectors.toMap(DbUser::getUserId, Function.identity()));
+
+      Set<Long> userIdsWithoutWorkspaces = usersById.keySet();
 
       log.info("collecting / converting all workspaces");
       for (DbWorkspace workspace : this.workspaceDao.findAll()) {
         rows.add(toWorkspaceExportRow(workspace, includeDemographics));
-        usersWithoutWorkspaces.remove(workspace.getCreator());
+        userIdsWithoutWorkspaces.remove(workspace.getCreator().getUserId());
```
calbach, line 195: Ah, I didn't notice this was `Set<DbUser>`, that would explain the behavioral change. Thanks for catching this.

calbach:
Script has been removed now

PR #3881, opened 2020-08-13
[no ticket][risk=no] Puppeteer Jupyter notebook tests

aweng98:
Adding new notebook tests:
* python-pyplot.spec.ts
* import-python-lib.spec.ts
* import-r-lib.spec.ts
(notebook R and Python code are retrieved from external text files in each test)

Two `import-*.spec.ts` notebook tests are derived from AoU release playbook notebook manual testing.
[Link](https://all-of-us-rw-staging.appspot.com/workspaces/aou-rw-staging-4ced74e4/notebooktesting/notebooks).

Anding new functions to support new test scenarios, and making incremental improvement to existing notebook functions to avoid code clutter and duplication.


e2e/tests/notebook/notebook-actions.spec.ts
```diff
@@ -85,14 +79,11 @@ describe('Jupyter notebook tests', () => {
 
       await workspaceAnalysisPage.duplicateNotebook(notebookName);
 
-      // Verify duplicate notebook exists
+      // Delete notebooks
       const duplNotebookName = `Duplicate of ${notebookName}`;
-      const resourceCard = new DataResourceCard(page);
-      const duplNotebookExists = await resourceCard.cardExists(duplNotebookName, CardType.Notebook);
-      expect(duplNotebookExists).toBe(true);
```
aweng98, line 92: unnecessary check because the dupled notebook is to be deleted.

e2e/tests/notebook/import-r-lib.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // Delete notebook
+    const analysisPage = await notebook.goAnalysisPage();
+    await dataPage.openTab(TabLabelAlias.Analysis);
+    await analysisPage.deleteNotebook(notebookName);
+
+    // UserEmail and Google-bucket is expected to be printed out.
+    const email = config.userEmail
+    expect(outputText).toEqual(expect.stringMatching(email));
+    expect(outputText).toEqual(expect.stringMatching('gs://fc-secure-'))
```
aweng98, line 45: assertions are done after delete notebooks. this is to make sure delete always happen even if assertion fail.
this new pattern is applied in other notebook tests as well.

e2e/app/page/dataset-save-modal.ts
```diff
@@ -71,4 +78,22 @@ export default class DatasetSaveModal extends Modal {
...
+    const previewButton = await Button.findByName(this.page, {name: 'See Code Preview'}, this);
+    await previewButton.click();
+    await waitUntilChanged(this.page, await previewButton.asElementHandle());
+
+    // Find Preview Code
+    const selector = `${this.getXpath()}//textarea[@data-test-id="code-text-box"]`;
+    const previewTextArea = new Textarea(this.page, selector);
+    // Has 'disabled' property
+    await waitForPropertyExists(this.page, selector, 'disabled');
+    const previewText = await previewTextArea.getTextContent();
```
calbach, line 95: nit: don't need this variable, just return

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+  Code = '.code_cell',
+  Markdown = '.text_cell',
+  Any = '',
+}
+
+export default class NotebookCell {
+
+  private readonly page: Page;
+  private iframe: Frame; // Jupyter notebook in iframe
+  private cellIndex: number;
```
calbach, line 15: The constructor indicates this is option, but the type here does not. Which should it be?

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+  private readonly cellType: CellType;
+
+
+  constructor(page: Page, cellType: CellType, cellIndex?: number) {
```
calbach, line 19: I don't think this variable assignment should be needed:

```
constructor(private page: Page, private cellType: CellType, private cellIndex = ..) {}
```

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+      this.iframe = await frame.contentFrame();
+    }
+    return this.iframe;
+  }
+
+  /**
+   * Set focus in notebook cell.
+   * @param {number} indx Cell index number.
+   */
+  async focusCell(indx?: number): Promise<ElementHandle> {
```
calbach, line 37: index

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+  async getIFrame(): Promise<Frame> {
+    if (this.iframe === undefined) {
+      const frame = await this.page.waitForSelector('iframe[src*="notebooks"]');
+      this.iframe = await frame.contentFrame();
+    }
+    return this.iframe;
+  }
+
+  /**
+   * Set focus in notebook cell.
```
calbach, line 34: Does this change this container to point to a different element on the page? Would it be better to just return a different handle to another cell, rather than letting this be mutable?

For example, an issue with this is that you haven't updated `cellType` to reflect the new cell

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * Set focus in last cell.
+   */
+  async focusLastCell(): Promise<ElementHandle> {
+    const selector = `${this.cellSelector()} .CodeMirror-code`;
+    const frame = await this.getIFrame();
+    const cells = await frame.$$(selector);
+    this.cellIndex = cells.length;
```
calbach, line 58: Not -1? Is this a 1-based index? 0-based is probably better.

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  private readonly page: Page;
+  private iframe: Frame; // Jupyter notebook in iframe
+  private cellIndex: number;
+  private readonly cellType: CellType;
+
+
+  constructor(page: Page, cellType: CellType, cellIndex?: number) {
+    this.page = page;
+    this.cellIndex = cellIndex;
```
calbach, line 21: cellIndex should be documented somewhere. For example, it's not clear to me whether this index includes both code and output cells.

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  outputAreaSelector(indx?: number): string {
+    return `${this.cellSelector(indx)} .output_subarea`;
+  }
+
+  /**
+   * Returns css selector for cell.
+   * @param {number} indx Cell index.
+   */
+  cellSelector(indx?: number): string {
```
calbach, line 132: I would probably make some of these private for clarity. I doubt there's a need to use this outside this class.

e2e/app/page/notebook-page.ts
```diff
@@ -28,145 +28,165 @@ export enum KernelStatus {
...
-   * @return {string} Output string.
+   * @param {string} code The code to run.
+   * @param {number} timeOut The timeout time in milliseconds.
+   *
    */
-  async runCodeCell(cellIndex: number, opts: { code?: string, timeOut?: number }): Promise<string> {
-    await this.codeCell.selectCell(cellIndex, opts.code);
-    await this.runButton();
+  async runCodeCell(cellIndex?: number, opts: { code?: string, timeOut?: number } = {}): Promise<string> {
+    const cell = await this.findCell(cellIndex);
```
calbach, line 130: cellIndex is optional. What does this do then?

e2e/resources/python-code/import-libs.txt
```diff
@@ -0,0 +1,23 @@
+import dateutil
```
calbach, line 1: .py and .R seem like more accurate file extensions here

e2e/tests/notebook/import-python-lib.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  beforeEach(async () => {
+    await signIn(page);
+  });
+
+  test('Import os', async () => {
+    let codeSnippet;
+    try {
+      codeSnippet = fs.readFileSync('resources/python-code/import-os.txt', 'utf8');
+    } catch(e) {
+      throw new Error(e);
```
calbach, line 21: Why bother catching this only to immediately rethrow it?

e2e/tests/notebook/import-python-lib.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    const outputText = await notebook.runCodeCell(cellIndex, {code: codeSnippet});
+
+    // Delete notebook
+    const analysisPage = await notebook.goAnalysisPage();
+    await dataPage.openTab(TabLabelAlias.Analysis);
+    await analysisPage.deleteNotebook(notebookName);
+
+    // Partial values check: UserEmail and python3 path is expected to be printed out.
+    const partialArray = [
+      '/usr/local/bin/python3',
```
calbach, line 42: Should we just have a second file for each of these inputs that includes the expected output?

Alternatively, should we have each of these snippets output a specific string on success? e.g. "success", to make this checking more consistent?

e2e/tests/notebook/import-python-lib.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      config.userEmail,
+    ];
+    const outputTextArray = outputText.split(/\n/);
+    expect(outputTextArray).toEqual(expect.arrayContaining(partialArray));
+  });
+
+  test('Import common lib', async () => {
+    let codeSnippet;
```
calbach, line 50: Please move most of this to a helper. This is 95% identical to the other test method

e2e/tests/notebook/import-r-lib.spec.ts
```diff
@@ -0,0 +1,84 @@
+import DataPage, {TabLabelAlias} from 'app/page/data-page';
+import {Language} from 'app/text-labels';
+import * as fs from 'fs';
+import {config} from 'resources/workbench-config';
+import {makeRandomName} from 'utils/str-utils';
+import {findWorkspace, signIn} from 'utils/test-utils';
+import {CellType} from 'app/page/notebook-cell';
+
+// Notebook server start may take a long time. Set maximum test running time to 20 minutes.
+jest.setTimeout(20 * 60 * 1000);
```
calbach, line 10: Won't this affect any test which gets run after this one?

e2e/tests/notebook/import-r-lib.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    await analysisPage.deleteNotebook(notebookName);
+
+    // UserEmail and Google-bucket is expected to be printed out.
+    const email = config.userEmail
+    expect(outputText).toEqual(expect.stringMatching(email));
+    expect(outputText).toEqual(expect.stringMatching('gs://fc-secure-'))
+  });
+
+  test('Import common lib', async () => {
+    let rCodeSnippet;
```
calbach, line 49: Please eliminate duplicate code here as well

e2e/tests/notebook/import-python-lib.spec.ts
calbach: Having a consistent "success" output should still be possible, for example:

```
assert os.getenv('OWNER_EMAIL')
assert os.getenv('WORKSPACE_CDR')
assert os.getenv('WORKSPACE_NAMESPACE')
assert os.getenv('GOOGLE_PROJECT') # same as WORKSPACE_NAMESPACE
assert os.getenv('CLUSTER_NAME')
assert os.getenv('WORKSPACE_BUCKET')

print('success')
```

e2e/tests/notebook/import-python-lib.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    const partialArray = [
+      '/usr/local/bin/python3',
+      config.userEmail,
+    ];
+    const outputTextArray = outputText.split(/\n/);
+    expect(outputTextArray).toEqual(expect.arrayContaining(partialArray));
+  });
+
+  test('Import common lib', async () => {
+    const workspaceCard = await findWorkspace(page);
```
calbach, line 37: I still think there's too much duplication. The only difference between these methods should be the input file. Even the output should be identical if you follow my other suggestion - else the helper can simply return the outputText and just handle the assertions in the test.

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/**
+ * Notebook Cell represents the root element that contains both the code input and output cells.
+ */
+export default class NotebookCell {
+
+  private readonly page: Page;
+  private iframe: Frame; // Jupyter notebook iframe
+  private cellIndex: number; // 1-based cell index because notebook cell prompt index starts from 1.
```
calbach, line 18: To be clear, the numbers show in the Jupyter UI are not cell indices. They correspond to the execution history of the notebook. For example, you can execute the 4th cell before you execute the first one. If you do this, you're notebook will look like:

```
[2] cell0
    cell1
    cell2
[1] cell3
```

When you start the notebook, no cell will have this label. If you always execute the notebook in a linear fashion, then they will look like they are just 1-indexed cell labels, so I can see how this could be confusing to someone who is debugging the test.

Personally, I would still prefer 0-indexed here. This is probably what I would assume the test code would do, rather than yielding to the UI.

e2e/app/page/notebook-cell.ts
```diff
@@ -0,0 +1,177 @@
...
+  private readonly cellType: CellType; // Markdown or Code cell.
+
+  /**
+   *
+   * @param {Page} page Puppeteer Page.
+   * @param {CellType} cellType Cell type: Code or Markdown cell. Default value is Code cell.
+   * @param {number} cellIndex 1-based cell index. Default value is 1.
+   */
+  constructor(page: Page, cellType: CellType = CellType.Code, cellIndex: number = 1) {
```
calbach, line 27: My constructor comment I don't think was addressed: inline private here to avoid boilerplate?

e2e/app/page/notebook-cell.ts
calbach: If I understand your argument, you're saying that you're making it easy for non-coders by using 1-based indexing which they are probably more familiar with than 0-indexing. If so, this is just doing a disservice to them: just use standard coding practices so they can learn the standards. The nth-child CSS selector is an implementation detail that the test writer does not need to concern themselves with.

So, I'm not convinced. But also I don't feel strongly enough to block the review. Fine to merge for now, happy to discuss more if you are not convinced whether this is worth changing in a follow-up.

aweng98:
Great. Thank you for your feedback. I'll have more time to think more about it.

PR #3882, opened 2020-08-14
[RW-5382][risk=no] Streaming BQ upload service

jaycarlton:
As a hedge against potential issues with batch inserts, this implementation gives the "easy way" of streaming to BigQuery. We need to watch out for duplication, but the random insert ID should help. 

Added some timing around forming the snapshot and each stream upload.

Had a heck of a time with timestamp conversions. It seems some places expect different formats on input, and I thought I observed another format coming off the QueryParameterValues yesterday.

---
**PR checklist**


api/src/main/java/org/pmiops/workbench/reporting/ReportingServiceImpl.java
```diff
@@ -1,24 +1,52 @@
...
   @Override
   public ReportingJobResult takeAndUploadSnapshot() {
     final ReportingSnapshot snapshot = reportingSnapshotService.takeSnapshot();
-    return reportingUploadService.uploadSnapshot(snapshot);
+    return getConfiguredUploadService().uploadSnapshot(snapshot);
+  }
+
+  private ReportingUploadService getConfiguredUploadService() {
+    final String uploadMethod = workbenchConfigProvider.get().reporting.uploadMethod;
+    if (UploadMethod.DML.name().equals(uploadMethod)) {
```
jaycarlton, line 43: Sadly we can't do this in the DI, because the config isn't available at bootstrap time. So I'm always injecting dead code unless there's a runtime change to it.

Though perhaps I could manual-wire it instead of using ctor injection. Not really worth that though.

api/src/main/java/org/pmiops/workbench/cohortbuilder/util/QueryParameterValues.java
```diff
@@ -14,6 +17,18 @@
 public final class QueryParameterValues {
   private static final int MICROSECONDS_IN_MILLISECOND = 1000;
 
+  // For creating a Timestamp QueryParameterValue, use this formatter.
+  // example error when using the RowToInsert version (below): "Invalid format:
+  // "1989-02-17 00:00:00.000000" is too short".
+  public static final DateTimeFormatter QPV_TIMESTAMP_FORMATTER =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSSZZ").withZone(ZoneOffset.UTC);
+  // For parsing the string value of a timestamp QPV, use this formatter (with colon).
+  public static final DateTimeFormatter QPV_TIMESTAMP_TO_STRING_FORMATTER =
```
jmthibault79, line 26: this one is not used

api/src/main/java/org/pmiops/workbench/reporting/ReportingJobResult.java
```diff
@@ -2,7 +2,6 @@
 
 public enum ReportingJobResult {
   SUCCEEDED,
-  ABORTED,
```
jmthibault79, line 5: ?

api/src/main/java/org/pmiops/workbench/reporting/ReportingSnapshotServiceImpl.java
```diff
@@ -1,45 +1,68 @@
...
   }
 
   @Override
   public ReportingSnapshot takeSnapshot() {
-    return new ReportingSnapshot()
-        .captureTimestamp(clock.millis())
-        .researchers(getResearchers())
-        .workspaces(getWorkspaces());
+    final Stopwatch stopwatch = stopwatchProvider.get().start();
+    final TransactionTemplate template = new TransactionTemplate(platformTransactionManager);
```
jmthibault79, line 53: Flagging this: needs review by someone more Spring savvy

api/src/main/java/org/pmiops/workbench/reporting/ReportingUploadServiceStreamingImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            researcherRequestBuilder.build(
+                TableId.of(projectId, dataset, "researcher"),
+                reportingSnapshot.getResearchers(),
+                fixedValues),
+            workspaceRequestBuilder.build(
+                TableId.of(projectId, dataset, "workspace"),
+                reportingSnapshot.getWorkspaces(),
+                fixedValues))
+        .stream()
+        .filter(r -> 0 < r.getRows().size())
```
jmthibault79, line 87: nit: could you flip this to size > 0?  I find that more readable

api/src/test/java/org/pmiops/workbench/reporting/ReportingUploadServiceTest.java
```diff
@@ -157,17 +197,54 @@ public void testUploadSnapshot_batchInserts() {
...
+    final List<InsertAllRequest> requests = insertAllRequestCaptor.getAllValues();
+
+    assertThat(requests).hasSize(2);
+
+    final List<RowToInsert> researcherRows = requests.get(0).getRows();
+    assertThat(researcherRows).hasSize(3);
+    assertThat(researcherRows.get(0).getId()).hasLength(16);
+    assertThat(researcherRows.get(0).getContent()).hasSize(RESEARCHER_COLUMN_COUNT + 1);
+
+    final List<RowToInsert> workspace1Rows = requests.get(1).getRows();
```
jmthibault79, line 238: this name is misleading - there is one row per workspace.  This is not Workspace 1's rows.

Please also assert that there are 3 workspaces here.

api/src/test/java/org/pmiops/workbench/reporting/ReportingUploadServiceTest.java
```diff
@@ -127,12 +161,18 @@ public void testUploadSnapshot() {
...
+            (double)
+                timestampQpvToInstant(jobs.get(1).getNamedParameters().get("creation_time__0"))
+                    .toEpochMilli())
+        .isWithin(500.0)
+        .of(THEN.toEpochMilli());
   }
 
   @Test
-  public void testUploadSnapshot_batchInserts() {
+  public void testUploadSnapshot_dmlBatchInserts() {
```
jmthibault79, line 175: should we test the streaming version of this?

api/src/main/java/org/pmiops/workbench/reporting/ReportingJobResult.java
```diff
@@ -2,7 +2,6 @@
 
 public enum ReportingJobResult {
```
freemabd, line 3: Do we use this enum for anything? Recovering from bad states or is that coming in a future PR?

jaycarlton: Not yet. It's either coming out, or will be incorporated into a more sophisticated return type. I haven't decided whether to roll up exceptions and keep going or not. But I have an error handling/correcting story in the backlog.

api/src/test/java/org/pmiops/workbench/reporting/ReportingUploadServiceTest.java
```diff
@@ -157,17 +197,54 @@ public void testUploadSnapshot_batchInserts() {
...
+
+    final QueryParameterValue creationTime =
+        jobs.get(5).getNamedParameters().get("creation_time__0");
+    assertThat(creationTime).isNotNull();
+    final Instant instant = QueryParameterValues.timestampQpvToInstant(creationTime);
+    assertThat((double) instant.toEpochMilli()).isWithin(500.0).of(THEN.toEpochMilli());
+  }
+
+  @Test
+  public void testUploadSnapshot_streaming() {
```
jaycarlton, line 219: @jmthibault79 here is the streaming test. It's actually a bit harder to unit test, because not all of the conversions are documented (which gave me fits).

api/src/main/java/org/pmiops/workbench/reporting/ReportingSnapshotServiceImpl.java
jaycarlton: I expect to revisit it when I do the [perf testing ticket](https://precisionmedicineinitiative.atlassian.net/browse/RW-5196). I think it's going to be important to schedule a couple of these soon after the initial delivery if we can.

Is there a  way to flag it as read-only if I don't use the template stuff?

api/src/main/java/org/pmiops/workbench/reporting/ReportingJobResult.java
jaycarlton: This was just a placeholder. After the alpha delivery, I want to implement an actual error [handling/reporting/mitigating strategy](https://precisionmedicineinitiative.atlassian.net/browse/RW-5384).

api/src/main/java/org/pmiops/workbench/reporting/ReportingServiceImpl.java
```diff
@@ -1,24 +1,52 @@
...
 import org.pmiops.workbench.model.ReportingSnapshot;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Service;
 
 @Service
 public class ReportingServiceImpl implements ReportingService {
 
-  private final ReportingUploadService reportingUploadService;
+  private static enum UploadMethod {
+    DML,
```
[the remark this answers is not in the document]
jaycarlton: Sure. It's data manipulation language, aka an Insert/Update/Drop/Create, or everything except a query. I think BigQuery is the only place I've seen this term.

api/src/main/java/org/pmiops/workbench/reporting/ReportingServiceImpl.java
```diff
@@ -1,24 +1,52 @@
...
   private final ReportingSnapshotService reportingSnapshotService;
+  private final ReportingUploadService reportingUploadServiceDmlImpl;
+  private final ReportingUploadService reportingUploadServiceStreamingImpl;
+  private final Provider<WorkbenchConfig> workbenchConfigProvider;
 
   public ReportingServiceImpl(
-      ReportingUploadService reportingUploadService,
-      ReportingSnapshotService reportingSnapshotService) {
-    this.reportingUploadService = reportingUploadService;
+      @Qualifier("REPORTING_UPLOAD_SERVICE_DML_IMPL")
```
jaycarlton, line 23: Note  I'm testing two different implementations of upload here. All this fanciness is really only necessary because my local DB doesn't have nearly as many rows as the Test DB, but I don't want to get into a one-off deployments. We can also compare test against Perf, for example.

api/src/test/java/org/pmiops/workbench/reporting/ReportingUploadServiceTest.java
jaycarlton: We do, right below. There's no manual batching with the stream approach though. The BQ client does that for us I'm assuming.

api/src/main/java/org/pmiops/workbench/reporting/ReportingSnapshotServiceImpl.java
freemabd: @Transactional(readOnly = true)

api/src/main/java/org/pmiops/workbench/cohortbuilder/util/QueryParameterValues.java
jaycarlton: Yeah, I had a use for it but now I can't find it.

api/src/main/java/org/pmiops/workbench/reporting/ReportingSnapshotServiceImpl.java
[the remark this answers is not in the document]
jaycarlton: It's nice to have an explicit transaction boundary. I haven't yet gotten to  the point where I trust `@Transactional`, but I may get there. I also didn't think I could set it to read only using the annotation.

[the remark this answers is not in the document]
freemabd: I prefer to stay away from programmatic transaction management in most cases as it ties the code to transaction management logic. This seems like a pretty straight forward situation that is perfect for `@Transactional`.

api/src/main/java/org/pmiops/workbench/reporting/ReportingSnapshotServiceImpl.java
```diff
@@ -1,59 +1,100 @@
...
+    stopwatch.stop();
+    log.info(LogFormatters.duration("Conversion to ReportingSnapshot", stopwatch.elapsed()));
+    return result;
   }
 
-  private List<ReportingWorkspace> getWorkspaces() {
+  // Retrieve all the data we need from the MySQL database in a single transaction for
+  // consistency.
+  @Transactional(readOnly = true)
+  protected EntityBundle getApplicationDbData() {
```
freemabd, line 90: Per Spring documentation: 
Method visibility and @Transactional

When using proxies, you should apply the @Transactional annotation only to methods with public visibility. If you do annotate protected, private or package-visible methods with the @Transactional annotation, no error is raised, but the annotated method does not exhibit the configured transactional settings. 

The annotation needs to exist on method called from outside the proxy class. So the `takeSnapshot` method would be a candidate for the annotation.

jaycarlton: Oh ok. I was hoping to avoid the time to convert the entries extending the transaction any longer, but I bet that's negligible in practice.

PR #3895, opened 2020-08-18
[DRAFT] Run puppeteer tests in staging

aweng98:
Changes to enable running Puppeteer tests in staging env.


.circleci/config.yml
```diff
@@ -659,6 +691,11 @@ workflows:
             - api-deps-check
             - api-integration-test
             - ui-unit-test
+      # Not ready
+      # - puppeteer-e2e-test-staging:
+          # filters: *filter_only_release_tags
+          # requires:
+            # - deploy-staging
```
aweng98, line 698: It is not enabled in staging workflow because the automation script that delete workspaces and billing projects are not working in staging.

.circleci/config.yml
```diff
@@ -625,6 +654,9 @@ workflows:
           requires:
             - api-deploy-to-test
             - ui-deploy-to-test
+      - puppeteer-e2e-test-staging
```
aweng98, line 657: Reminder to myself: Remove this call in workflow.

aweng98:
@calbach Can you help review circleci config changes?

e2e/app/page/workspace-edit-page.ts
```diff
@@ -245,6 +245,13 @@ export default class WorkspaceEditPage extends AuthenticatedPage {
     }
   }
 
+  /**
```
als364, line 248: What do these changes have to do with running tests in staging?

e2e/app/element/select.ts
```diff
@@ -23,18 +23,26 @@ export default class Select extends BaseElement {
     super(page, xpath);
   }
 
-  async selectOption(optionValue: string): Promise<string> {
+  async selectOption(value: string): Promise<string> {
```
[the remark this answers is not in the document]
aweng98: The original way of finding select option is relying on the `value`. however, values are different for the Dataset on test and staging. see screenshots below. I had to change it to use displayed texts instead of value.

test env.
Screen Shot 2020-08-18 at 10 30 24 PM

staging env.
Screen Shot 2020-08-20 at 2 15 17 PM

e2e/app/page/workspace-edit-page.ts
aweng98: result of changes in `Select` class.

e2e/tests/cohorts/cohorts-rename.spec.ts
```diff
@@ -43,24 +43,27 @@ describe('User can create, modify, rename and delete Cohort', () => {
     const minAge = 21;
     const maxAge = 95;
     const group1 = cohortBuildPage.findIncludeParticipantsGroup('Group 1');
-    const group1Count = await group1.includeAge(minAge, maxAge);
```
aweng98, line 46: In staging, there is a enabled UI feature flag which is disabled in test.
UI behaviors are different as result. I had to modify this test so it works in both env.

e2e/tests/conceptsets/conceptset-edit.spec.ts
```diff
@@ -98,28 +95,14 @@ describe('Editing and Copying Concept Sets', () => {
...
-      workspace1 = allWorkspaces[0];
-      workspace2 = allWorkspaces[1];
-    } else {
-      // Don't have two. Create two new workspaces.
-      workspace1 = await findWorkspace(page, true);
-      workspace2 = await findWorkspace(page, true);
-    }
-
-    // Note: workspace1 is the Copy to workspace. workspace2 is the Copy from workspace.
+
```
aweng98, line 100: Changes here also for running test consistently in both env. In test env, all existing workspaces were created using Synthetic Dataset 3. In staging env, most of existing workspaces were created on Synthetic Dataset 1. Reusing Synthetic Dataset 1 workspaces are causing problems. So the solution is modify the test code - create new workspaces.

e2e/tests/conceptsets/conceptset-edit.spec.ts
```diff
@@ -26,9 +25,7 @@ describe('Editing and Copying Concept Sets', () => {
    */
   test('Add to existing Concept Set in same workspace', async () => {
     // Create a workspace
-    const workspaceCard = await findWorkspace(page, true);
-    const workspaceName = await workspaceCard.getWorkspaceName();
-    await workspaceCard.clickWorkspaceName();
+    const workspaceName = await findWorkspace(page, true).then(card => card.clickWorkspaceName());
```
aweng98, line 28: unrelated change. while running and inspecting test code, I realized 3 lines of code can be rewritten in one line.

e2e/tests/workspace/workspace-edit.spec.ts
```diff
@@ -20,10 +20,17 @@ describe('Editing workspace thru workspace card ellipsis menu', () => {
    * - Verify Workspace Information in ABOUT tab.
    */
   test('User as OWNER can edit workspace', async () => {
-    const workspaceCard = await findWorkspace(page);
+    const workspaceCard = await findWorkspace(page, true);
```
aweng98, line 23: Avoid finding and using Synthetic Dataset 1 workspaces.

e2e/tests/cohorts/cohorts-rename.spec.ts
aweng98: I believe it was `enableCBAgeTypeOptions`. But I did a quick search now, flag was removed 3 days ago for all env. So I don't know what happens next.

What the feature flag did wass: the count on the age slider is pre-calculated and then the real count is calculated in bigquery when Finish button in Modal is clicked. Pre-calcuated count is different from calculated count.

als364: would `expect(totalCount).toEqual(group1Count)` not then fail with the feature flag enabled?

aweng98: No, test shouldn't fail with or without flag.
Old way was to get the count within Age modal then verify count after close Age modal. New way is get the calculated Group Count after close Age modal, and then compare it with the Total Count. Because only one group, Group Count and Total Count should match.

als364: Gotcha, thanks.

.circleci/config.yml
```diff
@@ -205,12 +205,25 @@ commands:
 
   run_e2e_test:
     description: "Run puppeteer e2e integration tests"
+    parameters:
+      env_name:
+        description: The target environment for run Puppeteer tests. Must be one of "dev", "staging".
+        default: "dev"
```
calbach, line 211: Our test environment is called "test"

.circleci/config.yml
```diff
@@ -205,12 +205,25 @@ commands:
...
+        enum: ["dev", "staging"]
     steps:
       - install_e2e_dependencies
       - run:
-          name: Update Puppeteer test environment variables
+          name: Update test user email env variable
+          command: |
+            case << parameters.env_name >> in
+              staging) echo 'export USER_NAME=$PUPPETEER_TEST_USER_STAGING' >> $BASH_ENV ;;
+              dev) echo 'export USER_NAME=$PUPPETEER_TEST_USER' >> $BASH_ENV ;;
```
calbach, line 221: Would prefer a consistent naming convention for these, e.g.

```
PUPPETEER_USER_STAGING
PUPPETEER_USER_TEST
```

.circleci/config.yml
```diff
@@ -565,21 +579,29 @@ jobs:
...
 
-  # Run Puppeteer e2e UI tests on deployed "test" environment.
+  # Run Puppeteer e2e UI tests on deployed "test" or "staging" environment.
   puppeteer-e2e-test:
+    parameters:
+      env_name:
+        description: The target environment for run Puppeteer tests. Must be one of "dev", "staging".
+        default: "dev"
+        type: enum
+        enum: ["dev", "staging"]
```
calbach, line 589: dev -> test

aweng98:
enum values changed to `"test", "staging"`.
renamed circleci env variables names to:
```
PUPPETEER_USER_STAGING
PUPPETEER_USER_TEST
```

PR #3904, opened 2020-08-21
[RW-5101][risk=low] Reactify notebook list

als364:
@calbach, @ericsong could I get some eyes on this - it is no longer in draft and is blocking the merging of two other router PRs that have been approved

I believe pr [3917](https://github.com/all-of-us/workbench/pull/3917) will resolve the issues seen with switch and protected routes. I have removed `Switch` from the router and pushed the guard check down to the individual routes similar to Ariel's proposal. The child component will only render if the path matches.

Having played with the `Switch` component some more I am inclined to avoid it altogether as the order of the routes would implicitly have meaning - only the first match under a switch would render. If we would like to render multiple components under a tiered path (e.g. workspace/:wsid/notebooks/:notebookId) it may be a better pattern to add the routing under the component that initially renders the workspace rather than in one monolithic routing component.

Ariel's Notes:
-----------------------------------------
Opening as draft cause I'm trying to figure out an elegant way to deal with `AppRoute`s...

React builds components from the outside in. react-router's `Switch` component performs matching in its `render()` function before rendering any of the children. It expects that the children will _be_ `Route`s instead of _evaluating to_ `Route`s. So we need to either 1) force the evaluation, or 2) switch to using bare routes. The reason this is working right now is because of duck typing.

Snippet in question:
```js
  _proto.render = function render() {
    var _this = this;

    return React.createElement(context.Consumer, null, function (context) {
      !context ? process.env.NODE_ENV !== "production" ? invariant(false, "You should not use <Switch> outside a <Router>") : invariant(false) : void 0;
      var location = _this.props.location || context.location;
      var element, match; // We use React.Children.forEach instead of React.Children.toArray().find()
      // here because toArray adds keys to all child elements and we do not want
      // to trigger an unmount/remount for two <Route>s that render the same
      // component at different URLs.

      React.Children.forEach(_this.props.children, function (child) {
        if (match == null && React.isValidElement(child)) {
          element = child;
          var path = child.props.path || child.props.from;
          match = path ? matchPath(location.pathname, _extends({}, child.props, {
            path: path
          })) : context.match;
        }
      });
      return match ? React.cloneElement(element, {
        location: location,
        computedMatch: match
      }) : null;
    });
  };
```

This function naively uses the children. It will not, for instance, do exact matching by default unless an `exact` prop exists on the child, regardless of what component the child is exactly.

---
**PR checklist**


ui/src/app/app-routing.module.ts
```diff
@@ -335,12 +337,12 @@ const routes: Routes = [
             data: {}
           },
           {
-            path: 'workspaceAudit',
+            path: 'workspace-audit',
```
calbach, line 340: Was this intentional? Does not seem related to the PR

ui/src/app/app-routing.module.ts
```diff
@@ -155,28 +156,29 @@ const routes: Routes = [
                     children: [
                       {
                         path: '',
-                        component: NotebookListComponent,
+                        component: AppRouting,
                         data: {
-                          title: 'View Notebooks',
                           breadcrumb: BreadcrumbType.Workspace,
                           helpContentKey: 'notebooks'
```
calbach, line 162: The ticket includes dealing with these two parameters. Do we no longer intend to move these over piecemeal?

petesantos: Yes, I added reviewers and took this out of draft state prematurely - I noticed this as I was typing up some notes in the description. I will take a look at this.

ui/src/app/pages/workspace/workspace-wrapper/component.ts
```diff
@@ -65,6 +67,11 @@ export class WorkspaceWrapperComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit(): void {
+    // This is allows the react-router conversion to utilize the helpContentKey
+    // Once we are fully converted the help sidebar and modals will need to be reworked a bit to eliminate
+    // This Angular code
```
calbach, line 72: nit: This -> this

ui/src/app/components/breadcrumb.tsx
```diff
@@ -179,7 +182,7 @@ export const Breadcrumb = fp.flow(
 
     trail(): Array<BreadcrumbData> {
       return getTrail(
-        this.props.routeConfigData.breadcrumb,
+        this.props.routeConfigData.breadcrumb || this.props.reactRouteData.breadcrumb,
```
calbach, line 185: Confirming: does the `reactRouteData` store get cleared with an empty object when you navigate away from the React routes? Or will this potentially contain stale data from the most recent route? Probably won't be a problem here..

petesantos: This will contain stale data when navigating away. Once a react route is engaged again the data in the store will be completely replaced.

PR #3907, opened 2020-08-24
[RW-4847][risk=low] Use swagger codegen plugin over custom gradle tasks

calbach:
Part 1 of 2 for supporting swagger3

## Why use this plugin?

This plugin supports [simultaneous usage of Swagger 2 and Swagger 3](https://github.com/int128/gradle-swagger-generator-plugin#switch-version-of-swagger-codegen). Despite my best efforts, I couldn't sort out the deps to make our current approach work on Swagger 2 / Swagger 3 simultaneously (tried subproject, imports into the same project), but this plugin seems preferable anyways as it comes with decent documentation and equivalent or better capabilities.

 ## Changes

- This PR should be a no-op, in terms of behavior of the application.
- Fixed an issue (I believe preexisting) which caused the appengineRun task to get cached, which meant that dev-up would not execute if you made no code changes since last run (possibly this is the issue you hit @jmthibault79)
- Made several changes in preparation for adding swagger3:
  - Renamed the codegen ignore file to support two
  - Switched codegen output directory to live under build/ instead of src/generated. This is more standard and allows flexibility to write swagger2 and swagger3 to separate build directories. This also meant we no longer can implicitly depend on the default srcDir for `compileGenerated`, which opaquely defaulted to `src/generated/java`.
  - Made corresponding swagger ignore file changes to account for the above
- Consistency change: swagger yaml file names
- As part of the swagger plugin, validation is now performed ahead of code generation.

## Why Swagger and not OpenAPI generator?

Turns out OpenAPI generator doesn't fully support `oneOf` yet, which is used heavily in the new Leo API. This resulted in generation of invalid Java code.


calbach:
Looks like I lost `default: false` functionality in the codegen somehow. Will look tomorrow.

api/build.gradle
```diff
@@ -1,12 +1,10 @@
...
 import org.pmiops.workbench.tooling.IncrementalHotSwapTask
 
 // Runs before all tasks. Sets up properties and dependencies for the build
 // process itself.
 buildscript {
-  // External propertiges on the default project. Values declared in ext blocks
+  // External properties on the default project. Values declared in ext blocks
   // outside of the buildscsript block aren't usable here.
   ext {
     ELASTICSEARCH_VERSION = '6.8.3'
```
jaycarlton, line 10: aside: I never quite grokked what needs to go here vs. an external properties file.

api/src/main/resources/captcha.yaml
```diff
@@ -3,10 +3,10 @@ info:
   title: Captcha
```
jaycarlton, line 3: why Captcha and not ReCaptcha?

api/src/main/resources/workbench-api.yaml
```diff
@@ -5055,10 +5055,9 @@ definitions:
         type: string
       contactEmailFailure:
         description: >
-          Whether or not contact email could be added to verification list. No longer in use.
+          DEPRECATED. Whether or not contact email could be added to verification list. No longer in use.
```
jaycarlton, line 5058: You should be able to add `deprecated: true` here so that it maps to `@Deprecated` in Java.

api/.swagger2-codegen-ignore
```diff
@@ -22,18 +22,22 @@
...
-settings.gradle
-git_push.sh
-gradle.properties
-src/main/AndroidManifest.xml
-src/main/resources/application.properties
-src/test/**
-gradlew
-gradlew.bat
-gradle/**
+build/swagger2/README.md
```
jaycarlton, line 25: nit: you could just do `**/*.md` in case you want another top-level markdown file.

api/build.gradle
```diff
@@ -69,251 +66,208 @@ apply plugin: 'org.owasp.dependencycheck'
...
+      language = 'java'
+      outputDir = file(swagger2CodegenDir)
+      templateDir = file(swaggerTemplateDir)
+      wipeOutputDir = false
+      rawOptions = rawOptsSwagger2 + [
+        '--model-name-prefix', 'Firecloud'
+      ]
+      additionalProperties = baseJavaProps + [
+        'invokerPackage'   : "${swaggerCodegenPackage}.firecloud",
+        'modelPackage'     : "${swaggerCodegenPackage}.firecloud.model",
```
jaycarlton, line 151: nit: you could declare constants like `MODEL_SUBPACKAGE = 'model'` and `API_SUBPACKAGE` so they're not repeated for each of these. Or even have a little method for generating all the additional package properties.

api/.swagger2-codegen-ignore
calbach: This is an ignore file for the output of Swagger codegen, it doesn't relate to the files that we check in to the repository. I also don't see a reason to make changes in this PR since only the base path changed.

api/src/main/resources/workbench-api.yaml
```diff
@@ -4199,7 +4199,6 @@ definitions:
         type: string
       approved:
         type: boolean
-        default:
```
calbach, line 4202: These were the cause of the test breakages - it was resulting in a default value of `false`. I suspect the Swagger plugin may convert to JSON before passing to the codegen CLI, which would explain why the resulting code was different. I didn't see a good reason to keep these lines, though perhaps I could have also done `default: null`, if you want to leave them around.

api/build.gradle
```diff
@@ -86,7 +86,45 @@ configurations {
   }
 }
 
-def swagger2CodegenDir = "$buildDir/swagger2"
+// Shared variables for Swagger codegen. Use "ext" rather than "def" to allow
+// access globally, including from within helper functions.
+ext.swaggerCodegenBasePkg = 'org.pmiops.workbench'
```
jaycarlton, line 91: Very  nice. If they're effectively constants, maybe use UPPER_SNAKE_CASE.

api/build.gradle
```diff
@@ -86,7 +86,45 @@ configurations {
...
+      outputDir = file(swagger2CodegenDir)
+      templateDir = file(swaggerTemplateDir)
+      // Since the output directory is shared among these swagger tasks, wiping
+      // the directory would result in task invalidation / churn.
+      wipeOutputDir = false
+      rawOptions = swagger2RawOpts + rawOptExtras
+      additionalProperties = [
+        'library'          : 'okhttp-gson',
+        'serializableModel': 'true',
+        'dateLibrary'      : 'java8',
```
jaycarlton, line 120: Is this a change? I  thought  it was defaulting to `joda` today for datetimes, at least in  our own  API models. `java8` is the recommended  replacement for joda, but there would be a bit of fixup  I'd think.

calbach: It's not a change, see the previous value above: https://github.com/all-of-us/workbench/pull/3907/files#diff-dd8fd6d45c4e2eccf0f278e121d714d8L150

api/build.gradle
calbach: sure, done

api/src/main/resources/workbench-api.yaml
jaycarlton: tl;dr, if you really want 3-state boolean semantics (🤮 ), then you need `default: null`.

I put in a PR a while back to default all booleans explicitly because the null default made them all tri-state, and since we don't  have proper constructors, we kept having  issues. There are only a couple of places where we want null booleans. (I'm planning on writing about this in one of the Java best practices guides: IMO if you need null booleans, you need an enum). It's also a painful mapping to typescript.

So the new convention I proposed was to put a default value for all booleans in Swagger files, and this new issue seems to me to be an even stronger argument  for it.

calbach: I tried `default: null` as well. It doesn't work with the plugin, it has the same behavior as `default: `, which both appear to be interpreted as default: false

jaycarlton: That's definitely a regression then. That, or we were relying on undocumented behavior, which I don't believe we were.

calbach: What is the regression? Having to omit the default rather than stating `default: null` would be a change, but it seems pretty insignificant and non-functional to me. If you're concerned about documentation, I can replace these instances with a comment.

jaycarlton: I mean a boolean with no stated default changing causing generated classes to initialize members to `false` instead of `null` seems like a potentially (silently) breaking change.

calbach: I'm not too surprised to see undefined behavior out of `default: null` and especially `default: `. I'm probably not going to spend more time investigating this issue unless you have clear documentation that Java code gen in supposed to respond to `default: null` in a specific way. Even then, it's probably a idiosyncrasy with how the plugin works - best I can do is file a bug, this issue is not severe enough to back out of this approach.

To be clear, this has nothing to do with Swagger 3. Everything is still on Swagger 2. The change here is to use a plugin instead of our own bespoke codegen task.

PR #3913, opened 2020-08-25
[risk=low][RW-4956] enable free credit limit override UX

jmthibault79:
Description:

Default case (no free tier credit limit override has been set yet) -> default values of 300-800 in dropdown
no-override

Existing override = 400 -> value of 400 is preselected, using the same default values
over-400

Existing override is not one of the default values -> this value is preselected and added to the dropdown choices, sorted appropriately. Examples for $234.50 and $1500
override-234 50
over-1500

---
**PR checklist**

### DIFF
--- a/ui/src/app/pages/admin/admin-user.tsx
+++ b/ui/src/app/pages/admin/admin-user.tsx
@@ -90,7 +90,6 @@ const ToggleWithLabelAndToggledText = ({label, initialValue, disabled, onToggle,
   </FlexColumn>;
 };
 
-
 const EmailValidationErrorMessage = ({emailValidationResponse, updatedProfile, verifiedInstitutionOptions}) => {
   if (updatedProfile && updatedProfile.verifiedInstitutionalAffiliation) {
     if (emailValidationResponse.isValidMember) {
@@ -111,6 +110,31 @@ const EmailValidationErrorMessage = ({emailValidationResponse, updatedProfile, v
     }
   }
   return null;
+};
+
+const FreeCreditsUsage = ({aboveLimit, usage}) => {
+  const inputStyle = aboveLimit ?
+  {...styles.textInput,
+    backgroundColor: colorWithWhiteness(colors.danger, .95),
+    borderColor: colors.danger,
+    color: colors.danger,
+  } :
+  {...styles.textInput,
+    ...styles.backgroundColorDark,
+    borderColor: colors.disabled,
+    color: colors.disabled};
+
+  return <React.Fragment>
+    <TextInputWithLabel
+      labelText={'Free credits used'}
+      value={usage}
+      inputId={'freeTierUsage'}
+      disabled={true}
+      inputStyle={inputStyle}
+      containerStyle={styles.textInputContainer}
+    />
+    {aboveLimit && <div style={{color: colors.danger}}>Update free credit limit</div>}
+  </React.Fragment>;
 };
 
 interface Props {
@@ -220,6 +244,16 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
     // gotcha: JS sorts numbers lexicographically by default
     const numericallySorted = defaultsPlusMaybeOverride.sort((a, b) => a - b);
     return fp.map((limit) => ({label: renderUSD(limit), value: limit}), numericallySorted);
+  }
+
+  getFreeCreditUsage(): string {
+    const {updatedProfile: {freeTierDollarQuota, freeTierUsage}} = this.state;
+    return `${renderUSD(freeTierUsage)} used of ${renderUSD(freeTierDollarQuota)} limit`;
+  }
+
+  usageIsAboveLimit(): boolean {
+    const {updatedProfile: {freeTierDollarQuota, freeTierUsage}} = this.state;
+    return freeTierDollarQuota < freeTierUsage;
   }
 
   async getInstitutions() {
@@ -501,13 +535,9 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
                 containerStyle={styles.textInputContainer}
                 onChange={email => this.setContactEmail(email)}
             />
-            <TextInputWithLabel
-                labelText={'Free credits used'}
-                placeholder={renderUSD(updatedProfile.freeTierUsage)}
-                inputId={'freeTierUsage'}
-                disabled={true}
-                inputStyle={{width: '6.5rem', ...styles.backgroundColorDark}}
-                containerStyle={styles.textInputContainer}
+            <FreeCreditsUsage
+              aboveLimit={this.usageIsAboveLimit()}
+              usage={this.getFreeCreditUsage()}
             />
           </FlexColumn>
           <FlexColumn style={{width: '33%'}}>
--- a/ui/src/app/pages/admin/admin-user.tsx
+++ b/ui/src/app/pages/admin/admin-user.tsx
@@ -217,7 +217,9 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
   getFreeCreditLimitOptions() {
     const {oldProfile: {freeTierDollarQuota}} = this.state;
     const defaultsPlusMaybeOverride = Array.from(freeCreditLimitDefaultOptions.add(freeTierDollarQuota));
-    return fp.map((limit) => ({label: renderUSD(limit), value: limit}), defaultsPlusMaybeOverride.sort());
+    // gotcha: JS sorts numbers lexicographically by default
+    const numericallySorted = defaultsPlusMaybeOverride.sort((a, b) => a - b);
+    return fp.map((limit) => ({label: renderUSD(limit), value: limit}), numericallySorted);
   }
 
   async getInstitutions() {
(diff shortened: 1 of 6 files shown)


ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -56,19 +57,7 @@ const styles = reactStyles({
...
-  {label: '$450', value: 450},
-  {label: '$500', value: 500},
-  {label: '$550', value: 550},
-  {label: '$600', value: 600},
-  {label: '$650', value: 650},
-  {label: '$700', value: 700},
-  {label: '$750', value: 750},
-  {label: '$800', value: 800}
-];
+const freeCreditLimitDefaultOptions = new Set([300, 350, 400, 450, 500, 550, 600, 650, 700, 750, 800]);
```
jmthibault79, line 60: can't hardcode the options list because we also need to include the current override value, if one exists and is not one of these

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -225,6 +214,12 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
     }
   }
 
+  getFreeCreditLimitOptions() {
+    const {oldProfile: {freeTierDollarQuota}} = this.state;
+    const defaultsPlusMaybeOverride = Array.from(freeCreditLimitDefaultOptions.add(freeTierDollarQuota));
```
jmthibault79, line 219: tested with quota:
* undefined
* equal to a predefined value: does not duplicate
* defined, with various values in and around the predefined range: orders correctly

ui/src/app/pages/profile/profile-page.tsx
```diff
@@ -577,8 +577,8 @@ export const ProfilePage = withUserProfile()(class extends React.Component<
                     <div>Remaining <i>All of Us</i> free credits:</div>
                 </FlexColumn>
                 <FlexColumn style={{alignItems: 'flex-end', marginLeft: '1.0rem'}}>
-                    <div style={{marginTop: '0.4rem'}}>{renderUSD(profile.freeTierUsage)}</div>
```
jmthibault79, line 580: renderUSD() is strictly a string processing function now.  The formatting is desired only here, so it was moved here.

ui/src/app/pages/admin/admin-user.tsx
jaycarlton: nit: consider [fp.range](https://lodash.com/docs/4.17.15#range).

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -124,6 +112,31 @@ const EmailValidationErrorMessage = ({emailValidationResponse, updatedProfile, v
...
+    color: colors.danger,
+  } :
+  {...styles.textInput,
+    ...styles.backgroundColorDark,
+    color: colors.disabled,
+  };
+
+  return <React.Fragment>
+    <TextInputWithLabel
+      labelText={'Free credits used'}
```
jaycarlton, line 129: nit: maybe indicate in the help text or hover  text that  this applies to  all workspaces owned by this user.

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -124,6 +112,31 @@ const EmailValidationErrorMessage = ({emailValidationResponse, updatedProfile, v
   return null;
 };
 
+const FreeCreditsUsage = ({aboveLimit, usage}) => {
```
jaycarlton, line 115: nit: type these? I don't know what rule we settled on for when we explicitly type stuff.

ui/src/app/pages/admin/admin-user.tsx
jaycarlton: I'm not sure if we want to tolerate undefined values or just assume we'll set whatever isn't there to `null`. IOW I never understood how `undefined` is a feature of JavaScript.

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -500,23 +535,18 @@ const AdminUser = withUrlParams()(class extends React.Component<Props, State> {
...
             />
-            <TextInputWithLabel
-                labelText={'Free credits used'}
-                placeholder={updatedProfile.freeTierUsage}
-                inputId={'freeTierUsage'}
-                disabled={true}
-                inputStyle={{width: '6.5rem', ...styles.backgroundColorDark}}
-                containerStyle={styles.textInputContainer}
+            <FreeCreditsUsage
+              aboveLimit={this.usageIsAboveLimit()}
```
jaycarlton, line 539: nit: I'd name this `isAboveLimit`. Generally booleans starting with forms of "be" are more recognizable IMO.

ui/src/app/utils/index.tsx
```diff
@@ -531,12 +531,12 @@ export function highlightSearchTerm(searchTerm: string, stringToHighlight: strin
 
 // render a float value as US currency, rounded to cents: 255.372793 -> $255.37
 // negative values are rendered as $0
-export function renderUSD(value: number) {
+export function renderUSD(value: number): string {
```
jaycarlton, line 534: `render` implies a side effect. I'd call this `formatCreditUsage`. (The negative to zero conversion indicates to me that maybe we shouldn't put it in a public utils file, as that's not standard, and would be an easy way to generate bugs.)

Is there a specific requirement to drop the cents when it's zero? Why not when it's `$1`? Or can we just round to the nearest dollar?

ui/src/app/utils/institutions.tsx
```diff
@@ -44,6 +44,10 @@ export const getRoleOptions = (institutions: Array<PublicInstitutionDetails>, in
     const {shortName} = institution;
     return shortName === institutionShortName;
   }, institutions);
+  if (matchedInstitution === null || matchedInstitution === undefined) {
```
jaycarlton, line 47: direct comparison to undefined makes me nervous. Is everything getting initialized correctly?

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: I attempted to, unsuccessfully.  My assumption was that it was not possible.  But now Googling tells me otherwise, so I will make an attempt.

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: Here I was imprecise: I do not mean JS `undefined`.  Rather, the DB user has `free_tier_credits_limit_dollars_override` set to NULL, which is the state for almost all users.

This results in `freeTierDollarQuota` here being equal to the config-file default value, which may or may not be one of the predefined dropdown options (300 to 800) but in current practice is 300.

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: I considered this, but was surprisingly unsuccessful: https://github.com/all-of-us/workbench/commits/joel/RW-4956-range

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: typed now, with a `FreeCreditsProps`

ui/src/app/utils/index.tsx
jmthibault79: Renamed away from "render" and with a name to indicate that this is specifically for free credit values (usage, balance, quota/override).

The requirement is to use the string "$0" for negative values.

ui/src/app/pages/admin/admin-user.tsx
```diff
@@ -124,6 +112,36 @@ const EmailValidationErrorMessage = ({emailValidationResponse, updatedProfile, v
...
+    color: colors.danger,
+  } :
+  {...styles.textInput,
+    ...styles.backgroundColorDark,
+    color: colors.disabled,
+  };
+
+  return <React.Fragment>
+    <TextInputWithLabel
+      labelText={'Free credits used'}
```
jaycarlton, line 134: nit: don't think you need the curly braces for a string literal with no interpolation.

ui/src/app/utils/index.tsx
```diff
@@ -531,12 +531,12 @@ export function highlightSearchTerm(searchTerm: string, stringToHighlight: strin
 
 // render a float value as US currency, rounded to cents: 255.372793 -> $255.37
 // negative values are rendered as $0
-export function renderUSD(value: number) {
+export function formatFreeCreditsUSD(value: number): string {
   value = value || 0.0;
   if (value < 0.0) {
-    return <div style={{fontWeight: 600}}>$0</div>;
+    return '$0';
```
jaycarlton, line 537: nit: if the value is exactly 0, it will print as `$0.00` but if negative it's `$0`?

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: Fixed, after Greg's surprising suggestion to randomly rearrange the params until it works.  Apparently this is a "fun quirk" of fp-lodash.  Added comment to explain.

ui/src/app/pages/admin/admin-user.tsx
jmthibault79: yep

ui/src/app/utils/index.tsx
jmthibault79: switching to `<=` makes sense

ui/src/app/utils/institutions.tsx
jmthibault79: This is a check of a call to `fp.find()` which returns undefined if it is not found.  I removed the null check since it's not relevant.

PR #3914, opened 2020-08-25
[RW-5324][risk=no] Move ConceptSetMapper into ConceptSetService

freemabd:
This PR finalizes separation of logic between ConceptSetController and ConceptSetService

- ConceptSetController only has access to services and only throws request level exceptions(BadRequestException)
- ConceptSetService has all daos and mappers and only throws service level/db level exceptions

Also added `ConceptSetMapper#dbModelToDbModel` for copy/clone of ConceptSets. Had to add an inner class `ConceptSetContext` to `ConceptSetMapper` to work around a problem with Mapstruct that throws an error: The types of `@Context` parameters must be unique. This was caused by having two `Timestamp` objects as parameters to the `@afterMapping` method.

### DIFF
--- a/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
+++ b/api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
@@ -4,21 +4,20 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Ordering;
 import java.sql.Timestamp;
 import java.time.Clock;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.inject.Provider;
-import javax.persistence.OptimisticLockException;
+import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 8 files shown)


api/src/main/java/org/pmiops/workbench/api/ConceptSetsController.java
```diff
@@ -221,18 +126,15 @@ private void addConceptsToSet(DbConceptSet dbConceptSet, List<Long> addedIds) {
...
-      throw new ConflictException("Failed due to concurrent concept set modification");
-    }
+    // Fail fast if request isn't valid
+    validateUpdateConceptSetConcepts(request);
+
+    workspaceService.getWorkspaceEnforceAccessLevelAndSetCdrVersion(
+        workspaceNamespace, workspaceId, WorkspaceAccessLevel.WRITER);
+
+    return ResponseEntity.ok(
+        conceptSetService.toHydratedConcepts(
```
NehaBroad, line 136: Since both at line 120 and here toHydratedConcept is being called after updateConeptSet, how about calling   **toHydratedConcept** inside an conceptSetService.updateConceptSet? Or maybe since service is taking the UpdateConceptSetRequest should we make the service return 'ConceptSet' where ever possible like update/create?

freemabd: Good idea. Making toHydratedConcept a private method in ConceptSetService.

PR #3916, opened 2020-08-25
[RW-5312][risk=no] Add genomics microarray field to cdrVersion

ericsong:
Description:
Adds a new field to the CdrVersion table and model.

I did not add it to the API model since it's not being used anywhere and many of the CdrVersion fields are not exposed anyways. Also, part of the A/C of the ticket was to add a new CdrVersion that contains a reference to the microarray dataset but I decided to postpone that until the dataset actually exists.

---
**PR checklist**


calbach:
`but I decided to postpone that until the dataset actually exists.`  can the 500 sample dataset not be used here?

RE passing to client: note that we will need to use this soon, for https://precisionmedicineinitiative.atlassian.net/browse/RW-5411 , but the work could be deferred until then.

ericsong:
@calbach sounds good, I added both of those

api/config/cdr_versions_local.json
```diff
@@ -24,5 +24,19 @@
     "numParticipants": 946237,
     "cdrDbName": "cdr",
     "elasticIndexBaseName": "cdr"
+  },
+  {
+    "cdrVersionId": 3,
+    "isDefault": false,
+    "name": "Synthetic Dataset v1.1",
```
calbach, line 31: nit: maybe "Synthetic Dataset v1 (with microarray)"

api/config/cdr_versions_local.json
```diff
@@ -24,5 +24,19 @@
...
+    "dataAccessLevel": 1,
+    "bigqueryProject": "fc-aou-cdr-synth-test",
+    "bigqueryDataset": "SR2019q4r3",
+    "microarrayBigqueryDataset": "microarray_data",
+    "creationTime": "2019-08-26 00:09:51Z",
```
calbach, line 36: nit: 2020 

(this doesn't really matter)

api/src/main/resources/workbench-api.yaml
```diff
@@ -3604,6 +3604,8 @@ definitions:
         "$ref": "#/definitions/DataAccessLevel"
       archivalStatus:
         "$ref": "#/definitions/ArchivalStatus"
+      microarrayBigqueryDataset:
```
calbach, line 3607: Now that I'm looking at this, the client won't really be able to do much with it (they don't know the BQ project). The only thing the client needs to use this for is to make a determination as to which features should be shown in a given workspace.

WDYT about changing this to be a boolean `hasMicroarrayData`?

api/src/main/java/org/pmiops/workbench/cdr/CdrVersionMapper.java
```diff
@@ -0,0 +1,18 @@
...
+import org.mapstruct.Mapping;
+import org.pmiops.workbench.db.model.DbCdrVersion;
+import org.pmiops.workbench.model.CdrVersion;
+import org.pmiops.workbench.utils.mappers.CommonMappers;
+import org.pmiops.workbench.utils.mappers.MapStructConfig;
+
+@Mapper(
+    config = MapStructConfig.class,
+    uses = {CommonMappers.class})
+public interface CdrVersionMapper {
```
calbach, line 13: Also - nice, thanks for adding the mapper

api/src/main/resources/workbench-api.yaml
ericsong: sounds good. I'll just make the change for the API model

api/src/main/java/org/pmiops/workbench/cdr/CdrVersionMapper.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Mapping(source = "microarrayBigqueryDataset", target = "hasMicroarrayData")
+  @Mapping(source = "dataAccessLevelEnum", target = "dataAccessLevel")
+  @Mapping(source = "archivalStatusEnum", target = "archivalStatus")
+  CdrVersion dbModelToClient(DbCdrVersion db);
+
+  default boolean isNonNull(String s) {
+    return s != null;
```
calbach, line 21: nit: null or empty might be slightly less error-prone. I'm worried about someone putting `microarrayBigQueryDataset: ""` in the JSON.

PR #3917, opened 2020-08-25
[risk=no][RW-5446] Correct issues with ProtectedRoutes routing component

petesantos:
Description:

1. Added the featured workspaces back in to the react based router

2. Fix the way `ProtectedRoutes` handles child routes and guards. Guards were being applied at a higher level in the component hierarchy causing issues like unintended redirects and routes not getting hit if they were not in the "right" location in the routing component. Guards are now applied at the individual "route level" component rather than at the parent. Only if the route matches and its render function is hit will a guard be checked. At that point if the guard fails it will redirect the user.

3. The issue with "next sibling" components of the protected route has been addressed by removing the react-router `Switch` component. When no path is provided to a child of a `Switch` it will match any path. Since `ProtectedRoutes` do not have a path, that component would always be considered a match, even if it did not render any children. Removing the `Switch` will allow all paths to be checked, not stopping at the first match.

The `Switch` can be useful if a component page wants to handle its own routing - but we do not need it at the top level.

---
**PR checklist**

### DIFF
--- a/ui/src/app/app-routing.tsx
+++ b/ui/src/app/app-routing.tsx
@@ -106,12 +106,11 @@ export const AppRoutingComponent: React.FunctionComponent<RoutingProps> = ({onSi
       */}
       <ProtectedRoutes guards={[registrationGuard]}>
         <AppRoute
-          path='/library'
+        path='/library'
           component={() => <WorkspaceLibraryPage routeData={{title: 'Workspace Library', minimizeChrome: false}}/>}
         />
       </ProtectedRoutes>
     </ProtectedRoutes>
-
   </AppRouter>;
 };
 
--- a/ui/src/app/app-routing.tsx
+++ b/ui/src/app/app-routing.tsx
@@ -99,14 +99,9 @@ export const AppRoutingComponent: React.FunctionComponent<RoutingProps> = ({onSi
       />
       <AppRoute path='/nih-callback' component={() => <HomepagePage routeData={{title: 'Homepage'}}/>} />
 
-      {/* Protected routes must be nested in order to work. This is a result of the way react-router handles paths
-          react-router expects all children under one of its top level components (the Switch component) to have a 'path' prop
-          The 'ProtectedRoutes' component does not currently support a path prop - though this may change.
-          The resultant behavior is that any routes that are "next siblings" of the top level ProtectedRoutes will not be seen.
-      */}
       <ProtectedRoutes guards={[registrationGuard]}>
         <AppRoute
-        path='/library'
+          path='/library'
           component={() => <WorkspaceLibraryPage routeData={{title: 'Workspace Library', minimizeChrome: false}}/>}
         />
       </ProtectedRoutes>
--- a/ui/src/app/components/app-router.tsx
+++ b/ui/src/app/components/app-router.tsx
@@ -38,21 +38,31 @@ const NavRedirect = ({path}) => {
   return null;
 };
 
-export const AppRoute = ({path, data = {}, component: Component}): React.ReactElement => {
+export const AppRoute = ({path, data = {}, guards = [], component: Component}): React.ReactElement => {
   const routeParams = useParams();
   const routeHistory = useHistory();
+  const { redirectPath = null } = fp.find(({allowed}) => !allowed(), guards) || {};
 
-  return <Route exact={true} path={path} >
-    <Component urlParams={routeParams} routeHistory={routeHistory} routeConfig={data}/>
+  return <Route exact={true} path={path} render={
+    () => {
+      return redirectPath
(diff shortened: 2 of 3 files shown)


ui/src/app/components/app-router.tsx
```diff
@@ -38,21 +38,36 @@ const NavRedirect = ({path}) => {
...
 
 export const ProtectedRoutes = (
   {guards, children}: {guards: Guard[], children: any }): React.ReactElement => {
-  const { redirectPath = null } = fp.find(({allowed}) => !allowed(), guards) || {};
 
-  return redirectPath ? <NavRedirect path={redirectPath}/>
-  : <Fragment>{children}</Fragment>;
+  // Pass the guards to the individual routes. Be sure not to overwrite any existing guards
+  const guardedChildren = fp.flow(
+    fp.flatten,
```
calbach, line 60: What's the purpose of putting `children` in an array, only to immediately flatten them?

ui/src/app/components/app-router.tsx
```diff
@@ -38,21 +38,36 @@ const NavRedirect = ({path}) => {
...
+    fp.toPairs,
+    fp.map(
+      ([key, element]: [string, React.ReactElement]) => {
+        const {guards: elementGuards = []} = element.props;
+        return React.cloneElement(element, {key, guards: [...elementGuards, ...guards]});
+      }
+    )
+  )([children]);
```
calbach, line 68: I gather children is a map of keys -> react elements. I'm a bit confused though, since the previous approach rendered this directly, and your new approach appears to instead render an array of elements. Does React just support either?

ui/src/app/components/app-router.tsx
```diff
@@ -38,21 +38,36 @@ const NavRedirect = ({path}) => {
...
-  return redirectPath ? <NavRedirect path={redirectPath}/>
-  : <Fragment>{children}</Fragment>;
+  // Pass the guards to the individual routes. Be sure not to overwrite any existing guards
+  const guardedChildren = fp.flow(
+    fp.flatten,
+    fp.toPairs,
+    fp.map(
+      ([key, element]: [string, React.ReactElement]) => {
+        const {guards: elementGuards = []} = element.props;
+        return React.cloneElement(element, {key, guards: [...elementGuards, ...guards]});
```
calbach, line 65: This is the opposite order of guard execution than what I would expect, i.e. was expecting `[...guards, ...elementGuards]`, which should track with increasing specificity.

ui/src/app/app-routing.tsx
```diff
@@ -23,11 +24,10 @@ const signInGuard: Guard = {
   redirectPath: '/login'
 };
 
-// This will be added back in after the protectedRoutes fix
-// const registrationGuard: Guard = {
-//   allowed: (): boolean => hasRegisteredAccess(profileStore.get().profile.dataAccessLevel),
-//   redirectPath: '/'
-// };
+const registrationGuard: Guard = {
+  allowed: (): boolean => hasRegisteredAccess(profileStore.get().profile.dataAccessLevel),
```
calbach, line 28: Existing issue, but we're not treating this as state, right? so there is nothing to cause a route to be rerendered as a result of a registration status change, for example?

I think this issue is purely hypothetical for now, since Angular routing should cause this to be reevaluated on every navigate anyways.

ui/src/app/components/app-router.tsx
petesantos: If a component has a single react child it will not create an array of children - it will be the individual item. If there are multiple children they will already be in an array. I want to ensure we are always dealing with an array of children, I flatten in case children is already an array, and it ends up getting nested.

ui/src/app/components/app-router.tsx
petesantos: `children` can be a react element or an array of react elements. React does support arrays of elements.

ui/src/app/components/app-router.tsx
petesantos: fixed

ui/src/app/app-routing.tsx
petesantos: That is correct. The guards are based on stores at the moment. We could use a hook in the top level router to subscribe to the store and re-render on update.

ui/src/app/components/app-router.tsx
petesantos: I updated the type for `children` to help clarify this in the code and I added a comment.
And to expand on my previous comment - children can be many different types in react, but for the purposes of ProtectedRoutes - it should be a react element or array of elements. I could make it more specifically an `AppRoute` / array of routes. But I would like to leave that as an exercise for a future PR to work on my Typescript.

PR #3919, opened 2020-08-25
[RW-4847][risk=low] Support Swagger 3 codegen, partially migrate notebooks.yaml

calbach:
Part 2 of 2 for supporting Swagger3 (follow-up to https://github.com/all-of-us/workbench/pull/3907).

## Changes

- This PR should be a no-op in terms of application behavior
- Support Swagger 3 codegen, in parallel with swagger 2
- Fork and rebase our custom swagger2 templates onto the Swagger 3 templates
  - Note: some changes to the files and file locations were needed to support the template renderer
- Import the new Leo OAS3 yaml with some noted modifications
- Migrate our server-side usage of the Cluster APIs to the new spec, removing overlapping sections of `notebooks.yaml`
- Continues to use the "Cluster" APIs, this PR does not yet switch us to the "Runtime" APIs
- Fixed status check API to actually use the proper baseUrl for the current environment

## Follow-up work

- Investigate typescript codegen for Swagger 3; currently client code still depends on some cluster/notebook APIs.
- Switch to the new "Runtime" API variants
- Model prefixes appear to be broken for Swagger3 codegen, would be nice to use this for Leo's APIs.
- Deal with oneOfs in the Leo spec
- Consider better distinguishing between Leo and Jupyter/Welder APIs.


api/src/main/resources/swagger3codegen/aouTracedMethodCall.mustache
```diff
@@ -0,0 +1,34 @@
+        // Generate an OpenCensus tracing span for this API method call.
```
calbach, line 1: @gjuggler FYI, I have spawned an additional set of templates here to support swagger 3. The patch was relatively straightforward to apply against the swagger3 codegen - these templates didn't change much.

api/src/main/java/org/pmiops/workbench/notebooks/NotebooksConfig.java
```diff
@@ -20,8 +20,13 @@
 public class NotebooksConfig {
   public static final String USER_CLUSTER_API = "userClusterApi";
   public static final String SERVICE_CLUSTER_API = "svcClusterApi";
+
+  // Identifiers for the older Swagger2 APIs for Jupyter and Welder.
```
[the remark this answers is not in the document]
calbach: Yep, clarified comment.

api/build.gradle
```diff
@@ -299,15 +339,18 @@ classes.dependsOn 'generatedClasses'
 project.tasks.ideaModule.dependsOn(generateSwaggerCode)
 
 // Java plugin's Generated Source Set gives us the compileGeneratedJava task.
-// We need to provide it with the swagger-codegen tasks that output Java.
+// We need to provide it with the swagger-codegen tasks that output Java; this
```
calbach, line 342: RE generateSwaggerCode: see new comment here

api/src/main/resources/notebooks.yaml
```diff
@@ -1,7 +1,11 @@
 # This file is a hand-crafted combination of Leo and its proxy APIs.
-#  - All of Leo: https://github.com/broadinstitute/leonardo/blob/develop/src/main/resources/swagger/api-docs.yaml
 #  - A subset of Welder (modified for cluster/path prefix): https://github.com/DataBiosphere/welder/blob/master/server/src/main/resources/api-docs.yaml
 #  - A subset of Jupyter (modified for cluster/path prefix): https://github.com/jupyter/notebook/blob/master/notebook/services/api/api.yaml
+#
+# TODO(calbach): Remove the following once the client-side codegen can support
```
calbach, line 5: RE notebooks.yaml: good catch, fixed.

api/src/main/java/org/pmiops/workbench/notebooks/LeonardoNotebooksClientImpl.java
```diff
@@ -214,15 +222,15 @@ public void localize(String googleProject, String clusterName, Map<String, Strin
...
+    return notebooksRetryHandler.run(
         (context) -> notebooksApi.welderCreateStorageLink(googleProject, clusterName, storageLink));
   }
 
   @Override
   public boolean getNotebooksStatus() {
     try {
-      new StatusApi().getSystemStatus();
-    } catch (ApiException e) {
+      serviceInfoApiProvider.get().getSystemStatus();
```
calbach, line 232: APparently this status API was exclusively checking prod... nice... Fixed that, which should resolve the integration test issues.

